PR #4, opened 2016-07-17
Modify Rx tests to work with real asynchronous observables

cypressious:


dzharkov:
Thanks for your contribution!

Issue #7, opened 2016-08-29
yieldAll and recursive generators


ilya-g:
Investigate whether it's possible to implement efficient tail-recursive generators with `yieldAll` suspension function and the nested iterators technique, described here: https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/specsharp-iterators.pdf

```
fun fromTo(from: Int, to: Int) = generate {
    if (from > to) return
    yield(from)
    yieldAll(fromTo(from + 1, to))
}
```

h0tk3y:
I'm currently investigating this. As to the tail-recursive generators, seems like we can only achieve that using `handleResult()`, which requires a coroutine to return the tail sequence:
```
fun fromTo(from: Int, to: Int) = generateWithTail f@{
    if (from > to) return@f emptySequence()
    yield(from)
    return@f fromTo(from + 1, to)
}
```
As far as I understand, there's no other way to learn whether a nested iterator is the last one yielded in the block, is there? Probably we could use some coroutine metadata here, something that will indicate that a continuation is actually a NOP that will immediately finish the coroutine block. Or is there already a way to learn that?

elizarov:
`buildSequence` is not a part of stdlib with both `yield` and `yieldAll` support provided out-of-the-box.

ilya-g:
Note that the mentioned nested iterator optimization is not implemented.

elizarov:
It actually _is_ available with suspending functions in _even nicer way_:

```
suspend tailrec fun SequenceBuilder.fromTo(from: Int, to: Int) {
    if (from > to) return
    yield(from)
    fromTo(from + 1, to)
}

then do `buildSequence { fromTo(1, 1000) }`

PR #8, opened 2016-10-24
yieldAll through nested iterators

h0tk3y:
Partial solution for #7 -- implementation of `yieldAll` and recursive generators through [SpecSharp nested iterators](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/specsharp-iterators.pdf).

Considerations:
- Needs performance measurement against original implementation (and, probably, extraction into separate API if this turns out to be considerably slower)
- Increased allocation overhead by one object per iterator: `RootIterator` is allocated when iterating over the generated sequence (negligible?)
- Probably some optimizations are possible that simplify the iterator behavior in cases when there were no `yieldAll` calls up to some point
- Tail recursion can be optimized, but not in a transparent way. One option is to add separate function `generateWithTail` (or something) that would take a coroutine returning `Sequence<T>`
  
  ```
  fun fromTo(from: Int, to: Int): Sequence<Int> = generateWithTail f@{
     if (from > to) return@f emptySequence()
     yield(from)
     return@f fromTo(from + 1, to)
  }
  ```
  
  Another option is to request a coroutines API that would allow one to find out that a `Continuation<Unit>` is basically a NOP that will do nothing but immediately finish the coroutine block. Once we know this about continuations we can easily optimize tail recursion.

### DIFF
--- a/kotlinx-coroutines-generate/src/main/kotlin/generate.kt
+++ b/kotlinx-coroutines-generate/src/main/kotlin/generate.kt
@@ -3,86 +3,6 @@ package kotlinx.coroutines
 import java.util.*
 
 /**
- * Manages [NestedIterator]s, arranging them in a stack and switching between them as they produce next
- * nested iterators or finish.
- * The concept is taken from
- * [this article](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/specsharp-iterators.pdf).
- */
-private class RootIterator<T> internal constructor(iterator: Iterator<T>) : AbstractIterator<T>() {
-    private val stack = Stack<Iterator<T>>().apply { push(iterator) }
-
-    override fun computeNext() {
-        while (true) {
-            if (stack.isEmpty()) {
-                done()
-                return
-            }
-            val i = stack.peek()
-            if (!i.hasNext()) {
-                stack.pop()
-            } else {
-                if (i is NestedIterator<T> && i.nextNestedIterator != null) {
-                    stack.push(i.nextNestedIterator)
-                    i.next() //state transfer to NotReady
-                } else {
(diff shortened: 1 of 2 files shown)


h0tk3y:
Unfortunately, this implementations seems to be about two times slower than original coroutines. Benchmark: [(link)](https://github.com/h0tk3y/kotlinx.collections.experimental), see the benchmarks suffixed by `n`. Any optimization suggestions?

My results:

```
Benchmark                                  Mode  Cnt     Score    Error  Units
SequenceCoroutineOperations.filter_ai      avgt   40   712.893 ±  8.580  us/op
SequenceCoroutineOperations.filter_c       avgt   40  1035.318 ± 11.017  us/op
SequenceCoroutineOperations.filter_ci      avgt   40   965.898 ± 25.101  us/op
SequenceCoroutineOperations.filter_cin     avgt   40  1811.378 ± 56.751  us/op
SequenceCoroutineOperations.filter_cn      avgt   40  1940.069 ± 44.974  us/op
SequenceCoroutineOperations.filter_std     avgt   40   511.533 ±  4.832  us/op
SequenceCoroutineOperations.filter_std_ai  avgt   40   761.549 ±  2.216  us/op
SequenceCoroutineOperations.map            avgt   40  1071.858 ± 12.418  us/op
SequenceCoroutineOperations.map_c          avgt   40  2590.374 ± 28.339  us/op
SequenceCoroutineOperations.map_ci         avgt   40  3086.386 ± 15.886  us/op
SequenceCoroutineOperations.map_cin        avgt   40  3497.331 ± 44.871  us/op
SequenceCoroutineOperations.map_cn         avgt   40  4274.558 ± 53.390  us/op
```

PR #13, opened 2016-12-21
asyncRx suspending iterator

elizarov:


elizarov:
Will remove a spurious modification to asyncRx.kt file

PR #20, opened 2017-01-17
Changed comment because I think it was backwards.

bdavisx:


elizarov:
Thank for PR. We've decided to drop examples from readme completely for now. There will be a separate tutorial in the future.

PR #36, opened 2017-03-14
Add Mutex.withMutex

fvasco:


fvasco:
Fixed in e73899d

fvasco:
Fixed https://youtrack.jetbrains.com/issue/KT-16448

PR #38, opened 2017-03-15
HandlerContext: check if the current thread is associated with the correct looper

cypressious:
if yes, run the block directly, if no, post it using the handler

### DIFF
--- a/ui/kotlinx-coroutines-android/src/main/kotlin/kotlinx/coroutines/experimental/android/HandlerContext.kt
+++ b/ui/kotlinx-coroutines-android/src/main/kotlin/kotlinx/coroutines/experimental/android/HandlerContext.kt
@@ -45,7 +45,11 @@ public class HandlerContext(
     private val name: String? = null
 ) : CoroutineDispatcher(), Delay {
     override fun dispatch(context: CoroutineContext, block: Runnable) {
-        handler.post(block)
+        if (Looper.myLooper() == handler.looper) {
+            block.run()
+        } else {
+            handler.post(block)
+        }
     }
 
     override fun scheduleResumeAfterDelay(time: Long, unit: TimeUnit, continuation: CancellableContinuation<Unit>) {


elizarov:
This is a tricky decision to make for every kind of dispatcher. The general rule is that UI dispatchers should not override `isDispatchNeeded`, but leave a default implementation that returns true. To understand the rationale beyond this recommendation, consider the following code:

```
fun asyncUpdateUI() = async(UI) {
    // do something here that updates something in UI
}
```

When you invoke `asyncUpdateUI` in some background thread, it immediately continues to the next line, while UI update happens asynchronously in the UI thread. However, if you invoke it in the UI thread itself, it updates UI synchronously if your `isDispatchNeeded` is overridden with a thread check. Checking if we are already in the UI thread seems more efficient (and it might indeed save a few CPU cycles), but this subtle and context-sensitive difference in behavior makes the resulting async code harder to debug.

Basically, the choice here is between “JS-style” asynchronous approach (async actions are always postponed to be executed later in the even dispatch thread) and “C#-style” approach (async actions are executed in the invoker thread until the first suspension point). While, C# approach seems to be more efficient, it ends up with recommendations like “use yield if you need to ….”. This is error-prone. JS-style approach is more consistent and does not require programmers to think about whether they need to yield or not.

cypressious:
The question is, which use case is more common: Starting `launch(UI)`/`async(UI)` from the main thread or from a background thread? For me, it would be the main thread when I want to subscribe to some network observables and then set some UI state. In this case, I would be surprised that my code doesn't run immediately.

In fact, calling `handler.post()` if you're already on the main thread can lead to subtle bugs, e.g. see https://medium.com/square-corner-blog/a-journey-on-the-android-main-thread-lifecycle-bits-d916bc1ee6b2#.f4z28arlg. As a counter argument, you could however say that the `job` should be canceled if the activity is being destroyed so the `post`ed code shouldn't run at all.

elizarov:
I really believe that we should heed the example of JS when making decisions about async programming in UI, because they've been doing it for years. All they have in JS is a single thread and they have to be cooperative to work with it. They had no questions on whether it is invoked from UI thread or not. It is always the UI thread in JS. 

JS community went though multiple models. There used to be multiple JS-promise libraries both "synchronous" (that would invoke your action directly in your thread if it is ready) and "asynchronous" (that would post it to be executed later). Asynchronous won and became the universal standard, because synchronous is subtly harder to reason about.

I don't think that lifecycle on Android gives any say in this discussion. You have to be life-cycle-aware on Android in anyway.

elizarov:
Let me also mention the stack size, that is somewhat imprtant on Android, too. If you don't dispatch, then you are at risk of "convoying" between multiple communicating coroutines. One coroutine sends something, resumes the other coroutine, but that one checked that it is already in the right thread, so it does it stuff in the same stack, does something else that resumes another coroutine, etc. The stack grows. 

If you always dispatch with `handler.post()`, then you always have a clean stack that is bound only by your code, instead of your code times then number of coroutines you have (worst case)

cypressious:
Ok, since I'm not too bothered by it, I'll accept the explanation and close the PR. We can come back to that topic again, if problems arise.

PR #45, opened 2017-03-24
Add support for Rx2 Maybe.

konrad-kaminski:
#44

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/src/main/kotlin/kotlinx/coroutines/experimental/rx2/RxAwait.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/main/kotlin/kotlinx/coroutines/experimental/rx2/RxAwait.kt
@@ -51,14 +51,8 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
  * If the [Job] of the current coroutine is completed while this suspending function is waiting, this function
  * immediately resumes with [CancellationException].
  */
-public suspend fun <T> MaybeSource<T>.await(): T? = suspendCancellableCoroutine { cont ->
-    subscribe(object : MaybeObserver<T> {
-        override fun onSubscribe(d: Disposable) { cont.disposeOnCompletion(d) }
-        override fun onComplete() { cont.resume(null) }
-        override fun onSuccess(t: T) { cont.resume(t) }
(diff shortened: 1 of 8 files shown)


elizarov:
Great work! I really appreciate it. I have just a few comments and question. I'll put them next to the corresponding sources.

reactive/kotlinx-coroutines-rx2/src/main/kotlin/kotlinx/coroutines/experimental/rx2/RxAwait.kt
```diff
@@ -40,6 +40,44 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
...
+/**
+ * Awaits for completion of the maybe without blocking a thread.
+ * Returns the resulting value, null if no value was produced or throws the corresponding exception if this
+ * maybe had produced error.
+ *
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is completed while this suspending function is waiting, this function
+ * immediately resumes with [CancellationException].
+ */
+public suspend fun <T> MaybeSource<T>.await(): T? = suspendCancellableCoroutine { cont ->
```
konrad-kaminski: You're completely right. I also added a test for `Maybe.empty().await() == null`.

reactive/kotlinx-coroutines-rx2/src/main/kotlin/kotlinx/coroutines/experimental/rx2/RxChannel.kt
```diff
@@ -36,6 +38,15 @@ public interface SubscriptionReceiveChannel<out T> : ReceiveChannel<T>, Closeabl
 }
 
 /**
+ * Subscribes to this [MaybeSource] and returns a channel to receive elements emitted by it.
+ */
+public fun <T> MaybeSource<T>.open(): SubscriptionReceiveChannel<T> {
```
konrad-kaminski: I was thinking about cases where you'd like to do sth with a value emitted from `Maybe` (or do nothing if it is empty). Then you could do sth like

```kotlin
maybe.consumeEach {
  doSth(it)
}
```

But then again you could probably do the same thing with

```kotlin
maybe.await()?.also {
  doSth(it)
}
```
The former code looks nicer and more explicit to me.

Issue #35, opened 2017-03-08
Add Repo-Info to Tutorial


s1monw1:
Hey Guys,

would make sense to add the repository information to your coroutines tutorial, wouldn't it?
"maven {
        url "http://dl.bintray.com/kotlin/kotlin-eap-1.1"
}" is necessary to find 
"dependencies {
    compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:0.12"
}"

@see http://kotlinlang.org/docs/tutorials/coroutines-basic-jvm.html

Thanks and kind regards,
Simon

ilya-g:
Thanks, we updated the current version of readme with the actual repository information.

PR #46, opened 2017-03-25
fix readme typo

sreich:
rm extraneous backticks in gradle code


ilya-g:
I have already changed that block to point to jcenter in #35, thanks.

PR #65, opened 2017-06-01
Basic CoroutineExceptionHandler implementation.

konrad-kaminski:
Currently there is no `CoroutineExceptionHandler` implementation. It'd be useful to have even a simple one.

With this PR it is possible to create a basic `CoroutineExceptionHandler` with this simple code:

```kotlin
val handler = CoroutineExceptionHandler { _, ex ->
    println("Got exception: ${ex.message}")
}

launch(CommonPool + handler) {
    // do sth
    // the exception thrown here will be handled by the println above
}
```

See also `CoroutineExceptionHandlerTest`.

### DIFF
--- a/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/CoroutineExceptionHandler.kt
+++ b/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/CoroutineExceptionHandler.kt
@@ -16,6 +16,7 @@
 
 package kotlinx.coroutines.experimental
 
+import kotlin.coroutines.experimental.AbstractCoroutineContextElement
 import kotlin.coroutines.experimental.CoroutineContext
 
 
@@ -51,7 +52,17 @@ public interface CoroutineExceptionHandler : CoroutineContext.Element {
     /**
      * Key for [CoroutineExceptionHandler] instance in the coroutine context.
      */
-    companion object Key : CoroutineContext.Key<CoroutineExceptionHandler>
+    companion object Key : CoroutineContext.Key<CoroutineExceptionHandler> {
+        /**
+         * Creates new [CoroutineExceptionHandler] instance.
+         * @param handler a function which handles exception thrown by a coroutine
+         */
+        public operator inline fun invoke(crossinline handler: (CoroutineContext, Throwable) -> Unit): CoroutineExceptionHandler =
+            object: AbstractCoroutineContextElement(Key), CoroutineExceptionHandler {
+                override fun handleException(context: CoroutineContext, exception: Throwable) =
+                    handler.invoke(context, exception)
+            }
+    }
 
     /**
      * Handles uncaught [exception] in the given [context]. It is invoked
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Doesn't handling exception in `CoroutineExceptionHandler` prevent job from completion/cancellation if we simply `println` for example? This seems to be a dangerous feature, isn't it?

cypressious:
@cy6erGn0m isn't that what the name CoroutineException**Handler** implies?

kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/CoroutineExceptionHandler.kt
```diff
@@ -51,7 +52,17 @@ public interface CoroutineExceptionHandler : CoroutineContext.Element {
...
      * Key for [CoroutineExceptionHandler] instance in the coroutine context.
      */
-    companion object Key : CoroutineContext.Key<CoroutineExceptionHandler>
+    companion object Key : CoroutineContext.Key<CoroutineExceptionHandler> {
+        /**
+         * Creates new [CoroutineExceptionHandler] instance.
+         * @param handler a function which handles exception thrown by a coroutine
+         */
+        public operator fun invoke(handler: (CoroutineContext, Throwable) -> Unit): CoroutineExceptionHandler =
+            object: AbstractCoroutineContextElement(Key), CoroutineExceptionHandler {
```
cypressious: Can't you make the function `inline`?

cy6erGn0m:
@cypressious I just wanted to say that people generally don't need to use their own exception handlers (except the case when adopt coroutines to frameworks and libraries) but use job's completion listeners, am I right, @elizarov ?

kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/CoroutineExceptionHandler.kt
konrad-kaminski: Why would you want to do that?

cypressious: Save an allocation. In the current form you'll have an allocation for the lambda and one for the anonymous object. With `inline` you would save one allocation for the lambda.

konrad-kaminski: What I meant was is this kind of optimisation really needed? How often will you create such exception handlers in real-life systems?

cypressious: Potentially on every `launch`. It is not unthinkable to start a coroutine on every frame in a UI application.

konrad-kaminski: I made it `inline`, though I can't imagine a scenario where I'd need to create a new context with a separate exception handler over and over again.

elizarov:
It might be more convenient to handle exception with *CoroutineExceptionHandler* than with the regular `try { ... } catch(e: Exception) { ... }`, even in your local code where you have some one-off needs (like you need to restart a coroutine if it fails for any reason).

elizarov:
I've merged it into `develop`, though rewrote a test w/o `CountDownLatch`

PR #68, opened 2017-06-09
Fix error that fail to dispose handler when job is in SINGLE/SINGLE+ state

uchuhimo:
When disposing a handler, the corresponding job node will be removed from job's state.

When job is in SINGLE/SINGLE+ state, we compare state with the job node to ensure we remove the correct job node. However, the origin code compares state with the job itself, which is always false, causing the job node still stays in job's state.

This pull request aims to fix the bug described above. The corresponding unit tests are also added (`testDisposeSingleHandler` will fail before fixed).


elizarov:
Great catch. Thanks a lot. I've merged it into `develop` branch for the next release.

Issue #79, opened 2017-07-05
IO thread pool for blocking calls


elizarov:
We need some kind of `IO` dispatcher in `kotlinx.coroutines` that will be optimized for offloading of blocking calls and will be backed by an unbounded thread-pool that is smart enough to create new threads only when needed and to shut them down on timeout. The goal is that you could always wrap blocking IO call in `withContext(IO) { ... }` (**UPDATED**: it was formerly named `run(IO)`) and be sure that you will not run into the problem of having not enough threads in your pool. However, it will be up to the user to control the maximal number of concurrent operations via some other means. 

It will be somewhat conceptually similar to `newThread()` scheduler in Rx, but it will be also designed to be used where `io()` scheduler in Rx is used.

fvasco:
Java 6 has `Executors.newCachedThreadPool`, it is easy to create and easy to customize:

    public static ExecutorService newCachedThreadPool() {
        return new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                                      60L, TimeUnit.SECONDS,
                                      new SynchronousQueue<Runnable>());
    }

We can tune `maximumPoolSize` at startup to limit "the maximum allowed number of threads" throught a System property.

However we should consider that specific use case should create a fine tuned IO executor without using the global one.

elizarov:
We need to make sure that there is no confusion in difference between `run(IO) { /* blocking code here */ }` and `runBlocking { ... }`.

fvasco:
Kotlin can provide an explicit annotation (like @DslMarker) to avoid invocation from coroutine.

So we can rename the function to

    @Blocking
    fun runAndWait( ... )

fvasco:
Using the code:

    val job = ...
    run(IO + job) { ... }
    job.cancel()

I/O thread should be interrupted?

elizarov:
We should not do interruption by default. It is dangerous. But we definitely need some easy way to adapt an interruptible blocking API (when we are sure it plays well with interrupts) to coroutines. I don't know yet what is going to be the best solution for this. This is also related to #57

elizarov:
@fvasco W.r.t to `@Blocking` annotation it is actually a larger issue that spans both Kotlin and Java. Here is a related issue in YT: https://youtrack.jetbrains.com/issue/KT-15525

voddan:
[...]

PR #83, opened 2017-07-12
Support for blocking calls inside coroutines (#79)

fvasco:
Added `blocking` and `blockingAsync` suspending functions for blocking code invocation inside a coroutine.

### DIFF
--- a/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Builders.kt
+++ b/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Builders.kt
@@ -16,6 +16,7 @@
 
 package kotlinx.coroutines.experimental
 
+import java.util.concurrent.Executor
 import java.util.concurrent.locks.LockSupport
 import kotlin.coroutines.experimental.*
 import kotlin.coroutines.experimental.intrinsics.COROUTINE_SUSPENDED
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Deferred.kt
```diff
@@ -161,6 +162,21 @@ public fun <T> async(context: CoroutineContext, start: Boolean, block: suspend C
...
+ *
+ * @param executor the executor for blocking code
+ * @param start start option, [CoroutineStart.UNDISPATCHED] is invalid
+ * @param block the blocking code
+ */
+public fun <T> blockingAsync(executor: Executor = DefaultBlockingExecutor,
+                             start: CoroutineStart = CoroutineStart.DEFAULT,
+                             block: () -> T): Deferred<T> {
+    require(start != CoroutineStart.UNDISPATCHED) { "Start blocking code undispatched is not supported" }
+    return async(executor.asCoroutineDispatcher(), start) { block() }
```
gildor: Maybe we should cache dispatcher created from DefaultBlockingExecutor to avoid dispatcher creation on each call of the function and it would be helpful to have DefaultBlockingExecutor dispatcher (BlockingPool?) available as object same way as we have CommonPool.

Also, I'm not sure that it's good, that we don't allow to pass custom context here because context can be used not only for dispatchers.

kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Builders.kt
```diff
@@ -25,6 +26,24 @@ import kotlin.coroutines.experimental.intrinsics.suspendCoroutineOrReturn
 // --------------- basic coroutine builders ---------------
 
 /**
+ * Suspend current coroutine and execute the [block] blocking code in [executor].
+ * The current coroutine is resumed after [block] execution.
+ *
+ * @param executor the executor for blocking code
+ * @param block the blocking code
+ */
+suspend fun <T> blocking(executor: Executor = DefaultBlockingExecutor, block: () -> T) =
```
gildor: I think blocking should be covered by tests as well.

fvasco:
Hi @gildor,
I agree with you.

I wish receive some feedbacks about the future of this patch before continue the work.

elizarov:
I'm sorry for dragging my feet for so long here. I did not saw it at as improvement versus `run(myContext) { ... }` for blocking calls and was thinking that we'd just provide an `IO` thread pool to solve it as explained in #79, which I did not like either, since providing too many contexts and switching back and forth between them is not really a user-friendly idea, until it clicked how we can have `blocking` and a single scheduler without any switching.

PR #91, opened 2017-07-26
'owner' support for Mutex.withLock

fvasco:


elizarov:
You should leave the old, one-parameter, version as `@Deprecated(..., level=DeprecationLevel.HIDDEN)` for the purpose for preserving backwards binary compatibility with the code that was compiled against an older version of the library.

Issue #90, opened 2017-07-26
Mutex fairness policy


fvasco:
The current `Mutex` implementation is always fair or not?
Documentation should explicit it.

If current implementation isn't fair then is it possible to add an optional parameter to define a fairness policy?

elizarov:
It is fair.

elizarov:
Released in version 0.18

TYZRPVX:
lock & unlock function make use of `LockFreeLinkedListNode` to ensure FIFO order

PR #95, opened 2017-07-31
Mutex fairness documentation (#90)

fvasco:
> The mutex created is fair: lock is granted in sequential order to the applicants.


kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/sync/Mutex.kt
```diff
@@ -100,6 +100,8 @@ public interface Mutex {
 
 /**
  * Creates new [Mutex] instance.
+ * The mutex created is fair: lock is granted in sequential order to the applicants.
```
elizarov: The technical name of this guarantee is "first come, first served" (FCFS), so the better way of saying it is that "lock is granted in first come, first served order"

PR #96, opened 2017-08-02
reactor.core.Cancellation is removed from project reactor from 3.1.0.M1

yxf07:
reactor.core.Cancellation is removed from project reactor from 3.1.0.M1

kotlinx.coroutines.experimental.reactor.mono will throw ClassNotFoundException

so in the next versions , we need to use Disposable as replacement and set method from setCancellation to onDispose


elizarov:
I've merged it into `develop` branch. I also had to upgrade Reactor's version to Aluminium-SR3

elizarov:
Released in version 0.18

Issue #103, opened 2017-08-20
ART crash with coroutines in obfuscated code


Miha-x64:
Hello. Thanks for your work!

I've got [a sample coroutine code in a tiny Android project](https://github.com/Miha-x64/Advanced-Kotlin-Patterns/blob/master/app/src/main/java/net/aquadc/advancedkotlinpatterns/feature/coroutines/CoroutinesFragment.kt#L51).

It works great, but obfuscated and optimized code crashes ART (I use Nexus 5x with Android 7.1.1). Surprisingly, `-useuniqueclassmembernames` ProGuard flag fixes the issue: [my aggressive proguard-rules.pro](https://github.com/Miha-x64/Advanced-Kotlin-Patterns/blob/master/app/proguard-rules.pro#L47).

Stack trace:
```
art/runtime/class_linker.cc:5452] Check failed: super_class->IsResolved() 
art/runtime/runtime.cc:422] Runtime aborting...
art/runtime/runtime.cc:422] Aborting thread:
art/runtime/runtime.cc:422] "main" prio=10 tid=1 Runnable
art/runtime/runtime.cc:422]   | group="" sCount=0 dsCount=0 obj=0x75600a68 self=0x7ce6095a00
art/runtime/runtime.cc:422]   | sysTid=10811 nice=-10 cgrp=default sched=0/0 handle=0x7cea043a98
art/runtime/runtime.cc:422]   | state=R schedstat=( 143266706 22457552 211 ) utm=7 stm=6 core=4 HZ=100
art/runtime/runtime.cc:422]   | stack=0x7fd84e6000-0x7fd84e8000 stackSize=8MB
art/runtime/runtime.cc:422]   | held mutexes= "abort lock" "mutator lock"(shared held)
art/runtime/runtime.cc:422]   native: #00 pc 000000000047e090  /system/lib64/libart.so (_ZN3art15DumpNativeStackERNSt3__113basic_ostreamIcNS0_11char_traitsIcEEEEiP12BacktraceMapPKcPNS_9ArtMethodEPv+220)
art/runtime/runtime.cc:422]   native: #01 pc 000000000047e08c  /system/lib64/libart.so (_ZN3art15DumpNativeStackERNSt3__113basic_ostreamIcNS0_11char_traitsIcEEEEiP12BacktraceMapPKcPNS_9ArtMethodEPv+216)
[...]

PR #107, opened 2017-08-31
ProGuard config

Miha-x64:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/103,
suggested by ProGuard team here: https://sourceforge.net/p/proguard/bugs/660/

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,3 +97,14 @@ buildscript {
     ext.kotlin_version = '1.1.4'
 }
 ```
+
+### ProGuard
+
+In obfuscated code, fields with different types can have the same names,
+and `AtomicReferenceFieldUpdater` may be unable to find the correct ones.
+To avoid field overloading by type during obfuscation, add this to your config:
+```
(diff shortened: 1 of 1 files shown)


cypressious:
If the `kotlinx-coroutines-android` project were an AAR artifact, you could directly bundle the rules.

Miha-x64:
@cypressious I don't think it's very useful.
Also, bad ProGuard rules lead to such hacks:
image

JakeWharton:
This is way too broad. You only want to prevent their renaming while currently it will also prevent their removal when unused.

Miha-x64:
@JakeWharton 
> [-keepclassmembernames class_specification](https://www.guardsquare.com/en/proguard/manual/usage#keepclassmembernames)
> Specifies class members whose names are to be preserved, if they aren't removed in the shrinking phase.

JakeWharton:
Perfect. Thought it said keepclassmembers

PR #204, opened 2018-01-11
Update coroutines-guide-ui.md

Vixb1122:
perform correct naive computation of Fibonacci numbers


elizarov:
I've rebuilt examples files with `gradlew knit` after those changes and pushed to `develop` branch.

Issue #212, opened 2018-01-30
`withTimeout` hangs on objects with exception-throwing `equals`


elizarov:
Run this code: 

```
import kotlinx.coroutines.experimental.*

class FooBar {
    override fun equals(obj: Any?): Boolean = TODO()
}

fun main(args: Array<String>) = runBlocking<Unit> {
    withTimeout(100L) {
        FooBar()
    }
}
```

It should return normally (returning an instance of `FooBar` from inside of `withTimeout` block), but it hangs. 

This problem was originally reported here: https://youtrack.jetbrains.com/issue/KT-21629

konrad-kaminski:
Fixed in #215. Object equality instead of reference equality was used when comparing with `COROUTINE_SUSPENDED`.

PR #215, opened 2018-02-01
Fix for #212.

konrad-kaminski:


elizarov:
I've added a test and merged the patch to `develop`

Issue #238, opened 2018-02-12
Feature - Priority Channel


wdroste:
It would great to have a priority based channel, such that I can use Comparable to make one object of a higher priority than another such as PriorityQueue. I was looking at this example.. and it would be trivial but of course this is not the core.
https://github.com/Kotlin/kotlin-coroutines/blob/master/examples/channel/channel.kt

wdroste:
@elizarov any questions, comments, concerns?

wdroste:
So I tried to port this to my own code base however because this Symbol is marked _internal_ I can't.. easily that is.. 

internal val ALREADY_SELECTED: Any = Symbol("ALREADY_SELECTED")

elizarov:
Can you, please, elaborate on the use-case for such a priority channel? Do you any specific usage example in mind? 

I'm also referencing PR #238 here.

wdroste:
The specific use case is I've got a high performance application that is sending messages to the channel, some messages are much more important than others. In our case some are human driven so they expect lower latency aka higher priority. The other messages can have higher latency. 

The messages in the channel are all the same type and just data.. but due to the fact that some are human driven there's an expectation even under **load** that latency or priority for those messages be elevated because there's a person waiting on the response.

elizarov:
The problem with the proposed implementation is that it is using a `PriorityQueue` that would loose FIFO order for messages at the same priority. It can be fixed. However, I would recommend to consider a totally different approach to your use-case. Instead of designing a special channel based on a priority queue use a separate channel for each priority instead. You can easily use `select { ... }` expression to receive from multiple channels (higher-priority goes first in the list) and this way you get your use-case covered without having to implement a separate channel kind. Moreover, you will not have to make all your messages `Comparable`, too.

wdroste:
Do you have an example? I struggling to see how that's better.. do I have more coroutines on the higher priority channels and rank them? The select statement is just routing.. FIFO on the same priority is not an issue for me.. Maybe your talking about having the select {} do it from the higher and then lower and lower and lower.. Anyway if you don't think its cool that's fine but could you at least make it so I can leverage the great work you guys have already done in the AbstractChannel

fvasco:
Hi @wdroste,
a prototype.

```kotlin
abstract class PriorityChannel<T>(protected val subschannels: List<Channel<T>>) : Channel<T> {
    /**
     * Send the specified [item] to a single subschannel
     */
[...]

PR #239, opened 2018-02-12
#238 -Added naive PriorityChannel implementation

wdroste:


elizarov:
What would be the use-cases for such a channel implementation? Can you give an example?

Issue #262, opened 2018-02-27
How should Java provide us with a `Deferred<T>`


ZakTaccardi:
I'm building an SDK in Kotlin, where I want the client to pass in a feature throttle. I'd like this to be a non-blocking future, like `Deffered<T>` or if it were Rx, a `Single<T>`.

```
MySdk.Builder()
  .setFeatureFlag(deferred) // where `deferred` is a `Deferred<Boolean>`
  .build()
```

Internally in the SDK, I wait to asynchronously check if the flag is `true` or `false`

```
launch {
  if(deferred.await()){
    // enable feature
  } else {
    //disable feature
  }
}
```

If the client app, which is injecting an instance of a `Deferred<T>` into the `MySdk.Builder.setFeatureFlag(deferred)` method, is written in Java, how can they best provide an instance of `Deferred<T>`?

If the client app was written in Kotlin, I believe an implementation of `CompletableFuture` would be appropriate (not sure on this).

If this were Rx, I'd expect the client app to inject a `Single.fromCallable { .. }`.

I think it's important for Java consumers to easily construct an instance of `Deferred<T>` (if this is not already possible)

jcornaz:
It is possible to call `async` from java with `DefferedKt.async()`, but this is cumbersome, because the caller would have to give all arguments that are usually optionnal (coroutine context, parent job and start mode)

But it is possible to add create a small and simple helper for that:
```kotlin
// Helper for java written in Kotlin
object DeferredHelper {

  @JvmStatic
  fun <T> create(future: CompletableFuture<T>): Deferred<T> = async { future.get() }

  @JvmStatic
  fun <T> create(supplier: Supplier<T>): Deferred<T> = async { supplier.get() }
}
```

Then one can create a deferred with `DeferredHelper.create(() -> 42)` from Java

elizarov:
In Java you should be able to convert `CompletableFuture` (Java's approach to async) into `Deferred` with something like `FutureKt.asDeferred(future)`, but we don't currently have this in a ready-to-use method.

We need to add extension `fun <T> CompletionStage<T>.asDeferred(): Deferred<T>` into `kotlinx-coroutines-jdk8` module.

ZakTaccardi:
That doesn't work for Android unfortunately, which is stuck on Java 7

elizarov:
[...]

PR #263, opened 2018-02-28
Add conversion of CompletionStage to Deferred

jcornaz:
Hello,

This PR Implements `fun <T> CompletionStage<T>.asDeferred(): Deferred<T>` as proposed by @elizarov  in #262.

Let me know what you think about it.

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/src/test/kotlin/kotlinx/coroutines/experimental/future/FutureTest.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/test/kotlin/kotlinx/coroutines/experimental/future/FutureTest.kt
@@ -25,6 +25,8 @@ import java.util.concurrent.CompletableFuture
 import java.util.concurrent.CompletionStage
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.concurrent.withLock
 import kotlin.coroutines.experimental.CoroutineContext
 
 class FutureTest : TestBase() {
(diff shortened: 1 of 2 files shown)


fvasco:
You should check `isDone`, invoking `whenComplete` on completed futures is expensive.

jcornaz:
Yes indeed.

As `CompletableStage` doesn't provide any `isDone`  function I wrote:
`if (this is Future<*> && isDone()) return CompletableDeferred(get() as T)`

fvasco:
Nice,
you should add the test for `CompletableFuture.failedFuture`.

Both `CompletableDeferred` and `CompletableFuture` have `completeExceptionally`, for clarity use `also` instead of `apply`.

jcornaz:
Okay, I fixed it. I wasn't expecting the correct exception type in the test. A future created as failed wrap the cause exception in `ExecutionException` while a failure during the task is wrapped in `CompletionException`.

integration/kotlinx-coroutines-jdk8/src/test/kotlin/kotlinx/coroutines/experimental/future/FutureTest.kt
```diff
@@ -255,6 +255,49 @@ class FutureTest : TestBase() {
...
+            fail("deferred.await() should throw an exception")
+        } catch (e: Exception) {
+            assertEquals("something went wrong", e.cause!!.message)
+        }
+    }
+
+    @Test
+    fun testCompletableFutureWithExceptionAsDeferred() = runBlocking {
+        val deferred: Deferred<Int> = CompletableFuture.supplyAsync {
+            Thread.sleep(100)
```
fvasco: This `sleep` is useful?

integration/kotlinx-coroutines-jdk8/src/test/kotlin/kotlinx/coroutines/experimental/future/FutureTest.kt
```diff
@@ -255,6 +255,49 @@ class FutureTest : TestBase() {
         assertThat(future.get(), IsEqual("OK"))
     }
 
+    @Test
+    fun testCompletableFutureStageAsDeferred() = runBlocking {
+        val deferred: Deferred<Int> = CompletableFuture.supplyAsync { Thread.sleep(100) ; 42 }.asDeferred()
```
fvasco: A Lock is more reliable than a `sleep`.
You can
1. use a Lock to pause the Supplier
2. check Deferred as incomplete
3. release the Lock to resume the Supplier
4. verify Deferred result

fvasco:
Hi @jcornaz,
all your commits refer to a single proposal, you should squash all them in a single one (using `git rebase` to rewrite your commit history).

jcornaz:
Thanks @fvasco. I wasn't convinced either by the `sleep`. And using a lock is indeed way better.

I used kotlinx.coroutines `Mutex` is that okay ?

fvasco:
Using a `Mutex` in a `Supplier` is a bit tricky, try yourself (like using a `delay` instead of a `sleep`).

jcornaz:
Of course I have to use it inside of `runBlocking` is that what you mean by "tricky" ?

jcornaz:
Yes it is more readable I agree. I've replace the `Mutex` by a Java `Lock`

elizarov:
Thanks a lot. Merged. I've made a few changes though. The most important one is that in the fast-path of `CompletionStage.asDeferred` when we are using `future.get()` we should unwrap `ExecutionException` that `get` uses just like it is done in `await` implementation. I've updated the tests correspondingly to make a stricter checks that the appropriate exception types are preserved.

I've also implemented two-way cancellation support `CompletionStage.asDeferred` in a similar vein to `Deferred.asCompletionStage`. It is just one-line addition. As a result, when the resulting deferred is cancelled, it now cancels the source `CompletionStage` is that is possible (if it implements `Future` interface, that is).

I've also took a fresh look on other functions in this file and made changes there, too. I've used the same `this is Future<*>` trick in `CompletionStage.await` implementation, made it cancellable (what I was originally thinking -- I don't know) and deprecated a specialized version of `await`.

jcornaz:
Cool, thank you.

pacher:
> invoking `whenComplete` on completed futures is expensive

@fvasco Hi, do you have anything I can read about performance issues of `whenComplete`? I could not find it and quick looking at the code did not help. And `CompletionStage<T>.await()` uses it under the hood.
Is it expensive only for completed futures? Is it fine to use it for "awating" the result of not yet completed ones?
Is there any better alternative? Should I use `CompletableDeffered` and `invokeOnCompletion` instead? In my experience it was even slower, but it was long time ago and just some observations, not a real benchmark.

fvasco:
Hi, @pacher,
`whenComplete` is more expensive than `get`, that's all.
I don't know your use case, but please consider avoiding conversions when possible.

PR #265, opened 2018-03-02
Improved Android's onClick

mice777:
It writes that there's no "MouseEvent" on Android, actually there is android.view.MotionEvent. This sounds misleading.
Here is proposal for improvement to pass through View parameter in onClick.


elizarov:
Thanks

PR #268, opened 2018-03-04
Update RxJava1 to RxJava2 in kotlinx-coroutines-rx-example

takahirom:
Update and use `io.reactivex.Single`  instead of `rx.Observable `


Ogiwara-CostlierRain464:
I'm really expecting for muluti-platform RxKotlin…

elizarov:
Thanks. Merged to develop branch.

PR #270, opened 2018-03-05
Add a context argument to Channel.filterNot

jcornaz:
Hello, 

This PR simply add a `CoroutineContext` argument to `filterNot` (with a default value of `Unconfined`)

It makes it more consistent with the other channel operators and especially with `filter` which already has a `CoroutineContext` argument.


elizarov:
You should add an overloaded version with the old signature marked with `@Deprecated(level = DeprecationLevel.HIDDEN)` and marked with `@suppress` in KDoc documentation to maintain binary compatibility with previously compiled code.

jcornaz:
Like this?

elizarov:
Yes. Thanks. Merged.

Issue #258, opened 2018-02-26
Android LiveData / architecture components integration


elizarov:
We need an integration module for Android `LiveData` architecture components that provides read-to-use extensions for their use with coroutine in a similar style to reactive integration modules. See also #232

LouisCAD:
@GeoffreyMetais We may help this get real by telling our experience using both `LiveData` and kotlinx.coroutines

elizarov:
@LouisCAD Please, do.

LouisCAD:
The first thing I'd like to say is about null safety and generics:
`LiveData` is not very Kotlin friendly as its `value` property can be `null` even if the type parameter is not nullable, and its `Observer` SAM passes an explicitly `@Nullable` value.

However, `LiveData` is not tightly coupled to the rest of the lifecycle part of Android Architecture Components, and it could be rewritten in a more Kotlin friendly way (and in Kotlin BTW) where it could take an initial value so it can be non null.

`LiveData` is based on `Lifecycle`, and you can see in the imports that it really only depends on java.util and `Lifecycle` (plus two internal classes that only depend on java.util that could be copied if needed), and it's pretty lightweight, so there's no roadblocks to do a more Kotlin friendly implementation where you could specify an initial non null value for a non null generic type.

I think this would be the first step, but the issue is that using the class name `LiveData` could be confusing. So like `Future` vs `Deferred`, there should be another one, unless Google refactors `LiveData` itself to be more Kotlin friendly nullability-wise.

The names I found (for now) are `AliveData` and `LifecycleData`, which are IMHO, a bit less nice than `LiveData`.

If you don't agree with me, or otherwise, if you have other name ideas, please, tell me.

GeoffreyMetais:
+1 for nullability, I'm starting my own `LiveData` for convenience because of this.

`ViewModel` would need the more coroutines treatment.

In my current refactoring, I have to set an actor to queue operations and guarantee data integrity
I'm not sure I'm doing it well but I'd love this to be managed by a helper `ViewModel` class.

LouisCAD:
I agree about `ViewModel`. I have a `JobViewModel` subclass in an app that has a protected `job` property that is cancelled in `onCleared()`, for use as a parent `Job`. This is really trivial to do.

[...]

Issue #254, opened 2018-02-21
Provide abstraction for cold streams


elizarov:
All the currently provided channel abstractions in `kotlinx.coroutines` are _hot_. The data is being produced regardless of the presence of subscriber. This is good for data sources and applications that are inherently hot, like incoming network and UI-events. 

However, hot streams are not an ideal solution for cases where data stream is produced on demand. Consider, for example, the following simple code that produces `ReceiveChannel<Int>`:

```
produce<Int> { 
    while (true) {
        val x = computeNextValue()
        send(x)
    } 
}
```

One obvious downside is the `computeNextValue()` is invoked before `send`, so even when receiver is not ready, the next value gets computed. Of course, it gets suspended in `send` if there is no receiver, but it is not as lazy as you get with cold reactive Publisher/Observable/Flowable/Flux/Flow.

We need the abstraction for cold streams in `kotlinx.coroutines` that is going to be just as lazy, computing data in "push" mode versus "pull" mode of hot channels that we have now.

There are the following related discussions:
* https://github.com/reactor/reactor-core/issues/979#issuecomment-351770494 describes preliminary performance test that indicates that "push" mode is much faster for same-thread cases.
* https://github.com/Kotlin/kotlinx.coroutines/issues/113 (SPSC channels) seems to get superseded by the support of cold streams.

elizarov:
Let me add here that we are looking at an asynchronous analogue of a  standard Kotlin's `Sequence` here. We _already_ can write the following code in Kotlin:

```
val s = buildSequence<Int> { 
    while (true) {
        val x = computeNextValue()
        yield(x)
    } 
}
```

and this code is perfectly lazy in the sense that `computeNextValue` is not invoked until sequence gets to be iterated and only actually requested values are ever compute. However, we cannot do arbitrary suspension from inside of `buildSequence`. We cannot `delay`, we cannot do asynchronous network requests, etc.

pull-vert:
Hi,
I was writing a kotlinx-coroutine reactive (=cold) library based on reactive-streams. But following Elizarov advice on #201 I switched to a pure kotlin suspend version, strongly inspired by your StreamBenchmarks project.
Here it is : [Reactivity](https://github.com/pull-vert/reactivity)

It provides 2 cold producers (Reactor inspiration) : 

[...]

Issue #232, opened 2018-02-09
Room integration


elizarov:
See #231. We should provide an out-of-the box integration with Room either via an integration module (if that is possible in Room at all) or via a PR to the Room library.

dmytrodanylyk:
Also, I would love to have extension functions for android `LiveData` to work with channels, not with `Observable`.

See [LiveData](https://developer.android.com/reference/android/arch/lifecycle/LiveData.html), [Observer](https://developer.android.com/reference/android/arch/lifecycle/Observer.html)

Kernald:
There's work in progress (and a first stage released as of Room 2.1.0-alpha03) tracked here: https://issuetracker.google.com/issues/69474692

elizarov:
I'm closing this issue as this integration work is being up on Google side.

PR #274, opened 2018-03-08
Add Subscribable interfaces

### LINKED ISSUE #245: ConflatedBroadcastChannel as ReadWriteProperty
Hello, 

In our code base, we found useful to define the folowing extension functions:
```kotlin
operator fun <T> ConflatedBroadcastChannel<T>.getValue(thisRef: Any, property: KProperty<*>): T = value

operator fun <T> ConflatedBroadcastChannel<T>.setValue(thisRef: Any, property: KProperty<*>, value: T) {
  offer(value)
}
```

They allow us to use the following pattern: 
```kotlin
class MyClass {
  private val _value = ConflatedBroadcastChannel(0)
  var value by _value 

  fun openValueSubscription() = _value.openSubscription()
}
```

May I ask what do you think about this ?
Do you think it is an anti-pattern ?
* If yes, why? And how would you achieve better generic observer pattern with coroutines?
* If no, what would you think about making `ConflatedBroadcastChannel` implement `ReadWriteProperty`?

jcornaz:
Resolve #245 

This PR add the following:
* interface `Subscribable`:
Represent anything from which it is possible to open a subscription
* interface `SubscribableValue`:
Represent a ` Subscriblable` from which it is always possible to get a *current value*.
* interface `SubscriblableVariable`:
Mutable `SubscribableValue` allowing to set the *current value*
* factory functions `SubscribableValue(value)`and `SubscribableVariable(intialValue)`:
Which create instances for the corresponding interfaces.

This PR also make `BroadcastChannel` a sub interface of `Subscribable`.

### Decisions made:
* We do not support late initialization. Therefore `ConflatedBroadacstChannel` do not implement `SusbcribableVariable`. This makes the overall usage simpler and safer to use.
* Operators such as `map` and `combine` or adapter for JavaFX `ObservableValue` won't be part of this PR, but may be made and discussed in another issue/PR

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/channels/BroadcastChannel.kt
+++ b/common/kotlinx-coroutines-core-common/src/channels/BroadcastChannel.kt
@@ -18,7 +18,7 @@ import kotlinx.coroutines.experimental.internalAnnotations.*
  * See `BroadcastChannel()` factory function for the description of available
  * broadcast channel implementations.
  */
-public interface BroadcastChannel<E> : SendChannel<E> {
+public interface BroadcastChannel<E> : SendChannel<E>, Subscribable<E> {
     /**
      * Factory for broadcast channels.
      * @suppress **Deprecated**
@@ -37,7 +37,8 @@ public interface BroadcastChannel<E> : SendChannel<E> {
      * The resulting channel shall be [cancelled][ReceiveChannel.cancel] to unsubscribe from this
      * broadcast channel.
      */
-    public fun openSubscription(): ReceiveChannel<E>
+    // TODO: Remove (already declared in Subscribable). Kept here only for binary compatibility
+    public override fun openSubscription(): ReceiveChannel<E>
 
     /**
      * @suppress **Deprecated**: Return type changed to `ReceiveChannel`, this one left here for binary compatibility.
--- a/common/kotlinx-coroutines-core-common/src/channels/Subscribables.kt
+++ b/common/kotlinx-coroutines-core-common/src/channels/Subscribables.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2016-2018 JetBrains s.r.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kotlinx.coroutines.experimental.channels
+
+import kotlinx.coroutines.experimental.Unconfined
+import kotlin.properties.ReadOnlyProperty
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+/**
+ * Interface for classes allowing to open subscriptions.
+ */
+public interface Subscribable<out T> {
+
+    /**
+     * Subscribes and returns a channel to receive elements broadcasted by this Subscribable.
+     * The resulting channel shall be [cancelled][ReceiveChannel.cancel] to unsubscribe from this subscribable.
+     */
+    public fun openSubscription(): ReceiveChannel<T>
+}
+
+/**
+ * Interface for a value holder where the value may change over time.
+ *
+ * The current value can be obtained with [value] and one can open a subscription in order to receive the new value when changed.
+ *
+ * The channel returned by [openSubscription] is "conflated" and start by the current value if any.
+ */
+public interface SubscribableValue<out T> : Subscribable<T>, ReadOnlyProperty<Any?, T> {
+
+    /**
+     * Current value
+     */
+    public val value: T
+
+    /**
+     * Returns the current value.
+     *
+     * @see value
+     */
+    public override fun getValue(thisRef: Any?, property: KProperty<*>): T = value
+}
+
+/**
+ * Mutable version of [SubscribableValue].
+ *
+ * Provide ability to get and set the current value with [value].
+ */
+public interface SubscribableVariable<T> : SubscribableValue<T>, ReadWriteProperty<Any?, T> {
+
+    /**
+     * Current value
+     */
+    public override var value: T
+
+    public override fun getValue(thisRef: Any?, property: KProperty<*>): T = value
+
+    /**
+     * Set the current value
+     */
+    public override fun setValue(thisRef: Any?, property: KProperty<*>, value: T) {
+        this.value = value
+    }
+}
+
+/**
+ * Create an instance of [SubscribableValue] with the given [value]
+ */
+public fun <T> SubscribableValue(value: T): SubscribableValue<T> = SubscribableValueImpl(value)
+
+/**
+ * Create an instance of [SubscribableVariable] initialized with the given [initialValue]
+ */
+public fun <T> SubscribableVariable(initialValue: T): SubscribableVariable<T> = SubscribableVariableImpl(initialValue)
+
+private class SubscribableValueImpl<T>(override val value: T): SubscribableValue<T> {
+    override fun openSubscription() = produce(Unconfined) { send(value) }
+}
+
+private class SubscribableVariableImpl<T>(initialValue: T): SubscribableVariable<T> {
+    private val broadcast = ConflatedBroadcastChannel(initialValue)
+
+    override var value: T
+        get() = broadcast.value
+        set(value) {
+            broadcast.offer(value)
+        }
+
+    override fun openSubscription() = broadcast.openSubscription()
+}
--- a/common/kotlinx-coroutines-core-common/test/channels/SubscribableValueTest.kt
+++ b/common/kotlinx-coroutines-core-common/test/channels/SubscribableValueTest.kt
@@ -1,0 +1,63 @@
+package kotlinx.coroutines.experimental.channels
+
+import kotlinx.coroutines.experimental.CoroutineStart
+import kotlinx.coroutines.experimental.TestBase
+import kotlinx.coroutines.experimental.Unconfined
+import kotlinx.coroutines.experimental.launch
+import kotlin.coroutines.experimental.coroutineContext
+import kotlin.test.Test
+
+class SubscribableValueTest : TestBase() {
+
+    @Test
+    fun testBasicScenario() = runTest {
+        expect(1)
+        val subscribable = SubscribableValue(42)
+        val variable by subscribable
+        check(subscribable.value == 42)
+        check(variable == 42)
+        launch(coroutineContext, start = CoroutineStart.UNDISPATCHED) {
+            expect(2)
+            val sub = subscribable.openSubscription()
+            check(sub.receive() == 42)
+            expect(3)
+            try {
+                sub.receive()
+                expectUnreached()
+            } catch (e: ClosedReceiveChannelException) {
+                // expected exception
+            }
+            expect(4)
+        }
+        finish(5)
+    }
+
+    @Test
+    fun testOpenSubscriptionAlwaysStartByTheCurrentValue() {
+        expect(1)
+        val subscribable = SubscribableValue(42)
+        launch(Unconfined) {
+            expect(2)
+            check(subscribable.openSubscription().first() == 42)
+            check(subscribable.openSubscription().first() == 42)
+            expect(3)
+        }
+        finish(4)
+    }
+
+    @Test
+    fun testObservableValue() {
+        expect(1)
+        val subscribable = SubscribableValue(42)
+        val value by subscribable
+        check(subscribable.value == 42)
+        check(value == 42)
+        launch(Unconfined) {
+            expect(2)
+            check(subscribable.openSubscription().first() == 42)
+            check(subscribable.openSubscription().first() == 42)
+            expect(3)
+        }
+        finish(4)
+    }
+}
--- a/common/kotlinx-coroutines-core-common/test/channels/SubscribableVariableTest.kt
+++ b/common/kotlinx-coroutines-core-common/test/channels/SubscribableVariableTest.kt
@@ -1,0 +1,65 @@
+package kotlinx.coroutines.experimental.channels
+
+import kotlinx.coroutines.experimental.CoroutineStart
+import kotlinx.coroutines.experimental.TestBase
+import kotlinx.coroutines.experimental.Unconfined
+import kotlinx.coroutines.experimental.launch
+import kotlinx.coroutines.experimental.yield
+import kotlin.coroutines.experimental.coroutineContext
+import kotlin.test.Test
+
+class SubscribableVariableTest : TestBase() {
+
+    @kotlin.test.Test
+    fun testBasicScenario() = runTest {
+        expect(1)
+        val subscribable = SubscribableVariable(42)
+        var variable by subscribable
+        check(subscribable.value == 42)
+        check(variable == 42)
+        launch(coroutineContext, start = CoroutineStart.UNDISPATCHED) {
+            expect(2)
+            val sub = subscribable.openSubscription()
+            check(sub.receive() == 42)
+            expect(3)
+            check(sub.receive() == 24) // suspends
+            expect(6)
+        }
+        expect(4)
+        variable = 24
+        check(subscribable.value == 24)
+        check(variable == 24)
+        expect(5)
+        yield() // to child
+        finish(7)
+    }
+
+    @Test
+    fun testOpenSubscriptionAlwaysStartByTheCurrentValue() {
+        expect(1)
+        val subscribable = SubscribableVariable(42)
+        launch(Unconfined) {
+            expect(2)
+            check(subscribable.openSubscription().first() == 42)
+            check(subscribable.openSubscription().first() == 42)
+            expect(3)
+        }
+        finish(4)
+    }
+
+    @Test
+    fun testObservableValue() {
+        expect(1)
+        val subscribable = SubscribableValue(42)
+        val value by subscribable
+        check(subscribable.value == 42)
+        check(value == 42)
+        launch(Unconfined) {
+            expect(2)
+            check(subscribable.openSubscription().first() == 42)
+            check(subscribable.openSubscription().first() == 42)
+            expect(3)
+        }
+        finish(4)
+    }
+}


jcornaz:
Oups....  I opened the PR for against the `master` branch. Should I close it an recreate it ?

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
```diff
@@ -38,7 +38,7 @@ import kotlinx.coroutines.experimental.selects.SelectInstance
  * [opening][openSubscription] and [closing][SubscriptionReceiveChannel.close] subscription takes O(N) time, where N is the
  * number of subscribers.
  */
-public class ConflatedBroadcastChannel<E>() : BroadcastChannel<E> {
+public class ConflatedBroadcastChannel<E>() : BroadcastChannel<E>, SubscribableVariable<E> {
```
fvasco: Should we inject an automatic item replay into all `ConflatedBroadcastChannel`?
This is a breaking change and a different behavior from channels' default.

In your design `SubscribableVariable` is not a `Channel`, so I prefer to left `ConflatedBroadcastChannel` as is.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
```diff
@@ -81,25 +81,29 @@ public class ConflatedBroadcastChannel<E>() : BroadcastChannel<E> {
...
+                        return state.value as E
+                    }
+                    else -> error("Invalid state $state")
                 }
-                else -> error("Invalid state $state")
             }
         }
-    }
+        set(value) {
+            sendBlocking(value)
```
fvasco: `offer` should be enough.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/BroadcastChannel.kt
```diff
@@ -47,7 +47,7 @@ public interface BroadcastChannel<E> : SendChannel<E> {
      * The resulting channel shall be [closed][SubscriptionReceiveChannel.close] to unsubscribe from this
      * broadcast channel.
      */
-    public fun openSubscription(): SubscriptionReceiveChannel<E>
+    override fun openSubscription(): SubscriptionReceiveChannel<E>
```
fvasco: You can remove this method and shift documentation into `Subscribable`'s one.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,36 @@
...
+
+  public val valueOrNull: T?
+
+  override fun getValue(thisRef: Any, property: KProperty<*>): T = value
+}
+
+public interface SubscribableVariable<T> : SubscribableValue<T>, ReadWriteProperty<Any, T> {
+  public override var value: T
+
+  override fun getValue(thisRef: Any, property: KProperty<*>): T = value
```
fvasco: Is it already defined in `SubscribableValue`?

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * Current value if any.
+   *
+   * Access this property may throws [IllegalStateException] if there is no value set yet.
+   *
+   * Use [valueOrNull] to get `null` when the value is not set or `openSubscription().consume { receive() }` to suspend until a value is available.
+   */
+  public val value: T
+
+  public val valueOrNull: T?
```
fvasco: We should support `lateinit`?

Is `Deferred<SubscribableValue>` a good replacement?

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
jcornaz: `offer` would never throw any exception. But I could do `if (!offer(value)) throw IllegalStateException()`.

Or do we prefer to silently fail if the broadcast channel is closed ?

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
jcornaz: Currently `ConflatedBroadcastChannel` does re-send the last item through the channel opened with `openSubscription()`. Or am I misunderstanding your point ?

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/BroadcastChannel.kt
jcornaz: Yes I'll do it.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
jcornaz: Yes, but Kotlin compiler enforce to redeclare it, because it is inherited from two interfaces (`SubscribableValue` and `ReadWriteProperty`)

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
fvasco: Are you sure?

https://github.com/Kotlin/kotlinx.coroutines/blob/fc6e31e206d3a024189b701903736429423fc767/core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt#L201

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
jcornaz: I think yes, because this is the actual current behavior of `ConflatedBroadcastChannel`.

https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.experimental.channels/-conflated-broadcast-channel/value.html

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
jcornaz: https://kotlinlang.org/docs/reference/classes.html#overriding-rules

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
fvasco: Sorry, my fault.

Please discard this comment.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
jcornaz: Yes sorry, I was wrong. I'll use `offer`. At least I learned something :-)

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
fvasco: Ok, discard this point.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
fvasco: Ok

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/BroadcastChannel.kt
jcornaz: Mmm. After a 2nd though I wonder if removing it couldn't cause binary incompatibilities?

Should I leave it for this reason?

jcornaz: Yes it could definitely cause binary incompatibility. I think we should leave it.

fvasco:
We should provide also

    fun SubscribableValue.map(...) : SubscribableValue

However this implementation should be lazy and GC friendly.

jcornaz:
>We should provide builders for SubscribableValue and SubscribableVariable.

Yes, I'll propose something.

jcornaz:
Definitely yes for the `map` operator. I was just not sure if it should be part of this PR or not.

If a `map` operator is provided, I think we should also provide:
```kotlin
fun <T1, T2, R> SubscribableValue<T1>.combineWith(other: SubscribableValue<T2>, combine: (T1, T2) -> R): SusbcribableValue<R>
```

Should I also provide an adapter for JavaFx: `fun <T> SubscriblableValue<T>.asObservableValue(): ObservableValue<T>` in the module `kotlinx-coroutines-javafx`? To me it makes a lot of sense and I would write it in any of my JavaFx application anyway. This is something that would make interoperability between coroutines and JavaFx seamless and very pleasant. But perhaps, I should do it in a separate PR?

jcornaz:
>We should provide builders for SubscribableValue and SubscribableVariable.

Here is a wild API suggestion. Is it looking like what you imagined @fvasco ?
```kotlin
interface SubscribableValueBuilderScope<in T> : CoroutineScope {
  suspend fun yield(value: T)
}

interface SubscribableValueJob<T> : SubscribableValue<T>, Job

fun <T> SubscribableVariable(initialValue: T): SubscribableVariable<T> = ConflatedBroadcastChannel(initialValue)

fun <T> buildSubscribableValue(
    context: CoroutineContext = DefaultDispatcher,
    builderAction: suspend SubscribableValueBuilderScope<T>.() -> Unit
): SubscribableValueJob<T> = TODO()

fun usageExample() {
  val variable = SubscribableVariable<Int>(0)

  variable.value = 42

  val currentTime: SubscribableValueJob<Instant> = buildSubscribableValue {
    while (isActive) {
      delay(1, TimeUnit.SECONDS)
      yield(Instant.now())
    }
  }

  currentTime.cancel()
}
```

fvasco:
`buildSubscribableValue` is really looks like the `actor` builder, I seen your code and probably the `ConflatedBroadcastChannel` class is enough, at least for my use case.

Instead I am considering another kind of actor

```kotlin
fun <E> SubscribableValue<E>.launchObserver(
        context: CoroutineContext = DefaultDispatcher,
        parent: Job? = null,
        block: suspend ActorScope<E>.() -> Unit
): Job
```

This requires explicit `cancel` so I am sceptic about it.
However this fits my use case, can you share your considerations?

jcornaz:
>buildSubscribableValue is really looks like the actor builder

Actually more like `produce` than `actor`. But create it would create broadcast instead of a channel.

>probably the ConflatedBroadcastChannel class is enough, at least for my use case

For my use case too. So I won't add `buildSubscribableValue` in this PR. Should I add the `SubscribableVariable()` function, which would only be an alias to `ConflatedBroadcastChannel` constructor?

About the `launchObserver`, we implemented the following in our code-base: 
```kotlin
/**
 * Start a job that consume each elements of the channel and execute [onEach] of each of them.
 */
fun <E> ReceiveChannel<E>.launchConsumer(context: CoroutineContext = DefaultDispatcher, onEach: suspend (E) -> Unit): Job =
    launch(context) { consumeEach { onEach(it) } }
```
I feel the same Idea than behind your `launchObserver`. The difference is that this one can be used for any `ReceiveChannel` regardless if it is coming from a broadcast or not.

I am not particular found of the `SubscribableValue.launchObserver` you proposed for the exact reason you mentioned: "it requires explicit cancel".

And anyway, I do not need it personally. Because I use JavaFx, I would most of the time simply bind a JavaFx component to the result of `SubscribableValue.asObservableValue()`. With this pattern when the JavaFx component would unsubscribe, everything would be cancelled. So it is easier and safer to use (in the context of JavaFX).

And for other usages, it is not really an hassle to do:

```kotlin
launch {
  observableValue.openSubscription().consumeEach {
    // do something usefull
  }
}
```

But this is purely personal opinion. I will add you guys want it.

fvasco:
Well,
so limit this pull request to "Subscribable interfaces", probably it is not useful dump all ideas here.

jcornaz:
Ok, I added some comments and a test for `ConflatedBroadcastChannel.setValue`

What should we do for `BroadcastChannel.openSubscription()`? We cannot remove it without breaking binary compatibility. But we cannot deprecated it either, because it is still the correct method to use.

Let me know if if I should correct something or do something.

I'll open new PRs for the `map` and `combine` operators and for the JavaFx adapter.

fvasco:
> What should we do for BroadcastChannel.openSubscription()?

Mark it for removal using a `:todo:` comment.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/BroadcastChannel.kt
fvasco: Resume `public` modifier.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
```diff
@@ -81,25 +81,25 @@ public class ConflatedBroadcastChannel<E>() : BroadcastChannel<E> {
...
      */
     @Suppress("UNCHECKED_CAST")
-    public val value: E get() {
-        _state.loop { state ->
-            when (state) {
-                is Closed -> throw state.valueException
-                is State<*> -> {
-                    if (state.value === UNDEFINED) throw IllegalStateException("No value")
-                    return state.value as E
+    override var value: E
```
fvasco: Resume `public` modifier.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/ConflatedBroadcastChannel.kt
```diff
@@ -81,25 +81,25 @@ public class ConflatedBroadcastChannel<E>() : BroadcastChannel<E> {
...
+            offer(value)
+        }
 
-    /**
-     * The most recently sent element to this channel or `null` when this class is constructed without
-     * initial value and no value was sent yet or if it was [closed][close].
-     */
     @Suppress("UNCHECKED_CAST")
-    public val valueOrNull: E? get() {
+    override val valueOrNull: E? get() {
```
fvasco: Resume `public` modifier.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,83 @@
...
+    public val valueOrNull: T?
+
+    /**
+     * Return the current value if any.
+     *
+     * May throw an exception if the implementation has been terminated with an error.
+     *
+     * @throws IllegalStateException If no value has been set yet
+     */
+    override fun getValue(thisRef: Any, property: KProperty<*>): T = value
```
fvasco: Add `public` modifier.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,83 @@
...
+     * Reading this property may throw an exception if the implementation has been terminated with an error.
+     *
+     * Use [valueOrNull] to get `null` when the value is not set or `openSubscription().consume { receive() }` to suspend until a value is available.
+     *
+     * @throws IllegalStateException When reading the property and no value has been set yet.
+     * @throws ClosedSendChannelException When setting the property and the implementation has been terminated.
+     */
+    public override var value: T
+
+    override fun getValue(thisRef: Any, property: KProperty<*>): T = value
```
fvasco: Add `public` modifier.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Set the current value.
+     *
+     * @throws ClosedSendChannelException If the implementation has been terminated
+     */
+    override fun setValue(thisRef: Any, property: KProperty<*>, value: T) {
```
fvasco: Add `public` modifier.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,83 @@
...
+ * The channel returned by [openSubscription] is "conflated" and start by the current value if any.
+ */
+public interface SubscribableValue<out T> : Subscribable<T>, ReadOnlyProperty<Any, T> {
+
+    /**
+     * Current value if any.
+     *
+     * Reading this property may throw the cause exception if the implementation has been terminated with an error.
+     *
+     * Use [valueOrNull] to get `null` when the value is not set or `openSubscription().consume { receive() }` to suspend until a value is available.
```
fvasco: `openSubscription().first()`?

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/channels/Subscribables.kt
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @throws IllegalStateException When reading the property and no value has been set yet
```
fvasco: What happen if channel is closed?

fvasco:
Can you test some negative paths?

jcornaz:
>Can you test some negative paths?

Yes, good Idea, I'll add some for `ConflatedBroadcastChannel`

jcornaz:
Ok, I added some test for negative paths (channel with no value set yet, channel closed, channel closed with a cause)

bohsen:
Just wanted to say really great work to you guys.
The ConflatedBroadcastChannel IMO is a really powerful tool.
I've used something similar where I had to combine it with Android Architecture Components and a custom LiveData class (look below, boilerplate removed), but having this in the library will be SO good. Could also help solving #232 and #258.

```
/**
 * [LiveData] object for observing changes in viewstate. Uses kotlin coroutine [channel]s for
 * communicating changes
 */
class UiStateLiveData : LiveData<UiStateModel>() {

    private val channel = ConflatedBroadcastChannel<UiStateModel>(UiStateModel.Error(IllegalArgumentException("Invalid Response")))
    private lateinit var subscription: SubscriptionReceiveChannel<UiStateModel>

    override fun onInactive() {
        subscription.close()
        super.onInactive()
    }

    override fun onActive() {
        super.onActive()
        subscription = channel.openSubscription()
        async { subscription.consumeEach { value -> postValue(value) } }
    }

    fun setUiState(state: UiStateModel) {
        async { channel.send(state) }
    }
}
```

fvasco:
Should we support better _lazy_ initialization?
Probably `valueOrNull` isn't good enough (`null` is a valid value for `T`).

```kotlin
public interface SubscribableValue<out T> : Subscribable<T>, ReadOnlyProperty<Any?, T> {

    public fun isInitialized(): Boolean

    public val value: T

    ...
}

public suspend fun <T> SubscribableValue<T>.awaitValue() =
        when {
            isInitialized() -> value
            else -> openSubscription().first()
        }
```

dave08:
Maybe `value` should be lazy if no initial value was passed, since I think in most cases, a subscribable value should really have an initial value... At least in my use case for Andoid MVP or MVVM it would always be that way... So it would make the more common use case easier to use... BTW nice work guys! This could be very very helpful!

jcornaz:
@bohsen and @dave08: thank you. It's good to hear that it'll be helpful.

@fvasco:
> Should we support better lazy initialization?
Probably valueOrNull isn't good enough (null is a valid value for T).

As it is already *supported* in the way that it is possible create a `ConflatedBroadacstChannel` with no initial value, I think yes it would be a good Idea to provide a `isInitialized` property.

fvasco:
The idea is to provide a _better support_ for the lazy initialization adding `isInitialized` and `awaitValue`.

(sorry for _unglish_ :)

jcornaz:
>The idea is to provide a better support for the lazy initialization adding isInitialized and awaitValue.

Yes I got it. Sorry if I wasn't clear ;-)

I added `isInitialized` and `awaitValue` with a test for it.

fvasco:
> I choose to make isInitialized fail if the channel is closed. Let me know what you think about it

It might be acceptable.

_Tip: remember to change `SubscribableValue.value` and `SubscribableVariable.value` documentations_

jcornaz:
> Tip: remember to change SubscribableValue.value and SubscribableVariable.value documentations

Yes good catch. I did it. 

> It might be acceptable.

I'm not especially found of it, but returning `false` is not semantically correct IMO (as it may have been initialized before being closed). So I guess we have to choose the less bad option.

jcornaz:
And what if we rename `isInitialized` by `hasValue`? If so, it would be semantically correct to return false if the channel is closed.

fvasco:
Sorry @jcornaz,
I back to my steps (https://github.com/Kotlin/kotlinx.coroutines/issues/245#issuecomment-370697855).
I consider difficult to understand why a `SubscribableValue` may have no value.

Can you consider this design instead?

```kotlin
public interface SubscribableValue<out T> : Subscribable<T>, ReadOnlyProperty<Any?, T> {

    public val value: T

    public override fun getValue(thisRef: Any?, property: KProperty<*>): T = value
}

public interface SubscribableVariable<T> : SubscribableValue<T>, ReadWriteProperty<Any?, T> {

    public override var value: T

    public override fun getValue(thisRef: Any?, property: KProperty<*>): T = value

    public override fun setValue(thisRef: Any?, property: KProperty<*>, value: T) {
        this.value = value
    }
}

fun <T> SubscribableValue(value: T) = SubscribableVariableImpl(value)
fun <T> SubscribableVariable(value: T) = SubscribableVariableImpl(value)

internal class SubscribableVariableImpl<T>(value: T) : SubscribableVariable<T> {

    private val channel = ConflatedBroadcastChannel(value)

    public override var value: T
        get() = channel.value
        set(value) {
            channel.offer(value)
        }

    public override fun openSubscription() = channel.openSubscription()
}
```

No errors, no checks, no corner cases.
It is possible cover more use case using `Deferred<SubscribableValue>>` or `SubscribableValue<Option>`.

Take your time.

jcornaz:
I think you make a very good point.

>No errors, no checks, no corner cases.

That's appealing indeed... It would be simpler and safer. Sounds very good.

In the other hand, I do need lazy initialization in practice. But it doesn't have to be in `kotlinx.coroutines`. I could write my own `LazySusbcribableValue` for my use-case (which we already have in our code-base). Or if it appears other have interest in `LazySubscribableValue` we could have an explicit interface for it, different from `SubscribableValue`, to avoid confusion.

jcornaz:
Here would be the API I'd propose: 

(implementation code of `ReadOnlyProperty` and `ReadWriteProperty` is not shown)
```kotlin
public interface Subscribable<out T> {
	fun openSubscription(): SubscribtionReceiveChannel<T>
}

public interface SubscribableValue<out T> : Subscribable<T>, ReadOnlyProperty<Any?, T> {
    
	// Never throw, always succeed
	public val value: T
}

public interface SubscribableVariable<T> : SubscribableValue<T>, ReadWriteProperty<Any?, T> {

	// Never throw, always succeed
       public override var value: T
}

public interface LazySubscribableValue<out T> : Subscribable<T>, ReadOnlyProperty<Any?, T> {

	// never throw
	public val isInitialized: Boolean
    
	// only the getter throw if no value set yet
        public val value: T
	
	// never throw
	public val valueOrNull: T?
}

// Never throw. May suspend until a value is available
suspend fun <T> LazySubscribableValue<T>.awaitValue(): T = openSubscription().first()

public interface LazySubscribableVariable<T> : LazySubscribableValue<T>, ReadWriteProperty<Any?, T> {

	// only the getter throw if no value set yet
       public override var value: T
}
```

fvasco:
It is possible to replace `LateinitSubscribableValue` with `Deferred<SubscribableValue>` to cover the same use case?

jcornaz:
> It is possible to replace LateinitSubscribableValue with Deferred<SubscribableValue> to cover the same use case?

Good point. Yes it would fit my use-case. While playing a bit with it I only found a problem arising when we want a `LateinitVariable`. Still it can be solved with `Deferred<SubscriblableValue>` like this:

```kotlin
class LateinitVariable<T> : Deferred<SubscribableValue<T>> {

  private lateinit var variable: SubscribableVariable<T>
  private val variableMutex = Mutex(true)

  private val deferred = async { variableMutex.withLock { variable } }

  var value: T
    get() = variable.value
    set(newValue) {
      synchronized(this) {
        if (!this::variable.isInitialized) {
          variable = SubscribableVariable()
          variableMutex.unlock()
        }
      }
      variable.value = newValue
    }

  // delegate all methods of [Deferred] to [deferred]
}
```

But I'm not even sure I'd need it. So here is my proposition:

We only implement `SubscribableValue` and `SusbcribableVariable` in this PR, without any support for lazy initialization. If it appears in the future that a `Lateinit` version would be useful, we'll start a new discussion about it (I'll use `Deferred<SubscribableValue>` for my use case). I don't see how it could lead to breaking changes to the design of `SubscribableValue and `SubscribableVariable` anyway.

dave08:
This reminds me of BehaviorSubject and https://github.com/JakeWharton/RxRelay, just removing the onComplete... I think this new way is the problem Relay was coming to solve... I wonder if there is a use case for closing a BehaviorSubject in the first place (unless everyone is unsubscribed and shouldn't be able to subscribe anymore...). I also think this is a better implementation than the first proposal... I think @jw was also looking for such a thing in Slack recently...

jcornaz:
@fvasco, I removed the lateinit support, and used the design you proposed. It is in a separate commit so it is still easy to roll it back.

I made a slightly different implementation of the function `SubscribableValue()`. Let me know what you think about it.

I'll squash the commits if we decide to keep this latest version.

core/kotlinx-coroutines-core/src/test/kotlin/kotlinx/coroutines/experimental/channels/SubscribableVariableTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.Assert.assertThat
+import org.junit.Test
+
+class SubscribableVariableTest : TestBase() {
+
+    @Test
+    fun testBasicScenario() = runBlocking {
+        expect(1)
+        val subscribable = SubscribableVariable(42)
+        var variable by subscribable
```
fvasco: can you use`val`?

jcornaz: Here no, because at line 26 we do `variable = 24`. But `testObservableValue` use a `val`.

dave08:
I think the two SubscribableValues top level functions are the same except for the param names initialValue or value, also the name of a fun should start with lower case?

jcornaz:
@dave08 no the result type is different too (even if backed by the same implementation). `SubscribableValue()` return an immutable type while `SubscribleVariable()` return a mutable one. This is the same kind of difference than between  `listOf` and `mutableListOf`.

> the name of a fun should start with lower case @jcornaz ?

Usually yes. Nevertheless it seems to be a convention in `kotlinx.courtines` to name builder function with the same name (including capitalizing the first letter) as the of the interface of it's result. This is the case for example of `Job()` and `Channel()`

fvasco:
Should the SubscribleVariable be synchronized?
We should provide a thread unsafe implementation for performance purpose and require external synchronization?

fvasco:
To be clear: synchronization is required for value assignation but not for reading.

jcornaz:
Value assignation is implemented by `offer`. So it is already thread safe.

And I'm not sure if it is necessary to provide an thread-unsafe method for it, because the implementation of the `ConflatedBroadcastChannel.offer` is already lock-free (so more performant that synchronizing).

Of course if one wants to read the value and then update it based on the result of its reading, then it is not safe, because not atomic. But in this case, I'd say that't the responsibility of the client code to use a mutex or an atomic variable.

Or am I misunderstanding your point?

fvasco:
Sorry @jcornaz,
I wrote my draft idea, however I not meant the `synchronized` in JVM.
No problem, I looked for a lightweight `ConflatedBroadcastChannel`, I submit you a draft of it.

```kotlin
private class SubscribableVariableImpl<T>(initialValue: T) : SubscribableVariable<T> {

    /**
     * Subscription list
     *
     * @see _subscriptionsUpdater
     */
    @Volatile
    private var _subscriptions: Array<Subscription>? = null

    /**
     * `1` when [value] is updating, `0` otherwise
     *
     * @see _updatingUpdater
     */
    @Volatile
    @Suppress("unused")
    private var _updating = 0

    public override var value: T = initialValue
        set(value) {
            val subscriptions = _subscriptions
            if (subscriptions == null) {
                // update field directly
                field = value
            } else if (_updatingUpdater.compareAndSet(this, 0, 1)) {
                // lock value and notify value update
                try {
                    field = value
                    subscriptions.forEach { it.offer(value) }
                } finally {
                    _updatingUpdater.set(this, 0)
                }
            }
        }

    public override fun openSubscription(): SubscriptionReceiveChannel<T> {
        val subscription = Subscription()
        do {
            val oldSubscriptions = _subscriptions
            val newSubscriptions = oldSubscriptions?.let { it + subscription } ?: arrayOf(subscription)
        } while (!_subscriptionsUpdater.compareAndSet(this, oldSubscriptions, newSubscriptions))
        // offer initial value
        subscription.offer(value)
        return subscription
    }

    /**
     * Remove [subscription] from [_subscriptions] list
     */
    private fun removeSubscription(subscription: Subscription) {
        do {
            val oldSubscriptions = _subscriptions!!
            val n = oldSubscriptions.size
            val i = oldSubscriptions.indexOf(subscription)
            check(i >= 0)

            var newSubscriptions: Array<Subscription?>?
            if (n == 1) {
                newSubscriptions = null
            } else {
                newSubscriptions = arrayOfNulls(n - 1)
                System.arraycopy(oldSubscriptions, 0, newSubscriptions, 0, i)
                System.arraycopy(oldSubscriptions, i + 1, newSubscriptions, i, n - i - 1)
            }
        } while (!_subscriptionsUpdater.compareAndSet(this, oldSubscriptions, newSubscriptions))
    }

    /**
     * Private [SubscriptionReceiveChannel] implementation
     */
    private inner class Subscription : ConflatedChannel<T>(), SubscriptionReceiveChannel<T> {
        override fun onClosed(closed: Closed<T>) {
            removeSubscription(this)
            super.onClosed(closed)
        }
    }

    private companion object {
        private val _subscriptionsUpdater: AtomicReferenceFieldUpdater<SubscribableVariableImpl<*>, Any?> =
                AtomicReferenceFieldUpdater.newUpdater(
                        SubscribableVariableImpl::class.java,
                        Any::class.java,
                        "_subscriptions"
                )

        private val _updatingUpdater: AtomicIntegerFieldUpdater<SubscribableVariableImpl<*>> =
                AtomicIntegerFieldUpdater.newUpdater(
                        SubscribableVariableImpl::class.java,
                        "_updating"
                )
    }
}
```

jcornaz:
>No problem, I looked for a lightweight ConflatedBroadcastChannel, I submit you a draft of it.

Ok, thank you for the precision and the implementation draft. So the goal would be to have a dedicated implementation for `SubscribableVariable`, instead of using `ConflatedBroadcastChannel`?

I'm a bit skeptical about it. Is the benefit provided big enough to justify the complexity it adds in `kotlinx.coroutines`?

I think it should be at least measured, to ensure that the use of `volatile` doesn't kill the benefit of having a *fast path* avoiding to update the atomic variable for example. However, if it appears that this implementation is indeed a good efficiency improvement, why not directly improve the implementation of `ConflatedBroadcastChannel`?

fvasco:
My first idea was without `_updating`, finally I copied that from `ConflatedBroadcastChannel`; without it this implementation is not _synchronized_.
The purpose of a dedicated implementation is to reduce memory footprint of `SubscribableVariableImpl` (24 bytes instead of 64).

> I'm a bit skeptical about it. Is the benefit provided big enough to justify the complexity it adds in kotlinx.coroutines?
Good point, unfortunately I fighted too much against the GC...

jcornaz:
>The purpose of a dedicated implementation is to reduce memory footprint of SubscribableVariableImpl (24 bytes instead of 64).

I admit the lowest memory footprint, the better.

But personally, I wouldn't introduce a dedicated implementation to save only 40 bytes. Because at the end it saves 40 bytes, but is a separate implementation to test and maintain. And it is likely that any futur change in `ConflatedBroadcastChannel` should be reported to `SubscribableVariable` anyway.

Personally, I usually deal with much larger memory footprints, due to data itself. I don't think I'd see a difference in practice for my usage.

But this is a very personal opinion. I'd let decide the maintainers of `kotlinx.coroutines`.

fvasco:
My proposal is not compilable due to https://youtrack.jetbrains.com/issue/KT-13924

jcornaz:
>My proposal is not compilable due to https://youtrack.jetbrains.com/issue/KT-13924

Why? I don't see any need to use `Array::class` in your proposition.

fvasco:
`subscriptionsUpdater` updates an `Array` type, not an `Any`.
However this can be a workaround

```kotlin
private companion object {
    private val _subscriptionsUpdater: AtomicReferenceFieldUpdater<SubscribableVariableImpl<*>, Any?> =
            AtomicReferenceFieldUpdater.newUpdater(
                    SubscribableVariableImpl::class.java,
                    SubscribableVariableImpl<*>::_subscriptions.javaClass,
                    "_subscriptions"
            )

    private val _updatingUpdater: AtomicIntegerFieldUpdater<SubscribableVariableImpl<*>> =
            AtomicIntegerFieldUpdater.newUpdater(
                    SubscribableVariableImpl<*>::_updating.javaClass,
                    "_updating"
            )
}
```

jcornaz:
Ah ok. sorry, I didn't pay enough attention.

So what do we decide? Should I update the PR to include this improvement? Or should I leave the PR as it is without it. (This is anyway implementation detail, and could be improved at any time without making breaking change, because the implementation class is private)

As already stated I'd personally choose to not do it, as it brings unnecessary complexity (for the maintainers of the *kotlinx.coroutines*) and too less benefit (for the users) in my opinion.

However, I'll do it with pleasure if you ask for it. Or even let @fvasco do it, if wants to?

fvasco:
In my opinion the proposed code is ready for merge, so don't change your PR for now.

A dedicated `SubscribableVariableImpl` implementation should be lighter, however my implementation can `offer` one item twice during subscription.
I don't have personal suggestion: the choice is a tradeoff.

So @elizarov  is able to merge this at any moment or propose further changes/investigation.

jcornaz:
I guess I should update the `openSubscribtion()` signature to return `ReceiveChannel` instead of `SubscriptionReceiveChannel`, as this latest interface is planned to be removed: #283

elizarov:
I'm working on #280 and the current version just returns `BroadcastChannel` from `broadcast { ... }` builder and `broadcast()` operator. It seems wrong. We need to declare them to return `Subscribable`. Can we already merge this PR?

jcornaz:
Totally agree, `broadcast` should return a `Subscribable`. Or may be a `CloseableSubscribable` which would be a sub interface of both `Subscribable` and `Cloeaseble`?

However, merging this PR is up to you @elizarov. I'm done with it unless you have change/investigations to request.

elizarov:
So I have this crazy idea toady. Instead of `Subscribable` interface that exposes `openSubscription` do instead `Consumable` interface that exposes `consumeEach` (make it a member). That would let us unify all 3 kinds of abstractions: channels, broadcast channels, and cold streams (#254) and then provide a single set of (efficient implemented!) operators like `fitler`, `map`, etc on the types of streams with one piece of code (#285). I'll leave it for a while here to think though all the implications.

jcornaz:
This is very interesting an sounds promising. However, `consumeEach` is actually specific to an iteration over all values. And some operators (those which don't need to iterate over all values) could be hard to implement efficiently. (For example, `first`, `any`, `contains` etc.)

But I still think it is a good Idea, and that it goes in the good direction.

So here is an alternative proposition: 

What if we make `ChannelIterator` implement `Closeable` and add a`Consumable` interface (as you propose) but exposing a `fun iterator(): ChannelIterator<E>` function instead of `consumeEach` ?

Here is a quick draft of the idea: https://github.com/Kotlin/kotlinx.coroutines/compare/master...jcornaz:spike/consumable

elizarov:
Iterators are inherently slow. My idea for `consumeEach` was that terminal operations that do not need to scan all elements would throw `CancellationException` when they are done. We can use a shared instance and that should be pretty efficient, however I'll need to do more benchmarking around this particular case.

smaldini:
Sounds like something we are planning to add in reactor-core as an extension. fromAsyncIterable/toAsyncIterable (in addition to fromCoroutine/toCoroutine). We might be able to use that new contract instead of creating a new AsyncIterable of our own.

fvasco:
I wrote a little example using `exit` to stop consuming.

```kotlin
interface Consumable<E> {
    fun consumeEach(consumer: suspend ConsumerScope<E>.(E) -> Unit): E?
}

interface ConsumerScope<E> {
    suspend fun exit(exitValue: E? = null): Nothing
}

fun Consumable<Int>.sum(): Int {
    var sum = 0
    consumeEach { sum += it }
    return sum
}

fun <E> Consumable<E>.first() = consumeEach { exit(it) }
```

elizarov:
We clearly need a prototype to make this decision. It is tightly bound to the decision on cold streams abstraction in #254, since it is basically _the same_ "subscribable" abstraction that we are talking about here. It does not look that we have place for two different abstractions here. We already have too many of them.

The proof of concept should demonstrate that all the interesting operators like `filter`, `map`, `first`, `take`, etc that are currently defined as extensions to `ReceiveChannel` can be rewritten as extension to the new subscribable/coldStreams abstraction while keeping the code simple. We also need POC to measure the performance of the resulting code.

SolomonSun2010:
Are you going to support  _Subscribable_ as dataflow concurrency in Oz ?https://www.info.ucl.ac.be/~pvr/GeneralOverview.pdf

I like it , Data-Driven Concurrency is cool !
Also, may I propose to rename these names similar with Groovy GPars ?
http://gpars.org/webapp/quickstart/index.html#__strong_dataflow_concurrency_strong
@jcornaz @fvasco @elizarov 

see also:
Building Scalable, Highly Concurrent & Fault Tolerant Systems - Lessons Learned
https://www.slideshare.net/jboner/building-scalable-highly-concurrent-fault-tolerant-systems-lessons-learned?from_action=save

Dataflow Concurrency
• Deterministic
• Declarative
• Data-driven
• Threads are suspended until data is available
• Lazy & On-demand
• No difference between:
• Concurrent code
• Sequential code
• Examples: Akka & GPars

jcornaz:
@SolomonSun2010 data-flow concurrency is already possible with channels. And there is active design discussion about how to provide *cold stream* in #254.

qwwdfsad:
When #254 is almost here, I think it is time to close this PR.

During our design phase, we have a lot of discussions on whether `Channel` should implement `Flow` or not and decided not to; and this separate interface carries more burden than a value when living side-by-side with `Flow`.

There is also a natural desire to use channels as subjects and for this, we have `Flow.broadcastIn` and `BroadcastChannel.asFlow`, so I'd better focus our efforts on shaping this API

fvasco:
No problem, all these consideration can be cherry picked when needed.

PR #294, opened 2018-03-20
fix: build fails if JDK_16 is not set

venkatperi:


jcornaz:
Wouldn't it better to simply specify in the readme that it is necessary to specify a `JDK_16` environment variable?

elizarov:
We cannot just skip the requirement for `JDK_16`. It has to be built against JDK 1.6 runtime in order to be  compatible with JDK 1.6. We should give a clear error message when `JDK_16` is not.

venkatperi:
Closing. Missing JDK_16 should throw an error.

Issue #278, opened 2018-03-12
Provide ExecutorService.asCoroutineDispatcher that returns a closeable dispatcher


elizarov:
This issue in inspired by this SO question: https://stackoverflow.com/questions/49217668/shut-down-the-underlying-executor-of-executorcoroutinedispatcher

The idea is to provide `ExecutorService.asCoroutineDispatcher` extension in addition to `Executor.asCoroutineDispatcher` that would return a subclass of `CoroutineDispatcher` that also implements `Closeable` interface and delegates `close` to `shutdown` of the underlying `ExecutorService`.

elizarov:
Thanks. I've merged it the the develop branch. I've made a small optimization. Instead of creating a separate class that implements CloseableCoroutineDispatcher, I've made the original class implement it.

PR #295, opened 2018-03-20
ExecutorService extension that return closable CoroutineDispatcher #278

deva666:
#278


elizarov:
Thanks. I've merged it the the develop branch. I've made a small optimization. Instead of creating a separate class that implements CloseableCoroutineDispatcher, I've made the original class implement it.

PR #296, opened 2018-03-21
 Fail if JDK_16 is not set

venkatperi:


elizarov:
Thanks. I'm made some stylistic changes in readme and in the build file (it should fail only when trying to build, not when building effecting project model) and merged into the `develop` branch.

PR #299, opened 2018-03-22
Avoid `moreTest` task on `gradle build -x test` build

fvasco:
Allow build without test if `JDK_16` environment variable is not set ( #291)


elizarov:
I've tried doing it like this. The problem is that you loose ability to use Gradle test runner IDEA properly and otherwise run single test from Gradle, because with this change `gradle test --tests <test-name>` runs both `test` and `moreTest`. I don't know a workaround, so I've defined `moreTest`. The downside is that if you want to run w/o tests you have to do `grade build -x test -x moreTest`

qwwdfsad:
Current workaround is to make JDK_16 point to JDK_HOME

Issue #283, opened 2018-03-13
Remove SubscriptionReceiveChannel


elizarov:
`SubscriptionReceiveChannel` interface is a legacy artifact. Currently every `ReceiveChannel` has `cancel` method that should be invoked after this channel was consumed. We should deprecate and remove `SubscriptionReceiveChannel `.

PR #311, opened 2018-04-04
Remove SubscriptionReceiveChannel

deva666:
#283 Issue says that `SubscriptionReceiveChannel` should be removed, so this PR is a breaking change. Not sure if this is wanted?


elizarov:
That is Ok. It is supposed to be breaking change in the source. However, for binary compatibility we need to keep old `SubscrpitonReceiveChannel` interface (as deprecated), keep old `openSubscription` function (as deprecated/hidden), so new `openSubscription` will probably have to use a different `@JvmName`.

Also, can you please rebase your changes onto the latest `develop` branch.

deva666:
@elizarov not sure how you can apply `JvmName` annotation in this case. And of course, there can't be two functions with different return types. If we deprecate the old `openSubscription` function can the new one be named differently, something like `openReceiveChannel` ?

elizarov:
Hm... I checked. Indeed, I cannot seem to find a way to preserve binary compatibility when changing return type. This is bad: https://youtrack.jetbrains.com/issue/KT-23750

elizarov:
I've found a way to preserve binary compatibility. Merged this PR into `develop`.

PR #313, opened 2018-04-09
updated dependencies for android sample apps

Dmitry-Borodin:
updated dependencies for android sample apps to up to date

### DIFF
--- a/ui/kotlinx-coroutines-android/animation-app/gradle.properties
+++ b/ui/kotlinx-coroutines-android/animation-app/gradle.properties
@@ -18,6 +18,6 @@ org.gradle.jvmargs=-Xmx1536m
 
 kotlin.coroutines=enable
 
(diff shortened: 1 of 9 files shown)


ui/kotlinx-coroutines-android/animation-app/gradle.properties
```diff
@@ -18,6 +18,6 @@ org.gradle.jvmargs=-Xmx1536m
 
 kotlin.coroutines=enable
 
-kotlin_version = 1.2.21
```
Dmitry-Borodin: this actually not used to define version of library used. We pick library version from Build.gradle file from top project kotlinx-coroutines-android.
Is that intentional?
Dependencies usually are not defined in gradle.properties, but in Build.gradle with other build configuration.

elizarov:
Thanks. Merged into `develop` branch.

PR #323, opened 2018-04-11
awaitAll and joinAll extensions

### LINKED ISSUE #171: Introduce awaitAll and joinAll extensions for collections for Deferreds/Jobs
`joinAll` is simple `map { it.join() }`, while `awaitAll` needs a slightly more involved implementation than a simple `map { it.await() }`. It shall actually wait on all of the deferreds from collection and crash as soon as any one of them crashes.

qwwdfsad:
Fixes #171


core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Await.kt
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Awaits for completion of given jobs without blocking a thread and resumes normally when all jobs computations are complete
+ * or resumes with the first thrown exception if any of computations completes exceptionally including cancellation.
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is l or completed while this suspending function is waiting,
+ * this function immediately resumes with [CancellationException].
+ */
+public suspend fun Iterable<Job>.awaitAll(): Unit {
+    val jobs = this as? Collection<Job> ?: this.toList()
```
ilya-g: What if `jobs` collection is changed after calling `.awaitAll()` extension on it?

qwwdfsad: Depends on mutation. In the worst case, `awaitAll` will hang (if concurrent mutator removes an element while `awaitAll` prepares its handles), but a non-trivial effort is required to write such code.

Behaviour in such case is undefined, what does user expect if he creates a collection of jobs, shares it and mutates in one coroutine while awaiting them in another?

elizarov: So, semantically we want `awaitAll` to operate on a some kind of snapshot. The challenge is how far we are ready to go for an optimal implementation. The easy, but not efficient, way it so to always take a snapshot of collection first, then await on it.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Await.kt
```diff
@@ -20,7 +20,8 @@ public suspend fun awaitAll(vararg jobs: Job): Unit = jobs.asList().awaitAll()
  */
 public suspend fun Collection<Job>.awaitAll() {
     if (isEmpty()) return
-    AwaitAll(this).await()
+    val snapshot = ArrayList(this)
```
ilya-g: Better to use toList, it can exploit collection immutability, when we have one.

core/kotlinx-coroutines-core/src/main/kotlin/kotlinx/coroutines/experimental/Await.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.atomicfu.atomic
+
+/**
+ * Awaits for completion of given jobs without blocking a thread and resumes normally when all deferred computations are complete
+ * or resumes with the first thrown exception if any of computations completes exceptionally including cancellation.
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting,
+ * this function immediately resumes with [CancellationException].
+ */
+public suspend fun awaitAll(vararg jobs: Job): Unit = jobs.asList().awaitAll()
```
ilya-g: This list can be awaited without taking snapshot.

elizarov:
Merged!

qwwdfsad:
Hi @olgachotinun 
We're almost ready for the next release, please stay tuned.

>if You could update the coroutines guide saying awaitAll is coming

We can't really do it, having "what's coming in the next release" in the guide is slightly confusing, especially for "planned" public API which is frequently changed during initial implementation

Issue #219, opened 2018-02-05
CancellableContinuation should not be a Job


elizarov:
`CancellableContinuation` interface should cease to extend `Job` interface. `Job` is too "heavy weight". Job has a concept of "children" and "completing" state where it waits for its children to complete. None of that is applicable to `CancellableContinuation`. It has much simpler state-machine. Thus, `Job` has many methods and extensions that are not truly applicable to `CancellableContinuation`.

From implementation standpoint, `CancellableContinuation` has to extend `JobSupport` for all that Job machinery which is not needed. This complicates the code itself and makes it harder to performance-optimize `CancellableContinuation` to its specific use-cases.

`CancellableContinuation ` will still keep many methods from the `Job`, notably `invokeOnCompletion`, `cancel` and `isActive`/`isCompleted`/`isCancelled`.

PR #334, opened 2018-04-17
Decouple AbstractContinuation and CancellableContinuation from Job interface

qwwdfsad:
Decouples `CancellableContinuation` from `Job`.

Pros:
1) Some workarounds from Job may be removed
2) State machine of `AbstractContinuation` becomes much simpler
3) `AbstractContinuation` state is now more compact and start/cancel produce less garbage (and state transitions are more lightweight)

2, 3 => all integration modules become faster (~25% win in cancellation benchmark).

Cons:
1) Binary compatibility may be broken in some places
2) Compatibility **is** broken due to changes in `invokeOnCompletion` contract
3) Small maintenance mess ("similar, but not the same" code)

In general, it boosts the performance of cancellable continuations, allows to do specific optimizations (e.g., it makes sense to get rid of `resumeMode` after merging this change and replace it with inheritance to reduce memory footprint even more) and allows us to move `CancellableContinuation` to stdlib


fvasco:
Release with https://github.com/Kotlin/kotlinx.coroutines/pull/311

qwwdfsad:
Fixes #219

PR #335, opened 2018-04-18
Implement ticker channel

qwwdfsad:
Delay channel, the basic building block for time-based operators.
Complex example of usage may be found in tests:

```
private fun ReceiveChannel<Int>.averageInTimeWindow(timespan: Long) = produce {
        val delayChannel = DelayChannel(delay = timespan, initialDelay = timespan)
        var sum = 0
        var n = 0
        whileSelect {
            this@averageInTimeWindow.onReceiveOrNull {
                when (it) {
                    null -> {
                        // Send leftovers and bail out
                        if (n != 0) send(sum / n.toDouble())
                        false
                    }
                    else -> {
                        sum += it
                        ++n
                        true
                    }
                }
            }

            // Timeout, send aggregated average and reset counters
            delayChannel.onReceive {
                send(sum / n.toDouble())
                sum = 0
                n = 0
                true
            }
        }

        delayChannel.cancel()
    }
```

Opinions on how else we can name `DelayChannel` and `FixedDelayChannel` for better discoverability are weclome


fvasco:
Invoke `delayChannel.cancel()` is optional, right?

`DelayChannel` looks good, but it may be `DelayProducer`.

PR #349, opened 2018-05-01
Allow to change the default dispatcher

soywiz:
I want to upgrade my libraries to use kotlinx.coroutines. But my main concern, and the only thing that is blocking me at this point, is that right now `launch` blocks (and anything using the 
`DefaultDispatcher` as default), by default use a ThreadPool on the JVM.

Use case: In my case I use a Node.JS/JS approach for building common projects:  the application login runs in a single thread. That prevents having to think about synchronizing non async primitives in the final user code and provides consistency between JS and JVM targets.

This behaviour might not be what people want, but what some people will want. So my proposal here is to allow to change the default dispatcher so people doing libraries or final code can chose what to do without forcing final users to specify always a dispatcher that is also prone to error.

On my side I would want to implement something like this:

```
fun main(args: Array<String>) = EventLoop {
    ...
}

// Or call this something like `SingleThread`?
fun EventLoop(callback: suspend () -> Unit) {
    val eventLoop = EventLoop()
    temporalSetDefaultDispatcher(eventLoop) {
        callback.startCoroutine(object : Continuation<Unit> {
            override val context: CoroutineContext = eventLoop
            override fun resume(value: Unit) = Unit
            override fun resumeWithException(exception: Throwable) = println(exception)
        })
        eventLoop.execute() // Loop in JVM targets
    }
}

inline fun temporalSetDefaultDispatcher(temporalDispatcher: CoroutineDispatcher, block: () -> Unit) {
    val oldDispatcher = DefaultDispatcherProvider
    DefaultDispatcherProvider = { temporalDispatcher }
    try {
        block()
    } finally {
        DefaultDispatcherProvider = oldDispatcher
    }
}
```

Note: EventLoop implements CoroutineDispatcher and Delay. Though Delay right now not provides a common interface and it is problematic for people trying to implement it in common code (not just calling delay).

I put a provider instead of just a simple global variable so I could later change it per thread. For example by being able to provide a default EventLoop per thread instead of only one global.

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/main/kotlin/kotlinx/coroutines/experimental/CommonCoroutineDispatcher.kt
+++ b/common/kotlinx-coroutines-core-common/src/main/kotlin/kotlinx/coroutines/experimental/CommonCoroutineDispatcher.kt
@@ -17,6 +17,7 @@
 package kotlinx.coroutines.experimental
 
 import kotlin.coroutines.experimental.*
+import kotlin.jvm.Volatile
 
 public expect abstract class CoroutineDispatcher constructor() : AbstractCoroutineContextElement, ContinuationInterceptor {
     public open fun isDispatchNeeded(context: CoroutineContext): Boolean
@@ -32,6 +33,7 @@ public expect val DefaultDefaultDispatcher: CoroutineDispatcher
 
 private fun getDefaultDefaultDispatcherHelper() = DefaultDefaultDispatcher
 
+@Volatile
 var DefaultDispatcherProvider: () -> CoroutineDispatcher = { getDefaultDefaultDispatcherHelper() }
 
 @Suppress("PropertyName")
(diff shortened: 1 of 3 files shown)


qwwdfsad:
We can't accept this PR for multiple reasons.
1) Right now `DefaultDispatcher` (and dispatching mechanism in general) is being reworked and it's not obvious how this change will interfere with that.
2) Changing default dispatcher is very dangerous operation.
Consider you're writing a library with method `foo`, which is written in the following way:
```
fun foo(someUserCallback: Runnable) = withTemporalDispatcher(LibraryDispatcher) {
  // Library code
  someUserCallback.invoke()
  // Library code
}
```
Now user's callback will be invoked in `LibraryDispatcher` and may block it indefinitely, causing heisenbugs.
Changing any implicit behaviour (and `DefaultDispatcher` **is** implicit behaviour) make written code less understandable and frequently becomes the source of non-obvious bugs due to different components/libraries interference.


>So my proposal here is to allow to change the default dispatcher so people doing libraries or final code can choose what to do without forcing final users to specify always a dispatcher **that is also prone to error.**
True. We're trying to fix that with a different approach (currently only in design draft): https://kotlinlang.slack.com/archives/C1CFAFJSK/p1524842434000397?thread_ts=1524835325.000240&cid=C1CFAFJSK


For `@Volatile` in common code we have `kotlinx.coroutines.experimental.internalAnnotations.Volatile` as workaround for `https://youtrack.jetbrains.com/issue/KT-23727`

Issue #340, opened 2018-04-19
Calling launch in kotlinjs causes a TypeError to be logged


ansman:
If you compile the following code and execute it:
```kotlin
fun main(args: Array<String>) {
  launch {}
}
```
You get the following exception printed to the terminal:
```
TypeError: throwCCE is not a function
```

qwwdfsad:
I can't reproduce this problem on my machine.
Could you please provide more details (node.js version, Kotlin version, build script etc.) or self-contained project example?

ansman:
Sure, no problem. I've attached a same project.

Just cd in to the dir and run `npm run bug`

[coroutines-bug.zip](https://github.com/Kotlin/kotlinx.coroutines/files/1935819/coroutines-bug.zip)

ansman:
Seems like it could be solved by fixing #339 then

SUPERCILEX:
@elizarov any idea when this will be released? Or are there snapshot builds we could use? Coroutines are basically unusable in JS, seems like a big deal. 😕

PR #350, opened 2018-05-01
Change kotlin to be a peer dependency

### LINKED ISSUE #339: kotlin should be defined as a peer dependency to kotlinx-coroutines-core
Currently kotlin is defined as a regular dependency which means that it's not using the same runtime as your app module and things like `instanceof` might not work as expected.

ansman:
This fixes #339 and #340


elizarov:
I've rebased it onto `develop` branch. Thanks.

ansman:
Any idea when this might be released? As it is right now I can't really use coroutines in javascript.

Issue #336, opened 2018-04-18
kotlinx-coroutines-test modules


pull-vert:
It would be nice to have separate modules dedicated to Tests, so external libraries can use it to test Coroutines by importing it like this in gradle (example for jvm) :
testCompile "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinx_coroutines_version"

This module would provide TestBase class, containing runTest builder and useful methods to track good execution in coroutines.

A MPP would be perfect (common, js and jvm, and why not native too).

Edit : it would be nice if this module can include TestCoroutineContext too (#297)

PR #355, opened 2018-05-05
kotlinx-coroutines-test modules

pull-vert:
add 3 new dedicated kotlinx-coroutines-test projects (common, JVM and JS)
Solves #336


elizarov:
I've reviewed this PR and I have few questions or concerns.

There is a number of APIs that were previously `internal` in core (like `resetCoroutineId`, `CommonPool.usePrivePool`, `DefaultExecutor.ensureStarted`, etc). They are now `public` which raises a number of issues. The most pressing issue is that they are not documented, but there is also a question on whether these low-level methods and classes should be exposed at all. I feel that some kind of higher-level test-oriented functionality to "setup" and "clear" the library before/after tests should be provided instead.

`TestBase` class itself provides both general-purpose testing facilities (`expect` and `finish`) as well as very project specific things like `isStressTest` and `stressTestMultiplier`. I think that they should be decoupled somehow.

Issue #359, opened 2018-05-14
Lost send in Channel during close


qwwdfsad:
The following code may lose `send` coroutine:

T1: `channel.send()`
T2 `channel.receive(); channel.close()`

`close` is not atomic and can be considered as two actions: "remove and cancel waiting receivers" and "add closed token". If `send` occurs between these actions, `send` will be stuck in the channel

Issue #260, opened 2018-02-27
Introduce job property for channels


elizarov:
When data is being consumed by an `actor` there is sometimes a need to wait until all the messages are processed after a channel was closed. Currently this requires the following code patten:

```
val job = Job() // create parent job
val actor = actor(parent = job) { ... } 
// ... work with actor
actor.close() // close an actor
job.join() // wait for actor's completion
```

The proposal is to introduce `ActorSendChannel<T>` interface that extends `SendChannel<T>` interface and introduces `job` property that directly gives access to actor's job, so that this code can be written in a slightly more straightforward way:

```
val actor = actor { ... } 
// ... work with actor
actor.close() // close an actor
actor.job.join() // wait for actor's completion
```

It makes sense to also introduce `ProducerReceiveChannel` for a symmetry, albeit its usefulness is not immediately apparent at the moment.

Note, that in some older version of `kotlinx.coroutines` library there ware an `ActorJob<T>` and `ProducerJob<T>` interfaces that extended both `SendChannel<T>`/`ReceiveChannel<T>` and `Job`. It served a similar role, but mixing both channel and job methods onto a single receiver had proved harmful (see #127). This issue basically reboots this idea, but in cleaner way, by moving job into a separate property.

fvasco:
A tiny consideration: the `Job` exposes many other useful informations (`context` and `children`s).
Is is possible to achieve the same goal introducing the `flush()` method into the `SendChannel` interface.

Error: `job.await -> job.join`

elizarov:
@fvasco Thanks for `join` correction. 

`flush()` concept is interesting, however I find it somewhat counter-initive that it actually "waits". Usually, `flush` (on a socket, for example) is associated with "send data to the underlying transport", but it does not actually wait for data to get physically transferred, yet alone processed. 

Also note, that we are looking for a common nomenclature between regular (object) channels and byte-channels that are work-in-progress in `kotlinx-coroutines-io`. There we already have a concept of `flush` that exists there because all byte channels are buffered and, unlike object channels, writing a few bytes does not immediately resume receiving coroutine (for performance reasons), but only stores bytes in the buffer and you must `flush` your byte channel to resume the coroutine on the other end of your channel.

In order for the `flush` concept to work on a regular `SendChannel` we'll need to figure out:
* A different name (`awaitProcessed`?)
[...]

PR #362, opened 2018-05-16
Channel fixes

qwwdfsad:
Channel `close` linearizability
`onClose` method
Fixes for #359 and #260


common/kotlinx-coroutines-core-common/src/main/kotlin/kotlinx/coroutines/experimental/channels/AbstractChannel.kt
```diff
@@ -243,23 +243,53 @@ public abstract class AbstractSendChannel<E> : SendChannel<E> {
...
+
+        helpClose(closed)
+        onClosed(closed)
+        afterClose(cause)
+        // Cancel it as the last action so if the channel is closed, then the job is cancelled as well
+        job.cancel(cause)
+        return true
+    }
+
+    private fun helpClose(closed: Closed<*>) {
```
qwwdfsad: Discuss: `helpClose` should probably try to cancel job as well, so it will be linearizable along with channel

LouisCAD: Given the current implementation doesn't try to cancel the job, could this lead to a deadlock if somehow, a `runBlocking { ... }` call is waiting for it because it'd be waiting for the job to complete?

qwwdfsad: The current implementation doesn't have a job, this branch is not going to `master`.

common/kotlinx-coroutines-core-common/src/test/kotlin/kotlinx/coroutines/experimental/channels/TestChannelKind.kt
```diff
@@ -16,54 +16,55 @@
...
         override fun toString(): String = "RendezvousChannel"
     },
     ARRAY_1 {
-        override fun create(): Channel<Int> = ArrayChannel(1)
+        override fun create(job: Job): Channel<Int> = ArrayChannel(1, job)
         override fun toString(): String = "ArrayChannel(1)"
     },
     ARRAY_10 {
-        override fun create(): Channel<Int> = ArrayChannel(8)
+        override fun create(job: Job): Channel<Int> = ArrayChannel(8, job)
```
LouisCAD: `ARRAY_10` with a size of 8??

LouisCAD: Reported in #438

Issue #119, opened 2017-09-13
Work with ThreadLocal-sensitive Components


ian4hu:
Many application trace framework use `ThreadLocal` to store the call context of a application, like the [SLF4j MDC](https://logback.qos.ch/manual/mdc.html), [CAT ](https://github.com/dianping/cat), and other distributed trace system(which use thread local for trace the context in some async situation).

But the coroutine will be dispatched to a unspecified thread to execute, how can i make the tracing work?

elizarov:
Coroutine's analog to `ThreadLocal` is `CoroutineContext`. 

To interoperate with ThradLocal-using libraries you need to implement a custom `ContinuationInterceptor` that supports framework-specific thread-locals. Here is an example. Assume that use some framework that relies on a specific `ThreadLocal` to store some application-specific data (`MyData` in this example):

```
val myThreadLocal = ThreadLocal<MyData>()
```

To use it with coroutines, you'll need to implement a context that keeps the current value of `MyData` and puts it the corresponding ThreadLocal every time the coroutine is resumed on a thread. The code should look like this:

```
class MyContext(
    private var myData: MyData,
    private val dispatcher: CoroutineDispatcher
) : AbstractCoroutineContextElement(ContinuationInterceptor), ContinuationInterceptor {
    override fun <T> interceptContinuation(continuation: Continuation<T>): Continuation<T> =
        dispatcher.interceptContinuation(Wrapper(continuation))

    inner class Wrapper<T>(private val continuation: Continuation<T>): Continuation<T> {
        private inline fun wrap(block: () -> Unit) {
            try {
                myThreadLocal.set(myData)
                block()
            } finally {
                myData = myThreadLocal.get()
            }
        }

        override val context: CoroutineContext get() = continuation.context
        override fun resume(value: T) = wrap { continuation.resume(value) }
        override fun resumeWithException(exception: Throwable) = wrap { continuation.resumeWithException(exception) }
    }
}
```

To use it in your coroutines, you wrap the dispatcher that you want to use with `MyContext` and give it the initial value of your data. This value will be put into the thread-local on the thread where the coroutine is resumed.  

```
launch(MyContext(MyData(), CommonPool)) {
    // do something...
}
```

The implementation above would also track any changes to the thread-local that was done and store it in this context, so this way multiple invocation can share "thread-local" data via context.

[...]

PR #403, opened 2018-06-20
implementation of a SLF4J MDC context

abendt:
We use the CoRoutines rxjava integration and stumbled across the necessity to make the SLF4J MDC Context within in the CoRoutine scope. We implemented a solution based on the the description here:
https://github.com/Kotlin/kotlinx.coroutines/issues/119

### DIFF
--- a/integration/kotlinx-coroutines-slf4j/README.md
+++ b/integration/kotlinx-coroutines-slf4j/README.md
@@ -1,0 +1,19 @@
+# Module kotlinx-coroutines-slf4j
+
+Integration with SLF4J [MDC](https://logback.qos.ch/manual/mdc.html).
+
+## Example
+
+Extends a given Coroutine context so that the SLF4J MDC context is passed into the coroutine.
+
+```kotlin
+MDC.put("kotlin", "rocks") // put a value into the MDC context
+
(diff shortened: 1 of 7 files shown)


oshai:
In case this is merged will add it to [kotlin-logging](https://github.com/MicroUtils/kotlin-logging) docs

x0a1b:
I think it's a really useful addition I am not sure if this change should be merged here or go to Kotlin-logging (unless it's using some internal classes from coroutines). I strongly think getting it in kotlin-logging rather than in core is the right option; and would let us iterate faster.

abendt:
fair enough, i created the PR here because of the suggestion of @elizarov: https://github.com/Kotlin/kotlinx.coroutines/issues/119
kotlin-logging also looks like a reasonable location

oshai:
I think that both here and in kotlin-logging this will be an integration module. @elizarov suggested: "We should really have it out-of-the box in a separate integration module." in #119 . 
In addition, kotlinx.coroutines already has other integration modules that are similar so it seems like a more natural place to look for, and also a case for example for the lib users.

elizarov:
I've rebased your code on top of new architecture to introduce thread-local context elements as explained in #119 and opened a new PR #455

Issue #167, opened 2017-11-20
Fail-fast when trying to consume a channel multiple times


elizarov:
Channels are designed to be used in fan-out fashion or in a pipeline fashion. Pipeline is conveniently performed with various operators like `filter`, `map`, etc (see PR #88) which are modeled after a `Sequence`. However, a sequence is a multi-shot abstraction, for the following is perfectly file:

```
val s: Sequence<T> = ...
s.filter { ... }.forEach { it.doSomething() }
s.map { ... }.forEach { it.doSomethingElse() }
```

However, a similar code will not work in the same way with a `ReceiveChannel`. `ReceiveChannel` is more like `Iterator`, than it is like a `Sequence`. Each element of the `ReceiveChannel` can be consumed only once. It is expected that this might be a common source of errors and misunderstanding, so the proposal is to make all the standard operators "fail-fast" when the channel is already being consumed by another operator. So, in the above example, if the types of `s` was to be replaced with `ReceiveChannel<T>`, then the invocation of `s.map` (which comes after `s.filter`) shall fail immediately. 

P.S. It is a separate issue whether a truly `Sequence`-like (Rx-like) higher-order abstraction shall be introduced for channels.

ZakTaccardi:
just want to confirm that the following would still work:

```
val channel: ReceiveChannel<T> = ...

channel.filter { .. }
    .map { .. }
    .consumeEach { .. }

```

bdavisx:
I'm confused by the description, will consumeEach only be able to be called once? If so how would you do a fan-out?

elizarov:
@bdavisx Thank you for observation. This indeed can be confusing and we should think on how to make it less confusing. They idea is that for fan-out you don't _consume_ the channel. Instead, you simply invoke `receive()` to receive elements from it as you need.

bdavisx:
Thanks and yes, very confusing because the [current guide/examples fan out section](https://github.com/Kotlin/kotlinx.coroutines/blob/master/coroutines-guide.md#fan-out) doesn't use `receive()`, it uses multiple calls to `channel.consumeEach()`.

bdavisx:
I changed the doc to use an `iterator` instead of `consumeEach()` - and in the end it looks like the iterator uses `receive()`. I think this is a better example, but I'm not sure.

https://github.com/Kotlin/kotlinx.coroutines/pull/406

jcornaz:
[...]

PR #406, opened 2018-06-22
Updated the guide fan-out section to use channel iterator

bdavisx:
The `consumeEach` function has a warning about multiple consumption but the `iterator` uses `receive()` in the end, so I think it's a better way to handle this.

See discussion around #167


elizarov:
Thanks. Pushed onto `master`.

Issue #94, opened 2017-07-31
Read-write Mutex


fvasco:
Is it in the roadmap?

elizarov:
It is on my to-do list, but not for the short-term. It is really up for grabs. This is a great feature for someone, who wants to figure how to write low-level lock-free implementation for `kotlinx.coroutines` primitives, to go ahead, study `Mutex` implementation, and figure out how to abstract (reuse) parts of that to implement `ReadWriteMutex`. It would be a pity if I write it myself and deprive someone else from this great learning opportunity.

bobvawter:
Here's a basic implementation of the 80% case for `ReadWriteMutex` built on top of `Mutex` where you just want to execute coroutines as readers or writers.

https://gist.github.com/bobvawter/4ff642d5996dfccb228425909f303306

fvasco:
The proposed implementation misses of support for "select" clause.

elizarov:
Adding "select" in not a critical problem. I have a different concern here -- `stateLock` approach does not seem efficient to me for a core synchronization primitive. The act of acquiring read lock should be fully lock-free for scalability purposes, so that it scales to multi-core machines where a lot of threads can be acquiring and releasing read lock concurrently.

bobvawter:
@elizarov If you agree with the general design of that ReadWriteMutex API, I can work on making the implementation more like `Mutex` (using an atomic queue of operations), adding additional state-management methods (e.g. `tryReadLock()`), and then put together a PR in the next couple of weeks.

Two questions before I start:
* Should state-change functions like `readLock()` accept an `owner` argument?
* If a user were to pass the same object twice to `readLock(owner:Any?)`, should that call succeed?

I can see arguments for `owner` representing a specific actor (possibly reentrant or concurrent), or for `owner` representing a specific read operation.  The former would look like some internal structure similar to `ConcurrentMap<Any, Int>` and the latter just a `ConcurrentSet<Any>`.

jcornaz:
@elizarov, I don't understand your statement:
>The act of acquiring read lock should be fully lock-free

How is it possible to guarantee that there will never be a write-lock and a read-lock open at the same time without locking something when we acquire a read-lock ?

I am not aware of other algorithms than theses two: https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock#Implementation

And neither of them alow to acquire a readlock in a lock-free manner.

[...]

PR #408, opened 2018-06-26
efficient read-write lock (#94) and semaphore implementations

btwilk:
These read-write lock and semaphore implementations are completely non-suspending in happy cases: i.e., when there are no writers or when permits are available.

The implementations are missing support for select, graceful handling of cancellation, ownership tracking, and stack-overflow avoidance for the Unconfined dispatcher.

This contribution includes a StateMachine abstraction which encapsulates a lot of the complexity of both implementations. It could eventually be used to simplify Mutex as well.

Addresses issue #94.


bohsen:
No tests?

btwilk:
Is there anything I can do to attract a reviewer for this PR?

elizarov:
P.S. Please, squash your contribution into a single commit to make it easier to review.

elizarov:
As for specific comments:

1. There is no need for a new `Lock` interface. We already have `Mutex` interface in this library. We only need `ReadWriteMutex` (that is how it should be named). The `Lock` name would conflict with `java.util.concurrent.locks.Lock`.

2. There is not need for a dedicated `StateMachine` class that boxes atomic ref inside of it. You can simply define `transition` function as an extension function on `AtomicRef<Any>` type. I found actually recommend to use any existing `loop` and `update` extension functions on `AtomicRef`. You don't need `Decision.Stay` and you can always just do a non-local return from an enclosing function. The whole `Decision` class (and the overhead with creation of its instances) should be removed. 

3. Please, abstract away all the existing tests on `Mutex` implementation and make sure that all pass on the `write` part of `ReadWrite` mutex and on the 0/1 `Semaphore` as well.

4. The code and tests shall be placed into `kotlinx-coroutines-core-common` module. It shall pass all the test both on Kotlin/JVM and on Kotlin/JS.

btwilk:
Thanks for the feedback. I switched to using the Mutex interface and the loop method of AtomicRef as in MutexImpl.

For the read lock I'm just throwing exceptions for anything related to owners.

I'm not too happy with how I addressed 3 as I had to override some of the tests for certain implementations. Let me know if you have any better ideas.

clarcharr:
This is awesome; thank you for the hard work! Any idea when this will be done/merged?

clarcharr:
Any plans to work on this soon?

btwilk:
I'm ready to make more progress if I get some more guidance on requirements.

qwwdfsad:
I am sorry for this delay; I will try to spend more time on reviewing PRs or at least on providing feedback as soon as possible even if the issue is not on the priority list.

Do you still want to work on this PR?
If so, please rebase on current`develop` branch and I will review it.

I apologize for this delay.

PR #409, opened 2018-06-26
Clarify RX subscription and cancellation example

rocketraman:
Clarify that `use` is called automatically rather than needing to be called manually. Demonstrate stream "completion".


elizarov:
Thanks. I've further clarified some terms, performed `gradlew knit` to update sample code according to updated text and pushed amended commit to the `master`.

PR #423, opened 2018-07-05
Use parent = contextJob instead of contextJob + UI

mtopolnik:
As discussed on [Stack Overflow](https://stackoverflow.com/questions/51028384/whats-the-difference-between-a-job-ui-context-and-using-parent-job), the idiom `actor(contextJob + UI)` got first-class support in the form of `actor(UI, parent = contextJob)`. This PR updates the docs to reflect this improvement.

Since the `parent` parameter is nullable, this change allows us to also simplify the definition of `contextJob` and avoid the usage of `NonCancellable` as the explicit substitution for `null`.


qwwdfsad:
Thanks!

PR #425, opened 2018-07-06
Provide HandlerContext.immediate extension to immediately execute tas…

### LINKED ISSUE #381: Android UI dispatcher and isDispatchNeeded
Could we change implementation of method `HandlerContext.isDispatchNeeded()`  to be more flexible:
```kotlin
public class HandlerContext(
    private val handler: Handler,
    private val name: String? = null,
    val alwaysDispatch: Boolean = true
) : CoroutineDispatcher(), Delay {

    override fun isDispatchNeeded(context: CoroutineContext): Boolean {
        return alwaysDispatch || Looper.myLooper() == handler.looper
    }

}
```

qwwdfsad:
…ks which are invoked from the right context

Fixes #381


elizarov:
@JakeWharton what is your concern about this _particular_ solution of providing `UI.immediate` extension that would avoid additional dispatch onto the looper if the current is the corresponding looper's thread?

JakeWharton:
Conditional threading is a giant code smell to me. When you define clear threading semantics of the layers of your application this is never needed. This "optimization" is one that becomes required only when you have ill-defined threading at the production site. The same was true in every proposal for an Rx scheduler that wanted to add this feature and it was solved with no library changes by fixing threading upstream rather than downstream which wound up creating more clear threading semantics for consumers.

elizarov:
I agree that this is an optimization. I've added a link to quite a lengthy documentation of why you should not do that, hopefully to deter most of the people.

PR #435, opened 2018-07-13
added note about FIFO order of `Mutex.lock()`

Groostav:
- Note that the OpenJDK [documents locks as "fair" when they release "the longest waiting thread"](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/concurrent/locks/ReentrantLock.java#L55)


elizarov:
Thanks. Rebased and pushed onto `develop`.

PR #442, opened 2018-07-19
Fix documentation

SahilLone:
The PR intends to improve the documentation for the segment `lazy couroutines`
The documentation already in place does not fully communicate the usage of lazy coroutines


elizarov:
Thanks for the idea. I've updated working a bit and merged it into the `develop` branch.

PR #453, opened 2018-07-24
Job exceptions rework

qwwdfsad:


qwwdfsad:
Things I don't like about current implementation (but it carries enough value anyway):

1) `cancel` contract is even more strange.
2) Communication via general cancellation is fragile. Probably some of coroutine builders are broken in terms of cancellation and we should test literally each of them
3) Suppression-based exception aggregation doesn't work on JDK 1.6, JS and Native (we can solve it for everything except 1.6).

1-2 can be resolved via more fine-graned cancellation using `attachChild(): ParentHandle`, so it can be postponed (current solution already carries enough value by itself)

3 can be resolved for JS and Native, but not for JDK 1.6
There a not so many use-cases when multiple exceptions are thrown, so it may be not a big deal, but it still may create exception cycles and duplicates.
I personally like Trio [solution](http://trio.readthedocs.io/en/latest/reference-core.html#errors-in-multiple-child-tasks):  multiple exceptions are wrapped into `MultiError` (thus leaking possible parallel decomposition). But it's rare, it's composable, it's immutable (not broken by design :) )

elizarov:
I've refactored some code around `JobSupport.tryFinalizeState`. Now all notifications that happen in final state occur inside `JobSupport.completeStateFinalization` which is clearly called once right after that state update. Please, review.

elizarov:
Oops... just saw a failure of `JobExceptionsStressTest.testMultipleChildrenThrows` on windows -- https://teamcity.jetbrains.com/viewLog.html?buildId=1559284&tab=buildResultsDiv&buildTypeId=KotlinTools_KotlinxCoroutines_BuildWindows

PR #454, opened 2018-07-25
Introduce ThreadContextElement API to integrate with thread-local sensitive code

### LINKED ISSUE #119: Work with ThreadLocal-sensitive Components

elizarov:
Fixes #119

### DIFF
--- a/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
+++ b/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
@@ -136,8 +136,6 @@ public final class kotlinx/coroutines/experimental/CoroutineContextKt {
 	public static final fun newCoroutineContext (Lkotlin/coroutines/experimental/CoroutineContext;)Lkotlin/coroutines/experimental/CoroutineContext;
 	public static final fun newCoroutineContext (Lkotlin/coroutines/experimental/CoroutineContext;Lkotlinx/coroutines/experimental/Job;)Lkotlin/coroutines/experimental/CoroutineContext;
 	public static synthetic fun newCoroutineContext$default (Lkotlin/coroutines/experimental/CoroutineContext;Lkotlinx/coroutines/experimental/Job;ILjava/lang/Object;)Lkotlin/coroutines/experimental/CoroutineContext;
-	public static final fun restoreThreadContext (Ljava/lang/String;)V
-	public static final fun updateThreadContext (Lkotlin/coroutines/experimental/CoroutineContext;)Ljava/lang/String;
 }
 
 public abstract class kotlinx/coroutines/experimental/CoroutineDispatcher : kotlin/coroutines/experimental/AbstractCoroutineContextElement, kotlin/coroutines/experimental/ContinuationInterceptor {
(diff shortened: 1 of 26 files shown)


core/kotlinx-coroutines-core/test/CoroutineContextThreadLocalTest.kt
```diff
@@ -0,0 +1,59 @@
...
+                assertSame(element, myThreadLocal.get())
+            }
+            assertTrue(mainThread != Thread.currentThread())
+            assertSame(element, coroutineContext[MyElement])
+            assertSame(element, myThreadLocal.get())
+        }
+        assertEquals(null, myThreadLocal.get())
+        job.join()
+        assertEquals(null, myThreadLocal.get())
+    }
```
qwwdfsad:
```
 @Test
    fun testUndispatched()= runTest {
        val element = MyElement()
        val job = launch(context = CommonPool + element, start = CoroutineStart.UNDISPATCHED) {
            assertNotNull(myThreadLocal.get())
            yield()
            assertNotNull(myThreadLocal.get())
        }

        assertNull(myThreadLocal.get())
        job.join()
        assertNull(myThreadLocal.get())
    }
```

This test fails

core/kotlinx-coroutines-core/src/CoroutineContextThreadLocal.kt
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Implementations on this interface are looked up via [java.util.ServiceLoader].
+ *
+ * Example usage looks like this:
+ *
+ * ```
+ * // declare custom coroutine context element
+ * class MyElement : AbstractCoroutineContextElement(Key) {
+ *     companion object Key : CoroutineContext.Key<MyElement>
+ *     // some state is kept here
```
qwwdfsad: [nitpick] worth adding some real value here, so it will be perfectly clear for users

core/kotlinx-coroutines-core/test/CoroutineContextThreadLocalTest.kt
elizarov: Fixed it. Added test. Turned out there was a combination of bugs. Not only the test did not work, but actually none of the tests worked because test resources were not included into test builds, but those failures were silently ignored.

core/kotlinx-coroutines-core/src/CoroutineContextThreadLocal.kt
elizarov: Updated examples.

elizarov:
I had to add lot more changes to clean up and fix the way thread context is installed. Test for debug thread names is also introduced.

elizarov:
@qwwdfsad I rewrote `CoroutineContextThreadLocal` API so that it requires `O(N)` overhead, where `N` is the size of the context of the current coroutine. It does not depends anymore on the number of registered thread locals. Each thread local is represented by a key in the coroutine context and only the ones which have their present in the context are being asked to `updateThreadContext`.

elizarov:
I've reworked it once more, getting rid of `ServiceLoader`. Now, all you have to do to use this API is to define a context element that implements `ThreadContextElement`.

PR #472, opened 2018-08-01
Defined specific dependencies for android sample apps

Dmitry-Borodin:
Opening project in android studio to check samples will generate additional files
- local.properties need to be local, so all android projects have it in .gitignore
- gradlew.properties defined gradle version and it have to be in the repository if we want consistent behavior of the samples 

gradlew files are scripts generated by AS, it is a common practice to have them in the repository 
Examples 
https://github.com/JakeWharton/timber/tree/master/gradle/wrapper
https://github.com/JakeWharton/timber
https://github.com/googlesamples/android-JobScheduler
While I don't see any specific settings there, so they can be added to gitignore as well, if there is some reason for them to not be in the repository.


elizarov:
Thanks

PR #477, opened 2018-08-06
corrected return-type of EventLoop pseudo ctor

Groostav:
without exposing `processNextEvent` the returned value from `EventLoop()` is not very useful!


elizarov:
I've rebased your code onto the latest `develop` and added a hidden function to restore binary compatibility with previously compiled code that used `EventLoop()` pseudo-ctor. It is in `event-loop-ctor` branch. Please, squash all three commits into one for better history and update your PR.

elizarov:
I've squashed and merged it into `develop` manually. Thanks.

Issue #169, opened 2017-11-28
Actor need to support termination message, stop accepting messages, discard, terminate


elizarov:
Actor implementation needs ability to stop accepting further messages (on receiving some kind of "final" message), discard all the queued messages (they may need to be returned to some kind of pool or deliver response), and do some normal termination procedure (without being cancelled).

elizarov:
It looks that if we support "closing channel with a value" it would fix both this issue and also #332

PR #485, opened 2018-08-09
Introduce typesafe actors abstraction

### LINKED ISSUE #87: Writing complex actors
I need to create a complex actor using some private methods to update the internal state, but I cannot find any useful base class to implement `ActorJob`.

What is the best practice in such case?

qwwdfsad:
Fixes #87 and #169

### DIFF
--- a/core/kotlinx-coroutines-core/src/actors/TypedActor.kt
+++ b/core/kotlinx-coroutines-core/src/actors/TypedActor.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.experimental.actors
+
+import kotlinx.coroutines.experimental.*
+import kotlinx.coroutines.experimental.channels.*
+import kotlin.coroutines.experimental.*
+
+/**
+ * [TypedActor] is the base for all stateful actors, which can process only one [type][T] of messages.
+ * [TypedActor] has well-defined lifecycle described in [ActorTraits].
+ * [TypedActor.receive] method is used to declare a message handler, which is parametrized by [T]
+ * to provide better compile-type safety.
+ *
+ * Example:
+ * ```
+ * class ExampleActor : TypedActor<String>() {
+ *
+ *   override suspend fun receive(string: String) = act {
+ *     println("Received $string")
+ *   }
+ * }
+ *
+ * // Sender
+ * exampleActor.send("foo")
+ * ```
+ *
+ * @param T type of the message this actor can handle
+ * @param context context in which actor's job will be launched
+ * @param parent optional parent of actor's job
+ * @param start start mode of actor's job
+ * @param channelCapacity capacity of actor's mailbox aka maximum count of pending messages
+ */
+@Suppress("EXPOSED_SUPER_CLASS")
+abstract class TypedActor<T>(
+    context: CoroutineContext = DefaultDispatcher,
+    parent: Job? = null,
+    start: CoroutineStart = CoroutineStart.LAZY,
+    channelCapacity: Int = 16
+) : AbstractActor<T>(context, parent, start, channelCapacity) {
+
+
+    /**
+     * Sends the message to the actor, which later will be sequentially processed by [receive].
+     * Sender is suspended, if actor's channel capacity is reached. This suspension is cancellable
+     * and has semantics similar to [SendChannel.send]
+     *
+     * @throws ClosedSendChannelException if actor is [closed][close]
+     */
+    suspend fun send(message: T) {
+        job.start()
+        mailbox.send(message)
+    }
+
+    /**
+     * Attempts to send message to the actor, which later will be sequentially processed by [receive].
+     * Attempt is successful if actor's channel capacity restriction is not violated.
+     * This method is intended to be used from synchronous callbacks with [Channel.UNLIMITED]
+     *
+     * @throws ClosedSendChannelException if actor is [closed][close]
+     * @return `true` if offer was successful, false otherwise
+     */
+    fun offer(message: T): Boolean {
+        job.start()
+        return mailbox.offer(message)
+    }
+
+    /**
+     * Handler, which handles all received messages.
+     *
+     * @throws ClassCastException if actor was casted to raw type and [send] was invoked with wrong type of the argument
+     */
+    protected abstract suspend fun receive(message: T)
+
+    internal override suspend fun onMessage(message: T) {
+        receive(message)
+    }
+}
--- a/coroutines-guide.md
+++ b/coroutines-guide.md
@@ -85,7 +85,10 @@ You need to add a dependency on `kotlinx-coroutines-core` module as explained
   * [Thread confinement fine-grained](#thread-confinement-fine-grained)
   * [Thread confinement coarse-grained](#thread-confinement-coarse-grained)
   * [Mutual exclusion](#mutual-exclusion)
-  * [Actors](#actors)
+* [Actors](#actors)
+  * [Implementing simple actor](#implementing-simple-actor)
+  * [Typesafe actors](#typesafe-actors)
+  * [Actors API](#actors-api)
 * [Select expression](#select-expression)
   * [Selecting from channels](#selecting-from-channels)
   * [Selecting on close](#selecting-on-close)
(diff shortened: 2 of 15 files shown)


coroutines-guide.md
```diff
@@ -2022,79 +2025,164 @@ The locking in this example is fine-grained, so it pays the price. However, it i
...
+    
+    override suspend fun onStart() {
+      println("CountingActor started")
+    }
+    
+    suspend fun increment() = act { // <- note act {} extension
+      counter++
+    }
+    
+    suspend fun counter(response: CompletableDeferred<Int>) = act {
```
fvasco: Is parameters as an output the final design?

Should the `counter` method return an `Int`?

https://github.com/Kotlin/kotlinx.coroutines/issues/87#issuecomment-397554104

qwwdfsad: Yes.
We discourage "send and wait" pattern as it's not an actor-ish way to do things, if you need it, you're probably using the wrong abstraction.
Actors come in a system, one actor is no actor. In the real world `counter` method would send its result to another actor (which then updates UI, prints it to the console, whatever).

Nevertheless, for some integrations with existing code it's more or less trivial to write `actAndResponse` on top of `act` in 5-10 lines of code. But we don't want to provide it in the library to avoid misusages

coroutines-guide.md
```diff
@@ -2022,79 +2025,164 @@ The locking in this example is fine-grained, so it pays the price. However, it i
...
+fun main(args: Array<String>) = runBlocking<Unit> {
+    val counter = CountingActor() // create the actor
+    println("Preparing to send a lot of inc requests")
+    massiveRun(CommonPool) {
+        counter.increment()
+    }
+    
+    val response = CompletableDeferred<Int>()
+    counter.counter(response)
+    println("Counter = ${response.await()}")
```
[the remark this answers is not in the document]
qwwdfsad: It's a matter of taste.
For me this pattern is obscure

qwwdfsad:
>do you consider a good idea providing an example of bad code?
>Should it be better describe your point of view in documentation and provide a better example?

I think I will add a short section.
The problem with documentation and a lot of examples is that it's already too big to be fully readable, we're thinking how to rewrite it

coroutines-guide.md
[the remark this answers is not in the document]
elizarov: Agree. We should actually rewrite actor example in the guide. Actors should never be written like that. You should never need to wait for a response to a particular actor's action.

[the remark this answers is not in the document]
fvasco: Same example here

https://github.com/Kotlin/kotlinx.coroutines/blob/master/coroutines-guide.md#actors

Following your reasoning the "counter" use case fits really better using a `Mutex`, otherwise you have to consider the feature of replying actor's messages.

dave08:
Is there a way to confine the current Actor implementation to say, Android's Service lifecycle? And also how would notifying progress to the UI be handled? I've tried with the previous implementation, and had code riddled with nested CompletableDeferreds, just to keep the service alive (by blocking its thread) while the actors were finishing their job... I didn't end up using the design because of the many hard to track down bugs that introduced...

qwwdfsad:
@dave08 
Bind android lifecycle to a `Job` (with defining handlers `onStop() = Job.cancel()` etc.) and then use that job as a parent of all actors.
Could you please clarify the rest of your statement? The problem doesn't look clear: you have to use `CompletableDeferred` to "send and wait"? Application was terminated before actors finished their work?

dave08:
@qwwdfsad I want Android to know when the actors are actually finished. If the actors are called on the main Service thread and don't block it, Android will think the Service finished its work. On the other hand, I need to fan out and fan in for different actor tasks. The main overseeing task should be reporting progress in a blocking way, until finished. So I had tried passing CompletableDeferreds to the actor chain, and await for them on the Services main thread, which blocks it and gives me progress... but it was a mess... Your point of using a Job is a different aspect, say if the Service must be interrupted even though it wasn't finished, then resources should be released etc... but I need that a regular run should complete normally.

qwwdfsad:
@dave08
>I want Android to know when the actors are actually finished. If the actors are called on the main Service thread and don't block it, Android will think the Service finished its work

`runBlocking { actor.join() }` from non-daemon thread is not applicable? Your architecture with overseeing task is not clear to me (I'm out of Android/your app context)

Could you extract your problem to a toy android app with a couple of classes? If so, I can rewrite it using actors (and probably will find a couple of bugs or API inconveniences) and add it as a sample project to the guide, some background job + UI progress bar + termination conditions are a perfect fit for actors and it would be great to have such example.

LouisCAD:
@dave08 The `Service` thread in an Android app is the same thread as everywhere else, the main (UI) thread, that you should *never*  block. If you're thinking about `IntentService`s, they have their own thread, but these are kind of deprecated since API 26.
You can't tell Android to wait when it's telling you to stop. You can just cancel work that should be cancelled in a best effort. If Android kills your process, everything is aborted anyway.

The best you can do in a `Service` is asking it to stop with `stopSelf()`, at which point `onDestroy()` will be invoked (but the application process may still stay alive)

kevinherron:
Is this going to make it into the release with 1.3? Even with experimental annotations (are those transitive, it relies on channels...) that would be preferable.

qwwdfsad:
@kevinherron 

We are focused on "for 1.0 release" issues in order to make release 1.0 soon and not to break non-experimental things in the future.
Actors will come in the follow-up release.

jcornaz:
Thank you very much for your work.

In the proposed design, I tend to consider error-prone the need to use `act`, since it can easily be forgotten without any help or warning from the compiler. And Forgetting to use `act` result in data-race, which is never easy to understand/spot.

```kotlin
class MyActor : Actor() {

   // I don't find very clear that doSomething may resume *before* the work being actually done. Since the function suspend, it would be natural (IMHO) to assume it resumes when the job is actually done.
   // It is also very important to not forget to use `act`. But it is easy to forget, compiler won't help and missing usages of `act` may be hard to spot.
   suspend fun doSomething() = act {
     // work...
   }
}
```

Personally I'd prefer something like this:
```kotlin
interface Actor<in T> {

  // Here it is explicit, that the method suspend only until the message has been accepted.
  suspend fun accept(message: T)
}

class MyActor : AbstractActor<MessageType>() {

  // Yes it is more verbose. But there's nothing special we can forget without getting compilation error.
  override suspend fun onMessage(message: MessageType) = when(message) {
	is DoSomething -> // work...
  }
}
```

qwwdfsad:
@jcornaz 
You can use `TypedActor` with method `protected abstract suspend fun receive(message: T)` (I will probably rename it when revisit this branch):

```
class MyActor : TypedActor<MessageType>() {
   override protected suspend fun receive(message: MessageType) = ...
}
```

`Actor` should be used when you need either multiple method parameters (without ugly pairs/sealed classes where they are not necessary) or multiple methods/handlers.

Your concern is understandable, I will update guide and documentation accordingly.
What about missing `act`, we will provide IDEA inspection for public methods on classes which extend `Actor`, but are not using `act`.

jcornaz:
>You can use TypedActor with method protected abstract suspend fun receive(message: T)

This is indeed somehow what I'm looking for. Except I still need to use `act` right?

>Your concern is understandable, I will update guide and documentation accordingly. What about missing act, we will provide IDEA inspection for public methods on classes which extend Actor, but are not using act.

That sounds great, thanks.

qwwdfsad:
>Except I still need to use act right?

No, `TypedActor` doesn't have that method

core/kotlinx-coroutines-core/src/actors/TypedActor.kt
```diff
@@ -0,0 +1,80 @@
...
+ * [TypedActor] is the base for all stateful actors, which can process only one [type][T] of messages.
+ * [TypedActor] has well-defined lifecycle described in [ActorTraits].
+ * [TypedActor.receive] method is used to declare a message handler, which is parametrized by [T]
+ * to provide better compile-type safety.
+ *
+ * Example:
+ * ```
+ * class ExampleActor : TypedActor<String>() {
+ *
+ *   override suspend fun receive(string: String) = act {
```
TimothyEarley: >> Except I still need to use act right?

> No, TypedActor doesn't have that method

Then I guess this example is wrong and `act` is not needed here.

qwwdfsad: yup, documentation leftover

qwwdfsad:
Leaving this one for a reference, I will return to the actors as soon as `Flow` stabilize a bit.
But this API needs careful rethinking in the terms of structured concurrency (and missing extension constructors :))

qwwdfsad:
https://www.reddit.com/r/Kotlin/comments/nm2eaf/kotlin_team_ama_3_ask_us_anything/gznacew?utm_source=share&utm_medium=web2x&context=3

PR #497, opened 2018-08-15
Class for exposing dispatcher's executor

qwwdfsad:
Introducing `ExecutorCoroutineDispatche`r instead of `CloseableCoroutineDispatcher` (ABI breaking change)
Remove deprecated for long enough methods

Also, I like the idea to make `ThreadPoolDispatcher` internal and return `ExecutorCoroutineDispatcher` everywhere instead, but it will break a lot of existing ABI (though, maybe it's not that relevant)

### DIFF
--- a/core/kotlinx-coroutines-core/src/Executors.kt
+++ b/core/kotlinx-coroutines-core/src/Executors.kt
@@ -10,16 +10,49 @@ import java.util.concurrent.*
 import kotlin.coroutines.experimental.*
 
 /**
- * [CoroutineDispatcher] that implements [Closeable]
+ * [CoroutineDispatcher] that has underlying [Executor] for dispatching tasks.
+ * Instances of [ExecutorCoroutineDispatcher] should be closed by the owner of the dispatcher.
+ *
+ * This class is generally used as a bridge between coroutine-based API and
+ * asynchronous API which requires instance of the [Executor].
  */
-abstract class CloseableCoroutineDispatcher: CoroutineDispatcher(), Closeable
+public abstract class ExecutorCoroutineDispatcher: CloseableCoroutineDispatcher(), Closeable {
+
+    /**
+     * Underlying executor of current [CoroutineDispatcher].
+     */
+    public abstract val executor: Executor
+}
+
+/**
+ * [CoroutineDispatcher] that implements [Closeable].
+ *
+ * @suppress **Deprecated**: Use [ExecutorCoroutineDispatcher].
+ */
+@Deprecated("Use ExecutorCoroutineDispatcher instead", replaceWith = ReplaceWith("ExecutorCoroutineDispatcher"))
+public abstract class CloseableCoroutineDispatcher: CoroutineDispatcher(), Closeable
+
+/**
+ * Converts an instance of [ExecutorService] to an implementation of [ExecutorCoroutineDispatcher].
+ */
+public fun ExecutorService.asCoroutineDispatcher(): ExecutorCoroutineDispatcher =
+    // we know that an implementation of Executor.asCoroutineDispatcher actually returns a closeable one
+    (this as Executor).asCoroutineDispatcher() as ExecutorCoroutineDispatcher
+
+/**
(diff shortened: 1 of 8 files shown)


core/kotlinx-coroutines-core/src/Executors.kt
```diff
@@ -98,7 +92,7 @@ private class ResumeUndispatchedRunnable(
  * An implementation of [DisposableHandle] that cancels the specified future on dispose.
  * @suppress **This is unstable API and it is subject to change.**
  */
-public class DisposableFutureHandle(private val future: Future<*>) : DisposableHandle {
+class DisposableFutureHandle(private val future: Future<*>) : DisposableHandle {
```
qwwdfsad: @elizarov any reason it's not internal?

core/kotlinx-coroutines-core/src/Executors.kt
```diff
@@ -10,43 +10,37 @@ import java.util.concurrent.*
...
 /**
- * [CoroutineDispatcher] that implements [Closeable]
+ * [CoroutineDispatcher] that has underlying [Executor] for dispatching tasks.
+ * Instances of [ExecutorCoroutineDispatcher] should be closed by the owner of the dispatcher.
+ *
+ * This class is generally used as a bridge between coroutine-based API and
+ * asynchronous API which requires instance of the [Executor]
  */
-abstract class CloseableCoroutineDispatcher: CoroutineDispatcher(), Closeable
+public abstract class ExecutorCoroutineDispatcher: CoroutineDispatcher(), Closeable {
```
elizarov: I'd suggest to keep maintaining binary compatibility with previously compiled code by using `@JvmName("CloseableCoroutineDispatcher")` to this declaration and adding a deprecated `typealias CloseableCoroutineDispatcher = ExecutorCoroutineDispatcher` with an appropriate `replaceWith`. We will clean it up in 1.0 release.

core/kotlinx-coroutines-core/src/Executors.kt
```diff
@@ -10,43 +10,37 @@ import java.util.concurrent.*
...
-/**
- * Converts an instance of [ExecutorService] to an implementation of [CloseableCoroutineDispatcher].
- */
-public fun ExecutorService.asCoroutineDispatcher(): CloseableCoroutineDispatcher =
-    // we know that an implementation of Executor.asCoroutineDispatcher actually returns a closeable one
-    (this as Executor).asCoroutineDispatcher() as CloseableCoroutineDispatcher
+    /**
+     * Underlying executor of current [CoroutineDispatcher].
+     */
+    abstract val executor: Executor
```
elizarov: Please, use explicit visibility (`public` here) in libraries code.

core/kotlinx-coroutines-core/src/Executors.kt
elizarov: Frankly I don't remember. If it compiles with `internal`, then it should be `internal`.

core/kotlinx-coroutines-core/src/Executors.kt
```diff
@@ -10,43 +10,37 @@ import java.util.concurrent.*
...
  * Converts an instance of [Executor] to an implementation of [CoroutineDispatcher].
  */
 public fun Executor.asCoroutineDispatcher(): CoroutineDispatcher =
-    ExecutorCoroutineDispatcher(this)
+    ExecutorCoroutineDispatcherImpl(this)
 
-private class ExecutorCoroutineDispatcher(override val executor: Executor) : ExecutorCoroutineDispatcherBase()
+private class ExecutorCoroutineDispatcherImpl(override val executor: Executor) : ExecutorCoroutineDispatcherBase()
+
+abstract class ExecutorCoroutineDispatcherBase : ExecutorCoroutineDispatcher(), Delay {
```
elizarov: This abstract class shall is public will be visible in generated documentation. It shall be either fully documented with KDoc or suppressed from documentation (as it was before).

elizarov:
I've restored binary compatibility. Waiting for tests on teamcity.

PR #499, opened 2018-08-16
Implement withTimeoutOrNull via withTimeout to avoid timing bugs and …

### LINKED ISSUE #498: withTimeoutOrNull throw TimeoutCancellationException
Hi,
I understand that ```withTimeoutOrNull``` never throw timeout exception, does it?

I will try to reproduce it.

The code sample is
```
val ch = cudChannel.openSubscription()
try {
    var lastPingTime = System.currentTimeMillis()
    while (true) {
        val cud = withTimeoutOrNull(1000) {
            ch.receive()
        }
        if (cud != null) {
            wc.writeString(cud.first)
            wc.writeBoolean(cud.second)
        } else if (System.currentTimeMillis() - lastPingTime > 6000) {
            wc.writeString("")
            wc.writeBoolean(false)
        } else {
            continue
        }
        lastPingTime = System.currentTimeMillis()
        withTimeout(30000) {
            rc.readInt()
        }
    }
} finally {
    ch.cancel()
}
```

The stacktrace is
```
kotlinx.coroutines.experimental.TimeoutCancellationException: Timed out waiting for 1000 MILLISECONDS
        at kotlinx.coroutines.experimental.ScheduledKt.TimeoutCancellationException(Scheduled.kt:202)
        at kotlinx.coroutines.experimental.TimeoutCoroutine.run(Scheduled.kt:100)
        at kotlinx.coroutines.experimental.EventLoopBase$DelayedRunnableTask.run(EventLoop.kt:322)
        at kotlinx.coroutines.experimental.EventLoopBase.processNextEvent(EventLoop.kt:148)
        at kotlinx.coroutines.experimental.DefaultExecutor.run(DefaultExecutor.kt:60)
        at java.lang.Thread.run(Thread.java:745)
```

qwwdfsad:
…races.

Remove deprecated API

Fixes #498

### DIFF
--- a/common/kotlinx-coroutines-core-common/test/WithTimeoutOrNullTest.kt
+++ b/common/kotlinx-coroutines-core-common/test/WithTimeoutOrNullTest.kt
@@ -12,7 +12,6 @@ import kotlin.coroutines.experimental.*
 import kotlin.test.*
 
 class WithTimeoutOrNullTest : TestBase() {
-
     /**
      * Tests a case of no timeout and no suspension inside.
      */
--- a/common/kotlinx-coroutines-core-common/test/WithTimeoutOrNullTest.kt
+++ b/common/kotlinx-coroutines-core-common/test/WithTimeoutOrNullTest.kt
@@ -7,10 +7,12 @@
 
 package kotlinx.coroutines.experimental
 
+import kotlinx.coroutines.experimental.channels.*
 import kotlin.coroutines.experimental.*
 import kotlin.test.*
 
 class WithTimeoutOrNullTest : TestBase() {
+
     /**
      * Tests a case of no timeout and no suspension inside.
      */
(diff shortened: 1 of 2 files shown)


common/kotlinx-coroutines-core-common/test/WithTimeoutOrNullTest.kt
```diff
@@ -7,10 +7,12 @@
 
 package kotlinx.coroutines.experimental
 
+import kotlinx.coroutines.experimental.channels.*
 import kotlin.coroutines.experimental.*
 import kotlin.test.*
 
 class WithTimeoutOrNullTest : TestBase() {
+
```
elizarov: Let's follow Kotlin code style with respect to empty lines (they should used to separate declarations in class, but not between class declaration line and first member). See https://kotlinlang.org/docs/reference/coding-conventions.html The only possible exception is when multi-line class declaration is formatted with a trailing `... {`, then an empty line should be used as a separator. However, I suggest we consistently prefer the second option from the conventions document and always put `{` on a separate line as a separator in such cases.

qwwdfsad:
Squashed and merged

Issue #74, opened 2017-06-19
Stack trace recovery


wetnose:
Look at the example below

```kotlin
fun main(args: Array<String>) = runBlocking {
    xxx()
}


suspend fun xxx() {
    delay(100)
    throw Exception()
}
```

The execution of the code results to an exception with the incomplete stack trace.

```
Exception in thread "main" java.lang.Exception
	at TestKt$xxx$1.doResume(Test.kt:20) // throw Exception()
	at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:54)
        ....
	at TestKt.main(Test.kt:13) //  runBlocking
```
The stack is starting from the bridge function and completes with the last continuation. All the intermediate calls are missing.

The problem makes it difficult to debug the app because it is not possible to determine the true continuation caller. The stack trace of an exception, thrown from a coroutine, becomes useless.

elizarov:
This is indeed a known issue on our plate. Thank you for starting the thread, though. I'm curious what kind of stack-track you would personally prefer to see in this case?

wetnose:
I want to see pointers to actual code lines with function calls. In my example above, the pointer to the line with call of `xxx` function is missing. If expand this example with a chain of calls (xxx -> yyy -> zzz -> suspend/resume -> throw Exception) then all the pointers will be missing in the stack trace excepting the last one witch throws the exception.

wetnose:
One more thing. I think the problem is critical for Kotlin coroutines project: non-informative stack traces makes debugging of coroutines too expensive as well as supporting of a program that uses coroutines. From my point of view the main goal of coroutines is to make asynchronous programming easier and cheaper. But with expensive debugging this goal cannot be achieved.

elizarov:
It is indeed important. Kotlin coroutines are _already_ easier than alternative async solutions which are all notoriously known for their hard-to-decipher stack traces, but we can also make Kotlin coroutines stack traces better. It is not a straightforward thing to do on JVM and will likely come at a considerable performance penalty (to be paid when exception is being propagated though coroutine), so the current idea is to make it available under a debug system property.

wetnose:
I think in the beginning the cost does not matter, especially that it will be optional.

molixiaoge:
The serious thing is the main func immediately exit, when some func throw exception!

Example Codes:

```
fun main(args: Array<String>) = runBlocking {
    println("start")
    xxx()
    println("end")
}


[...]

PR #502, opened 2018-08-20
[Prototype] exceptions stracktrace recovery

### LINKED ISSUE #493: Basic stacktrace augmentation
After #333 we can do a basic [pluggable? opt-in or opt-out?] stacktrace augmentation on `throw` call-site.

The basic idea is to attempt to reflectively instantiate exception of the same type at `throw` call-site with original exception as a cause.

For example currently this code
```
val deferred = async {
  throw IllegalStateException("Some invariant failed")
}

try {
  deferred.await()
} catch (t: Throwable) {
  t.printStackTrace()
}
```

will print
```
java.lang.IllegalStateException: Some invariant failed
  at kotlinx.coroutines.experimental.AsyncJvmTest$simpleTest$1$deferred$1.doResume(AsyncJvmTest. 
 kt:15)
  at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:42)
  at kotlinx.coroutines.experimental.DispatchedTask$DefaultImpls.run(Dispatched.kt:149)
```
after stacktrace augmentation it will be

```
java.lang.IllegalStateException: java.lang.IllegalStateException: Some invariant failed

  at kotlinx.coroutines.experimental.AsyncJvmTest$simpleTest$1.doResume(AsyncJvmTest.kt:21)
  at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:42)
  at kotlinx.coroutines.experimental.DispatchedTask$DefaultImpls.run(Dispatched.kt:149)
  at kotlinx.coroutines.experimental.DispatchedContinuation.run(Dispatched.kt:13)
  at kotlinx.coroutines.experimental.EventLoopBase.processNextEvent(EventLoop.kt:132)
  ....
  at kotlinx.coroutines.experimental.AsyncJvmTest.simpleTest(AsyncJvmTest.kt:13) // <- WHERE EXCEPTION IS ACTUALLY THROWN
Caused by: java.lang.IllegalStateException: Some invariant failed
  at kotlinx.coroutines.experimental.AsyncJvmTest$simpleTest$1$deferred$1.doResume(AsyncJvmTest.kt:15)
  at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:42)
  at kotlinx.coroutines.experimental.DispatchedTask$DefaultImpls.run(Dispatched.kt:149)

```

FJP for FJ-tasks does exactly the same: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/ForkJoinTask.java#l582

qwwdfsad:
1) Provide entry-points for stack-trace recovery at the call-site (not at suspension point), which can be improved in 1.3
2) First-class support in `await` and channel
3) Enable only in debug mode

Things to do
1) Discuss whether we want to hide this behavior under another system property
2) Implement caching on JDK 1.7 with ClassValueMap (we can't use CHM because it will be a memory leak)
3) Implement `withContext` support

Fixes #493 and (partially) #74


SUPERCILEX:
This would be amazing! I currently do something very similar by injecting the callsite stacktrace as a fake cause of the suspension point exception.

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -0,0 +1,88 @@
...
+                break
+            }
+        }
+    } catch (e: Exception) {
+        // Do nothing
+    }
+    return newException
+}
+
+private fun fillInStackTrace(continuation: Continuation<*>): ArrayList<StackTraceElement> {
```
elizarov: It does not "fill" anything, so it should be called `getStackTrace`.

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -0,0 +1,88 @@
...
+    var last = continuation
+    while (true) {
+        last = getCompletion(last) ?: break
+        stack.add(stackTraceElement(last))
+    }
+    return stack
+}
+
+// TODO basic stub before 1.3
+private fun stackTraceElement(continuation: Continuation<*>): StackTraceElement {
```
elizarov: Should return a nullable result. An object may not have a stack trace itself, but may be a completion. An good example here is `CancellableContinuationImpl` that should be supported.

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -0,0 +1,88 @@
...
+internal actual fun <E : Throwable> recoverStackTrace(exception: E): E {
+    if (!DEBUG || exception is CancellationException) {
+        return exception
+    }
+
+    return tryWrapException(exception) ?: exception
+}
+
+internal actual fun <E : Throwable> recoverStackTrace(exception: E, continuation: Continuation<*>): E {
+    if (!DEBUG || exception is CancellationException || continuation !is CoroutineImpl) {
```
elizarov: `continuation !is CoroutineImpl` is too restrictive for experimental coroutines. `CancellableContinuationImpl` shall be supported, too. It is needed for basic example that use `Deferred.await()`, for example.

PR #516, opened 2018-08-25
Add OSGI support

GisoBartels:
Add OSGI meta data to coroutines-core, so the jar can be used as an OSGI bundle.

Please let me know, if there is more meta data desired, e.g. bundle-description.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -152,6 +152,11 @@ configure(subprojects.findAll { !internal.contains(it.name) && it.name != 'kotli
     }
 }
 
+// configure OSGI manifest
+configure(subprojects.findAll { !internal.contains(it.name) && platformOf(it) == 'jvm' }) {
+    apply from: rootProject.file("gradle/osgi.gradle")
+}
+
 // --------------- Configure sub-projects that are published ---------------
 
 // todo: native is not published yet
--- a/core/kotlinx-coroutines-core/build.gradle
+++ b/core/kotlinx-coroutines-core/build.gradle
@@ -7,6 +7,13 @@ dependencies {
     testCompile "com.esotericsoftware:kryo:4.0.0"
 }
 
+jar {
+    manifest {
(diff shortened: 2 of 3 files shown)


elizarov:
It seems worthwhile to have to have OSGi bundles for all JVM modules (not just `kotlinx-coroutines-core`), that is add the corresponding plugin to `gradle/compile-jvm.gradle` file. It also makes sense to properly declare internal packages (so they are not exported) and to set `Bundle-Vendor` to `JetBrains`.

GisoBartels:
@elizarov: done.
I only found one internal package, so please let me know, if I missed one.

elizarov:
I've also added `scheduling` to internal packages, rebased and pushed to `osgi` branch for testing. Now, the question that I have is whether it all makes sense for `kotlinx.coroutines`. I've looked at the resulting OSGi bundle and I see all the imports for `kotlin.*` (kotlin runtime) packages, but kotlin runtime does not have the OSGi bundle annotations in its jar file, so there is no bundle to import those packages from. What is the the use of adding it to `kotlinx.coroutines`?

qwwdfsad:
@GisoBartels do you still need OSGi meta-data? Could you please answer @elizarov question?

GisoBartels:
rebased PR and your changes to v1.0.1

notiocide:
Has there been any progress on this issue?

I'm currently embedding the library inside each of my bundles that require it, but I worry that each of the bundles might end up using a different instance of the built-in dispatchers. I'm not quite sure if this is too much of a problem, since it seems to me that the `Default` dispatcher utilizes the common `ForkJoinPool` (which, as it is part of the standard library, should be the same across bundles), but I wonder if this behavior may become an issue in other places.

elizarov:
Supporting OSGI is more hassle than its worth for us in its current shape. It is not common practice to include OSGI metadata in popular JVM ecosystem libraries and we don't have any means to test it, so we will not even know if it works or not. Users can add they own metadata as needed. I'm closing this PR. We can reconsider if (at least) the corresponding PR would add some kind of tests.

Issue #360, opened 2018-05-15
Proposal: work with grpc-java team to provide binding layer and code generator for grpc-java


hsyed:
Grpc-java is a low-level api that is quite error prone to use without significant investment in code-generation and libraries encoding relevant concepts --e.g., [(pagination, batching, etc](`https://github.com/googleapis/gax-java`). 

The go use-case for grpc  is much easier to get started with for applications largely due to channels and context propagation being part of the stub generation. Kotlin Coroutines can provide an even better experience out of the box, completely eliminating the need for Streaming libraries or a lot of the utilities provided by a library like gax --e.g., Batching and Paging would be much easier to codify on a per application bases.  

At the moment we are prototyping with a glue layer that connects grpc-java concepts per request. The bidi stream use case in this context would code generate a method that creates a job that does: 

1. Glue a `Channel` to the input `StreamObserver` 
2. Glue `LinkedListChannel` to the out `StreamObserver`. 
3. Bind Grpc context to Coroutine Context or create a dispatcher using `Context::fixedContextExecutor`.
4. Propogate Context cancelation to the Job.

This is quite heavy handed. Ideally we should have a fully kotlinx.coroutine based layer perhaps directly above Netty that sidesteps the need to use the grpc-java layer and it would be fully coroutine based from the ground up.

elizarov:
@hsyed Seem like a great idea. Do you have any prototype to look at?

hsyed:
@elizarov our mono repo is built using Bazel I will try and share the code if I can.

I based the code generator in our example from[here](https://github.com/rouzwawi/grpc-kotlin). I've made quite some changes to the templates and what they generate but the principal hasn't diverged too far from it. What is good about this generator is that it the generated stubs are compatible with grpc-java. This code generator uses mustache templates and an additional external dep from salesforce to generate both of these deps aren't strictly needed. 

There is another code generator [here](https://github.com/marcoferrer/kroto-plus) -- this generates various utility code. It uses "kotlin poet" from square and uses the a proto file parser from square/wire to parse proto files -- kotlin poet is very useful but the proto file parser isn't strictly needed.


--- 

Here is some example code to see how our generators glue with grpc-java. The [binding classes](https://gist.github.com/hsyed/c27fc7089a1f84eed77c844100c26d8a) our generators use.  Note the generator is different to the open source generator I linked above.  

Server Stub: Server Streaming
```
[...]

PR #518, opened 2018-08-27
Initial prototype for grpc integration

marcoferrer:
This pr is in reference to issue #360 

While its still far from complete, I did want to open the pr early so that the proposed implementation can be discussed. 

With the code generation aspect of this being trivial, I wanted to focus more on the design of the integration before building the protoc plugin. 

This initial commit adds support for client stubs and bridging coroutines with the grpc ```StreamObserver``` interface. Code needed for supporting a ```BindableService``` impl base is not far behind.

Ive based this loosely off of the code from an existing [code generator](https://github.com/marcoferrer/kroto-plus) I authored, but I do feel there is room for improvement or maybe even better approaches.

The [example stub](https://github.com/marcoferrer/kotlinx.coroutines/blob/df783d01568cf03eaf8c29b44a0f63019df66e64/integration/kotlinx-coroutines-grpc/test/example/example-stub.kt) shows what the generated code for a client stub would look like.

---
This change is [Reviewable](https://reviewable.io/reviews/kotlin/kotlinx.coroutines/518)


qwwdfsad:
Hi @marcoferrer

Any update on the work of gRPC? No hurry, just to know current status.

marcoferrer:
I pushed back some of the work because I wanted to tackle it after I attending the coroutines workshop at kotlin conf. I got a lot of good insight there from Roman. Updating kroto-plus to stable coroutines showed me somethings that can be handled better. So Im currently rethinking certain parts so that they better align with the structured concurrency model, in addition to exploring lower level grpc apis for supporting back pressure correctly.  

Moving forward, I think you were right about getting a fully working client-server unary example before trying to reason about the quirks introduced with streaming rpcs. I'm planning on dialing back on the scope of this pr so that progress can be made on the parts that I'm comfortable with design wise.

I do feel like progress is picking up and Ill be pushing what I have soon. Thanks for checking in @qwwdfsad !

elizarov:
Thanks for your initial prototype, but we won't be maintaing GRPC integration in this repository.

PR #538, opened 2018-09-07
Add aWithLock, aReadBytes, aReadText extensions to AsynchronousFileChannel

clarcharr:
These extension functions match the existing ones on `File` (minus `aWithLock`, as there currently isn't a `File.withLock` method for a regular `File`), and should work as I've been using them already in my own code. Figured that these would be useful to have upstream.

### DIFF
--- a/integration/kotlinx-coroutines-nio/src/Nio.kt
+++ b/integration/kotlinx-coroutines-nio/src/Nio.kt
@@ -25,6 +25,11 @@ suspend fun AsynchronousFileChannel.aLock() = suspendCancellableCoroutine<FileLo
 }
 
 /**
+ * Wrapper around [aLock] that unlocks after a block is executed.
+ */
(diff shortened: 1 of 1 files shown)


qwwdfsad:
Sadly, we've decided to no longer support NIO integration as it carries no value at all.
This module will be removed in the future release

clarcharr:
Not exactly sure what you mean by "carries no value at all." Are you saying that there won't be asynchronous I/O support using Kotlin coroutines, or that this module will be moved to the kotlinx.io repo?

elizarov:
If you are using NIO `AsynchronousFileChannel` classes, then you should copy the corresponding supporting functions to your own code and/or maintain your own integration library. It requires only a thin wrapper on top of Java APIs. We don't plan to provide support for asynchronous file access in our own libraries in the near future. Sorry.

PR #557, opened 2018-09-13
Minor link fix for structured concurrency section

rocketraman:


qwwdfsad:
It's already fixed in develop, thanks

PR #558, opened 2018-09-13
Fix CoroutineScope.plus operator

r4zzz4k:


elizarov:
Thanks. That is embarrassing. The fix in not correct either (does not compile). I've fixed it correctly and added a test. See #560

r4zzz4k:
It happens. My screw for not checking PR is much woree than that, sorry.

PR #569, opened 2018-09-16
Fixed CoroutineScope.plus operator

zach-klippenstein:
The operator was just combining the passed-in `context` with itself, which blew away the current context from the receiver `CoroutineScope`.


JakeWharton:
Dupe of #560

PR #573, opened 2018-09-17
Fix code sample code for log() in first-read document

XinyueZ:


XinyueZ:
My mis, I should run with option -Dkotlinx.coroutines.debug

archerLj:
@XinyueZ hi, can you tell me where the option -Dkotlinx.coroutines.debug should be added? I put it in `gradle.properties` and also put it in `studio.vmoptions`, but none of them works.

PR #582, opened 2018-09-18
Aligned to RxJava2 contract - Observable and Single returns non nulla…

### LINKED ISSUE #347: rxSingle, rxObservable nullity
Currently the rxSingle and rxObservable methods return of type `T`. So code like this is valid:

```kotlin
val getString: () -> String? = { null }
rxSingle { getString() }.subscribe()
```

However in RxJava2 `null` is not allowed at all and it immediately error. I suggest to set the  bounds to `T: Any` to provide some compile time safety.

Dmitry-Borodin:
…ble type <T>.

Fixes #347 ( https://github.com/Kotlin/kotlinx.coroutines/issues/347 )

This is a breaking change in case if somebody used it for nullable results expecting onError() callback from Observable with NullPointerException

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
@@ -5,9 +5,11 @@
 package kotlinx.coroutines.experimental.rx2
 
 import io.reactivex.*
-import kotlinx.coroutines.experimental.*
-import kotlinx.coroutines.experimental.channels.*
-import kotlin.coroutines.experimental.*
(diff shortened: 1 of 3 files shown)


qwwdfsad:
The code doesn't compile.

Please respect our time and check your changes before doing a pull request.

Dmitry-Borodin:
My bad. Fixed compilation issues, 
While many tests still failing, they are failing on dev and up to date master as well (on old master worked).  Some multiplatform issues with my environment. Working on it.

qwwdfsad:
Rebased onto develop, thanks

PR #592, opened 2018-09-20
Fix typo in deprecation message of JavaFx dispatcher

jcornaz:
Deprecation message of `JavaFx` says: "Use Dispatchers.Main". This is a bit confusing, since there is no `Dispatchers.Main` available in `kotlinx-coroutines-javafx`.

Deprecation message should be "Use Dispatchers.JavaFx".


qwwdfsad:
Thank you!

PR #596, opened 2018-09-21
Provide suspendAtomicCancellableCoroutineReusable primitive, high-per…

### LINKED ISSUE #534: Provide high-performance and reusable CancellableContinuation implementation
Consider the following code that reads some bytes in a loop:

```
suspend fun foo() {
   while (true) {
       val bytes = readBytes(buf) // suspending
       if (bytes < 0) break
   }
}
```

where `readBytes` is written via `suspendCancellableCoroutine`. The problem is that each `readBytes` invocation allocates a new `CancellableContinuationImpl` object and some auxiliary ones to track parent's cancellation. This is unacceptable for high-performance IO code and a standard solution need to be provided for that. For 1.0 release we are looking for an experimental implementation of such an API that relies on new feature in 1.3 coroutines: `ContinuationInterceptor.releaseInterceptedContinuation`.

qwwdfsad:
…formance intrinsic which allows to reuse cancellation machinery and opens the way to garbage-free channels:

  * Make AbstractContinuation reusable
  * Add additional field to DispatchedContinuation to cache cancellable continuation
  * Use suspendAtomicCancellableCoroutineReusable in channels (+15% of performance on ping-pong)
  * Detect calls to tryResume and forbid cancellation reusing in order to make channels work with selects
  * Fix benchmarks on 1.3

Fixes #534


qwwdfsad:
Crazy idea: get rid of `_reusableCancellableContinuation` and turn `resumeMode` to `Any` with following states:
1) `resumeMode is Int && resumeMode >= 0`. It's just a mode.
2) `resumeMode is Int && resumeMode < 0`. Actual mode is `-resumeMode` and `-` indicates sealed reusability. (one may notice that `MODE_ATOMIC_DEFAULT` is zero, but it can easily be mitigated with shifting all modes).
3) `resumeMode` is `CancellableContinuation`, then actual mode is `(resumeMode as CancellableContinuation).resumeMode`. If cancellation is currently used, we can safely mutate its `resumeMode` if we will restore it later in `suspendAtomicCancellableCoroutineReusable`

PR #600, opened 2018-09-22
Added AsynchronousByteChannel.aRead and aWrite.

programingjd:
Useful for using [this](https://github.com/marianobarrios/tls-channel) for instance.

I removed the default timeout of `0L` for `AsynchronousSocketChannel` to make it obvious that if
no timeout is given, the `AsynchronousByteChannel` extension is used.


elizarov:
Sorry, but we've decided not to maintain integration for `AsynchronousXXXChannel` classes in 1.0 release. If you are using those, I'd suggest to copy-and-paste the corresponding integration source file into your own project. See #601 for details

PR #605, opened 2018-09-24
Upgrade Android sample to use structured concurrency

kingsleyadio:
- Updates Android sample/AnimationModel to use structured concurrency


elizarov:
Thanks. Looks good to me.

PR #619, opened 2018-09-26
Runnable docs

hhariri:


elizarov:
I'm closing this one. We already have the corresponding `<div ... >` around code block with the correct style, but making them actually runnable is more work than that (need to pull all the imports in the code).

Issue #584, opened 2018-09-19
Scheduled executor service metrics


jordillonch:
It would be nice if we would monitorize the `ScheduledExecutorService` found [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/core/kotlinx-coroutines-core/src/ThreadPoolDispatcher.kt#L67) using for example: [InstrumentedScheduledExecutorService (Metrics library)](https://github.com/dropwizard/metrics/blob/4.1-development/metrics-core/src/main/java/com/codahale/metrics/InstrumentedScheduledExecutorService.java).
Do you know if there is any way to add these metrics? Thanks! 😄

elizarov:
PR is welcome!

jordillonch:
I found a way to have `ScheduledExecutorService` instrumented:

```kotlin
fun newInstrumentedFixedThreadPoolContext(corePoolSize: Int,
                                          name: String,
                                          metricRegistry: MetricRegistry): CoroutineDispatcher =
        InstrumentedExecutorService(
                Executors.newScheduledThreadPool(corePoolSize)
                { runnable -> Thread(runnable, name) }, metricRegistry, name)
                .asCoroutineDispatcher()
```
What do you think if a prepare a PR with a little module in the `integration` folder called `kotlinx-coroutines-metrics`?

qwwdfsad:
I am closing this issue, see rationale in [PR](https://github.com/Kotlin/kotlinx.coroutines/pull/623#issuecomment-437874467)

PR #622, opened 2018-09-26
integration with metrics library

jordillonch:
Add integration with [Metrics](https://metrics.dropwizard.io/4.0.0/).

Issue: https://github.com/Kotlin/kotlinx.coroutines/issues/584


jordillonch:
it should merge against develop instead master

PR #623, opened 2018-09-26
integration with metrics library

jordillonch:
Add integration with [Metrics](https://metrics.dropwizard.io/4.0.0/).

Issue: https://github.com/Kotlin/kotlinx.coroutines/issues/584


IRus:
I think `kotlinx-coroutines-metrics` shouldn't depend on *Dropwizard Metrics*, since there are plenty of metrics libraries, for example http://micrometer.io/, so we ended up with integrating with all of these libraries? Instead we should define some interface and maybe some implementations for jvm: DropwizardCoroutinesMetricsAdapter, MicrometrCoroutinesMetricsAdapter, etc. Dependencies can be compileOnly, so single module can contains adapters for different metrics libraries.

dave08:
I agree, I think it's important to have Prometheus support, since Kubernetes and Docker work with it. Micrometer provides that and also provides Dropwizard.

dave08:
Also, the Context creators are now obsolete in 0.30.2, so maybe it could be implemented in `asDispatcher`?

jordillonch:
Thanks for your comments. I agree with all. I will try to implement all your advices when I have time 😄

qwwdfsad:
Sorry, but I cannot accept this PR. It creates a less value than it is required to maintain/release/document this module and provides a single method, which can be implemented manually without a lot of effort when it is required.

Any kind of monitoring/instrumenting around `Executor` service should use ` .asCoroutineDispatcher()` in a project where you are using it instead of providing a separate integration module. = 

Sorry for the inconvenience and your time, we should properly review feature and API surface before making it "up for grabs".

PR #634, opened 2018-09-28
Check cancellation when starting unconfined coroutine

### LINKED ISSUE #621: Cancelled scopes can be reused
It's a bit odd that cancelled scopes can be reused.

### Example
```
val scope = CoroutineScope(Dispatchers.Unconfined + Job())

scope.launch { println("hello!") }

scope.coroutineContext[Job]!!.cancel()

// I expect this not to print, because the scope has been cancelled
scope.launch { println("goodbye!") }
```

### Expected

```
hello!
```

### Actual

```
hello!
goodbye!
```

EDIT: It seems the issue is that before launching a coroutine, the launch does not check if the scope has been cancelled or not:

### yield example
```
val scope = CoroutineScope(Dispatchers.Unconfined + Job())

scope.launch { println("hello!") }

scope.coroutineContext[Job]!!.cancel()

scope.launch {
    yield()
    println("goodbye!")
}
```
### result
```
hello!
```

Is this intentional?

qwwdfsad:
Fixes #621

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/Dispatched.kt
+++ b/common/kotlinx-coroutines-core-common/src/Dispatched.kt
@@ -43,8 +43,9 @@ internal class DispatchedContinuation<in T>(
             _state = CompletedExceptionally(exception)
             resumeMode = MODE_ATOMIC_DEFAULT
             dispatcher.dispatch(context, this)
-        } else
+        } else {
             resumeUndispatchedWithException(exception)
+        }
     }
 
     @Suppress("NOTHING_TO_INLINE") // we need it inline to save us an entry on the stack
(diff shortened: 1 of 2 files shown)


common/kotlinx-coroutines-core-common/src/Dispatched.kt
```diff
@@ -65,8 +69,22 @@ internal class DispatchedContinuation<in T>(
             _state = CompletedExceptionally(exception)
             resumeMode = MODE_CANCELLABLE
             dispatcher.dispatch(context, this)
-        } else
-            resumeUndispatchedWithException(exception)
+        } else {
```
qwwdfsad: I'll merge it as `else if()` and rebase after review

PR #655, opened 2018-10-02
[Obsolete] Add experimental AndroidX Lifecycle integration

LouisCAD:
This new integration provides a few extensions to make `Lifecycle` play well with `Job` and `CoroutineScope`.

~It has no documentation and tests at the moment, so don't merge it yet. Any contribution on this PR is welcome, especially regarding tests.~ **EDIT: Ready for review**

Thanks to @SUPERCILEX for the `aar` configuration code for build.gradle.


LouisCAD:
Adding a test dependency for `androidx.lifecycle:lifecycle-runtime:2.0.0` doesn't work because it is an aar. Any help for this welcome.

integration/kotlinx-coroutines-androidx-lifecycle/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+
+ext.androidx_lifecycle_version = '2.0.0'
+
+repositories {
+    google()
+}
+
+def attr = Attribute.of("artifactType", String.class)
+configurations {
+    aar {
```
SUPERCILEX: Add another configuration called `aarTest` that only modifies the test classpath.

LouisCAD: Thanks for the hint! I had tried with `testAar`, but seems like groovy gradle messed it because of the dynamic test configurations. I have now passing tests.

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
```diff
@@ -0,0 +1,87 @@
...
+                    removeObserver(this)
+                    job.cancel()
+                }
+            }
+        })
+    }
+}
+
+private val allowedCancelEvents = arrayOf(ON_PAUSE, ON_STOP, ON_DESTROY)
+private val lifecycleJobs = ConcurrentHashMap<Lifecycle, Job>()
```
gildor: Why do you need this (and lifecycleCoroutineScopes) global states? Isn't would better to create new Job each time this significantly simplify code and also allows to avoid shared mutable state (so one part of the code cancels a job that required for another).
If someone really wants to share a job, he can just cache instance of job.

I think current implementation can cause bugs and not enough explicit

LouisCAD: When getting a property, I don't expect it to give me a new value each time, that's why I implemented caching, and it is more efficient if you call it multiple times.

Keep in mind that for scopes, the `job` is only used as a parent, so you should not cancel the scope's job yourself when it's already done for you, like here with `Lifecycle`s.

In the cases where you want your scope to be shorter lived, you can still use `createScope()` or `createJob()`, and you can use the latter to make your custom lifecycle aware scope.

About misuse, you can always write `coroutineContext[Job]?.cancel()` on any scope and possibly cause bugs. Do you think will use `lifecycle.job` and then call `cancel()` on it? Also, do you think I should make it private and expose only `coroutineScope` and the two create methods?

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Note that [cancelEvent] supports only "**down**" events. That means that only [ON_PAUSE],
+ * [ON_STOP] and [ON_DESTROY] are supported.
+ *
+ * If the [Lifecycle] is already destroyed, then the returned [Job] comes already cancelled.
+ */
+fun Lifecycle.createJob(cancelEvent: Lifecycle.Event = ON_DESTROY): Job {
+    require(cancelEvent in allowedCancelEvents) { "$cancelEvent is forbidden for createJob(…)." }
+    return Job().also { job ->
+        if (currentState == Lifecycle.State.DESTROYED) job.cancel()
```
pakoito: Unless done in the main thread, this check is not thread safe, as the currentState can change and the stateChanged event can happen between lines 24 and 25 without being captured.

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
pakoito: Do `Lifecycle` instances have an stable identity? It seems like it compares references, and as anonymous objects they can capture state and leak.

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
LouisCAD: You're commenting on outdated code, please use the "Files changed" tab instead of a specific commit.
However, the logic you're pinpointing has not changed so much. This is just a fast path. If it goes to the `else` case, then, the observer is added, and it'll receive the events, even if it's already destroyed. I just added `Test observer is called after destroy` to check this.

LouisCAD: In a few words: The way AndroidX lifecycles are implemented makes this thread safe.

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
LouisCAD: @pakoito Yes, `Lifecycle` instances are never reused. After it is destroyed, it will be garbage collected, unless you keep a strong reference to it. Regarding leaks, there's none because the entries are always removed from the map after the `currentState` is no longer at least the passed `activeWhile` state (and `DESTROYED` is forbidden as a parameter as it does not make sense and would create leaks anyway, both semantically and technically).

LouisCAD: Marking this as resolved, if you still have concerns, please, use the latest version (in the **Files changed** tab) before you make a review or add a comment.

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
pakoito: Okay, I didn't know the latest state would fire after subscribing. LGTM then.

LouisCAD:
Closed in favor of #726

PR #662, opened 2018-10-04
Added @Keep annotations for CoroutineExceptionHandler and MainDispatсherFactory which are loaded via ServiceLoader.load() on Android.

trashkalmar:
Android app crashes in runtime when obfuscated with default Proguard settings. This PR adds `@Keep` annotation for two classes which are located via `ServiceLoader.load()`.


fvasco:
Does this PR work for Kotlin/JS?

trashkalmar:
Hm, it seems, `@Keep` is Android specific, so it will not help on other platforms.

JakeWharton:
You need to define an @OptionalExpecation expect annotation named Keep in the common code and then actual typealias it to the real annotation in the JDK code.

fvasco:
The PR is on the core module, so it is not possible to import the `android.support.annotation.Keep` annotation.

Moreover this PR does not include the imports.

JakeWharton:
Hence my comment which shows you how to do it correctly

PR #667, opened 2018-10-05
Align rxFlowable nullability with the Reactive Streams spec

### LINKED ISSUE #347: rxSingle, rxObservable nullity

DmitriyZaitsev:
Resolves #347


qwwdfsad:
Thanks!

Issue #261, opened 2018-02-27
Sliceable dispatchers: Provide alternative to newSingle/FixedThreadPoolContext via a shared pool of threads 


elizarov:
### Background

`newFixedThreadPoolContext` is actively used in coroutines code as a concurrency-limiting mechanism.  For example, to limit a number of concurrent request to the database to 10 one typically defines:

```
val DB = newFixedThreadPoolContext(10, "DB")
```

and then wraps all DB invocation into `withContext(DB) { ... }` blocks.

This approach have the following problems:

* This `withContext(DB)` invocation performs an actual switch to a different thread which is extremely expensive.
* The result of `newFixedThreadPoolContext` references the underlying threads and must be explicitly closed when no longer used. This is quite error-prone as programmers may use `newFixedThreadPoolContext` in their code without realizing this fact, thus leaking threads.

### Solution

The plan is to reimplement `newFixedThreadPoolContext` from scratch so that it does not create any threads. Instead, there will be one shared pool of threads that creates new thread strictly when they are needed. Thus, `newFixedThreadPoolContext` does not create its own threads, but acts only as a semaphore that limits the number of concurrent operations running in this context. 

Moreover, `DefaultContext`, which is currently equal to `CommonPool` (backed by `ForkJointPool.commonPool`), is going to be redefined in this way:

```
val DefaultContext = newFixedThreadPoolContext(defaultParallelism, "DefaultContext")
```

> The current plan is to set `defaultParallelism` to `nCPUs + 1` as a compromise value that ensures utilization of the underlying hardware even if one coroutine accidentally blocks and helps us avoid issue #198 

Now, with this redefinition of `DefaultContext` the code that is used to define its own `DB` context continues to work as before (limiting the number of concurrent DB operations). However, both issues identified above are solved:

* This `withContext(DB)` invocation does not actually perform thread context switch anymore. It only switches coroutine context and separately keeps track of and limits the number of concurrently running coroutines in `DB` context.
* There is not need to close `newFixedThreadPoolContext` anymore, as it is not backed by any physical threads, no risk of leaking threads.

This change also affects `newSingleThreadContext` as its implementation is:

```
fun newSingleThreadContext(name: String) = newFixedThreadPoolContext(1, name)
```

[...]

PR #669, opened 2018-10-06
Mark newSingle/FixedThreadPoolContext as obsolete, document the reason

elizarov:
* The will be replaced by another mechanism in the future. See #261 for details.
* The proposed replacement is to use the standard java API: `Executors.newSingleThreadExecutor/newFixedThreadPool` and convert to dispatcher via `asCoroutineDispatcher()` extension.

### DIFF
--- a/core/kotlinx-coroutines-core/src/ThreadPoolDispatcher.kt
+++ b/core/kotlinx-coroutines-core/src/ThreadPoolDispatcher.kt
@@ -14,8 +14,19 @@ import kotlin.coroutines.experimental.*
  * **NOTE: The resulting [ExecutorCoroutineDispatcher] owns native resources (its thread).
  * Resources are reclaimed by [ExecutorCoroutineDispatcher.close].**
  *
+ * **NOTE: This API will be replaced in the future**. A different API to create thread-limited thread pools
+ * that is based on a shared thread-pool and does not require the resulting dispatcher to be explicitly closed
+ * will be provided, thus avoiding potential thread leaks and also significantly improving performance, due
+ * to coroutine-oriented scheduling policy and thread-switch minimization.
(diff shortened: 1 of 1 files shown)


Dico200:
I do not think that `singleThreadContext` should be deprecated and phased out. I think its functionality should remain a part of the API. 

This is because limiting concurrent invocation of some set of methods/code to 1, is not the same as limiting it to one specific thread. Somewhere, you or someone else mentioned that it would break code if it uses `ThreadLocal`, however, I would hope that use case is not very common, as that should be unnecessary when there's just one thread already.

The problem I see is with regular mutable fields/properties not updating immediately on other threads unless they are volatile. The `limitedThreadContext` does permit access from various threads, even if it's only one at the same time. If the thread changes, field reads from the new thread might be outdated.

While the replacements you suggest are appropriate, I think there should be an option within the API that does the same as the current `newSingleThreadContext`. Whether that's by keeping it as is or creating an alternative, doesn't matter for this argument.

elizarov:
@Dico200 Have you checked out issue #261? It explain the solution to the same problem (limiting # of threads), but without actually creating a separate thread pool (that would need to be disposed of) for that.

Dico200:
@elizarov Yes, what I said was in light of the change proposed in #261.

I checked it again and I noticed I skimmed past these words a bit too quickly:

> >... some code in the wild that assumed that everything working in `newSingleThreadContext` was indeed happening in the single instance of `Thread`...

I missed the part with "in the single instance of Thread", which is the use case I describe.

However, I don't see anyone suggesting that it should stay for the reason I specified.

qwwdfsad:
Let's move the discussion to #261.
I will merge this PR as it doesn't break anything and will not be removed in 1.0.0.

PR #674, opened 2018-10-07
Added type <Unit> to be consistent with the text

olbpetersson:
In the text it states that "We explicitly specificy its `Unit` return type" but the code example does not explicitly set it


qwwdfsad:
Thanks!

PR #683, opened 2018-10-08
Expose silent internal clock overflows in TestCoroutineContext.

spand:
When writing tests I find that its better to be strict and explicit. This particular change would have exposed that I called `advanceTimeBy` instead of `advanceTimeTo` when using `Long.MAX_VALUE` as a "much later" value.


qwwdfsad:
Closing this, test API from `kotlinx-coroutines-core` is obsolete

Issue #673, opened 2018-10-07
Need Instructions For Using Gradle Kotlin Plugin


russel:
The README gives instructions for using Gradle but with the Groovy script,need to add what to do for when using the Gradle Kotlin plugin to write the build script in Kotlin.

elizarov:
Good catch. PRs are welcome.

russel:
My worry is that my hack is not a good bit of Kotlin code for a Gradle script, so I am not sure I have the backing to make a PR, yet. If there was someone from the Gradle Kotlin Plugin group who could advise that would help.

PR #685, opened 2018-10-08
Need Instructions For Using Gradle Kotlin Plugin #673

IRus:
Sample project available here: https://github.com/IRus/gradle-kotlin-dsl-samples/tree/kotlinx-coroutines-readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,32 @@ repository {
 }
 ```
 
+### Gradle Kotlin DSL
+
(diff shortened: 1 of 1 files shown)


elizarov:
Thanks. I've merged that to `develop`. Wanted to restructure it a bit (to avoid must duplication between Gradle/Grade Kotlin section), but cannot figure out how.

PR #695, opened 2018-10-10
Release candidate

qwwdfsad:
From branch `ea13`.

* Migration to Kotlin 1.3
* All deprecated API is removed
* Internal API became `internal` where it is possible
* Some workarounds for compiler bugs are removed
* Coroutines API annotations are `@Experimental`


elizarov:
We need to check `ktor` against this RC to make sure nothing extra was removed.

qwwdfsad:
Will check against Ktor (already checked `kotlinx.io`) before merge

elizarov:
`ExecutorService.asCoroutineDispatcher` shall have `@JvmName("from")`, too.

elizarov:
For consistency, we should have `@JvmName("from")` on `ThreadLocal.asContextElement`, too.

elizarov:
Looks good for me now.

PR #708, opened 2018-10-16
Make ReceiveChannel<E>.iterator() extension function

ndkoval:
Make `ReceiveChannel<E>.iterator()` extension function and simplify the implementation.

This new implementation is a bit less efficient, what can be fixed after a function like `receiveOrClosed` is introduced.

Note that this change breaks the binary compatibility.

### DIFF
--- a/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
+++ b/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
@@ -439,6 +439,10 @@ public abstract interface class kotlinx/coroutines/channels/ActorScope : kotlinx
 	public abstract fun getChannel ()Lkotlinx/coroutines/channels/Channel;
 }
 
+public final class kotlinx/coroutines/channels/ActorScope$DefaultImpls {
+	public static synthetic fun iterator (Lkotlinx/coroutines/channels/ActorScope;)Lkotlinx/coroutines/channels/ChannelIterator;
+}
+
 public abstract interface class kotlinx/coroutines/channels/BroadcastChannel : kotlinx/coroutines/channels/SendChannel {
 	public abstract fun cancel (Ljava/lang/Throwable;)Z
 	public abstract fun openSubscription ()Lkotlinx/coroutines/channels/ReceiveChannel;
@@ -464,6 +468,10 @@ public abstract interface class kotlinx/coroutines/channels/Channel : kotlinx/co
 	public static final field Factory Lkotlinx/coroutines/channels/Channel$Factory;
 	public static final field RENDEZVOUS I
 	public static final field UNLIMITED I
+}
+
+public final class kotlinx/coroutines/channels/Channel$DefaultImpls {
+	public static synthetic fun iterator (Lkotlinx/coroutines/channels/Channel;)Lkotlinx/coroutines/channels/ChannelIterator;
 }
 
 public final class kotlinx/coroutines/channels/Channel$Factory {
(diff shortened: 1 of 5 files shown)


binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
```diff
@@ -643,7 +647,6 @@ public abstract interface class kotlinx/coroutines/channels/ReceiveChannel {
 	public abstract fun getOnReceiveOrNull ()Lkotlinx/coroutines/selects/SelectClause1;
 	public abstract fun isClosedForReceive ()Z
 	public abstract fun isEmpty ()Z
-	public abstract fun iterator ()Lkotlinx/coroutines/channels/ChannelIterator;
```
qwwdfsad: Can we have this method in 1.0.0-RC as hidden + LowPriorityInOverloadResolution?
Breaking binary compatibility (especially in iteration over the channel) in RC doesn't seem right

common/kotlinx-coroutines-core-common/src/channels/ChannelIterator.kt
```diff
@@ -0,0 +1,119 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.internal.Symbol
+import kotlinx.coroutines.yield
+
+/**
+ * Iterator for [ReceiveChannel]. Instances of this interface are *not thread-safe* and shall not be used
+ * from concurrent coroutines.
+ */
+interface ChannelIterator<out E> {
```
qwwdfsad: `public` should not be omitted

common/kotlinx-coroutines-core-common/src/channels/ChannelIterator.kt
```diff
@@ -0,0 +1,119 @@
...
+     */
+    suspend operator fun next(): E
+}
+
+/**
+ * Returns new iterator to receive elements from this channels using `for` loop.
+ * Iteration completes normally when the channel is [isClosedForReceive] without cause and
+ * throws the original [close][SendChannel.close] cause exception if the channel has _failed_.
+ */
+operator fun <E> ReceiveChannel<E>.iterator(): ChannelIterator<E> {
```
qwwdfsad: `public`

common/kotlinx-coroutines-core-common/src/channels/ChannelIterator.kt
```diff
@@ -0,0 +1,119 @@
...
+    return object: ChannelIterator<E> {
+        var result: Any? = NO_RESULT // NO_RESULT | E (next element) | ClosedResult
+
+        override suspend fun hasNext(): Boolean {
+            if (result != NO_RESULT) return checkNotClosed(result)
+            // Try to receive an element. Store the result even if
+            // receiving fails in order to process further [hasNext]
+            // and [next] invocations properly.
+            try {
+                result = c.receive()
```
qwwdfsad: please write here // TODO use receive or closed when it's here

common/kotlinx-coroutines-core-common/src/channels/ChannelIterator.kt
```diff
@@ -0,0 +1,119 @@
...
+                    this.result = NO_RESULT
+                    return result as E
+                }
+            }
+        }
+    }
+}
+
+private val NO_RESULT = Symbol("NO_RESULT")
+private class ClosedResult(val cause: Throwable?) {
```
qwwdfsad: `@JvmField val cause`

common/kotlinx-coroutines-core-common/src/channels/ChannelIterator.kt
ndkoval: Why should we add a redundant modifier?

binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
ndkoval: Sure. Look at the updated `kotlinx-coroutines-core.txt` please, I am not sure about these changes.

elizarov:
(I'm taking this, will fix this + few more style issues)

elizarov:
Stop. We need to study performance impact of this change. We don’t have to include it into release. Can wait further optimizations.

PR #709, opened 2018-10-16
Add workaround for androidx toolchain issues: try to load main dispat…

### LINKED ISSUE #657: Android app with coroutines 0.30.1-eap13 crashes in runtime
```
java.lang.IllegalStateException: Module with Main dispatcher is missing. Add dependency with required Main dispatcher, e.g. 'kotlinx-coroutines-android'
        at kotlinx.coroutines.Dispatchers.getMain(Dispatchers.kt:61)
```
`kotlinx-coroutines-android` is obviously added.
Version 0.30.0-eap13 works fine.

qwwdfsad:
…cher via Class.forName(..) if ServiceLoader failed to find dispatcher

Fixes #657

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ exception is logged before crashing Android application, similarly to the way un
 threads are handled by Android runtime. 
 
 ### R8 and ProGuard
+
 If you are using R8 or ProGuard add the options from [coroutines.pro](core/kotlinx-coroutines-core/resources/META-INF/proguard/coroutines.pro) file to your rules.
 
 ## Building 
--- a/README.md
+++ b/README.md
@@ -129,17 +129,8 @@ coroutine dispatcher and also makes sure that in case of crashed coroutine with 
 exception is logged before crashing Android application, similarly to the way uncaught exceptions in 
 threads are handled by Android runtime. 
 
-### ProGuard
-
-In obfuscated code, fields with different types can have the same names,
-and `AtomicReferenceFieldUpdater` may be unable to find the correct ones.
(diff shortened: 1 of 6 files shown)


JakeWharton:
Why this over a `@Keep` (with `actual`/`expect`) or embedding the ProGuard rules inside the jar?

JakeWharton:
You can avoid an annotation in the public API with an `internal expect annotation` in common and an `internal actual typealias` to the real one in JDK.

Only R8 works with embedded rules, but that shouldn't preclude their use. What's worse is shipping rules solely as documentation that require copy/pasting. Embed the rules in the JAR and have it be the source of truth. Have your documentation link to the rules for people using old toolchains. We should be optimizing for the future so that more and more people will get behavior that Just Works™.

README.md
```diff
@@ -129,17 +129,8 @@ coroutine dispatcher and also makes sure that in case of crashed coroutine with
...
-and `AtomicReferenceFieldUpdater` may be unable to find the correct ones.
-To avoid field overloading by type during obfuscation, add this to your config:
-
-```
--keepclassmembernames class kotlinx.** {
-    volatile <fields>;
-}
-```
+### R8 and ProGuard
+If you are using R8 or ProGuard add the options from [coroutines.pro](core/kotlinx-coroutines-core/resources/META-INF/proguard/coroutines.pro) file to your rules.
```
elizarov: Style note: let's put an empty line after titles (it reads better in source, since it is actually a separate paragraph).

PR #718, opened 2018-10-17
Rework Dispatchers.Unconfined to use thread-local event loop

qwwdfsad:


elizarov:
Let's move its implementation to `DispatchedContinuation`, support it for all dispatchers that return `false` from `isDispatchNeeded`, and change scheduling policy to FIFO queue.

common/kotlinx-coroutines-core-common/src/Dispatched.kt
```diff
@@ -182,12 +266,15 @@ internal fun <T> DispatchedTask<T>.dispatch(mode: Int = MODE_CANCELLABLE) {
...
     }
-    // slow-path - use delegate
-    val state = takeState()
-    val exception = getExceptionalResult(state)
-    if (exception != null) {
-        delegate.resumeWithExceptionMode(exception, useMode)
-    } else {
-        delegate.resumeMode(getSuccessfulResult(state), useMode)
+
+    UndispatchedEventLoop.execute(this) {
```
qwwdfsad: ^^^ bug

common/kotlinx-coroutines-core-common/src/Dispatched.kt
```diff
@@ -11,11 +11,66 @@ import kotlin.jvm.*
 @Suppress("PrivatePropertyName")
 private val UNDEFINED = Symbol("UNDEFINED")
 
+@NativeThreadLocal
+internal object UndispatchedEventLoop {
+    data class State(
```
elizarov: Rename, please, this class are the corresponding state to something other than `State`. It is confusing with `DispatchedContinuation.state`

common/kotlinx-coroutines-core-common/src/internal/ArrayQueue.kt
```diff
@@ -0,0 +1,46 @@
...
+    private var tail = 0
+
+    public fun addLast(element: T) {
+        elements[tail] = element
+        tail = (tail + 1) and elements.size - 1
+        if (tail == head) ensureCapacity()
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    public fun removeFirst(): T {
```
elizarov: Consider replacing it with `removeFirstOrNull`. It should simplify this code (no need for `isEmpty` and no need for `check` here).

common/kotlinx-coroutines-core-common/src/internal/ArrayQueue.kt
```diff
@@ -0,0 +1,46 @@
...
+        require(head != tail) { "Queue is empty" }
+        val element = elements[head]
+        elements[head] = null
+        head = (head + 1) and elements.size - 1
+        return element!! as T
```
elizarov: Consider dropping `<T: Any>` requirement and (leave just `<T>`). It would remove the need to have `!!` null-check here.

PR #726, opened 2018-10-18
New integration: AndroidX Lifecycle

LouisCAD:
This PR replaces #655. It targets non experimental coroutines.

Includes:
- Job and CoroutineScope extensions for Lifecycle and LifecycleOwner
- Transitive dependency to kotlinx-android artifact
- Tests and test supporting code
- Documentation
- Entry in binary-compatibility-validator build.gradle file

Note that the binary compatibility task didn't succeed for the `release-candidate` branch after my edits (it was working in #655). Also, the `knit` warns for this new module regarding dokka documentation and unresolved references. I don't know how to solve these, but the maintainers can edit this PR, and you're free to help.

### DIFF
--- a/integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
+++ b/integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
@@ -1,0 +1,92 @@
+package kotlinx.coroutines.androidx.lifecycle
+
+import androidx.lifecycle.GenericLifecycleObserver
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.Lifecycle.State.DESTROYED
+import androidx.lifecycle.Lifecycle.State.INITIALIZED
+import androidx.lifecycle.LifecycleOwner
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Returns a [CoroutineScope] that uses [Dispatchers.Main] by default, and that is cancelled when
+ * the [Lifecycle] reaches [Lifecycle.State.DESTROYED] state.
+ *
+ * Note that this value is cached until the Lifecycle reaches the destroyed state.
+ */
+val Lifecycle.coroutineScope: CoroutineScope
+    get() = lifecycleCoroutineScopes[this] ?: job.let { job ->
+        val newScope = CoroutineScope(job + Dispatchers.Main)
+        if (job.isActive) {
+            lifecycleCoroutineScopes[this] = newScope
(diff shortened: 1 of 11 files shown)


LouisCAD:
Should I rebase this on the master branch?

integration/kotlinx-coroutines-androidx-lifecycle/src/Lifecycle.kt
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+/**
+ * Returns a [CoroutineScope] that uses [Dispatchers.Main] by default, and that is cancelled when
+ * the [Lifecycle] reaches [Lifecycle.State.DESTROYED] state.
+ *
+ * Note that this value is cached until the Lifecycle reaches the destroyed state.
+ */
+val Lifecycle.coroutineScope: CoroutineScope
+    get() = lifecycleCoroutineScopes[this] ?: job.let { job ->
```
twyatt: I don't believe this is thread safe.

I think the following scenario could occur:

- `Thread 1` reads `coroutineScope` and `this` is not in `lifecycleCoroutineScopes` collection, so it starts creating a new scope
- `Thread 2` reads `coroutineScope` and it **also** determines that `this` is not in `lifecycleCoroutineScopes` collection
- `Thread 1` finishes creating `CoroutineScope` and adds it to `lifecycleCoroutineScopes` collection
- `Thread 3` reads `coroutineScope` (and it retrieves it from `lifecycleCoroutineScopes` collection) and launches a Coroutine under the scope
- `Thread 2` finishes creating `CoroutineScope` and adds it to `lifecycleCoroutineScopes` collection
    - So we've lost reference to the `CoroutineScope` that `Thread 3` used to launch it's Coroutine

LouisCAD: @twyatt It is still thread-safe, because this is only for cache. This is not a problem if multiple job are created because of multi-thread access, as an attempt to remove them from the cache is performed anyway when the lifecycle is destroyed (which is independent from the cache that may be overridden).

LouisCAD: @twyatt In your scenario, all threads accessed a valid instance of a scope, and all of these instances will eventually be destroyed with the `LifecycleOwner`.

hrach:
@qwwdfsad why closed without a note?

qwwdfsad:
Sorry, I've removed `released-candidate` branch and this issue had closed automatically by github.
@LouisCAD please rebase your branch on develop and reopen the PR.

qwwdfsad:
https://screencast.com/t/qoP56Lufa still can't :(
Please create a fresh PR then, sorry for the inconvenience

LouisCAD:
@qwwdfsad Done in #760! 😄

PR #739, opened 2018-10-20
Undeprecate Channel.receiveOrNull

elizarov:
* It is a valid method for channels of non-null elements that is
  also consistent with Kotlin stdlib tradition.
* However, leave Channel.onReceiveOrNull since this is a tricker one,
  because on failure it aborts the whole select without letting figure
  out which clause failed.

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/channels/Channel.kt
+++ b/common/kotlinx-coroutines-core-common/src/channels/Channel.kt
@@ -206,14 +206,7 @@ public interface ReceiveChannel<out E> {
      *
      * This function can be used in [select] invocation with [onReceiveOrNull] clause.
      * Use [poll] to try receiving from this channel without waiting.
-     *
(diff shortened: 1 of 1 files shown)


jcornaz:
What about having this kind of function instead?

```kotlin
suspend inline fun <E> ReceiveChannel<E>.receiveOr(action: () -> E): E {
  return try {
    receive()
  } catch (closedException: ClosedReceiveChannelException) {
    action()
  }
}
```

This let the user choose what to do if the the channel is closed.

```kotllin
val nullableValue = channel.receiveOr { null }
val nonNullableValue1 = channel.receiveOr { return }
val nonNullableValue2 = channel.receiveOr { throw Exception() }
```

qwwdfsad:
Fixed in develop.
Deprecation is removed, but "obsolete" marker is not.

PR #743, opened 2018-10-22
Make select-expression & shared state and concurrency runnable

AlexanderPrendota:
knit - OK
:kotlinx-coroutines-core:test - OK


elizarov:
Merged into master.

PR #750, opened 2018-10-22
Implement yield for unconfined dispatchers

### LINKED ISSUE #737: Yield support in unconfined and immediate dispatchers
Currently, `yield` is no-op for any dispatcher which returns `false` for `isDispatchNeeded`.
But after #704 we can introduce support for `yield`, which may be useful for dependent unconfined coroutines in a blocking API bridges.

qwwdfsad:
Fixes #737

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/Dispatched.kt
+++ b/common/kotlinx-coroutines-core-common/src/Dispatched.kt
@@ -21,10 +21,17 @@ internal object UndispatchedEventLoop {
     @JvmField
     internal val threadLocalEventLoop = CommonThreadLocal { EventLoop() }
 
-    inline fun execute(continuation: DispatchedContinuation<*>, contState: Any?, mode: Int, doYield: Boolean = false, block: () -> Unit) : Boolean {
+    /**
+     * Executes given [block] as part of current event loop, updating related to block [continuation]
+     * mode and state if continuation is not resumed immediately.
+     * [doYield] indicates whether current continuation is yielding (to provide fast-path if event-loop is empty).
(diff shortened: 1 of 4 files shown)


common/kotlinx-coroutines-core-common/src/Dispatched.kt
```diff
@@ -21,26 +21,33 @@ internal object UndispatchedEventLoop {
     @JvmField
     internal val threadLocalEventLoop = CommonThreadLocal { EventLoop() }
 
-    inline fun execute(continuation: DispatchedContinuation<*>, contState: Any?, mode: Int, block: () -> Unit) {
+    inline fun execute(continuation: DispatchedContinuation<*>, contState: Any?, mode: Int, doYield: Boolean = false, block: () -> Unit) : Boolean {
```
elizarov: 1. Boolean result needs documentation on its meaning
2. This line (with addition of doYield parameter) is too long.

common/kotlinx-coroutines-core-common/src/Dispatched.kt
```diff
@@ -227,6 +234,12 @@ internal interface DispatchedTask<in T> : Runnable {
     }
 }
 
+internal fun DispatchedContinuation<Unit>.yield(): Boolean {
```
elizarov: 1. Maybe rename to `yieldUndispatched` & document the meaning of boolean result. In the context of this particular file this extension function is hard to understand.
2. Also refactor to idiomatic expression body, please, since this function does not do anything else but delegates to `UndispatchedEventLoop.execute` and this intent should be immediately clear.

PR #762, opened 2018-10-25
Implement receiveOrClosed

### LINKED ISSUE #330: Implement receiveOrClosed for channels and select clause
We need `receiveOrClosed` method on top of channels and `onReceiveOrClosed` for `select` which will have return type `Either<E, CloseToken>`.

Currently, pattern to work with channels always has the following form which is clumsy and obfuscates code:
```
 try {
    source.onReceive { ... }
} catch (e: ClosedReceiveChannelException) {
  // source is closed
}
```

The better alternative may be 
```
source.onReceiveOrClosed {
  when (it) {
    CloseToken -> // end up consumption
    _ -> // proceed normally
  }
}
```

For implementation to be efficient we need inline classes (https://github.com/Kotlin/KEEP/issues/104)

qwwdfsad:
Fixes #330

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/channels/Channel.kt
+++ b/common/kotlinx-coroutines-core-common/src/channels/Channel.kt
@@ -12,6 +12,7 @@ import kotlinx.coroutines.channels.Channel.Factory.RENDEZVOUS
 import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
 import kotlinx.coroutines.selects.*
 import kotlin.jvm.*
+import kotlin.internal.*
 
 /**
  * Sender's interface to [Channel].
@@ -188,10 +189,6 @@ public interface ReceiveChannel<out E> {
     public val onReceive: SelectClause1<E>
 
     /**
-     * Retrieves and removes the element from this channel suspending the caller while this channel [isEmpty]
-     * or returns `null` if the channel is [closed][isClosedForReceive] without cause
-     * or throws the original [close][SendChannel.close] cause exception if the channel has _failed_.
-     *
      * This suspending function is cancellable. If the [Job] of the current coroutine is cancelled or completed while this
      * function is suspended, this function immediately resumes with [CancellationException].
      *
@@ -207,12 +204,14 @@ public interface ReceiveChannel<out E> {
      * This function can be used in [select] invocation with [onReceiveOrNull] clause.
      * Use [poll] to try receiving from this channel without waiting.
      *
-     * **Note: This is an obsolete api.**
-     * This function will be replaced with `receiveOrClosed: ReceiveResult<E>` and
-     * extension `suspend fun <E: Any> ReceiveChannel<E>.receiveOrNull(): E?`
-     * It is obsolete because it does not distinguish closed channel and null elements.
+     * This function is deprecated  because it does not distinguish closed channel and null elements.
+     * Use [receiveOrClosed] or [receiveOrNull] extension method.
      */
     @ObsoleteCoroutinesApi
+    @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
+    @LowPriorityInOverloadResolution
+    @Deprecated(message = "Deprecated in favor of receiveOrClosed and receiveOrNull extension",
+        level = DeprecationLevel.WARNING, replaceWith = ReplaceWith("receiveOrClosed"))
     public suspend fun receiveOrNull(): E?
 
     /**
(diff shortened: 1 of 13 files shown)


qwwdfsad:
I will revisit this PR tomorrow

common/kotlinx-coroutines-core-common/src/channels/Channel.kt
```diff
@@ -268,6 +301,75 @@ public interface ReceiveChannel<out E> {
...
+     * Returns `true` if this instance represents close cause.
+     * In this case [isValue] returns `false`.
+     */
+    public val isClosed: Boolean get() = holder is Closed
+
+    /**
+     * Returns received value if this instance represents received value or throws [IllegalStateException] otherwise.
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val value: T
```
elizarov: So we have the following methods on `ValueOrClosed`:
* `value`
* `valueOrNull`
* `valurOrThrow`

Which is kind of confusing, since the channel has the following `receive` variants:

* `receive` which corresponds to `receiveOrClosed().valueOrThrow`... huh?
* `receiveOrNull` which corresponds to `receiveOrClosed().valueOrNull`
*  `receiveOrClosed`

For the sake of consistency I'd suggest to drop `value` and rename `valueOrThrow` to simply `value`.

common/kotlinx-coroutines-core-common/src/channels/Channel.kt
```diff
@@ -188,10 +189,6 @@ public interface ReceiveChannel<out E> {
     public val onReceive: SelectClause1<E>
 
     /**
```
elizarov: Accidentally lost the first paragraph of `receiveOrNull` documentation.

common/kotlinx-coroutines-core-common/src/channels/Channel.kt
```diff
@@ -268,6 +301,75 @@ public interface ReceiveChannel<out E> {
...
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val value: T
+        get() = if (isClosed) throw IllegalStateException() else holder as T
+
+    /**
+     * Returns received value if this element represents received value or `null` otherwise.
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val valueOrNull: T?
```
elizarov: It seems logical to also make it an extension that is available on non-null T types only, so that generic code does not accidentally use it.

docs/select-expression.md
```diff
@@ -178,14 +179,16 @@ suspend fun selectAorB(a: ReceiveChannel<String>, b: ReceiveChannel<String>): St
...
+            if (valueOrClosed.isClosed) 
                 "Channel 'b' is closed"
             else    
-                "b -> '$value'"
+                "b -> '${valueOrClosed.value}'"
         }
     }
 ```
+Note that [onReceiveOrNull][ReceiveChannel.onReceiveOrNull] is an extension function defined only for channels with non-nullable elements.
+
```
elizarov: This seems to be a good place in documentation to add an extended rationale on _why_ it is defined to be so (so that generic functions that are designed to operation on any `T` do not accidentally confuse closed channel and the null element)

elizarov:
An awesome piece of work! I've successfully cherry-picked all three commits to PR #1343 (`consume-as-flow` branch) with a few minor exceptions:
* `Channel.receiveOrClosed/onReciveOrClosed` and `ValueOrClosed` are `internal` for now, until [KT-27524](https://youtrack.jetbrains.com/issue/KT-27524) is resolved. 
* Thus we cannot mention them full in documentation just yet, so doc changes to `select-expression.md` cannot be merged.
* There are a couple of outstanding (non-resolved) issues in this conversation that should be either resolved or moved to #1343

common/kotlinx-coroutines-core-common/src/channels/Channel.kt
elizarov: I've moved this comment to the source in #1343 for future decision.

common/kotlinx-coroutines-core-common/src/channels/Channel.kt
elizarov: I've moved this comment to the source in #1343 for future decision.

common/kotlinx-coroutines-core-common/src/channels/Channel.kt
elizarov: I've moved this comment to the source in #1343 for future decision.

elizarov:
It seems that with #330 we can finally close this PR for good.

PR #764, opened 2018-10-26
Add uncaught exception handler in TestBase.before and validate that n…

qwwdfsad:
…o uncaught exceptions are present in TestBase.after

### DIFF
--- a/core/kotlinx-coroutines-core/test/TestBase.kt
+++ b/core/kotlinx-coroutines-core/test/TestBase.kt
@@ -118,7 +118,11 @@ public actual open class TestBase actual constructor() {
         initPoolsBeforeTest()
         threadsBefore = currentThreads()
         originalUncaughtExceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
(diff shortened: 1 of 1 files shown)


core/kotlinx-coroutines-core/test/TestBase.kt
```diff
@@ -105,13 +113,12 @@ public actual open class TestBase actual constructor() {
...
 
-    private lateinit var threadsBefore: Set<Thread>
-    private val SHUTDOWN_TIMEOUT = 10_000L // 10s at most to wait
-
     @Before
     fun before() {
         initPoolsBeforeTest()
         threadsBefore = currentThreads()
+        originalUncaughtExceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
+        Thread.setDefaultUncaughtExceptionHandler({ t, e -> uncaughtExceptions.add(e) })
```
elizarov: It makes sense to print the stacktrace of this exception to the console (just like original handler does). This would help to trace a problem in case when the test hangs (also, the assertion at `after` does not print stacktraces either).

PR #767, opened 2018-10-29
LazyStandaloneCoroutine contains an `val` reference to block

fvasco:
The class `LazyStandaloneCoroutine` set a `private val` reference to the `block` lambda.
This reference must be used only one time, but it cannot be reclaimed from GC until the `Job` instance is fully unreferenced.
`Job`s, `Deferred`s and producers should release all potentially large references as soon at possible.


fvasco:
This commit changes `LazyDeferredCoroutine`, `LazyStandaloneCoroutine` and `LazyBroadcastCoroutine`

qwwdfsad:
`ChannelCoroutine` is not changed, but I think it is okay as they have a different usage pattern.
I'll merge it as soon as it pass all tests

qwwdfsad:
Could you please rebase your branch on top of current `develop`?

PR #781, opened 2018-10-31
Dispatcher fixes

qwwdfsad:


core/kotlinx-coroutines-core/src/scheduling/Dispatcher.kt
```diff
@@ -36,15 +36,18 @@ internal object DefaultScheduler : ExperimentalCoroutineDispatcher() {
...
     private val corePoolSize: Int,
     private val maxPoolSize: Int,
-    private val idleWorkerKeepAliveNs: Long
+    private val idleWorkerKeepAliveNs: Long,
+    private val schedulerName: String = "CoroutineScheduler"
 ) : ExecutorCoroutineDispatcher() {
     constructor(
         corePoolSize: Int = CORE_POOL_SIZE,
-        maxPoolSize: Int = MAX_POOL_SIZE
+        maxPoolSize: Int = MAX_POOL_SIZE,
```
elizarov: This is not a binary compatible change and it is going to break existing code in ktor that uses this constructor. A new constructor that accepts a `schedulerName` parameters shall be added. Old one (this one) shall be deprecated&hidden, so that that next (recompiled) version of ktor resolves to the new one.

PR #789, opened 2018-11-03
Fix the example

naartjie:
I had to change the code to the following to make it work


qwwdfsad:
You have to import `kotlin.concurrent.thread` for this builder.

PR #792, opened 2018-11-04
Stacktrace recovery

### LINKED ISSUE #493: Basic stacktrace augmentation

qwwdfsad:
Basic exception stacktrace recovery mechanism  …
  * Implement CoroutineStackFrame in CancellableContinuationImpl, DispatchedContinuation and ScopeCoroutine
  * On coroutine resumption try to reflectively instantiate exception instance of the same type, but with augmented stacktrace
  * Recover stacktrace by walking over CoroutineStackFrame
  * Recover stacktrace on fast-path exceptions without CoroutineStackFrame walking to provide more context to an exception
  * Unwrap exceptions when doing aggregation in JobSupport
  * Add kill-switch to disable stacktrace recovery, introduce method to recover stacktrace on the exceptional fast-path
  * Add `suspendCoroutineOrReturn` on exceptional fast-path in await in order to provide "real" stacktrace

Design rationale:
All recovery of *suspended* continuations takes place in Dispatched.kt file, the only place where all calls to "resume*" ends up, so we don't have to remember about stacktrace recovery in every primitive we are implementing. But to provide more accurate stacktraces we *have to* recover it on every fast-path for better debuggability.

Fixes #493

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/Dispatched.kt
+++ b/common/kotlinx-coroutines-core-common/src/Dispatched.kt
@@ -81,11 +81,13 @@ internal object UndispatchedEventLoop {
 internal class DispatchedContinuation<in T>(
     @JvmField val dispatcher: CoroutineDispatcher,
     @JvmField val continuation: Continuation<T>
-) : Continuation<T> by continuation, DispatchedTask<T> {
+) : Continuation<T> by continuation, DispatchedTask<T>, CoroutineStackFrame {
     @JvmField
     @Suppress("PropertyName")
     internal var _state: Any? = UNDEFINED
     public override var resumeMode: Int = 0
+    override val callerFrame: CoroutineStackFrame? = continuation as? CoroutineStackFrame
+    override fun getStackTraceElement(): StackTraceElement? = null
     @JvmField // pre-cached value to avoid ctx.fold on every resumption
     internal val countOrElement = threadContextElements(context)
 
@@ -168,7 +170,7 @@ internal class DispatchedContinuation<in T>(
     @Suppress("NOTHING_TO_INLINE") // we need it inline to save us an entry on the stack
     inline fun resumeUndispatchedWithException(exception: Throwable) {
         withCoroutineContext(context, countOrElement) {
-            continuation.resumeWithException(exception)
+            continuation.resumeWithStackTrace(exception)
         }
     }
 
@@ -191,7 +193,7 @@ internal fun <T> Continuation<T>.resumeCancellable(value: T) = when (this) {
 
 internal fun <T> Continuation<T>.resumeCancellableWithException(exception: Throwable) = when (this) {
     is DispatchedContinuation -> resumeCancellableWithException(exception)
-    else -> resumeWithException(exception)
+    else -> resumeWithStackTrace(exception)
 }
 
 internal fun <T> Continuation<T>.resumeDirect(value: T) = when (this) {
@@ -200,8 +202,8 @@ internal fun <T> Continuation<T>.resumeDirect(value: T) = when (this) {
 }
 
 internal fun <T> Continuation<T>.resumeDirectWithException(exception: Throwable) = when (this) {
-    is DispatchedContinuation -> continuation.resumeWithException(exception)
-    else -> resumeWithException(exception)
+    is DispatchedContinuation -> continuation.resumeWithStackTrace(exception)
+    else -> resumeWithStackTrace(exception)
 }
 
 internal interface DispatchedTask<in T> : Runnable {
@@ -230,7 +232,7 @@ internal interface DispatchedTask<in T> : Runnable {
                 else {
                     val exception = getExceptionalResult(state)
                     if (exception != null)
-                        continuation.resumeWithException(exception)
+                        continuation.resumeWithStackTrace(exception)
                     else
                         continuation.resume(getSuccessfulResult(state))
                 }
@@ -272,3 +274,7 @@ internal fun <T> DispatchedTask<T>.resume(delegate: Continuation<T>, useMode: In
         delegate.resumeMode(getSuccessfulResult(state), useMode)
     }
 }
+
+
+@Suppress("NOTHING_TO_INLINE")
+private inline fun Continuation<*>.resumeWithStackTrace(exception: Throwable) = resumeWith(Result.failure(recoverStackTrace(exception, this)))
--- a/common/kotlinx-coroutines-core-common/src/Dispatched.kt
+++ b/common/kotlinx-coroutines-core-common/src/Dispatched.kt
@@ -277,4 +277,4 @@ internal fun <T> DispatchedTask<T>.resume(delegate: Continuation<T>, useMode: In
 
 
 @Suppress("NOTHING_TO_INLINE")
-private inline fun Continuation<*>.resumeWithStackTrace(exception: Throwable) = resumeWith(Result.failure(recoverStackTrace(exception, this)))
+internal inline fun Continuation<*>.resumeWithStackTrace(exception: Throwable) = resumeWith(Result.failure(recoverStackTrace(exception, this)))
--- a/common/kotlinx-coroutines-core-common/src/JobSupport.kt
+++ b/common/kotlinx-coroutines-core-common/src/JobSupport.kt
@@ -242,8 +242,9 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         val seenExceptions = identitySet<Throwable>(exceptions.size)
         var suppressed = false
         for (exception in exceptions) {
-            if (exception !== rootCause && exception !is CancellationException && seenExceptions.add(exception)) {
-                rootCause.addSuppressedThrowable(exception)
+            val unwrapped = unwrap(exception)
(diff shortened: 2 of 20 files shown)


common/kotlinx-coroutines-core-common/src/JobSupport.kt
```diff
@@ -1077,7 +1078,9 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
             val state = this.state
             if (state !is Incomplete) {
                 // already complete -- just return result
-                if (state is CompletedExceptionally) throw state.cause
+                if (state is CompletedExceptionally) { // Slow path to recover stacktrace
+                    recoverAndThrow(state.cause)
```
qwwdfsad: Such check should be inserted in every fast-path method of the channel

qwwdfsad:
This is still work-in-progress. It is worth reviewing, but I am actively trying to adjust this branch for debug agent purposes and won't merge it until debug agent is ready.

For example, I've found that recovery is stable only if stacktrace is recovered from `Dispatched.kt` (where all `resume*` end up eventually), thus changing the initial implementation accordingly.

core/kotlinx-coroutines-core/src/internal/ExceptionsConstuctor.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package kotlinx.coroutines.internal
+
+import java.util.*
+import java.util.concurrent.locks.*
+import kotlin.concurrent.*
+
+private val cacheLock = ReentrantReadWriteLock()
+private val exceptionConstructors: WeakHashMap<Class<*>, (Throwable) -> Throwable?> = WeakHashMap()
```
qwwdfsad: @cy6erGn0m had some concerns about this approach, though I don't see how `WeakReference<Class>` can prevent class unloading here.

qwwdfsad:
support in `withTimeout` is implemented in `debug-agent` branch

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -0,0 +1,114 @@
...
+        last.getStackTraceElement()?.let { stack.add(sanitize(it)) }
+    }
+    return stack
+}
+
+internal fun sanitize(element: StackTraceElement): StackTraceElement {
+    if (!element.className.contains('/')) {
+        return element
+    }
+    // STE generated with debug metadata contains '/' as separators in FQN, while Java contains dots
```
elizarov: A link to KT-XXX issue would be helpful here.

common/kotlinx-coroutines-core-common/src/Dispatched.kt
```diff
@@ -275,3 +277,7 @@ internal fun <T> DispatchedTask<T>.resume(delegate: Continuation<T>, useMode: In
         delegate.resumeMode(getSuccessfulResult(state), useMode)
     }
 }
+
+
+@Suppress("NOTHING_TO_INLINE")
+internal inline fun Continuation<*>.resumeWithStackTrace(exception: Throwable) = resumeWith(Result.failure(recoverStackTrace(exception, this)))
```
elizarov: Line is too long. Shall be wrapped.

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -0,0 +1,114 @@
...
+    if (recoveryDisabled(exception)) {
+        return exception
+    }
+
+    val copy = tryCopyException(exception) ?: return exception
+    return copy.sanitizeStackTrace()
+}
+
+private fun <E : Throwable> E.sanitizeStackTrace(): E {
+    val size = stackTrace.size
```
elizarov: Every read of `stackTrace` clones an array. Read it once, then analyze.

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    var lastIntrinsic = -1
+    for (i in 0 until size) {
```
elizarov: Use `indexOfFirst { ... }` instead of the loop to clearly show what you are doing here and avoid `var`.

common/kotlinx-coroutines-core-common/src/JobSupport.kt
elizarov: Looks ok to me. It does not affect that speed of fast-path (still no state-machine) and only slows down exceptions.

elizarov:
It looks like `withTimeout` (`TimeoutCoroutines`) shall be supported in this PR.

core/kotlinx-coroutines-core/src/internal/Exceptions.kt
```diff
@@ -47,15 +49,68 @@ internal actual fun <E : Throwable> recoverStackTrace(exception: E, continuation
...
+     * IllegalStateException
+     *   at foo
+     *   at kotlinx.coroutines.resumeWith
+     *   (Current coroutine stacktrace)
+     *   at bar
+     *   ...real stacktrace...
+     * caused by "IllegalStateException" (original one)
+     */
+    // TODO optimizable allocations and passes
+    stacktrace.addFirst(artificialFrame("Current coroutine stacktrace"))
```
qwwdfsad: do you think it is worth to append `continuation` to this string?

core/kotlinx-coroutines-core/src/internal/ExceptionsConstuctor.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.locks.*
+import kotlin.concurrent.*
+
+private val cacheLock = ReentrantReadWriteLock()
+// Replace it with ClassValue when Java 6 support is over
+private val exceptionConstructors: WeakHashMap<Class<out Throwable>, (Throwable) -> Throwable?> = WeakHashMap()
+
+@Suppress("UNCHECKED_CAST")
+internal fun <E : Throwable> tryCopyException(exception: E): E? {
+    val cachedCtor = cacheLock.read {
```
elizarov: Should not try to find constructor if already tried and did not found.

common/kotlinx-coroutines-core-common/src/internal/StackTraceRecovery.common.kt
```diff
@@ -0,0 +1,50 @@
...
+@Suppress("NOTHING_TO_INLINE")
+internal expect suspend inline fun recoverAndThrow(exception: Throwable): Nothing
+
+/**
+ * The opposite of [recoverStackTrace].
+ * It is guaranteed that `unwrap(recoverStackTrace(e)) === e`
+ */
+internal expect fun <E: Throwable> unwrap(exception: E): E
+
+expect class StackTraceElement
```
elizarov: Need to `@suppress` from the docs or make `internal`.

PR #800, opened 2018-11-06
Clarify docs for CancellableContinuation

qwwdfsad:
Let's update documentation on 1.0.1 and provide a more precise mechanism (via another intrinsic) separately.


common/kotlinx-coroutines-core-common/src/CancellableContinuation.kt
```diff
@@ -29,8 +29,10 @@ import kotlin.jvm.*
  *
  * A [cancelled][isCancelled] continuation implies that it is [completed][isCompleted].
  *
- * Invocation of [resume] or [resumeWithException] in _resumed_ state produces [IllegalStateException]
- * but is ignored in _cancelled_ state.
+ * Invocation of [resume] or [resumeWithException] in _resumed_ state produces [IllegalStateException].
+ * Invocation of [resume] in _cancelled_ state is ignore (it is a trivial race between resume from the continuation owner and
```
elizarov: typo: "is ignore" -> "is ignored"

PR #834, opened 2018-11-14
CoroutineScope extensions and MainScope

qwwdfsad:


elizarov:
We need use-cases for those extensions, e.g. they should be really used in the examples we give in documentation or otherwise it is not clear if they are really needed. `MainScope()` look good in this respect, but I have concerns about the rest.

In particular:

*  Why would anyone need `xxxOrNull` extensions? It seems that in the context where I might need `jobOrNull` the `coroutineContext[Job]` looks more appropriate.

* I'm concerned about `job` extension which is a very short identifier and thus would pollute completion context in many places unnecessarily. Also, it exposes `Job` context to novice users who should not be needing to understand. The same with `dispatcher` extension. Why would we need to expose it? How would one actually use it?

PR #837, opened 2018-11-14
Wrap all Incomplete final state into box (and unbox it where necessar…

### LINKED ISSUE #835: Puzzling hang in runBlocking
From @konrad-kaminski 

This code hangs, but does not hang without `invokeOnCompletion`:

```
fun main() {
    runBlocking {
        val job = GlobalScope.launch {  }
        job.invokeOnCompletion {  }
    }
}
```

Initial investigation shows that it is timing-depended, e.g. adding `println` statements inside `runBlocking` before `}` affects whether it hangs or not, too.

qwwdfsad:
…y) because otherwise job machinery treats such state as intermediate

Fixes #835

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/JobSupport.kt
+++ b/common/kotlinx-coroutines-core-common/src/JobSupport.kt
@@ -1034,7 +1034,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         private val job: JobSupport
     ) : CancellableContinuationImpl<T>(delegate, MODE_CANCELLABLE) {
         override fun getContinuationCancellationCause(parent: Job): Throwable {
-            val state = job.state.unboxState()
+            val state = job.state
             /*
              * When the job we are waiting for had already completely completed exceptionally or
              * is failing, we shall use its root/completion cause for await's result.
@@ -1059,7 +1059,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
     public val isCompletedExceptionally: Boolean get() = state is CompletedExceptionally
 
     public fun getCompletionExceptionOrNull(): Throwable? {
-        val state = this.state.unboxState()
+        val state = this.state
         check(state !is Incomplete) { "This job has not completed yet" }
         return state.exceptionOrNull
     }
@@ -1068,10 +1068,10 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
      * @suppress **This is unstable API and it is subject to change.**
      */
     internal fun getCompletedInternal(): Any? {
-        val state = this.state.unboxState()
+        val state = this.state
         check(state !is Incomplete) { "This job has not completed yet" }
         if (state is CompletedExceptionally) throw state.cause
-        return state
+        return state.unboxState()
     }
 
     /**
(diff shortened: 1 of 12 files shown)


common/kotlinx-coroutines-core-common/src/JobSupport.kt
```diff
@@ -1029,7 +1034,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         private val job: JobSupport
     ) : CancellableContinuationImpl<T>(delegate, MODE_CANCELLABLE) {
         override fun getContinuationCancellationCause(parent: Job): Throwable {
-            val state = job.state
+            val state = job.state.unboxState()
```
elizarov: Unboxing is not needed here and even harmful, since this code would not work problem in case when `Finishing` instance was boxed inside.

common/kotlinx-coroutines-core-common/src/JobSupport.kt
```diff
@@ -1054,7 +1059,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
     public val isCompletedExceptionally: Boolean get() = state is CompletedExceptionally
 
     public fun getCompletionExceptionOrNull(): Throwable? {
-        val state = this.state
+        val state = this.state.unboxState()
```
elizarov: Unboxing is not needed and even harmful here. This method would not work problem with boxed incomplete state.

common/kotlinx-coroutines-core-common/src/JobSupport.kt
```diff
@@ -1063,7 +1068,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
      * @suppress **This is unstable API and it is subject to change.**
      */
     internal fun getCompletedInternal(): Any? {
-        val state = this.state
+        val state = this.state.unboxState()
```
elizarov: Wrong `unbox` place. Should be in `return state` below.

core/kotlinx-coroutines-core/src/Builders.kt
```diff
@@ -80,7 +80,7 @@ private class BlockingCoroutine<T>(
         }
         timeSource.unregisterTimeLoopThread()
         // now return result
-        val state = this.state
+        val state = this.state.unboxState()
```
elizarov: It is slightly more efficient to unwrap at `return state` below.

common/kotlinx-coroutines-core-common/src/JobSupport.kt
```diff
@@ -220,7 +225,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
             handleJobException(finalException)
         }
         // Then CAS to completed state -> it must succeed
-        require(_state.compareAndSet(state, finalState)) { "Unexpected state: ${_state.value}, expected: $state, update: $finalState" }
+        require(_state.compareAndSet(state, finalState.boxIncomplete())) { "Unexpected state: ${_state.value}, expected: $state, update: $finalState" }
```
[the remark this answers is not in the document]
qwwdfsad: We can't because later this final state is used in `completeStateFinalization`

common/kotlinx-coroutines-core-common/src/JobSupport.kt
qwwdfsad: You are right, this one is unnecessary.
`Finishing` state can be boxed. I wanted to add `assert` (one with -ea), but it's JVM only

PR #838, opened 2018-11-14
Guava: Support building with Guava 27.

mpetrov:
Several methods were deprecated, and removed in Guava 27. This updates
existing calls to enable compiling with newer versions of Guava.


qwwdfsad:
I will merge it manually to develop when tests pass

mpetrov:
The failing checks look like flakes to me, not sure if there's a way to retry them.

qwwdfsad:
Merged manually, thanks

PR #848, opened 2018-11-21
Updated readme to make it clear, that for Proguard manual actions req…

### LINKED ISSUE #807: consumer proguard config is not added.
When building an apk with `org.jetbrains.kotlinx:kotlinx-coroutines-android` the [coroutines.pro](https://github.com/Kotlin/kotlinx.coroutines/blob/1.0.0/core/kotlinx-coroutines-core/resources/META-INF/proguard/coroutines.pro) is not added/applied to proguard configuration so it gets removed by progruard.

Is possible to change `org.jetbrains.kotlinx:kotlinx-coroutines-android` to an AAR instead of an JAR, then the consumer file is automatically used.

Dmitry-Borodin:
…uired for android.

Fixes #807

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,13 @@ coroutine dispatcher and also makes sure that in case of crashed coroutine with 
 exception is logged before crashing Android application, similarly to the way uncaught exceptions in 
 threads are handled by Android runtime. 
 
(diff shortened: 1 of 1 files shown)


elizarov:
Thanks

PR #866, opened 2018-11-30
Introduce MainScope factory and CoroutineScope.cancel extension

### LINKED ISSUE #829: Provide MainScope factory
We should provide factory method `fun MainScope() = CoroutineScope(Dispatchers.Main + SupervisorJob())`.

It fixes multiple issues (especially with the #828):

1. Integration with UI components are simple both for implementing an interface and having it in a field
2. It protects newcomers from `async` behaviour when it is launched from UI component
3. It simplifies guides and documentation

qwwdfsad:
Fixes #829

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,10 @@ GlobalScope.launch {
 
 ## Modules
 
-* [common](common/README.md) &mdash; common coroutines across all backends:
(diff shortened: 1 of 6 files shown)


elizarov:
Added a few style/wording fixes, otherwise looks good

Issue #840, opened 2018-11-16
Dispatchers.Default consumes too much CPU in the face of short bursts of work


qwwdfsad:
See https://discuss.kotlinlang.org/t/default-dispatcher-creates-too-many-threads/10143

`CoroutineScheduler` may consume too much CPU when there is no work to do (a lot of spins, yields and short parks before the transition to the idle state). What even worse, it can burn a lot of CPU in the pattern "a short burst of work, small idle period" (I remember @e5l has noticed that as well).

We should increase the speed of spinning -> idle transition, ideally without losing any performance for task-heavy benchmarks.

We probably can safely remove `yield` usages (this is the most suspicious place) and try to gracefully detect quiescent period

Tolriq:
Facing this also in a basic android app, a dozen of threads on a snapdragon 835 taking lot's of CPU at start until the 1,5 sec park starts.

Using basic stuff with actor / a few channels & worker pools.

I'm not sure to understand the correlation between the yield / spinning issue and the number of threads?

DanielThomas:
The number of threads is a factor because at low-loads it appears there's small enough amount of work that threads will spend their maximum time in `cpuWorkerIdle` because they don't receive any work in that time, but enough work that all of the threads in the core pool get woken to do work.

That makes the idle problem scale up with the number of threads. For instance, here's that same example with 1300% CPU utilization on my 10-core, 10-thread iMac Pro:

<img width="839" alt="screen shot 2018-12-10 at 12 00 22 pm" src="https://user-images.githubusercontent.com/1479220/49758319-7fc73380-fc73-11e8-80b9-46e59cbd5f80.png">

DanielThomas:
The linked discussion mentions setting `kotlinx.coroutines.scheduler=off` to move `Dispatchers.Default` back to `CommonPool`, but that doesn't help with usages of `Dispatchers.IO`.

Tolriq:
My question was more about the fact that a 4 core processor should not start 12 threads as I thought it would be limited to 4 per the doc, so was wondering is that high number of thread was related to that or something else with maybe a simpler fix :)

fvasco:
It is available a spin-wait hints for Java 9+

http://openjdk.java.net/jeps/285

rumatoest:
Hi. In my current setup with coroutines 1.3.3 I've struggle with the same CPU consuming issue.
Setting system property `kotlinx.coroutines.scheduler` to off fixed CPU issue.

[...]

PR #884, opened 2018-12-12
Do not yield in CoroutineScheduler during transition to the idle stat…

qwwdfsad:
…e, introduce system properties to tune this behaviour.

Rationale:

Thread.yield has a significant CPU cost (especially relatively to spins) and provides no significant benefits compared with exponential parking.
Thus yields burn a lot of CPU due to JVM upcall + syscall, providing benefits neither for liveness property (as CPU is mostly busy with doing these calls) nor for latencies. Initial benchmarking shows significant CPU usage reduction (200-300%) in low-to-average load benchmarks with no degradation on target affinity benchmarks.

Partially addresses #840


qwwdfsad:
Note that change does not completely fix #840, a transition to the idle state still can be significantly improved

jcornaz:
Hi, 

Sorry to disturb, but how can a reduction be more than 100%?

PR #886, opened 2018-12-12
Remove unnecessary function declaration from example

joffrey-bion:
The produceSquares() function is probably a leftover, because only produceNumbers() 
and square() are actually used in the code sample.


elizarov:
Yes. Thanks.

Issue #810, opened 2018-11-08
Possibility to replace default dispatchers


dmytrodanylyk:
So Rx java has a way to inject scheduler _(RxJavaPlugins)_, is there a plan to have a similar thing for coroutines dispatchers?

@gildor  
> I would add +1 for this.
Replacing dispatcher implementation is pretty common thing which of course can be solved by DI, but this adds not necessary boilerplate especially for functions and Inline classes (I have use case when would like to use Inline class to wrap blocking API with coroutines, but cannot because want to allow switch dispatcher)
This is also crucial for UI dispatcher and unit tests.
I agree that in most cases it can be handled by DI but a way to replace default dispatchers sounds reasonable for me

Christoph Sturm _(from slack)_
> is there a blocking scope that i can use to inject into classes that i test and make them run like they were invoked with runBlocking?
I have some classes that implement interfaces that are not suspending, so they have their own coroutine scope and use scope.launch {}. but in my tests it would be nice to run them blocking, or to be able to wait on the scope

elizarov:
I am not in favor of `RxJavaPlugins` approach where you can just `setXxx`. Just like any form of global (static) mutable state this leads to lots of practical problem. However, for testing purposes this seems like a reasonable and useful request. See PR #749 that we are working on.

qwwdfsad:
Let me clarify: we are not going to implement it in the form of global settable dispatchers like `RxJavaPlugins` does. For a core library, the probability that other libraries will set these global variables is high and conflicts in such cases just cannot be reasonably resolved.
The fact that some global state which is used by literally every library builder does not make it easier.

If you want to have control over dispatchers, use dependency injection and inversion of control.
Scoped concurrency is here for the rescue as one doesn't have to specify dispatcher everywhere, only in scope declaration.

>Maybe good solution would be to include IntelliJ warning/inspection to not override dispatchers in production code?

Inspections and warnings won't help here, because people will use it anyway ("I get used to it in RxJava, so I can ignore this warning").

dmytrodanylyk:
Sounds good. In most cases in Android result is delivered to the main thread anyway.

matejdro:
Another use case for this is ability to replace dispatchers with `TestCoroutineContext` to globally control timeouts and delays.

yigit:
Side effects is another case that needs some help for testing.

For instance, you can write espresso tests that look like:
```
button.click()
assert(contentView).hasData("foo")
```
[...]

Issue #541, opened 2018-09-10
Integrate TestCoroutineContext with structured concurrency


elizarov:
`TestCoroutineContext` (#297) should be integrated with structured concurrency (#410), so that all builders can be used inside the test context (custom builder-extension to be deprecated).

qwwdfsad:
The good idea is to extract `TestCoroutineContext` and `TestBase` into a separate module, maybe getting rid of inheritance

objcode:
I'm taking a look at putting together a PR for this.

objcode:
See https://github.com/Kotlin/kotlinx.coroutines/pull/890 for a draft API that would do this integration. I expect it'll go through some changes during review.

Files are still in the same place - it's mostly up for API review right now. After https://github.com/Kotlin/kotlinx.coroutines/pull/749 lands if this API is promising it should migrate to `kotlinx-coroutines-test`.

Thanks,
Sean

qwwdfsad:
Fixed in 1.2.1

Issue #242, opened 2018-02-14
EspressoIdlingResource integration


ZakTaccardi:
Would be nice to have a wrapper to easily provide support for [IdlingResource](https://developer.android.com/reference/android/support/test/espresso/IdlingResource.html).

Some examples
- RxJava: [RxIdler](https://github.com/square/RxIdler/)
- [OkHttpIdlingResource](https://github.com/JakeWharton/okhttp-idling-resource)

**EDIT**: I'm currently injecting my `CoroutineDispatchers` like so:

```
open class AppDispatchers(
    val ui: CoroutineDispatcher = UI,
    val background: CoroutineDispatcher = backgroundDefault,
    val network: CoroutineDispatcher = networkDefault
)
```

For espresso testing, which monitors the async task pool and UI thread for idle conditions, I'm injecting the following:

```
class EspressoDispatchers : AppDispatchers(
    ui = UI,
    background = AsyncTask.THREAD_POOL_EXECUTOR.asCoroutineDispatcher(),
    network = AsyncTask.THREAD_POOL_EXECUTOR.asCoroutineDispatcher()
)
```

Here's the problem I'm experiencing, which happens about 1% of the time on our automated tests.

1. Network call completes.
    - monitored by [okhttp-idling-resource](https://github.com/JakeWharton/okhttp-idling-resource)
2. a repository level `ConflatedBroadcastChannel` is updated with the information from the network call.
    - done by the `background` dispatcher
3. Espresso thinks app is now idle, and throws exception because the app isn't idle yet (see number 4)
4. a `ConflatedBroadcastChannel` in the `ViewModel` (which has been observing the repository level  `ConflatedBroadcastChannel` the whole time) is updated
    - done by the `background` dispatcher
    - this happens after Espresso thinks the app is idle (espresso is wrong)

jcornaz:
Or maybe you meant a way to create an IdlingResource from a Job ?

```kotlin
fun Job.asIdlingResource() = object : IdlingResource {
  override fun getName() = "Coroutine job ${this@asIdlingResource}"

  override fun isIdleNow() = isCompleted

  override fun registerIdleTransitionCallback(callback: IdlingResource.ResourceCallback) {
    launch { join(); callback.onTransitionToIdle() }
  }
}
```

ZakTaccardi:
one problem is that the application can be idle when a coroutine is still running (you're observing a receive channel)

ZakTaccardi:
[...]

PR #890, opened 2018-12-14
Update TestCoroutineContext to support structured concurrency.

objcode:
# Updates `TestCoroutineContext` to support structured concurrency.

Issue: https://github.com/Kotlin/kotlinx.coroutines/issues/541

Changes:
- Slit up `TestCoroutineContext` into `TestCoroutineDispatcher`, `TestCoroutineExceptionHandler` and `TestCoroutineScope`
- Added `DelayController` interface for testing libraries to expose Dispatcher control
- Added `ExceptionCaptor` inerface for testing libraries to expose uncaught exceptions
- Added builder for testing coroutines `runBlockingTest` 
- Removed old builder `withTestContext`

### runBlockingTest
`runBlockingTest` is an immediate executor, delays will always auto-progress from either `dispatch` or `scheduleResumeAfterDelay`. Note that it provides a suspend lambda.

`fun runBlockingTest(context: CoroutineContext? = null, testBody: suspend CoroutineScope.() -> Unit)`

```
@Test
fun blockingTest() = runBlockingTest {
    delay(1_000)
    assertTrue(true)
}
```

### TestCoroutineDispatcher / TestCoroutineScope

When used in conjunction with https://github.com/Kotlin/kotlinx.coroutines/issues/810 most tests will declare a dispatcher as part of the setup method. As a convenience extension functions are provided

```
@Before
fun setup() {
    val dispatcher = TestCoroutineDispatcher()
    Dispatchers.setMain(dispatcher)
    scope = TestCoroutineScope(dispatcher)
}

@Test
fun oneTest() = scope.runBlockingTest {
   // for running an async test with control over execution
}
```

*needs review*: Tests that only have a dispatcher are allowed access to `runBlockingTest` as well without constructing a scope

```
val dispatcher: TestCoroutineDispatcher 

@Test
fun threeTests() = dispatcher.runBlockingTest {
   // for running a test with an immediate executor
}
```
### runBlockingTest with time control

Tests can use runBlockingTest with explicit time control if they use the pauseDispatcher function:

```
@Test
fun foo() {
    runBlockingTest(dispatcher) {
          pauseDispatcher {
              // use time control in runBlockingTest
          }
    }
}
```

Tests can use runBlockingTest with time control if they pass the dispatcher directly:

```
@Test
fun foo() {
    val dispatcher = TestCoroutineDispatcher()
    dispatcher.pauseDispatcher()
    runBlockingTest(dispatcher) {
          // use time control in runBlocking
    }
}
```

## Library friendly interfaces
To help testing libraries expose test control, the time control and exception handling are split into interfaces. This allows a testing library to expose a smaller interface to tests than `TestCoroutineDispatcher` as well as combine the two interfaces onto a control object. In the current impl., `TestCoroutineScope` delegates both interfaces.

Ideally, testing libraries would prefer to expose these smaller interfaces. For example a JUnit4 rule might expose `DelayController` and delegate it to an instance of `TestCoroutineDispatcher`.

```
val scope = TestCoroutineScope()
scope.runCurrent() // delegated through DelayController
```

## (Notes) API changes on this PR

- Removal of `asyncTest` that would start the test with explicit time control. This API was a close duplicate of `runBlockingTest`, and is more naturally handled by the block version of `pauseDispatcher` inside a test.
- Dispatcher starts in eager mode now, but can be explicitly paused and resumed.
- Removed `runBlocking` support. The overlap between `runBlocking` and `runBlockingTest` is large - but there are subtle differences that make `runBlockingTest` better suited for testing. There is no API change for `runBlockingTest` for this. The main advantage `runBlockingTest` will eagerly execute `launch` bodies even if they are never joined.


core/kotlinx-coroutines-core/src/test_/TestCoroutineExceptionHandler.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Rethrow the first uncaught coroutine exception immediately.
+     *
+     * This allows tests to use their preferred exception testing techniques.
+     *
+     * If a test generates uncaught exceptions, it must call this method, or clear [exceptions] prior to calling
+     * [cleanupTestCoroutines].
+     */
+    fun rethrowUncaughtCoroutineException(): Nothing
```
zach-klippenstein: The exception checking mechanisms provided by some common unit testing libraries (kotlin.test's `assertFailWith`, JUnit's `ExpectedException` rule) will already fail tests if the code that is expected to throw does not. I would rather not have this method throw its own exception in those cases, to allow testing libraries to fail consistently as they would in other cases.

core/kotlinx-coroutines-core/src/test_/TestCoroutineScope.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlinx.coroutines.*
+import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * A scope which provides detailed control over the execution of coroutines for tests.
+ *
+ * @param context an optional context that must provide delegates [ExceptionCaptor] and [DelayController]
+ */
+class TestCoroutineScope(
```
zach-klippenstein: Since it's a recommended practice to have classes implement `CoroutineScope` to provide convenient access to coroutine builders, it would be nice if the same pattern could be used for tests. E.g.

```kotlin
class MyTest : TestCoroutineScope {
  override val context = TestCoroutineDispatcher() + SomeCustomElement()

  @Test fun thingsWork() = asyncTest {
    …
  }
}
```

The way this API is designed, it requires custom scopes to always be expressed as explicit properties. Is this required verbosity intentional?

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,288 @@
...
+     *     // dispatchImmediately is false
+     * }
+     * ```
+     *
+     * Setting this to true will immediately execute any pending tasks and advance the virtual clock-time to the last
+     * pending delay. While true, dispatch will continue to execute immediately, auto-advancing the virtual clock-time.
+     *
+     * Setting it to false will resume lazy execution.
+     */
+    var dispatchImmediately: Boolean
```
zach-klippenstein: If you set `dispatchImmediately` to `true` inside an `asyncTest` block, is that effectively like using `runBlockingTest` but with the stricter exception and completion checking?"

objcode:
Thanks for review @zach-klippenstein!

core/kotlinx-coroutines-core/src/test_/TestBuilders.kt
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * In addition any unhandled exceptions thrown in coroutines must be rethrown by
+ * [TestCoroutineScope.rethrowUncaughtCoroutineException] or cleared via [TestCoroutineScope.exceptions] inside of
+ * [testBody].
+ *
+ * @throws UncompletedCoroutinesError If the [testBody] does not complete (or cancel) all coroutines that it launches
+ * (including coroutines suspended on await).
+ * @throws UnhandledExceptionsError If an uncaught exception is not handled by [testBody]
+ *
+ * @param dispatcher An optional dispatcher, during [testBody] execution [TestCoroutineDispatcher.dispatchImmediately] will be set to false
```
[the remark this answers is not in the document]
objcode: Thanks, refactor error.

core/kotlinx-coroutines-core/src/test_/TestBuilders.kt
```diff
@@ -0,0 +1,190 @@
...
+        dispatcher.dispatchImmediately = oldDispatch
+    }
+}
+
+/**
+ * Convenience method for calling runBlocking on an existing [TestCoroutineScope].
+ *
+ * [block] will be executed in immediate execution mode, similar to [runBlockingTest].
+ */
+fun <T> TestCoroutineScope.runBlocking(block: suspend CoroutineScope.() -> T): T {
```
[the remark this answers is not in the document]
objcode: Agreed. Renaming and passing through.

core/kotlinx-coroutines-core/src/test_/TestBuilders.kt
```diff
@@ -0,0 +1,190 @@
...
+}
+
+private fun CoroutineContext?.checkArguments(): Pair<CoroutineContext, ContinuationInterceptor> {
+    var safeContext= this ?: TestCoroutineExceptionHandler() + TestCoroutineDispatcher()
+
+    val dispatcher = safeContext[ContinuationInterceptor].run {
+        this?.let {
+            if (this !is DelayController) {
+                throw IllegalArgumentException("Dispatcher must implement DelayController")
+            }
```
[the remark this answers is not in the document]
objcode: Thanks!

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,288 @@
...
+     * @param unit The [TimeUnit] in which the clock-time must be returned.
+     * @return The virtual clock-time
+     */
+    fun currentTime(unit: TimeUnit = TimeUnit.MILLISECONDS): Long
+
+    /**
+     * Moves the Dispatcher's virtual clock forward by a specified amount of time.
+     *
+     * The returned delay-time can be larger than the specified delay-time if the code
+     * under test contains *blocking* Coroutines.
```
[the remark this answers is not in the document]
objcode: Yea, it is the delta from start to the final clock time (which could be progressed via blocking, or even a call to `advanceTimeBy` inside a coroutine).

I'll try to clean this comment up.

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,288 @@
...
+     * @param delayTime The amount of time to move the CoroutineContext's clock forward.
+     * @param unit The [TimeUnit] in which [delayTime] and the return value is expressed.
+     * @return The amount of delay-time that this Dispatcher's clock has been forwarded.
+     */
+    fun advanceTimeBy(delayTime: Long, unit: TimeUnit = TimeUnit.MILLISECONDS): Long
+
+    /**
+     * Moves the current virtual clock forward until the next pending delay.
+     *
+     * Any pending immediate dispatches will be executed, and the clock will be advanced to the next delayed dispatch.
```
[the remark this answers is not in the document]
objcode: Thanks! Was probably too deep in event loops when I wrote this :)

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,288 @@
...
+     * @return the amount of delay-time that this Dispatcher's clock has been forwarded.
+     */
+    fun advanceTimeToNextDelayed(): Long
+
+    /**
+     * Immediately execute all pending tasks and advance the virtual clock-time to the latest delay.
+     *
+     * @return the amount of delay-time that this Dispatcher's clock has been forwarded.
+     */
+    fun runUntilIdle(): Long
```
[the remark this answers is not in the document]
objcode: Good point. I was thinking about making all the clock moving functions advance (which would rename this `advanceUntilIdle` to be consistent).

Alternatively, everything could become run, which would make the following

```
runUntilNextDelayed()
runUntilDelay(delayTime, unit)
```

I don't think `runUntilDelay` is a great name. I'll change this to `advanceUntilIdle` and leave this comment up for more input.

core/kotlinx-coroutines-core/src/test_/TestCoroutineExceptionHandler.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import kotlinx.coroutines.CoroutineExceptionHandler
+import java.util.*
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Access uncaught coroutines exceptions captured during test execution.
+ *
+ * Note, tests executed via [runBlockingTest] or [TestCoroutineScope.runBlocking] will not trigger uncaught exception
+ * handling and should use [Deferred.await] or [Job.getCancellationException] to test exceptions.
```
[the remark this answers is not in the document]
objcode: Yes

core/kotlinx-coroutines-core/src/test_/TestCoroutineExceptionHandler.kt
objcode: Yea - agreed. This API is adding complexity for not much gain. I'll reduce it to the list of exceptions, and rethrow the first uncaught exception in the cleanup method.

objcode: Test code that wants this behavior can just do

`throw exceptions.firstOrNull()`

core/kotlinx-coroutines-core/src/test_/TestCoroutineScope.kt
objcode: I'm not a huge fan of that style in tests, since `TestCoroutineDispatcher` is stateful and it wouldn't take much to leak state between tests if coded like that.

However, if you preferred to code it that way, it would just require a this to ensure you're dispatching to the builder you intend, as well as ensure that dispatcher.cleanupTestCoroutines gets called @After.

From a real-world test perspective, using something like a JUnit4 rule would probably be the way to go here, which would already move out of this inheritance model.

```
@Test fun things() = this.asyncTest { }
```

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
objcode: Yea, I hadn't thought about it that way but it's true. 

Talking out loud: Is there some API simplification that can follow that?

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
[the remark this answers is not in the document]
zach-klippenstein: If the prefixes don't have distinct meanings, I don't have strong opinions on the color of the bikeshed as long as it's consistent. RxJava's `TestScheduler` uses `advance` but also only has 2 such methods.

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
zach-klippenstein: Yea that's what I was thinking too - feels like there's some redundancy here.

If you're making `runBlockingTest` do more of the same checks as `asyncTest`, how different are those two builders really, other than just setting this flag differently? Would it make sense to merge them (maybe add a parameter to conveniently set this flag), or at least make it clear in the docs that the only difference is this flag?

core/kotlinx-coroutines-core/src/test_/TestCoroutineExceptionHandler.kt
zach-klippenstein: Awesome, that also makes it really hard to forget that there might be multiple uncaught exceptions. And it's easy to add this function back later.

core/kotlinx-coroutines-core/src/test_/TestCoroutineScope.kt
zach-klippenstein: Agreed a rule is probably better. Figured this was intended to prevent that pattern, but since that's a pretty common pattern in the official docs it might be helpful to explicitly call it out in the docs (e.g. "Implementing the scope interface is not recommended, for {reasons}, instead use this pattern.").

elizarov:
I've did a first pass through the codes, docs, and example usage. All in all, I very much like this approach to testing. The issues that need to be addressed (in particular order) are:

* Since #810 has landed and we now have a dedicated `kotlinx-corutines-test` module, then this API should move there. Moreover,  all existing APIs in `test` package of `kotlinx-coroutines-core` module shall be deprecated with a message explaining how to migrate to the new ones in a test module.

* Implementations need to be a little big more careful with respect to thread-safety, just in case people start doing weird things in their tests (in particular, `TestCoroutineExceptionHandler.exceptions` shall be a thread-safe list).

* Integration with custom dispatchers needs to worked out. In particular I'm concerned about `Dispatchers.IO` that libraries are supposed to use to wrap around blocking IO calls and `Dispatchers.Default` that might be used to wrap around CPU-consuming background functions in libraries, too. These dispatchers need to be integrated with test dispatchers (we'll need to hook into their implementations) so that, at least, they report that they can report to test dispatcher that they are "busy" working on some coroutines, so that various auto-advancing features would know to wait for them. Moreover, when you do `delay` from inside for `Default` dispatcher it currently goes to internal `DefaultExecutor`, but should be intercepted by test dispatcher instead. You can take a look at how it is internally done in implementation of `withVirtualTimeSource` function. 

* As a part of `kotlinx-coroutines-test` module we can make a global default exception hander customizable in a similar way that we make `Dispatchers.Main` customizable. That will allow to catch all uncaught exception that happen during the test (even if they happen inside `GlobalScope.launch { ... }` block).

* Last, but not the least, naming. While I like `runBlockingTest` name, I don't think `asyncTest` name clearly reflects how it is different from `runBlockingTest`. I would suggest to think about keeping just one top-level method `runBlockingTest` and make immediate/non-immediate behavior customizable in some other way. Immediate execution seems to be a good (least-surprising) default behavior to me.

objcode:
Thanks for the review - I'm going to get it migrated to the new package then take a look at the other points once it's back on master.

objcode:
Ahoy!

Got it rebased to master (ended up doing a bit of manual git history since the merge conflict commit wasn't particularly useful).

> Since #810 has landed and we now have a dedicated `kotlinx-corutines-test` module, then this API 
> should move there. Moreover,  all existing APIs in `test` package of `kotlinx-coroutines-core` module 
> shall be deprecated with a message explaining how to migrate to the new ones in a test module.

Moved this over. Taking a look at the current master, subclassing `EventLoop` (and promoting it to a public API) seemed unnecessary. After a bit of consideration I made `TestCoroutineDispatcher` exposes `runBlockingTest` and not play with `runBlocking`. 

The `runBlocking` integration was always somewhat awkward, since you had to remember to pass through `coroutineContext` every time, which was  easy to forget (enough that I had multiple test cases testing the behavior when forgotten).

> Implementations need to be a little big more careful with respect to thread-safety, just in case people 
> start doing weird things in their tests (in particular, `TestCoroutineExceptionHandler.exceptions` shall be 
> a thread-safe list).

Done. As a result it's copy on read.

> Integration with custom dispatchers needs to worked out. In particular I'm concerned about 
> `Dispatchers.IO` that libraries are supposed to use to wrap around blocking IO calls and 
> `Dispatchers.Default` that might be used to wrap around CPU-consuming background functions in 
> libraries, too. These dispatchers need to be integrated with test dispatchers (we'll need to hook into 
> their implementations) so that, at least, they report that they can report to test dispatcher that they are 
> "busy" working on some coroutines, so that various auto-advancing features would know to wait for 
> them. Moreover, when you do `delay` from inside for `Default` dispatcher it currently goes to internal 
> `DefaultExecutor`, but should be intercepted by test dispatcher instead. You can take a look at how it is 
> internally done in implementation of `withVirtualTimeSource` function.

In both of those cases, I was imagining test code doing one of two things:

1. Injecting a test dispatcher that does time control in place of Dispatchers.IO, etc. It'd be worth exploring if it's useful to swap out the global Dispatchers from test code.

In this case everything works as expected, but multi-threading is removed. From a test design standpoint, the lack of multi-threading improves test reliability so I don't think it would be advantageous to support it from the TestCoroutineDispatcher.

2. Not injecting or mocking a library fn that uses `withContext`, which behaves normally as a suspend function in the test dispatcher.

In this case, the a delay in a library function will lead to a test delay. However, it's not clear that time can be auto-progressed in a discover-able way in this context.

That said, some work will need to integrate coroutines with Espresso (e.g. https://github.com/Kotlin/kotlinx.coroutines/issues/242). I started chatting about that but haven't figured out exactly what it will mean. The trivial implementation (let espresso know whenever the dispatcher is executing something) will capture CPU bound tasks, but won't be busy when all coroutines are suspended waiting on input. The next implementation I'm considering is "any active jobs" = busy, but that won't handle actors.

I'll give it some more thought and think about what can fit here.

> As a part of `kotlinx-coroutines-test` module we can make a global default exception hander 
> customizable in a similar way that we make `Dispatchers.Main` customizable. That will allow to catch 
> all uncaught exception that happen during the test (even if they happen inside `GlobalScope.launch { ... 
> }` block).

Good idea, will take a look at what that would mean. I think it fits into this PR.

> Last, but not the least, naming. While I like `runBlockingTest` name, I don't think `asyncTest` name 
> clearly reflects how it is different from `runBlockingTest`. I would suggest to think about keeping just 
> one top-level method `runBlockingTest` and make immediate/non-immediate behavior customizable in 
> some other way. Immediate execution seems to be a good (least-surprising) default behavior to me.

Agreed. I've completely reworked the API of the dispatcher to expose auto-advance by default and allow it to be paused for advanced use cases. `asyncTest` is pretty easy to build on top of that if a codebase prefers that style, but `runBlockingTest` consistently delivers the least surprising behavior ("run this test fast, don't surprise me with event loop internals").

I've removed `asyncTest` from this PR.

Thanks.

I'll ping again for a review after I take a look into the remaining issues discussed above.

core/kotlinx-coroutines-core/src/test_/TestCoroutineDispatcher.kt
objcode: Ended up resolving this in favor of `runBlockingTest` and more explicit naming here.

elizarov:
Thanks.

I totally agree that "single-threaded tests" is the great goal to have. That is how we write most of our library tests, too. It makes functional testing & debugging of various behaviors so much easier.

Let me clarify my concerns about interaction between `withContext(Dispatchers.IO) { ... }` and tests. The concern is composability in highly-modular code-bases. 

Suppose that I have a separate module in my project that implements some useful function. For example, it performs some long-running CPU computation in its own thread pool or using some built-in dispatcher (like `Dispatchers.Default`) and exposes just the suspending function to the outside world. Now, if this function is somehow used, even indirectly (somewhere deep in a call-stack), in a test, then it foils "time-auto-advance" feature of the test dispatcher. 

We can require, as a coding style practice, that such dispatchers are always get injected (so they can be replaced by a test dispatcher in tests), but, IMHO, that places too much burden on all the libraries. I would rather find some global, test-only mechanism that automatically replaces all such dispatchers with a test-dispatcher (or makes test dispatcher's time-advance feature aware of alternative dispatchers), so that we don't have to manually write this dispatcher-injecting code in our libraries just for the sake of being able to use them in tests.

elizarov:
As for integration with EspressoIdlingResource (#242) I've extracted a separate issues for replacing `Dispatchers.IO` and `Dispatchers.Default` (#982) should help with such integrations. It could also be used to replace them with test dispatcher when needed.

objcode:
That all makes sense to me. I've updated the code based on the remaining comments.

### Current API
I've added appropriate experimental and deprecation annotations throughout. One thing to get input on - there is an existing API for TestCoroutineContext that's being replaced here, does it make sense to add it on `DelayController` as a deprecated-on-addition API to enable `ReplaceWith` refactors?

The current API exposes `runBlockingTest` as the main builder. Advanced users can use `TestCoroutineScope`, `TestCoroutineExceptionHandler`, and `TestCoroutineDispatcher` to execute tests.

### Espresso
I did not put espresso in this PR, but looking at @yigit's code in https://github.com/Kotlin/kotlinx.coroutines/issues/242 it fits cleanly in `TestCoroutineDispatcher`. I'd recommend reviewing this PR with the assumption that a callback-based integration will follow shortly. 

### Dispatchers.IO/Default

Agreed completely, a service locator should be added for both of those otherwise it will be nearly impossible to control this for libraries you depend upon. I'm seeing the pattern of thread confinement with `withContext(Dispatchers.IO)` being used extensively on Android, and it's a pattern we're leaning into. There should be no dependencies from this code to the service locators.

### Exception Handling outside of the testing scoep
After playing around with it I decided /not/ to include a global exception handler hook in this PR. It would add a public API for coroutines (perhaps `CoroutineScope.defaultUncaughtExceptionHandler = ` that would be useful in contexts outside of tests.

Adding a global default exception handler is probably a good idea (especially considering the different implementations for uncaught exceptions of the [bytecode](https://github.com/Kotlin/kotlinx.coroutines/blob/master/core/kotlinx-coroutines-core/src/CoroutineExceptionHandlerImpl.kt#L35) and [js](https://github.com/Kotlin/kotlinx.coroutines/blob/master/js/kotlinx-coroutines-core-js/src/CoroutineExceptionHandlerImpl.kt) versions). There's two reasons for an exception handler like this:

1. Metrics tracking (crash reports etc)
2. Test validation (this is less of a clear choice)

Since this isn't really a test-first API it makes sense to consider it's addition for metrics / crash reporting, especially in JS where it looks like there's no way to intercept the exceptions another way (in bytecode the thread uncaught exception handler can be used).

objcode:
When you have a second @elizarov  can you review it and I'll loop back and address any comments you have!

Thanks,
Sean

core/kotlinx-coroutines-test/src/internal/ThreadSafeHeap.kt
```diff
@@ -0,0 +1,153 @@
+/*
```
objcode: @elizarov This file is a direct copy of the one in core. However, I'm not sure the best way to avoid code duplication here. Happy to update the import however makes sense.

core/kotlinx-coroutines-test/src/TestBuilders.kt
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Convenience method for calling [runBlockingTest] on an existing [TestCoroutineDispatcher].
+ */
+@ExperimentalCoroutinesApi
+fun TestCoroutineDispatcher.runBlockingTest(block: suspend TestCoroutineScope.() -> Unit) {
+    runBlockingTest(this, block)
+}
+
+private fun CoroutineContext?.checkArguments(): Pair<CoroutineContext, ContinuationInterceptor> {
+    var safeContext= this ?: TestCoroutineExceptionHandler() + TestCoroutineDispatcher()
```
yigit: nit, space after safeContext

core/kotlinx-coroutines-test/src/TestBuilders.kt
```diff
@@ -0,0 +1,118 @@
...
+    safeContext = safeContext + dispatcher + exceptionHandler + job
+    return Pair(safeContext, dispatcher)
+}
+
+/**
+ * This method is deprecated.
+ *
+ * @see [runBlocking]
+ */
+@Deprecated("This API has been deprecated to integrate with Structured Concurrency.",
```
yigit: why are we adding a deprecated API ?

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+    fun currentTime(unit: TimeUnit = TimeUnit.MILLISECONDS): Long
+
+    /**
+     * Moves the Dispatcher's virtual clock forward by a specified amount of time.
+     *
+     * The amount the clock is progressed may be larger than the requested delayTime if the code under test uses
+     * blocking coroutines.
+     *
+     * @param delayTime The amount of time to move the CoroutineContext's clock forward.
+     * @param unit The [TimeUnit] in which [delayTime] and the return value is expressed.
```
yigit: There are other coroutine methods (like `withTimeout`, `delay`) that just receive `timeMillis`. Even though I prefer an API that lets me pass `TimeUnit`, it might be better to stay consistent with existing APIs.

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Moves the current virtual clock forward just far enough so the next delay will return.
+     *
+     * @return the amount of delay-time that this Dispatcher's clock has been forwarded.
+     */
+    @ExperimentalCoroutinesApi
+    fun advanceTimeToNextDelayed(): Long
+
+    /**
+     * Immediately execute all pending tasks and advance the virtual clock-time to the last delay.
```
yigit: what is the use case to use `advanceTimeToNextDelay` when `advanceUntilIdle` already exists?
Unless there is a use case in mind, it might make sense to keep the API simpler and not provide `advanceTimeToNextDelay` ?

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Run any tasks that are pending at or before the current virtual clock-time.
+     *
+     * Calling this function will never advance the clock.
+     */
+    @ExperimentalCoroutinesApi
+    fun runCurrent()
+
+    /**
+     * Call after a test case completes.
```
yigit: Called ?

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+    @Throws(UncompletedCoroutinesError::class)
+    fun cleanupTestCoroutines()
+
+    /**
+     * Run a block of code in a paused dispatcher.
+     *
+     * By pausing the dispatcher any new coroutines will not execute immediately. After block executes, the dispatcher
+     * will resume auto-advancing.
+     *
+     * This is useful when testing functions that that start a coroutine. By pausing the dispatcher assertions or
```
yigit: nit "that that"

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+     * setup may be done between the time the coroutine is created and started.
+     */
+    @ExperimentalCoroutinesApi
+    suspend fun pauseDispatcher(block: suspend () -> Unit)
+
+    /**
+     * Pause the dispatcher.
+     *
+     * When paused the dispatcher will not execute any coroutines automatically, and you must call [runCurrent], or one
```
yigit: nit When paused,

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+     * of [advanceTimeBy], [advanceTimeToNextDelayed], or [advanceUntilIdle] to execute coroutines.
```
yigit: nit "or of" ?

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+     */
+    @ExperimentalCoroutinesApi
+    fun pauseDispatcher()
+
+    /**
+     * Resume the dispatcher from a paused state.
+     *
+     * Resumed dispatchers will automatically progress through all coroutines scheduled at the current time. To advance
+     * time and execute coroutines scheduled in the future use one of [advanceTimeBy], [advanceTimeToNextDelayed],
```
yigit: nit "... in the future, use"

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+     * or [advanceUntilIdle].
+     */
+    @ExperimentalCoroutinesApi
+    fun resumeDispatcher()
+}
+
+/**
+ * Thrown when a test has completed by there are tasks that are not completed or cancelled.
```
yigit: nit. by -> but

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+ */
+@ExperimentalCoroutinesApi
+class UncompletedCoroutinesError(message: String, cause: Throwable? = null): AssertionError(message, cause)
+
+/**
+ * [CoroutineDispatcher] that can be used in tests for both immediate and lazy execution of coroutines.
+ *
+ * By default, [TestCoroutineDispatcher] will be immediate. That means any tasks scheduled to be run without delay will
+ * be immediately executed. If they were scheduled with a delay, the virtual clock-time must be advanced via one of the
```
yigit: this behavior is a bit worrying.
Specifically, an asyc being dispatched immediately significantly alters my coroutine's execution order. Of course, there are no guaratees but  consider a buggy code like this:

```
val atomic = AtomicBoolean(false)
async {
    if (fetchData() == true) {
        atomic.set(true)
    }
}
// do something else
if (atomic.set()) {
    // do something.
}
``` 
normally, the code above would flake since `async` may or may not be scheduled immediately. By altering that behavior and making it consistent; this might hide bugs in applications. If we need to pick a consistent behavior, i would prefer asyncs and launches to be deferred as much as possible (put them end of the execution queue). 
That is not perfect either but i think it is better behavior since it will make it easier to catch any missing `await` calls for side effects.

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+            override fun dispose() {
+                queue.remove(node)
+            }
+        }
+    }
+
+    override fun toString(): String = "TestCoroutineDispatcher[time=${time}ns]"
+
+    private fun post(block: Runnable) =
+            queue.addLast(TimedRunnable(block, counter++))
```
yigit: looks like `post` might be called in any thread (via dispatch). should the `counter` be atomic ? alternatively, all posts could be enqueued into a separate thread?

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,296 @@
...
+
+    private fun postDelayed(block: Runnable, delayTime: Long) =
+            TimedRunnable(block, counter++, time + TimeUnit.MILLISECONDS.toNanos(delayTime))
```
yigit: ditto for counter.

core/kotlinx-coroutines-test/src/TestCoroutineExceptionHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Access uncaught coroutine exceptions captured during test execution.
+ */
+@ExperimentalCoroutinesApi
+interface UncaughtExceptionCaptor {
+    /**
+     * List of uncaught coroutine exceptions.
+     *
+     * The returned list will be a copy of the currently caught exceptions.
+     *
+     * During [cleanupTestCoroutines] the first element of this list will be rethrown if it is not empty.
```
yigit: might be nice to log all remaining exceptions.

core/kotlinx-coroutines-test/src/TestCoroutineExceptionHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    override fun handleException(context: CoroutineContext, exception: Throwable) {
+        synchronized(_exceptions) {
+            _exceptions += exception
+        }
+    }
+
+    override val key = CoroutineExceptionHandler
+
+    private val _exceptions = mutableListOf<Throwable>()
```
yigit: i'm not familiar w/ kotlin src code but `_` seems unusual :man_shrugging:

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
objcode: This is one I've gone back and forth on - in this specific example it would be prudent for a developer to write the test like this

```
- test of delayed async call
runBlockingTest {
    pauseDispatcher {
         subject()
         runCurrent()
    }
}

- test of early async call
runBlockingTest {
    subject()
}
```

I modeled the current behavior off the Robolectric scheduler which typically works for most code that doesn't introduce either race conditions or inappropriate side-effects at launch. For code that does, `pauseDispatcher` provides the ability to engage in explicit ordering control.

Eager execution in tests provides a benefit over the execution for tests for this typical code:

```
fun subject() { // don't return job
    scope.launch { // stuff }
}

@Test
fun testSubject() = scope.runBlockingTest {
    subject()
}
```

In execution in the `runBlocking` style, `launch` body won't be entered until the `job.join()`, which isn't available. With this dispatcher, the test could call `runCurrent()` after `subject()` which is similar to the Rx implementation.

As another detail: All of this only applies to launched coroutines. Regular suspend functions will automatically progress through both dispatch and delay (calling another suspend function or a delay function) in `runBlockingTest` (since they don't start a coroutine there is no way to return control back to the test case without inventing new semantics).

Again, I have gone back and forth on this a few times and would welcome other ideas.  One way to go would be to invert the behavior, and expose a way to enter auto-progressing dispatcher state (maybe `resumeDispatcher` could take a block?)

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
objcode: Agreed. Will pull this since `advanceTimeBy` will allow the developer to accomplish this with more precision.

core/kotlinx-coroutines-test/src/TestBuilders.kt
objcode: This is here to aid in automatic refactoring to the new library - the suggested `ReplaceWith` for `TestCoroutineContext` creates code that doesn't have an obvious refactor without knowing the internals of this library.

Not sure if this is a good practice though. Put the code up for more opinions.

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
objcode: Agreed, I am not sure I see a use case for this since delays will always be expressed in m. @elizarov is there a internal code path I don't know about that would trigger a non-ms delay?

ashdavies: I've found this standard a little unexpected since Java API's will usually expect a `TimeUnit`, but as @yigit rightly says most Coroutine API's expect only a `timeMillis` param, so I'd opt for consistency also

ashdavies:
I really love this PR as I've found it quite difficult to consistently use `TestCoroutineContext` with the current hierarchical standay with Coroutines, thank you so much @objcode for proposing it! If there's anything I can do to help 🙂

core/kotlinx-coroutines-test/src/TestCoroutineExceptionHandler.kt
ashdavies: It seems to be a common behaviour when using a private backing field to represent a narrower public type, I'm to understand it's also the standard practice in some other languages, but sadly I can't provide a citation

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
objcode: Rewrote

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
objcode: Migrated it to use Atomic, which will make the bulk of the problems go away. Need to make sure block.run in dispatch doesn't accidentally run things on weird threads too.

core/kotlinx-coroutines-test/src/TestCoroutineExceptionHandler.kt
objcode: This is a private backing field  for a public property, _name is the pattern for that afaik

objcode:
Thanks for taking a look @ashdavies ! Definitely looking for more input about the public API so please feel free to put comments on it!

Cheers,
Sean

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,294 @@
...
+            queue.addLast(TimedRunnable(block, counter.getAndIncrement()))
+
+    private fun postDelayed(block: Runnable, delayTime: Long) =
+            TimedRunnable(block, counter.getAndIncrement(), time.get() + delayTime)
+                    .also {
+                        queue.addLast(it)
+                    }
+
+
+    private fun doActionsUntil(targetTime: Long) {
```
yigit: can we make this return boolean (or int, # of executed coroutines blocks) OR provide a function to check if idle?

I'm using the old one in another test where i have multiple TestCoroutineContexts. I want to have a function that progresses all of them (and they enqueue messages into eachother) and i cannot write a custom function without reflecting into the queue.

yigit: to be clear, the version I need is expected to ignore delayed tasks if time is not ready.
Maybe i'm  missing a function that already exists :/.
Not proud of it but here is what i'm doing there:
```
private fun triggerAllActions() {
        do {
            mainContext.triggerActions()
            testContext.triggerActions()
            val allIdle = listOf(mainContext, testContext).all {
                it.isIdle()
            }
        } while (!allIdle)
    }

    private fun TestCoroutineContext.isIdle() : Boolean {
        val queueField = this::class.java
            .getDeclaredField("queue")
        queueField.isAccessible = true
        val queue = queueField.get(this)
        val peekMethod = queue::class.java
            .getDeclaredMethod("peek")
        val nextTask = peekMethod.invoke(queue) ?: return true
        val timeField = nextTask::class.java.getDeclaredField("time")
        timeField.isAccessible = true
        val time = timeField.getLong(nextTask)
        return time > now()
    }
```

objcode:
Write a test around MAX_INT

streetsofboston:
I Love to follow this. Thanks!

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,294 @@
...
+            }
+            current.run()
+        }
+    }
+
+    override fun currentTime() = time.get()
+
+    override fun advanceTimeBy(delayTimeMillis: Long): Long {
+        val oldTime = time.get()
+        advanceUntilTime(oldTime + delayTimeMillis)
```
streetsofboston: Be sure to clamp the sum `oldTime + delayTimeMillis` against Long.MAX_VALUE so that this sum would not accidentally become a negative value.

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -0,0 +1,294 @@
...
+
+    override fun toString(): String {
+        return "TestCoroutineDispatcher[currentTime=${time}ms, queued=${queue.size}]"
+    }
+
+    private fun post(block: Runnable) =
+            queue.addLast(TimedRunnable(block, counter.getAndIncrement()))
+
+    private fun postDelayed(block: Runnable, delayTime: Long) =
+            TimedRunnable(block, counter.getAndIncrement(), time.get() + delayTime)
```
streetsofboston: When writing the old `TestCoroutineContext`, I forgot to add tests for `Long` overflows in the dispatcher.

If the code-under-test tries to simulate an always-suspending coroutine by calling `delay(Long.MAX_VALUE)`, the TimedRunnable would be scheduled to run immediately, since `time.get() + delayTime` would result in a negative value. 

Be sure to clamp the sum `time.get() + delayTime` against Long.MAX_VALUE so that this sum would not accidentally become a negative value.

core/kotlinx-coroutines-test/src/internal/ThreadSafeHeap.kt
elizarov: I'm changing this to reuse from the core from internal module.

core/kotlinx-coroutines-test/src/TestBuilders.kt
elizarov: This is good.

core/kotlinx-coroutines-test/src/TestCoroutineExceptionHandler.kt
elizarov: Yes. Underscore is also used for atomics in our libraries code to represent "backing atomic" for a atomic value.

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
elizarov: I'm taking care of it

core/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
elizarov: I took care of it.

elizarov:
Here's what I've done to expedite release of this PR:
1. Merged this code with the most recent changes (most notably move to the new `kotlin-multiplatform` plugin and the corresponding changes to the source code organization)
2. Pushed the resulting code to [`pr/980`](https://github.com/Kotlin/kotlinx.coroutines/commits/pr/890) branch to test on our CI infrastructure. 
2. Corrected some style issues here and there, added some todos for the future. See this commit: https://github.com/Kotlin/kotlinx.coroutines/commit/d76456f1b6a92868e9f2f01309167003c94c66ca

There is still an unresolved issue around semantics of immediate `launch` and `doActionsUntil`. See discussions between @objcode and @yigit above. I've also added some minor todos in the code for future discussion. I'd suggest that we do not wait to resolve them, but merge and release this PR asap.

All new API is marked with `@ExperimentalCoroutinesApi` anyway, so we can discuss all the remaining issues separately and make adjustments before this API is stabilized. Please, submit separate issues to github with specific concerns. 

@objcode Please, take a look at my adjustments. We tentatively plan to merge `pr/980` branch, close this PR, and release it in version `1.2.1` next week.

elizarov:
By the way, pay attention to "expect" style sequence testing that `TestBase` infrastructure provides. Take a look at this test I've added in `pr/980` branch: https://github.com/Kotlin/kotlinx.coroutines/blob/d76456f1b6a92868e9f2f01309167003c94c66ca/kotlinx-coroutines-test/test/TestRunBlockingOrderTest.kt
It seems quite useful for all kinds of asynchronous testing when you want to ensure that events did happen in certain order. However, I'm not sure if this style of testing makes sense for general application-writing public, beyond testing of the core libraries.

objcode:
Thanks @elizarov taking a look at the updates and notes.

objcode:
Took a look at all of the changes, LGTM.

This will need a readme updates for the test package to cover the new API surface. I'll draft that in an isolated commit that can cherry-picked or merged after 890.

Minor nit:

Typo at https://github.com/Kotlin/kotlinx.coroutines/commit/d76456f1b6a92868e9f2f01309167003c94c66ca#diff-e3c98820170c8b47a0aad1428e940c2cL111

And inquiry, is the `_name` style the correct naming convention for all private val's in classes?

elizarov:
> And inquiry, is the `_name` style the correct naming convention for all private val's in classes?

No. Only for backing properties: https://kotlinlang.org/docs/reference/coding-conventions.html#property-names
We kind of use this backing-property convention to name "atomic" fields that are "backing" the real properties, sometime even though there is no real property there, simply because otherwise a use-site of `count.value` looks fancy ("why I would need to ask a value of count, when it is a value?").

objcode:
Put together a README.md for this PR in a cherry-pickable commit.

https://github.com/Kotlin/kotlinx.coroutines/commit/3c484cec51db6380e973cff9af9b0fdbb197f8f0

All of the url's are pending because I wasn't sure I figured out the url format for docs.

elizarov:
> Put together a README.md for this PR in a cherry-pickable commit.
> [objcode#1](https://github.com/objcode/kotlinx.coroutines/pull/1)

Great. I've cherry-picked it into `pr/980`
 
> All of the url's are pending because I wasn't sure I figured out the url format for docs.

URLs links to docs are automatically fixed by running `./gradlew knit`. I did it in `pr/980` branch.

elizarov:
🎉 Merged into `develop` branch. Closing this PR. Feel free to open separate issue for future improvements.

PR #896, opened 2018-12-16
Ignore exception on cancel/fail race in CancellableContinuation

### LINKED ISSUE #830: Introduce API to avoid cancel/fail races with CancellableContinuation
Consider this attempt to integrate coroutines with cancellable callback-based API using `suspendCancellableCoroutine`. This example is for Retrofit, but the same issue could appear with any cancellable API:

```
suspend fun <T> Call<T>.await(): T = 
    suspendCancellableCoroutine { cont ->
        enqueue(object : Callback<T> { // install callback
            override fun onResponse(call: Call<T>, response: Response<T>) {
                 // resume normally -- omitted for clarity
            }
    
            override fun onFailure(call: Call<T>, t: Throwable) {
                if (cont.isCancelled) return     // LINE (1) 
                cont.resumeWithException(t)      // LINE (2)
            }
        })
        // cancel Call when continuation is cancelled
        cont.invokeOnCancellation {
            cancel()
        }
    }
```

Resuming a `CancellableContinuation` with exception is documented to behave in the following way (see https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/index.html):

> Invocation of resume or resumeWithException in resumed state produces IllegalStateException. Invocation of resume in cancelled state is ignored (it is a trivial race between resume from the continuation owner and outer job cancellation and cancellation wins). **Invocation of resumeWithException in cancelled state triggers exception handling of passed exception.**

So, this code attempts to avoid calling `resumingWithException` on an already cancelled continuation by doing a check in LINE (1). This creates a "check & act" race in between LINE (1) and LINE (2). In particular, if continuation gets cancelled from a different thread when this code executes between LINE (1) and LINE (2), then `resumeWithException` is invoked on cancelled continuation and triggers handling of "uncaught" exception.

There is no public (non-experimental, non-internal) API at the moment to work around this race.

## Workarounds

**W1. Use internal API on `CancellableContinuation`**:

Replace LINE (1) and LINE(2) with the following code:

```
            @UseExperimental(InternalCoroutinesApi::class)
            cont.tryResumeWithException(t)?.let { cont.completeResume(it) }
```

This workaround ignores exceptions when continuation is cancelled and this is the disadvantage of this workaround at the same time. If a true exception (failure) happens concurrently with cancellation, then it is going to be ignored instead of being handled. Arguably this problem is much less severe, but still.

**W2. Use internal API on `Job`**:

1. Replace `suspendCancellableCoroutine` with `suspendCoroutine`.
2. Replace `cont.invokeOnCancellation { ... }` with the following code:

```
        @UseExperimental(InternalCoroutinesApi::class)
        cont.context[Job]?.invokeOnCompletion(onCancelling = true) {
            cancel()
        }
```

3. Replace LINE (1) with detection of "Cancelled" state on a `Call` (to see if the call had failed or was cancelled) and resume continuation with `CancellationException` in case of cancellation:

```
            if (call.isCanceled) {
                cont.resumeWithException(CancellationException("Cancelled"))
                return
            }
```

This workaround correctly handles exceptions that occur concurrently with cancellation (the call is either going to be cancelled or fails and we learn what had happened).

## Proposed solutions

There are several possible solutions:

**S1. Ignore exceptions when resuming a cancelled continuation using resumeWithException**. Technically, this is a breaking change, but that is not a major problem here. The problem is that there could be a genuine exception that needs to be handled and ignoring it in `resumeWithException` is as bad as any other instance of ignoring an exception. 

**S2. Introduce new `resumeWithExceptionOrIgnoreOnCancel` method** (name is purely provisional). It is not a breaking change, but it suffers from the same problem of ignoring a genuine exception when it happens concurrently with cancellation. This is not a severe problem, though, and this could still be one of the solutions.

**S3. New API for cancellable callbacks with an intermediate _cancelling_ state**.  The idea is that invocation of `cancel()` should not immediately result in resumption of the suspended coroutine, but shall wait until `resumeWith` (value or exception) is invoked. It gives a chance to see whether the operation was indeed cancelled successfully or failed while we trying to cancel it. It requires writing somewhat more code, though, similar to the code described in the second workaround.
  - This can be a new _mode_ operation for `suspendCancellableCoroutine`. 
  - This can be a new top-level function.

### LINKED ISSUE #892: withTimeout leaks exceptions caught inside it
This test fails:

```
  @Test
  fun `withTimeout leaks caught exceptions`() {
    val start = AtomicLong()
    try {
      runBlocking {
        start.set(System.nanoTime())
        withTimeout(100) {
          while (true) {
            try {
              GlobalScope.future {
                throw FooException()
              }.await()
            } catch (ignored: FooException) {
              println("caught FooException")
            }
          }
        }
      }
      // OK
    } catch (e: FooException) {
      val elapsed = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start.get())
      fail("expected CancellationException, got $e; elapsed time: ${elapsed}ms")
    }
  }

  class FooException : Exception("foo")

```

Output:

```
caught FooException (169 matches)
java.lang.AssertionError: expected CancellationException, got FooException: foo; elapsed time: 112ms
```

I would _never_ expect `withTimeout` to throw `FooException`, since it is caught inside it!

elizarov:
There seems to be no other way to satisfactory fix the problem of a race between cancellation on CancellationContinuation and concurrent failure of the corresponding background process in a way the distinguishes "failure from cancellation attempt" from a "true failure" without placing undue burden on anyone who is implementing `await()` extension function for various future types.

- Added testTimeoutCancellationFailRace that works as a perfect litmus test; being allowed to run for a while it realiable detects various  problems in alternative approaches.
- Optimized CancellableContinuationImpl; merged its code with AbstractContinuation class that is not needed as a separate class and it removed.
- Deprecated and hidden CancellableContinuation.initCancellability(); it is now always invoked after the end of the block that was passed to suspendCancellableCoroutine.
- Deprecated `holdCancellability` parameter to an internal suspendAtomicCancellableCoroutine function.

Fixes #892
Fixes #830

### DIFF
--- a/common/kotlinx-coroutines-core-common/src/CancellableContinuationImpl.kt
+++ b/common/kotlinx-coroutines-core-common/src/CancellableContinuationImpl.kt
@@ -17,10 +17,12 @@ private const val RESUMED = 2
 /**
  * @suppress **This is unstable API and it is subject to change.**
  */
-internal abstract class AbstractContinuation<in T>(
+@PublishedApi
+internal open class CancellableContinuationImpl<in T>(
     public final override val delegate: Continuation<T>,
     resumeMode: Int
-) : DispatchedTask<T>(resumeMode), Continuation<T> {
+) : DispatchedTask<T>(resumeMode), CancellableContinuation<T>, CoroutineStackFrame {
+    public override val context: CoroutineContext = delegate.context
 
     /*
      * Implementation notes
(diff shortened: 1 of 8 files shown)


common/kotlinx-coroutines-core-common/src/CancellableContinuationImpl.kt
```diff
@@ -66,40 +67,63 @@ internal abstract class AbstractContinuation<in T>(
...
 
-    public val isCancelled: Boolean get() = state is CancelledContinuation
+    public override val isCancelled: Boolean get() = state is CancelledContinuation
 
-    internal fun initParentJobInternal(parent: Job?) {
+    public override fun initCancellability() {
+        // This method does nothing. Leftover for binary compatibility with old compiled code
+    }
+
+    private fun installParentCancellationHandler() {
```
qwwdfsad: we should check `if (isCompleted)` as the first action to avoid performance degradation in channels and mutex

elizarov:
All review issues were addressed.

gildor:
Are there any estimates about release date with this fix?

qwwdfsad:
@gildor 1.1.0-alpha today or tomorrow and 1.1.0 on Friday if everything goes well

gildor:
@qwwdfsad Great news, just in time, I just didn't want to release kotlin-coroutines-retrofit with workaround that uses internal API.
Now I will just recommend all users to update coroutines to 1.1.0 and change nothing with my adapter
Thanks!

PR #898, opened 2018-12-17
Do not probabilistically steal blocking tasks as part of anti-starvat…

qwwdfsad:
…ion mechanism: stealing thread might be woken up in `blockingQuiescence` and already have a blocking task in its local queue


core/kotlinx-coroutines-core/src/scheduling/CoroutineScheduler.kt
```diff
@@ -840,9 +840,17 @@ internal class CoroutineScheduler(
...
-        private fun doPark(nanos: Long) {
+        private fun doPark(nanos: Long): Boolean {
+            /*
+             * Here we are trying to park, then check whether there are new blocking tasks
+             * (because submitting thread could have missed this thread in tryUnpark)
+             */
             parkedWorkersStackPush(this)
+            if (!blockingQuiescence()) {
+                return false
+            }
```
elizarov: ^ Code style: Format those "quick returns" without curly braces, especially since `if (!blockingQuiescence()) return` is one-liner in two other places).

PR #928, opened 2019-01-08
Update coroutines-guide-ui.md

rosuH:
[Update]
- typo: what you might want to express here is `JavaScript event dispatch thread `


qwwdfsad:
Thanks!

PR #929, opened 2019-01-10
Make SupervisorCoroutine implement ScopeCoroutine

qwwdfsad:


core/kotlinx-coroutines-core/src/Debug.kt
```diff
@@ -42,15 +42,7 @@ public const val DEBUG_PROPERTY_VALUE_ON = "on"
...
-internal val DEBUG = systemProp(DEBUG_PROPERTY_NAME).let { value ->
-    when (value) {
-        DEBUG_PROPERTY_VALUE_AUTO, null -> CoroutineId::class.java.desiredAssertionStatus()
-        DEBUG_PROPERTY_VALUE_ON, "" -> true
-        DEBUG_PROPERTY_VALUE_OFF -> false
-        else -> error("System property '$DEBUG_PROPERTY_NAME' has unrecognized value '$value'")
-    }
-}
-
+internal val DEBUG = true
```
elizarov: How come its always `true` now?

core/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -74,23 +74,30 @@ internal object DebugProbesImpl {
...
             }
         } else {
+            // Append coroutine with its last stacktrace element
             val element = state.lastObservedStackTrace().firstOrNull()
             val contState = state.state
-            builder.append("$str, continuation is $contState at line $element\n")
+            builder.append("$indent$str, continuation is $contState at line $element\n")
+            newIndent = indent + "\t"
         }
+        // Append children with new indent
```
elizarov: The code above is hard to understand to due to tie "print string and increase indent" logic that is duplicated across different if branches. I'd rewrite it somehow. One way (not sure the best one):

```
        val line = if (state == null) { // Append coroutine without stacktrace
            // Do not print scoped coroutines 
            @Suppress("INVISIBLE_REFERENCE")
            if (this is kotlinx.coroutines.internal.ScopeCoroutine<*>) null else "$indent$str\n"
        } else {
            // Append coroutine with its last stacktrace element
            val element = state.lastObservedStackTrace().firstOrNull()
            val contState = state.state
            "$indent$str, continuation is $contState at line $element\n"
        }
        if (line != null) builder.append(line)
        // Append children with new indent
        for (child in children) {
            child.build(map, builder, if (line == null) indent else indent + "\t")
        }
```

core/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -74,23 +74,30 @@ internal object DebugProbesImpl {
...
-        @Suppress("DEPRECATION_ERROR")
         val str = if (this !is JobSupport) toString() else toDebugString()
-        if (state == null) {
+        val newIndent: String
+        if (state == null) { // Append coroutine without stacktrace
+            // Do not print scoped coroutines and do not increase indentation level
             @Suppress("INVISIBLE_REFERENCE")
-            if (this !is kotlinx.coroutines.internal.ScopeCoroutine<*>) { // Do not print scoped coroutines
-                builder.append("$str\n")
+            if (this !is kotlinx.coroutines.internal.ScopeCoroutine<*>) {
```
elizarov: This `if` would look better inverted.

core/kotlinx-coroutines-core/src/Debug.kt
qwwdfsad: Workaround for IDEA bug slim through changeset

core/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
qwwdfsad: Still not so readable.
Extracted method

elizarov: Much better

core/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -76,22 +76,31 @@ internal object DebugProbesImpl {
...
+        }
+    }
+
+    /**
+     * Appends the current job to the resulting string builder if it does not represent scope coroutine.
+     * Returns whether the job was appended to [builder].
+     */
+    @Suppress("INVISIBLE_REFERENCE", "DEPRECATION_ERROR") // JobSupport and ScopeCoroutine
+    private fun Job.appendCoroutine(builder: StringBuilder, state: CoroutineState?, indent: String): Boolean {
+        val str = if (this !is JobSupport) toString() else toDebugString()
```
elizarov: This condition would read better inverted:
```
val str = if (this is JobSupport) toDebugString() else toString() 
```
Or event better, you can introduce `toDebugString` extension for all jobs and avoid `str` variable all together (it is computed by not used for scoped coroutines):
```
fun Job.toDebugString() = if (this is JobSupport) toDebugString() else toString()  
```

core/kotlinx-coroutines-core/src/Debug.kt
elizarov: It is still in the change set!

PR #946, opened 2019-01-18
Update channels.md

BattleShipPark:
Fix typo


qwwdfsad:
It is not a typo, it was made to avoid clock resolution issues

PR #947, opened 2019-01-21
Migration to new multiplatorm plugin

qwwdfsad:
Should be squashed before merging


gradle/compile-common.gradle
```diff
@@ -1,13 +1,15 @@
 /*
- * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-// Platform-specific configuration to compile common modules
+kotlin.sourceSets {
+    commonMain.dependencies {
+        api group: 'org.jetbrains.kotlin', name: 'kotlin-stdlib-common', version: kotlin_version
```
elizarov: It would be more consistent and thus easier to read when version is specified in the same format as on the next line:  `"org.jetbrains.kotlin:kotlin-stdlib-common:$kotlin_version"`

gradle/compile-native-multiplatform.gradle
```diff
@@ -0,0 +1,35 @@
...
+        if (project.ext.ideaActive) {
+            fromPreset(project.ext.ideaPreset, 'native')
+        } else {
+            fromPreset(presets.linuxX64, 'linuxX64')
+            fromPreset(presets.iosArm64, 'iosArm64')
+            fromPreset(presets.iosArm32, 'iosArm32')
+            fromPreset(presets.iosX64, 'iosX64')
+            fromPreset(presets.macosX64, 'macosX64')
+            fromPreset(presets.mingwX64, 'windowsX64')
+            // TODO mingw :
```
elizarov: TODO what?

build.gradle
```diff
@@ -47,131 +54,85 @@ buildscript {
...
+    
+    dependencies {
+        compile project(":$coreModule")
+        // the only way IDEA can resolve test classes
+        testCompile project(":$coreModule").kotlin.targets.jvm.compilations.test.output.allOutputs
+    }
 
-def internal = sourceless + ['benchmarks', 'knit', 'js-stub', 'stdlib-stubs', 'binary-compatibility-validator']
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompile).all {
+        kotlinOptions.freeCompilerArgs += ["-Xuse-experimental=kotlin.Experimental",
```
elizarov: Why it's not using `experimentalAnnotations` list from `experimental.gradle` here?

gradle/maven-central.gradle
```diff
@@ -34,4 +34,12 @@ project.ext.configureMavenCentralMetadata = {
     root.appendNode('description', 'Coroutines support libraries for Kotlin')
     root.appendNode('url', 'https://github.com/Kotlin/kotlinx.coroutines')
     root.children().last() + pomConfig
+
+    def dependenciesNode = asNode().getAt("dependencies")[0]
+    if (dependenciesNode != null) {
+        def dependencyNode = dependenciesNode.appendNode('dependency')
+        dependencyNode.appendNode('groupId', "$group")
+        dependencyNode.appendNode('artifactId', "$project.name")
+        dependencyNode.appendNode('version', "$project.version")
```
elizarov: What's the reason for the above section? What I see in the resulting .pom files:
* `kotlinx-coroutines-core` (JVM) artifact depends on _itself_
* `kotlinx-coroutines-core-js` artifact depends on `kotlinx-coroutines-core` (JVM)... very strange
* `kotlinx-coroutines-core-common` artifact depends on `kotlinx-coroutines-core` (JVM)... even stranger.
Something is not right here.

gradle/publish-bintray.gradle
```diff
@@ -12,72 +12,149 @@ apply plugin: "com.github.johnrengelman.shadow"
...
+    disableMetadataPublication()
 }
 
-def preparePublication(MavenPublication publication) {
-    if (!isNative()) {
-        if (project.name == "kotlinx-coroutines-debug") {
-            project.shadow.component(publication)
-        } else {
-            publication.from components.java
+private void disableMetadataPublication() {
```
elizarov: Something is wrong with the way this code is written. Metadata publishing is (correctly) disabled for `kotlinx-coroutines`, `kotlinx-coroutines-common`, and `kotlinx-coroutines-js`, but there is metadata for all additional (non-multiplatform) gradle subprojects (`kotlinx-coroutines-jdk8`, `kotlinx-coroutines-guava`, etc)

gradle/publish-bintray.gradle
```diff
@@ -12,72 +12,149 @@ apply plugin: "com.github.johnrengelman.shadow"
...
+        from kotlin.sourceSets.commonMain.kotlin
+    } else {
         from sourceSets.main.allSource
     }
+}
 
-    if (project.name == coroutines_core && platform != "common") {
-        // add common source, too
-        from rootProject.file("common/$project.name-common/src/main/kotlin")
+def configureTransitiveDependencies = { Project project, Publication publication ->
```
elizarov: Something is wrong with code. For example, `kotlinx-coroutines-jdk8-1.1.0-SNAPSHOT.pom` file has the following dependencies:
* `kotlin-stdlib` - Ok
* `kotlinx-coroutines-core-native` - Wrong.
* `kotlinx-coroutines-jdk8` - Wrong (but that seems to be a bug in different place).

qwwdfsad: I is a bug in multiplatform + single target plugins, the only solution is to rewrite poms by hand

elizarov:
Another bug (does not seem to critical, might be a pre-existing problem): `kotlinx-coroutines-debug` module has not dependencies whatsoever in its `.pom` file, but it shall depend on core.

qwwdfsad:
Verified, it is a pre-existing problem related to shadow plugin

PR #951, opened 2019-01-23
Version 1.1.1

qwwdfsad:


elizarov:
Release notes and changes log?

Tolriq:
@elizarov @qwwdfsad This is really strange the PR is still opened and not merged but JCenter says that 1.1.1 is already available? Is there a risk that a fake dependency is on JCenter?
What would be the source of truth repository for coroutines?

https://bintray.com/kotlin/kotlinx/kotlinx.coroutines/1.1.1

qwwdfsad:
It is not a fake dependency, the release wasn't yet announced. 
Dependency is here for less than a hour, give it some time :)

Tolriq:
Ok thanks was just wondering why a non merged PR was already available :)
Since Kotlin 1.3.20 was out I ran ben manes version plugin to check other dependencies related and always check the changelog, so was lost to see it's not yet released and released at same time.

PR #955, opened 2019-01-28
Custom ServiceLoader without jar checksum verification

### LINKED ISSUE #878: Slow android Dispatchers.Main init
It seems Android Main dispatcher initialization / discovery can be quite slow currently.

On a Nokia 8 device my application startup time went from 250ms to 500ms (on release mode). I know that 250ms does not sound much but it's pretty much 2 times slower startup time and it does impact the users as they often want the fastest possible startup time for this kind of application. And I still have to see on older devices if the impact is 250ms or linear with the device power.

See attached flamechart (It's from the debug app with full trace profiling so the times are insanely slow but was necessary to find the issue).

coroutine flame chart

Is there a way to bypass that slow factory discovery and statically create my own Main dispatcher on Android to avoid that slowdown?

SokolovaMaria:
Fixes #878


core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,116 @@
...
+import java.util.zip.ZipEntry
+import kotlin.NoSuchElementException
+import java.io.IOException
+import java.io.BufferedReader
+
+private const val PREFIX: String = "META-INF/services/"
+
+public fun <S> loadProviders(service: Class<S>, loader: ClassLoader) = CustomServiceLoader(service, loader)
+
+public class CustomServiceLoader<S> internal constructor (private val service: Class<S>, private val loader: ClassLoader?) : Iterable<S> {
```
qwwdfsad: It shouldn't be iterable or stateful, the only API we need is `CustomServiceLoader.load(service, loader): List<..>`..
Something like 

```
object CustomServiceLoader {
   fun load(...)
}

```
is enough

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,116 @@
...
+import java.util.*
+import java.util.jar.JarFile
+import java.util.zip.ZipEntry
+import kotlin.NoSuchElementException
+import java.io.IOException
+import java.io.BufferedReader
+
+private const val PREFIX: String = "META-INF/services/"
+
+public fun <S> loadProviders(service: Class<S>, loader: ClassLoader) = CustomServiceLoader(service, loader)
```
qwwdfsad: Shouldn't it be internal?

core/kotlinx-coroutines-core/src/internal/MainDispatchers.kt
```diff
@@ -12,9 +12,12 @@ internal object MainDispatcherLoader {
     private fun loadMainDispatcher(): MainCoroutineDispatcher {
         return try {
             val factories = MainDispatcherFactory::class.java.let { clz ->
-                ServiceLoader.load(clz, clz.classLoader).toList()
+                try {
+                    loadProviders(clz, clz.classLoader).toList()
+                } catch (e: ServiceConfigurationError) {
+                    ServiceLoader.load(clz, clz.classLoader).toList()
```
qwwdfsad: This logic is better to integrate into our service loader.
Also, killswitch is missing (e.g. see `Debug#RECOVER_STACKTRACES`)

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,116 @@
...
+
+public class CustomServiceLoader<S> internal constructor (private val service: Class<S>, private val loader: ClassLoader?) : Iterable<S> {
+
+    private fun parse(service: Class<S>, url: URL): Iterator<String> {
+        val names = mutableListOf<String>()
```
qwwdfsad: It can be rewritten with `iterator` builder

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            val string = url.toString()
+            val separatorIndex = string.indexOf('!')
+            val pathToJar = string.substring(0, separatorIndex)
+            val entry = string.substring(separatorIndex + 2, string.length)
+            (JarFile(pathToJar.substring("jar:file:/".length), false) as Closeable).use { file ->
```
qwwdfsad: It looks like ad-hoc checks. Please leave some comments about what is happening here and why it is legal to unconditionally assume prefixes and `!` symbols

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,116 @@
...
+        return object : Iterator<S> {
+            private var urls: Enumeration<URL>? = null
+            private var providersToLoad: Iterator<String>? = null
+            private var nextProviderName: String? = null
+
+            override fun hasNext(): Boolean {
+                nextProviderName?.let { return true }
+                urls = urls ?: try {
+                    val fullServiceName = PREFIX + service.name
+                    if (loader == null) {
```
qwwdfsad: loader is never null

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,116 @@
...
+        var line = r.readLine() ?: return -1
+        val commentPos = line.indexOf('#')
+        if (commentPos >= 0) line = line.substring(0, commentPos)
+        line = line.trim()
+        val n = line.length
+        if (n != 0) {
+            if (line.indexOf(' ') >= 0 || line.indexOf('\t') >= 0)
+                error(service, "$u : $lc : Illegal configuration-file syntax")
+            var cp = line.codePointAt(0)
+            if (!Character.isJavaIdentifierStart(cp))
```
qwwdfsad: Are all these checks really required? Something like `require(line.all {Character. isJavaIdentifierPart(it) })` should be enough.
Also, we can assume we have a single service per file and simplify it even more

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private fun <S> parse(service: Class<S>, url: URL): List<String> {
+        try {
+            val string = url.toString()
+            // the syntax of JAR URL is: jar:<url>!/{entry}
+            val separatorIndex = string.indexOf('!')
+            val pathToJar = string.substring(0, separatorIndex)
+            val entry = string.substring(separatorIndex + 2, string.length)
+            (JarFile(pathToJar.substring("jar:file:/".length), false) as Closeable).use { file ->
```
qwwdfsad: `pathToJar.substring("jar:file:/".length)` works only when relative path from current directory matches absolute path.

It worked in my prototype on Android accidentally, slash should not be trimmed.
Currently, all of our tests (when running from gradle) fail on this line and falls back to the default implementation.

Please make `loadProviders` internal and add tests on that (e.g. in android-unit-test module).
At least one test for same module resource and one for cross-module (e.g. Main dispatcher)

qwwdfsad: Additionally, to avoid this index manipulation, Koltin has `substringBefore` and `substringAfter`

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,111 @@
...
+            val commentPos = line.indexOf('#')
+            if (commentPos >= 0) line = line.substring(0, commentPos)
+            line = line.trim()
+            if (!line.isEmpty()) {
+                try {
+                    require(line.all { it == '.' || Character.isJavaIdentifierPart(it) })
+                } catch (e: IllegalArgumentException) {
+                    error(service, "Illegal configuration-file syntax", e)
+                }
+                names.add(line)
```
qwwdfsad: Parsing can be written in a more compact and idiomatic way:

```
val line = r.readLine() ?: break
val serviceName = line.substringBefore("#").trim().ifEmpty {}
require(serviceName.all {}, ...)
if (serviceName.isNotEmpty() {
    name.add(serviceName)
}
```

core/kotlinx-coroutines-core/src/internal/CustomServiceLoader.kt
```diff
@@ -0,0 +1,98 @@
...
+                BufferedReader(InputStreamReader((file as JarFile).getInputStream(ZipEntry(entry)), "UTF-8")).use { r ->
+                    return parseFile(service, url, r)
+                }
+            }
+        } catch (e: Throwable) {
+            error(service, "Error reading configuration file", e)
+        }
+    }
+
+    private fun parseFile(service: Class<*>, u: URL, r: BufferedReader): String {
```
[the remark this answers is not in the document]
qwwdfsad: This should be tested as well

core/kotlinx-coroutines-core/src/internal/FastServiceLoader.kt
```diff
@@ -0,0 +1,96 @@
...
+        urls.forEach {
+            val providerNames = parse(it)
+            providers.addAll(providerNames.map { getProviderInstance(it, loader, service) })
+        }
+        return providers
+    }
+
+    private fun <S> getProviderInstance(name: String, loader: ClassLoader, service: Class<S>): S {
+        val cl = Class.forName(name, false, loader)
+        require(service.isAssignableFrom(cl))
```
qwwdfsad: Please always provide a message with proper diagnostic in asserts/exceptions.
E.g. `require(service.isAssignableFrom(cl)) { "Expected service of class $service, but had $cl" } `

qwwdfsad: Please always provide a message with proper diagnostic in asserts/exceptions.
E.g. `require(service.isAssignableFrom(cl)) { "Expected service of class $service, but had $cl" } `

core/kotlinx-coroutines-core/src/internal/FastServiceLoader.kt
```diff
@@ -0,0 +1,96 @@
...
+        return service.cast(cl.getDeclaredConstructor().newInstance())
+    }
+
+    private fun parse(url: URL): List<String> {
+        val string = url.toString()
+        when {
+            string.startsWith("file") -> {
```
qwwdfsad: This when requires some explanation

core/kotlinx-coroutines-core/src/internal/FastServiceLoader.kt
```diff
@@ -0,0 +1,96 @@
...
+            else -> throw Exception("Error parsing configuration file URL $url")
+        }
+    }
+
+    private fun parseFile(r: BufferedReader): List<String> {
+        val names = mutableSetOf<String>()
+        while (true) {
+            val line = r.readLine() ?: break
+            val serviceName = line.substringBefore("#").trim()
+            require(serviceName.all { it == '.' || Character.isJavaIdentifierPart(it) })
```
qwwdfsad: ditto for error message

core/kotlinx-coroutines-core/src/internal/FastServiceLoader.kt
```diff
@@ -0,0 +1,96 @@
...
+            string.startsWith("file") -> {
+                val pathToFile = string.substringAfter("file:")
+                BufferedReader(FileReader(pathToFile)).use { r ->
+                    return parseFile(r)
+                }
+            }
+            string.startsWith("jar") -> {
+                val pathToJar = string.substringAfter("jar:file:").substringBefore('!')
+                val entry = string.substringAfter("!/")
+                (JarFile(pathToJar, false) as Closeable).use { file ->
```
qwwdfsad: ` as Closeable` is a workaround for compiler bug? Is it reported?

core/kotlinx-coroutines-core/service-loader-test/test/ServiceLoaderTest.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlinx.coroutines.InternalCoroutinesApi
+import kotlinx.coroutines.internal.FastServiceLoader
+import kotlinx.coroutines.internal.MainDispatcherFactory
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class ServiceLoaderTest {
+
+    @InternalCoroutinesApi
```
qwwdfsad: These annotations (InternalCoroutinesApi) are not necessary

core/kotlinx-coroutines-core/service-loader-test/test/ServiceLoaderTest.kt
```diff
@@ -0,0 +1,32 @@
...
+    @Test
+    fun testLoadingCrossModuleService() {
+        val providers = MainDispatcherFactory::class.java.let { FastServiceLoader.loadProviders(it, it.classLoader) }
+        assertEquals(providers.size, 2)
```
qwwdfsad: It is better to verify actual classes as well

settings.gradle
```diff
@@ -23,6 +23,7 @@ module('binary-compatibility-validator')
 module('common/kotlinx-coroutines-core-common')
 
 module('core/kotlinx-coroutines-core')
+module('core/kotlinx-coroutines-core/service-loader-test')
```
qwwdfsad: Why a separate module is required?
Current tests can be safely moved to kx-coroutines-core.

It is also worth to add a special service to core resources, but write a test that loads this service from `android-unit-tests` module to test cross-module loading.

In that case, you can get rid of `public` + `InternalCoroutinesApi`and additional module

core/kotlinx-coroutines-core/src/internal/MainDispatchers.kt
```diff
@@ -25,7 +24,8 @@ internal object MainDispatcherLoader {
 }
 
 /**
- * If anything goes wrong while trying to create main dispatcher (class not found,
+ * If anything goes wrong while trying to
+ * create main dispatcher (class not found,
```
qwwdfsad: accidental change?

settings.gradle
qwwdfsad: For a single cross-module test you can use `@Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")`.
It is better to suppress the test then exposing something in the API

core/kotlinx-coroutines-core/src/internal/FastServiceLoader.kt
SokolovaMaria: I added this ```when string.startsWith("jar")``` parsing,to avoid failing when service file is declared in resources of the same module. But then providers are loaded twice from this file and from the jar of this module, I think parsing jar urls is enough - I've changed that.

core/kotlinx-coroutines-core/service-loader-test/test/ServiceLoaderTest.kt
SokolovaMaria: How can I verify that?

settings.gradle
SokolovaMaria: I created that separate module to add dependencies to ```kotlinx-coroutines-test (android)``` modules. Now if I move those tests to ```kotlinx-coroutines-core``` adding dependencies to these modules then FailFastOnStart tests would try to load providers from those modules and fail.

core/kotlinx-coroutines-core/src/internal/FastServiceLoader.kt
SokolovaMaria: Yes, otherwie it fails with: ```Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
@InlineOnly public inline fun <T : Closeable?, R> ???.use(block: (???) -> List<String>): List<String> defined in kotlin.io```

But I haven't found the minimal example for report.
This one works fine:
```
JarFile("src/test.zip", false).use { file ->
        BufferedReader(InputStreamReader((file).getInputStream(ZipEntry("test")),"UTF-8")).use { r ->
            println(r.readLine())
        }
}
```

settings.gradle
qwwdfsad: `FailFastOnStart` loads only `MainDispatcherFactory`.
You can use another service for tests.

core/kotlinx-coroutines-core/service-loader-test/test/ServiceLoaderTest.kt
qwwdfsad: E.g. by class name

qwwdfsad:
Also, please change parent of your branch to `develop`

PR #971, opened 2019-02-06
Completable job

elizarov:


kotlinx-coroutines-core/common/src/CompletableJob.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
qwwdfsad: Please update copyright year

kotlinx-coroutines-core/common/src/CompletableJob.kt
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * Repeated invocations of this function have no effect and always produce `false`.
+     *
+     * Note, that if this job has children, then it transitions into _completing_ state and becomes _complete_
+     * once all its children are _complete_. See [Job] for details.
+     */
+    public fun complete(): Boolean
+
+    /**
+     * Completes this job exceptionally with a given [exception]. The result is `true` if this job was
```
qwwdfsad: Please describe cancellation semantics of `completeExceptionally` here and in `CompletableDeferred`.

qwwdfsad: Also, we can increase deprecation level of `Job.cancel(cause)`

elizarov: We cannot simply increase deprecation level of `Job.cancel(cause)`, see #975

PR #978, opened 2019-02-09
Re-add common sources to platform sources jar

andreisilviudragnea:
This change fixes a regression introduced by commit e1fa197ddef67768c1586328cbdb2e88c89a26f0, which changed the directory layout of the project.

The directory layout change reintroduced the bug in [KT-20971](https://youtrack.jetbrains.com/issue/KT-20971), which was initially fixed in commit 2bc0a1ba13bb2015c013a3429948e73218cce909.

### DIFF
--- a/gradle/publish-bintray.gradle
+++ b/gradle/publish-bintray.gradle
@@ -29,7 +29,7 @@ task sourcesJar(type: Jar) {
 
     if (project.name == coroutines_core && platform != "common") {
         // add common source, too
-        from rootProject.file("common/$project.name-common/src/main/kotlin")
+        from rootProject.file("common/$project.name-common/src")
     }
 }
 


qwwdfsad:
Thanks for a detailed write-up!
Could you confirm that common source are not visible on Kotlin below 1.3.20 and the problem is gone on 1.3.20?

andreisilviudragnea:
I have confirmed on my local machine by building the sources jar and manually placing it in my local Maven repository for version `1.1.1`. After this, Intellij showed the sources for classes in the common module (for example, the `CoroutineName` class) in the Project view when navigating between classes in the `kotlinx.coroutines` package.

elizarov:
We've recently switched the project to a different Kotlin plugin (Kotlin multiplatform plugin) in `develop` branch and I cannot reproduce the project with source in the `develop` branch anymore, so it seem that the problem is already fixed (will be fixed in the upcoming release).

PR #980, opened 2019-02-11
Add operator invoke for CoroutineDispatcher

### LINKED ISSUE #428: Weird proposal: operator invoke extension on context
Ok. This is a weird, but let me shoot it here. How about instead of writing:

```
withContext(UI) { 
    updateSomethingInMyUI()
}
```

we could just write:

```
UI { 
    updateSomethingInMyUI()
}
```

Does it make sense? 👍or 👎

LouisCAD:
Resolves #428.

### DIFF
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -160,7 +160,7 @@ public suspend fun <T> withContext(
  * This inline function calls [withContext].
  */
 @ExperimentalCoroutinesApi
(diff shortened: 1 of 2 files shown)


elizarov:
Please add a couple of tests (so that this operation is used in tests) and squash the commits. The function itself looks good to me.

LouisCAD:
Just pushed 2 tests I shamelessly copy-pasted from `WithContextTest`.

Regarding squashing the commits, isn't this something you can do when merging by using the dropdown in the green merge button from GitHub here, as shown in the screenshot below?
screenshot 2019-02-19 at 16 11 17

PR #992, opened 2019-02-18
Check that worker threads belongs to the same scheduler in dispatcher…

### LINKED ISSUE #990: ExperimentalCoroutineDispatcher.close can shutdown enclosing thread of the same type
```
 while (true) {
            Thread.sleep(100)
            CoroutineScope(Dispatchers.Default).launch {
                ExperimentalCoroutineDispatcher(4).close()
                println("created and closed")
            }
        }
```

will print `created and closed` N times (where N is cores count of target machine)

qwwdfsad:
…s.shutdown

Fixes #990

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -299,7 +299,7 @@ internal class CoroutineScheduler(
         // atomically set termination flag which is checked when workers are added or removed
         if (!_isTerminated.compareAndSet(0, 1)) return
         // make sure we are not waiting for the current thread
-        val currentWorker = Thread.currentThread() as? Worker
+        val currentWorker = currentWorker()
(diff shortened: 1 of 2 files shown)


avently:
Thanks for such quick fix!

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -533,6 +531,11 @@ internal class CoroutineScheduler(
         return ADDED_REQUIRES_HELP
     }
 
+    private fun currentWorker(): Worker? {
+        val worker = Thread.currentThread() as? Worker ?: return null
+        return if (worker.scheduler == this) worker else null
+    }
+
```
elizarov: Consider using Kotlin idioms here and highlighting the fact it is a simple function (expression):
```
private fun currentWorker(): Worker? =
    (Thread.currentThread() as? Worker)?.takeIf { it.scheduler == this }
```

PR #998, opened 2019-02-19
Obtain real stacktrace of running coroutines in debug agent

### LINKED ISSUE #997: Capture real stacktrace of active coroutines in debug agent
Additionally to coroutine-specific stacktrace, we can also capture thread-dump and carefully merge it with coroutine stacktrace in order to provide even more debug information.

qwwdfsad:
* Capture thread of coroutine when it's transited to running state
  * During dump retrieve the stacktrace of captured thread for running coroutines and merge it with metadata-based stacktrace

Fixes #997

### DIFF
--- a/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
+++ b/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
@@ -144,12 +144,11 @@ internal object DebugProbesImpl {
         state: CoroutineState,
         coroutineTrace: List<StackTraceElement>
     ): List<StackTraceElement> {
-        val coroutineStackTrace = coroutineTrace as ArrayList<StackTraceElement>
         val thread = state.lastObservedThread
-        if (state.state != State.RUNNING || thread == null) return coroutineStackTrace
+        if (state.state != State.RUNNING || thread == null) return coroutineTrace
         // Avoid security manager issues
-        val actualTrace = kotlin.runCatching { thread.stackTrace }.getOrNull()
-            ?: return coroutineStackTrace
+        val actualTrace = runCatching { thread.stackTrace }.getOrNull()
+            ?: return coroutineTrace
 
         /*
          * Here goes heuristic that tries to merge two stacktraces: real one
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -117,22 +117,92 @@ internal object DebugProbesImpl {
...
+     * Tries to enhance [coroutineTrace] (obtained by call to [CoroutineState.lastObservedStackTrace]) with
+     * thread dump of [CoroutineState.lastObservedThread].
+     *
+     * Returns [coroutineTrace] if enhancement was unsuccessful or the enhancement result.
+     */
+    private fun enhanceStackTraceWithThreadDump(
+        state: CoroutineState,
+        coroutineTrace: List<StackTraceElement>
+    ): List<StackTraceElement> {
+        val coroutineStackTrace = coroutineTrace as ArrayList<StackTraceElement>
```
elizarov: Please, declare thus function as accepting `coroutineStackTrace` of type `ArrayList`. This would work great if you declare `lastObservedStackTrace` to return `ArrayList` (which you can change to always do by simply removing `?: return emptyList()` part in it)

Note: that you did not actually use it anywhere in the code. So either update it in place (since you know it is `ArrayList`) or just drop this `as ArrayList` part.

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -117,22 +117,92 @@ internal object DebugProbesImpl {
...
+        val thread = state.lastObservedThread
+        if (state.state != State.RUNNING || thread == null) return coroutineStackTrace
+        // Avoid security manager issues
+        val actualTrace = kotlin.runCatching { thread.stackTrace }.getOrNull()
```
elizarov: Remove `kotlin.` in `kotlin.runCatching` (that's an IDE autocompletion bug)

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -117,22 +117,92 @@ internal object DebugProbesImpl {
...
+            it.className == "kotlin.coroutines.jvm.internal.BaseContinuationImpl" &&
+                    it.methodName == "resumeWith" &&
+                    it.fileName == "ContinuationImpl.kt"
+        }
+
+        // We haven't found "BaseContinuationImpl.resumeWith" resume call in stacktrace
+        // This is some inconsistency in machinery, do not fail, fallback
+        if (indexOfResumeWith < 1) {
+            return coroutineStackTrace
+        }
```
elizarov: A minor note on style: this kinds of of "early return ifs" a better written in single-line. Consider that, but I would not insist for this particular one. You can also make more explicit alignment of this check if the code on the next line after it (blow) by dropping this whole `if` and rewriting the like below:
```
val continuationFrame = actualTrace.getOrNull(indexOfResumeWith - 1) 
    ?: return coroutineStackTrace
```
The advantage is that reader will not have to think if the check was correct and whether this line can fail or not.

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -117,22 +117,92 @@ internal object DebugProbesImpl {
...
+         * 1) Dump lastObservedThread
+         * 2) Find the next frame after BaseContinuationImpl.resumeWith (continuation machinery).
+         *   Invariant: this method is called under the lock, so such method **should** be present
+         *   in continuation stacktrace.
+         *
+         * 3) Find target method in continuation stacktrace (metadata-based)
+         * 4) Prepend dumped stacktrace (trimmed by target frame) to continuation stacktrace
+         *
+         * Heuristic may fail on recursion and overloads, but it will be automatically improved
+         * with KT-29997
```
elizarov: It will not improve with KT-29997. See implementation of `CoroutineState.updateState`. It ignores updates to the last observed frame as it is currently written.

PR #999, opened 2019-02-19
Create NestedLaunchesTest.kt

ptmt:
Should a test like this pass, or should not by design or did I do something wrong?

### DIFF
--- a/kotlinx-coroutines-core/native/test/NestedLaunchesTest.kt
+++ b/kotlinx-coroutines-core/native/test/NestedLaunchesTest.kt
@@ -1,0 +1,32 @@
+package kotlinx.coroutines
+
+import kotlin.coroutines.CoroutineContext
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class NestedLaunchesTest {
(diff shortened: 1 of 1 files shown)


qwwdfsad:
It should and it passes.
Why do you think it should not?

Though I do not recommend invoking code in place in `dispatch` **and** return `true` from `isDispatchNeeded`.

ptmt:
Sorry, I copied the dispatcher from another test in `DelayExceptionTest` by mistake. I replaced it with `Dispatcher.Main` which uses `DefaultExecutor` and fails on `takeEventLoop()`.

qwwdfsad:
Are you talking about native?
Yes, it is expected.
`runBlocking` exits prematurely because your `launch` has a completely different scope and `runBlocking` does not know about that.

In order ot fix that replace `CoroutineScope(dispatcher).launch(dispatcher)` with `launch(dispatcher)` and join on your task

In the future please create an issue for that instead of PR.

ptmt:
Sorry, silly me! It's indeed very obvious, I shouldn't have posted it.

PR #1000, opened 2019-02-19
Update named arguments referenced issue

LouisCAD:
The issue previously reference was marked as fixed for one of the two problems it was about,
and a new one has been created for the remaining issue that is relevant here.


qwwdfsad:
Good catch, thanks

PR #1004, opened 2019-02-23
Check `toMillis` overflow in operators with `Duration`

### LINKED ISSUE #868: Time.kt uses Duration.toMillis in unsafe manner
Using `Duration.ofSeconds(Long.MAX_VALUE)` as the timeout argument for [`kotlinx.coroutines.withTimeout`](https://github.com/Kotlin/kotlinx.coroutines/blob/d92b0fa7e5e5b5c19665cc649e8372bbe9cdf6e7/integration/kotlinx-coroutines-jdk8/src/time/Time.kt) will cause an ArithmaticOverflowException

This is because, in pure java `Durations.ofSeconds(Long.MAX_VALUE).toMillis()` will cause an Arithmatic overflow.

I've got a fix that basically involves adding an extension function `toMillisOrMax` that returns `Long.MAX_VALUE` when `toMillis` would otherwise overflow.

Use case:

I'm a big fan of null objects, and the way I've structured things is such that the "default" timeout is `Duration.of(seconds = Long.MAX_VALUE, nanos = 999__999_999)` (which is effectively `Duration.MAX_VALUE`), such that our down-stack code need not consider the case where no timeout is specified. 

If you guys are interested in this fix I'll submit a pull request to that involves simply:

```kt
private fun Duration.toMillisOrMax(): Long = when(this.seconds) {
    in 0L .. (Long.MAX_VALUE / MILLIS_PER_SECOND) -> duration.toMillis()
    else -> Long.MAX_VALUE
}
```

as a function in `TimeKt`, invoked instead of `Duration.toMillis()`

fvasco:
This proposal respects ChronoUnit FOREVER duration.
Fix #868

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/src/time/Time.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/time/Time.kt
@@ -4,30 +4,36 @@
 package kotlinx.coroutines.time
 
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.selects.SelectBuilder
 import java.time.Duration
-import java.util.concurrent.TimeUnit
+import kotlin.coroutines.suspendCoroutine
 
 /**
  * "java.time" adapter method for [kotlinx.coroutines.delay]
  */
-public suspend fun delay(duration: Duration) =
-    kotlinx.coroutines.delay(duration.toMillis())
+public suspend fun delay(duration: Duration) {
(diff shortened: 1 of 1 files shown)


qwwdfsad:
I am strongly against this change: silently changing the semantics of the method depending on the parameter is a no-go. Not talking about the fact that it breaks integration with custom `delay` and virtual time sources. I've already explained it [in this comment](https://github.com/Kotlin/kotlinx.coroutines/issues/868#issuecomment-446920682)

PR #1006, opened 2019-02-25
Check `toMillis` overflow in operators with `Duration`

### LINKED ISSUE #868: Time.kt uses Duration.toMillis in unsafe manner

fvasco:
Fix #868

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/src/time/Time.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/time/Time.kt
@@ -6,28 +6,45 @@ package kotlinx.coroutines.time
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.selects.SelectBuilder
 import java.time.Duration
-import java.util.concurrent.TimeUnit
+import java.time.temporal.ChronoUnit
 
 /**
  * "java.time" adapter method for [kotlinx.coroutines.delay]
  */
 public suspend fun delay(duration: Duration) =
-    kotlinx.coroutines.delay(duration.toMillis())
+        kotlinx.coroutines.delay(duration.toMillisDelay())
 
 /**
  * "java.time" adapter method for [SelectBuilder.onTimeout]
  */
 public fun <R> SelectBuilder<R>.onTimeout(duration: Duration, block: suspend () -> R) =
-    onTimeout(duration.toMillis(), block)
+        onTimeout(duration.toMillisDelay(), block)
 
 /**
  * "java.time" adapter method for [kotlinx.coroutines.withTimeout]
  */
 public suspend fun <T> withTimeout(duration: Duration, block: suspend CoroutineScope.() -> T): T =
-    kotlinx.coroutines.withTimeout(duration.toMillis(), block)
+        kotlinx.coroutines.withTimeout(duration.toMillisDelay(), block)
 
 /**
  * "java.time" adapter method for [kotlinx.coroutines.withTimeoutOrNull]
  */
 public suspend fun <T> withTimeoutOrNull(duration: Duration, block: suspend CoroutineScope.() -> T): T? =
-    kotlinx.coroutines.withTimeoutOrNull(duration.toMillis(), block)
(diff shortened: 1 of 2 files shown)


integration/kotlinx-coroutines-jdk8/src/time/Time.kt
```diff
@@ -6,28 +6,44 @@ package kotlinx.coroutines.time
...
 public suspend fun <T> withTimeoutOrNull(duration: Duration, block: suspend CoroutineScope.() -> T): T? =
-    kotlinx.coroutines.withTimeoutOrNull(duration.toMillis(), block)
\ No newline at end of file
+        kotlinx.coroutines.withTimeoutOrNull(duration.toMillisOrMax(), block)
+
+/**
+ * Try to convert a [Duration] to milliseconds, negative durations are translated to `0`,
+ * too large durations are capped to `Duration.ofMillis(Long.MAX_VALUE)`.
+ */
+private fun Duration.toMillisOrMax(): Long {
```
qwwdfsad: Please add some tests that don't pass without this change

integration/kotlinx-coroutines-jdk8/src/time/Time.kt
```diff
@@ -6,28 +6,44 @@ package kotlinx.coroutines.time
...
 import kotlinx.coroutines.selects.SelectBuilder
 import java.time.Duration
-import java.util.concurrent.TimeUnit
 
 /**
  * "java.time" adapter method for [kotlinx.coroutines.delay]
  */
-public suspend fun delay(duration: Duration) =
-    kotlinx.coroutines.delay(duration.toMillis())
+public suspend fun delay(duration: Duration) {
```
qwwdfsad: Please do not change codestyle of the methods

fvasco:
Hi @qwwdfsad 
I added some test, I you need some extra ones please notify me.

I have to confide you a problem, `gradle test` does not work on any my computers.
So I had tested all `TimeTest` methods manually, but unfortunately I does not know if they compile. Sorry for my miss.
Providing a battle tested docker image is a great help for me.

fvasco:
Please wait,
I wish to add some more tests and handle sub millisecond duration for all operators.

fvasco:
Hi @qwwdfsad 
I catch this opportunity to fix another issue, any positive duration less than 1ms is truncated to 0ms, so no delay happens and `withTimeout` crashes.
To fix this issue all strictly positive duration are coerced in `1..Long.MAX_VALUE` milliseconds.
What do you think?

qwwdfsad:
I think this is a good idea, thank you.
I've pushed your branch, improved documentation and tests and will merge it manually as soon as tests are passed.

Thank you for your contribution!

qwwdfsad:
Merged manually

PR #1012, opened 2019-02-26
Cancellation in consumeEach should dispose Rx Observable

### LINKED ISSUE #1008: Job.cancel does not dispose Rx Observable when using consumeEach
As I'm looking to convert my `Singles`, `Maybes`, and `Completables` into `suspend` functions, while still relying on RxJava's `Observables`, I was hoping that `CoroutineScope` could manage the subscription to my `Observable` streams by relying on the `consumeEach` extension function. However, it turns out that when the `Job` gets cancelled, although no more values will be "consumed" by the underlying Channel, the `Observable` itself will not get disposed. This could potentially leak the `Observable`.

Sample Repro:
```Kotlin
@Test
fun repro() {
    val observable = Observable.interval(1L, TimeUnit.SECONDS)
        .doOnNext { println("onNext received $it") }
        .doOnDispose { println("Observable disposed!") }

    val job = GlobalScope.launch {
        observable.consumeEach{
            println("Value $it consumed")
        }
    }

    Thread.sleep(3000L)
    job.cancel()
    Thread.sleep(3000L)
}
```
When I run this test I get the following result:
```
onNext received 0
Value 0 consumed
onNext received 1
Value 1 consumed
onNext received 2
onNext received 3
onNext received 4
onNext received 5
```

elizarov:
Fixed bugs in MaybeSource/ObservableSource.consumeEach implementation
so that observable is disposed on cancellation.

Also optimized implementation of bridge function to avoid extra dispose
calls if possible (this is permissible by specification, though)

Fixes #1008

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/Channel.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Channel.kt
@@ -30,9 +30,8 @@ public fun <T> Publisher<T>.openSubscription(request: Int = 0): ReceiveChannel<T
 /**
  * Subscribes to this [Publisher] and performs the specified action for each received element.
  */
-public suspend inline fun <T> Publisher<T>.consumeEach(action: (T) -> Unit) {
+public suspend inline fun <T> Publisher<T>.consumeEach(action: (T) -> Unit) =
     openSubscription().consumeEach(action)
-}
 
 @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 private class SubscriptionChannel<T>(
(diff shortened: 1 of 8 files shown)


reactive/kotlinx-coroutines-reactive/src/Channel.kt
```diff
@@ -75,6 +74,7 @@ private class SubscriptionChannel<T>(
     @Suppress("CANNOT_OVERRIDE_INVISIBLE_MEMBER")
     override fun onClosedIdempotent(closed: LockFreeLinkedListNode) {
         subscription?.cancel()
+        subscription = null // optimization -- no need to cancel it again
```
qwwdfsad: Please use CAS instead to have a more consistent guarantee.
E.g. reactor does exactly the same, see `Operators.terminate`

PR #1015, opened 2019-02-27
Add ensureActive extension

### LINKED ISSUE #963: Add checkCancelled() extension for Job / CoroutineContext / CoroutineScope
Something like:
```kt
fun CoroutineContext.checkCancelled() {
    if (this[Job]?.isCancelled == true) throw job.getCancellationException()
}
```

AFAIK, currently the only way to explicitly check coroutine shall still proceed (at the same time throwing a proper CancellationException with a proper stack trace) is to invoke `yield()`, which has a check similar to the suggested one, yet marked `internal`: https://github.com/Kotlin/kotlinx.coroutines/blob/f62acb6835fb5001f9b52602ebf73a60088d1e4c/common/kotlinx-coroutines-core-common/src/Yield.kt#L29-L32
(`Job.isCancelled` is equivalent to `!Job.isActive` when the coroutine checks that itself).

The suggestion is to expose such extension functions for `Job`, `CoroutineContext`, `CoroutineScope` (similarly to `isActive` and `cancel()`).

qwwdfsad:
Fixes #963

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineScope.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineScope.kt
@@ -215,7 +215,7 @@ public inline fun CoroutineScope.cancel() {
  * Throws [IllegalStateException] if the context does not have a job in it.
  *
  * If the job is no longer active, throws [CancellationException].
- * If the job was cancelled, thrown exception contains original cancellation cause.
+ * If the job was cancelled, thrown exception contains the original cancellation cause.
  *
  * This method is a drop-in replacement for the following code, but with more precise exception:
  * ```
--- a/kotlinx-coroutines-core/common/src/CoroutineScope.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineScope.kt
@@ -209,3 +209,19 @@ public inline fun CoroutineScope.cancel() {
     val job = coroutineContext[Job] ?: error("Scope cannot be cancelled because it does not have a job: $this")
     job.cancel()
 }
+
+/**
+ * Ensures that current scope is [active][CoroutineScope.isActive].
+ * Throws [IllegalStateException] if the context does not have a job in it.
+ *
+ * If the job is no longer active, throws [CancellationException].
+ * If the job was cancelled, thrown exception contains original cancellation cause.
+ *
+ * This method is a drop-in replacement for the following code, but with more precise exception:
+ * ```
+ * if (!isActive) {
+ *     throw CancellationException()
+ * }
+ * ```
+ */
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -536,6 +536,41 @@ public fun CoroutineContext.cancel(): Unit {
     this[Job]?.cancel()
 }
 
+/**
+ * Ensures that current job is [active][Job.isActive].
+ * If the job is no longer active, throws [CancellationException].
+ * If the job was cancelled, thrown exception contains original cancellation cause.
```
elizarov: _the_ original cancellation cause

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -209,3 +209,19 @@ public inline fun CoroutineScope.cancel() {
     val job = coroutineContext[Job] ?: error("Scope cannot be cancelled because it does not have a job: $this")
     job.cancel()
 }
+
+/**
+ * Ensures that current scope is [active][CoroutineScope.isActive].
+ * Throws [IllegalStateException] if the context does not have a job in it.
+ *
+ * If the job is no longer active, throws [CancellationException].
+ * If the job was cancelled, thrown exception contains original cancellation cause.
```
elizarov: _the_ original cancellation cause

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -536,6 +536,41 @@ public fun CoroutineContext.cancel(): Unit {
...
+public fun Job.ensureActive(): Unit {
+    if (!isActive) throw getCancellationException()
+}
+
+/**
+ * Ensures that job in the current context is [active][Job.isActive].
+ * Throws [IllegalStateException] if the context does not have a job in it.
+ *
+ * If the job is no longer active, throws [CancellationException].
+ * If the job was cancelled, thrown exception contains original cancellation cause.
```
elizarov: _the_ original cancellation cause

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -536,6 +536,41 @@ public fun CoroutineContext.cancel(): Unit {
...
+ *
+ * This method is a drop-in replacement for the following code, but with more precise exception:
+ * ```
+ * if (!isActive) {
+ *     throw CancellationException()
+ * }
+ * ```
+ */
+public fun CoroutineContext.ensureActive(): Unit {
+    val job = get(Job) ?: error("Context cannot be check for liveness because it does not have a job: $this")
```
elizarov: cannot be _checked_ for liveness

PR #1021, opened 2019-03-01
Improve CoroutineScope documentation

qwwdfsad:


qwwdfsad:
I do not like "Getting it together" name, I am open to alternatives

digitalbuddha:
Reading the section gives me the feeling that you are describing scope management.

May I suggest "Scope lifecycle management" as a section title? I would not be suprised clicking that link and seeing the description you wrote.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -6,39 +6,34 @@ package kotlinx.coroutines
...
-import kotlin.coroutines.intrinsics.*
 import kotlin.coroutines.*
+import kotlin.coroutines.intrinsics.*
 
 /**
  * Defines a scope for new coroutines. Every coroutine builder
  * is an extension on [CoroutineScope] and inherits its [coroutineContext][CoroutineScope.coroutineContext]
  * to automatically propagate both context elements and cancellation.
  *
+ * The best way to obtain a standalone instance of the scope is [CoroutineScope()] and [MainScope()] factory functions.
```
elizarov: Grammatical number is not balanced. Rephrase to either "The best way is ... _using_ this and that" or "The best _ways_ _are_ ... this and that"

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -6,39 +6,34 @@ package kotlinx.coroutines
...
+ * Additional context elements can be appended to the scope using [plus][CoroutineScope.plus] operator.
+ *
+ * Manual implementation of this interface is not recommended, implementation by delegation should be preferred instead.
+ * By convention context of the scope should contain an instance of a [job][Job] to enforce structured concurrency.
```
elizarov: By convention, ... (comma)

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -6,39 +6,34 @@ package kotlinx.coroutines
...
- *     override fun onCreate(savedInstanceState: Bundle?) {
- *         super.onCreate(savedInstanceState)
- *         job = Job()
- *     }
- *
+ * class MyActivity : AppCompatActivity(), CoroutineScope by MainScope() {
  *     override fun onDestroy() {
- *         super.onDestroy()
- *         job.cancel() // Cancel job on activity destroy. After destroy all children jobs will be cancelled automatically
+ *         cancel() // cancel is extension on CoroutineScope
```
elizarov: I would also use a chance to fix `fun showSomeData()` in this example.  Using `withContext { ... }` inside of it does not add any value. It would be shorter and more on topic if it just a comment like `// here we can use suspending functions`

kotlinx-coroutines-core/common/src/CoroutineScope.kt
elizarov: "context of the scope" would benefit, IMO, from a hyper-link to `coroutineContext` member value.

docs/coroutine-context-and-dispatchers.md
```diff
@@ -487,48 +487,43 @@ I'm working in thread DefaultDispatcher-worker-1 @test#2
...
     }
     // to be continued ...
 ```
 
 </div>
 
-We also implement [CoroutineScope] interface in this `Actvity` class. We only need to provide an override
-for its [CoroutineScope.coroutineContext] property to specify the context for coroutines launched in its
-scope. We combine the desired dispatcher (we used [Dispatchers.Default] in this example) and a job:
+We can implement [CoroutineScope] interface in this `Actvity` class. The best way to do it is
```
elizarov: Need some introductory word here, like "Alternatively, we can implement ... "

docs/coroutine-context-and-dispatchers.md
```diff
@@ -487,48 +487,43 @@ I'm working in thread DefaultDispatcher-worker-1 @test#2
...
+We also can combine the desired dispatcher (we used [Dispatchers.Default] in this example) with the scope:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
 ```kotlin
-    // class Activity continues
-    override val coroutineContext: CoroutineContext
-        get() = Dispatchers.Default + job
-    // to be continued ...
+    class Activity : CoroutineScope by CoroutineScope(Dispatchers.Default)
```
elizarov: I'd suggest to add opening curly brace and `// to be continued ...` like in the previous version of `Activity` class for consistency.

PR #1023, opened 2019-03-01
Improve stacktrace merge heuristic: skip state machine meaningless fr…

qwwdfsad:
…ame when attempting to merge stacktraces

Use timeout in all tests

### DIFF
--- a/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
+++ b/kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
@@ -6,7 +6,6 @@ package kotlinx.coroutines.debug.internal
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.debug.*
-import kotlinx.coroutines.internal.artificialFrame
 import net.bytebuddy.*
 import net.bytebuddy.agent.*
 import net.bytebuddy.dynamic.loading.*
@@ -16,6 +15,7 @@ import java.util.*
 import kotlin.collections.ArrayList
 import kotlin.coroutines.*
 import kotlin.coroutines.jvm.internal.*
+import kotlinx.coroutines.internal.artificialFrame as createArtificialFrame // IDEA bug workaround
 
 /**
  * Mirror of [DebugProbes] with actual implementation.
@@ -134,7 +134,7 @@ internal object DebugProbesImpl {
 
                 append("\n\nCoroutine $key, state: $state")
                 if (observedStackTrace.isEmpty()) {
-                    append("\n\tat ${artificialFrame(ARTIFICIAL_FRAME_MESSAGE)}")
+                    append("\n\tat ${createArtificialFrame(ARTIFICIAL_FRAME_MESSAGE)}")
                     printStackTrace(value.creationStackTrace)
                 } else {
                     printStackTrace(enhancedStackTrace)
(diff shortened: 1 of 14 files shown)


kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -249,32 +314,39 @@ internal object DebugProbesImpl {
...
+    private fun <T> createOwner(completion: Continuation<T>, frame: CoroutineStackFrame): CoroutineOwner<T> {
+        val state = CoroutineState(completion, frame, ++sequenceNumber)
+        val owner = CoroutineOwner(completion, state, frame)
+        capturedCoroutines += owner
+        return owner
     }
 
     @Synchronized
-    private fun probeCoroutineCompleted(coroutine: ArtificialStackFrame<*>) {
+    private fun probeCoroutineCompleted(coroutine: CoroutineOwner<*>) {
```
elizarov: Maybe call parameter `owner` for consistency?

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -249,32 +314,39 @@ internal object DebugProbesImpl {
...
     }
 
-    private class ArtificialStackFrame<T>(
+    /**
+     * This class is injected as completion of all continuations in [probeCoroutineCompleted].
+     * It is owning the coroutine state and responsible for managing all its external state related to debug agent.
+     */
+    private class CoroutineOwner<T>(
         @JvmField val delegate: Continuation<T>,
+        @JvmField val state: CoroutineState,
```
elizarov: Let's consistently rename `CoroutineState` to `CoroutineInfo` and vals from `state` and `coroutineState` consistently to `info`.

Also, `CoroutineState.continuation` does not need to be public as it seems. It looks like it can be replaced with `context`.

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -219,23 +260,47 @@ internal object DebugProbesImpl {
...
+        }
+
         // Find ArtificialStackFrame of the coroutine
-        val owner = frame.owner()
+        val owner = frame.owner() ?: return
         updateState(owner, frame, state)
     }
 
+    @Synchronized // See comment to infoCache
+    private fun updateRunningState(continuation: Continuation<*>, state: State) {
```
elizarov: It would be more consistent to always name the parameter `frame` in all `updateState` functions. Also, consider using `frame: CoroutineStackFrame` type as a parameter in all internal function as they all seem to contain the same line of code: `val frame = continuation as? CoroutineStackFrame ?: return`

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
```diff
@@ -249,32 +314,40 @@ internal object DebugProbesImpl {
...
+        val owner = CoroutineOwner(completion, info, frame)
+        capturedCoroutines += owner
+        return owner
     }
 
     @Synchronized
-    private fun probeCoroutineCompleted(coroutine: ArtificialStackFrame<*>) {
-        capturedCoroutines.remove(coroutine)
+    private fun probeCoroutineCompleted(owner: CoroutineOwner<*>) {
+        capturedCoroutines.remove(owner)
```
elizarov: Need to also remove `owner.info.lastObservedFrame.realCaller()` from `infoCache`.

kotlinx-coroutines-debug/src/internal/DebugProbesImpl.kt
qwwdfsad: I've done it only for this method.
Otherwise code becomes less readable: somewhere `Continuation<*>` is required, `updateState` is ok with nullable `CoroutineStackFrame` etc.

PR #1030, opened 2019-03-07
Do not use reflective exception pre-handler on newer Androids, since …

### LINKED ISSUE #822: Throwing exceptions inside coroutines generates warnings on Android 9 devices
Running this code on Android 9 device/emulator:

```
GlobalScope.launch {
    throw IllegalStateException()
}
```

will result in a `Accessing hidden method Ljava/lang/Thread;->getUncaughtExceptionPreHandler()Ljava/lang/Thread$UncaughtExceptionHandler; (light greylist, reflection)` warning printed into Logcat.

Android 9 includes [crackdown on developers using non-public stuff via reflection](https://developer.android.com/about/versions/pie/restrictions-non-sdk-interfaces), and looks like coroutines got caught in this.

qwwdfsad:
…Pie runtime ensures that uncaught exception is always logged

Fixes #822

### DIFF
--- a/ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt
+++ b/ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt
@@ -4,30 +4,42 @@
 
 package kotlinx.coroutines.android
 
+import android.os.*
 import android.support.annotation.*
 import kotlinx.coroutines.*
 import java.lang.reflect.*
 import kotlin.coroutines.*
 
-private val getter =
-    try {
-        Thread::class.java.getDeclaredMethod("getUncaughtExceptionPreHandler").takeIf {
-            Modifier.isPublic(it.modifiers) && Modifier.isStatic(it.modifiers)
-        }
-    }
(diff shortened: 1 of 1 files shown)


ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt
```diff
@@ -4,30 +4,42 @@
...
 import kotlin.coroutines.*
 
-private val getter =
-    try {
-        Thread::class.java.getDeclaredMethod("getUncaughtExceptionPreHandler").takeIf {
-            Modifier.isPublic(it.modifiers) && Modifier.isStatic(it.modifiers)
-        }
-    }
-    catch (e: Throwable) { null /* not found */ }
+private val getter by lazy {
```
JakeWharton: Shouldn't this be defined inside `AndroidExceptionPreHandler` to avoid a second class being generated for the file members.

qwwdfsad: Good point, though moving it inside AEPH is not enough

PR #1034, opened 2019-03-08
Update awaitAll documentation wording

digitalbuddha:
* double `when`
* 2nd change can be either `wait for the failing deferred` or `await the failing deferred`. Sounds a bit off with `wait the failing deferred`.

Let me know which is preferred


qwwdfsad:
Thanks!

PR #1043, opened 2019-03-14
Add ThreadLocal.isPresent and ThreadLocal.ensurePresent methods

### LINKED ISSUE #1028: Provide safety mechanisms for working with thread-locals
Currently, `kotlinx.coroutines` provides [primitives](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/java.lang.-thread-local/index.html) to work with thread locals.

The problem is that they are not quite safe and it is easy to accidentally forget them (#1016).
Moreover, sometimes it is necessary to know whether thread local is present in the context in order to do some action (e.g. probabilistic request tracing using thread locals, the case from Circlet).

In all cases, it is impossible to check whether thread local element is actually present in the context, and checking thread local itself is not an option, because it can have a stale or default value.

We can provide following extensions (similar to `coroutineContext`) to cover that cases and promote their usages in the documentation:
```
suspend fun TheadLocal<*>.ensurePresent(): Unit
suspend fun TheadLocal<*>.isPresent(): Boolean
suspend inline fun <T> ThreadLocal<T>.ifPresent(block: suspend (T) -> Unit) // Or R?
```

qwwdfsad:
Fixes #1028

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
+++ b/kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
@@ -135,3 +135,39 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
  */
 public fun <T> ThreadLocal<T>.asContextElement(value: T = get()): ThreadContextElement<T> =
     ThreadLocalElement(value, this)
+
+/**
+ * Return `true` when current thread local is present in the coroutine context, `false` otherwise.
+ * Thread local can be present in the context only if it was added via [asContextElement] to the context.
+ *
+ * Example of usage:
+ * ```
+ * suspend fun processRequest() {
+ *   if (traceCurrentRequestThreadLocal.isPresent()) { // Probabilistic tracing
+ *      // Do some heavy-weight tracing
+ *   }
+ *   // Process request regularly
+ * }
+ * ```
+ */
+public suspend fun ThreadLocal<*>.isPresent(): Boolean = coroutineContext[ThreadLocalKey(this)] != null
+
+/**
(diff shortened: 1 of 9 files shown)


kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -135,3 +135,39 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
+ * ```
+ * suspend fun processRequest() {
+ *   if (traceCurrentRequestThreadLocal.isPresent()) { // Probabilistic tracing
+ *      // Do some heavy-weight tracing
+ *   }
+ *   // Process request regularly
+ * }
+ * ```
+ */
+public suspend fun ThreadLocal<*>.isPresent(): Boolean = coroutineContext[ThreadLocalKey(this)] != null
```
elizarov: Make it `inline` so that suspension point is not generated at its use-site.

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -135,3 +135,39 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
+ *   return get()
+ * }
+ *
+ * // Usage
+ * withContext(...) {
+ *   val value = threadLocal.getSafely() // Fail-fast in case of improper context
+ * }
+ * ```
+ */
+public suspend fun ThreadLocal<*>.ensurePresent(): Unit = check(isPresent()) { "ThreadLocal $this is missing from context $coroutineContext" }
```
elizarov: Make it `inline` so that suspension point is not generated at its use-site.

PR #1048, opened 2019-03-18
Deprecate `SendChannel.isFull` and `ReceiveChannel.isEmpty` properties

ndkoval:


kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -30,14 +30,15 @@ public interface SendChannel<in E> {
      * Returns `true` if the channel is full (out of capacity) and the [send] attempt will suspend.
      * This function returns `false` for [isClosedForSend] channel.
      *
-     * **Note: This is an experimental api.** This property may change its semantics and/or name in the future.
+     * **Note: will be removed in next releases, no replacement.**
```
elizarov: Write `@suppress` before this node, so that val is excluded from KDoc.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -151,13 +152,14 @@ public interface ReceiveChannel<out E> {
      * Returns `true` if the channel is empty (contains no elements) and the [receive] attempt will suspend.
      * This function returns `false` for [isClosedForReceive] channel.
      *
-     * **Note: This is an experimental api.** This property may change its semantics and/or name in the future.
+     * **Note: will be removed in next releases, no replacement.**
```
elizarov: Write `@suppress` before this node, so that val is excluded from KDoc.

PR #1050, opened 2019-03-19
Improved coroutine exception handling logic

elizarov:
* JobSupport.handleJobException (which is called before the coroutine goes
  to the final state) is now called only when parent did not handle
  exception and is used only in launch-like coroutines (launch and actor)
  to report uncaught exception.
* The final result of other types coroutines that, by default, have an
  object "to store exceptions" is consistently processed after all
  other notifications when the coroutine state is already complete and
  is reported as uncaught exception as a last-resort processing tactic
  when it cannot be otherwise handled.
* The above change makes the order of notifications for future { ... } and
  other async-like coroutine builders consistent with the order in which
  async { ... }.await() is notified.
* The "handled" state of exception (whether it was processed or not)
  is now stored in CompletedExceptionally marker class (pulled this
  field up from CancelledContinuation).
* AbstractCoroutine.onCompletedExceptionally is renamed to onCancelled
  and includes "handled" flag.
* protected val JobSupport.completionCauseHandled is introduced to access
  "handled" flag, which simplified reactive streams integration.
* "suppressed" flag is dropped from JobSupport.onCompletionInternal.
* Fixed exception processing in reactive integrations -- exception that
  is thrown from a cancelled (disposed) coroutine is not lost anymore.

### DIFF
--- a/kotlinx-coroutines-core/common/src/CompletedExceptionally.kt
+++ b/kotlinx-coroutines-core/common/src/CompletedExceptionally.kt
@@ -18,8 +18,12 @@ internal fun <T> Result<T>.toState(): Any? =
  *        or artificial [CancellationException] if no cause was provided
  */
 internal open class CompletedExceptionally(
-    @JvmField public val cause: Throwable
+    @JvmField public val cause: Throwable,
+    handled: Boolean = false
 ) {
+    private val _handled = atomic(handled)
+    val handled: Boolean get() = _handled.value
+    fun makeHandled(): Boolean = _handled.compareAndSet(false, true)
     override fun toString(): String = "$classSimpleName[$cause]"
 }
 
@@ -34,10 +38,7 @@ internal class CancelledContinuation(
     continuation: Continuation<*>,
     cause: Throwable?,
     handled: Boolean
-) : CompletedExceptionally(cause ?: CancellationException("Continuation $continuation was cancelled normally")) {
-    private val resumed = atomic(false)
-    private val handled = atomic(handled)
-
-    fun makeResumed(): Boolean = resumed.compareAndSet(false, true)
-    fun makeHandled(): Boolean = handled.compareAndSet(false, true)
+) : CompletedExceptionally(cause ?: CancellationException("Continuation $continuation was cancelled normally"), handled) {
+    private val _resumed = atomic(false)
+    fun makeResumed(): Boolean = _resumed.compareAndSet(false, true)
 }
--- a/kotlinx-coroutines-core/common/src/JobSupport.kt
+++ b/kotlinx-coroutines-core/common/src/JobSupport.kt
@@ -98,9 +98,9 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
          ~ active coroutine is working (or scheduled to execution)
          >> childCancelled / cancelImpl invoked
        ## CANCELLING: state is Finishing, state.rootCause != null
-        ------ cancelling listeners are not admitted anymore, invokeOnCompletion(onCancellation=true) returns NonDisposableHandle
+        ------ cancelling listeners are not admitted anymore, invokeOnCompletion(onCancelling=true) returns NonDisposableHandle
         ------ new children get immediately cancelled, but are still admitted to the list
-         + onCancellation
+         + onCancelling
          + notifyCancelling (invoke all cancelling listeners -- cancel all children, suspended functions resume with exception)
          + cancelParent (rootCause of cancellation is communicated to the parent, parent is cancelled, too)
          ~ waits for completion of coroutine body
@@ -203,7 +203,9 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         require(state.isCompleting) // consistency check -- must be marked as completing
         val proposedException = (proposedUpdate as? CompletedExceptionally)?.cause
         // Create the final exception and seal the state so that no more exceptions can be added
+        var wasCancelling = false // KLUDGE: we cannot have contract for our own expect fun synchronized
         val finalException = synchronized(state) {
+            wasCancelling = state.isCancelling
             val exceptions = state.sealLocked(proposedException)
             val finalCause = getFinalRootCause(state, exceptions)
             if (finalCause != null) addSuppressedExceptions(finalCause, exceptions)
@@ -223,6 +225,10 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
             val handled = cancelParent(finalException) || handleJobException(finalException)
             if (handled) (finalState as CompletedExceptionally).makeHandled()
         }
+        // Process state updates for the final state before the state of the Job is actually set to the final state
+        // to avoid races where outside observer may see the job in the final state, yet exception is not handled yet.
+        if (!wasCancelling) onCancelling(finalException)
+        onCompletionInternal(finalState)
         // Then CAS to completed state -> it must succeed
         require(_state.compareAndSet(state, finalState.boxIncomplete())) { "Unexpected state: ${_state.value}, expected: $state, update: $finalState" }
         // And process all post-completion actions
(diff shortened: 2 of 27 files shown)


kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -303,7 +298,7 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
          *    It should be last so all callbacks observe consistent state
          *    of the job which doesn't depend on callback scheduling.
          */
-        onCompletionInternal(update, mode, suppressed)
+        onCompletionInternal(update, mode)
```
qwwdfsad: It looks like this is the wrong place to handle an exception:
```
runBlocking {
  val job = Job()
  val ch = produce<Int>(job + Dispatchers.Default + CoroutineExceptionHandler {
         _, t -> Thread.sleep(100); println("Handled $t") }) {
     throw AssertionError()
  }
   ch.cancel()
   job.complete()
   job.join()
   println("Joined")
}
```

prints 
```
Joined
Handled java.lang.AssertionError
```

When completion handlers are invoked, it is already too late to handle exceptions

qwwdfsad: Note that now we have unused `onCancellationInternal` that looks like the right place to handle exceptions in produce and actor

qwwdfsad: ...looking at `onCompletionInternal` that eventually leads to `onCancelled` or `onCompleted`

Both functions have a meaningful side-effect (future completion, exception handling, flux completion etc.).
It is inconsistent to have successful `join` while coroutine is still doing meaningful work that changes observing state.

kotlinx-coroutines-core/common/src/CompletedExceptionally.kt
```diff
@@ -18,8 +18,12 @@ internal fun <T> Result<T>.toState(): Any? =
  *        or artificial [CancellationException] if no cause was provided
  */
 internal open class CompletedExceptionally(
-    @JvmField public val cause: Throwable
+    @JvmField public val cause: Throwable,
+    handled: Boolean = false
 ) {
+    private val _handled = atomic(handled)
+    val handled: Boolean get() = _handled.value
```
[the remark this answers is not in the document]
elizarov: This is just encapsulation of atomic. This way usage of `handled` elsewhere looks nice (just `state.handle` as opposed to `state.handled.value`)

kotlinx-coroutines-core/common/src/JobSupport.kt
elizarov: Answering in pieces. `onCancellation` is invoked too early, when the coroutine is just cancelled before we know its final exception. It is too early to **handle** exception there, because the exception is not final yet (coroutine might crash while handling it). The only legal use of `onCancellation` is to cancel whatever other work the coroutine might have been doing. Note, that `onCancellation` is always called _before_ cancelling all children. And it _is_ actually used in _actor_ builder to cancel the channel as soon as possible.

elizarov: Good catch with code example. I've further tweaked the order to resolve this problem. I think I made a break-through in understanding the problem: different kinds of coroutines need to plug into a different phases of "completion" processing:

* Various futures, produce, etc need to update their state even _before_ the state of the Job becomes final (to be race-free), so that a successful `join` always sees a "linked" object is a proper state. They all override `onCompletionInternal` (via `onCancelled`/`onCompleted`) to do so.
So far, I'm giving this guarantee only for failures, but we might consider extending this guarantee to successful completions, too (not sure we need to at the moment)

* Scoped coroutines need to resume the rest of the code _after_ everything is done. I've created a separate `afterCompletionInternal` for them (they also need to know `mode`, while `onCompletionInternal` does not need `mode`).

elizarov:
Hm... it is not that simple. Reactive integrations are currently written with the assumption that you can read coroutine's state from `onCompletionInternal` handlers and they'll need to be tweaked, too.

elizarov:
Fixed them!

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -257,6 +262,8 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         check(state is Empty || state is JobNode<*>) // only simple state without lists where children can concurrently add
         check(update !is CompletedExceptionally) // only for normal completion
         if (!_state.compareAndSet(state, update.boxIncomplete())) return false
+        onCancellation(null) // simple state is not a failure
```
qwwdfsad: If it is not, then maybe we should not invoke `onCancellation`?
Especially when its doc says "This function is invoked once when this job is being cancelled"

qwwdfsad: It is now used only in `actor`, so there are no bugs, but nevertheless, we contradict our own contract

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -257,6 +262,8 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
+        onCompletionInternal(update)
```
qwwdfsad: Proposal: move both `onCompletionInternal` and `onCancellation`  into `completeStateFinalization`.
Also, for the sake of simplicity, we can completely remove `onCancellation` and use `onCancelled` instead.
Having both `onCancellation` and `onCancelled` with more or less the same contract looks fragile 
WDYT?

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -294,11 +295,10 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
         }
         /*
-         * 4) Invoke onCompletionInternal: onNext(), timeout de-registration etc.
-         *    It should be last so all callbacks observe consistent state
-         *    of the job which doesn't depend on callback scheduling.
+         * 3) Resumes the rest of the code in scoped coroutines
+         *    (runBlocking, coroutineScope, withContext, withTimeout, etc)
          */
-        onCompletionInternal(update, mode)
+        afterCompletionInternal(update, mode)
```
qwwdfsad: Why some functions have "internal" suffix, while others haven't?

elizarov: This is a legacy distinction from the times we had "internal" `JobSupport` and "public" `AbstractCoroutine`. Functions with `Internal` suffix were never supposed to become part of public API. They are either taking `mode: Int` or operating on `state: Any` (or both).

kotlinx-coroutines-core/common/src/JobSupport.kt
elizarov: We cannot move them to `completeStateFinalization`, because `completeStateFinalization` is invoked _after_ that state is already final, so what happens if we do it in `completeStateFinalization`:

1. Job1's state becomes final.
2. Job1's `onCompletionInternal` is invoked (exception is handled).
3. Waiting coroutines are notified.

So, now if we have another coroutine Job2 doing `Job1.join()` then we have a race -- usually exception in Job1 is handled before the Job2 resumes, but sometimes it could check the state of Job1 in just the right moment (in between steps 1 and 2) and return without suspending when the exception was not handled yet.

kotlinx-coroutines-core/common/src/JobSupport.kt
elizarov: Good catch. I've updated it to be be invoked only on cancellation and also renamed it to `onCancelling` for consistency with `invokeOnCompletion(onCancelling=xxx)` parameter. Also, it turns out that the condition it was called under from `tryFinalizeFinishingState` was not correct, but it did not really matter (since it is used only from `actor` extra calls from there do not harm anyway).

elizarov:
Nope. This does not work this way. We need to cancel actor's channel on exit, too.

PR #1055, opened 2019-03-23
Update Channels docs to Koltin 1.3

mlegy:
Use `sequence` and `iterator` instead of `buildSequence` and `buildIterator`.


elizarov:
Thanks. Good catch.

PR #1074, opened 2019-04-09
Flow leftovers

qwwdfsad:


reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
```diff
@@ -76,3 +79,40 @@ public fun <T : Any> ReceiveChannel<T>.asObservable(context: CoroutineContext):
...
+     * ATOMIC is used here to provide stable behaviour of subscribe+dispose pair even if
+     * asObservable is already invoked from unconfined
+     */
+    val job = GlobalScope.launch(Dispatchers.Unconfined, start = CoroutineStart.ATOMIC) {
+        try {
+            collect { value -> emitter.onNext(value) }
+            emitter.onComplete()
+        } catch (e: Throwable) {
+            // 'create' provides safe emitter, so we can unconditionally call on* here if exception occurs in `onComplete`
+            if (e !is CancellationException) emitter.onError(e)
```
qwwdfsad: This behaviour is inherited from `CoroutineScope.rxObservable()`, we should discuss whether we want to change it

PR #1083, opened 2019-04-10
Update Tasks to support recent Play Services builds

mpetrov:
Play Services 16 changed the `task.result` to `@Nullable`, so some additional casting is
required to compile.

Note that the generic `T` in `Task<T>` is not currently annotated. The nullability of `result` should match that of `T`.


mpetrov:
@aaronj1335 @shmuelr FYI

elizarov:
I've merged it into `develop`

PR #1084, opened 2019-04-10
Introduce CancellableContinuation.resume with onCancelling lambda

### LINKED ISSUE #1044: Proper handling of closable resources after #896
This is a follow up of #896 that started on slack.

Let's take the following example.
```
    private suspend fun Call.await(): WorkerResult {
        return suspendCancellableCoroutine { continuation ->
            continuation.invokeOnCancellation {
                cancel()
            }
            enqueue(object : Callback {
                override fun onResponse(call: Call, response: Response) {
                    continuation.resume(WorkerResult.Success(response))
                }

                override fun onFailure(call: Call, e: IOException) {
                    if (call.isCanceled || !isActive) {
                        continuation.resume(WorkerResult.Error(CancellationException("Cancelled")))
                    } else {
                        continuation.resume(WorkerResult.Error(e))
                    }
                }
            })
        }
    }
```

How can we ensure that the resume from ```onResponse()```is either successful or that we can close the response.
Adding isActive tests is race condition prone, and I can't think of a way to avoid a race condition leading to leaks of the response.

cc: @elizarov

elizarov:
* Allows safe return of closeable resources from suspending
functions, as it provides a way to close a resource if the
corresponding job was cancelled.
* Documentation on the context and expected behavior of
CompletionHandler implementations is updated.

Fixes #1044

### DIFF
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -123,10 +123,16 @@ public interface CancellableContinuation<in T> : Continuation<T> {
      * with cancellation exception. Otherwise, the handler will be invoked once on cancellation if this
      * continuation is cancelled.
      *
-     * Installed [handler] should not throw any exceptions. If it does, they will get caught,
-     * wrapped into [CompletionHandlerException], and rethrown, potentially causing the crash of unrelated code.
+     * Installed [handler] should not throw any exceptions.
+     * If it does, they will get caught, wrapped into [CompletionHandlerException] and
+     * processed as uncaught exception in the context of the current coroutine
+     * (see [CoroutineExceptionHandler]).
      *
      * At most one [handler] can be installed on one continuation.
+     *
+     * **Note**: Implementation of `CompletionHandler` must be fast, non-blocking, and thread-safe.
+     * This handler can be invoked concurrently with the surrounding code.
+     * There is no guarantee on the execution context in which the [handler] is invoked.
      */
     public fun invokeOnCancellation(handler: CompletionHandler)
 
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -151,6 +156,33 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
+     * This function shall be used when resuming with a resource that must be closed by the
+     * code that had called the corresponding suspending function, e.g.:
+     *
+     * ```
+     * continuation.resume(resource) {
+     *     resource.close()
+     * }
+     * ```
+     *
+     * **Note**: Implementation of `CompletionHandler` must be fast, non-blocking, and thread-safe.
```
qwwdfsad: >and thread-safe.

Why so? This handler is invoked only once from one unspecified thread.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -151,6 +156,33 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
+     * There is no guarantee on the execution context in which the [onCancellation] handler is invoked.
+     */
+    @ExperimentalCoroutinesApi // since 1.2.0, tentatively graduates in 1.3.0
+    public fun resume(value: T, onCancellation: CompletionHandler)
```
qwwdfsad: nitpick: `CompletionHandler` has `(cause: Throwable?) -> Unit` type (and documented how to treat `cause` parameter), while here we should have either `(cause: Throwable) -> Unit` or `() -> Unit`

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -165,8 +173,18 @@ internal open class CancellableContinuationImpl<in T>(
...
-    override fun resumeWith(result: Result<T>) =
+    override fun resumeWith(result: Result<T>) {
         resumeImpl(result.toState(), resumeMode)
+    }
+
+    override fun resume(value: T, onCancellation: CompletionHandler) {
+        if (!resumeImpl(CompletedWithCancellation(value, onCancellation), resumeMode)) {
+            // too late to resume (was cancelled) -- call handler
+            invokeHandlerSafely {
+                onCancellation.invokeIt((state as? CompletedExceptionally)?.cause)
```
qwwdfsad: `null` shouldn't be allowed here, it is better to use `!!` or explicit invariant check than safe call

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -102,6 +102,14 @@ internal open class CancellableContinuationImpl<in T>(
 
     override fun takeState(): Any? = state
 
+    override fun cancelResult(state: Any?, cause: Throwable) {
```
qwwdfsad: It shouldn't leak to our ABI, please make it internal

PR #1089, opened 2019-04-12
Version 1.2.0

qwwdfsad:


gradle.properties
```diff
@@ -1,11 +1,11 @@
 # Kotlin
 version=1.2.0-alpha-2-SNAPSHOT
 group=org.jetbrains.kotlinx
-kotlin_version=1.3.21
+kotlin_version=1.3.30
```
elizarov: Need to search/replace version across all files (READMEs, example projects, etc)

PR #1113, opened 2019-04-18
CompletableJob.asCompletableFuture

### LINKED ISSUE #1104: Introduce CompletableJob.asCompletableFuture
There is a simillar method for `Deferred`:
```kotlin
public fun <T> Deferred<T>.asCompletableFuture(): CompletableFuture<T>
```

Would be nice to have a replacement to migrate from `CompletableDeferred<Unit>`.

qwwdfsad:
Fixes #1104

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
@@ -83,15 +83,15 @@ public fun <T> Deferred<T>.asCompletableFuture(): CompletableFuture<T> {
 }
 
 /**
- * Converts this completable job to the instance of [CompletableFuture].
+ * Converts this job to the instance of [CompletableFuture].
  * The job is cancelled when the resulting future is cancelled or otherwise completed.
  */
 public fun Job.asCompletableFuture(): CompletableFuture<Unit> {
     val future = CompletableFuture<Unit>()
     setupCancellation(future)
-    invokeOnCompletion { throwable ->
-        if (throwable === null) future.complete(Unit)
-        else future.completeExceptionally(throwable)
+    invokeOnCompletion { cause ->
+        if (cause === null) future.complete(Unit)
+        else future.completeExceptionally(cause)
     }
     return future
 }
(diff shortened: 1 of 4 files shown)


integration/kotlinx-coroutines-jdk8/src/future/Future.kt
```diff
@@ -86,6 +82,28 @@ public fun <T> Deferred<T>.asCompletableFuture(): CompletableFuture<T> {
     return future
 }
 
+/**
+ * Converts this completable job to the instance of [CompletableFuture].
+ * The job is cancelled when the resulting future is cancelled or otherwise completed.
+ */
+public fun CompletableJob.asCompletableFuture(): CompletableFuture<Unit> {
```
elizarov: It has to be extension on `Job`, not `CompletableJob`, just like we have extension on `Deferred` and we don't need `CompletableDeferred`, since the value link is only one way, and only cancellation goes the other way around.

integration/kotlinx-coroutines-jdk8/src/future/Future.kt
```diff
@@ -86,6 +82,28 @@ public fun <T> Deferred<T>.asCompletableFuture(): CompletableFuture<T> {
     return future
 }
 
+/**
+ * Converts this completable job to the instance of [CompletableFuture].
```
elizarov: "completable job" -> "job"

integration/kotlinx-coroutines-jdk8/src/future/Future.kt
```diff
@@ -86,6 +82,28 @@ public fun <T> Deferred<T>.asCompletableFuture(): CompletableFuture<T> {
...
+ * The job is cancelled when the resulting future is cancelled or otherwise completed.
+ */
+public fun Job.asCompletableFuture(): CompletableFuture<Unit> {
+    val future = CompletableFuture<Unit>()
+    setupCancellation(future)
+    invokeOnCompletion { throwable ->
```
elizarov: Shall parameter be called `cause` for consistency?

PR #1114, opened 2019-04-18
android-example Add Empty Tack, Tack With Progress, Cancelable Tack

jura73:
I think that there are not enough examples on Android. 

Then I would like to add examples with error handling, with request repetition, with Retrofit, tests, presenters and requests to the server.

### DIFF
--- a/ui/kotlinx-coroutines-android/example-app/app/build.gradle
+++ b/ui/kotlinx-coroutines-android/example-app/app/build.gradle
@@ -4,7 +4,6 @@ apply plugin: 'kotlin-android-extensions'
 
 android {
     compileSdkVersion 27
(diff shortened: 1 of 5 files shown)


elizarov:
We actually want to ultimately get rid of all Android example in this repository, since it is not the right place to host them. This is a general-purpose coroutines library and domain-specific examples only serve as distraction here.

Google is already doing great job of creating educational materials for Android programmers, see for example this codelab: 
https://codelabs.developers.google.com/codelabs/kotlin-coroutines/#0
Ultimately `developers.google.com` is much better place for all those examples.

jura73:
Thanks for the reply, close the pull request.

PR #1127, opened 2019-04-21
Fix a race in Job.join that sporadically results in normal completion

### LINKED ISSUE #1123: No CancellationException thrown when join on a crashed Job

elizarov:
The race happens in the slow-path of 'join' implementation when
parent invokes join on a child coroutines that crashes and cancels
the parent.

Fixes #1123

### DIFF
--- a/kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
@@ -170,6 +170,18 @@ internal open class CancellableContinuationImpl<in T>(
         // otherwise, onCompletionInternal was already invoked & invoked tryResume, and the result is in the state
         val state = this.state
         if (state is CompletedExceptionally) throw recoverStackTrace(state.cause, this)
+        // if the parent job was already cancelled, then throw the corresponding cancellation exception
+        // otherwise, there is a race is suspendCancellableCoroutine { cont -> ... } does cont.resume(...)
+        // before the block returns. This getResult would return a result as opposed to cancellation
+        // exception that should have happened if the continuation is dispatched for execution later.
+        if (resumeMode == MODE_CANCELLABLE) {
+            val job = context[Job]
+            if (job != null && !job.isActive) {
+                val cause = job.getCancellationException()
+                cancelResult(state, cause)
+                throw cause
+            }
+        }
         return getSuccessfulResult(state)
     }
 
--- a/kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
@@ -179,7 +179,7 @@ internal open class CancellableContinuationImpl<in T>(
             if (job != null && !job.isActive) {
                 val cause = job.getCancellationException()
                 cancelResult(state, cause)
-                throw cause
+                throw recoverStackTrace(cause, this)
             }
         }
         return getSuccessfulResult(state)
--- a/kotlinx-coroutines-core/jvm/test/JobStructuredJoinStressTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/JobStructuredJoinStressTest.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+import org.junit.*
+
+/**
+ * Test a race between job failure and join.
+ *
+ * See [#1123](https://github.com/Kotlin/kotlinx.coroutines/issues/1123).
+ */
+class JobStructuredJoinStressTest : TestBase() {
+    private val nRepeats = 1_000 * stressTestMultiplier
+
+    @Test
+    fun testStress() {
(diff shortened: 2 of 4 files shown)


LouisCAD:
Is this planned for 1.2.1?

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -170,6 +170,18 @@ internal open class CancellableContinuationImpl<in T>(
...
         val state = this.state
         if (state is CompletedExceptionally) throw recoverStackTrace(state.cause, this)
+        // if the parent job was already cancelled, then throw the corresponding cancellation exception
+        // otherwise, there is a race is suspendCancellableCoroutine { cont -> ... } does cont.resume(...)
+        // before the block returns. This getResult would return a result as opposed to cancellation
+        // exception that should have happened if the continuation is dispatched for execution later.
+        if (resumeMode == MODE_CANCELLABLE) {
+            val job = context[Job]
+            if (job != null && !job.isActive) {
+                val cause = job.getCancellationException()
```
qwwdfsad: We should recover stacktrace here as well (and in `DispatchedContinuation.run#232`)

kotlinx-coroutines-core/jvm/test/JobStructuredJoinStressTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        repeat(nRepeats) {
+            assertFailsWith<TestException> {
+                runBlocking {
+                    // launch in background
+                    val job = launch(Dispatchers.Default) {
+                        throw TestException("OK") // crash
+                    }
+                    try {
+                        job.join()
+                    } catch (e: CancellationException) {
```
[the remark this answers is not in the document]
elizarov: Good catch. Refactoring leftovers. Should be `assertFailsWith<CancellationException> { job.join() }`

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
elizarov: Yes!

PR #1132, opened 2019-04-22
Flow operators:

### LINKED ISSUE #1107: Provide Flow operators for UI programming
There is a family of operators which are widespread in UI but which are tricky to implement (especially with current `select` statement):
-- `switchMap`
-- `debounce` (`throttle` and `sample`?)
-- `takeUntil(flow)` (?)

qwwdfsad:
* switchMap
    * debounce
    * sample
    * Update RxJava version to 2.2.8

Partially fixes #1107


kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -32,3 +35,125 @@ public fun <T> Flow<T>.delayEach(timeMillis: Long): Flow<T> = flow {
...
 }
+
+/**
+ * Returns a flow that mirrors the original flow, but filters out values
+ * that are followed by the newer values within the given [timeout][timeoutMs].
+ * The latest value is always emitted.
+ * Example:
+ * ```
+ * flow {
+ *   emit(1)
```
qwwdfsad: indents

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -32,3 +35,125 @@ public fun <T> Flow<T>.delayEach(timeMillis: Long): Flow<T> = flow {
...
+        }
+    }
+}
+
+/**
+ * Returns a flow that emits only the latest value emitted by the original flow during the given sampling [period][periodMs].
+ * Example:
+ * ```
+ * flow {
+ *   repeat(10) {
```
qwwdfsad: indents

kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
```diff
@@ -89,15 +89,50 @@ public fun <T> Flow<Flow<T>>.flattenConcat(): Flow<T> = flow {
...
+ * ```
+ * will produce `aa bb b_last`
+ */
+@FlowPreview
+public fun <T, R> Flow<T>.switchMap(transform: suspend (value: T) -> Flow<R>): Flow<R> = flow {
+    coroutineScope {
+        var previousFlow: Job? = null
+        collect { value ->
+            // Linearize calls to emit as alternative to the channel. Bonus points for never-overlapping channels.
+            previousFlow?.cancelAndJoin()
```
streetsofboston: Is joining the previousFlow necessary when doing a switch-map? Wouldn't calling just `previousFlow?.cancel()` be enough?

kotlinx-coroutines-core/common/test/flow/VirtualTime.kt
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+import kotlin.coroutines.*
+import kotlin.jvm.*
+
+private class VirtualTimeDispatcher(enclosingScope: CoroutineScope) : CoroutineDispatcher(), Delay {
```
streetsofboston: You may be able to hook up to the `TestCoroutineScope` and `TestCoroutineDispatcher` from this PR https://github.com/Kotlin/kotlinx.coroutines/pull/890 instead of using this `VirtualTimeDispatcher` class.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -32,3 +35,125 @@ public fun <T> Flow<T>.delayEach(timeMillis: Long): Flow<T> = flow {
...
+
+            var isDone = false
+            var lastValue: Any? = null
+            while (!isDone) {
+                select<Unit> {
+                    values.onReceive {
+                        lastValue = it
+                    }
+
+                    onTimeout(timeoutMs) {
```
bnorm: As is, won't this select statement run every `timeoutMs` for an idle Flow? Wrapping with `if (lastValue != null)` still passes all the unit tests and should avoid extra checking? The same question applies to `Flow<T>.sample` and the `ticker.onReceive` select statement below.

kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
qwwdfsad: I'd rather say is necessary otherwise the flow that does not support cancellation (or e.g. does something like `finally { withContext(NonCancellable) { emit(theVeryLastItem) } }` will ovelap with a newer flow.

kotlinx-coroutines-core/common/test/flow/VirtualTime.kt
qwwdfsad: I was thinking about it, but at 1.2.1 I am not ready to make `TestCoroutineScope` multiplatform

kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
streetsofboston: True! 
But not joining doesn't affect the result and the previous 'collecting' of the flow is still canceled, I think. 
Would joining not 'slow down' the resulting stream?

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
elizarov: I'm taking care of it.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
elizarov: fixing it

kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
elizarov: The joining will slow down the resulting stream _for good_, because we always wait for previous resources to be released, so the thing like `swithMap { doExpensiveRemoteCall() }` does not risk running us out of expensive resources.

streetsofboston: That is true. But is it up to the`switchMap` implementation to guard against this?

I think that if `doExpsensiveRemoteCall()` is so expensive, the `doExpensiveRemoteCall()` should guard against that issue (or wrap something around the call to guard against it).

In Rx, switchMap cancels/disposes the previous subscription. It doesn't wait for the subscription to be cancelled/disposed has completed.

JakeWharton: That is not true. RxJava calls `cancel()` (which invokes `dispose()`) on the active subscription synchronously. All work required to stop the previous stream will happen prior to the mapper being called for the new item and thus prior to the subscription to the returned stream.

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java#L104-L107

elizarov: > I think that if `doExpsensiveRemoteCall()` is so expensive, the `doExpensiveRemoteCall()` should guard against that issue

That's exactly the point. `doExpsensiveRemoteCall()` _is guarded_, because it is coded like:
```
suspend funn doExpsensiveRemoteCall() {
    val con = openExpensiveConnection()
    try {
        // do call
    } finally {
        con.close() // Guarding against leak of expensive resource
    }
} 
```

So, now it is responsibility to of `switchMap` implementation to wait until `doExpsensiveRemoteCall` is complete (and releases all its resources)

streetsofboston: Then I stand corrected 😄 
When looking at the Rx docks, I thought that cancelation was done asynchronously:
http://reactivex.io/RxJava/javadoc/rx/Observable.html#switchMap-rx.functions.Func1-

Just out of curiosity:
Is it synchronous because the call to `inner.cancel()` is synchronous?
Or is it because the for-loop keeps checking until the inner has been canceled:
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/observable/ObservableSwitchMap.java#L121-L130

JakeWharton: It's the `inner.cancel()` which calls `dispose()` on the current upstream. That loop guards against the downstream canceling concurrently with an update from upstream.

PR #1133, opened 2019-04-22
AtomicFu version 0.12.5

### LINKED ISSUE #1064: duplicate entry: META-INF/atomicfu.kotlin_module
I've tried [this](https://github.com/Kotlin/kotlinx.coroutines/issues/1059#issuecomment-476805836) and it doesn't work. I'm a bit stuck. 😢

### LINKED ISSUE #1116: Proguard warning about AtomicFU#setInterceptor()
After updating to 1.2.0 I got 

`Warning: kotlinx.atomicfu.AtomicFU: can't find referenced method 'void setInterceptor(kotlinx.atomicfu.AtomicOperationInterceptor)' in program class kotlinx.atomicfu.InterceptorKt
`
from ProGuard, version 6.0.3. 

I solved it by adding  `-dontwarn kotlinx.atomicfu.AtomicFU` in proguard rules, but I'm in doubt if it's a safe way to fix it and even if it is. It's probably required to add this line to [coroutines.pro](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro) file.

elizarov:
AtomicFu version 0.12.5

This should really fix processing of class files

Fixes #1116
Fixes #1064

### DIFF
--- a/binary-compatibility-validator/test/PublicApiTest.kt
+++ b/binary-compatibility-validator/test/PublicApiTest.kt
@@ -55,13 +55,16 @@ class PublicApiTest(
 
     private fun getJarPath(libsDir: File): File {
         val regex = Regex("$moduleName-.+\\.jar")
-        val files = (libsDir.listFiles() ?: throw Exception("Cannot list files in $libsDir"))
+        var files = (libsDir.listFiles() ?: throw Exception("Cannot list files in $libsDir"))
             .filter { it.name.let {
                     it matches regex
                     && !it.endsWith("-sources.jar")
                     && !it.endsWith("-javadoc.jar")
                     && !it.endsWith("-tests.jar")}
                     && !it.name.contains("-metadata-")}
-        return files.singleOrNull() ?: throw Exception("No single file matching $regex in $libsDir:\n${files.joinToString("\n")}")
+        if (files.size > 1) // maybe multiplatform?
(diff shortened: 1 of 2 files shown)


binary-compatibility-validator/test/PublicApiTest.kt
```diff
@@ -55,13 +55,16 @@ class PublicApiTest(
...
-        val files = (libsDir.listFiles() ?: throw Exception("Cannot list files in $libsDir"))
+        var files = (libsDir.listFiles() ?: throw Exception("Cannot list files in $libsDir"))
             .filter { it.name.let {
                     it matches regex
                     && !it.endsWith("-sources.jar")
                     && !it.endsWith("-javadoc.jar")
                     && !it.endsWith("-tests.jar")}
                     && !it.name.contains("-metadata-")}
-        return files.singleOrNull() ?: throw Exception("No single file matching $regex in $libsDir:\n${files.joinToString("\n")}")
+        if (files.size > 1) // maybe multiplatform?
```
qwwdfsad: Is it maybe or definitely? :)

elizarov: This is just a kludge. It could be the case where you build multiple versions and so this code still will not be able to pick the right jar out of many. A better fix would be to pass the jar name from gradle build file.

PR #1136, opened 2019-04-23
Fix header documentation typo in "TestCoroutineContext"

ahmedmsvb:
Delete the duplicated "based" word from header doc


elizarov:
Thanks

PR #1138, opened 2019-04-24
Context passing between coroutines and Reactor Mono/Flux

SokolovaMaria:


reactive/kotlinx-coroutines-reactor/src/Flux.kt
```diff
@@ -32,9 +42,238 @@ import kotlin.coroutines.*
...
+    val coroutine = PublisherCoroutine(newContext, subscriber)
+    subscriber.onSubscribe(coroutine) // do it first (before starting coroutine), to avoid unnecessary suspensions
+    coroutine.start(CoroutineStart.DEFAULT, coroutine, block)
+}
+
+private const val CLOSED = -1L    // closed, but have not signalled onCompleted/onError yet
+private const val SIGNALLED = -2L  // already signalled subscriber onCompleted/onError
+
+@Suppress("CONFLICTING_JVM_DECLARATIONS", "RETURN_TYPE_MISMATCH_ON_INHERITANCE")
+private class PublisherCoroutine<in T>(
```
qwwdfsad: Copy-pasting the whole integration is definitely not an option: it is hard to maintain, hard to test and easy to mess up.
For example, we can make it public and mark with `InternalCoroutinesApi`

reactive/kotlinx-coroutines-reactor/src/Flux.kt
```diff
@@ -32,9 +42,238 @@ import kotlin.coroutines.*
...
 @ExperimentalCoroutinesApi
 fun <T> CoroutineScope.flux(
     context: CoroutineContext = EmptyCoroutineContext,
     @BuilderInference block: suspend ProducerScope<T>.() -> Unit
-): Flux<T> =
-    Flux.from(publish(newCoroutineContext(context), block = block))
+): Flux<T> = Flux.from(reactorPublish(newCoroutineContext(context), block = block))
+
+@ExperimentalCoroutinesApi
+public fun <T> CoroutineScope.reactorPublish(
```
qwwdfsad: Why do we have to introduce public `reactorPublish` API?
Integration should work with `flux` builder.

reactive/kotlinx-coroutines-reactor/src/Flux.kt
```diff
@@ -32,9 +42,238 @@ import kotlin.coroutines.*
...
  *        to cancellation and error handling may change in the future.
  */
+
 @ExperimentalCoroutinesApi
 fun <T> CoroutineScope.flux(
     context: CoroutineContext = EmptyCoroutineContext,
     @BuilderInference block: suspend ProducerScope<T>.() -> Unit
-): Flux<T> =
-    Flux.from(publish(newCoroutineContext(context), block = block))
+): Flux<T> = Flux.from(reactorPublish(newCoroutineContext(context), block = block))
```
qwwdfsad: It looks like instead of having `is CoreSubscriber` hack (that may stop working at some point, it looks like we are relying on implementation detail) we should use `Flux.create {}` API here in the same way we are using `Mono.create {}`

qwwdfsad: And this is where we should abstract some parts of `PublisherCoroutine` away to share most of the implementation details

reactive/kotlinx-coroutines-reactor/test/ReactorContextTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import kotlinx.coroutines.reactive.awaitFirst
+import kotlinx.coroutines.runBlocking
+import org.junit.Test
+import reactor.util.context.Context
+import kotlin.test.assertEquals
+
+class ReactorContextTest {
+    @Test
+    fun monoHookedContext() = runBlocking(Context.of(1, "1", 7, "7").asCoroutineContext()) {
```
qwwdfsad: code style: all our tests are started with `test` prefix

reactive/kotlinx-coroutines-reactor/src/ReactorContext.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import reactor.util.context.Context
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+
+class ReactorContext(val context: Context) : AbstractCoroutineContextElement(ReactorContext) {
+    companion object Key : CoroutineContext.Key<ReactorContext>
+}
+
+fun Context.asCoroutineContext(): CoroutineContext = ReactorContext(this)
```
qwwdfsad: code style: all public API should have explicit `public` modifier

reactive/kotlinx-coroutines-reactor/src/ReactorContext.kt
```diff
@@ -0,0 +1,11 @@
+package kotlinx.coroutines.reactor
+
+import reactor.util.context.Context
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+
+class ReactorContext(val context: Context) : AbstractCoroutineContextElement(ReactorContext) {
```
qwwdfsad: All public API should be properly documented, with corresponding examples.
A good example of such documentation is `ThreadLocal<T>.asContextElement` API.

I am very interested in examples of how this API is supposed to be used, maybe a @sdeleuze could point a few pseudocode-like examples?

qwwdfsad: @SokolovaMaria note that both examples should be present: one with accessing reactor context from coroutine builder and the other with passing coroutine context from suspend function (or `Flow`?) to reactor entity

reactive/kotlinx-coroutines-reactor/src/Flux.kt
```diff
@@ -32,9 +35,24 @@ import kotlin.coroutines.*
...
+
+
+private fun <T> CoroutineScope.reactorPublish(
+    context: CoroutineContext = EmptyCoroutineContext,
+    @BuilderInference block: suspend ProducerScope<T>.() -> Unit
+): Publisher<T> = Publisher { subscriber ->
+    // specification requires NPE on null subscriber
+    if (subscriber == null) throw NullPointerException("Subscriber cannot be null")
+    val currentContext = if (subscriber is CoreSubscriber) subscriber.currentContext() else Context.empty()
```
qwwdfsad: I've asked Sebastian to verify whether this approach is stable or not.
As a backup plan, `Flux.create` can be used by adapting `FluxSink` to `Subscriber` and using corresponding callbacks: https://www.screencast.com/t/2Z2fFBq6HSC

reactive/kotlinx-coroutines-reactor/test/ReactorContextTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import reactor.util.context.Context
+import kotlin.test.assertEquals
+
+class ReactorContextTest {
+    @Test
+    fun testMonoHookedContext() = runBlocking(Context.of(1, "1", 7, "7").asCoroutineContext()) {
+        val mono = mono {
+            val ctx = coroutineContext[ReactorContext]?.context
+            buildString {
+                (1 .. 7).forEach { append(ctx?.getOrDefault(it, "noValue")) }
```
qwwdfsad: code style: `1..7`. Autoformatter should have caught that

qwwdfsad: Also, probably you meant `2..7` because key `1` is already in the context

reactive/kotlinx-coroutines-reactor/src/Flux.kt
qwwdfsad: Reactor team responded that the current way is enough, let's stick to it.
But I'd rather throw an exception if a subscriber is not `CoreSubscriber`

reactive/kotlinx-coroutines-reactor/src/Flux.kt
```diff
@@ -32,9 +34,25 @@ import kotlin.coroutines.*
...
+): Flux<T> = Flux.from(reactorPublish(newCoroutineContext(context), block = block))
+
+
+private fun <T> CoroutineScope.reactorPublish(
+    context: CoroutineContext = EmptyCoroutineContext,
+    @BuilderInference block: suspend ProducerScope<T>.() -> Unit
+): Publisher<T> = Publisher { subscriber ->
+    // specification requires NPE on null subscriber
+    if (subscriber == null) throw NullPointerException("Subscriber cannot be null")
+    check (subscriber is CoreSubscriber) { "Subscriber is not an instance of CoreSubscriber, context can not be extracted." }
```
qwwdfsad: 1) Code style, should not have an indent
2) Should be `require` (-> IAE instead of ISE)

reactive/kotlinx-coroutines-reactor/test/ReactorContextTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun testContextFromReactorToCoroutine() = runBlocking(Context.of(1, "global 1").asCoroutineContext()) {
+        mono {
+            assertEquals(coroutineContext[ReactorContext]!!.context[2], "2")
+            assertEquals(coroutineContext[ReactorContext]!!.context[1], "global 1")
+        }.subscriberContext(Context.of(2, "2", 3, "3"))
+                .subscribe()
+        mono {
+            assertEquals(coroutineContext[ReactorContext]!!.context.getOrDefault(1, "no value"), "global 1")
+        }.subscribe()
+        println()
```
qwwdfsad: Debug leftovers? (`println`)

PR #1144, opened 2019-04-25
kotlinx-coroutines-test cleanup

qwwdfsad:
* Get rid of ThreadSafeHeap usages, make it internal again
  * Yield support in test dispatcher
  * Allow partial override of coroutine context
  * Code style fixes

### DIFF
--- a/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
+++ b/kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
@@ -136,6 +136,7 @@ public class TestCoroutineDispatcher: CoroutineDispatcher(), Delay, DelayControl
     // Storing time in nanoseconds internally.
     private val _time = atomic(0L)
 
+    /** @suppress */
     override fun dispatch(context: CoroutineContext, block: Runnable) {
         if (dispatchImmediately) {
             block.run()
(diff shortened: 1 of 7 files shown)


kotlinx-coroutines-core/jvm/src/internal/ThreadSafeHeap.kt
```diff
@@ -20,8 +19,7 @@ public interface ThreadSafeHeapNode {
  * Synchronized binary heap.
  * @suppress **This an internal API and should not be used from general code.**
  */
-@InternalCoroutinesApi
-public class ThreadSafeHeap<T> : SynchronizedObject() where T: ThreadSafeHeapNode, T: Comparable<T> {
+internal class ThreadSafeHeap<T> : SynchronizedObject() where T: ThreadSafeHeapNode, T: Comparable<T> {
```
elizarov: Using `ThreadSafeHeap` was my explicit ask to make sure it behaves consistently when tasks are submitted from other threads. It goes not mean it will actually work, but at least will not produce some spectacularly strange behaviors. We'll need to use our own class any way when we start to commonized Test APIs.

kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -128,14 +128,15 @@ public class TestCoroutineDispatcher: CoroutineDispatcher(), Delay, DelayControl
         }
 
     // The ordered queue for the runnable tasks.
-    private val queue = ThreadSafeHeap<TimedRunnable>()
+    private val queue = PriorityBlockingQueue<TimedRunnable>()
 
     // The per-scheduler global order counter.
-    private val _counter = atomic(0L)
+    private val counter = atomic(0L)
```
elizarov: I've named them with underscore for consistency with `_state` and `_decision` variables in `CancellableContinuationImpl`, because `_counter.value` looks subjectively better than `counter.value`. See also disucssion on this style here: https://github.com/Kotlin/kotlinx.coroutines/pull/890#issuecomment-485159899

PR #1160, opened 2019-04-30
Rename TimedRunnable to TimedRunnableObsolete in obsolete kotlinx.cor…

### LINKED ISSUE #1159: Duplicate class kotlinx.coroutines.test.TimedRunnable
```
> 1 exception was raised by workers:
  java.lang.RuntimeException: Duplicate class kotlinx.coroutines.test.TimedRunnable found in modules kotlinx-coroutines-core-1.2.1.jar (org.jetbrains.kotlinx:kotlinx-coroutines-core:1.2.1) and kotlinx-coroutines-test-1.2.1.jar (org.jetbrains.kotlinx:kotlinx-coroutines-test:1.2.1)
```

Maybe the class (maybe the whole `test` package) should be only part of the `kotlinx-coroutines-test` module?

qwwdfsad:
…outines.test package to avoid FQN clash

Fixes #1159

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/test_/TestCoroutineContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/test_/TestCoroutineContext.kt
@@ -43,7 +43,7 @@ class TestCoroutineContext(private val name: String? = null) : CoroutineContext 
     }
 
     // The ordered queue for the runnable tasks.
-    private val queue = ThreadSafeHeap<TimedRunnable>()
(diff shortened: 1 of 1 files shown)


ZakTaccardi:
Could this be hotfixed into a `1.2.2` release? We have a release coming up and we can't bump to `1.3.0-RC`, and need to release with `1.2.1` (or `1.2.2` if you patch this)

If not - is there a special gradle command that would allow this to build for us? Maybe rename the class on the fly at build time?

PR #1178, opened 2019-05-06
Ensure that there are no references to atomicfu in classes

elizarov:
* Updated to 0.12.7 version of atmoicfu with fixes.
* Added test that that is no references to atomicfu in any of the
  resulting class files.

### DIFF
--- a/binary-compatibility-validator/test/PublicApiTest.kt
+++ b/binary-compatibility-validator/test/PublicApiTest.kt
@@ -43,14 +43,18 @@ class PublicApiTest(
     @Test
     fun testApi() {
         val libsDir = File("../$rootDir/$moduleName/build/libs").absoluteFile.normalize()
-        val jarFile = getJarPath(libsDir)
+        val jarPath = getJarPath(libsDir)
         val kotlinJvmMappingsFiles = listOf(libsDir.resolve("../visibilities.json"))
         val visibilities =
                 kotlinJvmMappingsFiles
                         .map { readKotlinVisibilities(it) }
                         .reduce { m1, m2 -> m1 + m2 }
-        val api = getBinaryAPI(JarFile(jarFile), visibilities).filterOutNonPublic(nonPublicPackages)
-        api.dumpAndCompareWith(File("reference-public-api").resolve("$moduleName.txt"))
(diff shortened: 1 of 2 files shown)


binary-compatibility-validator/test/PublicApiTest.kt
```diff
@@ -68,3 +72,26 @@ class PublicApiTest(
...
+        if (e.name.endsWith(".class")) {
+            getInputStream(e).use {
+                val bytes = it.readBytes()
+                loop@for (i in 0 until bytes.size - ATOMIC_FU_REF.size) {
+                    for (j in 0 until ATOMIC_FU_REF.size) {
+                        if (bytes[i + j] != ATOMIC_FU_REF[j]) continue@loop
+                    }
+                    // found!
+                    println("Found atomicfu reference in ${e.name}")
+                    found = true
```
qwwdfsad: Please raise error right here and add classname to the error message

binary-compatibility-validator/test/PublicApiTest.kt
```diff
@@ -68,3 +72,26 @@ class PublicApiTest(
             error("No single file matching $regex in $libsDir:\n${files.joinToString("\n")}")
     }
 }
+
+private val ATOMIC_FU_REF = "Lkotlinx/atomicfu/".toByteArray()
+
+private fun JarFile.checkForAtomicFu() {
+    var found = false
+    for (e in entries()) {
+        if (e.name.endsWith(".class")) {
```
qwwdfsad: nit: reverting this condition will make the code a bit more readable (one less indentation level)

binary-compatibility-validator/test/PublicApiTest.kt
elizarov: It's hard to debug when it fails on the first class, but I'm rewriting it to raise one exception with a list of all affected classes instead of printing the list to stdout.

binary-compatibility-validator/test/PublicApiTest.kt
elizarov: Good catch. I've found a way to save one more indent level, too: `val bytes = getInputStream(e).use { it.readBytes() }`

PR #1194, opened 2019-05-15
kotlinx-coroutines-test README: use runBlockingTest for withTimeout sample code

satoshun:
Maybe this sample needs runBlockingTest method.


elizarov:
Good catch. Thanks.

PR #1198, opened 2019-05-16
Add combineLatest with multiple flows

### LINKED ISSUE #1193: Provide combineLatest overload with vararg flow parameter
See the original request with rationale: https://github.com/Kotlin/kotlinx.coroutines/issues/437#issuecomment-491237681

The key question is how (if) to provide a type-safe API.
Whether we need up to N overloads with N generic parameters or single vararg parameter of type `Flow<*>`

qwwdfsad:
* combineLatest(Iterable<Flow<T>>) is not added deliberately, use-case is unclear
  * All specific overloads are marked as inline to reduce binary compatibility pressure

Fixes #1193


Thomas-Vos:
> combineLatest(Iterable<Flow>) is not added deliberately, use-case is unclear

Here is my use case for this operator:
```kotlin
fun <T> combineLatest(flows: List<Flow<T>>): Flow<List<T>> = TODO()

// A list of connected devices.
val devices: List<Device> = getDevices()

// A list of Flows which receive the state of the device every time it changes.
val flows: List<Flow<DeviceState>> = devices.map { device ->
    device.observeState() // Flow<DeviceState>
}

// Combine the latest state of the devices into a single Flow.
val flow: Flow<List<DeviceState>> = combineLatest(flows)


// Somewhere else in the code
flow.collect { states ->
    for (state in states) {
        // TODO: do something with the states.
    }
}
```

elizarov:
Please, create a separate issue.

Issue #689, opened 2018-10-09
Handle exception only once in case of Job in the coroutines hierarchy


qwwdfsad:
Current exception handling mechanism can report the same exception multiple times.

For example, the exception is reported by every child if their parent can't handle it:
```
val job = Job()
launch(job) {
    delay(Long.MAX_VALUE)
}

launch(job) {
    throw IOException()
}
job.join()
```

`IOException` will be reported twice, though only the second coroutine is responsible for it.

Second problem is even less intuitive: when attaching to already failed job, exception *always* will be reported. This is the case from #686.

qwwdfsad:
More examples:
```
 CoroutineScope(Job()).launch {
            val job = Job(coroutineContext[Job])
            launch(job) {
                println("A")
                throw Exception()
            }
        }
```

Mygod:
From my crashlytics, it seems like something like this still happens in 1.3.0-RC...

qwwdfsad:
What exactly? The same exception is handled twice?

Mygod:
Never mind. It seems to be an error on my end.

PR #1199, opened 2019-05-16
Improve coroutine exception handling logic

qwwdfsad:
* Wrap exceptions into CE if parent cancels its child. This is possible because now we can be sure that no one is invoking 'cancel(Throwable)', thus if parent cancels its children, an exception is definitely handled.
  * This changes allows exceptions in hierarchies to be handled only once even when coroutines in the hierarchy are capable of handling exceptions
  * Breaking change: CompletableJob.completeExceptionally(exception) uses its parameter as additional debug information for children ('cause' of CE), thus changing the result of its children. For example, if CJ was a parent of CompleteableDeferred and CJ was completed exceptionally with exception E1, now CD.await() will throw CE instead of E1
  * Recursively check whether parent handles exception to avoid duplicates reporting when Job() is present in the hierarchy

Fixes for #689

### DIFF
--- a/kotlinx-coroutines-core/common/src/JobSupport.kt
+++ b/kotlinx-coroutines-core/common/src/JobSupport.kt
@@ -635,29 +635,20 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
     private fun createJobCancellationException() =
         JobCancellationException("Job was cancelled", null, this)
 
-    override fun getChildJobCancellationCause(): Throwable {
+    override fun getChildJobCancellationCause(): CancellationException {
         // determine root cancellation cause of this job (why is it cancelling its children?)
         val state = this.state
         val rootCause = when (state) {
-            is Finishing -> state.rootCause
+            is Finishing -> state.rootCause as? CancellationException ?: JobCancellationException("Parent job was cancelled", state.rootCause, this)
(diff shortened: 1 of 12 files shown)


kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -635,29 +637,20 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
-         * If this parent job handles exceptions, then wrap cause into JobCancellationException, because we
-         * don't want the child to handle this exception on more time. Otherwise, pass our original rootCause
-         * to the child for cancellation.
-         */
-        return if (rootCause == null || handlesException && rootCause !is CancellationException) {
-            JobCancellationException("Parent job is ${stateString(state)}", rootCause, this)
-        } else {
-            rootCause
-        }
+        return rootCause ?: JobCancellationException("Parent job is ${stateString(state)}", rootCause, this)
```
elizarov: Remove duplication in this function code: Use `when` to find `rootCause`, but don't convert it to cancellation exception just yet, then do wrapping in the resturn statement:
```
return rootCase as? CancellationException 
    ?: JobCancellationException("Parent job is ${stateString(state)}", rootCause, this)
```

Issue #1202, opened 2019-05-17
Provide a way to determine if TestCoroutineDispatcher is "idle"


objcode:
When writing an advanced test that uses multiple paused and related `TestCoroutineDispatcher` instances, there is a problem that can occur when one dispatches a task to another:

```
val dispatcherA = TestCoroutineDispatcher()
val dispatcherB = TestCoroutineDispatcher()

// inject those somehow to the class under test

@Test
fun test_withMultipleDispatchers() {
    dispatcherA.pause()
    dispatcherB.pause()
    
    foo()
    
    dispatcherA.runUntilIdle()
    dispatcherB.runUntilIdle()
    // executing tasks on B enqueued another task on A. Currently there is no way to determine if both of them are actually idle.
}

fun foo() {
    scope.launch(dispatcherA) {
        bar()
    }
}

fun bar() {
    scope.launch(dispatcherB) {
        foo()
    }
}
```

In this contrived example, there's an infinite recursion so it will never resolve, but this problem comes up in normal test code.

This also comes up in a more typical case of testing `withContext`. To test `withContext` is called, the easiest way is to inject a dispatcher to intercept the coroutine started by `withContext`.  However, in the current API for TestCoroutineDispatcher there is no way to determine if a task has been sent to a dispatcher.

In addition, this sort of inspection is important for implementing an Espresso IdlingResource as discussed in https://github.com/Kotlin/kotlinx.coroutines/issues/242.

A bit of a complication here  (since this should fit well with the Espresso integration which likely needs to work w/ regular dispatchers). A dispatcher has the ability to differentiate between two states without introducing extra tracking:
1. `Idle` (all coroutines are suspended, or delayed into the future)
2. `Busy` (some coroutine is currently running)

However, a TestCoroutineDispatcher can differentiate between three states
1. `Idle` (all coroutines are suspended)
2. `Busy` (some coroutine is currently running)
3. `DelayQueued(long)` (some coroutine will resume after long delay)

No concrete API proposals at this second - wanted to make a issue to track this.

elizarov:
But why mutliple test dispatchers are needed? Why one is not enough? What practical problem is being solved by using multiple test dispatchers?

objcode:
cc @yigit @manuelvicnt who have provided the feedback that they wanted use multiple dispatchers.

manuelvicnt:
> why mutliple test dispatchers are needed?

[...]

PR #1206, opened 2019-05-18
Support completing the test coroutine from outside the test thread.

objcode:
Implement an eventLoop for runBlockingTest.
    
    As a result, runBlockingTest is now correct in the presence of repeated
    time-delayed dispatch from other dispatchers.
    
    Changes:
    - runBlockingTest will now allow a 30 second timeout for other
    dispatchers to complete coroutines
    - Introduced WaitConfig, SingleDispatcherWaitConfig, and
    MultiDispatcherWaitConfig to configure runBlockingTest timeout behavior
    - Added DelayController.queueState as a ConflatedBroadcastChannel to
    observe or poll the queue status of a TestCoroutineDispatcher
    - Added queue status of Idle, HasCurrentTask and HasDelayedTask as
    public APIs for quering the status of a TestCoroutineDispatcher
    - Added dependency on runBlocking from runBlockingTest
    - Improved documentation for threading concerns around resumeDispatche

### DIFF
--- a/kotlinx-coroutines-test/src/DelayController.kt
+++ b/kotlinx-coroutines-test/src/DelayController.kt
@@ -2,6 +2,7 @@ package kotlinx.coroutines.test
 
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.channels.ConflatedBroadcastChannel
 
 /**
  * Control the virtual clock time of a [CoroutineDispatcher].
@@ -112,9 +113,75 @@ public interface DelayController {
      * Resumed dispatchers will automatically progress through all coroutines scheduled at the current time. To advance
      * time and execute coroutines scheduled in the future use, one of [advanceTimeBy],
      * or [advanceUntilIdle].
+     *
+     * When the dispatcher is resumed, all execution be immediate in the thread that triggered it. This means
+     * that the following code will not switch back from Dispatchers.IO after `withContext`
+     *
+     * ```
+     * runBlockingTest {
+     *     withContext(Dispatchers.IO) { doIo() }
+     *     // runBlockingTest is still on Dispatchers.IO here
+     * }
+     * ```
+     *
+     * For tests that need accurate threading behavior, [pauseDispatcher] will ensure that the following test dispatches
+     * on the correct thread.
+     *
+     * ```
+     * runBlockingTest {
+     *     pauseDispatcher()
+     *     withContext(Dispatchers.IO) { doIo() }
+     *     // runBlockingTest has returned to it's starting thread here
+     * }
+     * ```
      */
     @ExperimentalCoroutinesApi // Since 1.2.1, tentatively till 1.3.0
     public fun resumeDispatcher()
+
+    /**
+     * Represents the queue state of a DelayController.
+     *
+     * Tests do not normally need to use this API. It is exposed for advanced situations like integrating multiple
+     * [TestCoroutineDispatcher] instances or creating alternatives to [runBlockingtest].
+     */
+    public sealed class QueueState {
+        /**
+         * A [DelayController] that is idle does not currently have any tasks to perform.
+         *
+         * This may happen if all coroutines in this [DelayController] have completed, or if they are all suspended
+         * waiting on other dispatchers.
+         */
+        public object Idle: QueueState() {
+            override fun toString() = "Idle"
+        }
+
+        /**
+         * A [DelayController] that has a task that will execute in response to a call to [runCurrent].
+         *
+         * There may also be delayed tasks scheduled, in which case [HasCurrentTask] takes priority since current tasks
+         * will execute at an earlier virtual time.
+         */
+        public object HasCurrentTask: QueueState() {
+            override fun toString() = "HasCurrentTask"
+        }
+
+        /**
+         * A [DelayController] that has delayed tasks has a task scheduled for sometime in the future.
+         *
+         * If there are also tasks at the current time, [HasCurrentTask] will take priority.
+         */
+        public object HasDelayedTask: QueueState(){
+            override fun toString() = "HasDelayedTask"
+        }
+    }
+
+    /**
+     * A ConflatedBroadcastChannel that is up to date with the current [QueueState].
+     *
+     * Tests do not normally need to use this API. It is exposed for advanced situations like integrating multiple
+     * [TestCoroutineDispatcher] instances or creating alternatives to [runBlockingtest].
+     */
+    public val queueState: ConflatedBroadcastChannel<QueueState>
 }
 
 /**
--- a/kotlinx-coroutines-test/src/DelayController.kt
+++ b/kotlinx-coroutines-test/src/DelayController.kt
@@ -94,6 +94,9 @@ public interface DelayController {
      *
      * This is useful when testing functions that start a coroutine. By pausing the dispatcher assertions or
      * setup may be done between the time the coroutine is created and started.
+     *
+     * While in the paused block, the dispatcher will queue all dispatched coroutines and they will be resumed on
+     * whatever thread calls [advanceUntilIdle], [advanceTimeBy], or [runCurrent].
      */
     @ExperimentalCoroutinesApi // Since 1.2.1, tentatively till 1.3.0
     public suspend fun pauseDispatcher(block: suspend () -> Unit)
@@ -103,6 +106,9 @@ public interface DelayController {
      *
      * When paused, the dispatcher will not execute any coroutines automatically, and you must call [runCurrent] or
      * [advanceTimeBy], or [advanceUntilIdle] to execute coroutines.
+     *
+     * While paused, the dispatcher will queue all dispatched coroutines and they will be resumed on whatever thread
+     * calls [advanceUntilIdle], [advanceTimeBy], or [runCurrent].
      */
     @ExperimentalCoroutinesApi // Since 1.2.1, tentatively till 1.3.0
     public fun pauseDispatcher()
@@ -114,8 +120,9 @@ public interface DelayController {
      * time and execute coroutines scheduled in the future use, one of [advanceTimeBy],
      * or [advanceUntilIdle].
      *
-     * When the dispatcher is resumed, all execution be immediate in the thread that triggered it. This means
-     * that the following code will not switch back from Dispatchers.IO after `withContext`
+     * When the dispatcher is resumed, all execution be immediate in the thread that triggered it similar to
+     * [Dispatchers.Unconfined]. This means that the following code will not switch back from Dispatchers.IO after
+     * `withContext`
      *
      * ```
      * runBlockingTest {
@@ -125,7 +132,7 @@ public interface DelayController {
      * ```
      *
      * For tests that need accurate threading behavior, [pauseDispatcher] will ensure that the following test dispatches
-     * on the correct thread.
+     * on a controlled thread.
      *
      * ```
      * runBlockingTest {
--- a/kotlinx-coroutines-test/src/DelayController.kt
+++ b/kotlinx-coroutines-test/src/DelayController.kt
@@ -137,51 +137,6 @@ public interface DelayController {
      */
     @ExperimentalCoroutinesApi // Since 1.2.1, tentatively till 1.3.0
     public fun resumeDispatcher()
-
-    /**
-     * Represents the queue state of a DelayController.
-     *
-     * Tests do not normally need to use this API. It is exposed for advanced situations like integrating multiple
-     * [TestCoroutineDispatcher] instances or creating alternatives to [runBlockingtest].
-     */
-    public sealed class QueueState {
-        /**
-         * A [DelayController] that is idle does not currently have any tasks to perform.
-         *
-         * This may happen if all coroutines in this [DelayController] have completed, or if they are all suspended
-         * waiting on other dispatchers.
-         */
-        public object Idle: QueueState() {
-            override fun toString() = "Idle"
-        }
-
-        /**
-         * A [DelayController] that has a task that will execute in response to a call to [runCurrent].
-         *
-         * There may also be delayed tasks scheduled, in which case [HasCurrentTask] takes priority since current tasks
-         * will execute at an earlier virtual time.
-         */
-        public object HasCurrentTask: QueueState() {
-            override fun toString() = "HasCurrentTask"
-        }
-
-        /**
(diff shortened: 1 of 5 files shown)


objcode:
Note to reviewer: Reading the code, I believe invokeOnCompletion will get a invoked  result if the coroutine was already completed in every case (including completion by another thread). Can you confirm that this understanding is correct and the code here is sound?

Thanks,
Sean

Zhelyazko:
Any ETA on when this will be released?

kotlinx-coroutines-test/src/TestBuilders.kt
```diff
@@ -46,18 +46,56 @@ public fun runBlockingTest(context: CoroutineContext = EmptyCoroutineContext, te
...
+        throw UncompletedCoroutinesError("Test completed all jobs after cleanup code started. This is " +
+                "thrown to avoid non-deterministic behavior in tests (the next execution may fail randomly). Ensure " +
+                "all threads started by the test are completed before returning from runBlockingTest.")
+    }
+}
+
+private fun Deferred<Unit>.getResultIfKnown(): Pair<Boolean, Throwable?> {
+    var testError: Throwable? = null
+    var wasExecuted = false
+    invokeOnCompletion { errorFromTestOrNull ->
```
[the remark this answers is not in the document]
objcode: Thanks!

objcode:
Thank you for the detailed review! Agreed completely this is a surprising API and introduces an unfortunate non-determinism.

Putting together a new commit based on your suggestions.

objcode:
@qwwdfsad @elizarov I've uploaded a proposed change that adds an event loop to runBlockingTest. PTAL!

If you can take a look for API review I'll make any updates necessary. One big decision that I wasn't sure about - there is a way to implement this by making TestCoroutineDispatcher be an EventLoop. However, I do kind of like the current impl (using a state channel) since it also exposes the API requested in https://github.com/Kotlin/kotlinx.coroutines/issues/1202.

For the parameter, I ended up giving it a class since it was really confusing to pass a timeout of 0 for the expected failure case (it felt like it would lead to false positives when using).

See

```
runBlockingTest(timeout = 0) { }

vs.

runBlockingTest(waitConfig = SingleDispatcherWaitConfig) {}
```

Let me know if you think this works or if you would prefer to see it implemented with an EventLoop.

MariusBoepple:
Is it possible that this is also very important for testing `actor`? I've implemented something according to this https://kotlinlang.org/docs/reference/coroutines/shared-mutable-state-and-concurrency.html#actors and now getting:

```
java.lang.IllegalStateException: This job has not completed yet
at kotlinx.coroutines.JobSupport.getCompletionExceptionOrNull(JobSupport.kt:1131)
at kotlinx.coroutines.test.TestBuildersKt.runBlockingTest(TestBuilders.kt:53)
```

Is the `actor` pattern covered with tests?

kotlinx-coroutines-test/src/TestBuilders.kt
```diff
@@ -5,8 +5,88 @@
...
+
+/**
+ * A strategy for waiting on coroutines executed on other dispatchers inside a [runBlockingTest].
+ *
+ * Most tests should use [MultiDispatcherWaitConfig]. As an optimization, a test that executes coroutines only on a
+ * [TestCoroutineDispatcher] and never interacts with other dispatchers may use [SingleDispatcherWaitConfig].
+ *
+ * A test may subclass this to customize the wait in advanced cases.
+ */
+interface WaitConfig {
```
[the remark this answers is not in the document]
objcode: Agreed - this definitely seems too much to add for a simple parameter that can be addressed with documentation.

Do you have ideas for integrating with an arbitrary dispatcher? I'm not seeing an obvious place to do it but agreed it would be preferred.

kotlinx-coroutines-test/src/DelayController.kt
```diff
@@ -112,9 +113,75 @@ public interface DelayController {
      * Resumed dispatchers will automatically progress through all coroutines scheduled at the current time. To advance
      * time and execute coroutines scheduled in the future use, one of [advanceTimeBy],
      * or [advanceUntilIdle].
+     *
+     * When the dispatcher is resumed, all execution be immediate in the thread that triggered it. This means
+     * that the following code will not switch back from Dispatchers.IO after `withContext`
```
qwwdfsad: While we are here, could you please elaborate on why this behaviour is preferred over "classic" dispatching?
I am a bit concerned about how this behaviour is different from "release" builds

kotlinx-coroutines-test/src/DelayController.kt
```diff
@@ -112,9 +113,75 @@ public interface DelayController {
...
+     * ```
+     *
+     * For tests that need accurate threading behavior, [pauseDispatcher] will ensure that the following test dispatches
+     * on the correct thread.
+     *
+     * ```
+     * runBlockingTest {
+     *     pauseDispatcher()
+     *     withContext(Dispatchers.IO) { doIo() }
+     *     // runBlockingTest has returned to it's starting thread here
```
qwwdfsad: It's a bit counter-intuitive as well, `pauseDispatcher` does not pause the dispatcher, but changes its behaviour

kotlinx-coroutines-test/src/DelayController.kt
```diff
@@ -112,9 +113,75 @@ public interface DelayController {
...
+     * ```
      */
     @ExperimentalCoroutinesApi // Since 1.2.1, tentatively till 1.3.0
     public fun resumeDispatcher()
+
+    /**
+     * Represents the queue state of a DelayController.
+     *
+     * Tests do not normally need to use this API. It is exposed for advanced situations like integrating multiple
+     * [TestCoroutineDispatcher] instances or creating alternatives to [runBlockingtest].
```
qwwdfsad: runBlockingtest -> runBlockingTest

kotlinx-coroutines-test/src/DelayController.kt
```diff
@@ -112,9 +113,75 @@ public interface DelayController {
...
+     */
+    public sealed class QueueState {
```
qwwdfsad: It is for #1202, right?

If I was implementing this in some closed-world application, then I would write more or less the same solution, trying to abstract everything with interfaces and be as generic as possible.

But for the library, it is completely different. It should have as small as possible API surface, at the same time being elegant, approachable by novices and pragmatic (a lot of buzzwords here, sorry for that :))

I would rather remove this API and re-design it from scratch when we have a couple of real world problems that cannot be elegantly solved with existing API for tests

kotlinx-coroutines-test/src/TestBuilders.kt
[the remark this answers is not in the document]
qwwdfsad: I agree with Roman, this class seems to be a little bit overkill, integer constant should be enough.

qwwdfsad:
Minor notes: all `public` API should have explicit `public` modifier and `ExperimentalCoroutinesApi`

objcode:
Sorry delayed response (just got back to work after travel) - taking a look at comments above.

kotlinx-coroutines-test/src/DelayController.kt
objcode: Agreed  - I put it up for review before writing tests because I wasn't sure this was the right direction to take the API.

This logic can by making TestCoroutineDispatcher implement EventLoop which is probably the preferred approach.

This code will be backed out.

kotlinx-coroutines-test/src/DelayController.kt
objcode: So the main goal of this eager behavior was to make this test function testable without having to call runCurrent():

```
fun launchesCoroutine() {
    scope.launch {
        // do stuff
    }
}
```

In a test that doesn't use multiple threads (TestCoroutineDispatcher is used for all coroutines) it provides eager entry into the launch body (during dispatch the launch body is immediately executed).

As this is both the normal structure for this code (at least on Android) it is nice to avoid extra calls to `runCurrent()` in this test

```
@Test fun callLaunchesCoroutine() = runBlockingTest {
    // assume `scope` is using this TestCoroutineDispatcher
    launchesCoroutine()
    // coroutine has already launched here
}
```

This thread switching behavior is an undesired side effect of that API choice.

So basically, I think there are two options here:

1. Modify the behavior of `dispatch` to require a call to `runCurrent()` in this test. This does make common test cases require an extra call, but leads to correct threading behavior if the developer hasn't injected `TestCoroutineDispatcher` throughout their code under test.
2. Keep the behavior (prefer eager entry of launch blocks) with the assumption most tests will not execute coroutines on multiple threads.

WDYT?

kotlinx-coroutines-test/src/TestBuilders.kt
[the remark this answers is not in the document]
objcode: +1 to all of the above. After getting this feedback and having more time to think about it, I think this should be a simple timeout.

objcode:
Ok after reading all the feedback here's a rough plan for this PR:

1. I'll revise the current code to remove both of the new interfaces by switching to an EventLoop based implementation.
2. A decision needs to be made about the behavior of `TestCoroutineDispatcher.dispatch`.

The first will help us focus in on the API surface without the distractions of the extra classes/channel API.

The second seems important and should land before this library is considered stable.  I would like to advocate for keeping the current behavior since it makes single threaded testing easier (which is the recommended practice) but can see the complexity exposed by it in a multi-threading behavior change. I am curious if there is a way to unify both API goals? If so that would be the best approach.

elizarov:
Waiting for updates.

kotlinx-coroutines-test/src/DelayController.kt
qwwdfsad: Thanks for the explanation!

"Keep the behavior" part is, of course, more preferable as it really simplifies writing simple unit tests for end users. I've just realized that this is exactly why we thought to make `Uncofined` the default behaviour of `TestMainDispatcher`.

I think it's easier to change the docs and explain that test dispatcher acts like unconfined one (and maybe implementation should use it as well)

qwwdfsad:
Hi @objcode, do you have any update on this?
We don't mind finishing the work by ourselves if you are short on time to prevent this issue getting stuck

objcode:
Definitely - sorry delay had a few interruptions. Let me put together the event loop impl. tonight and then pass it over to you for finalization.

kotlinx-coroutines-test/src/TestBuilders.kt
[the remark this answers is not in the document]
objcode: Rewrote this to be a simple timeout.

kotlinx-coroutines-test/src/DelayController.kt
```diff
@@ -112,9 +113,75 @@ public interface DelayController {
...
+        }
+    }
+
+    /**
+     * A ConflatedBroadcastChannel that is up to date with the current [QueueState].
+     *
+     * Tests do not normally need to use this API. It is exposed for advanced situations like integrating multiple
+     * [TestCoroutineDispatcher] instances or creating alternatives to [runBlockingtest].
+     */
+    public val queueState: ConflatedBroadcastChannel<QueueState>
```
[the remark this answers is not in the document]
objcode: Reduced this to a single suspend function. The interface is exposed publicly to allow reimplementation but not on DelayController to avoid mixing interfaces.

kotlinx-coroutines-test/src/TestCoroutineDispatcher.kt
```diff
@@ -181,6 +200,25 @@ public class TestCoroutineDispatcher: CoroutineDispatcher(), Delay, DelayControl
...
+
+    private fun updateQueueObservers() {
+        // note: this will be called from any thread, and is safe lock-free in runBlockingTest but to guard against
+        // third party code code updating this will use a lock
+        synchronized(queue) {
+            val next = queue.peek()
+            when {
+                next == null-> queueState.offerIfChanged(Idle)
+                next.time <= currentTime -> queueState.offerIfChanged(HasCurrentTask)
+                next.time > currentTime -> queueState.offerIfChanged(HasDelayedTask)
```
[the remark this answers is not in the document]
objcode: +1

kotlinx-coroutines-test/src/DelayController.kt
objcode: Yea - this one I'm stuck on. It seems required to support the pauseDispatcher { } block in runBlockingTest but it has this side effect in the presence of multi-threading.

kotlinx-coroutines-test/src/DelayController.kt
objcode: Comment was deleted in refactor.

objcode:
@MariusBoepple The actor issue you're seeing above is likely because you're not terminating the actor before runBlockingTest completes. In the case of an infinite loop such as an actor, you'll need to cancel it and ensure it's complete before exiting runBlockingTest or it will detect the leaked coroutine.

kotlinx-coroutines-test/src/DelayController.kt
objcode: That's a pretty good way to explain it. I'll add that to the docs in this PR since it's become all about threading.

objcode:
@qwwdfsad @elizarov updated PR with requested changes. I was unable to find a working solution using EventLoop/unpark() since that's all implemented in EventLoopImplBase so I externalized parking via a small additional public interface https://github.com/Kotlin/kotlinx.coroutines/pull/1206/files#diff-e3c98820170c8b47a0aad1428e940c2cR255

PTAL!

winfoozmnayef:
Any update on this pull request?

qwwdfsad:
Moving it to `coroutines-test` branch, will open PR soon

anudeep3998:
Any updates on the release ?

mdenburger:
Where did the moved PR end up? This issue is closed, but kotlin-coroutines-test 1.3.8 still has this problem.

ymaniz09:
1.4.0-M1 still has this issue

totemcaf:
It seems that 1.4.0 still has this issue

zivkesten:
kotlinx-coroutines-test:1.4.1 still has this issue

Purfakt:
kotlinx-coroutines-test:1.4.2 still has this issue

ivanbartsov:
@qwwdfsad
The issue seems to still be valid with `kotlinx-coroutines-test:1.4.3`

>Moving it to coroutines-test branch, will open PR soon

Can I ask you to link the succeeding PR (if any) here for future references? A couple of open issues (#1204 #1222) lead here and this PR kinda makes an impression it's been closed and abandoned for good.

PR #1207, opened 2019-05-20
Fix overflow bug in Flow.drop

EdwarDDay:
Fixes [KT-31477](https://youtrack.jetbrains.com/issue/KT-31477)

The test runs very long, therefore I only added it to the `jvmTest` source set.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
@@ -23,7 +23,7 @@ public fun <T> Flow<T>.drop(count: Int): Flow<T> {
     return flow {
         var skipped = 0
         collect { value ->
-            if (++skipped > count) emit(value)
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
```diff
@@ -23,7 +23,7 @@ public fun <T> Flow<T>.drop(count: Int): Flow<T> {
     return flow {
         var skipped = 0
         collect { value ->
-            if (++skipped > count) emit(value)
+            if (skipped >= count) emit(value) else ++skipped
```
fvasco: `skipped` cannot become greater than `count`

elizarov: That's a good engineering practice to write `>=` in this case anyway, even though writing `==` would technically suffice.

qwwdfsad:
Merged manually, thanks for the report and the fix!

Issue #48, opened 2017-03-30
wrapping callbacks that fire more than once


beyondeye:
I am now trying to convert some Firebase code on Android to use coroutines. some Firebase callbacks are a bit weird in that they sometimes trigger more than once. For example for getting the download url from firebase storage I have written this:
```kotlin
    private suspend fun getFileDownloadUrlAsync(file_sref: StorageReference): Uri = suspendCoroutine { c ->
        with(file_sref.downloadUrl) { //<---- call to Firebase getDownloadUrl()
            addOnCompleteListener { //<--- this callback can trigger more than once!
                if (it.isSuccessful) {
                    c.resume(it.result)
                } else
                    c.resumeWithException(Exception("some error"))
            }
            addOnFailureListener {
                c.resumeWithException(it)
            }
        }
    }

```
Since the completion listener is triggered more than once I get an IllegalStateException("Already resumed")
As a work-around I have defined 
```kotlin
    class WrappedContinuation<T>(val c: Continuation<T>) : Continuation<T> {
        var isResolved = false
        override val context: CoroutineContext
            get() = c.context

        override fun resume(value: T) {
            if (!isResolved) {
                isResolved = true
                c.resume(value)
            }
        }

        override fun resumeWithException(exception: Throwable) {
            if (!isResolved) {
                isResolved = true
                c.resumeWithException(exception)
            }
        }

    }

    public inline suspend fun <T> suspendCoroutineW(crossinline block: (WrappedContinuation<T>) -> Unit): T =
            suspendCoroutine { c ->
                val wd = WrappedContinuation(c)
                block(wd)
            }
```
and I am using ```suspendCoroutineW``` instead of ```suspendCoroutine```. Would it be possible to 
modify ```SafeContinuation``` class in the library to offer similar functionality, since it already keeps the information about if the continuation has already been resolved or not, and expose it to the user?

fvasco:
After first "complete event" the completion listener is unuseful therefore you should remove it.

IMHO this isn't a Continuation issue.

beyondeye:
Unfortunately I could not find a way to remove the listener.

beyondeye:
ok so I find the solution: I should have used [suspendCancellableCoroutine](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.experimental/suspend-cancellable-coroutine.html)

elizarov:
[...]

PR #1211, opened 2019-05-20
Fix Tasks await extension cancellation support

### LINKED ISSUE #1197: Completion listener from Play Services integration can leak
The two lines of code shown below highlight the problem well:
`suspendCancellableCoroutine` is used, but the wrapped callback (lambda passed to `addOnCompleteListener`)  stays registered and still references the continuation after cancellation, on which it can illegally call ` resume` (which crashes the app).
https://github.com/Kotlin/kotlinx.coroutines/blob/7af7ede7a06ae1a644c17dbb2ca12172af316ec7/integration/kotlinx-coroutines-play-services/src/Tasks.kt#L96-L97

Unfortunately, the [Google Tasks API](https://developers.google.com/android/reference/com/google/android/gms/tasks/Task) doesn't provide a way to remove the listener manually, so a workaround is needed to ignore post-cancellation completion without crashing, or make it no longer cancellable using `suspendCoroutine` (which can be a bummer considering cancellation ability can often be relied on).

The simplest workaround is to wrap `resume` call with `runCatching`, or surround it with `try`/`catch`, ignoring `IlegalStateException`s.

I can make a PR once a decision is taken.

LouisCAD:
That fixes previous usage of `suspendCancellableCoroutine` that would lead to
crashes if cancelled and then the `addOnCompleteListener` callback was invoked.

The Task itself is not cancellable from the public API, but the continuation is put in a
wrapper where it is removed back in the finally block (setting it to null), so subsequent
calls to the `addOnCompleteListener` callback are ignored by returning early.

There is an extra try catch inside the `addOnCompleteListener` callback in case of multithread
usage, to avoid crashing the app if cancellation takes place in a thread and `addOnCompleteListener`
is invoked while still seing the wrapped continuation and resuming it a second time.

Fixes #1197, and relates to #48.

### DIFF
--- a/integration/kotlinx-coroutines-play-services/src/Tasks.kt
+++ b/integration/kotlinx-coroutines-play-services/src/Tasks.kt
@@ -77,31 +77,38 @@ public fun <T> Task<T>.asDeferred(): Deferred<T> {
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * stops waiting for the completion stage and immediately resumes with [CancellationException].
  */
-public suspend fun <T> Task<T>.await(): T {
-    // fast path
+public suspend fun <R> Task<R>.await(): R {
     if (isComplete) {
-        val e = exception
-        return if (e == null) {
-            if (isCanceled) {
+        when (val e = exception) {
+            null -> if (isCanceled) {
                 throw CancellationException("Task $this was cancelled normally.")
             } else {
-                @Suppress("UNCHECKED_CAST")
-                result as T
+                @Suppress("UNCHECKED_CAST") return result as R
             }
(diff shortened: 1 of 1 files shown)


qwwdfsad:
>That fixes previous usage of suspendCancellableCoroutine that would lead to
crashes if cancelled and then the addOnCompleteListener callback was invoked.

This is not the case. `CancellableContinuation` is designed to avoid this exact situation (a race between cancellation and `resume`) and ignores the first `resume` after cancellation.
The only way to trigger `ISE` is to invoke `resume` *twice* which should not happen for `OnCompleteListener`.

Could you please provide a reproducer? We should understand the problem before fixing it

PR #1214, opened 2019-05-21
New flow builder: channelFlow (and its alias callbackFlow) and supple…

### LINKED ISSUE #1210: Flow context preserving property, thread safety and context changing
### Context preservation 
`flow` has the [context preservation](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/index.html) property that is supposed to make programming with flows more predictable, less error-prone and simpler.

For example, the following code will produce ISE on the first attempt to launch it:
```
flow {
    withContext(MyCustomContext) { emit(42) }
}
```

This is the desired property, but it is not enough to provide the best possible experience for developers.

### Problem #⁠1
For example (and I've seen this in the wild), one may want to write the following `merge` implementation:
```
fun Flow<T>.merge(other: Flow<T>): Flow<T> {
    coroutineScope {
        launch {
            collect { emit(value) }
        }
        other.collect { emit(value) }
    }
}
```

it passes context preservation check, it works under the single-threaded dispatcher (and on Android most of the tests are single-threaded), but it produces the least expected error (probably `KotlinNullPointerException`) when running concurrently.
Of course, it is possible to make `flow` thread-safe, but it significantly reduces performance, even when no actual parallelism is present. The other option is to do a probabilistic data-race check, but this won't be sufficient because of its nature. It is impossible to catch a data-race when no real data race occurs, so such check will always be too late. 

### Problem #⁠2
Originally reported by @gildor 

Consider the following system-wide function to work with any resources:
`fun foo(...parameters, block: suspend (resource) -> Unit)` that uses `withContext` internally.
The natural desire is to write the following snippet:
```
flow {
    foo(...) { resource ->
        emit(transform(resource))
    }
}
```
And it doesn't work because it violates context preservation. And one cannot simply use `flowOn` because `foo` **encapsulates** its context.

### Solution

**Strengthen context preservation property.** Prohibit not only context element changes, but also situations when emission happens from the different coroutines (efficiently prohibiting situation from #⁠1).
Formulated more strictly, the rule sounds like `'collect' and 'emit' context should have transitively the same non-scoped Job in the hierarchy`.
Sounds complicated, but this won't leak to the users 99% of the time.

But to allow writing such simple `merge` operator and to resolve #⁠2, the additional builder is required: `channelFlow` (and this builder will be mentioned in the text of exception when context preservation will be violated).

The full signature: `fun channelFlow(capacity: Int = DEFAULT, block: suspend ProducerScope.() -> Unit)`. To avoid having two "identical" builders, `flowViaChannel` will probably be deprecated with replacement.

The only change required to write concurrent or context-ignorant builder is replacing `emit` with `send`.

It was decided to provide a channel-based builder instead of thread-safe one for the sake of consistent mental model:
 * `FlowCollector` **is not** thread-safe and never can be used concurrently
 * Channels are thread-safe and can be used in the fan-in fashion
 * The same builder can be used to both concurrent primitives and integrations with callback-based APIs

The only serious downside is Java interoperability: `block` is suspending and thus cannot be easily invoked from Java

qwwdfsad:
…mentary ProducerScope.await method

Rationale:

    * Can be used in different context without breaking context preservation
    * Can be used to build concurrent operators such as merge
    * Can be used to integrate with callbacks
    * Is less error-prone than flowViaChannel because requires explicit await() call

Partially fixes #1210


kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -25,6 +25,32 @@ public interface ProducerScope<in E> : CoroutineScope, SendChannel<E> {
...
+ * Example of usage:
+ * ```
+ * val callbackEventsStream = produce {
+ *     val disposable = registerChannelInCallback(channel)
+ *     await { disposable.dispose() }
+ * }
+ * ```
+ */
+@ExperimentalCoroutinesApi
+public suspend fun <T> ProducerScope<T>.await(block: () -> Unit = {}) {
```
elizarov: Design: It should be called `awaitClose` (consistently with `invokeOnClose`) or otherwise it is not clear what is being awaited when reading `await { disposable.dispose() }` code.

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -25,6 +25,32 @@ public interface ProducerScope<in E> : CoroutineScope, SendChannel<E> {
     val channel: SendChannel<E>
 }
 
+/**
+ * Suspends the current coroutine until the channel is either [closed][SendChannel.close] or [cancelled][ReceiveChannel.cancel]
+ * and invokes the given [block] before resuming the coroutine.
```
elizarov: Additional documentation would be helpful here: 

"Note that when producer channel is cancelled this function resumes with cancellation exception, so putting the code after call this function would not lead to its execution in case of cancellation. That is why this code takes a lambda parameter."

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -25,6 +25,32 @@ public interface ProducerScope<in E> : CoroutineScope, SendChannel<E> {
...
+ * }
+ * ```
+ */
+@ExperimentalCoroutinesApi
+public suspend fun <T> ProducerScope<T>.await(block: () -> Unit = {}) {
+    check(kotlin.coroutines.coroutineContext[Job] === this) { "await() can be invoke only from the producer context" }
+    suspendCancellableCoroutine<Unit> { cont ->
+        invokeOnClose {
+            try {
+                block()
```
elizarov: There is no guarantee for exectuion context of the `block` which might lead to nasty surprives when trying to "undo" something in UI on cancellation of UI-confined producer. I'd suggest the following reimplementation:

```
    try {
        suspendCancellableCoroutine<Unit> { cont ->
            invokeOnClose {
                cont.resume(Unit)
            }
        }
    } finally {
        block()
    }
```

kotlinx-coroutines-core/common/src/flow/Builders.kt
```diff
@@ -237,3 +213,88 @@ public fun <T> flowViaChannel(
...
+ *
+ * Example of usage:
+ * ```
+ * fun flowFrom(api: CallbackBasedApi): Flow<T> = callbackFlow {
+ *     val callback = object : Callback { // implementation of some callback interface
+ *         override fun onNextValue(value: T) {
+ *             offer(value) // Note: offer drops value when buffer is full
+ *         }
+ *         override fun onApiError(cause: Throwable) {
+ *             cancel("API Error", CancellationException(cause))
```
elizarov: There's no two-parameter `cancel`. Should be:
```
cancel(CancellationException("API Error", cause))
```

kotlinx-coroutines-core/common/src/flow/Builders.kt
```diff
@@ -237,3 +213,88 @@ public fun <T> flowViaChannel(
...
+ * To control backpressure, [bufferSize] is used and matches directly the `capacity` parameter of [Channel] factory.
+ * The provided channel can later be used by any external service to communicate with the flow and its buffer determines
+ * backpressure buffer size or its behaviour (e.g. in the case when [Channel.CONFLATED] was used).
+ *
+ * Example of usage:
+ * ```
+ * fun flowFrom(api: CallbackBasedApi): Flow<T> = callbackFlow {
+ *     val callback = object : Callback { // implementation of some callback interface
+ *         override fun onNextValue(value: T) {
+ *             offer(value) // Note: offer drops value when buffer is full
```
elizarov: Should probably mention `Channel.UNLIMITED` capacity parameter somewhere around here.

kotlinx-coroutines-core/common/src/channels/Produce.kt
qwwdfsad: Also, note that it is an extension on `ProducerScope` rather than `SendChannel`.

PR #1215, opened 2019-05-21
Retries cancellation

qwwdfsad:


kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -67,13 +71,19 @@ public fun <T> Flow<T>.retry(
...
-                if (fromDownstream) throw e
+                if (fromDownstream || e.isCancellationCause()) throw e
                 if (!predicate(e) || retries-- == 0) throw e
             }
         }
     }
 }
 
+private suspend fun Throwable.isCancellationCause(): Boolean {
+    val job = coroutineContext[Job]
```
elizarov: Make `coroutineContext` a parameter to this function and drop `suspend` modifier from it to save suspension point in callers of this function.

PR #1220, opened 2019-05-23
Consolidate NullSurrogate with Symbol

elizarov:
* JvmStatic on unbox did not make sense anyway
* unbox is useful for other symbols

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/internal/NullSurrogate.kt
+++ b/kotlinx-coroutines-core/common/src/flow/internal/NullSurrogate.kt
@@ -4,11 +4,13 @@
 
 package kotlinx.coroutines.flow.internal
 
+import kotlinx.coroutines.internal.*
 import kotlin.jvm.*
 
-internal object NullSurrogate {
-
-    @JvmStatic
-    @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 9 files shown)


kotlinx-coroutines-core/common/src/flow/internal/NullSurrogate.kt
```diff
@@ -4,11 +4,11 @@
...
 import kotlin.jvm.*
 
-internal object NullSurrogate {
-
-    @JvmStatic
-    @Suppress("UNCHECKED_CAST")
-    internal fun <T> unbox(value: Any?): T = if (value === NullSurrogate) null as T else value as T
-}
+// Note: it is conceptually the same as AbstractChannel.NULL_VALUE
+// todo: consolidate, move this constant to "common" kotlinx.coroutines.internal, rename consistently with other symbols
```
qwwdfsad: About todo, there is always a chance to accidentally mix two unrelated symbols (as we did with `Incompleted`), so I'd prefer it so stay into its own package. Renaming it to `NULL_VALUE` is reasonable

elizarov:
I've renamed it to `NULL` to avoid named conflicts.

Issue #1128, opened 2019-04-21
Scoped flow leaks the decomposition into behaviour


qwwdfsad:
Consider the following operators:
```
fun <T> Flow<T>.id(): Flow<T> = flow { collect { emit(it) } }

fun <T> Flow<T>.idScoped(): Flow<T> = flow {
    coroutineScope {
        val channel = Channel<T>()
        launch {
            try {
                for (i in channel) emit(i)
            } finally {
                channel.cancel()
            }
        }

        val result = runCatching {
            collect {
                channel.send(it)
            }
        }
        channel.close(result.exceptionOrNull())
        result.getOrThrow()
    }
}
```

Our goal, eventually, is to make them more or less indistinguishable (modulo dispatch order, of course).

Currently, it is not the case, this behavior leaks directly into user code:
```
flow {
    emit(1)
    delay(Long.MAX_VALUE)
}.id[Scoped]().collect { value ->
    kotlin.coroutines.coroutineContext.cancel()
}
```

Depending on the chosen operator, behaviour will be different (CE or hang).
To fix it, `invokeOnCompletion { if (it is CancellationException) cancel() }` should be used right after `launch` in a scoped variant.

While this can problem can be workarounded, it clearly indicates that we are missing an important primitive or cancellation mode.

elizarov:
It seems that we already have a corresponding primitive and it is called `produce`, and a correct implementation of `idScoped` seems to be the following: 

```
fun <T> Flow<T>.idScoped(): Flow<T> = flow {
    coroutineScope {
        val channel = produce<T> {
            collect {
                send(it)
            }
        }
        launch {
            try {
                for (i in channel) emit(i)
            } finally {
                channel.cancel()
            }
        }
    }
}
```

qwwdfsad:
I am not sure it is correct (I deliberately avoided using `produce` here).

With `produce` the problem just hides a bit better, e.g.
```
 flow {
    emit(1)
    kotlin.coroutines.coroutineContext.cancel()
}.idScoped().collect { value ->
    println(value)
}
```

still behaves differently depending on the implementation of `idScoped`

elizarov:
[...]

PR #1227, opened 2019-05-28
Flow scope

### LINKED ISSUE #1218: Wrong Flow.switchMap { }.take() behaviour
Kotlin 1.3.31, kotlinx.coroutines 1.2.1

This code:
```kotlin
runBlocking {
  println(
    flowOf(1, 2, 3, 4, 5)
      .switchMap { x -> flow { emit(x) } }
      .take(1)
      .toList()
  )
}
```
prints wrong result: [1, 2, 3, 4, 5].

It should prints [1].

qwwdfsad:
Fixes #1218 and #1128


kotlinx-coroutines-core/common/test/flow/operators/FlowOnTest.kt
```diff
@@ -6,6 +6,7 @@ package kotlinx.coroutines.flow
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
+import kotlin.math.*
```
qwwdfsad: unused import

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -319,6 +319,31 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         cancelParent(cause) // tentative cancellation -- does not matter if there is no parent
     }
 
+    /**
+     * The method that is invoked when the job is cancelled to possible propagate cancellation to the parent.
+     * Returns `true` if the parent is responsible for handling the exception, `false` otherwise.
+     *
+     * Invariant: never returns `true` for instances of [CancellationException], otherwise such exception
```
elizarov: Wrong doc: Never returns `false` for instance of CancellationException.

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -594,17 +619,27 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
     }
 
-    // Child was cancelled with cause
-    // It is overridden in supervisor implementations to ignore child cancellation
-    public open fun childCancelled(cause: Throwable): Boolean =
-        cancelImpl(cause) && handlesException
+    /**
+     * Returns `true` if job should cancel itself on child [CancellationException].
+     */
+    public open fun cancelOnChildCancellation(cause: CancellationException) = false
```
elizarov: It seems wasteful to have a separate overridable function only for the use by flow scopes. I suggest to have them override `childCancelled` just supervisor jobs do.

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -594,17 +619,27 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
+    /**
+     * Child was cancelled with a cause.
+     * In this method parent decides whether it cancels itself (e.g. on a critical failure) and
+     * whether it handles the exception of the child.
+     * It is overridden in supervisor implementations to completely ignore any child cancellation.
+     */
+    public open fun childCancelled(cause: Throwable): Boolean {
+        if (cause is CancellationException && !cancelOnChildCancellation(cause)) return true
+        return cancelImpl(cause) && handlesException
+    }
```
elizarov: I suggest to replace it with the following code to make "default" cancellation logic clearer and to fix a bug with `produce` coroutine (failure of a child shall immediately cancel the channel because it is doomed to fail and should not wait for main coroutine's failure):
```
    public open fun childCancelled(cause: Throwable): Boolean {
        if (cause is CancellationException) return true
        return cancelInternal(cause)
    }
```

kotlinx-coroutines-core/common/src/flow/internal/FlowScope.kt
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+internal class FlowScope<T, R>(
+    private val collector: FlowCollector<R>,
+    context: CoroutineContext,
+    uCont: Continuation<T>
+) : ScopeCoroutine<T>(context, uCont), FlowCollector<R> {
+
+    public override fun cancelOnChildCancellation(cause: CancellationException) = cause !is ChildCancelledException
```
elizarov: It should override `childCancelled` for consistency with supervisor scopes (that override it to return `false`) and to make the logic more "straight" (without double negations). Use the following code in overridden implementation:

```
    override fun childCancelled(cause: Throwable): Boolean {
        if (cause is ChildCancelledException) return true
        return cancelInternal(cause)
    }
```

kotlinx-coroutines-core/common/src/flow/internal/FlowScope.kt
```diff
@@ -0,0 +1,58 @@
...
+ *     launch {
+ *         throw CancellationException()
+ *     }
+ * } // <- CE will be rethrown here
+ * ```
+ *
+ * Basically, it is shorthand for:
+ * ```
+ * flow {
+ *     coroutineScope {
```
elizarov: I'll need a separate `flowScope` primitive, too, for other operators I'm working on, so it makes sense to have a separate internal `flowScope` primitive too and say here that `scopedFlow { ... }` is a shortcut for `flow { flowScope { ... } }` that sames on lambdas (one lambda instead of two)

kotlinx-coroutines-core/common/src/flow/internal/FlowScope.kt
```diff
@@ -0,0 +1,58 @@
...
+ *         ...
+ *     }
+ * }
+ * ```
+ * with additional constraint on cancellation.
+ * To cancel child without cancelling itself, `cancel(ChildCancelledException())` should be used.
+ */
+internal fun <R> scopedFlow(@BuilderInference block: suspend FlowScope<Unit, R>.() -> Unit): Flow<R> =
```
elizarov: More correctly it should be called `unsafeScopedFlow`. I would also consider an alternative signature of this function passing collector as value parameter, so that there is no need to go through delegated call on each emit:

```
fun <T> unsafeScopedFlow(@BuilderInference block: suspend CoroutineScope.(FlowCollector<T>) -> Unit): Flow<T> = 
    unsafeFlow { 
        val collector = this
        flowScope { block(collector) }
    }
```

qwwdfsad: I'd prefer to avoid "unsafe" in operators implementation

kotlinx-coroutines-core/common/src/flow/internal/FlowCoroutine.kt
```diff
@@ -0,0 +1,61 @@
...
+ * For example:
+ * ```
+ * flowScope {
+ *     launch {
+ *         throw CancellationException()
+ *     }
+ * } // <- CE will be rethrown here
+ * ```
+ */
+internal suspend fun <R> flowScope(@BuilderInference block: suspend CoroutineScope.() -> R): R =
```
qwwdfsad: An interesting question: do we want to guarantee the same behaviour for `channelFlow`?
For example:
```
fun Flow<Int>.bufferWithTimeout(): Flow<Int> = channelFlow {
    launch {
        delay(Long.MAX_VALUE)
    }
    launch {
        collect {
            withTimeout(-1) {
                send(it)
            }
        }
    }
}
```

do we expect this flow to ever complete?

kotlinx-coroutines-core/common/src/flow/internal/FlowCoroutine.kt
```diff
@@ -0,0 +1,82 @@
...
+private class FlowCoroutine<T>(context: CoroutineContext, uCont: Continuation<T>) :
+    ScopeCoroutine<T>(context, uCont) {
+
+    public override fun childCancelled(cause: Throwable): Boolean {
+        if (cause is ChildCancelledException) return true
+        return cancelImpl(cause)
+    }
+}
+
+private class FlowProduceCoroutine<T>(parentContext: CoroutineContext, channel: Channel<T>) :
```
qwwdfsad: indent

PR #1228, opened 2019-05-28
Use nanosleep in runBlocking's delay on Native

### LINKED ISSUE #1225: Kotlin/Native runBlocking & delay cause high CPU usage
Hi all,

I just started a very simple hello world console application with Kotlin/Native for a linuxX64 target.
I want to increment a counter and print it every two seconds to the command line:
```kotlin
fun main() = runBlocking {
    var i = 0
    while(isActive){
        println("Hello Kotlin/Native COROUTINE #${++i}")
        delay(2000)
    }
}
```

When I run the binary on a the target host I encounter that CPU usage increases by 25% (one core being totally occupied). So it seems this is blocking the thread totally.

When I replace the `delay(...)` suspend function with a `platform.posix.sleep(2)` there is no noticeable CPU usage increase (0%).

here my _build.gradle.kts_:
```gradle
plugins {
    kotlin("multiplatform") version "1.3.30"
}

repositories {
    mavenCentral()
}


kotlin {
    linuxX64("native") {
        binaries {
            executable()
        }
    }

    sourceSets {
        val nativeMain by getting {
            dependencies {
                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-native:1.2.1")
            }
        }

    }
}

tasks.withType<Wrapper> {
  gradleVersion = "5.3.1"
  distributionType = Wrapper.DistributionType.ALL
}
```

Is this a known issue ? or is just me being new to Kotlin/Native and coroutines... (bear with me if that's the case)

Many thanks for your support and the great efforts you put into bringing Kotlin coroutines to every platform !

Best regards,
Thomas

qwwdfsad:
Fixes #1225

### DIFF
--- a/kotlinx-coroutines-core/native/src/Builders.kt
+++ b/kotlinx-coroutines-core/native/src/Builders.kt
@@ -66,7 +66,8 @@ private class BlockingCoroutine<T>(
                 val parkNanos = eventLoop?.processNextEvent() ?: Long.MAX_VALUE
                 // note: process next even may loose unpark flag, so check if completed before parking
                 if (isCompleted) break
-                timespec.tv_nsec = parkNanos.convert()
+                timespec.tv_sec = parkNanos / 1000000000L // 1e9 ns -> sec
+                timespec.tv_nsec = (parkNanos % 1000000000L).convert() // % 1e9
                 nanosleep(timespec.ptr, null)
             }
         } finally { // paranoia
--- a/kotlinx-coroutines-core/native/src/Builders.kt
+++ b/kotlinx-coroutines-core/native/src/Builders.kt
@@ -4,6 +4,8 @@
 
 package kotlinx.coroutines
 
+import kotlinx.cinterop.*
+import platform.posix.*
 import kotlin.coroutines.*
 
 /**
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/native/src/Builders.kt
```diff
@@ -56,21 +58,23 @@ private class BlockingCoroutine<T>(
...
+    fun joinBlocking(): T = memScoped {
         try {
             eventLoop?.incrementUseCount()
+            val timespec = alloc<timespec>()
             while (true) {
                 val parkNanos = eventLoop?.processNextEvent() ?: Long.MAX_VALUE
                 // note: process next even may loose unpark flag, so check if completed before parking
                 if (isCompleted) break
-                // todo: LockSupport.parkNanos(this, parkNanos)
+                timespec.tv_nsec = parkNanos.convert()
```
thowimmer: tv_nsec should not be greater then 999999999.
Therefore you should set tv_sec as well right ?
```kotlin
timespec.tv_sec = parkNanos / 1000000000L
timespec.tv_nsec = (parkNanos % 1000000000L).convert()
```

qwwdfsad: good catch, thanks!

PR #1229, opened 2019-05-28
Documentation

qwwdfsad:


kotlinx-coroutines-core/README.md
```diff
@@ -84,14 +84,6 @@ Select expression to perform multiple suspending operations simultaneously until
 
 Low-level primitives for finer-grained control of coroutines.
 
-# Package kotlinx.coroutines.timeunit
-
-Optional time unit support for multiplatform projects.
-
-# Package kotlinx.coroutines.test
```
elizarov: This leaves `kotlinx.coroutines.test` package without documentation on this module overview page: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/ 
We should either leave it here (adding that it is depracated) or suporess all public API from this package, so it does not appear in the list of packages.

PR #1232, opened 2019-05-30
Enable R8 optimization of Dispatchers.Main loading

### LINKED ISSUE #1231: FastServiceLoader performs I/O on calling thread and prevents R8 optimization
The fix to #878 that introduces a FastServiceLoader to mitigate the fact that ServiceLoader implementation on Android is slow is still inadequate for Android apps and moreover, the solution prevents the [R8 optimization](https://issuetracker.google.com/issues/120436373) from working.

For Android apps to avoid doing I/O on the main thread, we should
1) let the user disable the FastServiceLoader (this can be done now via setting a systemProp now, but is quite hidden and requires action from the developer)
2) Enable the R8 optimization by making sure that any calls to ServiceLoader.load are of the form:
`ServiceLoader.load(MyClass::class.java, MyClass::class.java.classLoader).iterator()`
Everything matters here: 2 argument constructor, class constants as arguments, and the only call to the returned ServiceLoader has to be `.iterator()`.

wojtek-kalicinski:
The developer still has to disable the FastServiceLoader manually
if they're using R8.

R8 is then able to optimize away the ServiceLoader and reflection
entirely, resulting in direct class instantiation and no extra I/O
on calling thread.

Fixes #1231

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/CoroutineExceptionHandlerImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/CoroutineExceptionHandlerImpl.kt
@@ -13,10 +13,15 @@ import kotlin.coroutines.*
  * Note that Android may have dummy [Thread.contextClassLoader] which is used by one-argument [ServiceLoader.load] function,
  * see (https://stackoverflow.com/questions/13407006/android-class-loader-may-fail-for-processes-that-host-multiple-applications).
  * So here we explicitly use two-argument `load` with a class-loader of [CoroutineExceptionHandler] class.
+ *
+ * We are explicitly using the `ServiceLoader.load(MyClass::class.java, MyClass::class.java.classLoader).iterator()`
+ * form of the ServiceLoader call to enable R8 optimization when compiled on Android.
  */
-private val handlers: List<CoroutineExceptionHandler> = CoroutineExceptionHandler::class.java.let { serviceClass ->
-    ServiceLoader.load(serviceClass, serviceClass.classLoader).toList()
-}
+private val handlers: List<CoroutineExceptionHandler> = ServiceLoader.load(
+        CoroutineExceptionHandler::class.java,
+        CoroutineExceptionHandler::class.java.classLoader
+).iterator().asSequence().toList()
+
 
 internal actual fun handleCoroutineExceptionImpl(context: CoroutineContext, exception: Throwable) {
     // use additional extension handlers
--- a/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
@@ -7,11 +7,6 @@ import java.util.jar.*
 import java.util.zip.*
 
 /**
- * Name of the boolean property that enables using of [FastServiceLoader].
- */
-private const val FAST_SERVICE_LOADER_PROPERTY_NAME = "kotlinx.coroutines.fast.service.loader"
-
-/**
  * A simplified version of [ServiceLoader].
  * FastServiceLoader locates and instantiates all service providers named in configuration
  * files placed in the resource directory <tt>META-INF/services</tt>.
@@ -25,12 +20,7 @@ private const val FAST_SERVICE_LOADER_PROPERTY_NAME = "kotlinx.coroutines.fast.s
 internal object FastServiceLoader {
     private const val PREFIX: String = "META-INF/services/"
 
-    private val FAST_SERVICE_LOADER_ENABLED = systemProp(FAST_SERVICE_LOADER_PROPERTY_NAME, true)
-
     internal fun <S> load(service: Class<S>, loader: ClassLoader): List<S> {
-        if (!FAST_SERVICE_LOADER_ENABLED) {
-            return ServiceLoader.load(service, loader).toList()
-        }
         return try {
             loadProviders(service, loader)
         } catch (e: Throwable) {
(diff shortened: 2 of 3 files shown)


LouisCAD:
@wojtek-kalicinski Would it possible for the R8 team to disable `FastServiceLoader` so we don't have to add an extra configuration step for all the projects in the world (risking people forgetting it)?

wojtek-kalicinski:
I'm looking into ways we can do it, I agree that having a manual step that developers have to remember about does not solve the issue fully, but I just want to unblock the R8 optimization first. 
I'll follow up on this if I have a solution.

ansman:
@wojtek-kalicinski Not sure if out of scope for this PR but `CoroutineExceptionHandlerImpl.kt` also uses `ServiceLoader.load` to load the exception handler and I'm guessing that code would need to be updated as well.

ansman:
> I'm looking into ways we can do it, I agree that having a manual step that developers have to remember about does not solve the issue fully, but I just want to unblock the R8 optimization first.
> I'll follow up on this if I have a solution.

Couldn't the default value just be `false` since the `FastServiceLoader` is only really needed on Android? JAR verification seems like a useful thing on the server side so it would make sense to hade it use the regular service loader by default.

JakeWharton:
I'm working on a test for this which actually invokes R8 and verifies that all ServiceLoader calls are removed from the final dex to prevent regression or new occurrences in the wrong form. Feel free to merge without it though, as it might be a few days.

LouisCAD:
I agree, turning it off by default would make sense. Then, we might get AGP
turn it on by default when R8 is not enabled (i.e. minifyEnabled = false),
or if not, we might still do it ourselves, which is less of an issue.

kotlinx-coroutines-core/jvm/src/CoroutineExceptionHandlerImpl.kt
```diff
@@ -13,10 +13,15 @@ import kotlin.coroutines.*
...
+ * We are explicitly using the `ServiceLoader.load(MyClass::class.java, MyClass::class.java.classLoader).iterator()`
+ * form of the ServiceLoader call to enable R8 optimization when compiled on Android.
  */
-private val handlers: List<CoroutineExceptionHandler> = CoroutineExceptionHandler::class.java.let { serviceClass ->
-    ServiceLoader.load(serviceClass, serviceClass.classLoader).toList()
-}
+private val handlers: List<CoroutineExceptionHandler> = ServiceLoader.load(
+        CoroutineExceptionHandler::class.java,
+        CoroutineExceptionHandler::class.java.classLoader
+).iterator().asSequence().toList()
```
qwwdfsad: Just out of curiosity, could you please explain why `ServiceLoader.load(...).iterator()...` enables optimization, while `ServiceLoader.load(...).anythingElse()` does not?

wojtek-kalicinski:
I would argue that we should not disable FastServiceLoader by default just yet, and discuss this further.
We have two problems here:
1) we want apps that use R8 to be able to get the optimization - this is what my PR unblocks
2) a lot (A LOT) of apps are not using R8 or ProGuard (they have minification disabled). FastServiceLoader is actually a desireable optimization for those apps and it would be a shame if they lost it

It's difficult to come up with a solution that works for both groups, but I have some ideas that I'd like to further discuss. I'll follow up in a bit.

kotlinx-coroutines-core/jvm/src/CoroutineExceptionHandlerImpl.kt
wojtek-kalicinski: I wasn't involved in the design of the R8 optimization, but looking at the source code of R8, these are the necessary conditions. My guess is because they're actually replacing the ServiceLoader entirely with direct instantiation and putting the loaded services into a Collection (if I remember correctly), they want to make sure that only code which accesses the Iterator (which both the Collection and ServiceLoader implement) is safe to modify in this way. 

If an app did something else like `ServiceLoader.getClass()` (for whatever reason), then there would be a mismatch between the optimized and non-optimized version.

wojtek-kalicinski:
> Could you please specify since which R8 version is this optimization present?
Also to add to my previous comment, if I understand the R8 releases correctly, this optimization will be present in future R8 releases from the 1.5.X branch, so this further reinforces why we should not disable the fast loader for everyone just yet.

kotlinx-coroutines-core/jvm/src/CoroutineExceptionHandlerImpl.kt
JakeWharton: Yeah it was meant to handle this Java pattern:
```java
for (Service service : ServiceLoader.loader(Service.class, Service.class.getClassLoader()) {
  // ...
}
```
Unfortunately Kotlin's nice extensions like `toList()` have the `ServiceLoader` passed as an instance to a function rather than inlining `iterator()` to the call site. `.forEach` and friends would still work though.

We _might_ be able to relax the optimization where it sees that the `ServiceLoader` flows into a function as an `Iterable`. Although, as @wojtek-kalicinski states, someone could still be doing `getClass()` or `toString()` or, for some horrific reason, `instanceof ServiceLoader`.

Tolriq:
Out of curiosity is there a proper way to set system properties on Android?

Or it's to be done in code as early as possible in application.OnCreate?

qwwdfsad:
@wojtek-kalicinski good points, I'll merge it as is in that case.

Please let us know when R8 will be released so we could update our documentation

PR #1236, opened 2019-05-31
More operators

qwwdfsad:


kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
```diff
@@ -97,3 +99,44 @@ public fun <T> Flow<T>.onEach(action: suspend (T) -> Unit): Flow<T> = flow {
         emit(value)
     }
 }
+
+/**
+ * Reduces the given flow with [operation], emitting every intermediate result, including initial value.
+ * The first element is takes as initial value for operation accumulator.
```
elizarov: typo: is taken

kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
```diff
@@ -97,3 +99,44 @@ public fun <T> Flow<T>.onEach(action: suspend (T) -> Unit): Flow<T> = flow {
...
+ * Reduces the given flow with [operation], emitting every intermediate result, including initial value.
+ * An initial value is provided lazily by [initialSupplier] and is always immediately emitted.
+ * For example:
+ * ```
+ * flowOf(1, 2, 3).scan(::emptyList) { acc: List<Int>, value -> acc + value }.toList()
+ * ```
+ * will produce `[], [1], [1, 2], [1, 2, 3]]`.
+ */
+@FlowPreview
+public fun <T, R> Flow<T>.scan(initialSupplier: () -> R, @BuilderInference operation: suspend (accumulator: R, value: T) -> R): Flow<R> = flow {
```
elizarov: `suspend () -> R` ?  

Anyway, we need to consider its consistency with https://youtrack.jetbrains.com/issue/KT-7657
One comment worth noting here:

> To be consistent with existing fold and reduce functions we need to name variants of scan or accumulate with and without initial value differently.

Unlike a stock scan, this variant's use-case is highly questionable, so I propose to avoid it for now altogether.

kotlinx-coroutines-core/common/src/flow/terminal/Collect.kt
```diff
@@ -32,3 +32,9 @@ public suspend inline fun <T> Flow<T>.collect(crossinline action: suspend (value
     collect(object : FlowCollector<T> {
         override suspend fun emit(value: T) = action(value)
     })
+
+/**
+ * Collects all the values from the given [flow] and emits them to the collector.
+ * Shortcut for `flow.collect { value -> emit(value) }`.
+ */
+public suspend inline fun <T> FlowCollector<T>.emitAll(flow: Flow<T>) = flow.collect { value -> emit(value) }
```
elizarov: `flow.collect(this)`

SAGARSURI:
Awesome!! Looking for more such great operators. Hope to completely replace RxJava someday with coroutines.

PR #1239, opened 2019-05-31
Adjust behavior of conflated channel to deliver last value

### LINKED ISSUE #332: ConflatedChannel unexpected behavior
```kotlin
val conflatedChannel = ConflatedChannel<Int>()
conflatedChannel.offer(1)
conflatedChannel.offer(2)
conflatedChannel.close()

conflatedChannel.receive() // expected receive 2 but error "Channel was closed"
```
It's very confusing, because `close()` method:
> Conceptually, its sends a special "close token" over this channel.

Is the `"close token"` treated equality with an element of Channel? I think it shouldn't be, then it shouldn't replace latest element of ConflatedChannel.

### LINKED ISSUE #1235: Adjust behavior of conflated channel to match use-case of intermediate values
See long discussion in #332. Here is TL;DR. Current behavior of `Channel(CONFLATED)` (`ConflatedChannel`) can be demonstrated with the following code ([play](https://pl.kotl.in/lM29toYHW)):

```
fun main() = runBlocking {
    val channel = produce(capacity = Channel.CONFLATED) { 
        send(1); send(2); send(3); send(4)
    }
    channel.consumeEach { println(it) } // prints only 1
}
```

What happens here is that closing a channel is considered to send a "value" that overwrites last sent value. This is subtle behavior and it is not clearly spelled out in docs. In fact, on can argue that docs imply a different behavior. Anyway, this is not what one wants when conflated channel is used to send intermediate values of some computation. In this case one needs to ensure that the most recent value is always delivered. This use-cases seems to be most prominent one and this proposal is to change behavior of conflated channels to match it, so that the above code prints 1 and 4.

elizarov:
Fixes #1235
Fixes #332

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
@@ -25,7 +25,35 @@ internal open class ConflatedChannel<E> : AbstractChannel<E>() {
     protected final override val isBufferFull: Boolean get() = false
 
     override fun onClosedIdempotent(closed: LockFreeLinkedListNode) {
-        conflatePreviousSendBuffered(closed)
+        @Suppress("UNCHECKED_CAST")
+        (closed.prevNode as? SendBuffered<E>)?.let { lastBuffered ->
+            conflatePreviousSendBuffered(lastBuffered)
+        }
+    }
+
+    /**
+     * Queues conflated element, returns null on success or
+     * returns node reference if it was already closed or is waiting for receive.
+     */
+    private fun sendConflated(element: E): ReceiveOrClosed<*>? {
+        val node = SendBuffered(element)
+        queue.addLastIfPrev(node) { prev ->
+            if (prev is ReceiveOrClosed<*>) return@sendConflated prev
+            true
+        }
+        conflatePreviousSendBuffered(node)
+        return null
+    }
+
+    private fun conflatePreviousSendBuffered(node: SendBuffered<E>) {
+        // Conflate all previous SendBuffered, helping other sends to conflate
+        var prev = node.prevNode
+        while (prev is SendBuffered<*>) {
+            if (!prev.remove()) {
+                prev.helpRemove()
+            }
+            prev = prev.prevNode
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -24,31 +24,20 @@ internal open class ConflatedChannel<E> : AbstractChannel<E>() {
     protected final override val isBufferAlwaysFull: Boolean get() = false
     protected final override val isBufferFull: Boolean get() = false
 
-    override fun onClosedIdempotent(closed: LockFreeLinkedListNode) {
```
qwwdfsad: This change makes the channel non-linearizable (see `ChannelLinearizabilityTest.testConflatedChannelLinearizability`).

`onClosedIdempotent ` call should conflate all sent elements except the last one.
Otherwise, weird non-linearizable executions are possible when closing thread interacts with the channel after `close`.

qwwdfsad:
I think we should document this behaviour somewhere, e.g. around `Factory.CONFLATED`

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -24,31 +24,20 @@ internal open class ConflatedChannel<E> : AbstractChannel<E>() {
...
-        conflatePreviousSendBuffered(closed)
-    }
```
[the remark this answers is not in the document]
elizarov: Good catch!

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
elizarov: Yes hell, indeed.

elizarov:
I've fixed the linearizability & moved the code. As for docs, it seems to be _already_ documented this way right here: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/index.html

PR #1242, opened 2019-06-01
Update "Shared mutable state and concurrency" guide with best practices

### LINKED ISSUE #1241: Docs: Thread confinement coarse-grained example does not follow best-practises
See here https://kotlinlang.org/docs/reference/coroutines/shared-mutable-state-and-concurrency.html#thread-confinement-coarse-grained
The function `massiveRun` is declared as extension on coroutine scope but and suspends at the same time.

elizarov:
* Properly use structured concurrency in the example code.
* Also remove a special not on <= 2 core CPU, which is no longer needed
because default dispatcher has always at least two CPUs.

Fixed #1241


docs/shared-mutable-state-and-concurrency.md
```diff
@@ -73,30 +74,33 @@ multi-threaded [Dispatchers.Default] that is used in [GlobalScope].
...
 
+//sampleStart
 var counter = 0
 
-fun main() = runBlocking<Unit> {
-//sampleStart
-    GlobalScope.massiveRun {
-        counter++
+fun main() = runBlocking {
+    withContext(Dispatchers.Default) {
```
zach-klippenstein: Isn't this effectively the same as `suspend fun main()`?

elizarov: Yes. But it looks more consistent across various examples in this chapter when it is written this way.

PR #1245, opened 2019-06-03
Add integration test with R8 to ensure ServiceLoader optimization

JakeWharton:


ui/kotlinx-coroutines-android/build.gradle
```diff
@@ -4,6 +4,17 @@
 
 repositories {
     google()
+    // TODO Remove once R8 is updated to a 1.5.x version.
```
JakeWharton: R8 1.5.x will be released when AGP 3.5.0 stable is released which is probably on the order of 4-8 weeks from now

elizarov:
I'm closing this PR as it was folder into #1304

PR #1260, opened 2019-06-07
Fix flowViaChannel closing immediately.

### LINKED ISSUE #1259: Regression: channelFlow/flowViaChannel broken.
This is a bug.
I updated a project today where I used `flowViaChannel` to bridge from `Observable` from Google's Agera library for Android, and it stopped working properly.
Only reverting to the old implementation of `flowViaChannel` from 1.2.1 made it work.

The behavior I observe is that I only receive the first value.
Note that I use `distinctUntilChanged()` after it, and removing the `conflate()` operator applied before doesn't change the broken behavior.

For anyone that need the original working `flowViaChannel` until that is fixed, here's the code:
```kotlin
import kotlinx.coroutines.ExperimentalCoroutinesApi
import kotlinx.coroutines.channels.Channel
import kotlinx.coroutines.channels.SendChannel
import kotlinx.coroutines.channels.consumeEach
import kotlinx.coroutines.coroutineScope
import kotlinx.coroutines.flow.Flow
import kotlin.experimental.ExperimentalTypeInference

@ExperimentalCoroutinesApi
@UseExperimental(ExperimentalTypeInference::class)
fun <T> flowViaChannel(
    bufferSize: Int = 16,
    @BuilderInference block: CoroutineScope.(channel: SendChannel<T>) -> Unit
): Flow<T> {
    return flow {
        coroutineScope {
            val channel = Channel<T>(bufferSize)
            launch {
                block(channel)
            }

            channel.consumeEach { value ->
                emit(value)
            }
        }
    }
}
```

LouisCAD:
Resolves #1259

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Builders.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Builders.kt
@@ -214,6 +214,7 @@ public fun <T> flowViaChannel(
 ): Flow<T> {
     return channelFlow<T> {
         block(channel)
+        awaitClose()
     }.buffer(bufferSize)
 }
 
--- a/kotlinx-coroutines-core/common/src/flow/Builders.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Builders.kt
@@ -204,10 +204,10 @@ public fun LongRange.asFlow(): Flow<Long> = flow {
  */
 @FlowPreview
 @Deprecated(
-    message = "Use channelFlow instead",
-    level = DeprecationLevel.WARNING,
-    replaceWith = ReplaceWith("channelFlow(block)")
+    message = "Use channelFlow with awaitClose { } instead of flowViaChannel and invokeOnClose { }.",
+    level = DeprecationLevel.WARNING
 )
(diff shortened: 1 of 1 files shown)


elizarov:
We'll drop `flowViaChannel` from the final `1.3.0` release, so I'm not sure it is worth to fix at this point. Most likely there will be at most one more intermediate release where it is already "fixed", but not removed. What do you think?

LouisCAD:
I think it'd be good to have a hot-fix ASAP, so people updating from 1.2.1
to it don't encounter the almost unavoidable bug. That can help teams, or
people coming back from vacations, or other tasks.

Then, removing in 1.3.0 makes sense, but until then, people will already be
willing to update to non preview API. Also, the difference in source
between flowViaChannel and channelFlow will encourage people to be more
cautious about how they update (instead of being bitten by straightforward,
awaitClose-less update to channelFlow, like me yesterday).

It might make sense to remove the ReplaceWith too, and use ALL_CAPS
somewhere for the migration guide.

LouisCAD:
I removed the wrong ReplaceWith clause too, and updated the deprecation message to be more helpful.

PR #1277, opened 2019-06-18
Change default request size in Publisher.openSubscription to one

### LINKED ISSUE #1267: Flux onBackpressureDrop and coroutines collect extension does not called passed lambda
Simple code with coroutines and project reactor does not produce any values:

Expected: "test" printed
Actual: don't see any output on the console

Kotlin coroutines version: 1.2.1

```kotlin
import kotlinx.coroutines.reactive.collect
import reactor.core.publisher.Flux

    runBlocking {
        Flux.never<Long>()
            .startWith(0)
            .onBackpressureDrop()
            .collect {
                println("test")
            }
    }
}

```

elizarov:
This makes Publisher.collect consistent with Publisher.asFlow and
ensures smooth integration with reactive backpressure operators
"out of the box".

Fixes #1267

### DIFF
--- a/reactive/coroutines-guide-reactive.md
+++ b/reactive/coroutines-guide-reactive.md
@@ -292,19 +292,20 @@ OnSubscribe
 1
 2
 3
-4
 OnComplete
 Finally
+4
 5
 ```
 
 <!--- TEST -->
 
-Notice how "OnComplete" and "Finally" are printed before the last element "5". It happens because our `main` function in this
+Notice how "OnComplete" and "Finally" are printed before the lasts elements "4" and "5". 
+It happens because our `main` function in this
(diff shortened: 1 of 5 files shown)


reactive/coroutines-guide-reactive.md
```diff
@@ -292,19 +292,20 @@ OnSubscribe
...
 OnComplete
 Finally
+4
 5
 ```
 
 <!--- TEST -->
 
-Notice how "OnComplete" and "Finally" are printed before the last element "5". It happens because our `main` function in this
+Notice how "OnComplete" and "Finally" are printed before the lasts element "4" and "5".
```
qwwdfsad: last element**s**?

PR #1279, opened 2019-06-19
Decouple asFlow from batchSize and move it to buffer instead, promote…

qwwdfsad:
… it to experimental

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/flow/PublisherAsFlow.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/flow/PublisherAsFlow.kt
@@ -7,35 +7,62 @@ package kotlinx.coroutines.reactive.flow
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
 import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.flow.internal.*
 import org.reactivestreams.*
+import kotlin.coroutines.*
 
 /**
  * Transforms the given reactive [Publisher] into [Flow].
- * Backpressure is controlled by [batchSize] parameter that controls the size of in-flight elements
- * and [Subscription.request] size.
+ * Use [buffer] operator on the resulting flow to specify the size of the backpressure.
+ * More precisely, to it specifies the value of the subscription's [request][Subscription.request].
+ * `1` is used by default.
  *
  * If any of the resulting flow transformations fails, subscription is immediately cancelled and all in-flights elements
  * are discarded.
  */
-@FlowPreview
-@JvmOverloads // For nice Java API
 @JvmName("from")
-public fun <T : Any> Publisher<T>.asFlow(batchSize: Int = 1): Flow<T> =
-    PublisherAsFlow(this, batchSize)
+@ExperimentalCoroutinesApi
+public fun <T : Any> Publisher<T>.asFlow(): Flow<T> =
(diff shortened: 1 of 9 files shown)


reactive/kotlinx-coroutines-reactive/src/flow/PublisherAsFlow.kt
```diff
@@ -7,66 +7,111 @@ package kotlinx.coroutines.reactive.flow
...
-@JvmOverloads // For nice Java API
 @JvmName("from")
-public fun <T : Any> Publisher<T>.asFlow(batchSize: Int = 1): Flow<T> =
-    PublisherAsFlow(this, batchSize)
+@Deprecated(
+    message = "batchSize parameter is deprecated, use .buffer() instead to control the backpressure",
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith("asFlow().buffer(batchSize)", imports = ["kotlinx.coroutines.flow.*"])
+)
+public fun <T : Any> Publisher<T>.asFlow(batchSize: Int): Flow<T> = asFlow().buffer(batchSize)
```
qwwdfsad: It breaks binary compatibility around `asFlow$default`, but it should be okay for preview feature

elizarov: Look fine.

elizarov:
I was not looking to optimize this case, but to have a single `Subscriber` implementation. They are hard to get right, so having just one such implementation is easier it terms of testing and future maintenance. However, `Subscriber` implementation that `openSubscription` has (which I've used) it definitely more complex. Moreover, it has different behavior. It uses "batch size" only for initial call to `request` but then just makes constant `request` calls just for one element in the future. That is less efficient that the original `Subscriber` implementation in this PR and may worth a look at.

elizarov:
Double checked... that is not the case... `SubscriptionChannel` has similar batch-requesting logic (requests full batch when it is over), but it has weird behavior with respect to batch. I'll looking at how valid it is.

elizarov:
Anyway, this solution would not work with `publisher.asFlow().conflate()`. Needs to be fixed.

elizarov:
Ok. Having studied `openSubscription` in detail I've reverted to mostly original implementation of the subscriber for `publisher.asFlow()` because it is slightly more efficient. I've also fixed support for `conflate()` and properly integration with reactive contract for "requests everything" (shall be Long.MAX_VALUE!). Added tests for various case.

elizarov:
@qwwdfsad Please take a look again.

reactive/kotlinx-coroutines-reactive/src/flow/PublisherAsFlow.kt
```diff
@@ -7,66 +7,124 @@ package kotlinx.coroutines.reactive.flow
...
         try {
-            var consumed = 0
-            for (i in channel) {
-                collector.emit(i)
-                if (++consumed == batchSize) {
-                    consumed = 0
-                    subscriber.subscription.request(batchSize.toLong())
+            var consumed = 0L
+            while (true) {
+                val value = subscriber.takeNextOrNull() ?: break
```
qwwdfsad: this instead of the iterator to save an extra object?

elizarov: Yes

PR #1280, opened 2019-06-19
Flow exceptions handling facilities

elizarov:
Flow exceptions handling facilities:

* Flow documentation introduces and explains "exception transparency"
  concept in detail.
* Introduce exception handling operators:
  * catch intermediate operator.
  * retry and retryWhen intermediate operators.
  * collect() without lambda terminal operator.
* onErrorXxx operators deprecated and moved to migration file when
  it was appropriate.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Flow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Flow.kt
@@ -35,11 +35,11 @@ import kotlin.coroutines.*
  * ```
  *
  * By default, flows are _sequential_ and all flow operations are executed sequentially in the same coroutine,
- * with exception for a few operations specifically designed to introduce concurrency into flow
- * execution such a [buffer] and [flatMapMerge]. See their documentation for details.
+ * with an exception for a few operations specifically designed to introduce concurrency into flow
+ * the execution such a [buffer] and [flatMapMerge]. See their documentation for details.
  *
  * Flow interface does not carry information whether a flow is a truly a cold stream that can be collected repeatedly and
- * triggers execution of the same code every time it is collected or if it is a hot streams that emits different
+ * triggers execution of the same code every time it is collected or if it is a hot stream that emits different
  * values from the same running source on each collection. However, conventionally flows represent cold streams.
  * Transitions between hot and cold streams are supported via channels and the corresponding API:
  * [channelFlow], [produceIn], [broadcastIn].
@@ -55,14 +55,14 @@ import kotlin.coroutines.*
  * * [channelFlow { ... }][channelFlow] builder function to construct arbitrary flows from
  *   potentially concurrent calls to [send][kotlinx.coroutines.channels.SendChannel.send] function.
  *
- * ### Flow constrains
+ * ### Flow constraints
  *
  * All implementations of `Flow` interface must adhere to two key properties that are described in detail below:
  *
  * * Context preservation.
  * * Exception transparency.
  *
- * These properties ensure ability to perform local reasoning about the code with flows and modularize the code
+ * These properties ensure the ability to perform local reasoning about the code with flows and modularize the code
  * in such a way so that upstream flow emitters can be developed separately from downstream flow collectors.
  * A user of the flow does not needs to know implementation details of the upstream flows it uses.
  *
@@ -95,7 +95,7 @@ import kotlin.coroutines.*
  * }
  * ```
  *
- * From the implementation point of view it means that all flow implementations should
+ * From the implementation point of view, it means that all flow implementations should
  * emit only from the same coroutine.
  * This constraint is efficiently enforced by the default [flow] builder.
  * The [flow] builder should be used if flow implementation does not start any coroutines.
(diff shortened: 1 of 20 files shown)


kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -12,13 +12,17 @@ import kotlin.coroutines.*
...
- * Flow values can be collected in a suspending manner without actual blocking using the [collect] extension that
- * completes normally or exceptionally:
+ * _Intermediate operators_ on the flow such as [map], [filter], [take], [zip], and etc are functions that are
+ * applied to the _upstream_ flow or flows and return a _downstream_ flow where further operators can be applied to.
+ * Intermediate operations do not execute any code in the flow and are not suspending functions themselves.
+ * They only set up a chain of operations for future execution and quickly return.
+ * This is known as a _cold flow_ property.
+ *
+ * _Terminal operators_ on the flow are suspending functions such as [collect], [single], [reduce], [toList], and etc.
+ * They are are applied to the upstream flow and trigger execution of all operations.
```
qwwdfsad: are are

kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -12,13 +12,17 @@ import kotlin.coroutines.*
...
- * _Collecting the flow_ means executing all its operations.
- * Flow values can be collected in a suspending manner without actual blocking using the [collect] extension that
- * completes normally or exceptionally:
+ * _Intermediate operators_ on the flow such as [map], [filter], [take], [zip], and etc are functions that are
+ * applied to the _upstream_ flow or flows and return a _downstream_ flow where further operators can be applied to.
+ * Intermediate operations do not execute any code in the flow and are not suspending functions themselves.
+ * They only set up a chain of operations for future execution and quickly return.
+ * This is known as a _cold flow_ property.
+ *
+ * _Terminal operators_ on the flow are suspending functions such as [collect], [single], [reduce], [toList], and etc.
```
qwwdfsad: `and etc.` should be replaced with `etc.` 

>A good way to test whether etc. is appropriate is to substitute “and so on” or “and so forth.” If those synonyms make sense, you can use etc. You should never use “and et cetera.” Remember, et means “and.” “And et cetera” is redundant.

kotlinx-coroutines-core/common/src/flow/terminal/Collect.kt
```diff
@@ -8,13 +8,26 @@
...
+/**
+ * Terminal flow operator that collects the given flow but ignores all emitted values.
+ * If any exception occurs during collect or in the provided flow, this exception is rethrown from this method.
+ *
+ * It is a shorthand for `collect {}`.
+ *
+ * :todo: exception handling
+ */
+@ExperimentalCoroutinesApi
+public suspend fun Flow<*>.collect() = collect(NopCollector)
```
LouisCAD: Shouldn't this function be inline like `emitAll`?

elizarov: Making it inline does not add anything in terms of performance, but increases bytecode size for all users of this function.

LouisCAD: Oh, I just noticed `collect(…)` itself is `inline` too, I now understand how it'd make the bytecode bigger.

kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -108,29 +126,57 @@ import kotlin.coroutines.*
...
+ *
+ * ```
+ * flow { emitData() }
+ *     .map { computeOne(it) }
+ *     .catch { ... } // catches exceptions in emitData and computeOne
+ *     .map { computeTwo(it) }
+ *     .collect { process(it) } // throws exceptions from process and computeTwo
+ * ```
+ *
+ * Failure to adhere to the exception transparency requirement would strange behaviours that would make it hard
```
bohsen: >  Failure to adhere to the exception transparency requirement **_would strange behaviours_** that would make it hard

Is this correct?

Otherwise a good improvement to the documentation.

elizarov: Thanks for noticing. That was a typo (missing words, actually). I've also improved this whole sentence a bit having reread it from scratch.

kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -13,90 +13,210 @@ import kotlin.coroutines.*
...
+                collector.emit(it)
             } catch (e: Throwable) {
-                if (fromDownstream || e.isCancellationCause(coroutineContext)) throw e
-                if (!predicate(e) || retries-- == 0) throw e
+                fromDownstream = e
+                throw e
             }
         }
+    } catch (e: Throwable) {
+        if (e.isSameExceptionAs(fromDownstream) || e.isCancellationCause(coroutineContext))
```
zach-klippenstein: Is my understanding correct that you explicitly check for cancellation exception as well because there's no point in allowing the operator to see an exception if it has already caused the coroutine to be cancelled, since the operator would effectively be a no-op in that case anyway?

qwwdfsad: >since the operator would effectively be a no-op in that case anyway?

It won't be a no-op and that it the problem.
Consider the following code:
```
flow {
   // suspension point (1)
}
.catch({ println(it)  }) // Print an exception
.map {
  // suspension point (2)
}
.single()
```
If the enclosing coroutine is cancelled, `CancellationException` will be thrown either from `(1)` or `(2)`. `catch` (and `retry`) **should** be consistent in both cases.

PR #1282, opened 2019-06-19
Optimize the size of the coroutines library in Android projects

elizarov:
*updated* (see below)

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,10 +165,7 @@ threads are handled by Android runtime.
 #### R8 and ProGuard
 
 R8 and ProGuard rules are bundled into the [`kotlinx-coroutines-android`](ui/kotlinx-coroutines-android) module.
-For best results it is recommended to use a recent version of R8.
-R8 is a replacement for ProGuard in Android ecosystem, it is enabled by default since Android gradle plugin 3.4.0
-(3.3.0-beta also had it enabled). The upcoming AGP 4.0.0 has never R8 and additional rules enable
-more optimizations, producing smaller binary size. 
+For more details see ["Optimization" section for Android](ui/kotlinx-coroutines-android/README.md#optimization). 
 
 ### JS
 
--- a/README.md
+++ b/README.md
@@ -164,11 +164,11 @@ threads are handled by Android runtime.
 
 #### R8 and ProGuard
 
-For R8 no actions required, it will take obfuscation rules from the jar.
-
-For Proguard  you need to add options from [coroutines.pro](kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro) to your rules manually.
- 
-R8 is a replacement for ProGuard in Android ecosystem, it is enabled by default since Android gradle plugin 3.4.0 (3.3.0-beta also had it enabled).
+R8 and ProGuard rules are bundled into the [`kotlinx-coroutines-android`](ui/kotlinx-coroutines-android) module.
+For best results it is recommended to use a recent version of R8.
+R8 is a replacement for ProGuard in Android ecosystem, it is enabled by default since Android gradle plugin 3.4.0
+(3.3.0-beta also had it enabled). The upcoming AGP 4.0.0 has never R8 and additional rules enable
+more optimizations, producing smaller binary size. 
 
 ### JS
 
--- a/build.gradle
+++ b/build.gradle
@@ -137,10 +137,9 @@ apiValidation {
     ignoredPackages += "kotlinx.coroutines.internal"
 }
 
-
+// Configure repositories
 allprojects {
-    apply plugin: 'kotlinx-atomicfu' // it also adds all the necessary dependencies
-    def projectName = it.name
+    String projectName = it.name
     repositories {
         /*
          * google should be first in the repository list because some of the play services
@@ -158,30 +157,31 @@ allprojects {
         maven { url "https://kotlin.bintray.com/kotlin-eap" }
         maven { url "https://kotlin.bintray.com/kotlinx" }
     }
+}
 
-    if (projectName == rootModule || projectName == coreModule) return
-
-    // Add dependency to core source sets. Core is configured in kx-core/build.gradle
+// Add dependency to core source sets. Core is configured in kx-core/build.gradle
+configure(subprojects.findAll { !sourceless.contains(it.name) && it.name != coreModule }) {
     evaluationDependsOn(":$coreModule")
-    if (sourceless.contains(projectName)) return
-
     def platform = platformOf(it)
     apply from: rootProject.file("gradle/compile-${platform}.gradle")
-
     dependencies {
         // See comment below for rationale, it will be replaced with "project" dependency
         compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:$version"
-
         // the only way IDEA can resolve test classes
         testCompile project(":$coreModule").kotlin.targets.jvm.compilations.test.output.allOutputs
     }
+}
 
+// Configure subprojects with Kotlin sources
+configure(subprojects.findAll { !sourceless.contains(it.name) }) {
+    // Use atomicfu plugin, it also adds all the necessary dependencies
+    apply plugin: 'kotlinx-atomicfu'
+
+    // Configure options for all Kotlin compilation tasks
     tasks.withType(org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompile).all {
         kotlinOptions.freeCompilerArgs += experimentalAnnotations.collect { "-Xuse-experimental=" + it }
         kotlinOptions.freeCompilerArgs += "-progressive"
         kotlinOptions.freeCompilerArgs += "-XXLanguage:+InlineClasses"
-        // Binary compatibility support
-        kotlinOptions.freeCompilerArgs += ["-Xdump-declarations-to=${buildDir}/visibilities.json"]
         // Remove null assertions to get smaller bytecode on Android
         kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
     }
--- a/build.gradle
+++ b/build.gradle
@@ -182,6 +182,8 @@ allprojects {
         kotlinOptions.freeCompilerArgs += "-XXLanguage:+InlineClasses"
         // Binary compatibility support
         kotlinOptions.freeCompilerArgs += ["-Xdump-declarations-to=${buildDir}/visibilities.json"]
+        // Remove null assertions to get smaller bytecode on Android
+        kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
     }
 }
 
--- a/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 package kotlinx.coroutines.internal
 
 import kotlinx.coroutines.*
@@ -30,11 +34,12 @@ internal object MainDispatcherLoader {
                         MainDispatcherFactory::class.java.classLoader
                 ).iterator().asSequence().toList()
             }
+            @Suppress("ConstantConditionIf")
             factories.maxBy { it.loadPriority }?.tryCreateDispatcher(factories)
-                ?: MissingMainCoroutineDispatcher(null)
+                ?: createMissingDispatcher()
         } catch (e: Throwable) {
             // Service loader can throw an exception as well
-            MissingMainCoroutineDispatcher(e)
+            createMissingDispatcher(e)
         }
     }
 }
@@ -51,12 +56,29 @@ public fun MainDispatcherFactory.tryCreateDispatcher(factories: List<MainDispatc
     try {
         createDispatcher(factories)
     } catch (cause: Throwable) {
-        MissingMainCoroutineDispatcher(cause, hintOnError())
+        createMissingDispatcher(cause, hintOnError())
     }
 
 /** @suppress */
 @InternalCoroutinesApi
 public fun MainCoroutineDispatcher.isMissing(): Boolean = this is MissingMainCoroutineDispatcher
+
+// R8 optimization hook, not const on purpose to enable R8 optimizations via "assumenosideeffects"
+@Suppress("MayBeConstant")
+val SUPPORT_MISSING = true
+
+@Suppress("ConstantConditionIf")
+private fun createMissingDispatcher(cause: Throwable? = null, errorHint: String? = null) =
+    if (SUPPORT_MISSING) MissingMainCoroutineDispatcher(cause, errorHint) else
+        cause?.let { throw it } ?: throwMissingMainDispatcherException()
+
+internal fun throwMissingMainDispatcherException(): Nothing {
+    throw IllegalStateException(
+        "Module with the Main dispatcher is missing. " +
+            "Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android' " +
+            "and ensure it has the same version as 'kotlinx-coroutines-core'"
+    )
+}
 
 private class MissingMainCoroutineDispatcher(
     private val cause: Throwable?,
@@ -85,11 +107,7 @@ private class MissingMainCoroutineDispatcher(
 
     private fun missing(): Nothing {
         if  (cause == null) {
-            throw IllegalStateException(
-                "Module with the Main dispatcher is missing. " +
-                    "Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android' " +
-                        "and ensure it has the same version as 'kotlinx-coroutines-core'"
-            )
+            throwMissingMainDispatcherException()
         } else {
             val message = "Module with the Main dispatcher had failed to initialize" + (errorHint?.let { ". $it" } ?: "")
             throw IllegalStateException(message, cause)
--- a/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
@@ -65,7 +65,7 @@ public fun MainCoroutineDispatcher.isMissing(): Boolean = this is MissingMainCor
 
 // R8 optimization hook, not const on purpose to enable R8 optimizations via "assumenosideeffects"
 @Suppress("MayBeConstant")
-val SUPPORT_MISSING = true
+private val SUPPORT_MISSING = true
 
 @Suppress("ConstantConditionIf")
 private fun createMissingDispatcher(cause: Throwable? = null, errorHint: String? = null) =
--- a/kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 @file:Suppress("UNCHECKED_CAST")
@@ -52,7 +52,8 @@ private fun <E : Throwable> E.sanitizeStackTrace(): E {
     return this
 }
 
-internal actual fun <E : Throwable> recoverStackTrace(exception: E, continuation: Continuation<*>): E {
+@Suppress("NOTHING_TO_INLINE") // Inline for better R8 optimization
+internal actual inline fun <E : Throwable> recoverStackTrace(exception: E, continuation: Continuation<*>): E {
     if (!RECOVER_STACK_TRACES || continuation !is CoroutineStackFrame) return exception
     return recoverFromStackFrame(exception, continuation)
 }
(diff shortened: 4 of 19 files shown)


Tolriq:
Can you please comment a little on ```Use simpler and smaller CommonPool for both Dispatchers.Default and IO```?
I'm currently working on my app to try to better manage threads due to contention potentially starting unneeded threads with workerpool patterns. Maybe there's less needs with this.

elizarov:
It is just code-size optimization. `CommonPool` reuses platform code (either FJP or TPE) so its code-size is smaller. However, it actually makes thread usage worse, because those underlying thread-pool start all threads all the time, while more efficient `CoroutinesScheduler`  starts them only on demand. But CS has way more code, which might be critical for apps targeting extremely low-end devices with very tight budgets on .dex size.

Tolriq:
Hum OK thanks I'll continue to search why there's no many threads created and how to mitigate then :(

ui/kotlinx-coroutines-android/README.md
```diff
@@ -5,6 +5,63 @@ Provides `Dispatchers.Main` context for Android applications.
 Read [Guide to UI programming with coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/master/ui/coroutines-guide-ui.md)
 for tutorial on this module.
 
+# Optimization
+
+You can optimize the size of the coroutines library in Android projects by statically removing 
+debugging and other advanced features using R8 by adding the following rule to your 
+`proguard-rules.pro` file:
```
LouisCAD: Why not bundle these rules directly in `META_INF` so these optimizations are turned on by default on all projects, without manual configuration?

elizarov: Some people want to turn on debugging facilities. I'm not sure what would be the right default behavior and how configuration can be undone if we bundle it.

kotlinx-coroutines-core/jvm/src/Dispatchers.kt
```diff
@@ -106,5 +107,32 @@ public actual object Dispatchers {
...
+@Suppress("MayBeConstant")
+internal val IO_SCHEDULER = true
+
+@Suppress("NOTHING_TO_INLINE") // For better constant propagation in R8
+internal actual inline fun createDefaultDispatcher(): CoroutineDispatcher =
+    if (DEFAULT_SCHEDULER) DefaultScheduler else CommonPool
+
+@Suppress("NOTHING_TO_INLINE", "ConstantConditionIf") // For better constant propagation in R8
+internal inline fun createIoDispatcher(): CoroutineDispatcher =
+    if (IO_SCHEDULER) DefaultScheduler.IO else CommonPool
```
qwwdfsad: `CommonPool` cannot be used as IO dispatcher.
Either separate FJP/executor (NB: daemon threads) should be created or (if we are willing to implement "pool view" API) `ManagedBlocker`-based dispatcher should be used

build.gradle
```diff
@@ -137,6 +137,8 @@ allprojects {
         kotlinOptions.freeCompilerArgs += "-progressive"
         // Binary compatibility support
         kotlinOptions.freeCompilerArgs += ["-Xdump-declarations-to=${buildDir}/visibilities.json"]
+        // Remove null assertions to get smaller bytecode on Android
+        kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
```
qwwdfsad: Does it finally work?

kotlinx-coroutines-core/common/src/AbstractCoroutine.kt
```diff
@@ -71,7 +71,7 @@ public abstract class AbstractCoroutine<in T>(
      */
     protected open fun onStart() {}
 
-    internal final override fun onStartInternal() {
+    final override fun onStartInternal() {
```
qwwdfsad: What's the rationale of making a lot of API `public`, avoid name mangling and bridges?

kotlinx-coroutines-core/jvm/src/internal/LockFreeTaskQueue.kt
```diff
@@ -78,12 +79,17 @@ internal open class LockFreeTaskQueue<E : Any>(
  */
 internal class LockFreeTaskQueueCore<E : Any>(
     private val capacity: Int,
-    private val singleConsumer: Boolean // true when there is only a single consumer (slightly faster)
+    @JvmField // JvmField + public avoids generation of accessors
```
qwwdfsad: We are going to regret maintaining this.
Either R8 or kotlinc should have a flag "escalate visibility to avoid accessors" (dreaming of nestmates)

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
```diff
@@ -53,13 +58,29 @@ public fun MainDispatcherFactory.tryCreateDispatcher(factories: List<MainDispatc
...
+        createMissingDispatcher(cause, hintOnError())
     }
 
 /** @suppress */
 @InternalCoroutinesApi
 public fun MainCoroutineDispatcher.isMissing(): Boolean = this is MissingMainCoroutineDispatcher
 
+// R8 optimization hook, not const on purpose
+@Suppress("MayBeConstant")
+private val SUPPORT_MISSING = true
```
qwwdfsad: It's better to report this case to R8 instead of workarounding it here, especially with new SL optimization

qwwdfsad: Maybe @wojtek-kalicinski can verify that missing dispatcher is correctly removed when R8 optimizes SL?

kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
```diff
@@ -38,7 +38,8 @@ private fun <E : Throwable> E.sanitizeStackTrace(): E {
     return this
 }
 
-internal actual fun <E : Throwable> recoverStackTrace(exception: E, continuation: Continuation<*>): E {
+@Suppress("NOTHING_TO_INLINE") // Inline for better R8 optimization
+internal actual inline fun <E : Throwable> recoverStackTrace(exception: E, continuation: Continuation<*>): E {
```
qwwdfsad: Changes in this file should blow up bytecode size.
`inline` on `recoveryDisabled` should be enough, otherwise let's delegate it to R8

build.gradle
Tolriq: I'm using those in prod and can confirm that they work and reduce dex size by a large factor.

But I've exchanged a lot with R8 team on a few things and I was told that in the future it may be better to not apply those as R8 does use those intrinsics for some of their optimisations.
As of R8 1.5.51 (Last one currently for future AS 3.5 release) those compile flags still do a better job than R8, but it may not be the case for future 1.6 and it should be discussed with R8 tmea I suppose.

elizarov: They don't really work here. It seems to be a bug in `kotlin-multiplatform` plugin.

kotlinx-coroutines-core/common/src/AbstractCoroutine.kt
elizarov: Yes. It is a minor improvement in size, but still.

kotlinx-coroutines-core/jvm/src/internal/LockFreeTaskQueue.kt
elizarov: The classes where I did are internal anyway. Yes, it means public in JVM, but it does not make any practical difference, since no public API declares those types in their signature, so you cannot accidentally use them even from Java code (which is highly unlikely to be found around coroutines anyway).

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
elizarov: It is not removed now, but that is totally due to a different bug in R8. This constant cannot be made `const` anyway, because if it is `const` then it leaves no trace in the class files, making it impossible to configure via R8.

kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
elizarov: It has to be inline, or otherwise the `StrackTraceRecoveryKt.class` cannot be removed by R8. It is correctly removed now, which is verified by `-checkdiscard class kotlinx.coroutines.internal.StackTraceRecoveryKt`. 

Notice how `recoverStackTrace` function body checks for `recoveryDisabled` and the calls the actual `recoverFromStackFrame` implementation function. This way, when it is inlined and `recoveryDisabled` is also inline (check it out!) and statically knows that recovery is disabled, then no references to  `StrackTraceRecoveryKt.class` are left and it is removed completely.

kotlinx-coroutines-core/jvm/src/Dispatchers.kt
elizarov: The importance of a proper IO dispatcher for Android is questionable, but, indeed, it would be better to figure out a simple way to implement a better IO dispatcher. It might be based on some cached thread pool executor, for example. It will forfeit thread sharing, but that should not be an issue for Android anyway.

README.md
```diff
@@ -165,9 +165,15 @@ threads are handled by Android runtime.
 
 For R8 no actions required, it will take obfuscation rules from the jar.
 
-For Proguard  you need to add options from [coroutines.pro](kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro) to your rules manually.
+For Proguard  you need to add options from
```
wojtek-kalicinski: Actually, it's the Android Gradle Plugin that takes care of extracting rules from META-INF, so if you're compiling for Android using our Gradle build, the rules should work automatically with either ProGuard or R8
(since you're changing this file anyway...)

ui/kotlinx-coroutines-android/README.md
```diff
@@ -5,6 +5,63 @@ Provides `Dispatchers.Main` context for Android applications.
...
+-assumenosideeffects class kotlinx.coroutines.DispatchersKt {
+    boolean getDEFAULT_SCHEDULER() return false;
+    boolean getIO_SCHEDULER() return false;
+}
+
+-checkdiscard class kotlinx.coroutines.DispatchersKt
+-checkdiscard class kotlinx.coroutines.scheduling.CoroutinesScheduler
+
+# Remove the custom, fast service loader implementation
+-assumevalues class kotlinx.coroutines.internal.MainDispatcherLoader {
```
wojtek-kalicinski: we should add a comment to make sure only R8 users enable this, as ProGuard doesn't have the necessary ServiceLoader optimization

I'm also currently working on a mechanism for AGP 3.6.0 to enable putting seperate rules for PG and R8 in META-INF, so we'll be able to do this automatically for anyone on R8 1.5+ or R8 1.6+ (depending on which one we choose)

README.md
LouisCAD: I read that those rules worked only for R8. Does it depend on a minimum AGP version instead?

ui/kotlinx-coroutines-android/README.md
LouisCAD: It's very uncommon to enable R8 (using minifyEnabled) on a debug build.
I can think of someone wanting to debug coroutines post R8 processing though.

@wojtek-kalicinski are you aware of way to disable third-party rules for such a case?

kotlinx-coroutines-core/jvm/src/internal/LockFreeTaskQueue.kt
qwwdfsad: My concern is not `public` modifier, but the fact that now we should think about using `@JvmField + public` instead of `private` when the bridge will be generated.
This is fragile, easy to forget and hard to test.

That's why I think some compiler support is required here

kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
qwwdfsad: If `recoveryDisabled` is inline and statically known, optimizer should make `recoverStackTrace` a no-op and then safely remove.
Even if not (let's report it then), the only thing that should be left is empty `StrackTraceRecoveryKt` class that is not a big deal.

ui/kotlinx-coroutines-android/README.md
wojtek-kalicinski: we're working on a method to do it, but it won't be ready soon.

I'm wondering if just putting an opposite `-assumevalues ... return true;` in your project's proguard config would work, i.e. if it would be read after the library one and override it. I'd need to test that to be sure.

In any case, I also prefer rules to be applied automatically, otherwise 90% developers won't do it. It would be ideal if disabling the debug stripping would be possible though.

kotlinx-coroutines-core/jvm/src/Dispatchers.kt
qwwdfsad: >but that should not be an issue for Android anyway.

Apparently, it is an issue, according to different feedback.
We should try to minimize the number of threads, especially for "view" API.
I will investigate it further.

elizarov: Let's postone working of this PR until after 1.3.0 is out. We'll need to get a number of R8 fixes, too, in order to be able to trim the size further anyway. If I have time I'll see if I can separate some less questionable pieces of this big PR as smaller PRs.

kotlinx-coroutines-core/jvm/src/internal/LockFreeTaskQueue.kt
elizarov: Agreed. Let's not do this kind of changes and find a more sustainable strategy for this kind of optimizations. One approach is to consider a use of R8 as a standalone optimization tool (it can spit out class-files, too) that we would use on kotlinx-coroutines.jar that we deliver to the public to mangle and shorten the names of all private and internal identifiers and remove accessors. However, we have to be careful that we don't touch anything that might potentially become a part of a user-visible stack-trace. Needs a separate thinking.

kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
elizarov: It does not work this way. It makes it no-op, but does not remove, because it assumes that there might be class-initialization side-effects on calling this method. Removing it would require a separate `assumenosideeffects` to R8. Keeping `StrackTraceRecoveryKt` is a big deal, though, because it also contains various statics and their initialization. We have to make sure it is totally gone.

ui/kotlinx-coroutines-android/README.md
elizarov: I'm bundling _some_ of the rules (like turning off debug support). It looks like it would be Ok to say that you need to turn off optimization to support debugging.

build.gradle
elizarov: Fixed now.

kotlinx-coroutines-core/common/src/AbstractCoroutine.kt
elizarov: I've got rid of those tweaks. It makes no difference with obfuscation.

ui/kotlinx-coroutines-android/README.md
elizarov: I'm bundling all the rules in this PR.

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
elizarov: I'm leaving this `SUPPORT_MISSING`. The logic around it seems to bee too complex for R8 to analyze, but with `-assumenosideeffects class kotlinx.coroutines.internal.MainDispatchersKt { boolean SUPPORT_MISSING return false; }` I can verify that it gets removed.

README.md
elizarov: I reworded this part of readme.

elizarov:
I've cleaned up this PR, removed questionable optimizations, no more messing up with scheduler (scheduler is quite smaller now anyway). All the rules are bundles in the jar for R6 1.6.0+. Verified to work in a test project using AGP 4.0.0-alpha06. This PR saves additional 16Kb is DEX size in addition to savings you get by simply using a newer R8 version.

README.md
```diff
@@ -164,11 +164,11 @@ threads are handled by Android runtime.
...
 
-For R8 no actions required, it will take obfuscation rules from the jar.
-
-For Proguard  you need to add options from [coroutines.pro](kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro) to your rules manually.
- 
-R8 is a replacement for ProGuard in Android ecosystem, it is enabled by default since Android gradle plugin 3.4.0 (3.3.0-beta also had it enabled).
+R8 and ProGuard rules are bundled into the [`kotlinx-coroutines-android`](ui/kotlinx-coroutines-android) module.
+For best results it is recommended to use a recent version of R8.
+R8 is a replacement for ProGuard in Android ecosystem, it is enabled by default since Android gradle plugin 3.4.0
+(3.3.0-beta also had it enabled). The upcoming AGP 4.0.0 has never R8 and additional rules enable
```
egor-n:
```suggestion
(3.3.0-beta also had it enabled). The upcoming AGP 4.0.0 has newer R8 and additional rules enable
```
?

Tolriq: BTW @wojtek-kalicinski have fixed the version checks in AGP 3.6 too, so this works starting AGP 3.6.0

ui/kotlinx-coroutines-android/testdata/r8-test-rules.pro
```diff
@@ -0,0 +1,14 @@
+-include r8-test-common.pro
+
+-include ../resources/META-INF/com.android.tools/r8-from-1.6.0/coroutines.pro
+
+# Validate that service-loader & debugger classes are discarded
+-checkdiscard class kotlinx.coroutines.internal.FastServiceLoader
+-checkdiscard class kotlinx.coroutines.DebugKt
+-checkdiscard class kotlinx.coroutines.internal.StackTraceRecoveryKt
```
qwwdfsad: Great!

ui/kotlinx-coroutines-android/testdata/r8-test-rules.pro
```diff
@@ -0,0 +1,14 @@
...
+
+# Real android projects do not keep this class, but somehow it is kept in this test (R& bug)
```
qwwdfsad: R& -> R8 (here and below)

ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt
```diff
@@ -12,17 +12,24 @@ import kotlin.coroutines.*
...
-
-    // Reflectively lookup pre-handler. Implement Function0 to avoid generating second class for lambda
-    override fun invoke(): Method? = try {
-        Thread::class.java.getDeclaredMethod("getUncaughtExceptionPreHandler").takeIf {
-            Modifier.isPublic(it.modifiers) && Modifier.isStatic(it.modifiers)
+    // Reflectively lookup pre-handler.
+    private fun preHandler(): Method? {
+        val current = _preHandler
+        if (current !== this) return current as Method?
+        val declared =try {
```
qwwdfsad: missing indent

ui/kotlinx-coroutines-android/resources/META-INF/com.android.tools/r8-from-1.6.0/coroutines.pro
```diff
@@ -5,6 +5,18 @@
...
+# Disable support for "Missing Main Dispatcher", since we always have Android main dispatcher
+-assumenosideeffects class kotlinx.coroutines.internal.MainDispatchersKt {
+    boolean SUPPORT_MISSING return false;
+}
+
+# Statically turn off all debugging facilities and assertions
+-assumenosideeffects class kotlinx.coroutines.DebugKt {
+    boolean getASSERTIONS_ENABLED() return false;
+    boolean getDEBUG() return false;
+    boolean getRECOVER_STACK_TRACES() return false;
```
qwwdfsad: This fact should be reflected in documentation to android module, stacktracktrace recovery KDoc and related `.md` doc

README.md
LouisCAD: Also: enable -> enabled (end of the sentence)

LouisCAD:
```suggestion
(3.3.0-beta also had it enabled). The upcoming AGP 4.0.0 has newer R8 and additional rules enabled
```

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
```diff
@@ -51,13 +56,30 @@ public fun MainDispatcherFactory.tryCreateDispatcher(factories: List<MainDispatc
...
+        createMissingDispatcher(cause, hintOnError())
     }
 
 /** @suppress */
 @InternalCoroutinesApi
 public fun MainCoroutineDispatcher.isMissing(): Boolean = this is MissingMainCoroutineDispatcher
 
+// R8 optimization hook, not const on purpose to enable R8 optimizations via "assumenosideeffects"
+@Suppress("MayBeConstant")
+val SUPPORT_MISSING = true
```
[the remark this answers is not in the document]
elizarov: Good catch. Should be `private`.

README.md
elizarov: 👍

ui/kotlinx-coroutines-android/resources/META-INF/com.android.tools/r8-from-1.6.0/coroutines.pro
elizarov: Done

build.gradle
```diff
@@ -174,6 +174,8 @@ allprojects {
         kotlinOptions.freeCompilerArgs += "-XXLanguage:+InlineClasses"
         // Binary compatibility support
         kotlinOptions.freeCompilerArgs += ["-Xdump-declarations-to=${buildDir}/visibilities.json"]
+        // Remove null assertions to get smaller bytecode on Android
+        kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
```
qwwdfsad: This args are not applied to the compilation of the core artifact because of line 154

elizarov:
Works like a charm. Now also reports the size of the optimized android DEX file for basic coroutine usage: https://teamcity.jetbrains.com/project/KotlinTools_KotlinxCoroutines?mode=builds&projectTab=stats

PR #1285, opened 2019-06-20
Add distinctUntilChanged operator that uses a comparator function instead of a key extractor.

zach-klippenstein:
**Use case:** Deduping values that have some custom definition of equality that can't be expressed easily via a single key (e.g. compares multiple properties, or via a custom comparator interface).

This is implemented as a two-step comparison, first extracting the keys (which is the identity function by default) and then comparing keys using the comparator function (which is just `==` by default).

The `By` suffix doesn't really make sense without a key selector, so I called the comparator-only version simply `distinctUntilChanged`. This also disambiguates the operator when a parameter-less lambda is passed in, without having to use placeholders (e.g. `distinctUntilChanged{ _ -> ... }`).

| operator | equality defined as |
|---------|---------------------|
| `distinctUntilChanged()` | `old == new` |
| `distinctUntilChanged(areEquivalent)` | `areEquivalent(old, new)` |
| `distinctUntilChangedBy(keySelector)` | `keySelector(old) == keySelector(new)` |
| `distinctUntilChangedBy(keySelector, areEquivalent)` | `areEquivalent(keySelector(old), keySelector(new))` |

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Distinct.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Distinct.kt
@@ -19,16 +19,35 @@ import kotlinx.coroutines.flow.unsafeFlow as flow
 public fun <T> Flow<T>.distinctUntilChanged(): Flow<T> = distinctUntilChangedBy { it }
 
 /**
+ * Returns flow where all subsequent repetitions of the same value are filtered out, when compared
+ * with each other via the provided [areEquivalent] function.
+ */
+@FlowPreview
+public fun <T> Flow<T>.distinctUntilChanged(areEquivalent: (old: T, new: T) -> Boolean): Flow<T> =
+    distinctUntilChangedBy(keySelector = { it }, areEquivalent = areEquivalent)
+
+/**
  * Returns flow where all subsequent repetitions of the same key are filtered out, where
  * key is extracted with [keySelector] function.
  */
 @FlowPreview
 public fun <T, K> Flow<T>.distinctUntilChangedBy(keySelector: (T) -> K): Flow<T> =
+    distinctUntilChangedBy(keySelector = keySelector, areEquivalent = { old, new -> old == new })
+
+/**
+ * Returns flow where all subsequent repetitions of the same key are filtered out, where
+ * keys are extracted with [keySelector] function and compared with each other via the
+ * provided [areEquivalent] function.
+ */
+private inline fun <T, K> Flow<T>.distinctUntilChangedBy(
+    crossinline keySelector: (T) -> K,
+    crossinline areEquivalent: (old: K, new: K) -> Boolean
+): Flow<T> =
     flow {
         var previousKey: Any? = NULL
         collect { value ->
             val key = keySelector(value)
-            if (previousKey != key) {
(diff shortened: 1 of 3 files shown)


zach-klippenstein:
> I can see uses of distinctUntilChanged(areEquivalent) (though, some anecdotal evidence, we don't have Sequence.distinct overload with comparator in stdlib and we don't have a request for such overload

We have code (that is unfortunately private) that uses this overload to compare values in a number of ways: checks if the values implement an interface and use that interface to compare, else fall back to some default comparison logic. It might be possible to refactor that into using a temporary key, but it would be a much less elegant solution than just providing a comparator function directly, and we would mostly likely end up writing this operator internally in that case instead.

> public distinctUntilChangedBy(keySelector, areEquivalent) seems way too verbose and unpractical.

I don't have a use case for that, I am happy to make it private. It's just there to unify the implementations of the other two operators, but I could also remove it entirely since they're both so simple.

zach-klippenstein:
Just pushed some changes: made the two-function `distinctUntilChangedBy` private and inline (implementations of each other other functions are so similar it is still nice to have them factored out).

I will run the binary compatibility validator as soon as I get around to installing JDK 1.6 on my machine.

kotlinx-coroutines-core/common/src/flow/operators/Distinct.kt
```diff
@@ -18,17 +18,36 @@ import kotlinx.coroutines.flow.unsafeFlow as flow
 @ExperimentalCoroutinesApi
 public fun <T> Flow<T>.distinctUntilChanged(): Flow<T> = distinctUntilChangedBy { it }
 
+/**
+ * Returns flow where all subsequent repetitions of the same value are filtered out, when compared
+ * with each other via the provided [areEquivalent] function.
+ */
+@FlowPreview
+public fun <T> Flow<T>.distinctUntilChanged(areEquivalent: (old: T, new: T) -> Boolean): Flow<T> =
```
zach-klippenstein: Would like to confirm this name – `distinctUntilChangedBy` feels wrong, because this overload doesn't necessarily compare "by" a single thing, and using that name also makes `distinctUntilChangedBy { it.something }` not work, because the function call is ambiguous without explicit lambda parameters. Maybe `distinctUntilChangedWith`? Or is this just fine?

kotlinx-coroutines-core/common/src/flow/operators/Distinct.kt
```diff
@@ -18,17 +18,36 @@ import kotlinx.coroutines.flow.unsafeFlow as flow
...
 @FlowPreview
 public fun <T, K> Flow<T>.distinctUntilChangedBy(keySelector: (T) -> K): Flow<T> =
+    distinctUntilChangedBy(keySelector = keySelector, areEquivalent = { old, new -> old == new })
+
+/**
+ * Returns flow where all subsequent repetitions of the same key are filtered out, where
+ * keys are extracted with [keySelector] function and compared with each other via the
+ * provided [areEquivalent] function.
+ */
+private inline fun <T, K> Flow<T>.distinctUntilChangedBy(
```
zach-klippenstein: Since this is private now, getting this name right isn't as crucial, but open to changing it as well.

qwwdfsad:
Merged manually

PR #1291, opened 2019-06-25
Declarative flow operators

qwwdfsad:
`onCompletion` and basic `launchIn` support for 1.3.0-M2.

Open questions that we can work on later:
  * Enforce exception transparency in catch/retry/onCompletion operators
  * Optional DSL for `launchIn`, maybe with order enforcement

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
@@ -134,7 +134,7 @@ public fun <T> Flow<T>.onEach(action: suspend (T) -> Unit): Flow<T> = flow {
  *     .onCompletion { println(it) } // Can print exceptions from emitData, computeOne, onCompletion and computeTwo
  * ```
  */
-@ExperimentalCoroutinesApi
+@ExperimentalCoroutinesApi // tentatively stable in 1.3.0
 public fun <T> Flow<T>.onCompletion(action: suspend (cause: Throwable?) -> Unit): Flow<T> = flow {
     // Optimization: this var is used to detect exception from downstream and then as final cause
     var fromDownstream: Throwable? = null
--- a/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
@@ -105,7 +105,8 @@ public fun <T> Flow<T>.onEach(action: suspend (T) -> Unit): Flow<T> = flow {
  * Invokes the given [action] when the given flow is completed or cancelled, using
  * the exception from the upstream (if any) as cause parameter of [action].
  *
- * Conceptually, [onCompletion] is similar to wrapping the flow collection into a `finally` block, for example the following imperative snippet:
+ * Conceptually, [onCompletion] is similar to wrapping the flow collection into a `finally` block,
+ * for example the following imperative snippet:
  * ```
  * try {
  *     myFlow.collect { value ->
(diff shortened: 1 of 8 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
```diff
@@ -100,6 +101,75 @@ public fun <T> Flow<T>.onEach(action: suspend (T) -> Unit): Flow<T> = flow {
...
+ *     .map { computeOne(it) }
+ *     .onCompletion { println(it) } // Can print exceptions from emitData and computeOne
+ *     .map { computeTwo(it) }
+ *     .onCompletion { println(it) } // Can print exceptions from emitData, computeOne, onCompletion and computeTwo
+ * ```
+ */
+@ExperimentalCoroutinesApi // tentatively stable in 1.3.0
+public fun <T> Flow<T>.onCompletion(action: suspend (cause: Throwable?) -> Unit): Flow<T> = flow {
+    // Optimization: this var is used to detect exception from downstream and then as final cause
+    var fromDownstream: Throwable? = null
```
elizarov: Maybe renamed it to simply an `exception`? 

Alternatively, please consider strategies to avoid duplication of this quite a non-trivial code between `catch` and `onCompletion` operators. One way to achieve it is to change `catchImpl` so that it  _returns_ the resulting exception instead of being inline function with a lambda. This avoids repetition of its non-trivial bytecode in `catch`, `retryWhen`, and `onCompletion` operators:

```
internal suspend fun <T> Flow<T>.catchImpl(
    collector: FlowCollector<T>
): Throwable? {
    var fromDownstream: Throwable? = null
    try {
        collect {
            try {
                collector.emit(it)
            } catch (e: Throwable) {
                fromDownstream = e
                throw e
            }
        }
    } catch (e: Throwable) {
        /*
         * First check ensures that we catch an original exception, not one rethrown by an operator.
         * Seconds check ignores cancellation causes, they cannot be caught.
         */
        if (e.isSameExceptionAs(fromDownstream) || e.isCancellationCause(coroutineContext)) {
            throw e
        } else {
            return e // not from downstream
        }
    }
    return null // normal completion
}
```

In this case, both `catch` and `retryWhen` will be slightly refactored. Instead of having their logic in lambda, they'll have to check for the result of `catchImpl` being not null, for example:

```
public fun <T> Flow<T>.catch(action: suspend FlowCollector<T>.(cause: Throwable) -> Unit): Flow<T> =
    flow {
        val cause = catchImpl(this)
        if (cause != null) action(cause)
    }
```

Moreover `onCompletion` could be then written just like that with much smaller code:

```
public fun <T> Flow<T>.onCompletion(action: suspend (cause: Throwable?) -> Unit): Flow<T> = flow {
    var cause: Throwable? = null
    try {
        cause = catchImpl(this)
    } finally {
        // Separate method because of KT-32220
        invokeSafely(action, cause)
        cause?.let { throw it }
    }
}
```

kotlinx-coroutines-core/common/src/flow/terminal/Collect.kt
```diff
@@ -30,6 +30,30 @@ import kotlin.jvm.*
...
+/**
+ * Terminal flow operator that [launches][launch] the [collection][collect] of the given flow in the [scope].
+ * It is a shorthand for `scope.launch { flow.collect() }`.
+ *
+ * This operator is usually used with [onEach], [onCompletion] and [catch] operators to process all emitted values
+ * handle an exception that might occur in the upstream flow or during processing, for example:
+ * ```
+ * flow
+ *     .onEach { value -> updateUi(value) }
+ *     .onCompletion { cause -> if (cause == null) updateUi("Done") }
```
elizarov: Consider the following `onCompletion` to highlight the use-case of always updating ui on any kind of completion and to remove duplicated `updateUi` call from `catch`:
```
.onCompletion { cause -> updateUi(if (cause == null) "Done" else "Fail") }
```

PR #1296, opened 2019-06-25
Added CoroutineScope.asExecutor() to ListenableFuture.kt, and used it in place of directExecutor()

yorickhenning:
This should fix cases where CoroutineDispatchers and ContinuationInterceptors are missing state as a continuation executes. By using directExecutor(), CancellableContinuation adopts directExecutor()'s bad habit of bypassing dispatch for already-complete Futures.

### DIFF
--- a/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
+++ b/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
@@ -40,7 +40,7 @@ public fun <T> CoroutineScope.future(
     val newContext = newCoroutineContext(context)
     val future = SettableFuture.create<T>()
     val coroutine = ListenableFutureCoroutine(newContext, future)
-    Futures.addCallback(future, coroutine, MoreExecutors.directExecutor())
+    Futures.addCallback(future, coroutine, CoroutineScope(newContext).asExecutor())
     coroutine.start(start, coroutine, block)
     return future
 }
@@ -122,7 +122,7 @@ public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
         override fun onFailure(t: Throwable) {
             deferred.completeExceptionally(t)
         }
-    }, MoreExecutors.directExecutor())
+    }, CoroutineScope(deferred).asExecutor())
 
     deferred.invokeOnCompletion { cancel(false) }
     return deferred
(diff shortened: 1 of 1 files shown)


yorickhenning:
It wasn't correct of me to say it's missing context state - I had a misunderstanding about context propagation to child coroutines. _This_ issue is that the continuation sometimes isn't dispatched.

The Coroutine has a dispatcher and an interceptor in its context, and the Future being awaited is from a ListeningExecutorService, not an adapted Coroutine. I'm still working on boiling unit tests out of the failing business logic test cases.

I don't believe it's a race condition.

I _think_ the broken path is the `suspendCancellableCoroutine {}` in `await()`. If the Future is complete at call time, the stack grows and the continuation executes inline without going through `dispatch()` - `directExecutor()` runs the continuation's body inline. [`directExecutor()` turns `addCallback()` into a a synchronous caller-runs execution in this case](https://guava.dev/releases/23.3-jre/api/docs/com/google/common/util/concurrent/ListenableFuture.html#addListener-java.lang.Runnable-java.util.concurrent.Executor-), which has bitten me several times, in creative ways.

qwwdfsad:
>If the Future is complete at call time, the stack grows and the continuation executes inline without going through dispatch() - directExecutor() runs the continuation's body inline.

This is partially true.

> directExecutor() runs the continuation's body inline

Right. You can see that the body of the callback is `continuation.resume`. But this is not the original continuation, this is `CancellableContinuation` retrieved through `suspendCancellableCoroutine`.
`CancellableContinuation.resume` does a proper dispatch

yorickhenning:
Alright, I'll eat crow on this one. The equivalent of an Interceptor in the Executors I'm working with put state into the continuation's thread-local state in some weird ways. The true bug was in my ContinuationInterceptor implementation. Even with pretty extensive tests, I didn't manage to isolate it first time round.

Switching to a non-reentrant continuation-style Executor is probably still a good idea. It'll prevent bugs and make debugging Interceptors a lot easier to reason about. I'm not going to fight for that until we get a production problem, though. :)

I'll circle back with some issues around the Interceptor API. It's got some unexpected sharp edges.

qwwdfsad:
>Switching to a non-reentrant continuation-style Executor is probably still a good idea. It'll prevent bugs and make debugging Interceptors a lot easier to reason about.

If we are talking about continuations and interceptors (not Guava's executors), then they all properly unroll the stack before rereentrance as we dedicated a lot of effort to make `Dispatchers.Unconfined`, `MainDispatcher.immediate`, `CoroutineStart.UNDISPATCHED` to behave this way. If you can trigger SoE using only these interceptors, it's probably a bug worth considering.

PR #1304, opened 2019-06-28
Adds R8 optimization rule for FastServiceLoader

wojtek-kalicinski:
This anticipates a new mechanism in a future
Android Gradle Plugin (3.6.0+) that enables
version targeting of ProGuard/R8 rules.

It does not change any behavior for current users
of the library.

Once the new plugin is used, the R8 optimization
will be read automatically.

### DIFF
--- a/ui/kotlinx-coroutines-android/build.gradle
+++ b/ui/kotlinx-coroutines-android/build.gradle
@@ -4,7 +4,7 @@
 
 repositories {
     google()
-    // TODO Remove once R8 is updated to a 1.5.x version.
+    // TODO Remove once R8 is updated to a 1.6.x version.
     maven {
         url "http://storage.googleapis.com/r8-releases/raw/master"
         metadataSources {
@@ -25,38 +25,64 @@ dependencies {
     testImplementation 'org.robolectric:robolectric:4.0-alpha-3'
     testImplementation 'org.smali:baksmali:2.2.7'
 
-    // TODO Replace with a 1.5.x version once released to maven.google.com.
-    r8 'com.android.tools:r8:878ec25559ca91497517345ec3a6c16102b7025c'
+    // TODO Replace with a 1.6.x version once released to maven.google.com.
+    r8 'com.android.tools:r8:a7ce65837bec81c62261bf0adac73d9c09d32af2'
 }
 
-def optimizedDex = new File(buildDir, "dex/")
-task runR8(type: JavaExec, dependsOn: 'jar') {
-    classpath = configurations.r8
-    main = 'com.android.tools.r8.R8'
+class RunR8Task extends JavaExec {
 
-    doFirst {
-        if (optimizedDex.exists()) {
-            optimizedDex.deleteDir()
-        }
-        optimizedDex.mkdirs()
+    File outputDex
+    File inputConfig
+
+    @Override
+    Task configure(Closure closure) {
+        super.configure(closure)
+        classpath = project.configurations.r8
+        main = 'com.android.tools.r8.R8'
 
         def arguments = [
-            '--release',
-            '--no-desugaring',
-            '--output', optimizedDex.absolutePath,
-            '--pg-conf', file('r8-test-rules.pro').absolutePath
+                '--release',
+                '--no-desugaring',
+                '--output', outputDex.absolutePath,
+                '--pg-conf', inputConfig.absolutePath
         ]
-        arguments.addAll(configurations.runtimeClasspath.files.collect { it.absolutePath })
-        arguments.addAll(jar.archivePath.absolutePath)
+        arguments.addAll(project.configurations.runtimeClasspath.files.collect { it.absolutePath })
+        arguments.addAll(project.jar.archivePath.absolutePath)
 
         args = arguments
+        return this
     }
+
+    @Override
+    void exec() {
+        if (outputDex.exists()) {
+            outputDex.deleteDir()
+        }
+        outputDex.mkdirs()
+
+        super.exec()
+    }
+}
+
+def optimizedDex = new File(buildDir, "dex-optim/")
+def unOptimizedDex = new File(buildDir, "dex-unoptim/")
+
+task runR8(type: RunR8Task, dependsOn: 'jar'){
+    outputDex = optimizedDex
+    inputConfig = file('r8-test-rules.pro')
+}
+
+task runR8NoOptim(type: RunR8Task, dependsOn: 'jar'){
+    outputDex = unOptimizedDex
+    inputConfig = file('r8-test-rules-no-optim.pro')
 }
 
 test {
     // Ensure the R8-processed dex is built and supply its path as a property to the test.
     dependsOn(runR8)
+    dependsOn(runR8NoOptim)
     systemProperty 'dexPath', new File(optimizedDex, "classes.dex").absolutePath
+    systemProperty 'noOptimDexPath', new File(unOptimizedDex, "classes.dex").absolutePath
 }
 
 tasks.withType(dokka.getClass()) {
--- a/ui/kotlinx-coroutines-android/build.gradle
+++ b/ui/kotlinx-coroutines-android/build.gradle
@@ -31,8 +31,17 @@ dependencies {
 
 class RunR8Task extends JavaExec {
 
+    @OutputDirectory
     File outputDex
+
+    @InputFile
     File inputConfig
+
+    @InputFile
+    final File inputConfigCommon = new File('r8-test-common.pro')
+
+    @InputFiles
+    final File jarFile = project.jar.archivePath
 
     @Override
     Task configure(Closure closure) {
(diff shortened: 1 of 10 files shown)


wojtek-kalicinski:
actually, this still doesn't work with newest R8 in full mode for me,
@JakeWharton I lost our chat messages, do you remember what the problem was so I can file a bug?

JakeWharton:
The presence of the `-keepnames` rules in core are currently preventing the optimization from applying. Those are only needed for ProGuard or for R8 versions prior to 1.5.

wojtek-kalicinski:
> The presence of the `-keepnames` rules in core are currently preventing the optimization from applying. Those are only needed for ProGuard or for R8 versions prior to 1.5.

I think there's more to it than that, I summarized my findings in [this R8 bug](https://b.corp.google.com/issues/136452086). tl;dr: I think vertical class merging is preventing the optimization from working, but only on release builds in full mode.

Let's wait with this pull request until we have a fix from R8 or other guidance on how to make sure the optimization is applied in both non-full and full mode, on debug and release builds. I'll post updates here.

wojtek-kalicinski:
By the way, the only way I found that lets me consistently apply the optimization currently is if I add a dummy class to prevent vertical class merging:

```
@Keep
class DummyDispatcherFactory : MainDispatcherFactory {
    override val loadPriority: Int
        get() = TODO("not implemented")
    override fun createDispatcher(allFactories: List<MainDispatcherFactory>): MainCoroutineDispatcher {
        TODO("not implemented")
    }
}
```

I'm not sure if that's something you'd want in the codebase though.

elizarov:
Any update?

wojtek-kalicinski:
So far the R8 team acknowledged the bug but no ETA for a fix yet, so I don't know which version of R8 will have it.
I want to be sure that the optimization works for everyone (from a certain version of R8) before we enable it. We also have to wait for Android Gradle Plugin 3.6.0 for this to work, so maybe there will be more coroutines releases before that goes stable.

Do you want me to close this PR for now and create a new one in the future once we have everything in place?

elizarov:
Keep it, please. I'm just curious what the strategy I shall use to optimize the size of coroutines on android given the last of ability to target specific toolchain versions.

wojtek-kalicinski:
This is now ready for review.
I had to remove the `@Keep` from `AndroidDispatcherFactory`, but added a corresponding rule for proguard config and r8 config prior to version 1.6.

elizarov:
That looks good by itself. Now we need to consolidate it with #1245 by @JakeWharton to make sure those rules are actually tested so that we get a signal if we break/rename something in the future. We do need some kind of `-checkdiscard` rules that are automatically verified on build.

wojtek-kalicinski:
How about this?
I integrated Jake's test, and added more on top. We now check for:
- if the FastServiceLoader and ServiceLoader calls are removed with new R8 versions
- if the AndroidDispatcherFactory is kept with older R8/PG versions
- if the rules applied between older PG/R8 and AGP >= 3.6.0 / < 3.6.0 are the same, so that no one forgets to update all 3 files

ui/kotlinx-coroutines-android/test/R8ServiceLoaderOptimizationTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.jf.dexlib2.iface.ClassDef
+import org.jf.dexlib2.iface.DexFile
+import org.jf.util.IndentingWriter
+import org.junit.Test
+import java.io.File
+import java.io.StringWriter
+import java.util.stream.Collectors
+import kotlin.test.assertEquals
+
+class R8ServiceLoaderOptimizationTest {
```
elizarov: Can you, please reformat it to Kotlin Libraries code style:
It is basically standard Kotlin style (4 spaces, etc), but we also use `*` imports always.

ui/kotlinx-coroutines-android/test/R8ServiceLoaderOptimizationTest.kt
```diff
@@ -0,0 +1,69 @@
...
+  private val r8Dex = File(System.getProperty("dexPath")!!).asDexFile()
+  private val r8DexNoOptim = File(System.getProperty("noOptimDexPath")!!).asDexFile()
+
+  @Test fun noServiceLoaderCalls() {
+    val serviceLoaderInvocations = r8Dex.classes
+        // Create a map of type names to a list of their ServiceLoader invocations.
+        .associate { clz ->
+          clz.type to clz.toSmali().lines().filter { "Ljava/util/ServiceLoader;->" in it }
```
elizarov: It looks like the only test you are doing here is for textual occurence of the the service loader class. It seems that bringing the whole decompilation library here is a bit of overkill. May I suggest to simply scan the resulting classes for the corresponding string similarly to how we do it in this test: https://github.com/Kotlin/kotlinx.coroutines/blob/master/binary-compatibility-validator/test/PublicApiTest.kt#L78

ui/kotlinx-coroutines-android/r8-test-rules.pro
```diff
@@ -0,0 +1,19 @@
+# Entry point for retaining MainDispatcherLoader which uses a ServiceLoader.
```
elizarov: It looks like this rules file is mostly a copy of `r8-test-rules-no-optim.pro` with added discard check. Can we somehow make it reference the former file to ease maintenance in the future when we'll be adding more optimization rules and testing for them?

ui/kotlinx-coroutines-android/r8-test-rules.pro
```diff
@@ -0,0 +1,19 @@
...
+}
+
+# Ensure the custom, fast service loader implementation is removed. In the case of fast service
+# loader encountering an exception it falls back to regular ServiceLoader in a way that cannot be
+# optimized out by R8.
+-include resources/META-INF/com.android.tools/r8-min-1.6.0/coroutines.pro
+-checkdiscard class kotlinx.coroutines.internal.FastServiceLoader
+
+# We are cheating a bit by not having android.jar on R8's library classpath. Ignore those warnings.
+-ignorewarnings
```
elizarov: It seems that the warning are still printed on the console during build. Is there any way to suppress them?

ui/kotlinx-coroutines-android/build.gradle
```diff
@@ -12,6 +23,66 @@ dependencies {
...
+    testImplementation 'org.smali:baksmali:2.2.7'
+
+    // TODO Replace with a 1.6.x version once released to maven.google.com.
+    r8 'com.android.tools:r8:a7ce65837bec81c62261bf0adac73d9c09d32af2'
+}
+
+class RunR8Task extends JavaExec {
+
+    File outputDex
+    File inputConfig
```
elizarov: It seems that those files shall marked with Input/Ouput gradle annotations, because otherwise this task is not incremental and is being rerun every time even if nothing has changed.

ui/kotlinx-coroutines-android/r8-test-rules.pro
wojtek-kalicinski: I don't think so, unfortunately. If I disable the ErrorStream for the task we will lose the actual errors that we care about in case of failure.

ui/kotlinx-coroutines-android/test/R8ServiceLoaderOptimizationTest.kt
wojtek-kalicinski: We don't deal with classes in this case, as R8 outputs DEX (and that's what we need to check, as there are differences between optimizations that are applied if I were to switch R8 to output .class files). 
So it looks like being able to parse and read DEX files is essential here. 
I did simplify the test a lot though to not have to generate decompiled source Strings, I simply check if a reference to the type is present in the file.

JakeWharton: The reason that I chose the smali string representation is that the failure mode is much more useful than when you're querying directly. You're given the actual problems directly rather than having to go investigate separately.

elizarov:
Looks good now. One final problem, though. I've tested it with Android Studio `3.4.1` and Android Gradle Plugin version `3.4.0` and I see that all the `META-INF\xxx\*.pro` files end up in the resulting .apk file. Is it fine?

JakeWharton:
We'll have to add them to the automatic excludes (like we did with
META-INF/proguard/) if they weren't already added. They won't be
automatically excluded until 3.6 though. Otherwise there's no harm in them
landing in the APK for now.

wojtek-kalicinski:
Yeah, on older AGP we can suggest a rule to add to build.gradle manually to exclude the files, but in any case it's only a few hundred bytes. On 3.6.0+ once we have the new rules selection (it's not out in canary yet) the files will not be packaged.

wojtek-kalicinski:
The feature to read targeted rules was [merged](https://android.googlesource.com/platform/tools/base/+/9a0c2026d99f3c138fafe269c12259e4ac2aeb0f) into Android Studio and will be available to test in AS 3.6 Canary 7 in around two weeks

elizarov:
:tada: Squashed a merged to `develop`. Thank a lot: https://github.com/Kotlin/kotlinx.coroutines/commit/d6b0b0f90c601f86b2ac4030c02af4f4c4562ac1

PR #1305, opened 2019-06-28
 Consistently handle fatal exceptions in reactive integrations

### LINKED ISSUE #1297: RxJava2 "fatal" exceptions thrown by subscribers are swallowed by rxFlowable
This issue occurs under the following conditions:
1. RxJava2 `Flowable` created using the `rxFlowable` coroutine builder. Note that `rxObservable` is *not* affected.
2. The resulting `Flowable` is multicasted using either `replay` or `publish`. Method of connection doesn't matter (can use either `autoConnect`, `refCount`, or connect manually with `connect`).
3. `Flowable` is subscribed with single-argument `subscribe`, taking a `Consumer`, and the consumer throws a "fatal" exception, as defined by RxJava2's `Exceptions.throwIfFatal` method. Note that non-fatal exceptions get wrapped in `OnErrorNotImplemented` exceptions and re-thrown, as expected.

**Expected behavior:** The exception gets thrown as an `OnErrorNotImplementedException`, or at least reported to the default uncaught exception handler.
**Observed behavior:** The exception is swallowed without being reported in any way.

The context for the `rxFlowable` builder can be empty (`GlobalScope` + `EmptyCoroutineContext`), or use the `Unconfined` dispatcher, the issue occurs in both cases.

I have a Gradle project that reproduces the issue in the simplest form I could whittle it down to: https://github.com/zach-klippenstein/coroutines-rxflowable-bug-repo (see [Main.kt](https://github.com/zach-klippenstein/coroutines-rxflowable-bug-repo/blob/master/src/main/kotlin/Main.kt)).

qwwdfsad:
Fixes #1297

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/Publish.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Publish.kt
@@ -160,11 +160,11 @@ private class PublisherCoroutine<in T>(
         }
         // now update nRequested
         while (true) { // lock-free loop on nRequested
-            val cur = _nRequested.value
-            if (cur < 0) break // closed from inside onNext => unlock
-            if (cur == Long.MAX_VALUE) break // no back-pressure => unlock
-            val upd = cur - 1
-            if (_nRequested.compareAndSet(cur, upd)) {
+            val current = _nRequested.value
+            if (current < 0) break // closed from inside onNext => unlock
+            if (current == Long.MAX_VALUE) break // no back-pressure => unlock
+            val upd = current - 1
+            if (_nRequested.compareAndSet(current, upd)) {
                 if (upd == 0L) {
                     // return to keep locked due to back-pressure
                     return
--- a/reactive/kotlinx-coroutines-reactive/src/Publish.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Publish.kt
@@ -198,17 +198,32 @@ private class PublisherCoroutine<in T>(
                 if (cancelled) {
                     // If the parent had failed to handle our exception, then we must not lose this exception
                     if (cause != null && !handled) handleCoroutineException(context, cause)
-                } else {
-                    try {
-                        if (cause != null && cause !is CancellationException) {
+                    return
+                }
+
+                try {
+                    if (cause != null && cause !is CancellationException) {
+                        /*
+                         * Reactive frameworks have two types of exceptions: regular and fatal.
+                         * Regular are passed to onError.
+                         * Fatal can be passed to onError, but implementation **is free to swallow it** (e.g. see #1297).
+                         * Such behaviour is inconsistent, leads to silent failures and we can't possibly know whether
+                         * the cause will be handled by onError (and moreover, it depends on whether a fatal exception was
+                         * thrown by subscriber or upstream).
+                         * To make behaviour consistent and least surprising, we always handle fatal exceptions
+                         * by coroutines machinery, anyway, they should not be present in regular program flow,
+                         * thus our goal here is just to expose it as soon as possible.
+                         */
+                        if (cause.isFatal()) {
+                            if (!handled) handleCoroutineException(context, cause)
+                        } else {
                             subscriber.onError(cause)
                         }
-                        else {
-                            subscriber.onComplete()
-                        }
-                    } catch (e: Throwable) {
-                        handleCoroutineException(context, e)
+                    } else {
+                        subscriber.onComplete()
                     }
+                } catch (e: Throwable) {
+                    handleCoroutineException(context, e)
                 }
             }
         } finally {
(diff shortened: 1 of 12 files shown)


reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -250,4 +265,6 @@ private class PublisherCoroutine<in T>(
         cancelled = true
         super.cancel(null)
     }
+
+    private fun Throwable.isFatal() = this is VirtualMachineError || this is ThreadDeath || this is LinkageError
```
zach-klippenstein: This set is specific to RxJava2 – RxJava1 had different behavior, and other reactive frameworks might as well. A [comment in the RxJava2 source](https://github.com/ReactiveX/RxJava/blob/cc7fce89c46e67c4a124da9e206e4c770b6f8850/src/main/java/io/reactivex/exceptions/Exceptions.java#L67) says [this set was taken from Scala's Futures library](https://github.com/ReactiveX/RxJava/issues/748#issuecomment-32471495).

It's probably a reasonable default, but I'm not sure how this will interact with other reactive libraries that implement `Publisher`.

reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
```diff
@@ -164,4 +184,6 @@ private class RxObservableCoroutine<T: Any>(
     override fun onCancelled(cause: Throwable, handled: Boolean) {
         signalCompleted(cause, handled)
     }
-}
\ No newline at end of file
+}
+
+internal fun Throwable.isFatal() = this is VirtualMachineError || this is ThreadDeath || this is LinkageError
```
zach-klippenstein: It's less efficient, but to avoid duplicating this logic with RxJava2's `throwIfFatal`, you could do
```kotlin
internal fun Throwable.isFatal() = try {
  Exceptions.throwIfFatal(this)
  true
} catch (e: Throwable) {
  false
}
```

Since this is only called when handling an exception that has already been determined to not be a `CancellationException`, that overhead is probably not too much of a concern.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -176,17 +176,32 @@ private class PublisherCoroutine<in T>(
...
-                } else {
-                    try {
-                        if (cause != null && cause !is CancellationException) {
+                    return
+                }
+
+                try {
+                    if (cause != null && cause !is CancellationException) {
+                        /*
+                         * Reactive frameworks have two types of exceptions: regular and fatal.
```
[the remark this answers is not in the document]
qwwdfsad: Reactor does as well

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -176,17 +176,32 @@ private class PublisherCoroutine<in T>(
...
+                         * Regular are passed to onError.
+                         * Fatal can be passed to onError, but implementation **is free to swallow it** (e.g. see #1297).
```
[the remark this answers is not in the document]
qwwdfsad: This is a tricky part in Rx that is not properly documented.
But the reported test-case (thanks for that!) exposes the fact, that different **standard** subscribers implementation behave inconsistently in the face of fatal exception.
We, from our side, can only treat it as a black box and provide the most consistent (and least surprising) behaviour.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
qwwdfsad: Well, there an extended set in reactor, I will separately take care of it during #1138

qwwdfsad:
Ping @elizarov

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -176,17 +176,32 @@ private class PublisherCoroutine<in T>(
...
+                         * Such behaviour is inconsistent, leads to silent failures and we can't possibly know whether
+                         * the cause will be handled by onError (and moreover, it depends on whether a fatal exception was
+                         * thrown by subscriber or upstream).
+                         * To make behaviour consistent and least surprising, we always handle fatal exceptions
+                         * by coroutines machinery, anyway, they should not be present in regular program flow,
+                         * thus our goal here is just to expose it as soon as possible.
+                         */
+                        if (cause.isFatal()) {
```
elizarov: I'm worried that this `if` makes our implementation non compliant with the spec as we are not even trying to deliver those exceptions to the `subscriber.onError` as specification requires us to do. We don't terminate it in any way! We should first always _attempt_ to deliver the error to `subscriber.onError(cause)` and then, if the error if fatal, knowing that some reactive specs swallow then do `if (!handled && cause.isFatal()) handleCoroutineException(context, cause)` to make sure it is not lost.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -138,11 +138,11 @@ private class PublisherCoroutine<in T>(
...
-            val cur = _nRequested.value
-            if (cur < 0) break // closed from inside onNext => unlock
-            if (cur == Long.MAX_VALUE) break // no back-pressure => unlock
-            val upd = cur - 1
-            if (_nRequested.compareAndSet(cur, upd)) {
+            val current = _nRequested.value
+            if (current < 0) break // closed from inside onNext => unlock
+            if (current == Long.MAX_VALUE) break // no back-pressure => unlock
+            val upd = current - 1
+            if (_nRequested.compareAndSet(current, upd)) {
```
elizarov: Style nitpick here. `current`/`upd` pair does not look consistent anymore. Either leave as `cur`/`upd` or renamed to `current`/`updated`.

reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
```diff
@@ -131,16 +135,32 @@ private class RxObservableCoroutine<T: Any>(
...
+                        * thrown by subscriber or upstream).
+                        * To make behaviour consistent and least surprising, we always handle fatal exceptions
+                        * by coroutines machinery, anyway, they should not be present in regular program flow,
+                        * thus our goal here is just to expose it as soon as possible.
+                        */
+                        if (cause.isFatal()) {
+                            if (!handled) handleCoroutineException(context, cause)
+                        } else {
+                            subscriber.onError(cause)
+                        }
```
elizarov: Same concern in this RxObservable code about always calling `onError`.

reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
elizarov: Seems this is an Rx-specific piece of code in `RxObservable.kt` it seems sensible to go via Rx `throwIfFatal`.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
qwwdfsad: Makes sense

PR #1306, opened 2019-06-28
Add experimental contracts for withContext and coroutineScope

### LINKED ISSUE #1257: Add a contract to coroutineScope, withContext and operator fun invoke for CoroutineDispatcher
Now that there is an EAP kotlinx.coroutines version for 1.3.40 EAP, and since 1.3.40 supports contracts for suspending functions, we can add contracts for the lambdas of the following functions in this library:
- `coroutineScope`
- `withContext`
- `invoke` (extension function for `CoroutineDispatcher`)

I can make a PR on the 1.3.40-eap branch.

LouisCAD:
Resolves #1257

Note that I couldn't implement a contract for the operator function `invoke` for `CoroutineDispatcher` because these are not currently allowed for operator functions.
I requested the feature and asked why in [KT-32313](https://youtrack.jetbrains.com/issue/KT-32313).

### DIFF
--- a/gradle/experimental.gradle
+++ b/gradle/experimental.gradle
@@ -5,6 +5,7 @@
 // For new mpp
 ext.experimentalAnnotations = [
         "kotlin.Experimental",
+        "kotlin.contracts.ExperimentalContracts",
         "kotlin.experimental.ExperimentalTypeInference",
         "kotlin.ExperimentalMultiplatform",
         "kotlinx.coroutines.ExperimentalCoroutinesApi",
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -11,6 +11,8 @@ import kotlinx.atomicfu.*
 import kotlinx.coroutines.internal.*
 import kotlinx.coroutines.intrinsics.*
 import kotlinx.coroutines.selects.*
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
 import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
 import kotlin.jvm.*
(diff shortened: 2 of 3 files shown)


LouisCAD:
Cancellation means an exception is thrown, so you can't get past `withContext` call if cancellation happened. Consequently, `EXACTLY_ONCE` is the right contract for the parameter `block`, just like it is for [`apply` from the stdlib](https://github.com/JetBrains/kotlin/blob/4c91ba90bb2f6437d0e844cfd3b0860f4d0f251c/libraries/stdlib/src/kotlin/util/Standard.kt#L79-L85), where an exception could be thrown from inside the lambda.

For the build issue, I didn't expect it to trigger a compilation failure, but the CI did. Could reproduce locally, but used the handy link of the CI run in the issue report: https://youtrack.jetbrains.com/issue/KT-32322

I'm hoping 1.3.41 can fix it 🙏

zach-klippenstein:
I think it needs to be `AT_MOST_ONCE` so that this code won't compile:

```kotlin
val result: String
try {
  withContext(Cancelled) {
    result = "42"
  }
  // EXACTLY_ONCE guarantees that result has
  // been initialized here, but it won't be if 
  // the coroutine was already cancelled, and the
  // compiler should not let you read it.
  result.toInt()
}
```

LouisCAD:
No, EXACTLY_ONCE is the right one. This code won't compile anyway,
regardless of the contract (unless new capabilities are added and the
contract is changed the wrong way using those).

The contract is specified ~ this way: this function returning (not
throwing) implies that the passed lambda has been executed. After an
enclosing try block, you're out of the control flow scope, so there's no
way you can access an unitialized val there.

qwwdfsad:
Closing this PR while the corresponding issue in Kotlin is not fixed.
Pinged people responsible for this fix

PR #1308, opened 2019-06-28
Fix _size not incremented in CopyOnWriteList add

### LINKED ISSUE #1307: _size private property is not incremented in add w/ index in native CopyOnWriteList
In the `CopyOnWriteList` class for Kotlin/Native, the `_ size` private property is incremented in the `add(element: E): Boolean` function on line `16`:
https://github.com/Kotlin/kotlinx.coroutines/blob/7acaae66240e5fe5a0ebe9a2cea7bb971f078347/kotlinx-coroutines-core/native/src/internal/CopyOnWriteList.kt#L13-L19

but it is not in the `add(index: Int, element: E)` function:
https://github.com/Kotlin/kotlinx.coroutines/blob/7acaae66240e5fe5a0ebe9a2cea7bb971f078347/kotlinx-coroutines-core/native/src/internal/CopyOnWriteList.kt#L21-L28

I believe this is a bug.

LouisCAD:
FIxes #1307


LouisCAD:
Fixed conflict after manual merging of #1309

PR #1315, opened 2019-07-03
Add withLatestFrom operator.

zach-klippenstein:
**Use case:** The processing of one flow (the operator's receiver) requires using the latest value emitted from another flow at the time of emission from the first flow. My specific use case involves combining a flow of events with a flow of event sinks (`(Event) -> Unit`). Whenever an event occurs (is emitted from the first flow), it should be sent to the _most recent_ sink emitted from the second flow.

The resulting flow is entirely driven by the first flow. This operator conflates and caches values from the second flow and passes them to a lambda to be combined with values from the first flow whenever the first flow emits. The resulting flow will start collecting both flows immediately, but will not call the transform function or emit anything until both flows emit at least one value. The resulting flow will remain active until either flow throws an exception, or the first flow completes. If the first flow completes before the second, the second flow is cancelled. If the second flow completes before the first, the latest value will continue to be cached and passed to the transform function every time the first flow emits.

This operator is similar to `combineLatest`, but it always emits _immediately_ and _only_ when the first flow emits (values from the first flow are not cached). The implementation in this PR reuses as much of the infrastructure as possible from `combineLatest`. It collects the second flow using a `CONFLATED` channel since the operator only cares about caching the most recently emitted value, so the second flow does not need to have any backpressure applied.

This PR only includes a single overload of the operator, accepting a single other flow, since that is the only overload I have a concrete use case for.


kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
```diff
@@ -174,6 +175,53 @@ internal fun <T, R> Flow<T>.combineLatest(vararg others: Flow<T>, arrayFactory:
     }
 }
 
+/**
+ * Returns a [Flow] whose values are generated with [transform] function every time this flow emits by combining
+ * the value with the latest value emitted by `other`. Emissions are only triggered by this flow, not `other.
```
rjrjr: Missing close ` after _other_.

kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
```diff
@@ -174,6 +175,53 @@ internal fun <T, R> Flow<T>.combineLatest(vararg others: Flow<T>, arrayFactory:
...
+ * It does not call [transform], and the returned flow will not emit, until `other` emits – if it does not emit, the
```
rjrjr: _if it does not emit_ -- unclear at this point what "it" refers to.

rjrjr (approved):

kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
zach-klippenstein: Agreed this wording is confusing. Not sure if this is much better:

Returns a [Flow] whose values are generated by the [transform] function every time this flow emits by combining the value with the latest value emitted by `other`. Emissions are only triggered by this flow, not `other`, and the returned flow will remain active as long as this flow is, and only complete when this flow completes. The `other` flow will be cancelled if this flow completes first. The operator does not call [transform], and the returned flow will not emit, until `other` emits – if `other` does not emit, the returned flow will complete without emitting. However after `other` has emitted it's first value, the returned flow will ignore `other`'s completion and continue to cache the last value it emitted.

kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
zach-klippenstein: Fixed.

kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
rjrjr: Minor tweaks:

Returns a [Flow] whose values are generated by the [transform] function every time this flow emits by combining the value with the latest value emitted by other. Emissions are only triggered by this flow, not other, and the returned flow will remain active as long as this flow is, completing only when this flow completes. The other flow will be cancelled if this flow completes first. 

The operator does not call [transform], and the returned flow will not emit, until other emits – if other does not emit, the returned flow will complete without emitting. However after other has emitted its first value, the returned flow will ignore other's completion and continue to cache the last value it emitted.

zach-klippenstein: Updated.

zach-klippenstein:
Gentle ping: Is the use case I've given adequate, or would more information be helpful?

elizarov:
This will not be part of `1.3.0` release. We'll look at when planning for the next batch of operations in subsequent versions.

hrach:
Do I correctly understand that this is `combineTransform` with `Latest` extension? Actually, I'd love to see this.  

My usecase is for example: loading places to map; user is moving with map, which cancels the previous fetching, also the loaded data depend on filter or other Flows.

elizarov:
A simple implementation using only stable Flow APIs can be found here: https://pl.kotl.in/IYfZx_sKY

curioustechizen:
I'm having trouble writing unit tests for the implementation in https://github.com/Kotlin/kotlinx.coroutines/pull/1315#issuecomment-525338743

Here's what I've tried.

```kotlin
@Test
fun `When receiver flow emits then resulting flow emits`()  {

    val firstFlowChannel = ConflatedBroadcastChannel<Char>()
    val otherFlowChannel = ConflatedBroadcastChannel<Int>()

    val resultingFlow =
        firstFlowChannel.asFlow()
            .withLatestFrom(otherFlowChannel.asFlow()) { a, b -> "$a$b"}

    otherFlowChannel.offer(1)
    firstFlowChannel.offer('a')

    runBlocking {
        val receiveChannel = resultingFlow.produceIn(this)
        assertEquals("a1", receiveChannel.poll())
        receiveChannel.cancel()
    }
}
```

This always fails with 
>java.lang.AssertionError: 
Expected :a1
Actual   :null

I see `receiveChannel` is empty. If I try to add breakpoints it looks like the `withLatestFrom` is not even being called. I suspect I've misconfigured the test but I cannot spot the problem.

elizarov:
@curioustechizen Don't `poll` the receive channel. Use `receive`.

curioustechizen:
@elizarov Thanks. That worked. If I change `assertEquals(true, receiveChannel.isEmpty)` the test passes.

Do I understand correctly that `flow.produceIn()` does not actually trigger the processing; but rather it is `channel.receive()` that triggers it?

zach-klippenstein:
@curioustechizen  `produceIn` (may) launch a coroutine to collect upstream, but that coroutine probably won't actually start executing before `produceIn` returns and you call `poll` immediately on the next line (depending on dispatcher and thread scheduling details). If it doesn't, `poll` will return null.

PR #1316, opened 2019-07-03
Replace unneeded type parameter with projection

LouisCAD:
Type parameter of awaitClose is not used but produces a warning in
new type inference when used with builder inference as in callbackFlow
as seen in this issue: https://youtrack.jetbrains.com/issue/KT-32097

Using star projection removes the warning and is binary compatible
thanks to generics type erasure.


elizarov:
Thanks a lot!

PR #1320, opened 2019-07-04
Use the correct builder method `flowChannel` in the checkContext error description

PaulWoitaschek:
As there is no `channelFlow`


qwwdfsad:
Thank you

PR #1323, opened 2019-07-05
Semaphore cancellation bug

ndkoval:
Cancellation in `Semaphore` should resume the next waiting acquirer if and only if the canceled request has been resumed under a race. The current implementation _always_ resumes the next waiting acquirer, so that it violates the semaphore contract.

### DIFF
--- a/kotlinx-coroutines-core/common/src/sync/Semaphore.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Semaphore.kt
@@ -121,12 +121,14 @@ private class SemaphoreImpl(
     }
 
     override fun release() {
-        val p = _availablePermits.getAndUpdate { cur ->
-            check(cur < permits) { "The number of acquired permits cannot be greater than `permits`" }
-            cur + 1
-        }
+        val p = incPermits()
         if (p >= 0) return // no waiters
         resumeNextFromQueue()
+    }
+
+    internal fun incPermits() = _availablePermits.getAndUpdate { cur ->
+        check(cur < permits) { "The number of acquired permits cannot be greater than `permits`" }
+        cur + 1
     }
 
     private suspend fun addToQueueAndSuspend() = suspendAtomicCancellableCoroutine<Unit> sc@ { cont ->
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -181,11 +183,10 @@ private class SemaphoreSegment(id: Long, prev: SemaphoreSegment?): Segment<Semap
     @Suppress("NOTHING_TO_INLINE")
     inline fun cas(index: Int, expected: Any?, value: Any?): Boolean = acquirers[index].compareAndSet(expected, value)
 
-    inline fun getAndUpdate(index: Int, function: (Any?) -> Any?): Any? {
+    inline fun getAndUpdate(index: Int, value: Any?): Any? {
```
elizarov: This function need not be `inline` anymore and shall probably be renamed to `getAndSet`. Moreover, you don't need `cas` to implement it. You can directly call `acquirers[index].getAndSet(value)`, which is more efficient.

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -194,9 +195,7 @@ private class SemaphoreSegment(id: Long, prev: SemaphoreSegment?): Segment<Semap
 
     // Cleans the acquirer slot located by the specified index
     // and removes this segment physically if all slots are cleaned.
-    fun cancel(index: Int) {
-        // Clean the specified waiter
-        acquirers[index].value = CANCELLED
+    fun cancel(index: Int): Boolean = getAndUpdate(index, CANCELLED) !== RESUMED.also {
```
elizarov: With `also { }` this function now has the following order of operations:
1.  `cancelledSlots.incrementAndGet()` & `remove()`
2. `getAndUpdate(index, CANCELLED)`

Is it a bug (need a test to check it) or is it also Ok to do in this order? 
Anyway, please rewrite in a more explicit way.

ndkoval: I think it works, but I intended to perform `getAndUpdate` at first; otherwise, it is very complicated to prove that the code is correct.

Issue #1333, opened 2019-07-12
BroadcastChannel.asFlow() may leak memory


xsveda:
In our Android app we are broadcasting Activity Lifecycle Events:
```
sealed class ActivityEvent {
    data class ActivityCreate(val activity: Activity) : ActivityEvent()
    data class ActivityStart(val activity: Activity) : ActivityEvent()
    data class ActivityResume(val activity: Activity) : ActivityEvent()
    data class ActivityPause(val activity: Activity) : ActivityEvent()
    data class ActivityStop(val activity: Activity) : ActivityEvent()
    data class ActivityDestroy(val activity: Activity) : ActivityEvent()
}
```

The `ReceiveChannel` based implementation works fine:
```
class ActivityEvents {
    private val eventsChannel = BroadcastChannel<ActivityEvent>(1)
    fun post(event: ActivityEvent) { eventsChannel.offer(event) }
    fun get(): ReceiveChannel<ActivityEvent> = eventsChannel.openSubscription()
}
-----------------------------------------------------------------
events.get().consumeEach { event ->
    ...
}
```

When we try to expose the events as flow:
```
class ActivityEvents {
    private val eventsChannel = BroadcastChannel<ActivityEvent>(1)
    fun post(event: ActivityEvent) { eventsChannel.offer(event) }
    fun get(): Flow<ActivityEvent> = eventsChannel.asFlow()          // <-------- asFlow()
}
-----------------------------------------------------------------
events.get().collect { event ->
    ...
}
```
and user moves the app to background by tapping on system Back button, [LeakCanary](https://github.com/square/leakcanary) detects a memory leak as the last event `ActivityDestroy` is kept in the `BroadcastChannel.asFlow()` implementation.
LeakTrace:
```
LeakingInstance(referenceKey=d105ecdc-728d-455c-bbfa-761e1846357d, referenceName=, instanceClassName=app.system.MainActivity, watchDurationMillis=5348, retainedDurationMillis=309, exclusionStatus=null, leakTrace=
┬
├─ dalvik.system.PathClassLoader
│    Leaking: NO (it's a GC root and A ClassLoader is never leaking)
│    ↓ PathClassLoader.runtimeInternalObjects
├─ java.lang.Object[]
│    Leaking: NO (ActivityModuleKt↓ is not leaking)
│    ↓ array Object[].[1582]
├─ app.activity.di.ActivityModuleKt
│    Leaking: NO (a class is never leaking)
[...]

PR #1343, opened 2019-07-16
Consume as flow

elizarov:


kotlinx-coroutines-core/common/test/flow/channels/ChannelBuildersFlowTest.kt
```diff
@@ -9,6 +9,51 @@ import kotlinx.coroutines.channels.*
...
+        val channel = produce(NonCancellable) { // otherwise failure will cancel scope as well
+            repeat(10) {
+                send(it + 1)
+            }
+            throw TestException()
+        }
+        val flow = channel.consumeAsFlow()
+        assertFailsWith<TestException> { flow.sum() }
+        assertFailsWith<IllegalStateException> { flow.collect() }
+        finish(2)
```
qwwdfsad: not sure why these `expect(1)` + `finish(2)` are needed

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -242,6 +243,19 @@ public interface ReceiveChannel<out E> {
...
+     * and [cancels][cancel] the channel after the execution of the block.
+     * If you need to iterate over the channel without consuming it, a regular `for` loop should be used instead.
+     *
+     * This function provides a more efficient shorthand for `consumeEach { value -> collector.emit(value) }`.
+     * See [consumeEach].
+     *
+     * @suppress **This an internal API and should not be used from general code.**
+     */
+    @InternalCoroutinesApi
+    public suspend fun consumeEachTo(collector: FlowCollector<E>)
```
qwwdfsad: I wonder why we need such intrusive method for `@FlowPreview` API.
We can write `cosumeAsFlow` on top of an iterator with nulling-out the slot to workaround #1333.
I realize that such an approach is a bit slower, but at least it does not require to add a new internal method to already overloaded channels

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -663,6 +640,57 @@ internal abstract class AbstractChannel<E> : AbstractSendChannel<E>(), Channel<E
...
+                // :KLUDGE: This "run" call is resolved to an extension function "run" and forces the size of
+                // spilled state to increase by an additional slot, so there are 4 object local variables spilled here
+                // which makes the size of spill state equal to the 4 slots that are spilled around subsequent "emit"
+                // call, ensuring that the previously emitted value is not retained in the state while receiving
+                // the next one.
+                //     L$0 <- this
+                //     L$1 <- collector
+                //     L$2 <- cause
+                //     L$3 <- this$run (actually equal to this)
+                val result = run { receiveInternal() }
```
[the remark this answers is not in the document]
elizarov: Tried it, but it does not work. It does not spill the old result before call `receveInternal` even if it a `var` (it knows it is going to be overwritten anyway on resume), so setting it to `null` offers no benefit.

kotlinx-coroutines-core/common/test/flow/channels/ChannelBuildersFlowTest.kt
elizarov: Not really needed. Just for consistency.

kotlinx-coroutines-core/common/src/channels/Channel.kt
elizarov: For a fix to #1333 it can be indeed just an extension. However, converting to/from channels seems to be important part of a flow API surface and writing it in such a way offers better performance as it avoids iterator altogether. Also, the whole state machine and spilling around iterator-using code is much more complicated and the resulting code would be even more kludgy. 

However, there are other way to do the fix. An alternative but also quite performant way would be to introduce `ReceiveOrClosed` (even as experimental/iternal API that would be good here), and then `consumeEachTo` can be written as extension in the flow package. Can we resurrect PR #762 ?

zach-klippenstein: I would selfishly love this alternative, I've got one or two other use cases for `ReceiveOrClosed` that I'm working around awkwardly right now.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +323,82 @@ public interface ReceiveChannel<out E> {
...
+
+    /**
+     * Returns received value if this element represents received value or `null` otherwise.
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val valueOrNull: T?
+        get() = if (isClosed) null else holder as T
+
+    @Suppress("UNCHECKED_CAST")
+    public val valueOrThrow: T
```
zach-klippenstein: How is this actually different from `value` in the closed case? It's possible to figure out that the exceptions are different from reading the code, but would be helpful to have that document in the kdoc. Even reading the code, I'm not sure when I'd use this property instead of just `value`.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +323,82 @@ public interface ReceiveChannel<out E> {
...
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val value: T
+        get() = if (isClosed) throw IllegalStateException() else holder as T
+
+    /**
+     * Returns received value if this element represents received value or `null` otherwise.
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val valueOrNull: T?
```
zach-klippenstein: Nit: If this were an extension function on `<T: Any> ReceiveResult<T>` (i.e. when `T` is non-nullable), it would prevent the mistake of conflating closed with null, which this class is meant to solve in the first place.  Is that too restrictive? Given that `value` already does the safe thing anyway, I guess this verbose name makes it really clear you're acknowledging that.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +310,82 @@ public interface ReceiveChannel<out E> {
...
+    internal class Closed(@JvmField val cause: Throwable?) {
+        val exception: Throwable get() = cause ?: ClosedReceiveChannelException(DEFAULT_CLOSE_MESSAGE)
+        override fun equals(other: Any?): Boolean = other is Closed && cause == other.cause
+        override fun hashCode(): Int = cause.hashCode()
+        override fun toString(): String = "Closed($cause)"
+    }
+
+    internal companion object {
+        @Suppress("NOTHING_TO_INLINE")
+        internal inline fun <E> value(value: E): ReceiveResult<E> =
```
zach-klippenstein: Would you consider making these public once this API is not longer `@InternalCoroutinesApi`? Could be useful to be able to create these for unit tests.

elizarov: Don't see any problem with, but let's track that separately as a future enchancement. We cannot make it all public anyway now, due to instability of inline classes ABI.

kotlinx-coroutines-core/common/src/channels/Channel.kt
elizarov: Pulled in PR #762 into here and implemented it in a less intrusive way.

kotlinx-coroutines-core/common/src/channels/Channel.kt
elizarov: This is not critical to resolve right now (since it is internal). I'm leaving todo there.

kotlinx-coroutines-core/common/src/channels/Channel.kt
elizarov: Leaving todo.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +322,101 @@ public interface ReceiveChannel<out E> {
...
+     * In this case [isValue] returns `false`.
+     */
+    public val isClosed: Boolean get() = holder is Closed
+
+    /**
+     * Returns received value if this instance represents received value or throws [IllegalStateException] otherwise.
+     *
+     * :todo: Decide if it is needed how it shall be named with relation to [valueOrThrow]:
+     *
+     * So we have the following methods on ValueOrClosed: `value`, `valueOrNull`, `valueOrThrow`.
```
qwwdfsad: I am ok with postponing these TODOs for now while this API is unstable, so feel free to resolve my comments regarding it

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +322,101 @@ public interface ReceiveChannel<out E> {
...
+     * On the other hand, the channel has the following receive variants:
+     *  * `receive` which corresponds to `receiveOrClosed().valueOrThrow`... huh?
+     *  * `receiveOrNull` which corresponds to `receiveOrClosed().valueOrNull`
+     *  * `receiveOrClosed`
+     * For the sake of consider dropping this version of `value` and rename [valueOrThrow] to simply `value`.
```
qwwdfsad: I'd rather avoid dropping this version. `valueOrThrow` **rethrows** the `cause` adding no meaningful information about the callsite, so if one is sure that `ValueOrClosed` is always `value`, this very property should be used. Otherwise, it could be very hard to locate the source of the exception

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +322,101 @@ public interface ReceiveChannel<out E> {
...
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val value: T
+        get() = if (isClosed) throw IllegalStateException() else holder as T
+
+    /**
+     * Returns received value if this element represents received value or `null` otherwise.
+     * :todo: Decide if it shall be made into extension that is available only for non-null T.
+     * (note: it might become inconsistent with kotlin.Result)
```
qwwdfsad: I'd rather make it inconsistent with `Result`, but safe.
`Result.getOrNull` was added as member deliberately?

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -271,6 +322,101 @@ public interface ReceiveChannel<out E> {
...
+     * So we have the following methods on ValueOrClosed: `value`, `valueOrNull`, `valueOrThrow`.
+     * On the other hand, the channel has the following receive variants:
+     *  * `receive` which corresponds to `receiveOrClosed().valueOrThrow`... huh?
+     *  * `receiveOrNull` which corresponds to `receiveOrClosed().valueOrNull`
+     *  * `receiveOrClosed`
+     * For the sake of consider dropping this version of `value` and rename [valueOrThrow] to simply `value`.
+     */
+    @Suppress("UNCHECKED_CAST")
+    public val value: T
+        get() = if (isClosed) throw IllegalStateException() else holder as T
```
qwwdfsad: `if (isClosed) error(...)`

kotlinx-coroutines-core/common/src/flow/Channels.kt
```diff
@@ -7,14 +7,81 @@
...
+/**
+ * Represents the given receive channel as a hot flow and [consumes][ReceiveChannel.consume] the channel
+ * on the first collection from this flow. The resulting flow can be collected just once and throws
+ * [IllegalStateException] when trying to collect it more than once.
+ *
+ * ### Cancellation semantics
+ * 1) Flow consumer is cancelled when the original channel is cancelled.
+ * 2) Flow consumer completes normally when the original channel completes (~is closed) normally.
+ * 3) If the flow consumer fails with an exception, channel is cancelled.
+ *
```
qwwdfsad: extra indent

kotlinx-coroutines-core/common/src/flow/Channels.kt
```diff
@@ -7,14 +7,81 @@
...
+ */
+@FlowPreview
+public fun <T> ReceiveChannel<T>.consumeAsFlow(): Flow<T> = object : Flow<T> {
```
qwwdfsad: [optional] I find out that reading flow code is really unpleasant.
E.g. this method adds a frame named `kotlinx.coroutines.flow.FlowKt__ChannelsKt$consumeAsFlow$1.collect(Channels.kt:81)`. Not the best user experience :) 

While we can avoid such frames for inline methods, we can at least do a best-effort on non-inline ones

elizarov: How can it be fixed? Move the code to a named class?

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -91,7 +93,8 @@ public interface SendChannel<in E> {
      * on the side of [ReceiveChannel] starts returning `true` only after all previously sent elements
      * are received.
      *
-     * A channel that was closed without a [cause] throws [ClosedSendChannelException] on attempts to send or receive.
+     * A channel that was closed without a [cause] throws [ClosedSendChannelException] on attempts to send or
+     * and [ClosedReceiveChannelException] on attempts to receive.
```
romansl: Typo: odd `or` before `and`.

kotlinx-coroutines-core/common/src/channels/Channel.kt
elizarov: Added the above comment to todos.

kotlinx-coroutines-core/common/src/flow/Channels.kt
elizarov: Where?

kotlinx-coroutines-core/common/src/channels/Channel.kt
elizarov: Thanks!

elizarov:
I've squashed and rearranged all the old commit into two commits, fixing all the review issues:

1. "Channel.receiveOrNull becomes extension, internal receiveOrClosed added" combines all the improvement to channels machinery and introduces internal `receiveOrClosed`.
2. "Introduce ReceiveChannel.consumeAsFlow and FlowCollector.emitAll(chan)" implements `consumeAsFlow` using those channel improvements.

I've added a 3rd (new) commit that makes `consumeAsFlow` fuseable with all the channel-using flow operators so it avoids creation of an extra channel when possible and unless explicitly requested. Please, take a look.

kotlinx-coroutines-core/common/src/flow/Channels.kt
qwwdfsad: line 74

PR #1344, opened 2019-07-16
Recover stacktraces for no-dispatched continuations, so recovery work…

### LINKED ISSUE #1328: Stacktrace recovery doesn't work for suspend main
I took the example https://raw.githubusercontent.com/Kotlin/kotlinx.coroutines/master/kotlinx-coroutines-debug/test/RecoveryExample.kt and add at line 16 a `println("${Thread.currentThread().name}")`, so I get
```
package example

import kotlinx.coroutines.CoroutineName
import kotlinx.coroutines.async

object PublicApiImplementation : CoroutineScope by CoroutineScope(CoroutineName("Example")) {

    private fun doWork(): Int {
        error("Internal invariant failed")
    }

    private fun asynchronousWork(): Int {
        println("${Thread.currentThread().name}")
        return doWork() + 1
    }

    public suspend fun awaitAsynchronousWorkInMainThread() {
        val task = async(Dispatchers.Default) {
            asynchronousWork()
        }

        task.await()
    }
}

suspend fun main() {
    // Try to switch debug mode on and off to see the difference
    PublicApiImplementation.awaitAsynchronousWorkInMainThread()
}
```
I run it with 

1. -ea
2. -Dkotlinx.coroutines.debug -Dkotlinx.coroutines.stacktrace.recovery=true
3. -Dkotlinx.coroutines.debug

I never get an recovered stacktrace. How ever I got
`DefaultDispatcher-worker-1 @Example#1`
so `-Dkotlinx.coroutines.debug` seems to work.

What does I'm missing?
Here my version
kotlin 1.3.41 
coroutines.version 1.2.2

qwwdfsad:
…s in 'suspend fun main' cases to further improve user experience

Fixes #1328

### DIFF
--- a/kotlinx-coroutines-core/common/src/Dispatched.kt
+++ b/kotlinx-coroutines-core/common/src/Dispatched.kt
@@ -307,7 +307,14 @@ internal fun <T> DispatchedTask<T>.resume(delegate: Continuation<T>, useMode: In
     val state = takeState()
     val exception = getExceptionalResult(state)
     if (exception != null) {
-        delegate.resumeWithExceptionMode(exception, useMode)
+        /*
+         * Recover stacktrace for non-dispatched tasks.
+         * We usually do not recover stacktrace in a `resume` as all resumes go through `DispatchedTask.run`
+         * and we recover stacktraces there, but this is not the case for a `suspend fun main()` that knows nothing about
+         * kotlinx.coroutines and DispatchedTask
+         */
+        val recovered = if (delegate is DispatchedTask<*>) exception else recoverStackTrace(exception, delegate)
+        delegate.resumeWithExceptionMode(recovered, useMode)
     } else {
         delegate.resumeMode(getSuccessfulResult(state), useMode)
     }
--- a/kotlinx-coroutines-core/jvm/test/exceptions/StackTraceRecoveryTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/exceptions/StackTraceRecoveryTest.kt
@@ -6,9 +6,11 @@ package kotlinx.coroutines.exceptions
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.intrinsics.*
 import kotlinx.coroutines.selects.*
 import org.junit.Test
 import java.util.concurrent.*
+import kotlin.coroutines.*
 import kotlin.test.*
 
 /*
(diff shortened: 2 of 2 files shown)


kotlinx-coroutines-core/common/src/Dispatched.kt
```diff
@@ -307,7 +307,14 @@ internal fun <T> DispatchedTask<T>.resume(delegate: Continuation<T>, useMode: In
...
     val exception = getExceptionalResult(state)
     if (exception != null) {
-        delegate.resumeWithExceptionMode(exception, useMode)
+        /*
+         * Recover stacktrace for non-dispatched tasks.
+         * We usually do not recover stacktrace in a `resume` as all resumes go through `DispatchedTask.run`
+         * and we recover stacktraces there, but this is not the case for a `suspend fun main()` that knows nothing about
+         * kotlinx.coroutines and DispatchedTask
+         */
+        val recovered = if (delegate is DispatchedTask<*>) exception else recoverStackTrace(exception, delegate)
```
elizarov: This `DispatchedTask<T>.resume` extension is called only in two places. Once case is when `delegate` is not a `DispatchedContinuation`, but it is also called, for example, when there is a mismatch in resume modes. There other is from `resumeUnconfined` which is similar. Note that `DispatchTask.run` always recovers stack-trace when calling resume, so it looks like this function should unconditionally recover stack-trace, too.

Replacing this line with `val recovered = recoverStackTrace(exception, delegate)` does not break stack-track recovery test.

qwwdfsad: >Once case is when delegate is not a DispatchedContinuation, but it is also called, for example, when there is a mismatch in resume modes.

Yes, exactly. But when there is a mismatch in resume modes, `delegate` is still `DispatchedContinuation`, thus `delegate.resume` will trigger `run` and stacktrace recovery as well.

>Replacing this line with val recovered = recoverStackTrace(exception, delegate) does not break stack-track recovery test.

Sure, because of a smart enough `StackTraceRecoveryKt.mergeRecoveredTraces`. Adding unconditional recovery will only increase the amount of unnecessary work

elizarov: Indeed!

elizarov: Indeed!

Issue #1346, opened 2019-07-16
kotlinx.coroutines.guava ListenableFuture cancellation


yorickhenning:
I've found a few issues in `ListenableFuture.kt` with respect to `Future`'s interface contract and with cancellation propagation:

- Propagation of cancellation through `asListenableFuture()` is unidirectional; cancellation should be bidirectionally cooperative for any chain of `Deferred` and/or `LF<>` through transformation
- Cancellation of a `Future` from `asListenableFuture()` through `future.setFuture(someCancelledFuture)` is broken; this should cancel `future` and doesn't
- `Future`'s happens-after contract with `isCancelled()` is (I _think_) broken in some cases, but this is hard to nail down; a Future can never return true from any more than one call to `Future.cancel()`, but I think the current `asListenableFuture()` `Future` can
- Cancelling `asListenableFuture()` without interruption doesn't propagate to the deferred without interruption, which seems wrong? It's possible to cleanly cancel in this case.
- Cancelling a `Deferred` with a `CancellationException` cause won't propagate the cause to an `asListenableFuture()` created from the `Deferred` - the cause and message are lost
- Most of these subtler semantics aren't documented; there are some strange intersections between of what `Job` cancellation means and what `Future` cancellation means

These corner cases matter when using `.guava` as a migration layer in codebases where other implementations of `Future` or `ListenableFuture`/`AbstractFuture` are present. This is especially true for production code relying on `cancel()` behaviour to perform atomic work, or to efficiently clean up servlets or other somewhat-long-running tasks.

I've nailed down the problematic properties with unit tests. I've got a PR fixing them and documenting `.guava` pending.

There's an [existing discussion](https://github.com/Kotlin/kotlinx.coroutines/issues/515) on some related semantics around `await()` that don't address `asListenableFuture()` but do intersect a bit.

From the Guava perspective, the PR's semantics are "more correct" for `ListenableFuture` semantics, and should be a no-op or an improvement for `Deferred` semantics. There are some optimizations around atomic reads, too.

qwwdfsad:
Fixed in 1.3.1

PR #1347, opened 2019-07-16
Repaired some of ListenableFuture.kt's cancellation corner cases

yorickhenning:
This fixes:
  - Cancellation without an untrapped CancellationException propagating
    through a Callback; isCancelled() is the correct way to check for
    cancellation
  - Bidirectional propagation of cancellation through
    `asListenableFuture()`
  - The cause getting lost in the `asListenableFuture()` future  when
    cancelling its `Deferred` parent with a cause

This also:
  - Extensively documents the package and the contracts created by the
    promise-creating extension methods and `future()`
  - Uses `getUninterruptibly()` for speed
  - Uses `AbstractFuture` to make as certain as possible that
    `Future.cancel()` will return `true` at most once
  - Should clear up rare spooky race conditions around
    cancellation/interruption in hybrid Coroutines/Guava Futures
    codebases

There are probably a few more interesting corner cases hiding in here,
but this should be a good start improving the correctness of `.guava`'s
adapters.


yorickhenning:
@charlesmunger
@cpovirk

yorickhenning:
Should resolve #1346

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -5,31 +5,51 @@
...
+ * See [newCoroutineContext][CoroutineScope.newCoroutineContext] for a description of debugging
+ * facilities.
  *
- * @param context additional to [CoroutineScope.coroutineContext] context of the coroutine.
+ * Note that the error semantics of [future] are _subtly different_ than [asListenableFuture].
+ * By calling [future], any error that occurs after successfully cancelling the returned
+ * [ListenableFuture] will be passed to the [CoroutineExceptionHandler] from the context. The
+ * contract of [Future] does not permit it to return an error after it is successfully cancelled.
+ *
+ * By calling [asListenableFuture] on a [Deferred], any error that occurs after successfully
```
qwwdfsad: This paragraph is not necessary for the public API, it's better to move to implementation

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,81 +58,63 @@ public fun <T> CoroutineScope.future(
...
+    if (isDone && !isCancelled) {
         return try {
-            @Suppress("UNCHECKED_CAST")
-            CompletableDeferred(get() as T)
-        } catch (e: Throwable) {
-            // unwrap original cause from ExecutionException
-            val original = (e as? ExecutionException)?.cause ?: e
-            CompletableDeferred<T>().also { it.completeExceptionally(original) }
+            // getUninterruptibly() avoids one volatile read over Futures.getDone().
+            CompletableDeferred(Uninterruptibles.getUninterruptibly(this))
```
qwwdfsad: Guava has a bad reputation with `@Beta` API compatibility. It's better to use stable API, especially if it only costs one volatile read

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
```diff
@@ -205,6 +336,59 @@ class ListenableFutureTest : TestBase() {
...
+        val asDeferredAsFuture = asDeferred.asListenableFuture()
+
+        assertTrue(asDeferredAsFuture.isCancelled)
+        assertFailsWith<CancellationException> {
+            val value: Int = asDeferredAsFuture.await()
+        }
+    }
+
+    @Test
+    fun testCancelledDeferredAsListenableFutureAsDeferredJoinDoesNotThrow() = runTest {
```
qwwdfsad: `join` throws if and only if a caller of the `join` was cancelled

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
```diff
@@ -311,9 +495,9 @@ class ListenableFutureTest : TestBase() {
 
     @Test
     fun testUnhandledExceptionOnExternalCancellation() = runTest(
-        unhandled = listOf(
-            { it -> it is TestException } // exception is unhandled because there is no parent
-        )
+      unhandled = listOf(
+        { it -> it is TestException } // exception is unhandled because there is no parent
+      )
```
qwwdfsad: this is not the library code-style, please revert this change

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
charlesmunger: This method is not `@Beta`:

https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/Uninterruptibles.java#L232

qwwdfsad: Good to know, apparently we use a slightly outdated version of Guava

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -129,36 +131,289 @@ public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
...
+private class ListenableFutureCoroutine<T>(
+    context: CoroutineContext,
+    private val future: SettableFuture<T>
+) : AbstractCoroutine<T>(context)  {
+
+    override fun onCompleted(value: T) {
+        future.set(value)
+    }
+
+    // TODO: This doesn't actually cancel the Future. There doesn't seem to be bidi cancellation?
```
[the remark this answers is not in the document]
charlesmunger: A Future has four states:

1. Pending (isDone == false, isCancelled == false, get will block)
2. Cancelled (isDone == true, isCancelled == true, get will throw CancellationException)
3. Failed (isDone == true, isCancelled == false, get will throw ExecutionException)
4. Succeeded (isDone == true, isCancelled == false, get will return a value)

Thus there is a distinction between "the future is cancelled, and throws CancellationException from get()" and "The future is failed, and thus will throw an ExecutionException with a cause of CancellationException from get()"

A Future should only throw CancellationException from get() if isCancelled() returns true, and vice versa. 

This distinction doesn't seem to exist with Deferred, and so there's a bit of a mismatch. This behavior is what we came up with to best bridge the gap; if the Deferred itself is cancelled (by the Deferred's definition, which is failing by CancellationException), then the future reports itself is cancelled. 

This java code:
```java
Future<?> future = executorService.submit(() -> {throw new CancellationException();});
future.get();
```

Will throw an ExecutionException with a cause of CancellationException. That behavior matches your latter case.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,81 +58,63 @@ public fun <T> CoroutineScope.future(
...
+ *
+ * When either promise successfully completes, it will attempt to synchronously complete its
+ * counterpart with the same value. This will succeed barring a race with cancellation.
+ *
+ * When either promise completes with an Exception, it will attempt to synchronously complete its
+ * counterpart with the same Exception. This will succeed barring a race with cancellation.
+ *
+ * Cancellation is propagated bidirectionally.
+ *
+ * When `this` [ListenableFuture is successfully cancelled - meaning [ListenableFuture.cancel]
```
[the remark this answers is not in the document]
yorickhenning: Done.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,81 +58,63 @@ public fun <T> CoroutineScope.future(
...
-
 /**
- * Converts this listenable future to an instance of [Deferred].
- * It is cancelled when the resulting deferred is cancelled.
+ * Returns a [Deferred] that is completed or failed by `this` [ListenableFuture].
+ *
+ * Completion is non-atomic between the two promises.
+ *
+ * When either promise successfully completes, it will attempt to synchronously complete its
+ * counterpart with the same value. This will succeed barring a race with cancellation.
```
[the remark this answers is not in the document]
yorickhenning: By "completes" I mean "reaches Cancelled or Completed state" for a `Deferred`, or "finishes" for a `Future`. I'm referring to the promise ending, not its being completed in the sense of calling `CompletableDeferred.complete()`.

Is there another way I can phrase this to make that intention clearer?

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
yorickhenning: Autoformatter strikes, I think. Unintentional diff. Reverted.

Is there a documented style? I'm working from IntelliJ's autoformatter.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
yorickhenning: Moved to `CancellationToCoroutine` class doc and crosslinked.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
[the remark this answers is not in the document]
yorickhenning: Done.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,81 +58,63 @@ public fun <T> CoroutineScope.future(
 ): ListenableFuture<T> {
     require(!start.isLazy) { "$start start is not supported" }
     val newContext = newCoroutineContext(context)
+    // TODO: It'd be nice not to leak this SettableFuture reference, which is easily blind-cast.
     val future = SettableFuture.create<T>()
     val coroutine = ListenableFutureCoroutine(newContext, future)
-    Futures.addCallback(future, coroutine, MoreExecutors.directExecutor())
+    future.addListener(
+      CancellationToCoroutine(future, coroutine),
```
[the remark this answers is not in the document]
yorickhenning: I noticed that optimization but figured it might be a bit obtuse to open with it. :)

Done.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,81 +58,63 @@ public fun <T> CoroutineScope.future(
...
+            // ExecutionException is the only kind of exception that can be thrown from a gotten
+            // Future. Anything else showing up here indicates a very fundamental bug in a
+            // Future implementation.
+            CompletableDeferred<T>().also { it.completeExceptionally(e.nonNullCause()) }
         }
     }
+
     val deferred = CompletableDeferred<T>()
     Futures.addCallback(this, object : FutureCallback<T> {
         override fun onSuccess(result: T?) {
```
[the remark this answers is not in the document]
yorickhenning: True. Good catch.

In my first read, I thought the unchecked cast was equivalent to the !! assert. Re-reading, the previous behaviour was to catch and `completeExceptionally()` with a KNPE. I've restored that behaviour with a version that won't try/catch for the null case. I've added a test and documentation to make sure the behaviour doesn't change unnoticed again.

The fundamental problem is that `ListenableFuture<T>` can complete with a null value, and `CompletableDeferred<T>` (and Deferred<T>) cannot. Offhand, I don't think there's a better solution than this. Using the type `LF<T?>` in the method seems like a high cost.

I started addressing cancellation, not nullability, but here we are with bonus test cases. :)

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
yorickhenning: Am I understanding right that you mean the test case's name is redundant? I think I see tha. I've picked a new name to make it clearer what I'm trying to test for here.

To paraphrase, my intention is to test that the correct `Deferred` cancellation completion behaviour tunnels through a `Deferred` -> `Future` -> `Deferred` transformation. That's a good guarantee that the two transforms interoperate properly.

Let me know if I've misunderstood your perspective here.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
yorickhenning: Move to `ListenableFutureCoroutine` after merging `CancellationToCoroutine`.

yorickhenning:
Any further feedback? I'd like to get this finished up.

qwwdfsad:
Sorry, I'm busy with 1.3 release and internal activities, will return to your PR on Friday

yorickhenning:
> Sorry, I'm busy with 1.3 release and internal activities, will return to your PR on Friday

No worries! Just wanted to make sure it's not forgotten.

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
qwwdfsad: There isn't, but we are also using IntelliJ's autoformatter ¯\_(ツ)_/¯

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -77,88 +292,143 @@ private class ListenableFutureCoroutine<T>(
...
+ *     [ListenableFuture]'s documented and implicit contracts is surprisingly difficult to achieve.
+ *     The best way to be correct, especially given the fun corner cases from
+ *     [AsyncFuture.setAsync], is to just use an [AsyncFuture].
+ *   - To maintain sanity, this class implements [ListenableFuture] and uses an inner [AsyncFuture]
+ *     around its input [deferred] as a state engine to establish happens-after-completion. This
+ *     could probably be compressed into one subclass of [AsyncFuture] to save an allocation, at the
+ *     cost of the implementation's readability.
  */
-public fun <T> Deferred<T>.asListenableFuture(): ListenableFuture<T> = DeferredListenableFuture<T>(this)
+private class OuterFuture<T>(val deferred: Deferred<T>): ListenableFuture<T> {
```
qwwdfsad: `private val deferred` will save two methods in the bytecode :)

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -77,88 +292,143 @@ private class ListenableFutureCoroutine<T>(
...
+ *
+ * - Correct implementation of [Future]'s happens-after semantics documented for [get], [isDone]
+ *   and [isCancelled] methods
+ * - Cancellation propagation both to and from [Deferred]
+ * - Correct cancellation and completion semantics even when this [ListenableFuture] is combined
+ *   with different concrete implementations of [ListenableFuture]
+ *   - Fully correct cancellation and listener happens-after obeying [Future] and
+ *     [ListenableFuture]'s documented and implicit contracts is surprisingly difficult to achieve.
+ *     The best way to be correct, especially given the fun corner cases from
+ *     [AsyncFuture.setAsync], is to just use an [AsyncFuture].
```
qwwdfsad: There is no such class as `AsyncFuture`

qwwdfsad:
Almost there :)
In the same time, I will start to do the same exercise for `CompletableFuture`

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
JakeWharton: Is there a YouTrack for fixing this to happen automatically? Even if it's in the new IR backend only, you shouldn't be required to know things like this.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
yorickhenning: AbstractFuture. Not sure how I got that one wrong.

Fixed.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
yorickhenning: Done.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,28 +45,235 @@ public fun <T> CoroutineScope.future(
...
+ * the task is cancelled via the [Deferred] reference to it. This is unavoidable, so make sure
+ * no idempotent work is performed by any users of the [ListenableFuture] reference to the task
+ * if they're the thread that succeeded in cancelling the task - it may not get done.
+ *
+ * This is inherently a race. See [Future.cancel] for a description of `Future` cancellation
+ * semantics. See [Job] for a description of coroutine cancellation semantics.
+ */
+public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
+    // Fast path if already completed.
+    // TODO: There's a faster fast path - InternalFutureFailureAccess - in Guava >= 27.0
```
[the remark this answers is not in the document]
yorickhenning: Done. Will send another PR with the latest requested changes.

Let's see if I still know how to hold git right...

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
elizarov: @JakeWharton See https://youtrack.jetbrains.com/issue/KT-16454

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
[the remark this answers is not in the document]
yorickhenning: Done. Moved the details inside the method as a block comment.

yorickhenning:
Rebase on develop w/ some improvements: https://github.com/Kotlin/kotlinx.coroutines/pull/1441

PR #1348, opened 2019-07-16
Add Flow.onStart, support emit in onCompletion

### LINKED ISSUE #1168: Flow.concatWith
I often times need to concat two flows. In RxJava this is called `concatWith`.

A implementation could look like:
```kotlin
fun <T> Flow<T>.concatWith(other: Flow<T>): Flow<T> {
  return flow {
    collect { emit(it) }
    other.collect { emit(it) }
  }
}
```

elizarov:
Add Flow.onStart, support emit in onCompletion

* All "emitting" operators (onStart, transform, onCompletion) are moved
  to Emitters.kt and are implemented in a safe way, making sure that
  user-specified block is called on a SafeCollector only.
* Unsafe internal version of transform (unsafeTransform) is introduced.  
* All transformations in Transform.kt are rewritten via unsafeTransform. 
* Added migration for startWith and concatWith.
* Consistent docs for all migration functions.
* JvmMultifileClass for Migration.kt so that renamed/deprecated 
  functions (when they moved there) continue to resolve.

Fixes #1168

### DIFF
--- a/kotlinx-coroutines-core/common/test/flow/operators/OnStartTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/OnStartTest.kt
@@ -2,10 +2,9 @@
  * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-package kotlinx.coroutines.flow.operators
+package kotlinx.coroutines.flow
 
 import kotlinx.coroutines.*
-import kotlinx.coroutines.flow.*
 import kotlin.test.*
 
 class OnStartTest : TestBase() {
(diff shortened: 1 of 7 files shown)


kotlinx-coroutines-core/common/test/flow/operators/OnStartTest.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow.operators
```
qwwdfsad: wrong package

kotlinx-coroutines-core/common/test/flow/operators/TransformTest.kt
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow.operators
```
qwwdfsad: wrong package

PR #1351, opened 2019-07-17
Cancel extensions for CoroutineScope and Job

### LINKED ISSUE #1338: Consider CoroutineScope.cancel(String, Throwable?)
This is a simple extension, I found verbose to write `cancel(CancellationException("socket closed", ioException))`

So I propose to consider something like `cancel("socket closed", ioException)`

```kotlin
public inline fun CoroutineScope.cancel(description: String, cause: Throwable? = null) {
    cancel(CancellationException(description, cause))
}
```

This function is similar to deprecated: `cancel(Trowable?)`

qwwdfsad:
Fixes #1338


qwwdfsad:
Extension on `CoroutineContext` is not added deliberately as it's too heavy-weight and verbose

PR #1352, opened 2019-07-17
Stabilize core flow

qwwdfsad:
Only core interfaces and the most trivial operators are promoted to stable.

`switchMap` and `combienLatest` are demoted (we can do it within milestones, but not across different versions)

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Builders.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Builders.kt
@@ -44,12 +44,12 @@ import kotlin.jvm.*
  * ```
  * If you want to switch the context of execution of a flow, use the [flowOn] operator.
  */
-@ExperimentalCoroutinesApi
-public fun <T> flow(@BuilderInference block: suspend FlowCollector<T>.() -> Unit): Flow<T> {
-    return object : Flow<T> {
-        override suspend fun collect(collector: FlowCollector<T>) {
-            SafeCollector(collector, coroutineContext).block()
-        }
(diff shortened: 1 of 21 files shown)


kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -195,7 +194,6 @@ public interface Flow<out T> {
  * }
  * ```
  */
-@FlowPreview
 public abstract class AbstractFlow<T> : Flow<T> {
```
elizarov: I'm still not satisfied by `collectSafely` abstract method here that is becoming public with this change. It's name name has some non-trivial connotations. I suggest to leave it in preview. The safe `flow { ... }` builder does not internally use this class, so it has a stable ABI. We don't _have_ to make it public on the first release.

kotlinx-coroutines-core/common/src/flow/Builders.kt
```diff
@@ -90,7 +90,6 @@ public fun <T> (suspend () -> T).asFlow(): Flow<T> = unsafeFlow {
 /**
  * Creates a flow that produces values from the given iterable.
  */
-@ExperimentalCoroutinesApi
 public fun <T> Iterable<T>.asFlow(): Flow<T> = unsafeFlow {
```
elizarov: I makes sense to move `unsafeFlow` builder to a separate file, so that all the builders have can be written as `flow { ... }` for the source-browsing pleasure.

kotlinx-coroutines-core/common/src/flow/Flow.kt
qwwdfsad: Good point

Issue #1261, opened 2019-06-07
Flow.share operator


elizarov:
The `share()` operator operates on `Flow<T>` and returns `Flow<T>`. It shall have the following semantics. The resulting flow is cold, but when one collector shart collecting from it, the it starts to collect from the upstream flow, activating the emitter upstream. The trick of the `share` operator is that when additional collectors appear in the downstream, they all "share" the same upstream emitter. 

For example, consider the flow:

```
val flow = flow { 
    var i = 0
    while(true) { 
        delay(1000) 
        println("Emit $i")
        emit(i++)
    }
}
```

If you launch two collectors:

```
launch { flow.collect { println("A: got $it") } }
launch { flow.collect { println("B: got $it") } }
```

Then you shall see "Emit 0 / A: got 0 / Emit 0 / B: got 0 / Emit 1 / A: got 1 / Emit 1 / B: got 1 /  ...".

However, if you change the flow to `val flow = flow { /* same */ }.share()`, then you shall see "Emit 0 / A: got 0 / B: got 0 / Emit 1 / A: got 1 / B: got 1 /  ...", that is one emission gets delivered to both collectors. 

Now if you need to artificially "start" the shared flow simply to keep it active, then you can always launch a dummy collector: `launch { flow.collect {} }` that works as a "reference" which is active until you cancel the resulting job.

**TBD**: Share operator might need some configuration with the respect to how much "history" to keep in memory for "late collectors". So far it seems that one non-negative integer is enough (with zero -- new collector don't get any history, with one -- only the most recent value, with more -- the specified number of recent values). What is unclear is what shall be the default value (if any).

**UPDATE**: It will have to be, actually, a `shareIn(scope)` operator. Otherwise, the scope it works in will be bound the the first collectors and when this callector is cancelled it will not be able to maintain emissions to other collectors.

matejdro:
This looks good as #1086 implementation. One thing missing is the sharing timeout, which use case is described in https://github.com/Kotlin/kotlinx.coroutines/issues/1086#issuecomment-485675107

streetsofboston:
One question about 'starting' a shared/published flow through `launch { flow.collect {} }`.

In #1086 , starting a shared Flow could only happen on a Flow that was already shared, because `connect()` could only be invoked on a `ConnectableFlow`. Any subsequent subscribers/cosumers/collectors to this shared flow would not restart the producer of the flow. 

[...]

PR #1354, opened 2019-07-18
[DRAFT] DataFlow implementation

### LINKED ISSUE #1082: Consider making BroadcastChannels to implement Flow
Unlike point-to-point channels that are somewhat tricky to use, various kinds of `BroadcastChannel` implementations seem well suited to directly implement `Flow` interface. That would make easier and slightly more efficient to use them as "data model" classes in MVVM architectures, being able to directly use full set of flow operators to transform them and wire to UI.

Also, we might consider a shorter name for `ConflatedBroadcastChannel` that would focus on its "dataflow variable" nature. `DataFlow` could be a good starting point for this name bike-shedding.

elizarov:
`DataFlow` is a `Flow` analogue to `ConflatedBroadcastChannel`. Since Flow API is simpler than channels APIs, the implementation of `DataFlow` is simpler. It consumes and allocates less memory, while still providing full deadlock-freedom (even though it is not lock-free internally).

The name reflects the fact that `DataFlow` represent an cell in a "data flow programming" model (think of an electronic spreadsheet). Dependent values can be defined by using various operators on the flows, with `combineLatest` being especially useful to combine multiple data flows using arbitrary functions.

It is also a play of Android's `LiveData`, providing very similar API.

Fixes #1082

### Open issues

1. The name. Shall we stick with `DataFlow` or pick something else? 

2. I've made it fusible with `flowOn` and `conflated` operators (they shall have no effect on the `DataFlow`). As a result, the sequence of `dataFlow.flowOn(ctx).buffer()` becomes different from a sequence of `dataFlow.buffer().flowOn(ctx)`. In the former case cxt that is "applied" to dataflow is ignored. In the later case ctx is applied to the buffering coroutine is created between the data flow and the downstream collector. Is it clear why there is difference?

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/DataFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/DataFlow.kt
@@ -1,0 +1,335 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow
+
+import kotlinx.atomicfu.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.flow.internal.*
+import kotlinx.coroutines.internal.*
+import kotlin.coroutines.*
+import kotlin.jvm.*
+
+/**
+ * A [Flow] that contains a single updatable data [value].
+ *
+ * A data flow can be created with `DataFlow()` constructor function either without an initial value
+ * or with an initial value. Data value can be updated by setting its [value] property.
+ * Updates to the [value] are always [conflated][Flow.conflate]. So a slow collector will skip fast updates,
+ * but always collects the most recently emitted value.
+ *
+ * The name reflects the fact that [DataFlow] represents an cell in a "data flow programming" model
+ * (think of an electronic spreadsheet). Dependent values can be defined by using various operators on the
+ * flows, with [combineLatest] being especially useful to combine multiple data flows using arbitrary functions.
+ *
+ * A data flow can be [closed][close] and all its collectors will complete.
+ *
+ * A read-only interface to data flow is a [Flow]. It is supposed to be collected by the consumers of the data.
+ * The ability to read the current [value] is provided only for convenience of the code that updates the data.
+ * For example, the following class encapsulates an integer data flow and increments its value on each call to `inc`:
+ *
+ * ```
+ * class CounterModel {
+ *     private val _counter = DataFlow(0) // private data flow
+ *     val counter: Flow<Int> get() = _counter // publicly exposed as a flow
+ *
+ *     fun inc() {
+ *         _counter.value++
+ *     }
+ * }
+ * ```
+ *
+ * Having two instances of the above `CounterModel` class one can define the sum of their counters like this:
+ *
+ * ```
+ * val aModel = CounterModel()
+ * val bModel = CounterModel()
+ * val sum = aModel.counter.combineLatest(bModel.counter) { a, b -> a + b }
+ * ```
+ *
+ * Conceptually data flow is similar to
+ * [ConflatedBroadcastChannel][kotlinx.coroutines.channels.ConflatedBroadcastChannel]
+ * but it is simpler, because it does not have to implement all the channel APIs.
+ *
+ * All methods of data flow are **thread-safe** and can be safely invoked from concurrent coroutines without
+ * external synchronization.
+ *
+ * ### Operator fusion
+ *
+ * Application of [flowOn], [conflate], and
+ * [buffer] with [CONFLATED][Channel.CONFLATED] or [RENDEZVOUS][Channel.RENDEZVOUS] capacity
+ * has no effect on the data flow.
+ * 
+ * ### Implementation notes
+ *
+ * Data flow implementation is optimized for memory consumption and allocation-freedom. It uses a lock to ensure
+ * thread-safety, but suspending collector coroutines are resumed outside of this lock to avoid dead-locks when
+ * using unconfined coroutines. Adding new collectors has `O(1)` amortized code, but updating a [value] has `O(N)`
+ * cost, where `N` is the number of collectors.
+ */
+@FlowPreview
+public interface DataFlow<T> : Flow<T> {
+    /**
+     * Returns `true` if this data flow is initialized.
+     *
+     * A data flow is initialized when its [value] was set at least once.
+     * A [closed][isClosed] data flow is considered to be initialized.
+     */
+    public val isInitialized: Boolean
+
+    /**
+     * Returns `true` if this data flow is closed.
+     *
+     * See [close].
+     */
+    public val isClosed: Boolean
+
+    /**
+     * Value of this data flow.
+     *
+     * Setting a value that is [equal][Any.equals] to the previous one does nothing. 
+     *
+     * Getting a value from an [uninitialized][isInitialized] flow produces [IllegalStateException].
+     * Getting a value from or setting a value to a [closed][isClosed] flow produces [IllegalStateException].
+     */
+    public var value: T
+
+    /**
+     * Closes this data fl9w.
+     * This is an idempotent operation -- subsequent invocations of this function have no effect and return `false`.
+     * Conceptually, its sets a value to a special "close token" and all active and future collector from
+     * this [Flow] complete either normally (when [cause] is null) or with the corresponding exception that
+     * was specified as a [cause].
+     */
+    public fun close(cause: Throwable? = null): Boolean
+}
+
+/**
+ * Creates an uninitialized [DataFlow].
+ */
+@Suppress("FunctionName")
+@FlowPreview
+public fun <T> DataFlow(): DataFlow<T> = DataFlowImpl(NONE)
+
+/**
+ * Creates a [DataFlow] with a given initial [value].
+ *
+ * This is a shorthand for `DataFlow().apply { this.value = value }`.
+ */
+@Suppress("FunctionName")
+@FlowPreview
+public fun <T> DataFlow(value: T): DataFlow<T> = DataFlowImpl(value ?: NULL)
+
+@SharedImmutable
+private val NONE = Symbol("NONE")
+
+private const val INITIAL_SIZE = 2 // optimized for just a few collectors
+
+// DataFlow slots are allocated for its collectors
+private class DataFlowSlot {
+    /**
+     * Each slot can have one of the following states:
+     *
+     * * `null` -- it is not used right now. Can [allocate] to new collector.
+     * * `NONE` -- used by a collector, but neither suspended nor has pending value.
+     * * `T` | `Closed` -- pending value or closed token. [NULL] is used for `null` values.
+     * * `CancellableContinuationImpl<T>` -- suspended waiting for value.
+     *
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-core/common/src/flow/DataFlow.kt
```diff
@@ -0,0 +1,335 @@
...
+ * but always collects the most recently emitted value.
+ *
+ * The name reflects the fact that [DataFlow] represents an cell in a "data flow programming" model
+ * (think of an electronic spreadsheet). Dependent values can be defined by using various operators on the
+ * flows, with [combineLatest] being especially useful to combine multiple data flows using arbitrary functions.
+ *
+ * A data flow can be [closed][close] and all its collectors will complete.
+ *
+ * A read-only interface to data flow is a [Flow]. It is supposed to be collected by the consumers of the data.
+ * The ability to read the current [value] is provided only for convenience of the code that updates the data.
```
fvasco: It is not specified how to get the current value using the `flow` _read-only_ interface, is `counter.first()` work?

May multiple `counter.first()` invocations return the same value?

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
```diff
@@ -0,0 +1,335 @@
...
+     *
+     * Setting a value that is [equal][Any.equals] to the previous one does nothing. 
+     *
+     * Getting a value from an [uninitialized][isInitialized] flow produces [IllegalStateException].
+     * Getting a value from or setting a value to a [closed][isClosed] flow produces [IllegalStateException].
+     */
+    public var value: T
+
+    /**
+     * Closes this data fl9w.
```
fvasco: typo: flow

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
fvasco: How happen if I read the `value` on uninitialized or closed `DataFlow`?

fvasco: Is CAS not supported by design? The `CounterModel` example is not _thread-safe_.

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
```diff
@@ -0,0 +1,335 @@
...
+     * Returns `true` if this data flow is closed.
+     *
+     * See [close].
+     */
+    public val isClosed: Boolean
+
+    /**
+     * Value of this data flow.
+     *
+     * Setting a value that is [equal][Any.equals] to the previous one does nothing.
```
qwwdfsad: Where this restriction came from? Are we sure this is desired behavior for most of the users?

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
```diff
@@ -0,0 +1,335 @@
...
+ * Creates a [DataFlow] with a given initial [value].
+ *
+ * This is a shorthand for `DataFlow().apply { this.value = value }`.
+ */
+@Suppress("FunctionName")
+@FlowPreview
+public fun <T> DataFlow(value: T): DataFlow<T> = DataFlowImpl(value ?: NULL)
+
+@SharedImmutable
+private val NONE = Symbol("NONE")
```
qwwdfsad: Please move implementation to a different file (so users won't see it when exploring our public API)

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
```diff
@@ -0,0 +1,335 @@
...
+ * 
+ * ### Implementation notes
+ *
+ * Data flow implementation is optimized for memory consumption and allocation-freedom. It uses a lock to ensure
+ * thread-safety, but suspending collector coroutines are resumed outside of this lock to avoid dead-locks when
+ * using unconfined coroutines. Adding new collectors has `O(1)` amortized code, but updating a [value] has `O(N)`
+ * cost, where `N` is the number of collectors.
+ */
+@FlowPreview
+public interface DataFlow<T> : Flow<T> {
```
qwwdfsad: `DataFlow` sounds more like a `BroacastChannel` with (possible cached or replayable) *data*.
This primitive represents a single state as a flow of consecutive updates, I'd rather name it `StateFlow`

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
qwwdfsad: In other Reactive frameworks, such entity does not only represent `Flow`, but also `FlowCollector`.
Example of usages:
 * [Hystrix, caching example](https://github.com/Netflix/Hystrix/blob/7f5a0afc23aa5ff82320560a04d4c81a45efd67c/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCachedObservable.java#L15)
 * [Kickstarted, binding subject to observable](https://github.com/kickstarter/android-oss/blob/e73c5585948b3d9435359f42a3892290c9850f6f/app/src/main/java/com/kickstarter/viewmodels/DeepLinkViewModel.java)

Note that I (currently) do not advocate for doing the same thing, but it is something definitely worth consideration (maybe @gildor can give us a few pointers here?).

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
```diff
@@ -0,0 +1,335 @@
...
+
+    /**
+     * Value of this data flow.
+     *
+     * Setting a value that is [equal][Any.equals] to the previous one does nothing. 
+     *
+     * Getting a value from an [uninitialized][isInitialized] flow produces [IllegalStateException].
+     * Getting a value from or setting a value to a [closed][isClosed] flow produces [IllegalStateException].
+     */
+    public var value: T
```
qwwdfsad: Regarding my previous comment about `FlowCollector`.

I am afraid that mutable `value` will mix imperative and declarative approaches.
If we designed something like Compose, then yes, updating a `var` is a way to go.
But we are already in the world of `Flow` that requires more ceremonies, but is pretty approachable and clear in return. I am afraid that the ability to mutate `var` instead of writing `state.emit(update)` will make end-users code less clear and understandable.

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
elizarov: Yes. `counter.first()` does compute and return the current value.

elizarov: `IllegalStateException`. See the docs for the `value`.

elizarov: Yes. It is assume that the owner of `LiveData` that makes updates is single-threaded. Do you have compelling use-cases to support CAS there to allow coordinate updates from multiple threads?

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
elizarov: I've thought about making it implement `FlowCollector` (it is easy to add), but I could not find a complessing use-case to do it. In Kotlin word it will be less convenient, since `FlowCollector.emit` is a suspending function, do you do not actually need suspending update function. You can simply update `DataFlow.value` from any context, even non-suspending one. 

P.S. If we create some kind of `EventFlow` as a complement (which would replace `ArrayBroadcastChannel` and would not conflate events), then it would be logical to call its updating function `emit` and naturally implement `FlowCollector` interface there.

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
fvasco: I often use a `Channel` or a `Mutex` to resolve this kind of issues, so these use cases can be resolved outside the `DataFlow` class.

`Flow` is a safe abstraction, but the example `counter.value++` is not. I am not proposing a some kind of CAS operation, however there is not possible to declare that `counter.value++` simply works.

Similar to other @qwwdfsad's considerations, we can consider `DataFlow<T> : Flow<T>, FlowCollector<T>`, this adds a lot of _cerimonies_ but it make evident any issue in `counter.emit( counter.first() + 1)`.

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
fvasco: Hi @elizarov

> FlowCollector.emit is a suspending function, do you do not actually need suspending update function

Overriding a suspending function with a non-suspending one is a current Kotlin limitation.
`DataFlow` shares this issue with `ConflatedChannel` and `UnlimitedChannel`.

On the other hand in current design to update value (`dataFlow.value=42`) is cheap, instead to read a value using the public interface is really expensive (`flow.first()`).

gildor: Personally, I do not have good use cases for FlowCollector for DataFlow (same as rarely use io.reactivex.Observer that implemented by Subject). so my opinion that some simple adapter function that allows forwarding events from Flow/Channel/Lambda to DataFlow would be enough

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
gildor: I don't think that StateFlow is better, for more use cases it's essentially observable data field than some abstract state. At least it how we usually use it

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
gildor: Should `emit` be suspendable or regular function/

qwwdfsad: It's an open question. Either we can play the same trick as with channels (suspendable `send` + non-suspendable `offer` and guarantee, that for non-closed conflated channel `offer` always succeeds) and provide two methods (naming TBD) or make `emit` non-suspendable

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
yigit: +1 for allowing same value to be set twice. If someone wants to dedup it, they can as well use `distinctUntilChanged`.
One use case for dispatching the same value is using it for UI actions, like a refresh.

```
private val refreshTrigger = DataFlow<Unit>().also {
   it.value = Unit
}
val data = refreshTrigger.flatMapLatest {
    repositroy.loadData()
}
fun refresh() { 
    refreshTrigger.value = Unit
}
```
If there is an `EventFlow`, it could also fulfill this need.

kotlinx-coroutines-core/common/src/flow/DataFlow.kt
yigit: +1 for `DataFlow`. State sounds more limiting.

elizarov: @yigit @gildor @qwwdfsad Bikeshedding the name here: https://github.com/Kotlin/kotlinx.coroutines/issues/1082#issuecomment-525650399

JakeWharton:
That's https://github.com/Kotlin/kotlinx.coroutines/issues/1261

elizarov:
This prototype PR is superseded by the actual state flow design that is explained in issue #1973 and PR for `StateFlow` implementation in #1974.

PR #1368, opened 2019-07-23
Fully copy CoroutineInfo for DebugProbes.dumpCoroutinesInfo, it is re…

qwwdfsad:
…quired for IDEA integration (field is left as internal deliberately)

### DIFF
--- a/kotlinx-coroutines-debug/src/CoroutineInfo.kt
+++ b/kotlinx-coroutines-debug/src/CoroutineInfo.kt
@@ -91,9 +91,7 @@ public class CoroutineInfo internal constructor(
         }
     }
 
-    override fun toString(): String {
-        return "CoroutineInfo(state=$state,context=$context)"
-    }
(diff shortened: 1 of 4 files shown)


qwwdfsad:
I've decided to make `CoroutineInfo` non-data instead. Data class doesn't have any value here, but generates toString/hc/eq with a pretty weird (and mostly useless) contract.

kotlinx-coroutines-debug/src/CoroutineInfo.kt
```diff
@@ -94,6 +90,10 @@ public data class CoroutineInfo internal constructor(
             lastObservedThread = null
         }
     }
+
+    override fun toString(): String {
+        return "CoroutineInfo(state=$state,context=$context)"
+    }
```
elizarov: It would read in a much more Kotlin way (less boilerplate to parse to eyes) like this:
```
    override fun toString(): String = "CoroutineInfo(state=$state,context=$context)"
```

PR #1369, opened 2019-07-23
Better docs on coroutines debugging property

elizarov:
* Mention in the guide that `-ea` turns it on.
* Give a link to DEBUG_PROPERTY_NAME instead of newCoroutineContext.
  The later does not have details on its page anymore, since
  the details were only mentioned in JVM version.
* Move description in of debugging facilities to DEBUG_PROPERTY_NAME
  in the code.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/Debug.kt
+++ b/kotlinx-coroutines-core/jvm/src/Debug.kt
@@ -12,7 +12,26 @@ import java.util.concurrent.atomic.*
 import kotlin.internal.InlineOnly
 
 /**
- * Name of the property that controls coroutine debugging. See [newCoroutineContext][CoroutineScope.newCoroutineContext].
+ * Name of the property that controls coroutine debugging.
+ *
+ * ### Debugging facilities
+ *
+ * In debug mode every coroutine is assigned a unique consecutive identifier.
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/jvm/src/Debug.kt
```diff
@@ -12,7 +12,26 @@ import java.util.concurrent.atomic.*
 import kotlin.internal.InlineOnly
 
 /**
- * Name of the property that controls coroutine debugging. See [newCoroutineContext][CoroutineScope.newCoroutineContext].
+ * Name of the property that controls coroutine debugging.
+ *
+ * **Debugging facilities:** In debug mode every coroutine is assigned a unique consecutive identifier.
+ * Every thread that executes a coroutine has its name modified to include the name and identifier of the
+ * currently running coroutine.
+ * When one coroutine is suspended and resumes another coroutine that is dispatched in the same thread,
```
qwwdfsad: >When one coroutine is suspended and resumes another coroutine that is dispatched in the same thread, then the thread name displays the whole stack of coroutine descriptions that are being executed on this thread.

We are not doing it, only the last coroutine description is reflected in thread name

elizarov: Good catch!

PR #1371, opened 2019-07-23
Cover all publications in bom

Prototik:
Currently kx-coroutines-core-{common,js,native} publications missing

### DIFF
--- a/kotlinx-coroutines-bom/build.gradle
+++ b/kotlinx-coroutines-bom/build.gradle
@@ -10,8 +10,14 @@ def name = project.name
 dependencyManagement {
     dependencies {
         rootProject.subprojects.each {
-            if (!ext.unpublished.contains(it.name) && it.name != name) {
-                dependency(group: it.group, name: it.name, version: it.version)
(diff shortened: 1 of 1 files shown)


Prototik:
But BOMs can be applied at any project, not only jvm-only. If I want to build kotlin/js module it still valid to use bom for dependency resolution, isn't it?

So BOMs isn't bound to jvm world, it just used to resolve correct versions of libraries.

[Kotlin itself](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-bom/pom.xml) provides js/common artifacts too

Prototik:
@qwwdfsad any info?

qwwdfsad:
I will merge your PR soon, your concerns regarding non-jvm usages are completely valid.

>Also discovered that io.spring.dependency-management could be replaces with built-in gradle java-platform plugin (with rich gradle metadata info), but that requires gradle 5.2

Thanks for the info! We hopefully update to Gradle 5.2 as soon as figure out what to do with Gradle metadata.

qwwdfsad:
Great job, thanks!

PR #1376, opened 2019-07-24
Use RxJava Spelling for BehaviorSubject.

runningcode:
This represents a class in the RxJava library. Not saying US English is
better, but it is. 🤣 

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/subjects/BehaviorSubject.java


qwwdfsad:
Good catch

PR #1377, opened 2019-07-24
Coroutine context propagation for Reactor to coroutines API migration

SokolovaMaria:
We've supported the following cases of the coroutine context propagation for Reactor to coroutines API migration: 

- `kotlinx.coroutines.reactive` contains extension functions for (`Mono`/`Flux`) to suspending function: ```suspend fun <T> Publisher<T>.await*(): T?```. Coroutine context of these `await` functions is propagated into `Publisher`'s context. Consider the use case: 
```
suspend fun foo() {
    ...
    bar().awaitFirst()
}

fun bar(): Mono<T> { ... } 
```
- The same idea introduced for migration from Reactor's `Publisher` to `Flow`: suspending `collect` call propagates it's coroutine context to the source Publisher context. 
```
// function that uses new coroutines Flow API
suspend fun bar(): Flow<T> {
     return oldBar().asFlow()
}
// legacy function that returns Publisher
fun oldBar(): Flux<T> {...}

suspend fun foo() {
     bar().collect() 
}
```

This propagation is implemented the following way: 

- We provide `ContextInjector` service API that contains the only function to inject the given coroutine context into the `Publisher`: 
```
public interface ContextInjector {
    public fun <T> injectCoroutineContext(publisher: Publisher<T>, coroutineContext: CoroutineContext): Publisher<T>
}
```
- `reactor` module contains the implementation of this interface (`ReactorContextInjector`) that enriches Reactor's `Mono`/`Flux` contexts with values from `coroutineContext[ReactorContext]`, instances of other `Publisher` types are returned unchanged.

- `Publisher<T>.await*()` functions search for the available `ContextInjector` services (if `reactor` module is included - `ReactorContextInjector` is available, otherwise no services found) and enrich `Publisher`'s context if possible.
`PublisherAsFlow` injects the coroutine context of `collect` call the same way.

Available services are searched once, if `reactor` module is not included - the list of services is empty, so no overhead added.


reactive/kotlinx-coroutines-reactor/src/ReactorContextInjector.kt
```diff
@@ -0,0 +1,28 @@
...
+
+/** @suppress */
+@InternalCoroutinesApi
+class ReactorContextInjector : ContextInjector {
+    /**
+     * Injects all values from the [ReactorContext] entry of the given coroutine context
+     * into the downstream [Context] of Reactor's [Publisher] instances of [Mono] or [Flux].
+     */
+    override fun <T> injectCoroutineContext(publisher: Publisher<T>, coroutineContext: CoroutineContext): Publisher<T> {
+        val reactorContext = coroutineContext[ReactorContext]?.context
```
elizarov: Please, use `?: return this` idiom on this line. There is no need to perform `when` below if there is no context and no need to duplicate `?: Context.empty()` code twice.

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -77,10 +80,21 @@ public suspend fun <T> Publisher<T>.awaitLast(): T = awaitOne(Mode.LAST)
...
+public suspend fun <T> Publisher<T>.awaitSingle(): T = injectCoroutineContext().awaitOne(Mode.SINGLE)
 
 // ------------------------ private ------------------------
 
+// ContextInjector service is implemented in `kotlinx-coroutines-reactor` module only.
+// If `kotlinx-coroutines-reactor` module is not included, the list is empty.
+private val contextInjectors: List<ContextInjector> =
+    ServiceLoader.load(ContextInjector::class.java, ContextInjector::class.java.classLoader).toList()
+
+private suspend fun <T> Publisher<T>.injectCoroutineContext(): Publisher<T> {
```
elizarov: This function must have `context: CoroutineContext` as its parameter and it must not have `suspend` modifier. Note that all `awaitXxx` suspending functions remain ammendeable to tail-call optimization and do not have a state-machine. But you'd also move call to `injectCoroutineContext` to inside of `awaitOne` implementation and there you'd get coroutine context from the continuation via `cont.coroutineContext`.

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -77,10 +80,21 @@ public suspend fun <T> Publisher<T>.awaitLast(): T = awaitOne(Mode.LAST)
...
+    return if (contextInjectors.isNotEmpty()) {
+        contextInjectors[0].injectCoroutineContext(this, coroutineContext)
```
elizarov: Please, iterate over all context injectors here using `fold(this) { pub -> pub.injectCoroutineContext(pub, coroutineContext) }`. No need for a separate `isEmpty` check, since `fold` is quite an efficient inline function.

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -22,7 +23,7 @@ import kotlin.coroutines.*
  *
  * @throws NoSuchElementException if publisher does not emit any value
  */
-public suspend fun <T> Publisher<T>.awaitFirst(): T = awaitOne(Mode.FIRST)
+public suspend fun <T> Publisher<T>.awaitFirst(): T = injectCoroutineContext().awaitOne(Mode.FIRST)
```
elizarov: Just use `injectCoroutineContext` inside `awaitOne` implementation. No need to call it from every `awaitXxx` function.

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -52,7 +53,7 @@ public suspend fun <T> Publisher<T>.awaitFirstOrNull(): T? = awaitOne(Mode.FIRST
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * immediately resumes with [CancellationException].
  */
-public suspend fun <T> Publisher<T>.awaitFirstOrElse(defaultValue: () -> T): T = awaitOne(Mode.FIRST_OR_DEFAULT) ?: defaultValue()
+public suspend fun <T> Publisher<T>.awaitFirstOrElse(defaultValue: () -> T): T? = awaitOne(Mode.FIRST_OR_DEFAULT) ?: defaultValue()
```
qwwdfsad: Why was the return type changed?

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -81,6 +82,14 @@ public suspend fun <T> Publisher<T>.awaitSingle(): T = awaitOne(Mode.SINGLE)
 
 // ------------------------ private ------------------------
 
+// ContextInjector service is implemented in `kotlinx-coroutines-reactor` module only.
+// If `kotlinx-coroutines-reactor` module is not included, the list is empty.
+private val contextInjectors: List<ContextInjector> =
```
qwwdfsad: It makes sense to invoke `first` then to have an instance instead of a foldable list.
Note that `first` should be used, not `single`, otherwise some third-party code can make library unusable.

reactive/kotlinx-coroutines-reactive/src/ContextInjector.kt
```diff
@@ -0,0 +1,14 @@
+package kotlinx.coroutines.reactive
+
+import kotlinx.coroutines.InternalCoroutinesApi
+import org.reactivestreams.Publisher
+import kotlin.coroutines.CoroutineContext
+
+/** @suppress */
+@InternalCoroutinesApi
+public interface ContextInjector {
```
qwwdfsad: Could you please make it internal and suppress it in the reactor module?
This API looks dangerous enough not to expose it

reactive/kotlinx-coroutines-reactor/src/FlowAsFlux.kt
```diff
@@ -0,0 +1,21 @@
...
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.reactive.flow.FlowSubscription
+import reactor.core.CoreSubscriber
+import reactor.core.publisher.Flux
+
+/**
+ * Transforms the given flow to [Flux].
+ */
+@ExperimentalCoroutinesApi
+public fun <T: Any> Flow<T>.asFlux(): Flux<T> = FlowAsFlux(this)
```
qwwdfsad: Please provide more detailed documentation (like in `Publisher.asFlow`)

reactive/kotlinx-coroutines-reactor/src/ReactorContextInjector.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.coroutines.reactive.ContextInjector
+import org.reactivestreams.Publisher
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
+import reactor.util.context.Context
+import kotlin.coroutines.CoroutineContext
+
+/** @suppress */
+@InternalCoroutinesApi
+class ReactorContextInjector : ContextInjector {
```
qwwdfsad: `internal`

reactive/kotlinx-coroutines-reactor/src/FlowAsFlux.kt
```diff
@@ -0,0 +1,21 @@
+package kotlinx.coroutines.reactor
```
qwwdfsad: We have `Flow.asPublisher` in `kotlinx.coroutines.reactive.flow` and `asFlux` in `kotlinx.coroutines.reactor`.
We should add or remove `flow` prefix and I'm thinking about its removal.
WDYT @elizarov ?

reactive/kotlinx-coroutines-reactive/src/flow/FlowAsPublisher.kt
```diff
@@ -23,81 +23,83 @@ public fun <T : Any> Flow<T>.asPublisher(): Publisher<T> = FlowAsPublisher(this)
...
+    }
 
-        private suspend fun consumeFlow() {
-            flow.collect { value ->
-                if (!coroutineContext.isActive) {
-                    subscriber.onComplete()
-                    coroutineContext.ensureActive()
-                }
+    private suspend fun consumeFlow() {
+        val ctx = coroutineContext
```
qwwdfsad: leftover?

reactive/kotlinx-coroutines-reactive/src/flow/FlowAsPublisher.kt
```diff
@@ -23,81 +23,83 @@ public fun <T : Any> Flow<T>.asPublisher(): Publisher<T> = FlowAsPublisher(this)
...
-                consumeFlow()
-                subscriber.onComplete()
-            } catch (e: Throwable) {
-                // Failed with real exception, not due to cancellation
-                if (!coroutineContext[Job]!!.isCancelled) {
-                    subscriber.onError(e)
-                }
+/** @suppress */
+@InternalCoroutinesApi
+public class FlowSubscription<T>(val flow: Flow<T>, val subscriber: Subscriber<in T>) : Subscription {
```
qwwdfsad: val's should be `public` + `JvmField` to avoid excess getter/setters

reactive/kotlinx-coroutines-reactive/src/flow/PublisherAsFlow.kt
```diff
@@ -70,7 +71,7 @@ private class PublisherAsFlow<T : Any>(
 
     override suspend fun collect(collector: FlowCollector<T>) {
         val subscriber = ReactiveSubscriber<T>(capacity, requestSize)
-        publisher.subscribe(subscriber)
+        publisher.injectCoroutineContext(coroutineContext).subscribe(subscriber)
```
qwwdfsad: This injection is missing from the `produceImpl`.
Please add a failing test first

reactive/kotlinx-coroutines-reactive/src/Await.kt
elizarov: `first` will not save us from 3rd party code, as there will be no guarantee on the order. This way it is, at least, extensible (theoretically). I've changed the type to array to get a little bit of extra performance (removing additional wrapping layer)

reactive/kotlinx-coroutines-reactive/src/ContextInjector.kt
elizarov: I'd rather lean into the other direction -- try to stop using supporessions as it teaches the bad precedent and just as easy to suppress from the user code.

reactive/kotlinx-coroutines-reactor/src/FlowAsFlux.kt
elizarov: Copied docs from Flow.asPublisher/asFlowable

reactive/kotlinx-coroutines-reactor/src/ReactorContextInjector.kt
elizarov: See above

reactive/kotlinx-coroutines-reactor/src/FlowAsFlux.kt
elizarov: Removed the `flow` prefix.

elizarov:
I've resolved some of the issues, squashed commits, dropped `.flow` package from reactive modules, remove `@JvmName("from")` from conversion methods (as it is not used in other conversion methods of reactive module), but moved them to a more future-proof `FlowKt` JVM class.

Force-pushed

reactive/kotlinx-coroutines-reactor/src/ReactorContextInjector.kt
qwwdfsad: This class should not be public at all.

elizarov: Fixed

reactive/kotlinx-coroutines-reactive/src/flow/PublisherAsFlow.kt
elizarov: Test added, injection fixed.

elizarov:
Fixed remaining issues

reactive/kotlinx-coroutines-reactive/src/FlowAsPublisher.kt
```diff
@@ -0,0 +1,112 @@
...
+                    subscriber.onError(e)
+                }
+            } catch (e: Throwable) {
+                // Last ditch report
+                handleCoroutineException(coroutineContext, e)
+            }
+        }
+    }
+
+    private suspend fun consumeFlow() {
```
elizarov: Please, add comment there that there is only _one_ caller of _consumeFlow_.

reactive/kotlinx-coroutines-reactive/src/FlowAsPublisher.kt
```diff
@@ -0,0 +1,112 @@
...
+        flow.collect { value ->
+            /*
+             * Flow is scopeless, thus if it's not active, its subscription was cancelled.
+             * No intermediate "child failed, but flow coroutine is not" states are allowed.
+             */
+            coroutineContext.ensureActive()
+            if (requested.get() == 0L) {
```
elizarov: Please, use `<= 0L` comparison here. It does not really make any difference, but remind the reader that in normal life `requested` is never supposed to be negative.

reactive/kotlinx-coroutines-reactive/src/FlowAsPublisher.kt
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/** @suppress */
+@InternalCoroutinesApi
+public class FlowSubscription<T>(
+    @JvmField val flow: Flow<T>,
+    @JvmField val subscriber: Subscriber<in T>
+) : Subscription {
+    @Volatile
+    private var canceled: Boolean = false
```
elizarov: It looks like we can completely remove `cancelled` and all the related checked without any harm.

reactive/kotlinx-coroutines-reactive/src/FlowAsPublisher.kt
```diff
@@ -0,0 +1,114 @@
...
+            return
+        }
+        start()
+        var snapshot: Long
+        var newValue: Long
+        do {
+            snapshot = requested.value
+            newValue = snapshot + n
+            if (newValue <= 0L) newValue = Long.MAX_VALUE
+        } while (!requested.compareAndSet(snapshot, newValue))
```
elizarov: Seems shorter (and easier to understand) now is to write:
```
        requested.update { value ->
            val newValue = value + n
            if (newValue <= 0L) Long.MAX_VALUE else newValue
        }
```

reactive/kotlinx-coroutines-reactive/src/FlowAsPublisher.kt
```diff
@@ -0,0 +1,114 @@
...
+
+        val prev = producer.value
+        if (prev == null || !producer.compareAndSet(prev, null)) return
```
elizarov: It might be slightly better to write here:
```
val prev = producer.getAndSet(null) ?: return
```

reactive/kotlinx-coroutines-reactive/src/FlowAsPublisher.kt
qwwdfsad: nice!

PR #1381, opened 2019-07-25
Combine & switch rework

### LINKED ISSUE #1224: [Flow] combineLatest with custom transformer (e.g. not map transformer)
I miss `combineLatest()` which transformer behaves more like [Flow<E>.transform](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/transform.html) than [Flow<E>.map](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/map.html).

My usecase: I ('d like to) combine multiple flows and emit loading & success state between suspensions in trasformer:

```kotlin
val filter = ConflatedBroadcastChannel<Filter>()
val search = ConflatedBroadcastChannel<String>()

filter.asFlow().combineLatest(search.asFlow()) { f, s ->
    emit(LoadingResource())
   
    val data = fetchData(f, s) // suspending function
    emit(SuccessResource(data))
}
```

I may rewrite it to do a transformation to `Pair`, however it seems quite unnecessary and personally I have use cases for 4 flows (E.g. I would have to create a data class each time).

I could try to prepare a PR.

(Also, I would do this myself but the building-blocks of combineLatest are internal. :X)

### LINKED ISSUE #1262: combineLatest(Iterable<Flow>) operator
Please add a `combineLatest` operator which accepts a list of Flows. Here is an example of my use case:
```kotlin
fun <T> combineLatest(flows: Iterable<Flow<T>>): Flow<List<T>> = TODO()

// A list of connected devices.
val devices: List<Device> = getDevices()

// A list of Flows which receive the state of the device every time it changes.
val flows: List<Flow<DeviceState>> = devices.map { device ->
    device.observeState() // Flow<DeviceState>
}

// Combine the latest state of the devices into a single Flow.
val flow: Flow<List<DeviceState>> = combineLatest(flows)


// Somewhere else in the code
flow.collect { states ->
    for (state in states) {
        // TODO: do something with the states.
    }
}
```

### LINKED ISSUE #1335: Decouple switching logic from switchMap operator and consistent naming
We currently have a `switchMap { ... }` operator that is similar to the corresponding [swithMap operator from Rx](http://reactivex.io/RxJava/javadoc/io/reactivex/Flowable.html#switchMap-io.reactivex.functions.Function-) but from the standpoint of Kotlin coroutines looks like an amalgam of different operations that might need to be decoupled. The most basic operation seems to be this one, tentatively named as `switchTransform`:

```
fun <T, R> Flow<T>.switchTransform(transform: suspend FlowCollector<R>.(value: T) -> Unit): Flow<R>
```

It would be similar to [`Flow.transform`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/transform.html) but with a difference that `transform` waits for transformation to complete for each element while `switchTransfrom` would cancel the ongoing transformation on new incoming element and start it again.

Now, note that `transform` operator can be used to implement the host of basic operators:

```
filter(predicate) = transform { if (predicate(it)) emit(it) }
map(block) = transform { emit(block(it)) }
flattenConcat() = transform { emitAll(it) }
flatMapConcat(block) = map(block).flattenConcat() = transform  { emitAll(block(it)) } 
onEach(block) = transform { block(it); emit(it) }
collect(block) = transform { block(it) }.collect()
```

So, if we have a basic `switchTransform` operator, then it would be fitting to potentially have (at least reserve the names) the whole family of `switchXxx` operators that are similar to the above above when you replace `transform` with `switchTransform` in their implementations.

Moreover, there is a use case (see #1269) for the `collectLatest` terminal operator that should be called `switchCollect` under this proposed nomenclature, because:

```
swithCollect(block) = switchTransform { block(it) }.collect()
```

However, this `switchXxx` nomenclature has the following problem. 

Under the proposed `switchXxx` nomenclature `switchMap` operator shall have the same signature as `map` (operates on `Flow<T>`) and should be equivalent to `switchTransform { emit(block(it)) }`. But we already have experimental `switchMap` operator that operates on `Flow<Flow<T>>` and does `switchTransform { emitAll(block(it)) }` which under the new nomenclature should have been called `switchFlattenMapConcat`. It is not clear how to proceed.

qwwdfsad:
* Operator renamed to combine
  * Introduced combineTransform operator with custom transformer
  * Decouple API and implementation details to improve user experience from IDE
  * combine(Iterable<Flow>) and combineTransform(Iterable<Flow>) are introduced

Fixes #1224
Fixes #1262
Fixes #1335


hrach:
Will this get into 1.3? (🙏)

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -0,0 +1,142 @@
...
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.internal.Symbol
+import kotlinx.coroutines.selects.*
+
+internal fun getNull(): Symbol = NULL // Workaround for JS BE bug
+
+internal suspend inline fun <T1, T2, R> FlowCollector<R>.combineTransformInternal(
+    first: Flow<T1>, second: Flow<T2>,
+    crossinline transform: suspend FlowCollector<R>.(a: T1, b: T2) -> Unit
+) {
```
elizarov: To reduce bytecode size I'd suggest to make it non-inline and in both call-sited directly bass the corresponding lambda parameter to this `combineTransformInternal` function which would avoid lambda creation at the call-site.

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -0,0 +1,142 @@
...
+        } catch (e: AbortFlowException) {
+            // complete
+        } finally {
+            if (!second.isClosedForReceive) second.cancel(AbortFlowException())
+        }
+    }
+}
+
+// Channel has any type due to onReceiveOrNull. This will be fixed after receiveOrClosed
+private fun CoroutineScope.asChannel(flow: Flow<*>): ReceiveChannel<Any> = produce {
```
elizarov: It can be made fuseable with preceding buffer/flowOn calls by implementing it like this:

```
asChannelFlow().buffer(0).produceImpl(this)
```

As a bit of additional optimization (saving one small allocation) `asChannelFlow().buffer(0)` can be merged into `asZeroBufferChannelFlow()` function that can be implemented like this:

```
internal fun <T> Flow<T>.asZeroBufferChannelFlow(): ChannelFlow<T> =
    when (this) {
        is ChannelFlow -> this.buffer(0)
        else -> ChannelFlowOperatorImpl(this, capacity = 0)
   }
```

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -0,0 +1,142 @@
...
+    if (isClosed) return
+    channel.onReceiveOrNull {
+        // TODO onReceiveOrClosed when boxing issues are fixed
+        if (it === null) onClosed()
+        else onReceive(it)
+    }
+}
+
+// Channel has any type due to onReceiveOrNull. This will be fixed after receiveOrClosed
+private fun CoroutineScope.asFairChannel(flow: Flow<*>): ReceiveChannel<Any> = produce {
```
elizarov: It can be made fuseable with preceding buffer/flowOn calls just like `asChannel` by implementing it like this:

```
asChannelFlow().buffer(0).produceImpl(this, fair=true)
```

And adding an option `fair` parameter and threading it thorough the `ChannelFlow` method does to `collectTo` function where `SendingCollector` instance is created, so that it creates a separate `FairSendingCollector` instead.
```

File: kotlinx-coroutines-core/common/src/flow/operators/Zip.kt [hunk 4]
```diff
@@ -137,65 +238,52 @@ public inline fun <T1, T2, T3, T4, T5, R> Flow<T1>.combineLatest(
...
  * the most recently emitted values by each flow.
  */
-@FlowPreview
-public inline fun <reified T, R> Flow<T>.combineLatest(vararg others: Flow<T>, crossinline transform: suspend (Array<T>) -> R): Flow<R> =
-    combineLatest(*others, arrayFactory = { arrayOfNulls(others.size + 1) }, transform = { transform(it) })
+@ExperimentalCoroutinesApi
+public inline fun <reified T, R> combine(
+    vararg flows: Flow<T>,
+    crossinline transform: suspend (Array<T>) -> R
+): Flow<R> = combine(*flows, arrayFactory = { arrayOfNulls(flows.size) }, transform = { emit(transform(it)) })
```
elizarov: This `combine` implementation calls array-based `combine` which is actually a combine-transform that is using _safe_ collector that is not needed here.

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+}
+
+@PublishedApi
+internal fun <T, R> combine(
+    vararg flows: Flow<T>,
+    arrayFactory: () -> Array<T?>,
+    transform: suspend FlowCollector<R>.(Array<T>) -> Unit
+): Flow<R> = flow {
```
elizarov: TooManyFunctionsCalledCombineException. I'd suggest to rename this one to `combineImpl`.

kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
```diff
@@ -209,46 +297,22 @@ private fun CoroutineScope.asFairChannel(flow: Flow<*>): ReceiveChannel<Any> = p
...
-                emit(transform(NULL.unbox(value), NULL.unbox(secondValue)))
-            }
-        } catch (e: AbortFlowException) {
-            // complete
-        } finally {
-            if (!second.isClosedForReceive) second.cancel(AbortFlowException())
-        }
-    }
-}
+public fun <T1, T2, R> Flow<T1>.zip(other: Flow<T2>, transform: suspend (T1, T2) -> R): Flow<R> = zipImpl(this, other, transform)
```
elizarov: I'd suggest to get rid of top-level `zip`. Its name is too short to have it a top level function and, unlike, `combine`, there is no "consistency" argument for having it at the top level.

kotlinx-coroutines-core/common/src/flow/Migration.kt
```diff
@@ -380,7 +380,57 @@ public fun <T> Flow<T>.concatWith(value: T): Flow<T> = noImpl()
...
+    other4: Flow<T5>,
+    crossinline transform: suspend (T1, T2, T3, T4, T5) -> R
+): Flow<R> = combine(this, other, other2, other3, other4, transform)
+
+@Deprecated(
+    level = DeprecationLevel.ERROR,
+    message = "Flow analogues of 'switchMap' are 'transformLatest', 'flatMapLatest' and 'mapLatest'",
+    replaceWith = ReplaceWith("this.flatMapLatest(transform)")
+)
+public fun <T, R> Flow<T>.switchMap(transform: suspend (value: T) -> Flow<R>): Flow<R> = noImpl()
```
elizarov: Since we had `switchMap` in the previous release (even as "flow preivew") it would be helpful to preserve binary compatibility here by delegating to `flatMapLatest` here.

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
qwwdfsad: `combineInternal` to be consistent with the rest of the API

kotlinx-coroutines-core/common/src/flow/operators/Zip.kt
qwwdfsad: Good point

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
elizarov: After some further thought, the `capactiy` parameter can be made a default parameter to `Flow.asChannelFlow` extension (with default of `OPTIONAL_CHANNEL`).

kotlinx-coroutines-core/common/src/flow/terminal/Collect.kt
```diff
@@ -86,9 +86,45 @@ public suspend inline fun <T> Flow<T>.collectIndexed(crossinline action: suspend
         override suspend fun emit(value: T) = action(checkIndexOverflow(index++), value)
     })
 
+/**
+ * Terminal flow operator that collects the given flow with a provided [action].
+ * The crucial difference from [collect] is that when the original flow emits a new value, [action] block for previous
+ * value is cancelled.
+ * It can be demonstrated by the following example:
```
ilya-g: Makes sense to end the summary here.

qwwdfsad: What do you mean by "summary"?

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
qwwdfsad: It's a bit more complicated than that.
For example, `produceIn`/`produceImpl` is not enough because we handle nulls in a special way and `onReceiveOrClosed` creates a non-eliminable box, it will be working only "partially" etc.

kotlinx-coroutines-core/common/src/flow/terminal/Collect.kt
ilya-g: kdoc has summary and details parts separated by an empty line. The example should go in details.

PR #1383, opened 2019-07-25
Add more Flow migrations for Rx types

artem-zinnatullin:
Migrations already included:

- BehaviourSubject
- ReplaySubject
- PublishSubject

This PR adds more types to ease migration to Flow with potential to migrations for static and non-static functions (that's why they're classes with private constructor).


qwwdfsad:
We've decided that top-level functions are way too intrusive and make the experience of simultaneous use of Rx and coroutines unpleasant, so we are going to remove all top-level migrations in 1.3.

PR #1384, opened 2019-07-25
Fix naming for Flow BehaviorSubject migration

artem-zinnatullin:
It is `BehaviorSubject` in [RxJava](https://github.com/ReactiveX/RxJava/blob/v2.2.10/src/main/java/io/reactivex/subjects/BehaviorSubject.java) and [ReactiveX](http://reactivex.io/documentation/subject.html).


qwwdfsad:
Thanks, but it is already fixed in `develop`

PR #1387, opened 2019-07-26
Get rid of top-level functions in Migration.kt to improve experience …

qwwdfsad:
…of users who depend on any reactive library and kotlinx.coroutines

### DIFF
--- a/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
+++ b/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
@@ -826,9 +826,6 @@ public abstract interface class kotlinx/coroutines/flow/FlowCollector {
 
 public final class kotlinx/coroutines/flow/FlowKt {
 	public static final field DEFAULT_CONCURRENCY_PROPERTY_NAME Ljava/lang/String;
-	public static final fun BehaviorSubject ()Ljava/lang/Object;
-	public static final fun PublishSubject ()Ljava/lang/Object;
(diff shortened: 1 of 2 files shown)


digitalbuddha:
Would it be possible to include these optionally? I feel there is value in offering assistance for those who want to migrate from other reactive streams implementations

qwwdfsad:
I would like to give users a choice here, but I honestly don't know how, at least in the current state of the ecosystem.
On the one hand, these migrations are really valuable and can save a lot of time without google "what is XXX replacement in Flow". On the other hand, it seems to really piss off people who use reactive streams and coroutines simultaneously and tooling (IDEA) is not smart enough to help us here.

digitalbuddha:
Can they be in an optional artifact? I see this similar to packaging lint rules seperately

qwwdfsad:
They sure can, but then it makes much less sense, they are hard to discover and cannot be added to the project automatically

ZacSweers:
For what it's worth - I would have preferred they be a separate artifact to begin with. Document it well as a migratory tool and I think teams will adopt it as needed. At the end of the day, even if it's niche, I think that's sort of the goal

LouisCAD:
Adding a dependency for migration sounds reasonable to me. If it's properly advertised, target people will use it. I think it's the best that can be relatively quickly done in the current state of the tooling.

qwwdfsad:
Status update: current migrations are left as is.
We may provide a migration artifact in the future as long as it will contain more than just subjects without a replacement.

PR #1388, opened 2019-07-26
Allocate underlying buffer in ArrayChannel in on-demand manner

qwwdfsad:
Rationale:
Such change will allow us to use huge buffers in various flow operators without having a serious footprint in suspension-free scenarios

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.internal.*
 import kotlinx.coroutines.selects.*
 import kotlin.jvm.*
+import kotlin.math.*
 
 /**
  * Channel with array buffer of a fixed [capacity].
@@ -29,10 +30,14 @@ internal open class ArrayChannel<E>(
     }
 
     private val lock = ReentrantLock()
-    private val buffer: Array<Any?> = arrayOfNulls<Any?>(capacity)
+    /*
+     * Guarded by lock.
+     * Allocate minimum of capacity and 16 to avoid excess memory pressure for large channels when it's not necessary.
+     */
+    private var buffer: Array<Any?> = arrayOfNulls<Any?>(min(capacity, 16))
     private var head: Int = 0
     @Volatile
-    private var size: Int = 0
+    private var size: Int = 0 // Invariant: size <= capacity
 
     protected final override val isBufferAlwaysEmpty: Boolean get() = false
     protected final override val isBufferEmpty: Boolean get() = size == 0
@@ -64,6 +69,7 @@ internal open class ArrayChannel<E>(
                         }
                     }
                 }
+                ensureCapacity(size)
                 buffer[(head + size) % capacity] = element // actually queue element
                 return OFFER_SUCCESS
             }
@@ -112,6 +118,7 @@ internal open class ArrayChannel<E>(
                     this.size = size // restore size
                     return ALREADY_SELECTED
                 }
+                ensureCapacity(size)
                 buffer[(head + size) % capacity] = element // actually queue element
                 return OFFER_SUCCESS
             }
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -123,6 +130,16 @@ internal open class ArrayChannel<E>(
         return receive!!.offerResult
     }
 
+    // Guarded by lock
+    private fun ensureCapacity(currentSize: Int) {
+        if (currentSize == buffer.size) {
+            val newSize = min(buffer.size * 2, capacity)
+            val newBuffer = arrayOfNulls<Any?>(newSize)
+            buffer.copyInto(newBuffer)
+            buffer = newBuffer
```
elizarov: This does not work properly when buffer is current wrapped around (`head != 0`) as the `newBuffer` will not contain a contiguous segment of values. The easiest way to fix it is to copy all the values from `buffer` into `newBuffer` using a loop into zero offset and reset head to zero like this:

```
            for (i in 0 until currentSize)
                newBuffer[i] = buffer[(head + i) % buffer.size]
            head = 0
```

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -123,6 +130,16 @@ internal open class ArrayChannel<E>(
         return receive!!.offerResult
     }
 
+    // Guarded by lock
+    private fun ensureCapacity(currentSize: Int) {
+        if (currentSize == buffer.size) {
```
elizarov: Nitpick: It is a good style use `>=` in comparisons like this (aka "defensive programming"). Not that it really matters, but it makes code slightly easier to understand.

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -64,6 +69,7 @@ internal open class ArrayChannel<E>(
                         }
                     }
                 }
+                ensureCapacity(size)
                 buffer[(head + size) % capacity] = element // actually queue element
```
elizarov: Here it should use `buffer.size` as a modulo instead of `capacity` when storing to the buffer. Also, in all other places in the code `foo % capacity` shall be replaced with `foo % buffer.size` (all the code that reads/write to/from `buffer` and the code that updates `head`).

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -29,10 +30,14 @@ internal open class ArrayChannel<E>(
     }
 
     private val lock = ReentrantLock()
-    private val buffer: Array<Any?> = arrayOfNulls<Any?>(capacity)
+    /*
+     * Guarded by lock.
+     * Allocate minimum of capacity and 16 to avoid excess memory pressure for large channels when it's not necessary.
+     */
+    private var buffer: Array<Any?> = arrayOfNulls<Any?>(min(capacity, 16))
```
elizarov: Live hack: Default allocation size can be made even smaller (I suggest the value of `8` -- it will still produce "quite large" arrays). This way, all the existing stress tests that test for array channels of capacity `10` will also be testing resizing. Unfortunately, they do not actually fail with this buggy code, so a dedicated resizing test is still needed (need to send/receive a bunch a values to wrap around the buffer, then increase the number of buffered values to resize wrapped buffer)

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
fvasco: It is possible to split the loop and avoid the modulo operation.

```kotlin
val firstSliceSize = minOf(head + currentSize, buffer.size) - head
for (i in 0 until firstSliceSize) {
    newBuffer[i] = buffer[head + i]
}
for (i in 0 until head + currentSize - buffer.size) {
    newBuffer[firstSliceSize + i] = buffer[i]
}
```

elizarov: @fvasco At this point, where you have performed the mental gymnastics of getting rid of module and splitting, you can make it even better by replacing each loop with one call of `copyInto`.

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
qwwdfsad: Shame, fixed

PR #1396, opened 2019-07-30
Check for cancellation in concurrent flow merge on each element

### LINKED ISSUE #1392: Canceling flatMapMerge not canceling the source?
In the following code, the source Flow is not aborted as I'd expect:

```kotlin
@Test
fun flatMapCancel() = runBlocking {
    flow {
        try {
            println("-> 1")
            emit(1)
            println("-> 2")
            emit(2)
            println("-> 3")
            emit(3)
            println("-> 4")
            emit(4)
            println("-> 5")
        } catch (ex: Throwable) {
            ex.printStackTrace()
        }
    }
    .flatMapMerge(2) { v -> flow { emit(v) } }
    .take(2)
    .collect { println (it) }
}
```

prints:

```
-> 1
-> 2
-> 3
1
2
-> 4
-> 5
```

To check, setting concurrency to 1, thus using `flatMapConcat` underneath does abort.

Is this an expected property of `flatMapMerge`?

qwwdfsad:
* Implementation detail (launch on each value) is leaking into upstream behaviour
  * The overhead is negligible compared to launching a new coroutines and sending to channel, but it provides a much approachable mental model when no suspension in the upstream flow happens (note: upstream never sends elements to the channel)

Fixes #1392

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
@@ -151,8 +151,14 @@ private class ChannelFlowMerge<T>(
     // The actual merge implementation with concurrency limit
     private suspend fun mergeImpl(scope: CoroutineScope, collector: ConcurrentFlowCollector<T>) {
         val semaphore = Semaphore(concurrency)
-        @Suppress("UNCHECKED_CAST")
+        val job: Job? = coroutineContext[Job]
         flow.collect { inner ->
+            /*
+             * We launch a coroutine on each emitted element and the only potential
+             * suspension point in this collector is `semaphore.acquire` that rarely suspends,
+             * so we manually check for cancellation to propagate it to the upstream in time.
+             */
+            job?.ensureActive()
             semaphore.acquire() // Acquire concurrency permit
             scope.launch {
                 try {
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/test/flow/operators/FlowOnTest.kt
```diff
@@ -261,6 +261,21 @@ class FlowOnTest : TestBase() {
...
+    @Test
+    fun testCancellation() = runTest {
+        val result = flow {
+            emit(1)
+            emit(2)
+            emit(3)
+            expectUnreached()
+            emit(4)
+        }.flowOn(wrapperDispatcher())
+            .buffer(0)
```
elizarov: This `buffer(0)` shall be removed. Otherwise, this test does not fail without the fix. Alternatively, a test that check for cancellation with default buffer size shall be used here.

kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
```diff
@@ -151,8 +151,9 @@ private class ChannelFlowMerge<T>(
     // The actual merge implementation with concurrency limit
     private suspend fun mergeImpl(scope: CoroutineScope, collector: ConcurrentFlowCollector<T>) {
         val semaphore = Semaphore(concurrency)
-        @Suppress("UNCHECKED_CAST")
+        val job: Job? = coroutineContext[Job]
         flow.collect { inner ->
+            job?.ensureActive()
```
elizarov: A comment explaining the problem here would help a lot in the future.

PR #1401, opened 2019-08-01
Fix documentation of (suspend () -> T).asFlow()

joffrey-bion:
The function calling asFlow() doesn't need to be suspending, and the
doc seems to suggest otherwise.


qwwdfsad:
Thanks!

PR #1403, opened 2019-08-01
Cross platform support for scheduled coroutines

ascheja:
I've tried to mimic the behavior of `ScheduledExecutor` on the jvm, but with suspending functions instead of `Runnable` (and without `TimeUnit`).

### DIFF
--- a/kotlinx-coroutines-core/common/src/Schedule.kt
+++ b/kotlinx-coroutines-core/common/src/Schedule.kt
@@ -1,0 +1,47 @@
+package kotlinx.coroutines
+
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+suspend fun <T> scheduleAsync(
+    initialDelay: Long,
+    context: CoroutineContext = EmptyCoroutineContext,
+    block: suspend () -> T
+): Deferred<T> = coroutineScope {
+    async(Dispatchers.Default + Job()) {
+        delay(initialDelay)
+        withContext(context) { block() }
+    }
+}
+
+suspend fun scheduleWithFixedDelay(
+    initialDelay: Long,
+    successiveDelay: Long,
+    context: CoroutineContext = EmptyCoroutineContext,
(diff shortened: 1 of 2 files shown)


fvasco:
First question: we should ask ourself if this functions are -generally- useful.

Second: is it useful to provide these functions here, or we can achieve a better result using a third part library? (for example: adding cron expression support, also)

Third: are these function working as you expect?

ascheja:
My (personal) impression on these questions:
1. As of now there is no platform independent way of scheduling in kotlin yet. The standard library provides only a wrapper for `Timer` on the jvm, but when using suspending functions you have to wrap your code in `runBlocking` again, which seems rather inconvenient to me.

2. At least `scheduleAtFixedRate` makes use of two internal functions (`delayNanosToMillis` and `nanoTime`), which would require them to be implemented again in a third party library (for each platform). Adding something like cron expression (or `java.time.Duration`) support in a third party library would then be easier, as that library could concentrate on parsing these expressions and then just has to call `scheduleAtFixedRate`.

3. From what I've tried they seem to work fine, but on the other hand I'm not a coroutines power user yet, so If something is looking hinky for you I'm happy to learn.

fvasco:
> 1. As of now there is no platform independent way of scheduling in kotlin yet.

I agree with your statement, but it does not answer to my question, nor I don't have a personal answer.

> 2. At least scheduleAtFixedRate makes use of two internal functions (delayNanosToMillis and nanoTime), which would require them to be implemented again in a third party library (for each platform).

`nanoTime` is a larger issue, shoul be addressed in a KEEP, like https://github.com/Kotlin/KEEP/issues/190
After it was addressed, the question "2" remains open.

> 3. From what I've tried they seem to work fine

The main criticize that I see is the completely lack of structured concurrency, a [supervisorScope](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/supervisor-scope.html) should be introduced.
A plain `Timer` replacement is not enough for me.

PR #1407, opened 2019-08-04
Remove no longer needed tests excludes

### LINKED ISSUE #1405: Remove no longer needed excludes
Issue mentioned below was fixed in 1.3.40 (duplicate of [KT-31354](https://youtrack.jetbrains.com/issue/KT-31354))
https://github.com/Kotlin/kotlinx.coroutines/blob/ed97260407f59f1a91aea0b7bf5b4276105b4ac2/build.gradle#L150

…and 1.3.41 is greater than 1.3.31
https://github.com/Kotlin/kotlinx.coroutines/blob/ed97260407f59f1a91aea0b7bf5b4276105b4ac2/build.gradle#L155

so these two "hacks"(?) should be safe to remove.

LouisCAD:
These exclusions are obsolete since Kotlin 1.3.40.
Resolves #1405

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,12 +147,10 @@ if (build_snapshot_train) {
     allprojects {
         tasks.withType(Test).all {
             exclude '**/*LinearizabilityTest*'
-            exclude '**/*PublicApiTest*' // KT-30956
             exclude '**/*LFTest*'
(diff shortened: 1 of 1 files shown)


qwwdfsad:
Thanks!

PR #1408, opened 2019-08-05
Flow.kt: fix typos and rephrase some expressions for better readability

Inego:


kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -162,18 +162,18 @@ public interface Flow<out T> {
...
      * of the developer mistakes related to concurrency, inconsistent flow dispatchers and cancellation.
      */
     @InternalCoroutinesApi
     public suspend fun collect(collector: FlowCollector<T>)
 }
 
 /**
- * Base class to extend to have a stateful implementation of the flow.
- * It tracks all the properties required for context preservation and throws [IllegalStateException]
+ * The base class for stateful implementations of `Flow`.
```
elizarov: We don't usually use definite article in the first senses of KDocs. See also: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/index.html
Also, I could not find an example in JavaDoc that would shart with "The ...". Somehow it does not feel right. The closes example I was able to find:
https://docs.oracle.com/javase/8/docs/api/java/util/stream/BaseStream.html
Note, that we don't really follow JavaDocs naming conventions, as we find then too verbose (which is fitting for Java, though). For example, may docs start with "This class is ...." which is a total boilerplate which just makes it harder to get to the substance of what the corresponding class is doing.

Inego: I see, thanks. Removed the definite article.

elizarov:
Thanks!

PR #1417, opened 2019-08-07
Fix SOE with select onSend/onReceive clauses on the same channel

### LINKED ISSUE #1411: StackOverflowError when trying to select onSend and onReceive on the same channel
Steps to reproduce:
```
@Test
fun foo() = runBlocking {
    val c = Channel<Int>()
    select<Unit> {
        c.onSend(42) {
            println("Sent")
        }
        c.onReceive {
            println("R $it")
        }
    }
}
```

elizarov:
* Instead of StackOverflowError we throw IllegalStateException and leave
  the channel in the original state.

Fixes #1411


elizarov:
It does not current passes stress-tests on windows.

PR #1428, opened 2019-08-09
Use 'Class.forName($name).canonicalName' instead of '$name' in stackt…

### LINKED ISSUE #1416: Stacktrace recovery works incorrectly when Android minifier is enabled
To properly recover stacktraces we use the [heuristic](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt#L25) based on frame names.

When minifier is enabled, heuristic starts to fail and produces improperly merged stacktraces with internal symbols.

We can at least provide a proper `.pro` configuration and investigate alternative ways to recover stacktraces on Android

qwwdfsad:
…race recovery to properly deal with Android's minifier

Fixes #1416

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
@@ -11,6 +11,21 @@ import java.util.*
 import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
 
+/*
+ * `Class.forName(name).canonicalName` instead of plain `name` is required to properly handle
+ * Android's minifier that renames these classes and breaks our recovery heuristic without such lookup.
+ */
+private const val baseContinuationImplClass = "kotlin.coroutines.jvm.internal.BaseContinuationImpl"
+private const val stackTraceRecoveryClass = "kotlinx.coroutines.internal.StackTraceRecoveryKt"
+
+private val baseContinuationImplClassName = runCatching {
+    Class.forName(baseContinuationImplClass).canonicalName
+}.getOrElse { baseContinuationImplClass }
+
+private val stackTraceRecoveryClassName = runCatching {
+    Class.forName(stackTraceRecoveryClass).canonicalName
+}.getOrElse { stackTraceRecoveryClass }
+
 internal actual fun <E : Throwable> recoverStackTrace(exception: E): E {
     if (recoveryDisabled(exception)) return exception
     // No unwrapping on continuation-less path: exception is not reported multiple times via slow paths
@@ -21,10 +36,9 @@ internal actual fun <E : Throwable> recoverStackTrace(exception: E): E {
 private fun <E : Throwable> E.sanitizeStackTrace(): E {
     val stackTrace = stackTrace
     val size = stackTrace.size
-
-    val lastIntrinsic = stackTrace.frameIndex("kotlinx.coroutines.internal.StackTraceRecoveryKt")
+    val lastIntrinsic = stackTrace.frameIndex(stackTraceRecoveryClassName)
     val startIndex = lastIntrinsic + 1
-    val endIndex = stackTrace.frameIndex("kotlin.coroutines.jvm.internal.BaseContinuationImpl")
+    val endIndex = stackTrace.frameIndex(baseContinuationImplClassName)
     val adjustment = if (endIndex == -1) 0 else size - endIndex
     val trace = Array(size - lastIntrinsic - adjustment) {
         if (it == 0) {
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
```diff
@@ -11,6 +11,21 @@ import java.util.*
 import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
 
+/*
+ * `Class.forName(name).canonicalName` instead of plain `name` is required to properly handle
+ * Android's minifier that renames these classes and breaks our recovery heuristic without such lookup.
+ */
+private const val baseContinuationImplClass = "kotlin.coroutines.jvm.internal.BaseContinuationImpl"
```
qwwdfsad: It would be very nice if @JakeWharton or @wojtek-kalicinski could verify this approach.

Additionally, I've found out that with enabled minifier (build tools version 28.0.3) all line numbers from [coroutines metadata](https://github.com/JetBrains/kotlin/blob/deb416484c5128a6f4bc76c39a3d9878b38cec8c/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/DebugMetadata.kt#L41) are removed, but I am not sure whether it's a bug or intended behaviour

JakeWharton: Currently R8 will rewrite constant strings which flow into `Class.forName` if they match a type on the classpath (and that type is being renamed).

This is pretty easy to demo:
```
$ cat Reflection.java
class Reflection {
  public static void main(String... args) throws Exception {
    String name = Class.forName("BaseContinuationImpl").getCanonicalName();
    System.out.println(name);
  }
}

class BaseContinuationImpl {}

$ javac *.java

$ cat rules.txt
-keepclasseswithmembers class * { 
  public static void main(...); 
} 

$ java -jar ~/dev/android/r8/build/libs/r8.jar \
      --lib $ANDROID_HOME/platforms/android-28/android.jar \
      --release \
      --output . \
      --pg-conf rules.txt \
      *.class

$ $ANDROID_HOME/build-tools/29.0.1/dexdump -d classes.dex
(snip)
Class #0            -
  Class descriptor  : 'LReflection;'
[0001b4] Reflection.main:([Ljava/lang/String;)V
0000: const-string v1, "a"
0002: invoke-static {v1}, Ljava/lang/Class;.forName:(Ljava/lang/String;)Ljava/lang/Class;
0005: move-result-object v1
0006: invoke-virtual {v1}, Ljava/lang/Class;.getCanonicalName:()Ljava/lang/String;
0009: move-result-object v1
000a: sget-object v0, Ljava/lang/System;.out:Ljava/io/PrintStream;
000c: invoke-virtual {v0, v1}, Ljava/io/PrintStream;.println:(Ljava/lang/String;)V
000f: return-void

(snip)

Class #1            -
  Class descriptor  : 'La;'
```

Note how "a" flows into `Class.forName` and the second class was renamed to "a".

Unfortunately there's a huge missed opportunity here to rewrite the constant string flowing into `Class.forName` into a class literal. That's tracked by https://issuetracker.google.com/issues/117545367 (among other reflection-y optimizations). Once that is done, we can enable https://issuetracker.google.com/issues/118536394 which will rewrite calls to `getCanonicalName` which come from a class constant. Which means the _entire_ initializer expressions for the two properties below will be rewritten to a single constant string and then subsequently inlined into their call sites.

I will use the presence of this code in coroutines to hopefully push forward support for the `Class.forName` optimization.

JakeWharton: And for completeness, comment out the second class and you'll see the original string kept.
```diff
 // Reflection.java

-class BaseContinuationImpl {}
+//class BaseContinuationImpl {}
```
```
[000154] Reflection.main:([Ljava/lang/String;)V
0000: const-string v1, "BaseContinuationImpl"
0002: invoke-static {v1}, Ljava/lang/Class;.forName:(Ljava/lang/String;)Ljava/lang/Class;
0005: move-result-object v1
0006: invoke-virtual {v1}, Ljava/lang/Class;.getCanonicalName:()Ljava/lang/String;
0009: move-result-object v1
000a: sget-object v0, Ljava/lang/System;.out:Ljava/io/PrintStream;
000c: invoke-virtual {v0, v1}, Ljava/io/PrintStream;.println:(Ljava/lang/String;)V
000f: return-void
```

jsjeon: Glad to see that R8 identifies identifier name correctly and replaces it with a renamed class name, which is WAI. :)

Just finished `Class<?>#forName` rewriting, and will revisit whether we can enable class name computation. (Thanks for the reminder, btw.)

PR #1432, opened 2019-08-09
Flow docs

elizarov:


docs/flow.md
```diff
@@ -0,0 +1,1684 @@
...
+
+The emitter can use [catch] operator that preserves this exception transparency and allows encapsulation
+of its exception handling. The body of the `catch` operator can analyze an exception
+and react to it in different ways depending on which exception was caught:
+
+* Exceptions can be rethrown using `throw`.
+* Exceptions can be turned into emission of values using [emit][FlowCollector.emit] from the body of [catch].
+* Exceptions can be ignored, logged, or processed by some other code.
+
+For example, let us emit a text of cathing an exception:
```
zach-klippenstein: Typo: "cathing"

docs/flow.md
```diff
@@ -0,0 +1,1684 @@
...
+
+### Launching flow
+
+It is convenient to use flows to represent asynchronous events that are coming from some source.
+In this case we need an analogue of `addEventListener` function that registers a piece of code with reaction
+on incoming events and continues further work. The [onEach] operator can serve this role. 
+However, `onEach` is an intermediate operator. We also need a terminal operator to collect the flow. 
+Otherwise, just calling `onEach` has no effect.
+ 
+If we use [collect] terminal operator after `oneEach` then code after it waits until the flow is collected:
```
zach-klippenstein: Typo: "oneEach"

docs/flow.md
```diff
@@ -0,0 +1,1684 @@
...
+Caught java.lang.IllegalStateException: Collected 2
+-->
+
+### Launching flow
+
+It is convenient to use flows to represent asynchronous events that are coming from some source.
+In this case we need an analogue of `addEventListener` function that registers a piece of code with reaction
+on incoming events and continues further work. The [onEach] operator can serve this role. 
+However, `onEach` is an intermediate operator. We also need a terminal operator to collect the flow. 
+Otherwise, just calling `onEach` has no effect.
```
zach-klippenstein: I don't see what the discussion of events contributes to the explanation of `onEach`/`collect`/`launchIn` – it just feels distracting to me. After the first sentence, I expected this section to be about `callbackFlow`. What's the concrete use case for events that these operators make more convenient than alternative approaches? If there are specific use cases this section is trying to address, it would make more sense to me to give those use cases their own section and document these operators independently.

docs/flow.md
```diff
@@ -0,0 +1,1684 @@
...
+<!--- TEST -->
+
+The required parameter to `launchIn` must specify a [CoroutineScope] in which the coroutine to collect the flow is 
+launched. In the above example this scope comes from [runBlocking]
+coroutine builder, so while the flow is running this [runBlocking] scope waits for completion of its child coroutine
+and keeps the main function from returning and terminating this example. 
+
+In real applications a scope is going to come from some entity with a limited 
+lifetime. As soon as the lifetime of this entity is terminated the corresponding scope is cancelled, cancelling
+collection of the corresponding flow. This way the pair of `onEach { ... }.collectIn(scope)` works
```
zach-klippenstein: Typo: "collectIn" should be "launchIn".

docs/flow.md
```diff
@@ -0,0 +1,1684 @@
...
+Since `foo().collect` is called from the main thread, the body of `foo`'s flow is also called in the main thread.
+This is a perfect default for fast-running or asynchronous code that does not care about the execution context and
+does not block the caller. 
+
+#### Wrong emission withContext
+
+However, long-running CPU-consuming code might need to be executed in the context of [Dispatchers.Default] and UI-updating
+code might need to be executed in the context of [Dispatchers.Main]. Usually, [withContext] is used
+to change the context in code using Kotlin coroutines, but code in the `flow { ... }` builder has to honor context
+preservation property and is not allowed to [emit][FlowCollector.emit] from a different context.
```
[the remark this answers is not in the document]
elizarov: We still do not allow emit from inside of `launch { ... }`. You are only allowed to emit from inside the stuff like `coroutineScope { ... }` which technically creates a new `Job`, but it is a too low-level detail to mention.

docs/flow.md
elizarov: That's the best rationale I was able to come with to explain `launchIn`. Ideally, in the future we'll need a totally separate section of "flow in UI programming" and just move `launchIn` to there.

elizarov:
Reviewed and edited additional section about `onCompletion`. Added one more example, fixed review notes from @zach-klippenstein

docs/flow.md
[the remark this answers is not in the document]
zach-klippenstein: Ah, I was thinking of `ChannelFlowTransformLatest` and `ChannelFlowMerge` but those are internal and bypass the external-facing safety checks.

docs/flow.md
zach-klippenstein: Because UI programming is the main use case for needing to collect a flow from non-suspending code?

elizarov: Yes. I don't a use-case for `launchIn` in server-side business-logic (like in Spring app). `launchIn` was specifically designed for UI code.

zach-klippenstein: TIL

PR #1441, opened 2019-08-14
Repaired some of ListenableFuture.kt's cancellation corner cases.

yorickhenning:
This fixes:
  - Cancellation without an untrapped CancellationException propagating
    through a Callback; isCancelled() is the correct way to check for
    cancellation
  - Bidirectional propagation of cancellation through
    `asListenableFuture()`
  - The cause getting lost in the `asListenableFuture()` future  when
    cancelling its `Deferred` parent with a cause

This also:
  - Extensively documents the package and the contracts created by the
    promise-creating extension methods and `future()`
  - Uses `getUninterruptibly()` for speed
  - Uses `AbstractFuture` to make as certain as possible that
    `Future.cancel()` will return `true` at most once
  - Should clear up rare spooky race conditions around
    cancellation/interruption in hybrid Coroutines/Guava Futures
    codebases

There are probably a few more interesting corner cases hiding in here,
but this should be a good start improving the correctness of `.guava`'s
adapters.

This only addresses the `asFoo()` adapters, not the `future()` direct path. That also needs some cancellation changes. TBD.

This is a squash commit of kotlin/pr/1347 with some more improvements, rebased on develop:

- Incorporated first-round feedback.

- Merged CancellationToCoroutine into ListenableFutureCoroutine to save an
allocation.

- Documented and tested for null completion of asDeferred()'s parent
Future.

- Renamed a cancellation test case for clarity of purpose.

- Split asDeferred() documentation between KDoc/details

- Implemented InternalFutures faster-fast path. Documented.

### DIFF
--- a/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
+++ b/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
@@ -5,31 +5,40 @@
 package kotlinx.coroutines.guava
 
 import com.google.common.util.concurrent.*
-import kotlinx.coroutines.*
+import com.google.common.util.concurrent.internal.InternalFutureFailureAccess
+import com.google.common.util.concurrent.internal.InternalFutures
 import java.util.concurrent.*
-import java.util.concurrent.CancellationException
 import kotlin.coroutines.*
+import kotlinx.coroutines.*
+import java.util.concurrent.CancellationException
 
 /**
- * Starts new coroutine and returns its results an an implementation of [ListenableFuture].
- * The running coroutine is cancelled when the resulting future is cancelled or otherwise completed.
+ * Starts [block] in a new coroutine and returns a [ListenableFuture] pointing to its result.
  *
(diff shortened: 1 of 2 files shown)


yorickhenning:
@qwwdfsad

yorickhenning:
Filed a [followup issue](https://github.com/Kotlin/kotlinx.coroutines/issues/1442) so `future()` doesn't slip through unfixed.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -38,81 +47,98 @@ public fun <T> CoroutineScope.future(
 ): ListenableFuture<T> {
     require(!start.isLazy) { "$start start is not supported" }
     val newContext = newCoroutineContext(context)
+    // TODO: It'd be nice not to leak this SettableFuture reference, which is easily blind-cast.
```
lowasser: A nice short way to do this, if you like: `return object: ListenableFuture<T> by future {}`.

PR #1444, opened 2019-08-15
Fix coroutines guide table

kazemihabib:
The Supervision is a child of Exception handling but it wasn't in
the previous version because of the omitted indententation.


elizarov:
This file is auto-generated and it does not do much use to update it. Running `./gradlew knit` will revert it to a previous state. Please, see this PR where this problem is already fixed #1432

PR #1445, opened 2019-08-17
Optimize select expression registration phase

elizarov:
There is no need for multi-word atomic performAtomicIfNotSelected
operation when enqueuing select node for operation. We can simply
enqueue (addLast) xxxSelect node (SendSelect, ReceiveSelect, LockSelect).
If the coroutine that rendezvous with this node finds out that the
select expression was already selected, then it'll try again.

* Removed SelectInstance.performAtomicIfNotSelected function
* Removed Mutex.TryEnqueueLockDesc class, simplified onLock
* Removed AbstractSendChannel.TryEnqueueSendDesc class, simpler onSend
* Removed AbstractChannel.TryEnqueueReceiveDesc class, simpler onReceive

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
@@ -112,9 +112,10 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
         queue: LockFreeLinkedListHead,
         element: E
     ) : AddLastDesc<SendBuffered<E>>(queue, SendBuffered(element)) {
-        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? {
-            if (affected is ReceiveOrClosed<*>) return OFFER_FAILED
-            return null
+        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? = when (affected) {
+            is Closed<*> -> affected
+            is ReceiveOrClosed<*> -> OFFER_FAILED
+            else -> null
         }
     }
 
@@ -332,10 +333,10 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
     ) : RemoveFirstDesc<ReceiveOrClosed<E>>(queue) {
         @JvmField var resumeToken: Any? = null
 
-        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? {
-            if (affected !is ReceiveOrClosed<*>) return OFFER_FAILED
-            if (affected is Closed<*>) return affected
-            return null
+        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? = when (affected) {
+            is Closed<*> -> affected
+            !is ReceiveOrClosed<*> -> OFFER_FAILED
+            else -> null
         }
 
         override fun validatePrepared(node: ReceiveOrClosed<E>): Boolean {
@@ -633,10 +634,10 @@ internal abstract class AbstractChannel<E> : AbstractSendChannel<E>(), Channel<E
         @JvmField var resumeToken: Any? = null
         @JvmField var pollResult: E? = null
 
-        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? {
-            if (affected is Closed<*>) return affected
-            if (affected !is Send) return POLL_FAILED
-            return null
+        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? = when (affected) {
+            is Closed<*> -> affected
+            !is Send -> POLL_FAILED
+            else -> null
         }
 
         @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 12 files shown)


elizarov:
Also fixed a bug with unlimited channel select onSend on closed channel. It was hanging not being able to properly see that the channel was already closed at that send attempt should fail.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -381,27 +356,30 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
...
+                val node = SendSelect(element, this, select, block)
+                when (val enqueueResult = enqueueSend(node)) {
+                    null -> { // enqueued successfully
+                        select.disposeOnSelect(node)
                         return
                     }
-                    offerResult is Closed<*> -> throw recoverStackTrace(offerResult.sendException)
-                    else -> error("offerSelectInternal returned $offerResult")
+                    is Closed<*> -> {
+                        helpClose(enqueueResult)
```
qwwdfsad: Similar `helpClose` is missing from `registerSelectSend`, making close non-linearizable with `onSend`

qwwdfsad: `sendSuspend` and `registerSelectSend` look very similar, though it's not clear whether we can get rid of copy-paste here.
But this bug becomes obvious if these two methods are grouped together.

Ditto for receive

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -206,12 +205,12 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
...
      */
-    private fun enqueueSend(send: SendElement): Any? {
+    private fun enqueueSend(send: Send): Any? {
         if (isBufferAlwaysFull) {
-            queue.addLastIfPrev(send, { prev ->
+            queue.addLastIfPrev(send) { prev ->
                 if (prev is ReceiveOrClosed<*>) return@enqueueSend prev
                 true
-            })
+            }
```
qwwdfsad: Line 220: `return ENQUEUE_FAILED`

Nobody checks for `ENQUEUE_FAILED`. It doesn't look like a bug because of the current state machine (`null` -> success, `Closed` -> closed, everything else -> retry), but kind of error-prone.

For example, `null` can become `ENQUEUE_SUCCESS` and `ENQUEUE_FAILED` will become `null` instead. Code flow will be more straightforward and explicit

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -633,10 +634,10 @@ internal abstract class AbstractChannel<E> : AbstractSendChannel<E>(), Channel<E
         @JvmField var resumeToken: Any? = null
         @JvmField var pollResult: E? = null
 
-        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? {
-            if (affected is Closed<*>) return affected
-            if (affected !is Send) return POLL_FAILED
-            return null
+        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? = when (affected) {
```
qwwdfsad: It is possible to get rid of the copy-paste here as well, for example:
```
override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? = enqueueFailure(affected) { POLL_FAILED.takeIf { it !is Send } }
```

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -112,9 +112,10 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
         queue: LockFreeLinkedListHead,
         element: E
     ) : AddLastDesc<SendBuffered<E>>(queue, SendBuffered(element)) {
-        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? {
-            if (affected is ReceiveOrClosed<*>) return OFFER_FAILED
-            return null
+        override fun failure(affected: LockFreeLinkedListNode, next: Any): Any? = when (affected) {
```
qwwdfsad: `next` parameter is no longer used by production code.
It is only used in `describeRemove` for tests

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
elizarov: Fixed `helpClose` in `registerSelectSend`. Postone sharing code to later optimization.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
elizarov: I've rewritten the corresponding `when` statements in a more explicit way, checking all possible outcomes and added `error` on unexpected result.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
elizarov: Current code is easier to understand. Postone extraction of common code for future optimization.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
elizarov: Removed `next`, removed `LockFreeLinkedList.describeRemove` (as not really needed in production) and the corresponding tests.

PR #1451, opened 2019-08-20
Implemented parallel bfs benchmark

fmixing:


benchmarks/src/main/kotlin/bfschannel/BfsChannelBenchmark.kt
```diff
@@ -0,0 +1,276 @@
...
+                onFinish.arrive()
+            }
+        }
+    }
+    onFinish.arriveAndAwaitAdvance()
+    // Return the result
+    return distances.map { long -> long.toLong() }
+}
+
+private suspend fun processEdge(distances: Array<AtomicLong>, currentNode: Node, nextNode: Node, queue: Channel<Node>) {
```
ndkoval: why `suspend`? use `queue.offer(..)` instead

benchmarks/src/main/kotlin/bfschannel/BfsChannelBenchmark.kt
```diff
@@ -0,0 +1,276 @@
...
+                distances[edge.to.id] = newDistance
+                queue.add(edge.to)
+            }
+        }
+    }
+    return distances.toList()
+}
+
+fun bfsParallel(start: Node, nodes: Int, parallelism: Int = 0): List<Long> {
+    val workers = if (parallelism != 0) parallelism else Runtime.getRuntime().availableProcessors()
```
ndkoval: Please, specify `parallelism` implicitly

benchmarks/src/main/kotlin/bfschannel/BfsChannelBenchmark.kt
```diff
@@ -0,0 +1,276 @@
...
+                    for (edge in currentNode.outgoingEdges) {
+                        processEdge(distances, currentNode, edge.to, queue)
+                    }
+                }
+            } finally {
+                onFinish.arrive()
+            }
+        }
+    }
+    onFinish.arriveAndAwaitAdvance()
```
ndkoval: It is better to `join` on all the created by `GlobalScope.launch` tasks.

PR #1452, opened 2019-08-20
Restore context preservation invariant in flatMapMerge

qwwdfsad:


kotlinx-coroutines-core/common/src/flow/internal/Merge.kt
```diff
@@ -68,12 +72,15 @@ internal class ChannelFlowMerge<T>(
         }
     }
 
-    // Fast path in ChannelFlowOperator calls this function (channel was not created yet)
     override suspend fun flowCollect(collector: FlowCollector<T>) {
-        // this function should not have been invoked when channel was explicitly requested
-        assert { capacity == Channel.OPTIONAL_CHANNEL }
-        flowScope {
-            mergeImpl(this, collector.asConcurrentFlowCollector())
+        assert { collector !is SendingCollector<*> }
```
elizarov: This `flowCollect` implementation is never called, because there is now "fast path" in this impl (channel is always created) and also because `collectTo` is overridden (default impl would have delegate to `flowCollect`, but not anymore).

qwwdfsad: Good catch! Legacy of the ChannelFlowOperator

PR #1453, opened 2019-08-20
Replace kotlinx.coroutines.reactive.FlowKt and kotlinx.coroutines.rea…

qwwdfsad:
…ctor.FlowKt with properly named files

Otherwise, it leads to a weird classname clash and forces to write FQ imports


reactive/kotlinx-coroutines-reactive/src/Migration.kt
```diff
@@ -0,0 +1,36 @@
...
+package kotlinx.coroutines.reactive
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+import org.reactivestreams.*
+
+// Binary compatibility with Spring 5.2 RC
+@Deprecated(
+    message = "Replaced in favor of ReactiveFlow extension, please import kotlinx.coroutines.reactive.* instead of kotlinx.coroutines.reactive.FlowKt",
+    level = DeprecationLevel.ERROR
```
elizarov: It should have `DeprecationLevel.HIDDEN` since it is for binary compatibility one. Same for all other "Migration" function with the exception of `asFlow(batchSize)` which is also needed for source migration.

PR #1457, opened 2019-08-21
Introduce CoroutineDispatcher.asExecutor() extension

### LINKED ISSUE #1450: Convert `CoroutineContext` into `Executor`
I have to use an API that accepts an `Executor` as a parameter. How can I create an `Executor` that is backed by `Dispatchers.Default`? 

Can a `CoroutineDispatcher.asExecutor(): Executor` function be added to support this use case?

```kotlin
fun externalApi(executor: Executor)

fun example() {
  val executor: Executor = Dispatchers.Default.asExecutor()
  
  externalApi(executor)
}
```

Even better would be a `fun CoroutineContext.asExecutor(): Executor` function. I would imagine if the receiver `CoroutineContext` does not have a `ContinuationInterceptor`, this function would throw.

elizarov:
Fixes #1450

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -47,7 +47,7 @@ public fun Executor.asCoroutineDispatcher(): CoroutineDispatcher =
 /**
  * Converts an instance of [CoroutineDispatcher] to an implementation of [Executor].
  *
- * It returns an original executor when used on the result of [Executor.asCoroutineDispatcher] extensions.
+ * It returns the original executor when used on the result of [Executor.asCoroutineDispatcher] extensions.
  */
 public fun CoroutineDispatcher.asExecutor(): Executor =
     (this as? ExecutorCoroutineDispatcher)?.executor ?: DispatcherExecutor(this)
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.coroutines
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
```diff
@@ -32,14 +33,32 @@ class ExecutorsTest : TestBase() {
...
+    }
+
+    @Test
+    fun testDefaultDispatcherToExecutor() {
+        val latch = CountDownLatch(2)
+        Dispatchers.Default.asExecutor().execute {
+            checkThreadName("DefaultDispatcher")
+            latch.countDown()
+        }
+        latch.countDown()
```
JakeWharton: await?

kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -45,6 +45,19 @@ public fun ExecutorService.asCoroutineDispatcher(): ExecutorCoroutineDispatcher
 public fun Executor.asCoroutineDispatcher(): CoroutineDispatcher =
     ExecutorCoroutineDispatcherImpl(this)
 
+/**
+ * Converts an instance of [CoroutineDispatcher] to an implementation of [Executor].
+ *
+ * It returns an original executor when used on the result of [Executor.asCoroutineDispatcher] extensions.
```
paulblessing:
```suggestion
 * It returns the original executor when used on the result of [Executor.asCoroutineDispatcher] extensions.
```

kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
elizarov: Good catch.

kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
```diff
@@ -32,14 +34,49 @@ class ExecutorsTest : TestBase() {
...
         }
         executor.shutdown()
     }
 
+    @Test
+    fun testConvertedDispatcherToExecutor() {
+        val executor: ExecutorService = Executors.newSingleThreadExecutor { r -> Thread(r, "TestExecutor") }
+        val dispatcher: CoroutineDispatcher = executor.asCoroutineDispatcher()
+        assertSame(executor, dispatcher.asExecutor())
+        executor.shutdown()
```
qwwdfsad: Note: `shutdown` doesn't await actual shutdown

elizarov: That's Ok. Infrastrusture in `TestBase` awaits.

JakeWharton: It looks like it only awaits DefaultExecutor (and other dispatchers). You would need to register this out-of-band `Executor` with it somehow or put a factory method in `TestBase`.

elizarov: https://github.com/Kotlin/kotlinx.coroutines/blob/cd5e60e9f2470dced9687cd908b13c9c9aef887d/kotlinx-coroutines-core/jvm/test/Threads.kt#L47-L53

JakeWharton: Clever!

PR #1462, opened 2019-08-22
Version 1.3.0

qwwdfsad:


CHANGES.md
```diff
@@ -1,5 +1,25 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.3.0
+
+### Flow
+
+This version is the first stable release with [`Flow`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/index.html) API.
+
+All `Flow` API not marked with `@FlowPreview` or `@ExperimentalCoroutinesApi` annotations is stable and here to stay.
+Flow declarations marked with `@ExperimentalCoroutinesApi` has [the same guarantees](/docs/compatibility.md#experimental-api) as regular experimental API.
```
elizarov: "declarations ... has" -> "declarations ... have". I'd also replace "API ... is" in the previous sentence with "APIs ... are stable"

CHANGES.md
```diff
@@ -1,5 +1,25 @@
...
+Please note that API marked with `@FlowPreview` has [weak guarantees](/docs/compatibility.md#flow-preview-api) on source, binary and semantic compatibility.
```
elizarov: "API has" reads funny for me. I'd replace it is with "APIs have"

CHANGES.md
```diff
@@ -1,5 +1,25 @@
...
+
+### Changelog
+
+* A new [guide section](/docs/flow.md) about Flow.
+* `CoroutineDispatcher.asExecutor` extension (#1450).
+* Fixed bug when `select` statement could report the same exception twice (#1433).
+* Fixed context preservation in `flatMapMerge` in a case when collected values were immediately emitted to another flow (#1440).
+* Reactive Flow integrations enclosing files are renamed for better interoperability with Java.
+* Default buffer size in all Flow operator is increased to 64.
```
elizarov: _operators_

CHANGES.md
qwwdfsad: I was using stdlib style. E.g. "API is" in https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/time/ExperimentalTime.kt#L13

elizarov: It is a different context. When you are looking at a specific API has see it annotated as @SomethingExpreimental, then you go do checkout the docs for @SomethingExperimental and see "this API is ...". It makes sense. However, in release statement you are making a more general statement about all annotated APIs.

PR #1470, opened 2019-08-25
Update shared-mutable-state-and-concurrency.md

AlexBlokh:
Main func should return Unit, otherwise it won't run


AlexBlokh:
Screen Shot 2019-08-25 at 13 35 57

elizarov:
Thanks. I've also rerun `./gradlew knit` to update the corresponding auto-generate examples and merged it into `master` (since it is only docs update that does not touch lib sources).

PR #1477, opened 2019-08-27
Fix of a bug in the semaphore

dkhalanskyjb:
This bug surfaced about a month ago during the formal proof. The bug is demonstrated in the provided test that used to fail, but no longer does with this patch.

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
@@ -59,47 +59,37 @@ class SemaphoreStressTest : TestBase() {
         assertEquals(1, semaphore.availablePermits)
     }
 
-    /* This checks if repeated releases that race with cancellations put
-       the semaphore into an incorrect state where permits are leaked. */
+    /**
+     * This checks if repeated releases that race with cancellations put
+     * the semaphore into an incorrect state where permits are leaked.
+     */
     @Test
-    fun stressReleaseCancellation() = runTest {
+    fun stressReleaseCancelRace() = runTest {
         val n = 10_000 * stressTestMultiplier
         val semaphore = Semaphore(1, 1)
-
         newSingleThreadContext("SemaphoreStressTest").use { pool ->
-
             repeat (n) {
                 /* Initially, we hold the permit and no one else can `acquire`,
                    otherwise it's a bug. */
                 assertEquals(0, semaphore.availablePermits)
-
                 var job1_entered_critical_section = false
-
-                val job1 = launch {
+                val job1 = launch(start = CoroutineStart.UNDISPATCHED) {
                     semaphore.acquire()
                     job1_entered_critical_section = true
                     semaphore.release()
                 }
-
-                // Let `job1` run and sleep in `acquire`
-                yield()
-
+                // check that `job1` didn't finish the call to `acquire()`
                 assertEquals(false, job1_entered_critical_section)
-
                 val job2 = launch(pool) {
                     semaphore.release()
                 }
-
                 /* Because `job2` executes in a separate thread, this
                    cancellation races with the call to `release()`. */
                 job1.cancelAndJoin()
-
                 job2.join()
-
                 assertEquals(1, semaphore.availablePermits)
                 semaphore.acquire()
             }
-
         }
     }
 
--- a/kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
@@ -69,8 +69,8 @@ class SemaphoreStressTest : TestBase() {
         val semaphore = Semaphore(1, 1)
         newSingleThreadContext("SemaphoreStressTest").use { pool ->
             repeat (n) {
-                /* Initially, we hold the permit and no one else can `acquire`,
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
```diff
@@ -2,10 +2,18 @@ package kotlinx.coroutines.sync
 
 import kotlinx.coroutines.*
 import org.junit.Test
+import org.junit.After
 import kotlin.test.assertEquals
 
 class SemaphoreStressTest : TestBase() {
 
+    private val pool = newSingleThreadContext("SemaphoreStressTest")
```
elizarov: Please, move `pool` into `stressReleaseCancellation()` test function, since other functions to do not need this single-thread context at the moment.

kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
```diff
@@ -57,4 +65,36 @@ class SemaphoreStressTest : TestBase() {
...
+    @Test
+    fun stressReleaseCancellation() = runTest {
+        val n = 10_000 * stressTestMultiplier
+        val semaphore = Semaphore(1, 1)
+        repeat (n) {
+            assertEquals(0, semaphore.availablePermits)
+
+            var shared = false
+
+            val job1 = launch(pool) {
```
elizarov: A comment would be helpful here, explaining what is being tested and why we need a single-threaded pool.

kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
```diff
@@ -66,35 +59,48 @@ class SemaphoreStressTest : TestBase() {
...
+                val job1 = launch {
+                    semaphore.acquire()
+                    job1_entered_critical_section = true
+                    semaphore.release()
+                }
 
-            job1.cancelAndJoin()
-            job2.join()
+                // Let `job1` run and sleep in `acquire`
+                yield()
```
elizarov: You can achieve the same effect with less code -- just launch `job1` with `launch(start = CoroutineStart.UNDISPATCHED) { ... }`. You'll have no need to use `yield()`.

kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
```diff
@@ -66,35 +59,48 @@ class SemaphoreStressTest : TestBase() {
         assertEquals(1, semaphore.availablePermits)
     }
 
+    /* This checks if repeated releases that race with cancellations put
```
elizarov: style: This should be KDoc commend formatted like a regual KDoc

kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
```diff
@@ -66,35 +59,48 @@ class SemaphoreStressTest : TestBase() {
...
-            assertEquals(1, semaphore.availablePermits)
-            semaphore.acquire()
-        }
+                assertEquals(false, job1_entered_critical_section)
 
-        pool.close()
+                val job2 = launch(pool) {
+                    semaphore.release()
+                }
+
```
elizarov: style: try to avoid extra blank lines inside the code functions. Longer vertical size makes it harder to understand. Separate logical blocks of code with comments or not separate them at all.

kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
elizarov: Ping! Either format this as KDoc:
```
/*
 * Text here (not leading star!)
 */
```
Or use multi-line commens in code (preferred for short comments):
```
// text
// more text
```

dkhalanskyjb: I thought I have fixed this comment with the latest commit: https://github.com/Kotlin/kotlinx.coroutines/pull/1477/commits/63d563df22bb17d3121bdf5f0b830897c34b5a61

Do you mean that I should also replace the block comments in code with a group of single-line comments?

elizarov: You still have badly formatted comments in code like this one:
```
                /* Initially, we hold the permit and no one else can `acquire`,
                   otherwise it's a bug. */
```

elizarov:
🚀 Very good. I'll squash and merge it when it passes tests.

glasser:
How concerned should I be about the use of Semaphores in production code without this bugfix? We are about to deploy a change that introduces the use of Semaphore for the first time and I'm unsure from looking at this change and test what to worry about. It says the problem is that permits are "leaked" — does that mean that the number of available permits may end up being too high or too low?  ie, is the failure mode "my coroutines will run with a higher level of concurrency than I expected" or "eventually my coroutines will all block because there are no permits left for them"?

elizarov:
@glasser The bug only appears when semaphore  `acquire` is cancelled and this reaes with its `release`.  It only affect the code that would be still using a semaphore after that. E.g. it in not a problem in the way we use a `Semaphore` in `Flow` implementation, since there we never reuse a semaphore instance after cancellation. 

P.S. We'll get patch release out soon.

PR #1492, opened 2019-08-31
Fix incorrect currentTimeMillis() call

adamszadkowski:
I have made those changes to fix issue in here: https://kotlinlang.org/docs/reference/coroutines/flow.html

Problem:
image


elizarov:
Good catch, thanks. But changes shall be applied to sources in `docs/flow.md` file, not to the auto-generated `.kt` file (which are actually correct). I've fixed it. See #1496

PR #1494, opened 2019-09-01
Update flow.md

HKhademian:
use `System.currentTimeMillis()` instead of internal one

if you run the previous code one playground you get an error like:

> Cannot access 'currentTimeMillis': it is internal in 'kotlinx.coroutines'


elizarov:
Thanks! Already fixed.

PR #1518, opened 2019-09-05
Add CompletableDeferred.completeWith(result: Result<T>)

glasser:
The naming for this convenience function is inspired by Continuation.resumeWith.

### DIFF
--- a/kotlinx-coroutines-core/common/src/CompletableDeferred.kt
+++ b/kotlinx-coroutines-core/common/src/CompletableDeferred.kt
@@ -46,6 +46,18 @@ public interface CompletableDeferred<T> : Deferred<T> {
 }
 
 /**
+ * Completes this deferred value with the value or exception in the given [result]. Returns `true` if this deferred
+ * was completed as a result of this invocation and `false` otherwise (if it was already completed).
+ *
+ * Subsequent invocations of this function have no effect and always produce `false`.
+ *
+ * This function transitions this deferred in the same ways described by [CompletableDeferred.complete] and
+ * [CompletableDeferred.completeExceptionally].
+ */
+@ExperimentalCoroutinesApi // since 1.3.2, tentatively until 1.4.0
+public fun <T> CompletableDeferred<T>.completeWith(result: Result<T>) = result.fold({ complete(it) }, { completeExceptionally(it) })
+
+/**
  * Creates a [CompletableDeferred] in an _active_ state.
  * It is optionally a child of a [parent] job.
  */
--- a/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
+++ b/binary-compatibility-validator/reference-public-api/kotlinx-coroutines-core.txt
@@ -119,6 +119,7 @@ public final class kotlinx/coroutines/CompletableDeferredKt {
 	public static final fun CompletableDeferred (Ljava/lang/Object;)Lkotlinx/coroutines/CompletableDeferred;
 	public static final fun CompletableDeferred (Lkotlinx/coroutines/Job;)Lkotlinx/coroutines/CompletableDeferred;
 	public static synthetic fun CompletableDeferred$default (Lkotlinx/coroutines/Job;ILjava/lang/Object;)Lkotlinx/coroutines/CompletableDeferred;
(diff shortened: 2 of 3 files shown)


kotlinx-coroutines-core/common/src/CompletableDeferred.kt
```diff
@@ -45,6 +46,17 @@ public interface CompletableDeferred<T> : Deferred<T> {
...
+/**
+ * Completes this deferred value with the value or exception in the given [result]. Returns `true` if this deferred
+ * was completed as a result of this invocation and `false` otherwise (if it was already completed).
+ *
+ * Subsequent invocations of this function have no effect and always produce `false`.
+ *
+ * This function transitions this deferred in the same ways described by [CompletableDeferred.complete] and
+ * [CompletableDeferred.completeExceptionally].
+ */
+public fun <T> CompletableDeferred<T>.completeWith(result: Result<T>) = result.fold({ complete(it) }, { completeExceptionally(it) })
```
elizarov: Style nit: `fold` better works for functions that return something depending on success/failure, and not so good for side-effecting code like. Stylistically, `onSuccess { complete(it) }.onFailure { completeExceptionaly(it) }` would look better here.

glasser: But I am returning the same true/false value that complete/completeExceptionally returns.

kotlinx-coroutines-core/common/src/CompletableDeferred.kt
```diff
@@ -63,18 +75,22 @@ public fun <T> CompletableDeferred(value: T): CompletableDeferred<T> = Completab
  */
 @Suppress("UNCHECKED_CAST")
 private class CompletableDeferredImpl<T>(
-    parent: Job?
+        parent: Job?
```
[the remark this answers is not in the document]
glasser: Sorry, fixed.

kotlinx-coroutines-core/common/src/CompletableDeferred.kt
ZakTaccardi: I find the following to be the most readable:

```kotlin
= result.fold(
  onSuccess = { complete(it) }, 
  onFailure = { completeExceptionally(it) }
)
```

kotlinx-coroutines-core/common/src/CompletableDeferred.kt
```diff
@@ -45,6 +45,17 @@ public interface CompletableDeferred<T> : Deferred<T> {
...
+/**
+ * Completes this deferred value with the value or exception in the given [result]. Returns `true` if this deferred
+ * was completed as a result of this invocation and `false` otherwise (if it was already completed).
+ *
+ * Subsequent invocations of this function have no effect and always produce `false`.
+ *
+ * This function transitions this deferred in the same ways described by [CompletableDeferred.complete] and
+ * [CompletableDeferred.completeExceptionally].
+ */
+public fun <T> CompletableDeferred<T>.completeWith(result: Result<T>) = result.fold({ complete(it) }, { completeExceptionally(it) })
```
elizarov: That looks great. To make it directly mergeable, please add the following annotation and comment:

```
@ExperimentalCoroutinesApi // since 1.3.1, tentatively until 1.4.0
```

Also, you need to run `./gradlew clean :binary-compatibility-validator:test -Poverwrite.output=true` and add resulting change to public API declarations data.

glasser: Should that be 1.3.2? 1.3.1 is already out, right?

qwwdfsad: Yes, it has to be 1.3.2 now

glasser: Thanks, done. It would be helpful if that step was described in the contributions section of the README.

qwwdfsad: This step is non-deterministic :)

qwwdfsad:
Thanks for your contribution!

kotlinx-coroutines-core/common/src/CompletableDeferred.kt
glasser: Sorry, I just meant regenerating the binary compatibility files.

PR #1520, opened 2019-09-06
Update lincheck to 2.5

ndkoval:
Update lincheck to 2.5 and re-write tests with `suspend` operations using it


ndkoval:
It also fixes a linearizability bug in `ArrayChannel.isClosedForReceive`: #1530

kotlinx-coroutines-core/jvm/test/linearizability/ChannelsLCStressTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    Param(name = "value", gen = IntGen::class, conf = "1:5"),
+    Param(name = "closeToken", gen = IntGen::class, conf = "1:3")
+)
+class ChannelsLCStressTestImpl: VerifierState() {
+    private val c = channelKind.create()
+
+    @Operation
+    suspend fun send(@Param(name = "value") value: Int) = try {
+        c.send(value)
+    } catch (e : NumberedCancellationException) {
```
elizarov: Style nit: No space before `:` when used to specify variable type (in all the other cases in this file, too)

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeTaskQueueLCStressTest.kt
```diff
@@ -5,89 +5,54 @@
...
+            .check(this::class)
     }
-
-    override fun hashCode(): Int = Snapshot(q).hashCode()
 }
 
-@Param(name = "value", gen = IntGen::class, conf = "1:3")
-class MCLockFreeTaskQueueLCStressTest : LockFreeTaskQueueLCTestBase() {
-    private val q: LockFreeTaskQueue<Int> = LockFreeTaskQueue(singleConsumer = false)
+private var singleConsumer = false
```
elizarov: I don't see any need to make it top-level var add `singleConsumer = when (this::class) { ... }` above. It can be a constructor parameter to `LockFreeTaskQueueWithoutRemoveLCStressTest` class that both inherited classes (`MCLockFreeTaskQueueWithRemoveLCStressTest` and `SCLockFreeTaskQueueWithRemoveLCStressTest`) that are actually being tested pass an appropriate value of.

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -40,7 +40,7 @@ internal open class ArrayChannel<E>(
     private var size: Int = 0 // Invariant: size <= capacity
 
     protected final override val isBufferAlwaysEmpty: Boolean get() = false
-    protected final override val isBufferEmpty: Boolean get() = size == 0
+    protected final override val isBufferEmpty: Boolean get() = lock.withLock { size == 0 }
```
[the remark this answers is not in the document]
ndkoval: I would not be so scared since
1. this is actually not a _fast path_ because it suspends and suspensions are expensive
2. lock-based channel implementations in Golang are not worse in practice
3. this code will be completely removed with the new fast and scalable channels
4. this extra lock acquisition does not affect uncontended scenarios 

At the same time, your suggestion breaks the semantics of `isBufferEmpty` and makes the code harder to understand.

kotlinx-coroutines-core/common/test/TestBase.common.kt
```diff
@@ -10,10 +10,10 @@ import kotlinx.coroutines.flow.*
 import kotlin.coroutines.*
 import kotlin.test.*
 
-public expect open class TestBase constructor() {
-    public val isStressTest: Boolean
-    public val stressTestMultiplier: Int
+public expect val isStressTest: Boolean
```
elizarov: Please, extract this into a separate commit and a separate PR. It would great to have it asap and separately.

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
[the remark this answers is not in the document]
ndkoval: >  Moreover, it looks scary, since this is an atomic multi-word operation that other threads might be helping with.
This change adds only one read under the lock -- seems like a safe change

kotlinx-coroutines-core/common/test/TestBase.common.kt
ndkoval: See #1579

kotlinx-coroutines-core/jvm/test/LCStressOptionsDefault.kt
```diff
@@ -0,0 +1,19 @@
...
+package kotlinx.coroutines
+
+import org.jetbrains.kotlinx.lincheck.*
+import org.jetbrains.kotlinx.lincheck.strategy.stress.*
+import kotlin.reflect.*
+
+class LCStressOptionsDefault : StressOptions() {
+    init {
+        iterations(100 * stressTestMultiplierSqrt)
+        invocationsPerIteration(20_000 * stressTestMultiplierSqrt)
```
elizarov: This makes tests too slow on regular builds. See https://teamcity.jetbrains.com/viewLog.html?buildId=2571931&buildTypeId=KotlinTools_KotlinxCoroutines_BuildAggregated&tab=testsInfo
For example, `LockFreeListLCStressTest.testAddRemoveLinearizability` now takes 23 seconds.

kotlinx-coroutines-core/jvm/test/linearizability/ChannelsLCStressTest.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    @Operation
+    fun isClosedForReceive() = c.isClosedForReceive
+
+    @Operation
+    fun isClosedForSend() = c.isClosedForSend
+
+    // TODO: this operation should be (and can be!) linearizable, but is not
+    // @Operation
+    fun isEmpty() = c.isEmpty
```
elizarov: Can we still test `isEmpty` for rendezvous channels where it should be linearizable? It is not a short-stopper if we cannot, but would be a nice thing to have.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeTaskQueueLCStressTest.kt
```diff
@@ -5,89 +5,48 @@
...
 
     @Operation
     fun addLast(@Param(name = "value") value: Int) = q.addLast(value)
 
-    /**
-     * Note that removeFirstOrNull is not linearizable w.r.t. to addLast, so here
-     * we test only linearizability of close.
-     */
-//    @Operation(group = "consumer")
-//    fun removeFirstOrNull() = q.removeFirstOrNull()
```
elizarov: Please, leave this commented out section together with comment in the code.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeTaskQueueLCStressTest.kt
```diff
@@ -5,89 +5,48 @@
...
+    fun testWithoutRemove() = LCStressOptionsDefault().check(this::class)
 
-        other as SCLockFreeTaskQueueLCStressTest
-
-        return Snapshot(q) == Snapshot(other.q)
+    @Test
+    fun testWithRemoveForQuiescentConsistency() {
+        LCStressOptionsDefault()
+            .verifier(QuiescentConsistencyVerifier::class.java)
+            .actorsPerThread(if (isStressTest) 3 else 2)
```
elizarov: This seems to be a default `actorsPerThread` setting and it can be safely removed from here.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeListLCStressTest.kt
```diff
@@ -5,16 +5,16 @@
...
 import kotlinx.coroutines.internal.*
+import org.jetbrains.kotlinx.lincheck.annotations.*
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck.paramgen.*
+import org.jetbrains.kotlinx.lincheck.verifier.*
 import kotlin.test.*
 
-@Param(name = "value", gen = IntGen::class, conf = "1:3")
-class LockFreeListLCStressTest : TestBase() {
+@Param(name = "value", gen = IntGen::class, conf = "1:5")
```
elizarov: This test should use a narrow range of values (`1:3`), because the key operation whose linearizability is checked here is `removeFirstOrPeekIfNotSame` which must have a high change of using the same value as the recently added nodes.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeListLCStressTest.kt
```diff
@@ -44,29 +44,11 @@ class LockFreeListLCStressTest : TestBase() {
...
-        return curElements == other.curElements
-    }
+    fun testAddRemoveLinearizability() = LCStressOptionsDefault().check(this::class)
 
-    override fun hashCode(): Int {
-        return curElements.hashCode()
+    override fun extractState(): Any {
+        val elements = ArrayList<Int>()
+        q.forEach<Node> { elements.add(it.value) }
+        return elements
```
elizarov: It would be also very useful to add call to `q.validate()` here which also checks that the resulting doubly-linked list has consistent links in both directions.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeListLCStressTest.kt
```diff
@@ -44,29 +44,11 @@ class LockFreeListLCStressTest : TestBase() {
...
-            q.forEach<Node> { _curElements!!.add(it.value) }
-        }
-        return _curElements!!
-    }
-
-    override fun equals(other: Any?): Boolean {
-        other as LockFreeListLCStressTest
-        return curElements == other.curElements
-    }
+    fun testAddRemoveLinearizability() = LCStressOptionsDefault().check(this::class)
```
elizarov: Note, that previous version of LinCheck was running this test with 5 actors per thread and it turns out this was quite important to increase probability of catching certain problems. I'd recommend to leave it this way for this particular test.

kotlinx-coroutines-core/jvm/test/linearizability/ChannelsLCStressTest.kt
ndkoval: Technically: yes, we can. For example, we can either always return `false` for other channels or create a special test for rendezvous one. However, it makes the code uglier. I would NOT test `isEmpty` since we have never had a test for it, and everything will be fixed with new channels.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeListLCStressTest.kt
ndkoval: This code is used only for verification, to specify a state machine. Unfortunately, _lincheck_ does not have a mechanism to validate the data structure at the end of each invocation; we can add this feature to _licheck_.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeListLCStressTest.kt
ndkoval: OK.

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeTaskQueueLCStressTest.kt
ndkoval: We have tests with `removeFirstOrNull`

kotlinx-coroutines-core/jvm/test/linearizability/LockFreeListLCStressTest.kt
elizarov: We don't have to wait for new lincheck. We can add it later.

elizarov:
Still fails on tests.

ndkoval:
I also added a sequential channel implementation

PR #1521, opened 2019-09-07
Fixes linearizability of Channel.close in advanced receive+send case

elizarov:
We cannot resume closed receives until all receivers are removed from the list.
Consider channel state: head -> [receive_1] -> [receive_2] -> head
 - T1 called receive_2, and will call send() when it's receive call resumes
 - T2 calls close()

Now if T2's close resumes T1's receive_2 then it's receive gets
"closed for receive" exception, but its subsequent attempt to send
successfully rendezvous with receive_1, producing non-linearizable
execution.

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/InlineList.kt
+++ b/kotlinx-coroutines-core/common/src/internal/InlineList.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+@file:Suppress("UNCHECKED_CAST")
+
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.assert
+
+/*
+ * Inline class that represents a mutable list, but does not allocate an underlying storage
+ * for zero and one elements.
+ * Cannot be parametrized with `List<*>`.
+ */
+internal inline class InlineList<E>(private val holder: Any? = null) {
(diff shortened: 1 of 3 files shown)


qwwdfsad:
Extracted "in-place list" pattern into InlineList class, PTAL.

kotlinx-coroutines-core/common/src/internal/InlineList.kt
```diff
@@ -0,0 +1,46 @@
...
+ * Inline class that represents a mutable list, but does not allocate an underlying storage
+ * for zero and one elements.
+ * Cannot be parametrized with `List<*>`.
+ */
+internal inline class InlineList<E>(private val holder: Any? = null) {
+    public operator fun plus(element: E): InlineList<E>  {
+        assert { element !is List<*> } // Lists are prohibited
+        return when (holder) {
+            null -> InlineList(element)
+            is MutableList<*> -> {
```
elizarov: I would suggest to check for `is ArrayList` here for performance reason.

kotlinx-coroutines-core/common/src/internal/InlineList.kt
```diff
@@ -0,0 +1,46 @@
...
+                list.add(element)
+                InlineList(list)
+            }
+        }
+    }
+
+    public inline fun forEachReversed(action: (E) -> Unit) {
+        when (holder) {
+            null -> return
+            !is List<*> -> action(holder as E)
```
elizarov: Same here.

PR #1523, opened 2019-09-09
Minor documentation improvements

qwwdfsad:


kotlinx-coroutines-core/common/src/flow/Builders.kt
```diff
@@ -285,6 +285,9 @@ public fun <T> channelFlow(@BuilderInference block: suspend ProducerScope<T>.()
  *     awaitClose { api.unregister(callback) }
  * }
  * ```
+ *
+ * This function is an alias for [channelFlow], it has a separate name to reflect
```
qwwdfsad: This one raised a lot of questions in Slack

elizarov: There was also a proposal to deprecate `callbackFlow` and leave only `channelFlow`. It might have a merit.

PR #1538, opened 2019-09-16
Optimize Flow.take

qwwdfsad:


benchmarks/src/jmh/kotlin/benchmarks/flow/TakeBenchmark.kt
```diff
@@ -0,0 +1,141 @@
...
+        (0L until size).asFlow().consume()
+    }
+
+    @Benchmark
+    fun originalTake() = runBlocking<Int> {
+        (0L..Long.MAX_VALUE).asFlow().originalTake(size).consume()
+    }
+
+    @Benchmark
+    fun fastPathTake() = runBlocking<Int> {
```
qwwdfsad: Both `fastPathTake` and `mergedStateMachineTake` yield (mostly) the same performance characteristics.
In theory, merged SM can be faster because it avoids int boxing and (potentially) can avoid allocating `emitFun`, though it will complicate things (especially because this approach is not multiplatform), so I've decided to stick with `fastPath`

benchmarks/src/jmh/kotlin/benchmarks/flow/TakeBenchmark.kt
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        return unsafeFlow {
+            var consumed = 0
+            try {
+                collect { value ->
+                    if (++consumed < count) {
+                        return@collect emit(value)
+                    } else {
+                        return@collect emitAbort(value)
```
qwwdfsad: https://youtrack.jetbrains.com/issue/KT-33869

benchmarks/src/jmh/kotlin/benchmarks/flow/TakeBenchmark.kt
elizarov: Good.

benchmarks/src/jmh/kotlin/benchmarks/flow/TakeBenchmark.kt
```diff
@@ -0,0 +1,141 @@
...
+            } catch (e: StacklessCancellationException) {
+                // Nothing, bail out
+            }
+        }
+    }
+
+
+    public fun <T> Flow<T>.fastPathTake(count: Int): Flow<T> {
+
+        suspend fun FlowCollector<T>.emitAbort(value: T) {
```
elizarov: You'd get slightly more efficient bytecode is you make `emitAbort` a top-level private function. Local functions currently get cached in the outer function instance even if they don't capture anything from the outer scope.

elizarov:
Otherwise it is good to go.

PR #1546, opened 2019-09-17
Non-linearizable cancel

qwwdfsad:
This changeset does not make `cancel` completely linearizable.

It seriously reduces the amount of non-linearizable executions by mimicking `close` and makes `cancel` more approachable and understandable for unprepared users.

I have a separate changeset that makes cancel completely linearizable, but it requires a lot of ceremonies:

  * OpDescriptor and helping to make `cancel` and `close` linearizable with each other
  * Every receive-like operation should check whether the channel was cancelled: `Closed` token is added to the end of the list and fast-path of, e.g. ArrayChannel don't even touch the linked list

Let's discuss the overhead of complete approach separately, but these changes (maybe modulo renaming `onCancelIdempotent ` because it's currently not required to be so) are worth integrating anyway


elizarov:
Looks good to me!

PR #1549, opened 2019-09-17
Promote reactive bridges for Flow to stable API

qwwdfsad:


qwwdfsad:
`Flow.asFlowable()` -- looks reasonable, it piggybacks on `asPublisher` anyway.
`Flow.asObservable()` -- not really, I am not happy with the implementation (even though each choice in the implementation is properly tested). I'd rather wait for demand on this API to become stable before promoting it.
`RxConvert.kt` -- the same story. I am not sure if these extensions are useful at all.

PR #1560, opened 2019-09-22
Migrate BOM module to java-platform plugin

Prototik:


Prototik:
Main goal of #1371 was adding -js and -native libs, not platform-specific (like -linuxx64 and -macosx64).

That caused inconsistency when generating bom on different platforms: on linux box bom it will include both -native and -linuxx64, but not -macosx64 (and on macos it will include -native and -macoxx64, but no linuxx64) as kotlin plugin doesn't register publications for unavailable targets (and this makes sense: there is nothing to publish).

Consumers of the lib doesn't need to use specific -linuxx64 artifact, but use -native one and gradle will choose correct variant by using metadata, so it no any reason to keep -linuxx64, -macosx64 (and other native-specific) artifacts in bom - only -native is enough, thus removing inconsistency when releasing on different platforms.

That PR drop -linuxx64, but keeps -native and -js artifacts, so everything is ok.

PR #1563, opened 2019-09-22
Make the list of segments more abstract

ndkoval:
Make the list of segments more abstract, so that it can be used for stacks and channels


kotlinx-coroutines-core/common/src/internal/SegmentList.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    // returns `true` if this segment is logically removed after the decrement
+    inline fun decPointers() = cleanedAndPointers.addAndGet(-(1 shl POINTERS_SHIFT)) == maxSlots
+
+    /**
+     * This functions should be invoked on each slot clean-up;
+     * should not be invoked twice for the same slot.
+     */
+    fun onSlotCleaned() {
+        if (cleanedAndPointers.incrementAndGet() == maxSlots) remove()
```
dkhalanskyjb: In the same vein, this method can be called on a physical tail, and `remove()` may be run even though `removed` is not yet true. Hence, the assertion in `remove()` may fail. Once again, it is fine if we make certain assumptions about how pointers and cells behave: that is, the pointers only move to the right and never disappear, and no cells in segment `n` may be cancelled unless some pointer points to a segment with id of `n` or above. Still, this should probably be kept in mind.

kotlinx-coroutines-core/common/src/internal/SegmentList.kt
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Returns `true` if this segment is logically removed from the queue.
+     * The segment is considered as removed if all the slots are cleaned,
+     * there is no pointers to this segment from outside, and
+     * it is not a physical tail in the linked list of segments.
+     */
+    inline val removed get() = cleanedAndPointers.value == maxSlots && _next.value !== null
+
+    // increments the number of pointers if this segment is not logically removed\
+    inline fun tryIncPointers() = cleanedAndPointers.addConditionally(1 shl POINTERS_SHIFT) { it != maxSlots }
```
dkhalanskyjb: The segment may still be present (not logically removed) when this method returns `false` if it's the tail segment.

This may be a significant problem. Assume we have segments 1 and 2, with `tail` pointing to `segment_1`, and all the cells in 2 are cancelled (as with the comment for `moveForward`, this can't ever happen in the semaphore implementation, for the same reasons). If we call `tail.findSegmentAndMoveForward(2, segment_1, ...)`, `findSegmentInternal` returns `segment_2`. It isn't closed, and `moveForward` fails because the call `tryIncPointers` fails, so we loop. The situation hasn't changed, so we may loop forever.

kotlinx-coroutines-core/common/src/internal/SegmentList.kt
```diff
@@ -0,0 +1,213 @@
...
+// Returns `false` if the segment `to` is logically removed, `true` on successful update.
+private inline fun <S : Segment<S>> AtomicRef<S>.moveForward(to: S): Boolean = loop { cur ->
+    if (cur.id >= to.id) return true
+    if (!to.tryIncPointers()) return false
+    if (compareAndSet(cur, to)) {
+        // the segment is moved
+        if (cur.decPointers()) cur.remove()
+        return true
+    } else {
+        if (to.decPointers()) to.remove()
```
dkhalanskyjb: Assume the following execution:
* `cur` points to segment 0, and `to` has id 2 and is the physical tail. There are no pointers to `to` yet; for example, `cur` could be the tail pointer.
* `cur.moveForward(to)` is called. All the checks before the CAS pass.
* `cur.moveForward(segment_1)` is called and completed, where `segment_1` is the segment with `id = 1`.
* The CAS fails.
* The last cell is cancelled in the segment `to`.
* `to.decPointers()` returns `true`, and `to.remove()` is called, even though `to.removed` is not true.

This can never happen in the provided semaphore implementation because in order to cancel the cell, one must first leave the call to `suspend`, which moves the head pointer at least to the segment of the cell, and so the `head` points either to `to`, in which case `decPointer()` returns `false`, or to something to the right of it, in which case `to` is not a physical tail anymore and the problem isn't present.

kotlinx-coroutines-core/common/src/internal/SegmentList.kt
```diff
@@ -0,0 +1,213 @@
...
+    val isClosed = _next.value === CLOSED
+
+    /**
+     * Removes this segment physically from the segment queue. The segment should be
+     * logically removed (so [removed] returns `true`) at the point of invocation.
+     */
+    fun remove() {
+        assert { removed } // The segment should be logically removed at first
+        // Read `next` and `prev` pointers.
+        var next = this.nextOrClosed.segment ?: return // tail cannot be removed
```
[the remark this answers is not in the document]
ndkoval: Since we move pointers using only `findSegmentAndMoveForward` function, we guarantee that there is always a pointer which references to the last segment, so that  `cleanedAndPointers != 0` for the last segment and we do not invoke `remove` on it.

kotlinx-coroutines-core/common/src/internal/SegmentList.kt
ndkoval: See https://github.com/Kotlin/kotlinx.coroutines/pull/1563#discussion_r345505399

kotlinx-coroutines-core/common/src/internal/SegmentList.kt
ndkoval: Changed the condition: `{ it != maxSlots || _next.value == null }`

kotlinx-coroutines-core/common/src/internal/SegmentList.kt
ndkoval: The problem occurs because `decPointers` has a bug -- it should not return true if the segment is the physical tail (see the documentation). Fixed this bug.

ndkoval:
I have updated the data structure in a way so that it can be used for an efficient list of cancellable `JobNode`s implementation.

PR #1565, opened 2019-09-23
Simpler and faster lock-free linked list

elizarov:
Lock-free list implementation is considerably simplified, taking into
account a limited number of operations that it needs to support.

* prev pointers in the list are not marked for removal, since we
  don't need to support linearizable backwards iteration.
* helpDelete method is completely removed. All "delete-helping" is
  performed only by correctPrev method.
* correctPrev method bails out when the node it works on is removed to
  reduce contention during concurrent removals.
* Special open methods "isRemoved" and "nextIfRemoved" are introduced
  and are overriden in list head class (which is never removed).
  This ensures that on long list "removeFist" operation (touching head)
  does not interfere with "addLast" (touch tail). There is still
  sharing of cache-lines in this case, but no helping between them.

All in all, this improvement reduces the size of implementation code
and makes it considerably faster. Operations on LinkedListChannel are
now much faster (see timings of ChannelSendReceiveStressTest).

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
@@ -40,14 +40,18 @@ public actual typealias PrepareOp = LockFreeLinkedListNode.PrepareOp
  * Doubly-linked concurrent list node with remove support.
  * Based on paper
  * ["Lock-Free and Practical Doubly Linked List-Based Deques Using Single-Word Compare-and-Swap"](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.140.4693&rep=rep1&type=pdf)
- * by Sundell and Tsigas.
+ * by Sundell and Tsigas with considerable changes.
+ *
+ * The core idea of the algorithm is to maintain a doubly-linked list with an ever-present sentinel node (it is
+ * never removed) that serves both as a list head and tail and to linearize all operations (both insert and remove) on
+ * the update of the next pointer. Removed nodes have their next pointer marked with [Removed] class.
  *
  * Important notes:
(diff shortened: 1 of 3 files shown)


elizarov:
DO NOT MERGE. Fails on stress test: https://teamcity.jetbrains.com/viewLog.html?buildId=2544675&buildTypeId=KotlinTools_KotlinxCoroutines_NightlyStressWindows

elizarov:
ConflatedChannel starvation is too pronounced with this version. Need to do something about it. See https://teamcity.jetbrains.com/viewLog.html?buildId=2546814&buildTypeId=KotlinTools_KotlinxCoroutines_NightlyStressWindows

elizarov:
@qwwdfsad Ready for review! Works like a charm. Faster, less code.

elizarov:
@qurbonzoda Woot. Fixed now.

elizarov:
It is ready for review now. Due to the previously merged improvements to `ConflatedChannel` this change is now straightforward and reliably passes all tests, tested on stress-tests, too.

kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
```diff
@@ -233,55 +236,71 @@ public actual open class LockFreeLinkedListNode {
...
+                return null
             }
         }
     }
 
+    // Helps with removal of this node
     public actual fun helpRemove() {
-        val removed = this.next as? Removed ?: error("Must be invoked on a removed node")
-        finishRemove(removed.ref)
+        // Note: this node must be already removed
```
qwwdfsad: it should be asserted then?

elizarov: Cast on the next line will fail if that is not the case.

PR #1570, opened 2019-09-24
Hid the SettableFuture of future {}.

yorickhenning:
This should prevent successful casts to type SettableFuture, meaning
client code can't access and complete the internal Future without
resorting to reflection.


qwwdfsad:
Thanks!

PR #1572, opened 2019-09-25
User Class.forName instead of ServiceLoader to instantiate Dispatchers.Main on Android

### LINKED ISSUE #878: Slow android Dispatchers.Main init

### LINKED ISSUE #1557: Crash after upgrade from 1.3.0 to 1.3.1
Since I updated to coroutines 1.3.1 on Android projects it started to crash on all existing 3rd libraries that using coroutines too. 

It is just minor bugfix version I would not expect backward incompatibility on such release.

But there is this small unimportant note:

> Note: Kotlin/Native artifacts are now published with Gradle metadata format version 1.0, so you will need Gradle version 5.3 or later to use this version of kotlinx.coroutines in your Kotlin/Native project.

Can this be a reason?

My project is using Gradle 5.6.2, but the library is using an older version than 5.3

```
    java.lang.IllegalStateException: Module with the Main dispatcher is missing. Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android'
        at kotlinx.coroutines.internal.MissingMainCoroutineDispatcher.missing(MainDispatchers.kt:90)
        at kotlinx.coroutines.internal.MissingMainCoroutineDispatcher.isDispatchNeeded(MainDispatchers.kt:71)
        at kotlinx.coroutines.DispatchedKt.resumeCancellable(Dispatched.kt:420)
        at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:26)
        at kotlinx.coroutines.CoroutineStart.invoke(CoroutineStart.kt:109)
        at kotlinx.coroutines.AbstractCoroutine.start(AbstractCoroutine.kt:154)
        at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch(Builders.common.kt:54)
        at kotlinx.coroutines.BuildersKt.launch(Unknown Source:1)
        at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch$default(Builders.common.kt:47)
        at kotlinx.coroutines.BuildersKt.launch$default(Unknown Source:1)
```

I have both dependencies

```
    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.1'
    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.1'

```

qwwdfsad:
`ServiceLoader` has proven itself as a very unstable mechanism on Android (see all the issues in a `loadMainDispatcherFactory` comment). Moreover, failures in SL are considered by `kotlinx.coroutines` users as errors in the library itself, making it look unstable. We are not really happy with that, so here is an attempt to avoid `ServiceLoader` on Android for `Dispatchers.Main`.

Though I am not sure whether this change should be merged in `kotlinx.coroutines`. Moreover, I am not sure how exactly to evaluate the potential impact of this change.
@wojtek-kalicinski and @JakeWharton could you please help with reviewing this change first?

Fixes #1557
Fixes #878

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
@@ -5,6 +5,12 @@ import java.net.*
 import java.util.*
 import java.util.jar.*
 import java.util.zip.*
+import kotlin.collections.ArrayList
+
+/**
+ * Don't use JvmField here to enable R8 optimizations via "assumenosideeffects"
+ */
+internal val ANDROID_DETECTED = runCatching { Class.forName("android.os.Build") }.isSuccess
 
 /**
  * A simplified version of [ServiceLoader].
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
```diff
@@ -20,7 +27,59 @@ import java.util.zip.*
...
+     * that may already be injected vis SL, so we are not using the same technique for it.
+     */
+    internal fun loadMainDispatcherFactory(): List<MainDispatcherFactory> {
+        val clz = MainDispatcherFactory::class.java
+        if (!ANDROID_DETECTED) {
+            return load(clz, clz.classLoader)
+        }
+
+        return try {
+            val result = ArrayList<MainDispatcherFactory>()
```
JakeWharton: nit: pre-size to 2

kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
```diff
@@ -5,6 +5,13 @@ import java.net.*
 import java.util.*
 import java.util.jar.*
 import java.util.zip.*
+import kotlin.collections.ArrayList
+
+/**
+ * TODO investigate possible java.vendor values for fallbacks
+ * Don't use JvmField here to enable R8 optimizations via "assumenosideeffects"
+ */
+internal val ANDROID_DETECTED = systemProp("java.vendor")?.toLowerCase()?.contains("android") ?: false
```
JakeWharton: I've seen this used in the past, but I'm not sure anything actually guarantees that this string is present. Another reliable way to check is to try to load any Android class (e.g., `Class.forName("android.os.Build")` and then set true or false. R8 will eliminate the lookup so it will become a constant value which should then be inlined so the `if` and thus `load` are removed.

kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
```diff
@@ -20,7 +27,59 @@ import java.util.zip.*
...
+    /*
+     * This method is inline to have a direct Class.forName("string literal") in the byte code to avoid weird interactions with ProGuard/R8.
+     */
+    @Suppress("NOTHING_TO_INLINE")
+    private inline fun createInstanceOf(
+        baseClass: Class<MainDispatcherFactory>,
+        serviceClass: String
+    ): MainDispatcherFactory? {
+        return try {
+            val clz = Class.forName(serviceClass, true, baseClass.classLoader)
```
JakeWharton: We should confirm that calls to `forName` in this three-argument form can be replaced with class constants by R8. I'll do that.

qwwdfsad: Thanks!

wojtek-kalicinski:
In principle it looks good. I'm wondering if you then still want to have the R8-specific `assumenosideeffects class kotlinx.coroutines.internal.MainDispatcherLoader { boolean FAST_SERVICE_LOADER_ENABLED` configs... unless you treat them as fallback in case `Class.forName` fails, though I'm not sure if that would ever happen

kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt
JakeWharton: So far found at least one problem: https://issuetracker.google.com/issues/141678220

The use of the three-arg variant of this method also means that R8 cannot (currently) rewrite this into a `const-class` bytecode at compile time. This is probably okay, since this code only runs once. And it's something that can be investigated in the future. I actually hope to get code like this eventually rewritten as a direct call to the constructor!

PR #1579, opened 2019-09-27
Make `isStressTest` and the related properties top-level

ndkoval:


kotlinx-coroutines-core/common/test/TestBase.common.kt
```diff
@@ -10,10 +10,10 @@ import kotlinx.coroutines.flow.*
 import kotlin.coroutines.*
 import kotlin.test.*
 
-public expect open class TestBase constructor() {
-    public val isStressTest: Boolean
-    public val stressTestMultiplier: Int
+public expect val isStressTest: Boolean
```
qwwdfsad: What's the rationale behind this change?

elizarov: It is needed in two places:
* For lincheck tests (which do not extend TestBase)
* To tweak a list of parameters to parameterized tests, which has to happen in a static context using `isStressTest`.

PR #1590, opened 2019-10-01
Make MapTest - testErrorCancelsUpstream - actually test map operator

circusmagnus:
We were testing onEach {} instead of map {} operator there.


qwwdfsad:
Good catch, thanks

PR #1595, opened 2019-10-03
Improve sentence grammar by adding 'an'

itsandreramon:
**The original sentence was:**

"If a coroutine encounters exception other than `CancellationException`, it cancels its parent with that exception."

**I added an additional 'an' before 'exception':**

"If a coroutine encounters **an** exception other than `CancellationException`, it cancels its parent with that exception."


qwwdfsad:
Thanks!

PR #1597, opened 2019-10-03
StackTrace recovery rework

qwwdfsad:
* Stacktraces are no longer recovered in DispatchedTask.run
* Callers of uCont.resumeWithException are responsible for recovering the stacktrace
* CancellableContinuation.resumeWithException recovers stacktrace if necessary


qwwdfsad:
Fixed select, added the test, your changes look good.
Also made cancellation stacktraces much shorter (see testCancelledOffer.txt)

io.svg
```diff
@@ -0,0 +1 @@
+Started [cpu] profiling
```
elizarov: Extra committed file, should not be there.

PR #1598, opened 2019-10-06
Typo in CoroutineScope's javadoc

OlegAndreych:
Seems like a javadoc points that all other children would be cancelled in case any other child coroutine would be cancelled or failed. Changed javadoc accordingly.


qwwdfsad:
Thanks!

PR #1605, opened 2019-10-09
Introduce Stream.consumeAsFlow

### LINKED ISSUE #1601: java.util.Stream.consumeAsFlow() operator
Given a `java.util.Stream`, it can be converted to a coroutine `Flow` in a way that ensures the stream is closed after the flow is finished (consumed or cancelled) like follows:
```
result.iterator().asFlow().onCompletion { result.close() }
```
or, alternatively:
```
flow {
    result.use {
        for (edge in it) {
            emit(edge)
        }
    }
}
```

Would be slightly more convenient to have a `Stream.consumeAsFlow()` implementation in the coroutines libraries that does so even less verbosely.

qwwdfsad:
* Even though java.lang.Stream is collected rather than consumed, collectAsFlow will clash with Flow terminology where collect is a terminal operator
    * Close the stream in the end of collection despite the fact that regular terminal operations don't do that. We are already in suspending world (empty close() call won't make any difference in a common case) and "consume" implies closing the underlying resource (note that we already do it for channels)
    * Remove obsolete examples from the module

Fixes #1601


elizarov:
LGTM.

PR #1611, opened 2019-10-12
Fixes typo

ashishkujoy:


elizarov:
This is not a typo, but an emphasis on the word "ARE". Any, I'd agree it is worth removing, but you'll also need to run `./gradlew knit` and commit the resulting changes in all the automatically generated files.

ashishkujoy:
@elizarov To build the project i will require both java6 and java8 currently i am unable to install java6 on my machine. Do we have any docker image setup for dependency to build project inside container. I guess that would be helpful other contributors as well.

PR #1622, opened 2019-10-18
Fix JobSupport comment about new state.

langara:


qwwdfsad:
It's not really a typo, the initial state could be both `EMPTY_NEW` and `EMPTY_ACTIVE`

langara:
I understand that job can be created as active, but I still think it's better to use EMPTY_NEW in this example, especially that there is a comment above saying that this is the longest possible scenario.

qwwdfsad:
Good point, thanks for the contribution!

PR #1627, opened 2019-10-22
TimeoutCancellationException recovery

qwwdfsad:


kotlinx-coroutines-core/common/src/Timeout.kt
```diff
@@ -96,19 +96,21 @@ private open class TimeoutCoroutine<U, in T: U>(
...
     /**
      * Creates a timeout exception with the given message.
      * This constructor is needed for exception stack-traces recovery.
      */
     @Suppress("UNUSED")
     internal constructor(message: String) : this(message, null)
+
+    // message is never null in fact
+    override fun createCopy(): TimeoutCancellationException? =
+        TimeoutCancellationException(message ?: "", coroutine).also { it.initCause(this) }
```
vlsi: Does it make sense to retain `suppressed` exceptions as well?

qwwdfsad: They will be retained in `cause`

vlsi: Consider:

```kotlin
try {
    delay(100);
} catch(e: TimeoutCancellationException) {
    throw e.also { it.addSuppressed("Oh, suppressed") }
}
```

The suppressed exceptions are not added to `cause`, so the copy would strip the suppressed one.

qwwdfsad: Then it should be fixed not in `createCopy` method, but in our internal machinery because here you already have a copied instance of the exceptions.

Suppressed exceptions are a subtle topic that we want to address more thoughtfully as soon as we give up Java 6 (hopefully, in 1.4 release)

PR #1629, opened 2019-10-22
TestCoroutineDispatcher Intentional Time Skew

Bill:
TestCoroutineDispatcher always picks the same interleaving of concurrent tasks. This is good since it is deterministic. But it is bad because there is no way to explore other interleavings (run order) of the tasks.

If TestCoroutineDispatcher could explore different interleavings, more bugs could be found. The challenge is to add the exploration ability without losing determinism.

This draft PR (for issue #1630) adds a test `TestConcurrentCounterTest.kt`. The test sets up a coroutine that acts an integer register. Then it runs a couple "incrementer" coroutines to read-modify-write to that "register" (through channels). Of course this read-modify-write approach is classically wrong and leads to "lost updates" under certain interleavings. The purpose of the various `@Test` functions in that file, is to show:

1. that a nondeterministic test often finds the bug (see `nondeterministicTestN()`)
2. a deterministic test using the built-in `TestCoroutineDispatcher` behavior (before the proposed changes) fails to find the bug&mdash;the test _passes_ every time (see `deterministicTestN()`)
3. a deterministic test using the simplest-possible change to `TestCoroutineDispatcher` that finds the bug&mdash;the test _fails_ every time (see `deterministicInaccurateDispatcherTestN()`)

That simplest-possible change to `TestCoroutineDispatcher` is to introduce random time "skew" into `postDelayed()` and `post()`. The randomness is driven by a pseudo-random number generator injected through the `TestCoroutineScope` constructor.

If you look at `deterministicInaccurateDispatcherTestN()` you'll see that this arrangement lets us explore different random seeds to give us various task interleavings. When we find an interleaving we want to replay we can "lock it down". You might imagine budgeting a certain amount of test resources for this exploration. When "troublesome" interleavings were identified we might record them as part of a regression test suite.

## Related Work

This work was inspired by the work of the FoundationDB folks. See Will Wilson's 2014 Strange Loop talk _Testing Distributed Systems w/ Deterministic Simulation_ https://www.youtube.com/watch?v=4fFDFbi3toc

See the FoundationDB page on _Simulation and Testing_: https://apple.github.io/foundationdb/testing.html#

See Carl Lerche's October, 2019 blog post _Making the Tokio scheduler 10x faster_ https://tokio.rs/blog/2019-10-scheduler/ Specifically the _Fearless (unsafe) concurrency with Loom_ section https://tokio.rs/blog/2019-10-scheduler/#fearless-unsafe-concurrency-with-loom Loom is a "a permutation testing tool for concurrent code" for Rust programs.

## Opportunities for Improvement

The approach offered in this PR is intentionally minimalistic. It could be refactored so that different (interleaving) exploration mechanisms and policies could be tried.

One such policy that might be interesting to explore is dynamic partial-order reduction described here https://users.soe.ucsc.edu/~cormac/papers/popl05.pdf and used in Lerche's Loom (mentioned above).

### DIFF
--- a/docs/exception-handling.md
+++ b/docs/exception-handling.md
@@ -189,7 +189,7 @@ Parent is not cancelled
 
 <!--- TEST-->
 
-If a coroutine encounters exception other than `CancellationException`, it cancels its parent with that exception. 
+If a coroutine encounters an exception other than `CancellationException`, it cancels its parent with that exception. 
 This behaviour cannot be overridden and is used to provide stable coroutines hierarchies for
 [structured concurrency](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/composing-suspending-functions.md#structured-concurrency-with-async) which do not depend on 
 [CoroutineExceptionHandler] implementation.
--- a/docs/flow.md
+++ b/docs/flow.md
@@ -708,13 +708,15 @@ fun main() = runBlocking<Unit> {
 
 This code produces the following exception:
 
-<!--- TEST EXCEPTION
+```text
 Exception in thread "main" java.lang.IllegalStateException: Flow invariant is violated:
 		Flow was collected in [CoroutineId(1), "coroutine#1":BlockingCoroutine{Active}@5511c7f8, BlockingEventLoop@2eac3323],
 		but emission happened in [CoroutineId(1), "coroutine#1":DispatchedCoroutine{Active}@2dae0000, DefaultDispatcher].
 		Please refer to 'flow' documentation or use 'flowOn' instead
 	at ...
--->
+``` 
+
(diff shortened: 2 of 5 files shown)


qwwdfsad:
We're currently completely rewriting our test coroutine dispatcher and it seems that this PR completely fell off from our sight, I"m sorry for that :(

We actually discussed this issue in our design (kudos to @dkhalanskyjb for taking care of it!) and decided that it's not is our first priorities in our test module, but we may eventually return to it.
The biggest problem here is to provide a concise, but not too intrusive (relatively other test routines) API that is hard to get wrong by end-users. Alas we haven't succeeded to do it in a short timeframe and decided to postpone it

PR #1638, opened 2019-10-29
Consistently handle undeliverable exceptions in RxJava and Reactor in…

### LINKED ISSUE #252: rxAwait calls the uncaught exception handler
Why does the following code calls the uncaught exception handler?

 As there is an onError supplied I expect it to be called instead.

```kotlin
import io.reactivex.Observable
import io.reactivex.Single
import kotlinx.coroutines.experimental.rx2.await
import kotlinx.coroutines.experimental.rx2.rxSingle
import java.io.IOException
import java.util.concurrent.TimeUnit

  Observable
    .interval(1, TimeUnit.MILLISECONDS)
    .switchMapSingle {
      rxSingle {
        timeBomb().await()
      }
    }
    .subscribe({}, {})

  Thread.sleep(1000)
}

private fun timeBomb() = Single.timer(1, TimeUnit.MILLISECONDS)
  .doOnSuccess { throw IOException("") }
```

### LINKED ISSUE #1614: Leaked exception in rx integration
Repro:
```
RxJavaPlugins.setErrorHandler { }
val flow = rxSingle<Unit> { throw IOException("") }.toFlowable().asFlow()
runBlocking {
    repeat(10000) {
        combine(flow, flow) { _, _ -> Unit }
            .catch {}
            .collect { }
    }
}
```

qwwdfsad:
…tegrations

Use tryOnError in RxJava to make exception delivery check-and-act race free.
Deliver undeliverable exceptions via RxJavaPlugins instead of handleCoroutineException.
This is a deliberate choice for a multiple reasons:
    * When using Rx (whether with coroutines or not), undeliverable exceptions are inevitable and users should hook into RxJavaPlugins anyway. We don't want to force them using Rx-specific CoroutineExceptionHandler all over the place
    * Undeliverable exceptions provide additional helpful stacktrace and proper way to distinguish them from other unhandled exceptions
    * Be consistent with reactor where we don't have try*, thus cannot provide a completely consistent experience with CEH (at least, without wrapping all the subscribers)

Do the similar in Reactor integration, but without try*, Reactor does not have notion of undeliverable exceoptions and handles them via Operators.* on its own.

Also, get rid of ASCII tables that does not properly render in IDEA

Fixes #252
Fixes #1614

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/Publish.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Publish.kt
@@ -17,17 +17,14 @@ import kotlin.internal.LowPriorityInOverloadResolution
 
 /**
  * Creates cold reactive [Publisher] that runs a given [block] in a coroutine.
- * Every time the returned publisher is subscribed, it starts a new coroutine.
- * Coroutine emits items with `send`. Unsubscribing cancels running coroutine.
+ * Every time the returned flux is subscribed, it starts a new coroutine in the specified [context].
+ * Coroutine emits ([Subscriber.onNext]) values with `send`, completes ([Subscriber.onComplete])
+ * when the coroutine completes or channel is explicitly closed and emits error ([Subscriber.onError])
+ * if coroutine throws an exception or closes channel with a cause.
+ * Unsubscribing cancels running coroutine.
  *
  * Invocations of `send` are suspended appropriately when subscribers apply back-pressure and to ensure that
  * `onNext` is not invoked concurrently.
- *
- * | **Coroutine action**                         | **Signal to subscriber**
- * | -------------------------------------------- | ------------------------
- * | `send`                                       | `onNext`
- * | Normal completion or `close` without cause   | `onComplete`
- * | Failure with exception or `close` with cause | `onError`
  *
  * Coroutine context can be specified with [context] argument.
  * If the context does not have any dispatcher nor any other [ContinuationInterceptor], then [Dispatchers.Default] is used.
@@ -43,7 +40,7 @@ public fun <T> publish(
 ): Publisher<T> {
     require(context[Job] === null) { "Publisher context cannot contain job in it." +
             "Its lifecycle should be managed via subscription. Had $context" }
-    return publishInternal(GlobalScope, context, block)
+    return publishInternal(GlobalScope, context, DEFAULT_HANDLER, block)
 }
 
 @Deprecated(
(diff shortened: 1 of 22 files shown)


qwwdfsad:
Can't say I am excited or happy about this fix, but it looks like the lesser evil

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -17,18 +17,15 @@ import kotlin.internal.LowPriorityInOverloadResolution
 
 /**
  * Creates cold reactive [Publisher] that runs a given [block] in a coroutine.
- * Every time the returned publisher is subscribed, it starts a new coroutine.
- * Coroutine emits items with `send`. Unsubscribing cancels running coroutine.
+ * Every time the returned flux is subscribed, it starts a new coroutine in the specified [context].
+ * Coroutine [emits][Subscriber.onNext] values with `send`, [completes][Subscriber.onComplete]
```
elizarov: The names of subscriber methods (`onNext`, `onComplete`, and `onError`) are now hidden from the readers of docs (you can only see them if you click a link), while "emits", "completes", and "emits error" terminology might not be truly idiomatic for reactive world. I'd suggest to give them in parenthesis like this:

"Coroutines emits (`onNext`) values with `send` ...."

(Same applies to `flux`, `mono`, and `rxXxx` builders)

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -64,18 +59,24 @@ private class MonoCoroutine<in T>(
...
 ) : AbstractCoroutine<T>(parentContext, true), Disposable {
-    var disposed = false
+    @Volatile
+    private var disposed = false
 
     override fun onCompleted(value: T) {
-        if (!disposed) {
-            if (value == null) sink.success() else sink.success(value)
-        }
+        if (value == null) sink.success() else sink.success(value)
```
elizarov: Removal of `if (!disposed) ...` check around this emission and also inside `onCancelled` function below makes behavior of `mono` builder inconsistent with the behavior of `flux` , which does not call any methods on subscriber after it was disposed. 

Note, that the reactive streams specs does not actually prohibit calling any `onXxx` methods in this case. It just says in Rule 8: "If a Subscription is cancelled its Subscriber MUST eventually stop being signaled."

Same question applies to `RxCompletableCoroutine` changes.

qwwdfsad: It is, in fact, a really small inconsistency.
To make this race happen, a subscriber should manually call `dispose` on its subscription while coroutine is in its completion sequence (CAS in JobSupport is successful, so `job.cancel` doesn't affect anything). In this case, `if (!disposed)` makes a really small race window even smaller, but does not prevent it anyway. 

Adding `if (!isDisposed)` adds a cognitive load for readers, because it's "check-and-act" on some volatile state that is hard to reason about (yeah, we can leave a comment, but what's the point?), but does not change the semantics at all (because "race" is still possible).

Similar reasoning can be applied to `onCancelled` and Rx.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
qwwdfsad: Made it clear in multi-value builders, left no references at all in single-value (mono, single, etc.) because it's pretty straightforward

PR #1643, opened 2019-11-02
Clarify documentation around Dispatcher.IO parallelism

wojtek-kalicinski:
Reference: https://youtrack.jetbrains.com/issue/KT-34671


kotlinx-coroutines-core/jvm/src/Dispatchers.kt
```diff
@@ -101,6 +101,11 @@ public actual object Dispatchers {
      * "`kotlinx.coroutines.io.parallelism`" ([IO_PARALLELISM_PROPERTY_NAME]) system property.
      * It defaults to the limit of 64 threads or the number of cores (whichever is larger).
      *
+     * Moreover, the maximum configurable number of threads is capped by the
+     * `kotlinx.coroutines.scheduler.max.pool.size` system property.
+     * If you need a higher number of parallel threads,
```
fvasco: "If you need a higher number of ~~parallel~~ threads"

PR #1652, opened 2019-11-08
CoroutineScheduler rework

### LINKED ISSUE #840: Dispatchers.Default consumes too much CPU in the face of short bursts of work

### LINKED ISSUE #1046: High cpu utilization with ktor on Android using `GlobalScope.launch`
After switching to the fixed thread pool dispatcher in ktor(https://github.com/ktorio/ktor/commit/990ccfb2ab5371d6b1babf7645b2c5da17a46236) `GlobalScope.launch` cpu utilization is much higher than `client.launch`:
```kotlin
val client = HttpClient {
    install(WebSockets)
}

/* client.launch */
GlobalScope.launch {
    client.ws {
        while(true) {
           // ping
           session?.send(Frame.Text("Hello"))
           // pong
           val frame = session?.incoming?.receive()
        }
    } 
}
```

### LINKED ISSUE #1286: Android Thread handling issue
After trying to better understand why there was so much threads created for coroutines, I pushed the tests a little further and face something strange.

For tests I only use a dedicated ThreadPool with ```asCoroutineDispatcher()``` and a copy of the Main dispatcher to avoid disk access until 1.3 is out.

0 calls to Dispatchers.IO/Default/Main.

At application startup I have 3 to 4 DefaultDispatcher threads created, then over the course of the usage of the application more will be created until reaching the device max core count (8 in that cases).
Meaning 8 threads started that are not really used. (Probably a little for dispatching?)

Source of the issue was that when using also Dispatchers.IO the app always had 12 to 16 DefaultDispatchers threads even when doing nearly nothing. IO threads seemed to never be released and should probably not even have been started. (All long running tasks are started in a specific dispatcher with a lower thread priority, most IO calls are made via a worker pool limiting concurrency)

qwwdfsad:
Fixes #840
Fixes #1046
Fixes #1286


kotlinx-coroutines-core/jvm/test/scheduling/BlockingCoroutineDispatcherThreadLimitStressTest.kt
```diff
@@ -0,0 +1,71 @@
...
+            tasks.forEach { it.await() }
+            for (i in CORES_COUNT + 1..CORES_COUNT * 2) {
+                require(i !in observedConcurrency.keys) { "Unexpected state: $observedConcurrency" }
+            }
+            checkPoolThreadsCreated(0..CORES_COUNT + 1)
+        }
+    }
+
+    @Test
+    @Ignore
```
qwwdfsad: Limit parallelism tests are ignored because of the current implementation of `LimitingDispatcher`.
It should avoid signaling additional work from `afterTask` to avoid overprovision.

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -23,15 +26,13 @@ internal const val MASK = BUFFER_CAPACITY - 1 // 128 by default
...
- * As an alternative, offloading directly to some [CoroutineScheduler.Worker] may be used, but then the strategy of selecting any idle worker
- * should be implemented and implementation should be aware multiple producers.
- *
- * @suppress **This is unstable API and it is subject to change.**
+ * ### Algorithm and implementation details
+ * This is a regular SPMC bounded queue with the additional property that tasks can be removed from the middle of the queue
+ * (scheduler workers without a CPU permit steal blocking tasks via this mechanism). Such property enforces us to use CAS in
+ * order to properly claim value from the buffer.
+ * Moreover, [Task] objects are reusable, so it may seem that this queue is prone to ABA problem.
+ * Indeed it formally has ABA-problem, but the whole processing logic is written in the way that such ABA is harmless.
```
qwwdfsad: NB this is the most subtle part of the algorithm

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -78,23 +86,26 @@ internal class CoroutineScheduler(
         }
     }
 
-    private val globalQueue: GlobalQueue = GlobalQueue()
+    @JvmField
+    val globalCpuQueue = GlobalQueue()
+    @JvmField
+    val globalBlockingQueue = GlobalQueue()
```
qwwdfsad: When I am going to incorporate long scanning, we could get rid of global queues and replace it with sparse `WorkerQueue` at uneven indices in `workers` with CAS-tryLock protection and push external tasks to these queues

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -579,17 +525,19 @@ internal class CoroutineScheduler(
                     "CPU = $cpuWorkers, " +
                     "blocking = $blockingWorkers, " +
                     "parked = $parkedWorkers, " +
-                    "retired = $retired, " +
+                    "retired = $dormant, " +
```
fvasco: missing refactoring?

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -216,70 +230,70 @@ internal class CoroutineScheduler(
...
+    private inline fun blockingTasks(state: Long): Int = (state and BLOCKING_MASK shr BLOCKING_SHIFT).toInt()
+    public inline fun availableCpuPermits(state: Long): Int = (state and CPU_PERMITS_MASK shr CPU_PERMITS_SHIFT).toInt()
 
     // Guarded by synchronization
     private inline fun incrementCreatedWorkers(): Int = createdWorkers(controlState.incrementAndGet())
     private inline fun decrementCreatedWorkers(): Int = createdWorkers(controlState.getAndDecrement())
 
-    private inline fun incrementBlockingWorkers() { controlState.addAndGet(1L shl BLOCKING_SHIFT) }
-    private inline fun decrementBlockingWorkers() { controlState.addAndGet(-(1L shl BLOCKING_SHIFT)) }
+    private inline fun incrementBlockingTasks() = controlState.addAndGet(1L shl BLOCKING_SHIFT)
```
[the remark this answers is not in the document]
qwwdfsad: Indeed. But it is a private API and it's much easier to have two inconsistent methods than two consistent, but with clumsy names

kotlinx-coroutines-core/jvm/test/scheduling/BlockingCoroutineDispatcherThreadLimitStressTest.kt
qwwdfsad: TODO: create an issue before merging

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: Nah, apparently it can be done properly only with access to `threadLocalRandomProbe`

elizarov:
Does not seem that we need `@Volatile` on `state`:

elizarov:
Note, a bug was introduced somewhere in shutdown sequence. It can hang on shutdown now: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStressWindows/2652563
```
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
	at java.lang.Object.wait(Native Method)
	at java.lang.Thread.join(Thread.java:1260)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.shutdown(CoroutineScheduler.kt:346)
	at kotlinx.coroutines.scheduling.ExperimentalCoroutineDispatcher.shutdown$kotlinx_coroutines_core(Dispatcher.kt:125)
	at kotlinx.coroutines.TestBase.shutdownPoolsAfterTest(TestBase.kt:173)
	at kotlinx.coroutines.TestBase.onCompletion(TestBase.kt:150)
	at kotlinx.coroutines.debug.DebugTestBase.tearDown(DebugTestBase.kt:29)
```

qwwdfsad:
I was avoiding adding long-scanning (dd30af0ea6b206feb887a428ae3528c6f798d8af) because it could have masked all the bugs on the rendezvous of parking/unparking. Ironically, it has exposed existing problems :) It seems like it is impossible to ensure strict liveness in loads like `BlockingCoroutineDispatcherMixedStealingStressTest.`. I can't say this is a "bug", but rather something we should be aware of before merge.
I will elaborate on details later, for now just will fix an execution that leads to hanging 

-- 1 core, 2 threads (one with CPU token), both ends scanning queues before pushing themselves to the stack.
-- External submitter submits CPU task to the global queue, can't find anyone to unpark, can't create a new thread, bails out
-- Thread without CPU token pushes itself to the stack
-- Thread with CPU token pushes itself to the stack
-- Thread with CPU token finds external CPU tasks and starts executing it
-- Thread without token parks 
-- External submitter submits blocking tasks, tries to compensate workers, but can't, pops the stack, founds the thread with CPU token (that is executing CPU task right now), successfully "unparks" it, bails out
-- CPU task never completes (due to the nature of the test), blocking task is kept in the queue, blocking worker is parked forever

elizarov:
@qwwdfsad 
....
> * External submitter submits blocking tasks, tries to compensate workers, but can't, pops the stack, founds the thread with CPU token (that is executing CPU task right now), successfully "unparks" it, bails out
Hm. If the thread is executing CPU task right now, then it _should not_ be unparked by `tryUnpark` as `worker.tryForbidTermination()` should return false if the thread was claimed before, because `terminationState.value = TERMINATION_ALLOWED` assignment is only done in the actual `worker.park()` method after CPU task finished executing. 

Combining both `terminationState` and `parkingState` to a single state-machine that is reset after parking should be able to fix this problem.

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -730,23 +705,17 @@ internal class CoroutineScheduler(
...
         private fun tryPark() {
             if (!inStack()) {
-                parkingState.value = PARKING_ALLOWED
                 parkedWorkersStackPush(this)
                 return
-
             }
             assert { localQueue.size == 0 }
-            // Failed to get a parking permit => we are not in the stack
             while (inStack()) {
```
elizarov: Performance nit: Before this line we've already checked that `inStack()` is true, so this loop can be replaced with 
`do { ... } while (inStack())`

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -730,23 +705,17 @@ internal class CoroutineScheduler(
...
                 if (isTerminated || state == WorkerState.TERMINATED) break
-                if (parkingState.value != PARKED && !parkingState.compareAndSet(PARKING_ALLOWED, PARKED)) {
-                    return
-                }
                 tryReleaseCpu(WorkerState.PARKING)
+                workerCtl.value = PARKED
+                if (!inStack()) return
```
elizarov: This `if (!inStack()) return` works as "park avoidance", but the window between the previous check for `inStack` and this check is too short to be practically working. I'd suggest removing this line to simplify code.

Issue #1657, opened 2019-11-12
Enforce exception transparency in SafeFlow


qwwdfsad:
We can enforce it by caching and checking any exception that was thrown by `downstream.emit`.
The trick is to have (almost) the same performance as before while giving up on tail-call.
For that, we can use a trick with "reusable continuation" as in [`TakeFlowCollector`](https://github.com/Kotlin/kotlinx.coroutines/blob/develop/benchmarks/src/jmh/kotlin/benchmarks/flow/TakeBenchmark.kt#L105).

Enforcing this invariant is crucial, otherwise, it may lead to really weird interactions, for example, consider the following pipeline:
```
val flow = flow {
    try {
        emit(1)
    } catch (e: Throwable) {
        emit(2)
    }
}

// Somewhere on consumer side
flow.collect {
    println("Processed '$it' and expecting no more elements")
    throw CancellationException()
}
```

`collect` lambda will be invoked twice even though it threw an exception on the first `emit`

PR #1658, opened 2019-11-12
Rethrow downstream exception during "onCompletion" emissions

### LINKED ISSUE #1654: flowOf("a", "b").onCompletion { emit("Done") }.first() returns "Done"
If I run this program:

```
suspend fun main() {
    flowOf("a", "b")
            .onCompletion { emit("Done") }
            .first()
            .also { println(it) }
}
```

I expect to see the output `a`, but the actual output is `Done`.

qwwdfsad:
* We cannot allow emitting elements when downstream exception occurred, otherwise it may lead to a weird side-effects when "collect" block (or any other terminal operator) has thrown an exception, but keeps receiving new values
    * Another solution may be to silently ignore emitted values, but it may lead to a postponed cancellation and surprising behaviour for users

Fixes #1654


qwwdfsad:
I expect to address this issue even more during fix for #1657

PR #1667, opened 2019-11-20
Support yield in immediate dispatchers

### LINKED ISSUE #1474: Android: await frame
It seems like #737 might have regressed, yielding on the immediate main dispatcher while on the main thread is currently a no-op as shown by this example:
```kotlin
        Log.d("Application", "Before")
        GlobalScope.launch(Dispatchers.Main.immediate) {
            Log.d("Application", "Before yield")
            yield()
            Log.d("Application", "After yield")
        }
        Log.d("Application", "After")
```

The example above is placed inside the `Application.onCreate` which is already running on the main thread.

The expected output would be:
```
Before
Before yield
After
After yield
```

But the actual output is:
```
Before
Before yield
After yield
After
```

elizarov:
Fixes #1474

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
@@ -78,8 +78,9 @@ public abstract class CoroutineDispatcher :
      * may leave the coroutines that use this dispatcher in the inconsistent and hard to debug state.
      *
      * **Note**: This method must not immediately call [block]. Doing so would result in [StackOverflowError]
-     * when [yield] is repeatedly called from a loop. However, an implementation can delegate this function
-     * to `dispatch` method of [Dispatchers.Unconfined], which is integrated with [yield] to avoid this problem.
+     * when [yield] is repeatedly called from a loop. However, an implementation that returns `false` from
+     * [isDispatchNeeded] can delegate this function to `dispatch` method of [Dispatchers.Unconfined], which is
+     * integrated with [yield] to avoid this problem.
      */
     public abstract fun dispatch(context: CoroutineContext, block: Runnable)
 
--- a/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
@@ -76,6 +76,10 @@ public abstract class CoroutineDispatcher :
      *
      * This method should generally be exception-safe. An exception thrown from this method
      * may leave the coroutines that use this dispatcher in the inconsistent and hard to debug state.
+     *
+     * **Note**: This method must not immediately call [block]. Doing so would result in [StackOverflowError]
+     * when [yield] is repeatedly called from a loop. However, an implementation can delegate this function
+     * to `dispatch` method of [Dispatchers.Unconfined], which is integrated with [yield] to avoid this problem.
      */
     public abstract fun dispatch(context: CoroutineContext, block: Runnable)
 
--- a/kotlinx-coroutines-core/common/src/Yield.kt
+++ b/kotlinx-coroutines-core/common/src/Yield.kt
@@ -20,23 +20,30 @@ import kotlin.coroutines.intrinsics.*
  *
  * If the coroutine dispatcher is [Unconfined][Dispatchers.Unconfined], this
  * functions suspends only when there are other unconfined coroutines working and forming an event-loop.
- * For other dispatchers, this function does not call [CoroutineDispatcher.isDispatchNeeded] and
- * always suspends to be resumed later. If there is no [CoroutineDispatcher] in the context, it does not suspend.
+ * For other dispatchers, this function calls [CoroutineDispatcher.dispatch] and
+ * always suspends to be resumed later regardless of the result of [CoroutineDispatcher.isDispatchNeeded].
+ * If there is no [CoroutineDispatcher] in the context, it does not suspend.
  */
 public suspend fun yield(): Unit = suspendCoroutineUninterceptedOrReturn sc@ { uCont ->
     val context = uCont.context
     context.checkCompletion()
     val cont = uCont.intercepted() as? DispatchedContinuation<Unit> ?: return@sc Unit
-    // This code detects the Unconfined dispatcher even if it was wrapped into another dispatcher
-    val yieldContext = YieldContext()
-    cont.dispatchYield(context + yieldContext, Unit)
-    // Special case for the unconfined dispatcher that can yield only in existing unconfined loop
-    if (yieldContext.dispatcherWasUnconfined) {
-        // Means that the Unconfined dispatcher got the call, but did not do anything.
-        // See also code of "Unconfined.dispatch" function.
-        return@sc if (cont.yieldUndispatched()) COROUTINE_SUSPENDED else Unit
+    if (cont.dispatcher.isDispatchNeeded(context)) {
+        // this is a regular dispatcher -- do simple dispatchYield
+        cont.dispatchYield(context, Unit)
+    } else {
+        // This is either an "immediate" dispatcher or the Unconfined dispatcher
+        // This code detects the Unconfined dispatcher even if it was wrapped into another dispatcher
+        val yieldContext = YieldContext()
+        cont.dispatchYield(context + yieldContext, Unit)
+        // Special case for the unconfined dispatcher that can yield only in existing unconfined loop
+        if (yieldContext.dispatcherWasUnconfined) {
+            // Means that the Unconfined dispatcher got the call, but did not do anything.
+            // See also code of "Unconfined.dispatch" function.
+            return@sc if (cont.yieldUndispatched()) COROUTINE_SUSPENDED else Unit
+        }
+        // Otherwise, it was some other dispatcher that successfully dispatched the coroutine
     }
-    // It was some other dispatcher that successfully dispatched the coroutine
     COROUTINE_SUSPENDED
 }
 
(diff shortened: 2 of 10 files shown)


kotlinx-coroutines-core/common/test/ImmediateYieldTest.kt
```diff
@@ -0,0 +1,35 @@
...
+package kotlinx.coroutines
+
+import kotlin.coroutines.*
+import kotlin.test.*
+
+class ImmediateYieldTest : TestBase() {
+
+    // See https://github.com/Kotlin/kotlinx.coroutines/issues/1474
+    @Test
+    fun testImmediateYield() = runTest {
```
qwwdfsad: It would be nice to have a more realistic test as well. 
E.g. like `HandlerDispatcherTest` and `JavaFxTest`

kotlinx-coroutines-core/common/src/Yield.kt
```diff
@@ -19,7 +19,9 @@ public suspend fun yield(): Unit = suspendCoroutineUninterceptedOrReturn sc@ { u
     val context = uCont.context
     context.checkCompletion()
     val cont = uCont.intercepted() as? DispatchedContinuation<Unit> ?: return@sc Unit
-    if (!cont.dispatcher.isDispatchNeeded(context)) {
+    // Special case for Unconfined dispatcher that does not support the concept of "dispatch" and where
+    // "dispatchYield" function cannot be used.
+    if (!cont.dispatcher.isDispatchSupported(context)) {
```
qwwdfsad: If it is internal and applicable only for `Unconfined` dispatcher and `isDispatchSupported` is internal anyway, maybe we should compare `cont.dispatcher` to it instead of adding one more public method?

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -79,6 +79,15 @@ public abstract class CoroutineDispatcher :
...
     public abstract fun dispatch(context: CoroutineContext, block: Runnable)
 
+    /**
+     * Reserved for [Dispatchers.Unconfined] that returns `false` here because its
+     * [dispatch] method throws [UnsupportedOperationException]. It is used in [yield] implementation.
+     *
+     * @suppress **This an internal API and should not be used from general code.**
+     */
+    @InternalCoroutinesApi
+    public open fun isDispatchSupported(context: CoroutineContext): Boolean = true
```
qwwdfsad: What's the point of `context` parameter?

kotlinx-coroutines-core/common/test/ImmediateYieldTest.kt
elizarov: Added tests for Android, JavaFx, and Swing

kotlinx-coroutines-core/common/src/Yield.kt
elizarov: Got rid of the method, added equality comparison to dispatcher. Also, fixed a problem where yielding from a wrapped dispatcher around Unconfined would crash.

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
elizarov: Irrelevant, got rid of the method.

kotlinx-coroutines-core/common/src/Yield.kt
fvasco: Alternatively, both methods can being replaced by a single one with enumeration: `NEEDED`, `SUPPORTED`, `UNAVAILABLE`.

elizarov: @fvasco  That would be a binary API change that would break 3rd party dispatcher implementations. It is easier to special-case `Unconfined` here since we don't really have used-cases at the table for user-defined dispatchers with unconfined-like behavior.

elizarov:
@JakeWharton Note, that if somebody creates a wrapper around `Unconfined` dispatcher or writes their own dispatcher like this:
```
object MyUnconfined : CoroutineDispatcher() {
    override fun isDispatchNeeded(context: CoroutineContext): Boolean = false
    override fun dispatch(context: CoroutineContext, block: Runnable) = block.run() // !!!
}
```

The `yield()` in such an `MyUnconfined` dispatcher implementation would never work. It will not be able to actually "yield". So, using `yield()` to work-around `java.lang.reflect.Proxy` checked exception problem is not 100% fool-proof. I wonder if we can figure out a solution that works all the time.

JakeWharton:
The first thing I reached for was `delay(1)`, but it made me sad.

elizarov:
> The first thing I reached for was `delay(1)`, but it made me sad.

The other potential problematic case would be coroutine running without a dispatcher at all (a truly unconfined coroutine). The existing and updated version of `yield` is implemented as "nop" in this case. It does not do anything. So, I see two conceptual ways to solve it:

1. Retrofit-like libraries that use `Proxy` can rely on _their own_ thread-pool to schedule `continuation.resumeWithException` to happen later in their own thread.

2. We can update implementation of `yield` in such a way that it ether use dispatcher-provided ability to yield thread or if there is no dispatcher or dispatcher is unconfined, then `yield()` would use our built-in `DefaultExecutor` to yield --- that same executor that is used as a fallback to implement `delay(...)` for dispatchers that do not support it natively.

elizarov:
@JakeWharton I gave a bit more thought to the second idea. It is not really good. The change in **this** PR (support `yield` for `immediate`) is good. But if we provide a hard guarantee that "yield always suspends" then it means that in some simple situations without a dispatcher execution after `yield()` must resume in another thread. This would be surprising. 

However, it is not surprising if some code that uses a remote-call library like retrofit without a dispatcher gets resumed in another thread. So, the first option to work around `Proxy` exception handling seems better to me. I'd suggest to copy and use the following function instead of `yield()` as an exception work-around:

```
suspend fun ensureSuspension() = suspendCoroutine<Unit> {
    Dispatchers.Default.dispatch(it.context, Runnable { it.resume(Unit) })
}
```

You'd use just like you use `yield()` now inside your `Proxy`:

```
try { 
   // do something that might throw exception w/o suspending
} catch (e: Throwable) {
   ensureSuspension()
   throw e // rethrow exception after guaranteed suspension 
}
```

kotlinx-coroutines-core/common/src/Yield.kt
```diff
@@ -19,9 +19,8 @@ public suspend fun yield(): Unit = suspendCoroutineUninterceptedOrReturn sc@ { u
     val context = uCont.context
     context.checkCompletion()
     val cont = uCont.intercepted() as? DispatchedContinuation<Unit> ?: return@sc Unit
-    if (!cont.dispatcher.isDispatchNeeded(context)) {
-        return@sc if (cont.yieldUndispatched()) COROUTINE_SUSPENDED else Unit
-    }
+    // Special case for the unconfined dispatcher that can yield only in existing unconfined loop
```
qwwdfsad: When squashing, please fix the type in KDoc: "but also check" -> "but also checks".

elizarov: I've squashed and totally updated KDoc for `yield` to better document its actual behavior.

elizarov:
I've found a problem with this change. We try to dispatch to dispatchers that return `false` from `isDispatchNeed` to support immediate dispatcher. Now consider a user that had accidentally wrapped `Unconfined` dispatcher for whatever reason. Maybe they have some tracking framework and need their own `CorouteDispatcher` wrapper or whatever. With such a wrapper on top of `Unconfined` calling `yield` from a loop would cause `StackOverflowError`. I've added a test to demonstrate it. 

As a fix, `yield` now adds a special coroutine context element that is detected by `Unconfined` dispatcher so they still work correctly in tandem even when `Unconfined` dispatcher is wrapped. I've also updated docs for `CoroutineDispatcher.dispatch`. 

The downside of this change is that a few extra objects are created on each `yield` call. I have another implementation strategy based on `ThreadLocal` which I've tried and it also works, but I've found the resulting code to be somewhat more cryptic. 

Please, review.

kotlinx-coroutines-core/common/src/AbstractCoroutine.kt
```diff
@@ -133,7 +133,7 @@ public abstract class AbstractCoroutine<in T>(
      *
      * * [DEFAULT] uses [startCoroutineCancellable].
      * * [ATOMIC] uses [startCoroutine].
-     * * [UNDISPATCHED] uses [startCoroutineUndispatched].
+     * * [UNCONFINED] uses [startCoroutineUndispatched].
```
qwwdfsad: Accidental change?

kotlinx-coroutines-core/common/src/Yield.kt
```diff
@@ -8,21 +8,35 @@ import kotlin.coroutines.*
...
+ * For other dispatchers, this function does not call [CoroutineDispatcher.isDispatchNeeded] and
+ * always suspends to be resumed later. If there is no [CoroutineDispatcher] in the context, it does not suspend.
  */
 public suspend fun yield(): Unit = suspendCoroutineUninterceptedOrReturn sc@ { uCont ->
     val context = uCont.context
     context.checkCompletion()
     val cont = uCont.intercepted() as? DispatchedContinuation<Unit> ?: return@sc Unit
-    if (!cont.dispatcher.isDispatchNeeded(context)) {
+    // This code detects the Unconfined dispatcher even if it was wrapped into another dispatcher
+    val yieldContext = YieldContext()
```
qwwdfsad: YieldContext can be conditionally allocated based on `isDispatchNeeded`

kotlinx-coroutines-core/common/src/AbstractCoroutine.kt
elizarov: Good catch!

PR #1672, opened 2019-11-25
Improve documentation

### LINKED ISSUE #1634: Docs: Coroutine basics compiler error "Unresolved reference: thread"
On https://kotlinlang.org/docs/reference/coroutines/basics.html it states for the given example code

```
import kotlinx.coroutines.*

fun main() {
    GlobalScope.launch { // launch a new coroutine in background and continue
        delay(1000L) // non-blocking delay for 1 second (default time unit is ms)
        println("World!") // print after delay
    }
    println("Hello,") // main thread continues while coroutine is delayed
    Thread.sleep(2000L) // block main thread for 2 seconds to keep JVM alive
}
```

>You can achieve the same result replacing GlobalScope.launch { ... } with thread { ... } and delay(...) with Thread.sleep(...). Try it.

Changing the example per the instructions

```
import kotlinx.coroutines.*

fun main() {
    thread { // launch a new coroutine in background and continue
        Thread.sleep(1000L) // non-blocking delay for 1 second (default time unit is ms)
        println("World!") // print after delay
    }
    println("Hello,") // main thread continues while coroutine is delayed
    Thread.sleep(2000L) // block main thread for 2 seconds to keep JVM alive
}
```

and running gives the error

>Unresolved reference: thread

Am I misunderstanding the instructions?

### LINKED ISSUE #1650: Android: Main.immediate causes a dispatch using Unconfined inside
`Dispatchers.Main.immediate` forces child coroutine with `Dispatchers.Unconfined` interceptor to dispatch.

Example code
```
GlobalScope.launch(Dispatchers.Main.immediate) {
    GlobalScope.launch(Dispatchers.Unconfined) {
        println("$coroutineContext: before")
    }
    println("$coroutineContext: after")
}
```
This code prints
```
[CoroutineId(1), "coroutine#1":StandaloneCoroutine{Active}@229841e, Main [immediate]]: after
[CoroutineId(2), "coroutine#2":StandaloneCoroutine{Active}@f2294ff, Unconfined]: before
```
But expected would be backwards.

I can force not to dispatch if I use `start = CoroutineStart.UNDISPATCHED`. Then I get the expected result
```
[CoroutineId(4), "coroutine#4":StandaloneCoroutine{Active}@229841e, Unconfined]: before
[CoroutineId(3), "coroutine#3":StandaloneCoroutine{Active}@f2294ff, Main [immediate]]: after
```

But shouldn't `GlobalScope.launch(Dispatchers.Unconfined)` and `GlobalScope.launch(start = CoroutineStart.UNDISPATCHED, context = Dispatchers.Unconfined)` be interchangable?

### LINKED ISSUE #1651: [Question] Which dispatchers does CoroutineExceptionHandler use?
I'm new to coroutines, and I'm curious which `Dispatchers` `CoroutineExceptionHandler` uses. I'd like to perform UI work, and display an Android `SnackBar` from the exception handler context. But, I want to confirm or not that the safety of doing so as I can't use `withContext(Dispatchers.UI)` to explicitly state the context. Thanks in advance!

qwwdfsad:
* Clarifications of CoroutineExceptionHandler execution
    * Clarifications of MainCoroutineDispatcher.immediate
    * Outdated documentation (pointed out by Google AndoidX team) for isDispatchNeeded is rewritten

Fixes #1650
Fixes #1651
Fixes #1634

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
@@ -27,46 +27,30 @@ import kotlin.coroutines.*
  *
  * This class ensures that debugging facilities in [newCoroutineContext] function work properly.
  */
-public abstract class CoroutineDispatcher :
-    AbstractCoroutineContextElement(ContinuationInterceptor), ContinuationInterceptor {
+public abstract class CoroutineDispatcher
+    : AbstractCoroutineContextElement(ContinuationInterceptor), ContinuationInterceptor {
+
     /**
-     * Returns `true` if the execution shall be dispatched onto another thread.
+     * Returns `true` if the execution of the coroutine should be performed with [dispatch] method.
      * The default behavior for most dispatchers is to return `true`.
      *
-     * This method should never be used from general code, it is used only by `kotlinx.coroutines`
-     * internals and its contract with the rest of the API is an implementation detail.
+     * If this method returns `false`, the coroutine is resumed immediately in the current thread,
+     * potentially forming an event-loop to prevent stack overflows.
+     * The event loop is an advanced topic and its implications can be found in [Dispatchers.Unconfined] documentation.
      *
-     * UI dispatchers _should not_ override `isDispatchNeeded`, but leave the default implementation that
-     * returns `true`. To understand the rationale beyond this recommendation, consider the following code:
+     * A dispatcher can override this method to provide a performance optimization and avoid paying a cost of an unnecessary dispatch.
+     * E.g. [MainCoroutineDispatcher.immediate] checks whether we are already in the required UI thread in this method and avoids
+     * an additional dispatch when it is not required. But this method should not return `false` by default, because
+     * it may expose unexpected behaviour (e.g. with interleaved event-loops) and unexpected order of events.
      *
-     * ```kotlin
-     * fun asyncUpdateUI() = async(Dispatchers.Main) {
-     *     // do something here that updates something in UI
-     * }
-     * ```
-     *
-     * When you invoke `asyncUpdateUI` in some background thread, it immediately continues to the next
-     * line, while the UI update happens asynchronously in the UI thread. However, if you invoke
-     * it in the UI thread itself, it will update the UI _synchronously_ if your `isDispatchNeeded` is
-     * overridden with a thread check. Checking if we are already in the UI thread seems more
-     * efficient (and it might indeed save a few CPU cycles), but this subtle and context-sensitive
-     * difference in behavior makes the resulting async code harder to debug.
-     *
-     * Basically, the choice here is between the "JS-style" asynchronous approach (async actions
-     * are always postponed to be executed later in the event dispatch thread) and "C#-style" approach
-     * (async actions are executed in the invoker thread until the first suspension point).
-     * While the C# approach seems to be more efficient, it ends up with recommendations like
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -27,46 +27,30 @@ import kotlin.coroutines.*
  *
  * This class ensures that debugging facilities in [newCoroutineContext] function work properly.
  */
-public abstract class CoroutineDispatcher :
-    AbstractCoroutineContextElement(ContinuationInterceptor), ContinuationInterceptor {
+public abstract class CoroutineDispatcher
+    : AbstractCoroutineContextElement(ContinuationInterceptor), ContinuationInterceptor {
```
elizarov: Style (quote): "For classes with a long supertype list, _put a line break after the colon_ "

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -27,46 +27,30 @@ import kotlin.coroutines.*
...
-     * Basically, the choice here is between the "JS-style" asynchronous approach (async actions
-     * are always postponed to be executed later in the event dispatch thread) and "C#-style" approach
-     * (async actions are executed in the invoker thread until the first suspension point).
-     * While the C# approach seems to be more efficient, it ends up with recommendations like
-     * "use `yield` if you need to ....". This is error-prone. The JS-style approach is more consistent
-     * and does not require programmers to think about whether they need to yield or not.
+     * A dispatcher can override this method to provide a performance optimization and avoid paying a cost of an unnecessary dispatch.
+     * E.g. [MainCoroutineDispatcher.immediate] checks whether we are already in the required UI thread in this method and avoids
+     * an additional dispatch when it is not required. But this method should not return `false` by default, because
+     * it may expose unexpected behaviour (e.g. with interleaved event-loops) and unexpected order of events.
```
elizarov: This documentation now lacks rationale on why the default (non-immeidate) behavior works the way it is. There is no other place where it is currently documented. I'd suggest to either leave it here or move it somewhere.

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -27,46 +27,30 @@ import kotlin.coroutines.*
...
-     * whenever it is needed for efficiency.
+     * Coroutine builders like [launch][CoroutineScope.launch] and [async][CoroutineScope.async] accept an optional [CoroutineStart]
+     * parameter that allows one to optionally choose the [undispatched][CoroutineStart.UNDISPATCHED] behavior to start coroutine immediately,
+     * but to be resumed only in the provided dispatcher.
      *
      * This method should generally be exception-safe. An exception thrown from this method
      * may leave the coroutines that use this dispatcher in the inconsistent and hard to debug state.
      *
-     * **Note: This is an experimental api.** Execution semantics of coroutines may change in the future when this function returns `false`.
+     * **This is an experimental api.** Execution semantics of coroutines may change in the future when this function returns `false`.
```
elizarov: Maybe it's time to drop experimental and leave it the way it is?

qwwdfsad: On the one hand, we still change event order in event loops. On the other hand, `Unconfined` and `immediate` are stable and documentation states "order of events is an implementation detail that can be changed" . Let's drop experimental status

PR #1674, opened 2019-11-26
ConflatedChannel with synchronized access to the one-element buffer

SokolovaMaria:


kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -15,80 +12,116 @@ import kotlinx.coroutines.internal.*
...
+            if (!select.trySelect())
+                return ALREADY_SELECTED
+            result = buffer
+            buffer = null
+            size = 0
+        }
+        return result
+    }
+
+    protected override fun onCancelIdempotent(wasClosed: Boolean) {
```
elizarov: It should call `super.onCancelIdempotent`.

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -15,80 +12,116 @@ import kotlinx.coroutines.internal.*
...
+        if (wasClosed) {
+            lock.withLock {
+                buffer = null
+                size = 0
             }
         }
     }
-}
+}
```
elizarov: For easier debugging, please also override `bufferDebugString` here.

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -15,80 +12,116 @@ import kotlinx.coroutines.internal.*
...
  */
 internal open class ConflatedChannel<E> : AbstractChannel<E>() {
-    protected final override val isBufferAlwaysEmpty: Boolean get() = true
-    protected final override val isBufferEmpty: Boolean get() = true
-    protected final override val isBufferAlwaysFull: Boolean get() = false
-    protected final override val isBufferFull: Boolean get() = false
+    protected final override val isBufferAlwaysEmpty: Boolean = false
+    protected final override val isBufferEmpty: Boolean get() = lock.withLock { size == 0 }
+    protected final override val isBufferAlwaysFull: Boolean = false
+    protected final override val isBufferFull: Boolean get() = lock.withLock { size == 1 }
```
elizarov: This override of `isBufferFull` should not be needed. Conceptually, a conflated channel is never full, since you can also successfully send to it without suspension. Correspondingly, the override of `isFullImpl` should not be needed either.

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -15,80 +13,118 @@ import kotlinx.coroutines.internal.*
...
- *
- * This implementation is fully lock-free.
  */
 internal open class ConflatedChannel<E> : AbstractChannel<E>() {
-    protected final override val isBufferAlwaysEmpty: Boolean get() = true
-    protected final override val isBufferEmpty: Boolean get() = true
-    protected final override val isBufferAlwaysFull: Boolean get() = false
+    protected final override val isBufferAlwaysEmpty: Boolean = false
+    protected final override val isBufferEmpty: Boolean get() = lock.withLock { value === EMPTY }
+    protected final override val isBufferAlwaysFull: Boolean = false
```
elizarov: Both `isBufferAlwaysEmpty` and `isBufferAlwaysFull` should be written with `get() = false` as they were before to avoid an extra field that is not needed here.

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -15,80 +13,118 @@ import kotlinx.coroutines.internal.*
...
  * This channel is created by `Channel(Channel.CONFLATED)` factory function invocation.
- *
- * This implementation is fully lock-free.
  */
 internal open class ConflatedChannel<E> : AbstractChannel<E>() {
-    protected final override val isBufferAlwaysEmpty: Boolean get() = true
-    protected final override val isBufferEmpty: Boolean get() = true
-    protected final override val isBufferAlwaysFull: Boolean get() = false
+    protected final override val isBufferAlwaysEmpty: Boolean = false
+    protected final override val isBufferEmpty: Boolean get() = lock.withLock { value === EMPTY }
```
elizarov: The lock on `isBufferEmpty` should not be taken here, but in overrides for `isEmpty` and for `enqueueReceiveInternal` instead (see how `ArrayChannel` does it). This way, `isEmpty` will be linearizable, too.

kotlinx-coroutines-core/jvm/test/linearizability/ChannelsLCStressTest.kt
```diff
@@ -102,13 +100,13 @@ abstract class ChannelLCStressTestBase(private val c: Channel<Int>, private val
     fun close(@Param(name = "closeToken") token: Int): Boolean = c.close(NumberedCancellationException(token))
 
     // TODO: this operation should be (and can be!) linearizable, but is not
-    // @Operation
+    //@Operation
```
elizarov: No need to change this line (it looked better with space between `//` and `@`).

elizarov:
Stress-test on Windows found non-linearizable execution: 
```
ava.lang.AssertionError: Invalid interleaving found:
= Invalid execution results: =
Parallel part:
| *send(2):             SUSPENDED + CANCELLED | close(2):             true      |
| isClosedForReceive(): false                 | isClosedForReceive(): false     |
|                                             | poll():               Closed(2) |
java.lang.AssertionError: Invalid interleaving found:
= Invalid execution results: =
Parallel part:
| *send(2):             SUSPENDED + CANCELLED | close(2):             true      |
| isClosedForReceive(): false                 | isClosedForReceive(): false     |
|                                             | poll():               Closed(2) |
```
See https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStressWindows/2746407?mode=builds

PR #1685, opened 2019-12-02
Fix MDCContext to keep changed MDCContextMap

qbx2:
Hello. While I don't understand why `MDCContextMap` discards updates of MDC context inside of coroutine, I found this change will work well to keep MDC changes. My only concern is that `contextMap` could conflict when race condition occurs. However, I think it is up to the developer to avoid the issue. Please review if this change is ok.


qwwdfsad:
It is not going to work because `MDCContext` is shared

PR #1689, opened 2019-12-05
To string

qwwdfsad:


kotlinx-coroutines-core/common/src/internal/Scopes.kt
```diff
@@ -35,4 +35,6 @@ internal open class ScopeCoroutine<in T>(
 
 internal class ContextScope(context: CoroutineContext) : CoroutineScope {
     override val coroutineContext: CoroutineContext = context
+    // CoroutineScope is used intentionally for user-friendly representation
+    override fun toString(): String = "CoroutineScope(coroutineContext = $coroutineContext)"
```
elizarov: Nit: Extra space around `=`. We don't generally have them in our toString's nor there's any space in a default toString for data classes.

PR #1577, opened 2019-09-27
Restore thread context elements when directly resuming to parent

elizarov:
This fix solves the problem of restoring thread-context when returning to another context in undispatched way. It impacts suspend/resume performance of coroutines that use ThreadContextElement since we have to walk up the coroutine completion stack in search for
parent UndispatchedCoroutine. However, there is a fast-path to ensure that there is no performance impact in cases when ThreadContextElement is not used by a coroutine.

Fixes #985

Issue #985, opened 2019-02-15
Possible bug with `ThreadLocal.asContextElement`


alamothe:
Hello,
I'm observing a strange behavior with the following code:
```kotlin
	val t: ThreadLocal<String> = ThreadLocal.withInitial { "a" }

	@Test
	fun testContext1() = runBlocking {
		println(t.get())
		withContext(coroutineContext.plus(t.asContextElement("b"))) {
			println(t.get())
		}
		println(t.get())
	}

	@Test
	fun testContext2() = runBlocking {
		println(t.get())
		withContext(coroutineContext.plus(t.asContextElement("b"))) {
			println(t.get())
			delay(Random.nextLong(100))
			println(t.get())
		}
		println(t.get())
	}

	@Test
	fun testContext3() = runBlocking {
		println(t.get())
		async(coroutineContext.plus(t.asContextElement("b"))) {
			println(t.get())
			delay(Random.nextLong(100))
			println(t.get())
		}.await()
		println(t.get())
	}
```

The result is:

- `testContext1` prints `aba` - **expected**
- `testContext2` prints `abbb` - **It looks like the value is not restored for some reason. Is this a bug? What's the explanation?**
- `testContext3` prints `abba` - **expected**

Please help!

I'm using `org.jetbrains.kotlinx:kotlinx-coroutines-core:1.1.1`

elizarov:
It is not exactly a bug, but a side-effect of the way `asContextElement` is implemented. See, your `main` coroutine (in `runBlocking`) does not have t's context element, so this is interpreted as "main coroutine do not care about this thread local". So then, depending on specific ways you get back to the main tread, the value of `t` can be either restored or not restored for the main coroutine.

If you write `runBlocking(t.asContextElement("a")) { ... }` in all your tests, then they all are going to run as expected. 

Also note, then when you use `withContext` function you don't have to write `withContext(coroutineContext.plus(other)) { ... }`, because that merging of current context with the other context is exactly what `withContext` does (see its [docs](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-context.html)), so you can simply write `withContext(other) { ... }`.

Does it help?

alamothe:
Hey @elizarov thank you for your reply & the tip for `withContext`.

I changed the test like this:
```
	@Test
	fun testContext2() {
		println(t.get())
		runBlocking {
			println(t.get())
[...]

PR #1696, opened 2019-12-10
Add ability to inject context elements into GlobalScope

frost13it:
See motivation in #985.

To work reliably, `ThreadContextElement`s need to be contained in a context of each coroutine that uses them, until merge of #1577.

This PR implements an alternative [solution]( https://github.com/Kotlin/kotlinx.coroutines/issues/985#issuecomment-535432215) proposed by @elizarov: to give libraries an ability to inject elements into the context of `GlobalScope`.

Note that it does not solve the [issue](https://github.com/Kotlin/kotlinx.coroutines/issues/985#issuecomment-534929877) with `suspend fun main`.

### DIFF
--- a/integration/kotlinx-coroutines-slf4j/resources/META-INF/services/kotlinx.coroutines.InitialContextProvider
+++ b/integration/kotlinx-coroutines-slf4j/resources/META-INF/services/kotlinx.coroutines.InitialContextProvider
@@ -1,0 +1,1 @@
+kotlinx.coroutines.slf4j.MdcInitialContextProvider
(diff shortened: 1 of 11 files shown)


elizarov:
Unfortunately, solutions based on `ServiceLoader` (even fast implementation of one) are known to be problematic in Android ecosystem. See the recent hacks that were introduced to make sure that `Main` dispatcher impl if loaded without service loader on Android. We'll have the same problem with `GlobalScope`.

If the problem we are solving only relevant for SLFJ4 mapped diagnostic context? If so, could be just provide MDC-specific `EmptyGlobalScope` in that library only?

frost13it:
> If the problem we are solving only relevant for SLFJ4 mapped diagnostic context? If so, could be just provide MDC-specific EmptyGlobalScope in that library only?

@elizarov, my use case explained in https://github.com/Kotlin/kotlinx.coroutines/issues/985#issuecomment-535171594, so hard-coding MDC support this way won't be suitable for me.
Is there a chance that #1577 will be merged eventually or it will be possible to access current coroutine's context via a `ThreadLocal`?

elizarov:
We'll get back to #1577. There are chances we can make it work.

frost13it:
It would be great!

PR #1716, opened 2019-12-16
shareIn and cache operators

### LINKED ISSUE #1261: Flow.share operator

RBusarow:
Fixes #1261


kotlinx-coroutines-core/common/src/flow/internal/SharedFlow.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    private fun Flow<T>.cacheIfNeeded(): Flow<T> = if (cacheHistory > 0) {
+        onEach { value ->
+            // While flowing, also record all values in the cache.
+            cache.add(value)
+        }
+    } else this
+
+    private fun reset() {
+        cache = CircularArray(cacheHistory)
```
LouisCAD: Existing channel is not closed, which can leak the shared flow each time `refCount` reaches `0`.
This function should close `lazyChannelRef.value` before assigning a new value to the `lazyChannelRef` variable.

RBusarow: Thanks!  

Believe it or not I removed that just before making the PR, thinking "well no one's collecting it, so what could leak?".  Except of course the leak can go two ways.

elizarov:
In light of #2069 I'm closing this one. Thanks!

PR #1717, opened 2019-12-17
ReceiveChannel.broadcast shall start lazy coroutine when closed

### LINKED ISSUE #1713: Broadcast close with CoroutineStart.DEFAULT produces an unexpected exception
Kotlin: 1.3.61
Coroutines: 1.3.3

```kotlin
runBlocking {
    val a = produce { 
        while (true) send(1) 
    }
    val b = a.broadcast(start = CoroutineStart.DEFAULT)
    b.close()
}
```

Problem: `send(1)` throws an exception after `b.close()` is called.
Expected behavior: `send(1)` should suspend instead of throwing an exception.

elizarov:
Fixes #1713

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Broadcast.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Broadcast.kt
@@ -10,9 +10,11 @@ import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
 import kotlinx.coroutines.intrinsics.*
 import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
+import kotlin.native.concurrent.*
 
 /**
  * Broadcasts all elements of the channel.
+ * This function [consumes][ReceiveChannel.consume] all elements of the original [ReceiveChannel].
  *
  * The kind of the resulting channel depends on the specified [capacity] parameter:
  * when `capacity` is positive (1 by default), but less than [UNLIMITED] -- uses `ArrayBroadcastChannel` with a buffer of given capacity,
@@ -20,17 +22,37 @@ import kotlin.coroutines.intrinsics.*
  *   Note that resulting channel behaves like [ConflatedBroadcastChannel] but is not an instance of [ConflatedBroadcastChannel].
  *   otherwise -- throws [IllegalArgumentException].
  *
+ * ### Cancelling broadcast
+ *
+ * **To stop broadcasting from the underlying channel call [cancel][BroadcastChannel.cancel] on the result.**
+ *
+ * Do not use [close][BroadcastChannel.close] on the resulting channel.
+ * It causes eventual failure of the broadcast coroutine and cancellation of the underlying channel, too,
+ * but it is not as prompt.
+ *
+ * ### Obsolete
+ *
+ * This function has an inappropriate result type of [BroadcastChannel] which provides
+ * [send][BroadcastChannel.send] and [close][BroadcastChannel.close] operations that interfere with
+ * the broadcasting coroutine in hard-to-specify ways. It will be replaced with
+ * sharing operators on [Flow][kotlinx.coroutines.flow.Flow].
+ *
  * @param start coroutine start option. The default value is [CoroutineStart.LAZY].
  */
+@ObsoleteCoroutinesApi // since version 1.4.0
 fun <E> ReceiveChannel<E>.broadcast(
     capacity: Int = 1,
     start: CoroutineStart = CoroutineStart.LAZY
-): BroadcastChannel<E> =
-    GlobalScope.broadcast(Dispatchers.Unconfined, capacity = capacity, start = start, onCompletion = consumes()) {
+): BroadcastChannel<E> {
+    val scope = GlobalScope + Dispatchers.Unconfined + CoroutineExceptionHandler { _, _ -> }
+    // We can run this coroutine in the context that ignores all exceptions, because of `onCompletion = consume()`
+    // which passes all exceptions upstream to the source ReceiveChannel
+    return scope.broadcast(capacity = capacity, start = start, onCompletion = consumes()) {
         for (e in this@broadcast) {
             send(e)
         }
     }
+}
 
 /**
  * Launches new coroutine to produce a stream of values by sending them to a broadcast channel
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/common/src/channels/Broadcast.kt
```diff
@@ -28,7 +29,12 @@ fun <E> ReceiveChannel<E>.broadcast(
 ): BroadcastChannel<E> =
     GlobalScope.broadcast(Dispatchers.Unconfined, capacity = capacity, start = start, onCompletion = consumes()) {
         for (e in this@broadcast) {
-            send(e)
+            try {
+                send(e)
+            } catch (e: ClosedSendChannelException) {
```
qwwdfsad: Please add a test that shows the necessity of this `catch` block. It is not clear it is really required

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
```diff
@@ -119,6 +125,13 @@ private open class BroadcastCoroutine<E>(
         val processed = _channel.close(cause)
         if (!processed && !handled) handleCoroutineException(context, cause)
     }
+
+    // The BroadcastChannel could be also closed
+    override fun close(cause: Throwable?): Boolean {
```
qwwdfsad: This change is not enough. 
Now if you'll close `BroadcastChannel` with exception different from `CancellationException`, it will be reported to global exception handler, because `onCancelled` will try to close _channel second time

elizarov:
I've totally reworked and force-pushed this update.

* Documentation on broadcast operators is added that explains that the resulting BroadcastChannel shall be cancelled if it is not needed anymore.
* More tests added for various broadcast cancel/close cases.
* The only functional change is that closing a broadcast channel for lazy coroutine shall start the corresponding coroutine to give it a chance to promptly fail.
* Mark broadcast operators as obsolete. To be replaced with sharing operators on flows (see #1716).

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
```diff
@@ -63,12 +85,28 @@ fun <E> ReceiveChannel<E>.broadcast(
...
+ * the broadcasting coroutine in hard-to-specify ways. It will be replaced with
+ * sharing operators on [Flow][kotlinx.coroutines.flow.Flow].
+ *
  * @param context additional to [CoroutineScope.coroutineContext] context of the coroutine.
  * @param capacity capacity of the channel's buffer (1 by default).
  * @param start coroutine start option. The default value is [CoroutineStart.LAZY].
  * @param onCompletion optional completion handler for the producer coroutine (see [Job.invokeOnCompletion]).
  * @param block the coroutine code.
  */
+@ObsoleteCoroutinesApi // since version 1.4.0
```
qwwdfsad: `broadcastIn` should be marked obsolete as well then?
Though it's quite popular nowadays, so it would be nice to provide a decent explanation or not to mark it as obsolete at all

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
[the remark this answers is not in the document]
elizarov: Ok. It should merge cleanly with date changes in other branches.

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
elizarov: Indeed.

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
```diff
@@ -107,8 +145,9 @@ private open class BroadcastCoroutine<E>(
     }
 
     override fun cancelInternal(cause: Throwable) {
-        _channel.cancel(cause.toCancellationException()) // cancel the channel
-        cancelCoroutine(cause) // cancel the job
+        val exception = cause.toCancellationException()
+        _channel.cancel(exception) // cancel the channel
+        cancelCoroutine(exception) // cancel the job
```
qwwdfsad: Could you please elaborate why did you change `cancelCoroutine(cause)` to `cancelCoroutine(cause.toCancellationException())`?

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
```diff
@@ -10,27 +10,49 @@ import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
...
+ * ### Future replacement
+ *
+ * This function has an inappropriate result type of [BroadcastChannel] which provides
+ * [send][BroadcastChannel.send] and [close][BroadcastChannel.close] operations that interfere with
+ * the broadcasting coroutine in hard-to-specify ways. It will be replaced with
+ * sharing operators on [Flow][kotlinx.coroutines.flow.Flow] in the future.
+ *
  * @param start coroutine start option. The default value is [CoroutineStart.LAZY].
  */
+//@ObsoleteCoroutinesApi // since version 1.4.0
```
qwwdfsad: Why have you decided to leave `CoroutineScope.broadcastIn` obsolete, but not this one?

elizarov: I've changed my mind but forgot about the other version. Fixed. None of them will be obsolete.

kotlinx-coroutines-core/common/src/channels/Broadcast.kt
elizarov: I've changed it for consistency with `ChannelCoroutine`, so that `cancelInternal` in both `BroadcastCoroutine` and `ChannelCoroutine` are the same (it is easier to compare them side-by-side this way). 

It should not really matter since the only method that can call `cancelInternal` with non-CancellationException cause is `JobSupport.cancel(cause: Throwable?)` which has been deprecated & hidden long time ago.

PR #1728, opened 2019-12-20
Do not request additional worker from 'yield' calls and during post-e…

### LINKED ISSUE #1704: Scheduler: Avoid unnecessary work signalling in LimitingDispatcher
To avoid liveness problems, at the end of each "limited" we double-check the queue and dispatch any new tasks from it.
At this point, a scheduler is not aware of the fact that thread it effectively free and issues one more unnecessary thread to help.

### LINKED ISSUE #1706: Scheduler: do not request additional thread for yield calls

qwwdfsad:
…xecution phase in LimitingDispatcher

Fixes #1704
Fixes #1706

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -377,8 +377,8 @@ internal class CoroutineScheduler(
      * If `true`, then  the task will be dispatched in a FIFO manner and no additional workers will be requested,
      * but only if the current thread is a corresponding worker thread.
      * Note that caller cannot be ensured that it is being executed on worker thread for the following reasons:
-     *   -- [CoroutineStart.UNDISPATCHED]
-     *   -- Concurrent [close] that effectively shutdowns the worker thread
+     *   * [CoroutineStart.UNDISPATCHED]
+     *   * Concurrent [close] that effectively shutdowns the worker thread
      */
     fun dispatch(block: Runnable, taskContext: TaskContext = NonBlockingContext, tailDispatch: Boolean = false) {
         trackTask() // this is needed for virtual time support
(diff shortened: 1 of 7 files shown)


qwwdfsad:
The biggest impact here is reducing CPU usage of `ActorSelectorManager` in Ktor in idle state by a factor of two

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -372,25 +372,34 @@ internal class CoroutineScheduler(
      * Dispatches execution of a runnable [block] with a hint to a scheduler whether
      * this [block] may execute blocking operations (IO, system calls, locking primitives etc.)
      *
-     * @param taskContext concurrency context of given [block]
-     * @param fair whether the task should be dispatched fairly (strict FIFO) or not (semi-FIFO)
+     * [taskContext] -- concurrency context of given [block].
+     * [tailDispatch] -- whether this [dispatch] call is the last action the (presumably) worker thread does in its current task.
```
zach-klippenstein: Just curious, why move away from using param tags?

elizarov: Please, replace `--` with `*` for a proper markdown formatting of bullet list. It does not render properly in the current shape.

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -481,19 +491,19 @@ internal class CoroutineScheduler(
      * Returns `null` if task was successfully added or an instance of the
      * task that was not added or replaced (thus should be added to global queue).
      */
-    private fun submitToLocalQueue(task: Task, fair: Boolean): Task? {
-        val worker = currentWorker() ?: return task
+    private fun Worker?.submitToLocalQueue(task: Task, tailDispatch: Boolean): Task? {
+        if (this === null) return task
```
elizarov: Style nit: `this == null`.

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: Lines are shorter :) 
For non-public API I personally find it more readable

elizarov: 👍 It also gives more freedom in the layout of the doc. You can have parameters documented in the course of the prose.

PR #1730, opened 2019-12-23
Ensure that flowOn does not resume downstream after cancellation

### LINKED ISSUE #1265: Flow.collects receives items after it was cancelled on a single thread
Could anyone explain to me why this code throws an exception (on Android)? The `collect` and `cancel` functions are both called on the main thread. I tried this with both version 1.2.1 and 1.3.0-M1
```kotlin
var cancelled = false
val job = GlobalScope.launch(Dispatchers.Main) {
    val flow = flow {
        while (true) {
            emit(Unit)
        }
    }
    flow.flowOn(Dispatchers.IO)
        .collect {
            if (cancelled) { // main thread
                throw IllegalStateException() // this exception is thrown
            }
        }
}

GlobalScope.launch(Dispatchers.Main) {
    delay(1000)
    job.cancel() // main thread
    cancelled = true // set to true after the Flow was cancelled.
}
```
[Here is an Android sample project](https://github.com/Kotlin/kotlinx.coroutines/files/3271166/KotlinFlowExample.zip), just run the app and it will crash.

I'm not sure if this is intended behaviour or a bug. I need to make sure all Flows are cancelled in the Android `onDestroy` so they do not try to update the view if it is already gone. So if this is not a bug I am not sure how I would need to handle this otherwise.

There is a long discussion about this with more code examples on Slack: https://kotlinlang.slack.com/archives/C1CFAFJSK/p1559908916083600

We had some trouble trying to reproduce this in a unit test (it did not throw the exception), so that's why I attached an Android project.

elizarov:
This is a problematic for Android when Main dispatcher is cancelled on destroyed activity. Atomic nature of the underlying channels is designed to prevent loss of elements, which is really not an issue for a typical application, but creates problem when used with channels.

This change introduces an optional `atomic` parameter to an internal `ReceiveChannel.receiveOrClosed` method to control cancellation atomicity of this function and tweaks `emitAll(ReceiveChannel)` implementation to abandon atomicity in favor of predictable cancellation.

Fixes #1265

### DIFF
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -227,17 +227,20 @@ public suspend inline fun <T> suspendAtomicCancellableCoroutine(
  *  Suspends coroutine similar to [suspendAtomicCancellableCoroutine], but an instance of [CancellableContinuationImpl] is reused if possible.
  */
 internal suspend inline fun <T> suspendAtomicCancellableCoroutineReusable(
+    resumeMode: Int = MODE_ATOMIC_DEFAULT,
     crossinline block: (CancellableContinuation<T>) -> Unit
 ): T = suspendCoroutineUninterceptedOrReturn { uCont ->
-        val cancellable = getOrCreateCancellableContinuation(uCont.intercepted())
+        val cancellable = getOrCreateCancellableContinuation(uCont.intercepted(), resumeMode)
         block(cancellable)
         cancellable.getResult()
     }
 
-internal fun <T> getOrCreateCancellableContinuation(delegate: Continuation<T>): CancellableContinuationImpl<T> {
+internal fun <T> getOrCreateCancellableContinuation(
+    delegate: Continuation<T>, resumeMode: Int
+): CancellableContinuationImpl<T> {
     // If used outside of our dispatcher
     if (delegate !is DispatchedContinuation<T>) {
-        return CancellableContinuationImpl(delegate, resumeMode = MODE_ATOMIC_DEFAULT)
+        return CancellableContinuationImpl(delegate, resumeMode)
     }
     /*
      * Attempt to claim reusable instance.
@@ -253,8 +256,8 @@ internal fun <T> getOrCreateCancellableContinuation(delegate: Continuation<T>): 
      *    thus leaking CC instance for indefinite time.
      * 2) Continuation was cancelled. Then we should prevent any further reuse and bail out.
      */
-    return delegate.claimReusableCancellableContinuation()?.takeIf { it.resetState() }
-        ?: return CancellableContinuationImpl(delegate, MODE_ATOMIC_DEFAULT)
+    return delegate.claimReusableCancellableContinuation()?.takeIf { it.resetState(resumeMode) }
+        ?: return CancellableContinuationImpl(delegate, resumeMode)
 }
 
 /**
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -204,30 +204,18 @@ public suspend inline fun <T> suspendCancellableCoroutine(
     }
 
 /**
- * Suspends the coroutine like [suspendCancellableCoroutine], but with *atomic cancellation*.
+ * Suspends the coroutine similar to [suspendCancellableCoroutine], but an instance of
+ * [CancellableContinuationImpl] is reused.
  *
- * When the suspended function throws a [CancellationException], it means that the continuation was not resumed.
- * As a side-effect of atomic cancellation, a thread-bound coroutine (to some UI thread, for example) may
- * continue to execute even after it was cancelled from the same thread in the case when the continuation
- * was already resumed and was posted for execution to the thread's queue.
- *
- * @suppress **This an internal API and should not be used from general code.**
+ * * when [resumeMode] is [MODE_CANCELLABLE_REUSABLE] works like [suspendCancellableCoroutine].
+ * * when [resumeMode] is [MODE_ATOMIC_REUSABLE] it has *atomic cancellation*.
+ *   When the suspended function throws a [CancellationException], it means that the continuation was not resumed.
+ *   As a side-effect of atomic cancellation, a thread-bound coroutine (to some UI thread, for example) may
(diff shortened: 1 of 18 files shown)


kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -227,17 +227,20 @@ public suspend inline fun <T> suspendAtomicCancellableCoroutine(
  *  Suspends coroutine similar to [suspendAtomicCancellableCoroutine], but an instance of [CancellableContinuationImpl] is reused if possible.
  */
 internal suspend inline fun <T> suspendAtomicCancellableCoroutineReusable(
```
qwwdfsad: This method is no longer "atomic" as its name suggests

kotlinx-coroutines-core/common/src/internal/DispatchedTask.kt
```diff
@@ -8,12 +8,17 @@ import kotlinx.coroutines.internal.*
...
-@PublishedApi internal const val MODE_CANCELLABLE = 1    // schedule cancellable dispatch for suspendCancellableCoroutine
-@PublishedApi internal const val MODE_UNDISPATCHED = 2   // when the thread is right, but need to mark it with current coroutine
+@PublishedApi
+internal const val MODE_ATOMIC_DEFAULT = 0 // schedule non-cancellable dispatch for suspendCoroutine
+@PublishedApi
+internal const val MODE_CANCELLABLE = 1    // schedule cancellable dispatch for suspendCancellableCoroutine
 
-internal val Int.isCancellableMode get() = this == MODE_CANCELLABLE
-internal val Int.isDispatchedMode get() = this == MODE_ATOMIC_DEFAULT || this == MODE_CANCELLABLE
+internal const val MODE_CANCELLABLE_REUSABLE = 2 // same as MODE_CANCELLABLE but supports reused
```
qwwdfsad: I am a bit concerned about the maintainability of this machinery.
E.g. it is easy to use `if (mode == MODE_CANCELLABLE) doSomeLogic()` instead of `mode.isCancellableMode`.
Let's at least leave a precautious comment with a suggestion to avoid using a comparison with constants directly.

qwwdfsad: >E.g. it is easy to use if (mode == MODE_CANCELLABLE) doSomeLogic() instead of mode.isCancellableMode.

Speaking of which, please take a look at `CancellableContinuationImpl#getResult:237`

kotlinx-coroutines-core/common/src/internal/DispatchedTask.kt
```diff
@@ -120,7 +125,7 @@ internal fun <T> DispatchedTask<T>.resume(delegate: Continuation<T>, useMode: In
     val result = if (exception != null) Result.failure(exception) else Result.success(state as T)
     when (useMode) {
         MODE_ATOMIC_DEFAULT -> delegate.resumeWith(result)
```
qwwdfsad: Makes sense to make extensions `inline` and use them there instead of hand-rolled comparison?

kotlinx-coroutines-core/common/test/flow/operators/FlowOnTest.kt
```diff
@@ -314,4 +314,19 @@ class FlowOnTest : TestBase() {
             assertEquals(expected, value)
         }
     }
+
+    @Test
+    fun testCancelledFlowOn() = runTest {
+        assertFailsWith<CancellationException> {
```
qwwdfsad: Thinking about the original issue (#1265), I am concerned about the fact that other channel-related operators may expose unexpected behaviour as well (aka touching view after its destruction).
E.g. the similar test with `combine` operator will execute its `collect` block.

I'd say we should re-visit our channel-related operators, measure the performance impact of checking cancellation in `collect` and either do it, or do it only in channel-related operators (and, maybe, provide a `checkedCollect` or something like this)

elizarov: It's not about checking in `collect`. It is about checking _on resume after suspension_. All our suspending function do check for cancellation on resume with exception of `Mutex.lock` , `Semaphore.acquire`, and `Channel.send/receive` function that do not do so, we need to revisit operation that use those functions (`flowOn` was one of them). In fact, we should really revisit atomicity of those underlying operations in the first place.

kotlinx-coroutines-core/common/src/internal/DispatchedTask.kt
elizarov: Good catch. Added test for that particular omission.

kotlinx-coroutines-core/common/src/internal/DispatchedTask.kt
elizarov: I don't want to lose a nice table-switch in this bytecode. Moreover, this particular code is "safe", since it has "catch all" else branch just in case some option is missing.

kotlinx-coroutines-core/common/test/flow/operators/FlowOnTest.kt
qwwdfsad: Do you want to do it separately or as part of this PR?
E.g. the following code:
```
coroutineScope {
    val flow =  flow {
        emit(Unit) // emit to buffer
        cancel() // now cancel
    }

    flow.combine(flow) { u, _ -> u }.collect {
        // should not be reached, because cancelled before it runs
        expectUnreached()
    }
}
```

will throw `ISE(Should not be reached)`

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
elizarov: Fixed

kotlinx-coroutines-core/common/test/flow/operators/FlowOnTest.kt
elizarov: I'll fix it all as a part of this PR.

elizarov:
We need a more thorough solution as fixing all those issue one-by-one creates quite an ugly code. Here's a separate issue: #1813. Let's fix it as part of this PR, too.

elizarov:
Aborting this PR in favor of #1937 that completely gets rid of "atomic cancellation", thus making `flowOn` automatically correct.

PR #1732, opened 2019-12-24
Breaking change: Experimental Flow.onCompletion contract for cause

### LINKED ISSUE #1693: Flow: Operator for action on successful completion
According to the [doc](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/on-completion.html), the `onCompletion()` "operator is transparent to exceptions that occur in downstream flow and does not observe exceptions that are thrown to cancel the flow".

If I want to do something only when the flow is "successfully completed" (i.e. no exception, not cancelled), how can I do that? Thanks.

It can be reproduced with the following code:
```
import kotlinx.coroutines.*

fun foo(): Flow<Int> = flow {
    var i = 0
    while (true) {
        emit(i++)
        delay(100)
    }
}

suspend fun main() {
    val job = GlobalScope.launch {
        foo().onEach { println("onEach: $it")}
            .onCompletion {
                println("onCompletion: $it")
            }
            .collect()
    }
    
    delay(500)
    println("after delay()")
    job.cancel()
    
    delay(100)
    println("after second delay()")
}
```

elizarov:
Flow.onCompletion now reports all failures and cancellation in its cause just like invokeOnCompletion. A null cause is reported if and only if flow had completed successfully (no failure, no cancellation). Emission of additional elements after the end of the flow is only possible from inside of onCompletion block in case of successful completion.

Also fixed a bug where onCompletion implementation was trying to add exception to its own list of suppressed exceptions, which is not allowed.

Fixes #1693

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt
@@ -81,8 +81,8 @@ public fun <T> Flow<T>.onStart(
 }
 
 /**
- * Invokes the given [action] when the given flow is completed or cancelled, using
- * the exception from the upstream (if any) as cause parameter of [action].
+ * Invokes the given [action] when the given flow is completed or cancelled, passing
+ * the cancellation exception or failure as cause parameter of [action].
  *
  * Conceptually, `onCompletion` is similar to wrapping the flow collection into a `finally` block,
  * for example the following imperative snippet:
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt
```diff
@@ -101,48 +101,55 @@ public fun <T> Flow<T>.onStart(
...
+ * ```
+ * myFlow
+ *     .onEach { println(it) }
+ *     .onCompletion { if (it == null) println("Completed successfully") }
  *     .collect()
  * ```
  *
  * The receiver of the [action] is [FlowCollector] and this operator can be used to emit additional
- * elements at the end of the collection. For example:
+ * elements at the end if it had completed successfully. For example:
```
qwwdfsad: Just `completed`

PR #1733, opened 2019-12-24
Build project using JDK 11

elizarov:


ui/kotlinx-coroutines-javafx/build.gradle
```diff
@@ -1,4 +1,40 @@
...
+        if (!System.env.JDK_18) {
+            throw new GradleException("JDK_18 environment variable is not defined. " +
+                    "Can't build against JDK 8 runtime and run JDK 8 compatibility tests. " +
+                    "Please ensure JDK 8 is installed and that JDK_18 points to it.")
+        }
+    }
+}
+
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile) {
+    kotlinOptions.jdkHome = System.env.JDK_18
```
elizarov: You don't actually need this. There's not really need to force its compilation under JDK_18 as it would create additional barriers for contrbitors who'd need JDK8 installed on their machines. It is Ok to only test it under JDK8.

dkhalanskyjb: Here I simply copied the code that concerned `JDK_16`. Isn't the solution to the problem of viability external contributions here basically the same, that is, to allow external contributors to just have `JDK_18=$JAVA_HOME`?

elizarov:
Passes all tests on JDK11 & also tested to pass stress tests. Ready for merge!

PR #1735, opened 2019-12-25
Generate site using Jekyll via Docker

elizarov:
This makes site-building process more stable and less dependent on the environment, requiring only Docker to be installed locally.


elizarov:
> Please update the building instructions as well (README in `site/`)

Good catch. They were even more outdated than this change. Fixed.

PR #1747, opened 2020-01-03
Remove the second yield()

tatsuyafujisaki:
Because it does not seem to make any difference.


fvasco:
`yield` resumes if the job is active (this is a side effect), it should be replaced by `ensureActive`.

tatsuyafujisaki:
Thanks for the input!

Now I understand that calling the second yield() is to indicate that the parent is not canceled.

PR #1748, opened 2020-01-03
Add test to ensure parent scope is not frozen by child coroutine

brendanw:
If this test fails, then we have no have a reasonable means of communicating information from one thread to another. Freezing should enforce that the object returned from another thread is frozen; it should not freeze the callback that receives the object.


elizarov:
Added similar thing here: https://github.com/Kotlin/kotlinx.coroutines/commit/10979a350426f23c02356c66acfda1e765d23450

PR #1754, opened 2020-01-07
Introduce SWT dispatchers for SWT UI applications

brudaswen:
1. Add `Dispatchers.SWT` that dispatches execution onto SWT event dispatching thread of the **default** SWT  `Display`.
2. Add `Dispatchers.swt(display: Display)` that disaptches to arbitrary SWT `Display`.
3. Implement `MainDispatcherFactory` for SWT that dispatches to default `Display`.

Is there an interest in having SWT support directly in the `Kotlin/kotlinx.coroutines` repository? Otherwise I would release `kotlinx-coroutines-swt` separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ suspend fun main() = coroutineScope {
 
 * [core](kotlinx-coroutines-core/README.md) &mdash; common coroutines across all platforms:
   * [launch] and [async] coroutine builders returning [Job] and [Deferred] light-weight futures with cancellation support;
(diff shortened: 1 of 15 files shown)


fvasco:
> Otherwise I would release kotlinx-coroutines-swt separately

👍

elizarov:
Thanks! We would appreciate if you release the corresponding module separately. It does not look like you would need any kind of internal API, so the result should be pretty stable. If you miss something, then please fill free to open a separate issue.

brudaswen:
In case anyone is looking for SWT support:
https://github.com/brudaswen/kotlinx-coroutines-swt

PR #1759, opened 2020-01-13
Call BroadcastChannel.asFlow.onStart() after openSubscription()

### LINKED ISSUE #1758: BroadcastChannel.asFlow().onStart(...) is invoked before the subscription is opened
In the docs, we have this example of `onStart(...)`:

```Kotlin
flowOf("a", "b", "c")
    .onStart { emit("Begin") }
    .collect { println(it) } // prints Begin, a, b, c
```

Not just with `emit(...)`, this works with sending to a `Channel` as well.

```Kotlin
val channel = Channel<String>(4)

channel.consumeAsFlow()
    .onStart { 
      channel.send("Begin")
      channel.send("a")
      // etc.
    }
    .collect { println(it) } // prints Begin, a, b, c
```

However, with a `BroadcastChannel` we won't get anything:

```Kotlin
val channel = broadcast {
  send(2)
}

channel.asFlow()
    .onStart { channel.send(1) }
    .collect { println(it) } // prints 2
}
```

If `onStart(...)` is primarily a callback to say that "we're collecting data", instead of a fancy way to prepend data to the `Flow`, then this feels wrong.

The difference is because of how the subscription is created.  The subscription channel isn't created until the builder is reached, which is after `onStart(...)`:

```Kotlin
public fun <T> BroadcastChannel<T>.asFlow(): Flow<T> = flow {
    emitAll(openSubscription())
}
```

`onStart(...)` is being invoked at the same time as always, and the value is being sent, but the subscription's `ReceiveChannel` is created _afterwards_ and never gets it.

### Proposed Solution

Make `asFlow()` return its own internal type (`BroadcastFlow`?), and give it an `update(...)` function similar to `ChannelFlow`.  This function will just accumulate the `action`(s) and store it/them until the final `collect(...)` is called.

Then make `onStart(...)` do a type check and just call `update(...)` if it's a `BroadcastFlow`.
When it's time to start collecting, create the subscription, `then` invoke the action, then call `emitAll(...)`.

```Kotlin
public fun <T> BroadcastChannel<T>.asFlow(): Flow<T> = BroadcastFlow(this)

internal class BroadcastFlow<T>(
    private val source: BroadcastChannel<T>,
    private val _startAction: suspend FlowCollector<T>.() -> Unit = {}
) : Flow<T> {

    fun update(
        startAction: suspend FlowCollector<T>.() -> Unit
    ): BroadcastFlow<T> = BroadcastFlow(channel) {
        startAction()
        _startAction()
    }

    override suspend fun collect(collector: FlowCollector<T>) {
        val channel = source.openSubscription()
        collector._startAction()
        collector.emitAll(channel)
    }
}

public fun <T> Flow<T>.onStart(
    action: suspend FlowCollector<T>.() -> Unit
): Flow<T> = if (this is BroadcastFlow) {
    update(action)
} else {
    unsafeFlow {
        // Note: unsafe flow is used here, but safe collector is used to invoke start action
        SafeCollector<T>(this, coroutineContext).action()
        collect(this) // directly delegate
    }
}
```

RBusarow:
Fixes #1758

Introduces the `BroadcastChannelAsFlow` internal class to accumulate `onStart` callbacks and only invoke them inside `collect(...)`, after `openSubscription()` has been called.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Channels.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Channels.kt
@@ -13,7 +13,6 @@ import kotlinx.coroutines.channels.*
 import kotlinx.coroutines.flow.internal.*
 import kotlin.coroutines.*
 import kotlin.jvm.*
-import kotlinx.coroutines.flow.internal.unsafeFlow as flow
 
 /**
  * Emits all elements from the given [channel] to this flow collector and [cancels][cancel] (consumes)
@@ -128,7 +127,7 @@ private class ChannelAsFlow<T>(
             check(!consumed.getAndSet(true)) { "ReceiveChannel.consumeAsFlow can be collected just once" }
         }
     }
-    
+
     override fun create(context: CoroutineContext, capacity: Int): ChannelFlow<T> =
         ChannelAsFlow(channel, consume, context, capacity)
 
@@ -170,9 +169,7 @@ private class ChannelAsFlow<T>(
  * 3) If the flow consumer fails with an exception, subscription is cancelled.
  */
 @FlowPreview
-public fun <T> BroadcastChannel<T>.asFlow(): Flow<T> = flow {
-    emitAll(openSubscription())
-}
+public fun <T> BroadcastChannel<T>.asFlow(): Flow<T> = BroadcastChannelAsFlow(this)
 
 /**
  * Creates a [broadcast] coroutine that collects the given flow.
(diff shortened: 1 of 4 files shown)


elizarov:
Discussion in #1758

qwwdfsad:
Closing as obsolete. Fixed in `SharedFlow` via `onSubscription`

PR #1774, opened 2020-01-21
Fix context support in Publisher.asFlow.flowOn

### LINKED ISSUE #1765: PublisherAsFlow ignores the CoroutineContext passed in from flowOn?
**kotlin version: 1.3.61**
**kotlinx.coroutines version: 1.3.3**

Hi, I found that `PublisherAsFlow` ignores the CoroutineContext when created. It always uses an `EmptyCoroutineContext` instead.

```kotlin
private class PublisherAsFlow<T : Any>(
    private val publisher: Publisher<T>,
    capacity: Int
) : ChannelFlow<T>(EmptyCoroutineContext, capacity) {
    // Just ignore the context, use EmptyCoroutineContext for ChannelFlow
    override fun create(context: CoroutineContext, capacity: Int): ChannelFlow<T> =
        PublisherAsFlow(publisher, capacity)
    ...
}
```

 So if I call `flowOn` on it, nothing happens.

```kotlin
Flowable.create<Int>({
        ... // run on ui thread, io thread expected!!!
    }, ...)
    .asFlow()
    .flowOn(Dispatchers.IO) // will be ignored!!!
    .launchIn(mainScope)
```

elizarov:
* When using asFlow().flowOn(...) context is now properly tracked and taken into account for both execution context of the reactive subscription and for injection into Reactor context.
* Publisher.asFlow slow-path implementation is simplified. It does not sure specialized openSubscription anymore, but always uses the same flow request logic.

Fixes #1765

### DIFF
--- a/reactive/kotlinx-coroutines-reactor/test/FluxContextTest.kt
+++ b/reactive/kotlinx-coroutines-reactor/test/FluxContextTest.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.reactor
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.reactive.*
+import org.junit.*
+import org.junit.Test
+import reactor.core.publisher.*
+import kotlin.test.*
+
+class FluxContextTest : TestBase() {
+    private val dispatcher = newSingleThreadContext("FluxContextTest")
+
+    @After
+    fun tearDown() {
+        dispatcher.close()
+    }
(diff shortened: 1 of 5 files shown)


reactive/kotlinx-coroutines-reactor/test/FluxContextTest.kt
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
qwwdfsad: Copyright year.
By the way, you can use `$today.year` in copyright template

elizarov: Good catch. Thanks.

Issue #1727, opened 2019-12-20
Java9 reactive streams integration


elizarov:
`kotlinx-coroutines-jdk9` integration module

dkhalanskyjb:
Solved by #1783.

Issue #162, opened 2017-11-12
Support for Java 9 Flow API


fvasco:
Rename kotlinx-coroutines-reactive package to Java 9 package.

elizarov:
Why rename? `kotlinx-coroutines-reactive` support Reactive Streams specification interfaces. It does not work with Java 8 `Flow.*` interfaces. We should define a new module for that. The only tricking part is how name it: either `kotlinx-coroutines-jdk9` or `kotlinx-coroutines-flow`.

fvasco:
Hi @elizarov,
I said that badly, I am fully agree with you.

Both names are nice,
I consider a bit better to offer a complete jdk 9 integration as a single dependency, however this goal can be reached by putting `kotlinx-coroutines-flow` as dependency of `kotlinx-coroutines-jdk9`.

elizarov:
What else there is to jdk 9 integration beyond `Flow.*` interfaces?

fvasco:
Nothing,
so I prefer the name `kotlinx-coroutines-jdk9`.

Flow api should not look different from time api.

qwwdfsad:
Fixed in 1.3.5

PR #1783, opened 2020-02-04
JDK9 Flow integration

dkhalanskyjb:
Solves https://github.com/Kotlin/kotlinx.coroutines/issues/162, https://github.com/Kotlin/kotlinx.coroutines/issues/1727


fvasco:
Even though the `reactive-streams-flow-adapters` simplifies this module a lot, it is an _important_ dependency.
This PR looks like a shortcut library to the `FlowAdapters.toPublisher` method, not so much in my humble opinion.

reactive/kotlinx-coroutines-jdk9/src/Await.kt
```diff
@@ -0,0 +1,44 @@
...
+package kotlinx.coroutines.jdk9
+
+import java.util.concurrent.Flow.Publisher
+import org.reactivestreams.FlowAdapters
+import kotlinx.coroutines.reactive.*
+
+/**
+ * A thin wrapper around [kotlinx.coroutines.reactive.awaitFirst] for JDK 9 Flow [Publisher].
+ */
+public suspend fun <T> Publisher<T>.awaitFirst(): T = FlowAdapters.toPublisher(this).awaitFirst()
```
qwwdfsad: When someone is going to use this integration, it's likely that they have only JDK Flow without an RS library.
Thus KDoc should be self-contained, it is completely okay to copy it.

And top-level module implementation can note that it's written via RS adapters

reactive/kotlinx-coroutines-jdk9/src/ReactiveFlow.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+public fun <T : Any> Flow<T>.asPublisher(): Publisher<T> {
+    val reactivePublisher : org.reactivestreams.Publisher<T> = this.asPublisher<T>()
+    return FlowAdapters.toFlowPublisher(reactivePublisher)
+}
+
+/**
+ * A thin wrapper around [kotlinx.coroutines.reactive.collect] for JDK 9 Flow [Publisher].
+ */
+public suspend fun <T> Publisher<T>.collect(action: (T) -> Unit) =
```
qwwdfsad: This function should be `inline` to avoid suspension in `action` and to avoid excess lambda allocation

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
```diff
@@ -0,0 +1,150 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.jdk9
```
qwwdfsad: It would be to have at least a couple of smoke tests for each public declaration in this module even though it is implemented via delegation.

reactive/kotlinx-coroutines-jdk9/README.md
```diff
@@ -0,0 +1,9 @@
+# Module kotlinx-coroutines-jdk9
```
qwwdfsad: Please mention this module in a top-level readme as well

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
```diff
@@ -0,0 +1,150 @@
...
+    private inline fun assertIAE(block: () -> Unit) {
+        try {
+            block()
+            expectUnreached()
+        } catch (e: Throwable) {
+            assertThat(e, IsInstanceOf(IllegalArgumentException::class.java))
+        }
+    }
+
+    private inline fun assertNSE(block: () -> Unit) {
```
qwwdfsad: Kotlin has `assrtFailsWith<NoSuchElementException>` instead

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
```diff
@@ -0,0 +1,150 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.jdk9
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.flowOn
+import org.hamcrest.MatcherAssert.*
+import org.hamcrest.core.*
```
qwwdfsad: Please avoid using `hamcrest`. There is nothing wrong with it, but all the tests don't use it and it's better to be consistent across the single code base

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
dkhalanskyjb: I didn't actually write any of these tests, just copied what was in `kotlinx-coroutines-reactive` with little change. Should I fix it there, too, in this PR?

reactive/kotlinx-coroutines-jdk9/src/ReactiveFlow.kt
dkhalanskyjb: Fixed.

reactive/kotlinx-coroutines-jdk9/README.md
dkhalanskyjb: Done.

reactive/kotlinx-coroutines-jdk9/src/Await.kt
dkhalanskyjb: > without an RS library

I'm not sure how this is possible, given that this library is implemented on top of the RS one. Am I missing something?

qwwdfsad: The library is present in the dependencies, but users may not use RS primitives directly or even be aware of them. So referring to RS documentation may seem a bit weird for them ("why documentation for standard JDK extensions and their documentation is superseded by extensions to some "reactivestreams" library?)

dkhalanskyjb: I see. I dropped the commit that replaced the comments with references to the `kotlinx-coroutines-reactive` docs.

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
dkhalanskyjb: I edited the commit that removed the tests; now it only removes the TCK integration tests, all the other ones are left intact.

reactive/kotlinx-coroutines-jdk9/src/Await.kt
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Awaits for the first value from the given observable or the [default] value if none is emitted without blocking a
+ * thread and returns the resulting value or throws the corresponding exception if this observable had produced error.
+ *
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
+ * immediately resumes with [CancellationException].
+ */
+public suspend fun <T> Publisher<T>.awaitFirstOrDefault(default: T): T =
```
dkhalanskyjb: Should these, too, be inline?

fvasco: There is nothing to inline

LouisCAD: @fvasco Actually, there's the `ContinuationImpl` to inline. You can even see that for suspending functions, there's no "nothing_to_inline" warning.

fvasco: @LouisCAD
Are you considered the tail call optimization?

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
dkhalanskyjb: Same here: should I also fix this in the original tests?

reactive/kotlinx-coroutines-jdk9/src/Await.kt
LouisCAD: I'm not, but I expect inlining a suspending function like that optimizes further as there's no "nothing_to_inline" warning. Maybe @elizarov can confirm?

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
elizarov: Yes, it would be great to fix tests there, too.

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
elizarov: Yes, it would be a great fix while we are at it.

reactive/kotlinx-coroutines-jdk9/src/Await.kt
elizarov: It does not have to be inline because there is a tail-call optimization.

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
dkhalanskyjb: Fixed in another PR.

reactive/kotlinx-coroutines-jdk9/test/IntegrationTest.kt
dkhalanskyjb: I replaced such methods where I found them.

reactive/kotlinx-coroutines-jdk9/src/Await.kt
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.jdk9
+
+import java.util.concurrent.Flow.Publisher
```
elizarov: Let's avoid this kind of import. The code here and in other classes would be easier to read if we only have `import java.util.concurrent.*` and the in the code explicitly write `Flow.Publisher<T>`.

reactive/kotlinx-coroutines-jdk9/src/Publish.kt
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Coroutine context can be specified with [context] argument.
+ * If the context does not have any dispatcher nor any other [ContinuationInterceptor], then [Dispatchers.Default] is used.
+ * Method throws [IllegalArgumentException] if provided [context] contains a [Job] instance.
+ *
+ * **Note: This is an experimental api.** Behaviour of publishers that work as children in a parent scope with respect
+ *        to cancellation and error handling may change in the future.
+ */
+@ExperimentalCoroutinesApi
+public fun <T> publish(
```
elizarov: Let's give it a diffferent name from a version that return reactive streams publisher. It returns `Flow.Publisher<T>`, so let's call it `flowPublish`.

reactive/kotlinx-coroutines-jdk9/src/Await.kt
dkhalanskyjb: Fixed.

reactive/kotlinx-coroutines-jdk9/src/Publish.kt
dkhalanskyjb: Done.

dkhalanskyjb:
Since this subproject outright requires to be built under JDK 9+, calling `gradle build` here with JDK 1.8 will not work properly. It should probably be dealt with, either by dropping support for JDK 1.8 and changing the build instructions in the top-level README to require JDK 11, or by building this subproject only conditionally. @elizarov , what would be the right thing to do?

elizarov:
> Since this subproject outright requires to be built under JDK 9+, calling `gradle build` here with JDK 1.8 will not work properly. It should probably be dealt with, either by dropping support for JDK 1.8 and changing the build instructions in the top-level README to require JDK 11, or by building this subproject only conditionally. @elizarov , what would be the right thing to do?

It is already mentioned in top-level readme:

> * JDK >= 11 referred to by the `JAVA_HOME` environment variable.

PR #1785, opened 2020-02-04
Remove extraneous word in comment

EricAtPlanGrid:
Just a quick drive-by discovered while browsing the source.


qwwdfsad:
Thanks

Issue #1695, opened 2019-12-10
Add .asFlow() extensions for ObservableValue to kotlinx-coroutines-javafx library


qwwdfsad:
>It would be nice if all observable properties could be easily converted to Flow when coding with TornadoFX. This could be added either for `javafx.beans.value.ObservableValue` or (maybe even better) for the entire `javafx.beans.Observable` interface if possible.

Originally reported here: https://youtrack.jetbrains.com/issue/KT-35389

LittleLightCz:
Thank you, I am watching like a hawk :eagle: :relaxed:

PR #1789, opened 2020-02-07
Implement ObservableValue<T>.asFlow()

dkhalanskyjb:
Solves https://github.com/Kotlin/kotlinx.coroutines/issues/1695

It seems there is no way to implement this for `Observable` as it doesn't even expose the type of the observable value.

### DIFF
--- a/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
+++ b/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
@@ -1,0 +1,38 @@
+package kotlinx.coroutines.javafx
+
+import javafx.beans.value.ChangeListener
+import javafx.beans.value.ObservableValue
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.conflate
+import kotlinx.coroutines.flow.flowOn
+
+/**
+ * Creates an instance of a cold [Flow] that subscribes to the given [ObservableValue] and produces
+ * its values as they change.
+ *
+ * The resulting flow is conflated, meaning that if several values arrive in a quick succession, only
+ * the last one will be produced.
+ *
+ * It produces at least one value.
+ *
+ * Since this implementation uses [ObservableValue.addListener], even if this [ObservableValue]
+ * supports lazy evaluation, eager computation will be enforced while the flow is being collected.
+ */
+@ExperimentalCoroutinesApi
+public fun <T: Any> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
+    val listener = ChangeListener<T> { _, _, newValue ->
+        try {
+            offer(newValue)
+        } catch (e: CancellationException) {
+            // In case the event fires after the channel is closed
+        }
+    }
+    addListener(listener)
+    send(value)
+    awaitClose {
+        removeListener(listener)
+    }
+}.flowOn(Dispatchers.JavaFx).conflate()
--- a/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
+++ b/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
@@ -8,10 +8,7 @@ import javafx.beans.value.ChangeListener
 import javafx.beans.value.ObservableValue
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.awaitClose
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.callbackFlow
-import kotlinx.coroutines.flow.conflate
-import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.flow.*
 
 /**
  * Creates an instance of a cold [Flow] that subscribes to the given [ObservableValue] and produces
--- a/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
+++ b/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
@@ -11,18 +11,12 @@ import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.flow.*
 
 /**
- * Creates an instance of a cold [Flow] that subscribes to the given [ObservableValue] and produces
- * its values as they change.
- *
- * The resulting flow is conflated, meaning that if several values arrive in a quick succession, only
- * the last one will be produced.
- *
- * It produces at least one value.
- *
+ * Creates an instance of a cold [Flow] that subscribes to the given [ObservableValue] and emits
+ * its values as they change. The resulting flow is conflated, meaning that if several values arrive in quick
+ * succession, only the last one will be emitted.
  * Since this implementation uses [ObservableValue.addListener], even if this [ObservableValue]
  * supports lazy evaluation, eager computation will be enforced while the flow is being collected.
- *
- * All the calls to JavaFX API are performed in the JavaFX application thread.
+ * All the calls to JavaFX API are performed in [Dispatchers.JavaFx].
  *
  * ### Operator fusion
  *
--- a/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
+++ b/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.flow.*
  * Since this implementation uses [ObservableValue.addListener], even if this [ObservableValue]
  * supports lazy evaluation, eager computation will be enforced while the flow is being collected.
  * All the calls to JavaFX API are performed in [Dispatchers.JavaFx].
+ * This flow emits at least the initial value.
  *
  * ### Operator fusion
  *
--- a/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
+++ b/ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
@@ -1,3 +1,7 @@
+/*
+ *  Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 package kotlinx.coroutines.javafx
 
 import javafx.beans.value.ChangeListener
@@ -20,9 +24,16 @@ import kotlinx.coroutines.flow.flowOn
  *
  * Since this implementation uses [ObservableValue.addListener], even if this [ObservableValue]
  * supports lazy evaluation, eager computation will be enforced while the flow is being collected.
+ *
+ * All the calls to JavaFX API are performed in the JavaFX application thread.
+ *
+ * ### Operator fusion
+ *
+ * Adjacent applications of [flowOn], [buffer], [conflate], and [produceIn] to the result of `asFlow` are fused.
+ * [conflate] has no effect, as this flow is already conflated; one can use [buffer] to change that instead.
  */
 @ExperimentalCoroutinesApi
-public fun <T: Any> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
+public fun <T> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
     val listener = ChangeListener<T> { _, _, newValue ->
         try {
             offer(newValue)
--- a/ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
+++ b/ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
@@ -1,0 +1,86 @@
+package kotlinx.coroutines.javafx
+
+import javafx.beans.property.SimpleIntegerProperty
+import kotlinx.coroutines.TestBase
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+import org.junit.Before
+import org.junit.Test
+import kotlin.test.*
+
+
+class JavaFxObservableAsFlowTest : TestBase() {
+
+    @Before
+    fun setup() {
+        ignoreLostThreads("JavaFX Application Thread", "Thread-", "QuantumRenderer-", "InvokeLaterDispatcher")
+    }
+
+    @Test
+    fun testFlowOrder() = runTest {
+        if (!initPlatform()) {
+            println("Skipping JavaFxTest in headless environment")
+            return@runTest // ignore test in headless environments
+        }
+
+        val integerProperty = SimpleIntegerProperty(0)
+        val n = 10000 * stressTestMultiplier
+        val flow = integerProperty.asFlow().takeWhile { j -> j != n }
+        newSingleThreadContext("setter").use { pool ->
+            launch(pool) {
+                for (i in 1..n) {
+                    launch(Dispatchers.JavaFx) {
+                        integerProperty.set(i)
+                    }
+                }
+            }
+            var i = -1
+            flow.collect { j ->
+                assertTrue(i < (j as Int), "Elements are neither repeated nor shuffled")
+                i = j
+            }
+        }
+    }
+
+    @Test
+    fun testConflation() = runTest {
+        if (!initPlatform()) {
+            println("Skipping JavaFxTest in headless environment")
+            return@runTest // ignore test in headless environments
+        }
+
+        withContext(Dispatchers.JavaFx) {
+            val END_MARKER = -1
+            val integerProperty = SimpleIntegerProperty(0)
+            val flow = integerProperty.asFlow().takeWhile { j -> j != END_MARKER }
+            launch {
+                yield() // to subscribe to [integerProperty]
+                yield() // send 0
+                integerProperty.set(1)
+                expect(3)
+                yield() // send 1
+                expect(5)
+                integerProperty.set(2)
+                for (i in (-100..-2)) {
+                    integerProperty.set(i) // should be skipped due to conflation
+                }
+                integerProperty.set(3)
+                expect(6)
+                yield() // send 2 and 3
+                integerProperty.set(-1)
+            }
+            expect(1)
+            flow.collect { i ->
+                when (i) {
+                    0 -> expect(2)
+                    1 -> expect(4)
+                    2 -> expect(7)
+                    3 -> expect(8)
+                    else -> fail("i is $i")
+                }
+            }
+            finish(9)
+        }
+    }
+
+}
--- a/ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
+++ b/ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
@@ -24,7 +24,7 @@ class JavaFxObservableAsFlowTest : TestBase() {
         }
 
         val integerProperty = SimpleIntegerProperty(0)
-        val n = 10000 * stressTestMultiplier
+        val n = 1000
         val flow = integerProperty.asFlow().takeWhile { j -> j != n }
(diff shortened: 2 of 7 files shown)


ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,43 @@
...
+@ExperimentalCoroutinesApi
+fun <T: Any> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
+    val listener = ChangeListener<T> { observable, oldValue, newValue ->
+        try {
+            offer(newValue)
+        } catch (e: CancellationException) {
+            // In case the event fires after the channel is closed
+        }
+    }
+    withContext(Dispatchers.JavaFx) {
```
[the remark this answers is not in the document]
qwwdfsad: Also, could you please port [this test](https://github.com/ReactiveX/RxJavaFX/blob/a78ca7d15f7d82d201df8fafb6eba732ec17e327/src/test/java/io/reactivex/rxjavafx/RxJavaFXTest.java) to `ObservableValue<T>.asFlow()`? It looks like we are missing a decent integration test

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Since this implementation uses [ObservableValue.addListener], even if this [ObservableValue]
+ * supports lazy evaluation, eager computation will be enforced while the flow is being collected.
+ */
+@ExperimentalCoroutinesApi
+fun <T: Any> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
+    val listener = ChangeListener<T> { observable, oldValue, newValue ->
+        try {
+            offer(newValue)
+        } catch (e: CancellationException) {
```
qwwdfsad: What is going to happen if  `offer` throws a non-cancellation exception, e.g. because `collect {}` block has thrown it? Is it behaviour consistent and non-racy?

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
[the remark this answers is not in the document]
dkhalanskyjb: > Is it necessary to add and remove listener in JavaFx thread?

It turns out that JavaFX is supposed to be used in a single thread for *everything*. It did catch me, too.

> Could you please point out why?

Sure: `removeListener` and `set` are not synchronized at all, and, as seen here https://java-browser.yawk.at/org.openjfx/javafx-base/11/com/sun/javafx/binding/ExpressionHelper.java#348, and here https://java-browser.yawk.at/org.openjfx/javafx-base/11/com/sun/javafx/binding/ExpressionHelper.java#311, if `removeListener` is run somewhere other than JavaFX thread, an NPE or an OutOfRange could happen.

Likewise with `addListener`: if another `addListener` was running in the JavaFX thread, similar problems would occur.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,43 @@
...
+            // In case the event fires after the channel is closed
+        }
+    }
+    withContext(Dispatchers.JavaFx) {
+        addListener(listener)
+        send(value)
```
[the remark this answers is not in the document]
dkhalanskyjb: No one would dare to send a value in a way that could race with an `addListener`, because it could lead to an NPE, for example. I could describe a specific execution if needed, but that is highly elaborate and relies on happens-before reasoning. My point is, JavaFX is really not thread-safe, so if someone tries to do something outside of the JavaFX thread, they are bound to find trouble no matter what.

ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+
+        flow.collect { i ->
+            assertTrue(i == 1 || i == 2)
+        }
+    }
+
+    @Test
+    fun cancellationRaceStressTest() = runTest {
```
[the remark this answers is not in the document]
dkhalanskyjb: What constitutes a stress-test? Many tests in this code base use `stressTestMultiplier` without being called stress-tests. Should I rename this test not to have `StressTest` in the name or should I move it to a separate file? What is the difference?

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,43 @@
...
+ * The resulting flow is conflated, meaning that if several values arrive in a quick succession, only
+ * the last one will be produced.
+ *
+ * It produces at least one value.
+ *
+ * Since this implementation uses [ObservableValue.addListener], even if this [ObservableValue]
+ * supports lazy evaluation, eager computation will be enforced while the flow is being collected.
+ */
+@ExperimentalCoroutinesApi
+fun <T: Any> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
```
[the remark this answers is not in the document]
dkhalanskyjb: Understood, thanks. I told IDEA not to chastise me for redundant `public` qualifiers, and will fix this.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
dkhalanskyjb: As usual for this PR, I think that this is fine because everything here (this offer here, `addListener`, `removeListener` and so on) happens in the same JavaFX thread. So it's probably fully consistent and not racy.

I found a problem though: if someone runs `collect` in the JavaFX main thread, everything hangs. Should we consider this case? And can we?

dkhalanskyjb: Removing `runBlocking`, as recommended above, did solve the hanging problem.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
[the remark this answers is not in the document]
dkhalanskyjb: Fixed.

ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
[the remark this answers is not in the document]
dkhalanskyjb: Fixed.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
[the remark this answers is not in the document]
dkhalanskyjb: This version does work nicely! I replaced mine with it.

I have not yet understood how to adapt the test though. I am submitting a new version of code for review anyway because introducing a new test is independent from other changes.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,38 @@
...
+public fun <T: Any> ObservableValue<T>.asFlow(): Flow<T> = callbackFlow<T> {
+    val listener = ChangeListener<T> { _, _, newValue ->
+        try {
+            offer(newValue)
+        } catch (e: CancellationException) {
+            // In case the event fires after the channel is closed
+        }
+    }
+    addListener(listener)
+    send(value)
```
dkhalanskyjb: There is a problem with this behavior:
```kotlin
    @Test
    fun failure() = runTest {
        withContext(Dispatchers.JavaFx) {
            val END_MARKER = "xxx"
            val integerProperty = SimpleStringProperty()
            val flow = integerProperty.asFlow().takeWhile { j -> j != END_MARKER }
            launch {
                yield()
                yield()
                integerProperty.set("xxx")
            }
            flow.collect {

            }
        }
    }
```

dkhalanskyjb: This test fails, because the initial value is `null`, and `null` can't be passed to `send` for some reason.

However, the problem is deeper:
```kotlin
    @Test
    fun failure() = runTest {
        withContext(Dispatchers.JavaFx) {
            val END_MARKER = "xxx"
            val integerProperty = SimpleStringProperty("xyzzy")
            val flow: Flow<String?> = integerProperty.asFlow().takeWhile {
                it != null && it == END_MARKER
            }
            launch {
                yield()
                integerProperty.set(null)
                yield()
                integerProperty.set(END_MARKER)
            }
            flow.collect {

            }
        }
    }
```
also fails, and this can't be easily fixed: https://docs.oracle.com/javase/8/javafx/api/javafx/beans/value/WritableIntegerValue.html#setValue-java.lang.Number- JavaFX is, as seen in many places, fine with storing `null` in properties.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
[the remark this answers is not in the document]
dkhalanskyjb: Added the integration test.

ui/kotlinx-coroutines-javafx/test/examples/FxAsFlow.kt
```diff
@@ -0,0 +1,101 @@
...
+ * https://github.com/ReactiveX/RxJavaFX/blob/a78ca7d15f7d82d201df8fafb6eba732ec17e327/src/test/java/io/reactivex/rxjavafx/RxJavaFXTest.java
+ */
+class FxAsFlowApp: Application(), CoroutineScope {
+
+    private var job = Job()
+    override val coroutineContext: CoroutineContext
+        get() = JavaFx + job
+
+    private  val  incrementBttn = Button("Increment")
+    private  val  incrementLabel = Label("")
```
qwwdfsad: Please format it properly

ui/kotlinx-coroutines-javafx/test/JavaFxStressTest.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class JavaFxStressTest : TestBase() {
+
+    @Before
+    fun setup() {
+        ignoreLostThreads("JavaFX Application Thread", "Thread-", "QuantumRenderer-", "InvokeLaterDispatcher")
+    }
+
+    @Test
+    fun cancellationRaceStressTest() = runTest {
```
qwwdfsad: `testCancellationRace`: `test` should be in the prefix and `stressTest` is already present in the class name

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
qwwdfsad: This test does not pass because the compiler inserts null-checks in `ChangeListener`.
And it does so because you have explicit `Any` bound: `public fun <T: Any> ObservableValue<T>.asFlow()`.

It looks like you can remove `: Any` upper bound and ignore the initial null value.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,38 @@
+package kotlinx.coroutines.javafx
```
qwwdfsad: Missing copyright. You can setup one in `Preferences -> Copyright profiles`:
`Copyright 2016-$today.year JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.`

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,38 @@
...
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.conflate
+import kotlinx.coroutines.flow.flowOn
+
+/**
+ * Creates an instance of a cold [Flow] that subscribes to the given [ObservableValue] and produces
+ * its values as they change.
```
qwwdfsad: It's worth to mention that is all happening on Fx main thread.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * The resulting flow is conflated, meaning that if several values arrive in a quick succession, only
```
qwwdfsad: It's worth to mention that it is possible to change this behaviour with `[buffer]` operator, similarly to other operators

ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    @Test
+    fun testFlowOrder() = runTest {
+        if (!initPlatform()) {
+            println("Skipping JavaFxTest in headless environment")
+            return@runTest // ignore test in headless environments
+        }
+
+        val integerProperty = SimpleIntegerProperty(0)
+        val n = 10000 * stressTestMultiplier
```
qwwdfsad: please remove stressTestMultiplier and make the constant smaller

qwwdfsad:
Please, rebase all the changes on current develop

ui/kotlinx-coroutines-javafx/test/examples/FxAsFlow.kt
dkhalanskyjb: Done.

ui/kotlinx-coroutines-javafx/test/JavaFxStressTest.kt
dkhalanskyjb: Fixed.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
dkhalanskyjb: Yes, that helped!

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
dkhalanskyjb: Configured and fixed.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
dkhalanskyjb: I added a mention, but depending on why it's needed, maybe it should be worded differently.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
dkhalanskyjb: Added some text about operator fusion.

ui/kotlinx-coroutines-javafx/test/JavaFxObservableAsFlowTest.kt
dkhalanskyjb: Done.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,46 @@
...
+package kotlinx.coroutines.javafx
+
+import javafx.beans.value.ChangeListener
+import javafx.beans.value.ObservableValue
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.*
+
+/**
+ * Creates an instance of a cold [Flow] that subscribes to the given [ObservableValue] and produces
```
qwwdfsad: KDoc wording: Flow does not *produce*, it *emits*

qwwdfsad: `a quick succession` -> `quick succession`

qwwdfsad: A lot of newlines are not really necessary

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
qwwdfsad: Please use `Dispatchers.JavaFx` wording instead of `in the JavaFX application thread.`
Referring library primitives is always beneficial

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
```diff
@@ -0,0 +1,46 @@
...
+ * its values as they change.
+ *
+ * The resulting flow is conflated, meaning that if several values arrive in a quick succession, only
+ * the last one will be produced.
+ *
+ * It produces at least one value.
```
qwwdfsad: >It produces at least one value.

This guarantee is a bit weird and leads to an inconvenient API.
Basically, it means that one **cannot** have `ObservableValue<ObjectSubtype>` without a default value (a common case in UI) and non-null bound.
Have you considered an option to ignore the very first value (line 42) if it is a `null` like Rx adapter does?

dkhalanskyjb: >  Basically, it means that one cannot have `ObservableValue<ObjectSubtype>` without a default value (a common case in UI) and non-null bound.

True, but I don't see how this would be possible in any case:
https://docs.oracle.com/javase/8/javafx/api/javafx/beans/value/ObservableValue.html#getValue-- is it even allowed to implement an `ObservableValue` that throws an exception on calls to `getValue()`?

> This guarantee is a bit weird and leads to an inconvenient API.

The reason why I even added this guarantee was convenience: this would allow one to collect this flow in a thread aside from the JavaFX main thread and always have an initial value to work with, not worrying with synchronization. Otherwise, in order to always have an up-to-date version of an `ObservableValue` in a separate thread, one would have to block the JavaFX thread, which is a no-go.

>  Have you considered an option to ignore the very first value (line 42) if it is a null like Rx adapter does?

This would violate the guarantee that at least one value is produced, and if that guarantee does not hold, what would be a good reason to emit a value at all?

In short, if there is such a use case—collecting from a separate thread and being sure that you are not terribly out-of-date—then, I believe, at least one value should be emitted; otherwise, I don't see why emit an initial value in any case, but the solution with sending an initial value only if it is not null is seemingly not worse.

ui/kotlinx-coroutines-javafx/test/JavaFxStressTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        if (!initPlatform()) {
+            println("Skipping JavaFxTest in headless environment")
+            return@runTest // ignore test in headless environments
+        }
+
+        val integerProperty = SimpleIntegerProperty(0)
+        val flow = integerProperty.asFlow()
+        var i = 1
+        val n = 1000 * stressTestMultiplier
+        newSingleThreadContext("collector").use { pool ->
```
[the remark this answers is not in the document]
dkhalanskyjb: Could you please elaborate? grepping the whole codebase for `TestExecutor` only shows something seemingly irrelevant.

ui/kotlinx-coroutines-javafx/src/JavaFxConvert.kt
dkhalanskyjb: I took the word "produce" from https://kotlinlang.org/docs/reference/coroutines/flow.html where it is used not only in the context of producing an output of the whole program, so maybe it, too, should be fixed there.

In any case, I committed an improved KDoc.

ui/kotlinx-coroutines-javafx/test/JavaFxStressTest.kt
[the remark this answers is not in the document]
qwwdfsad: My bad, meant `ExecutorRule`

PR #1792, opened 2020-02-10
Remove all traces of Hamcrest

dkhalanskyjb:
Did it in a semi-automated manner, doing minimal fixes over the output of the command:
```sh
git grep hamcrest |
cut -d: -f1 |
xargs sed -iE -e 's/assertThat(\(.*\),\s*IsEqual(\(.*\)))/assertEquals(\2, \1)/' \
-e 's/assertThat(\(.*\),\s*IsInstanceOf(\(.*\)::class.java))/assert(\1 is \2)/' \
-e 's/assertThat(\(.*\),\s*IsNull())/assertEquals(null, \1)/' \
-e 's/assertThat(\(.*\),\s*IsNot(IsEqual(\(.*\))))/assert(\1 != \2)/' \
-e '/import.*hamcrest/d'
```

There are still some problems with style: in some cases `org.junit.Assert.*` are used, while in some others the native versions provided by Kotlin are.

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/channels/ActorTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/channels/ActorTest.kt
@@ -5,12 +5,11 @@
 package kotlinx.coroutines.channels
 
 import kotlinx.coroutines.*
-import org.hamcrest.core.*
-import org.junit.*
-import org.junit.Assert.*
+import org.junit.Test
 import org.junit.runner.*
 import org.junit.runners.*
 import java.io.*
+import kotlin.test.*
 
 @RunWith(Parameterized::class)
 class ActorTest(private val capacity: Int) : TestBase() {
@@ -28,14 +27,14 @@ class ActorTest(private val capacity: Int) : TestBase() {
             expect(3)
         }
         actor as Job // type assertion
-        assertThat(actor.isActive, IsEqual(true))
-        assertThat(actor.isCompleted, IsEqual(false))
-        assertThat(actor.isClosedForSend, IsEqual(false))
+        assertTrue(actor.isActive)
+        assertFalse(actor.isCompleted)
+        assertFalse(actor.isClosedForSend)
         expect(2)
         yield() // to actor code
-        assertThat(actor.isActive, IsEqual(false))
-        assertThat(actor.isCompleted, IsEqual(true))
-        assertThat(actor.isClosedForSend, IsEqual(true))
+        assertFalse(actor.isActive)
+        assertTrue(actor.isCompleted)
+        assertTrue(actor.isClosedForSend)
         finish(4)
     }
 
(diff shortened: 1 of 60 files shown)


reactive/kotlinx-coroutines-reactor/test/FluxTest.kt
```diff
@@ -23,7 +22,7 @@ class FluxTest : TestBase() {
         expect(2)
         flux.subscribe { value ->
             expect(5)
-            Assert.assertThat(value, IsEqual("OK"))
+            Assert.assertEquals("OK", value)
```
elizarov: `Assert.` prefix is wrong and should be removed.

kotlinx-coroutines-core/jvm/test/channels/ConflatedBroadcastChannelNotifyStressTest.kt
```diff
@@ -5,10 +5,9 @@
 package kotlinx.coroutines.channels
 
 import kotlinx.coroutines.*
-import org.hamcrest.MatcherAssert.*
-import org.hamcrest.core.*
 import org.junit.*
 import java.util.concurrent.atomic.*
+import kotlin.test.assertEquals
```
elizarov: Start imports should be used (can be configured once, then do "optimize import" on files that were changed by commit)

reactive/kotlinx-coroutines-reactor/test/MonoTest.kt
```diff
@@ -52,8 +51,8 @@ class MonoTest : TestBase() {
             expectUnreached()
         }, { error ->
             expect(5)
-            assertThat(error, IsInstanceOf(RuntimeException::class.java))
-            assertThat(error.message, IsEqual("OK"))
+            assert(error is RuntimeException)
```
elizarov: that should be replaced with `assertTrue(...)`, not `assert(...)`

reactive/kotlinx-coroutines-reactor/test/SchedulerTest.kt
```diff
@@ -24,13 +21,13 @@ class SchedulerTest : TestBase() {
         val mainThread = Thread.currentThread()
         withContext(Schedulers.single().asCoroutineDispatcher()) {
             val t1 = Thread.currentThread()
-            assertThat(t1, IsNot(IsEqual(mainThread)))
+            assert(t1 != mainThread)
```
elizarov: Should use `assertNotSame(....)` here.

kotlinx-coroutines-core/jvm/test/channels/ActorTest.kt
```diff
@@ -28,14 +27,14 @@ class ActorTest(private val capacity: Int) : TestBase() {
             expect(3)
         }
         actor as Job // type assertion
-        assertThat(actor.isActive, IsEqual(true))
-        assertThat(actor.isCompleted, IsEqual(false))
-        assertThat(actor.isClosedForSend, IsEqual(false))
+        assertEquals(true, actor.isActive)
```
dkhalanskyjb: These are also easy to replace with `assertTrue` and `assertFalse`. Would it be better?

reactive/kotlinx-coroutines-reactor/test/FluxTest.kt
dkhalanskyjb: Do you mean that `assertEquals` should be taken from `kotlin.test` or that `Assert` should be part of the import at the top of the file?

kotlinx-coroutines-core/jvm/test/channels/ConflatedBroadcastChannelNotifyStressTest.kt
dkhalanskyjb: I've never heard of "start imports". Google suggests that it meant "starred imports". Did it guess right?

dkhalanskyjb:
I think I fixed the problems. I also automatically replaced `assertEquals(null, X)` with `assertNull(X)` and changed uses of `org.junit.Assert` to `kotlin.test`. I have doubts about `ConvertTest` though.

kotlinx-coroutines-core/jvm/test/channels/ConflatedBroadcastChannelNotifyStressTest.kt
elizarov: Yes. That was type. I meant "star imports" as in `import kotlin.test.*`

PR #1793, opened 2020-02-10
Make SafeCollector platform-specific declaration and enforce exceptio…

qwwdfsad:
…n transparency invariant on JVM

    * Make it in allocation-free manner by using crafty trick with casting KSuspendFunction to Function and pass reusable object as a completion


kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -149,8 +149,8 @@ import kotlin.coroutines.*
  * it hard to reason about the code because an exception in the `collect { ... }` could be somehow "caught"
  * by an upstream flow, limiting the ability of local reasoning about the code.
  *
- * Currently, the flow infrastructure does not enforce exception transparency contracts, however, it might be enforced
- * in the future either at run time or at compile time.
+ * Flow machinery enforces exception transparency at runtime an throws [IllegalStateException] on any attempt to emit value,
+ * if an exception has been thrown on previous attemp.
```
elizarov: Typos: 
* "_an_ throws" -> "and throws"
* "on previous _attemp_" -> "on a previous attempt" (also article)

kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -199,7 +199,7 @@ public abstract class AbstractFlow<T> : Flow<T> {
 
     @InternalCoroutinesApi
     public final override suspend fun collect(collector: FlowCollector<T>) {
-        collectSafely(SafeCollector(collector, collectContext = coroutineContext))
+        collectSafely(SafeCollector(collector, coroutineContext))
```
elizarov: `try { .. } finally { ... }` should be here, too.

kotlinx-coroutines-core/jvm/src/flow/internal/SafeCollector.kt
```diff
@@ -0,0 +1,143 @@
...
+    private val emitFun = run {
+        collector::emit as Function2<T, Continuation<Unit>, Any?>
+    }
+
+    /*
+     * Implementor of ContinuationImpl (that will be preserved by ABI implementation nearly forever)
+     * in order to properly control 'intercepted()' lifecycle.
+     */
+    @Suppress("CANNOT_OVERRIDE_INVISIBLE_MEMBER", "INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+    private inner class InterceptedContinuationOwner : ContinuationImpl(NoOpContinuation, EmptyCoroutineContext) {
```
elizarov: `SafeCollector` can just extend `ContinuationImpl` directly, without having to allocate a separate `intercepted` field.

kotlinx-coroutines-core/jvm/src/flow/internal/SafeCollector.kt
```diff
@@ -0,0 +1,143 @@
...
+internal actual class SafeCollector<T> actual constructor(
+    @JvmField internal actual val collector: FlowCollector<T>,
+    @JvmField internal actual val collectContext: CoroutineContext
+) : FlowCollector<T> {
+
+    @JvmField // Note, it is non-capturing lambda, so no extra allocation during init of SafeCollector
+    internal actual val collectContextSize = collectContext.fold(0) { count, _ -> count + 1 }
+    private var lastEmissionContext: CoroutineContext? = null
+    private var intercepted: InterceptedContinuationOwner = InterceptedContinuationOwner()
+    private val emitFun = run {
```
elizarov: There's no need to use `run { ... }` and this whole `emitFun` should be a top-level constant in this file. Hint: you don't have to bind `collector` ahead of time, you can do:
```
private val emitFun =
    FlowCollector<*>::emit as Function3<FlowCollector<*>, Any?, Continuation<Unit>, Any?>
```
and the substitute an actual collector at the call-site.

kotlinx-coroutines-core/common/src/flow/Flow.kt
qwwdfsad: good catch

kotlinx-coroutines-core/jvm/src/flow/internal/SafeCollector.kt
qwwdfsad: Good point. Though for wildcards compiler explicitly puts checkcast to `Void` class, so `Any?` should be used instead

PR #1795, opened 2020-02-10
Indicate that root README should be updated with new contributions

MariusVolkhart:


elizarov:
Thanks! Good addition.

integration/README.md
```diff
@@ -23,5 +23,6 @@ Follow the following simple guidelines when contributing integration with your f
   * List of modules in this document.
   * List of modules in top-level [`settings.gradle`](../settings.gradle).
   * List of modules at the root of documentation site in [`site/docs/index.md`](../site/docs/index.md).
+  * List of integrations in the root [README.md](../README.md)
```
elizarov: Nit: should end the sentence with comma.

MariusVolkhart: @elizarov I've made a change, but used a period rather than a comma, since the surrounding code used the periods as well. If you'd still like a comma, I'm happy to use that instead.

Issue #1078, opened 2019-04-09
Add first { … } terminal operator to Flow


LouisCAD:
Would be much like `first { … }` from `kotlin.collections`, collecting all elements until the predicate returns true, after which it would cancel the flow.

I'm not sure what the behavior should be when the flow is empty though.

kevinherron:
and `firstOrNull` that doesn't throw?

elizarov:
Add some use-case, please

LouisCAD:
I made a library to return scan results from the Bluetooth Low Energy API of Android, and often, the use case is to scan for all devices, or devices responding to a filter list passed to the system, to find one device, the one you want, which effectively, is the first one, or the first one matching a predicate.

I currently do this using a `Channel` that becomes hot as soon as it is returned (scan is started before returning), then results are sent to the channel, and scan is stopped when the channel is stopped.

I would like to try to use `flow` as an alternative, and see what is a better fit between `Flow` and `Channel`, keeping both if they each have valid use cases, in line with the semantics of `Flow` and `Channel`.

Here's the relevant lines of code for the channel based API: https://github.com/Beepiz/BleScanCoroutines/blob/5904bc44089e722145f987d653f1c77f479b9af2/core/src/main/java/com/beepiz/bluetooth/scancoroutines/experimental/BluetoothLeScanner.kt#L18-L29

LouisCAD:
I'm not used to reactive streams at all though. Maybe @JakeWharton or others can bring some use cases, or tell if that doesn't make sense or is a recipe for undesirable use of reactive streams.

JakeWharton:
Not sure I've used it much except for in tests.

If you're doing something with a nested stream (like `concat(memoryCache, diskCache, network)`) you can always just `flatMap` a `take(1)` instead of `map`-ing `first()` to merge it back in an enclosing stream. But aside from that and maybe bridging with some legacy system that isn't reactive, the need for something like that should be relatively minor.

PaulWoitaschek:
> Add some use-case, please

I have a function where I need to fetch the next new element receied from a ConflatedBroadcastChannel. So I use `.drop(1).first()`.

```kotlin
private object NullSurrogate

private object FirstException : CancellationException("First received")

suspend fun <T> Flow<T>.first(): T {
  var result: Any? = NullSurrogate
  try {
    collect { value ->
      result = value
      throw FirstException
    }
  } catch (ignored: FirstException) {
  }

[...]

PR #1796, opened 2020-02-11
Flow firstOrNull support

bradynpoulsen:
Adds implementations for `Flow`:
- `suspend fun firstOrNull(): T?`
- `suspend fun firstOrNull(predicate: suspend (T) -> Boolean): T?`

It is not clear why they were intentionally left out in the original proposal (#1078) but introduced with `single`/`singleOrNull`

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
+++ b/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
@@ -142,7 +142,7 @@ public suspend fun <T> Flow<T>.firstOrNull(): T? {
  * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
  * Returns [null] if the flow did not contain an element matching the [predicate].
  */
-public suspend fun <T> Flow<T>.firstOrNull(predicate: suspend (T) -> Boolean): T? {
+public suspend fun <T : Any> Flow<T>.firstOrNull(predicate: suspend (T) -> Boolean): T? {
     var result: Any? = NULL
     try {
         collect { value ->
--- a/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
+++ b/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
@@ -125,7 +125,7 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
  * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
  * Returns [null] if the flow was empty.
  */
-public suspend fun <T> Flow<T>.firstOrNull(): T? {
+public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
     var result: Any? = NULL
     try {
         collect { value ->
--- a/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
+++ b/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
@@ -120,3 +120,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
     if (result === NULL) throw NoSuchElementException("Expected at least one element matching the predicate $predicate")
     return result as T
 }
+
+/**
+ * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
+ * Returns [null] if the flow was empty.
+ */
+public suspend fun <T> Flow<T>.firstOrNull(): T? {
+    var result: Any? = NULL
+    try {
+        collect { value ->
+            result = value
+            throw AbortFlowException(NopCollector)
+        }
+    } catch (e: AbortFlowException) {
+        // Do nothing
+    }
+    return result.takeUnless { it == NULL } as T?
+}
+
+/**
+ * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
+ * Returns [null] if the flow did not contain an element matching the [predicate].
+ */
+public suspend fun <T> Flow<T>.firstOrNull(predicate: suspend (T) -> Boolean): T? {
+    var result: Any? = NULL
+    try {
+        collect { value ->
+            if (predicate(value)) {
+                result = value
+                throw AbortFlowException(NopCollector)
+            }
+        }
+    } catch (e: AbortFlowException) {
+        // Do nothing
+    }
+    return result.takeUnless { it == NULL } as T?
+}
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -900,6 +900,8 @@ public final class kotlinx/coroutines/flow/FlowKt {
 	public static final fun filterNotNull (Lkotlinx/coroutines/flow/Flow;)Lkotlinx/coroutines/flow/Flow;
 	public static final fun first (Lkotlinx/coroutines/flow/Flow;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun first (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function2;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun firstOrNull (Lkotlinx/coroutines/flow/Flow;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun firstOrNull (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function2;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun flatMap (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/Flow;
 	public static final fun flatMapConcat (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/Flow;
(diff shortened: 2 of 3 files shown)


kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +120,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
...
+        // Do nothing
+    }
+    return result.takeUnless { it == NULL } as T?
+}
+
+/**
+ * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
+ * Returns [null] if the flow did not contain an element matching the [predicate].
+ */
+public suspend fun <T> Flow<T>.firstOrNull(predicate: suspend (T) -> Boolean): T? {
```
LouisCAD: Could be inline, with crossinline lambda, just like most Flow operators.

bradynpoulsen: Super odd that wasn't done with the existing `first` operator as well

bradynpoulsen: Looks like I won't be able to inline this because it uses the internal `NULL` and `AbortFlowException` (which is why `first` isn't inlined either)

elizarov:
What's your use-case for this function? Where does it show up?

dave08:
My use case is I have a Flow that could emit 0 or 1 elements matching a certain criteria, `singleOrNull { }` will crash if there's 0 and `first { }` too.

elizarov:
@dave08 I'm not sure I understand. Could you just use `.filter { ... }.singleOrNull()` ? Are you really looking for `singleOrNull { ... }` (a variant with prediacate, just like available for standard collections) to cover your use-case?

dave08:
That throws if there's more than one element that matches `filter { }` I need only the first one if there is, and if not, just `null`.

LouisCAD:
@elizarov `firstOrNull` returns null only if the flow completes and is empty. Otherwise, it returns the first value, and cancels the upstream flow. I currently see no combination of operators that can conveniently reproduce the behavior or `firstOrNull`.

elizarov:
@LouisCAD Indeed, that difference between `firstOrNull` and `singleOrNull` is that the former cancels the flow after receives the first element, while the latter receives more at throws an exception if there's more than one. And that's why I'm puzzled. @dave08 said that this PR addresses a use-case of a flow that is known to have 0 or 1 elements. So why this PR is adding `firstOrNull` instead of `singleOrNull`?

fvasco:
It isn't too sexy, but it works.

```kotlin
suspend fun <T> Flow<T>.firstOrNull() = take(1).fold(null as T?) { _, v -> v }
```

A generic version

```kotlin
suspend fun <T> Flow<T>.getOrNull(index: Int) = take(index + 1).fold(null as T?) { _, v -> v }
suspend fun <T> Flow<T>.firstOrNull() = getOrNull(0)
```

LouisCAD:
@elizarov I understand why you're puzzled now. I think `singleOrNull` is a different use case, with semantics that are more complex because of implications: "single" implies assertion, but "null" implies the contrary. That's not straightforward if you add a predicate, it's no longer just 0 or 1 value or throw.

I think it is out of the scope of this PR that is for `firstOrNull`, which has simpler semantics (accepts any flow, never throws by itself, no kind of contradiction in the wording).

If `singleOrNull { ... }` is needed, people should probably open an issue about their use case first before possibly adding it if it makes sense.

elizarov:
I'm not just puzzled. I'm looking for use-cases. This PR proposes to add `firstOrNull`, but I don't have a use-case for it on the table. All I have for now is a use-case for `singleOrNull`.

LouisCAD:
@bradynpoulsen Can you give yours?

RBusarow:
I've been using a no-arg version of `firstOrNull` for a while now.  

With Android Bluetooth, we can only set the callback once (upon connection), so all responses are returned as a stream (Flow) and we need to parse them out.  Hence our use for `firstOrNull()`.

We're writing some data over Bluetooth (discover services, write a descriptor, write a characteristic), then observing a Flow for the response.  Packets can be dropped or something can happen on the peripheral's end, so there's no guarantee of actually getting that response.  In `send`, we suspend until we get a value or the coroutine is cancelled upstream.  In the case of writing to the `Characteristic`, it's wrapped in a timeout.

``` kotlin
interface GattFlows : BluetoothConnectionElement {
  ...
  fun characteristicWrite(uuid: UUID): Flow<CharacteristicWrite>
  ...
}
```

``` kotlin
suspend fun sendCharacteristic(
  characteristic: Characteristic,
  data: ByteArray
) = withTimeoutOrNull(TIMEOUT) {

  send(gattFlows.characteristicWrite(characteristic.uuid)) {
    characteristic.value = data
    gatt?.writeCharacteristic(characteristic)
  }
}

internal suspend inline fun <E> send(
  flow: Flow<GattEvent<E>>, crossinline writeBlock: () -> Unit = {}
): E? = withContext(executor) {

  flow.onStart { writeBlock() }
    .firstOrNull()                             // < -- Here's the use-case
    ?.let { response ->
      if (response.isSuccessful) {
        response.value
      } else null
    }
}
```

_Coincidentally, it's the same function as the use-case as for #1758, but for a different reason._

dave08:
My original use case was to listen for the first finished `WorkInfo` in Android `WorkManager`'s `getWorkInfosByTagLiveData().asFlow()`, I don't want it to crash if by mistake there's another one following the first that is also finished, but I don't know if it will for some reason complete the flow without any finished results, I just don't need the extra check whether it's single, but I also don't want it to crash on none like `first { }`.

I guess my case is really just for prevention of unnecessary crashing (which might not really be a concern, but why take that chance?) when not knowing a third party api well enough... I could technically use a try/catch for that though it might be less clear that I'm only really interested in 0 or 1 emissions (even though there could be more) that match the criteria.

LouisCAD:
@dave08 Using `first` is safe in your case as the LiveData never ends, making the Flow never complete.

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +120,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
     if (result === NULL) throw NoSuchElementException("Expected at least one element matching the predicate $predicate")
     return result as T
 }
+
+/**
+ * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
+ * Returns [null] if the flow was empty.
+ */
+public suspend fun <T> Flow<T>.firstOrNull(): T? {
```
mhernand40: Looking at `singleOrNull()`, it takes in a `T : Any`. Wouldn't we want to do the same as well for `firstOrNull()`? Otherwise, given a `Flow<Foo?>`, there would be ambiguity between a `null` emission and a completion without any emissions.

bradynpoulsen: I was wondering about that. It would be ambiguous if you needed to distinguish between a null element and a null fallback, but it comes down to semantics on whether you actually care. If we look at the [`collections`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first-or-null.html) and [`sequences`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/first-or-null.html) APIs, there is no `T : Any` requirement since there may be a totally valid reason you don't care how you got `null`.

To be consistent, I'm also fine with `T : Any` purely for consistency with existing `Flow` APIs.

elizarov: We've discussed it today and had decided to leave `xxxOrNull` on Kotlin Flow as extensions to `<T : Any>`. Note, that this is not consistent with stdlib (where you can use `Any?` in similar operators) and it was made so because of pre-v1.0 reports that it is not convenient. We don't know if the same would apply to Flow, so leaving it as `<T : Any>` for now is the conservative decision here. If we find use-cases where it causes inconvenience it is always safe to wide the upper bound to `Any?` in future releases.

bradynpoulsen: Perfect, I'll apply that update now

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +120,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
...
+    var result: Any? = NULL
+    try {
+        collect { value ->
+            result = value
+            throw AbortFlowException(NopCollector)
+        }
+    } catch (e: AbortFlowException) {
+        // Do nothing
+    }
+    return result.takeUnless { it == NULL } as T?
```
elizarov: Must use `===` here. It would be also really helpful if a test that catches this bug is added. See how it is tested for in other places (search for `BadClass` in code).

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +120,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
...
+        collect { value ->
+            if (predicate(value)) {
+                result = value
+                throw AbortFlowException(NopCollector)
+            }
+        }
+    } catch (e: AbortFlowException) {
+        // Do nothing
+    }
+    return result.takeUnless { it == NULL } as T?
```
elizarov: Same here.

qwwdfsad:
Superseded by #1869

PR #1798, opened 2020-02-12
Introduce missing backpressure drop Flow operator

kevincianfarini:
This is a PR to introduce the lacking [RX Java onBackPressureDrop()](http://reactivex.io/RxJava/javadoc/rx/Observable.html#onBackpressureDrop--) Flow analogue. 

drop diagram

Although I would have liked to call this operator `drop` I think that name is both ambiguous and already taken. Therefore, I've named it `discard`. I'm open to name suggestions.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Context.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Context.kt
@@ -209,6 +209,21 @@ public fun <T> Flow<T>.flowOn(context: CoroutineContext): Flow<T> {
 }
 
 /**
+ * Returns a flow which checks cancellation status on each emission and throws
+ * the corresponding cancellation cause if flow collector was cancelled.
+ * Note that [flow] builder is [cancellable] by default.
+ */
+public fun <T> Flow<T>.cancellable(): Flow<T> {
+    if (this is AbstractFlow<*>) return this // Fast-path, already cancellable
+    return unsafeFlow {
+        collect {
+            currentCoroutineContext().ensureActive()
+            emit(it)
+        }
+    }
+}
+
+/**
  * The operator that changes the context where all transformations applied to the given flow within a [builder] are executed.
  * This operator is context preserving and does not affect the context of the preceding and subsequent operations.
  *
@@ -256,7 +271,7 @@ public fun <T, R> Flow<T>.flowWith(
          * All builders are written using scoping and no global coroutines are launched, so it is safe not to provide explicit Job.
          * It is also necessary not to mess with cancellation if multiple flowWith are used.
          */
-        val originalContext = coroutineContext.minusKey(Job)
+        val originalContext = currentCoroutineContext().minusKey(Job)
         val prepared = source.flowOn(originalContext).buffer(bufferSize)
         builder(prepared).flowOn(flowContext).buffer(bufferSize).collect { value ->
             return@collect emit(value)
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 **/.idea/*
 !/.idea/icon.png
 !/.idea/vcs.xml
+!/.idea/copyright
+!/.idea/codeStyleSettings.xml
+!/.idea/codeStyles
 *.iml
 .gradle
 .gradletasknamecache
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectCodeStyleSettingsManager">
+    <option name="PER_PROJECT_SETTINGS">
+      <value />
+    </option>
+  </component>
+</project>
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,15 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JetCodeStyleSettings>
+      <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="1" />
+      <option name="NAME_COUNT_TO_USE_STAR_IMPORT_FOR_MEMBERS" value="1" />
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+    </JetCodeStyleSettings>
+    <codeStyleSettings language="kotlin">
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
+      <indentOptions>
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+      </indentOptions>
+    </codeStyleSettings>
+  </code_scheme>
+</component>
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+  </state>
+</component>
--- a/.idea/copyright/kotlinx_coroutines.xml
+++ b/.idea/copyright/kotlinx_coroutines.xml
@@ -1,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright 2016-&amp;#36;today.year JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license." />
+    <option name="myName" value="kotlinx.coroutines" />
+  </copyright>
+</component>
--- a/.idea/copyright/profiles_settings.xml
+++ b/.idea/copyright/profiles_settings.xml
@@ -1,0 +1,3 @@
+<component name="CopyrightManager">
+  <settings default="kotlinx.coroutines" />
+</component>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.3.7
+
+* Fixed problem that triggered Android Lint failure (#2004).
+* New `Flow.cancellable()` operator for cooperative cancellation (#2026).
+* Emissions from `flow` builder now check cancellation status and are properly cancellable (#2026).
+* New `currentCoroutineContext` function to use unambiguously in the contexts with `CoroutineScope` in receiver position (#2026). 
+* `EXACTLY_ONCE` contract support in coroutine builders.
+* Various documentation improvements.
+
 ## Version 1.3.6
 
 ### Flow
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![official JetBrains project](https://jb.gg/badges/official.svg)](https://confluence.jetbrains.com/display/ALL/JetBrains+on+GitHub)
 [![GitHub license](https://img.shields.io/badge/license-Apache%20License%202.0-blue.svg?style=flat)](https://www.apache.org/licenses/LICENSE-2.0)
-[![Download](https://api.bintray.com/packages/kotlin/kotlinx/kotlinx.coroutines/images/download.svg?version=1.3.6) ](https://bintray.com/kotlin/kotlinx/kotlinx.coroutines/1.3.6)
+[![Download](https://api.bintray.com/packages/kotlin/kotlinx/kotlinx.coroutines/images/download.svg?version=1.3.7) ](https://bintray.com/kotlin/kotlinx/kotlinx.coroutines/1.3.7)
 
 Library support for Kotlin coroutines with [multiplatform](#multiplatform) support.
 This is a companion version for Kotlin `1.3.71` release.
@@ -84,7 +84,7 @@ Add dependencies (you can also add other modules that you need):
 <dependency>
     <groupId>org.jetbrains.kotlinx</groupId>
     <artifactId>kotlinx-coroutines-core</artifactId>
-    <version>1.3.6</version>
+    <version>1.3.7</version>
 </dependency>
 ```
 
@@ -102,7 +102,7 @@ Add dependencies (you can also add other modules that you need):
 
 ```groovy
 dependencies {
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.6'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.7'
 }
 ```
 
@@ -128,7 +128,7 @@ Add dependencies (you can also add other modules that you need):
 
 ```groovy
 dependencies {
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.6")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.7")
 }
 ```
 
@@ -147,7 +147,7 @@ Make sure that you have either `jcenter()` or `mavenCentral()` in the list of re
 Core modules of `kotlinx.coroutines` are also available for 
 [Kotlin/JS](#js) and [Kotlin/Native](#native).
 In common code that should get compiled for different platforms, add dependency to  
-[`kotlinx-coroutines-core-common`](https://search.maven.org/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-core-common/1.3.6/jar)
+[`kotlinx-coroutines-core-common`](https://search.maven.org/artifact/org.jetbrains.kotlinx/kotlinx-coroutines-core-common/1.3.7/jar)
 (follow the link to get the dependency declaration snippet).
 
 ### Android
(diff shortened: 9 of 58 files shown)


elizarov:
But what is the use-case? Why and in what cases would you need it?

kevincianfarini:
The use case I can see is that this is the inverse of `debounce` that had a dynamic timeout period based on the speed of the consumer. 

Given 4 rapid clicks of a button, you could:
* `debounce` with a given timeout. At best, you debounce three of the clicks based on timeout and consume the last element. At worst, the button is clicked in such a way that multiple events are queued up. 
* `discard` extra button clicks. Given the scenario that the button consumer speed varies, setting a static timeout using debounce isn't much help and would be useful to base element drops on the consumer speed.

elizarov:
@kevincianfarini It looks like your use-case of filter rapind button clicks is already covered by `conflate()` operator. Please, take a look: https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/conflate.html

kevincianfarini:
Wouldn't the conflate operator start by processing the first click, and then conflating the remaining three? Then, after the consumer is done processing the first click it would then process the fourth as well. 

Am I misunderstanding how conflate works?

elizarov:
> Wouldn't the conflate operator start by processing the first click, and then conflating the remaining three? Then, after the consumer is done processing the first click it would then process the fourth as well.

Yes. As far as can see exactly the same behavior is shown on the picture for `onBackpressureDrop`, isn't it?

kevincianfarini:
> Yes. As far as can see exactly the same behavior is shown on the picture for onBackpressureDrop, isn't it?

I don't think so. I tried to illustrate this scenario similar to how the RX docs do. 

diagram

In this example, `discard` only processes the first click because all of the other clicks happen during the `consumption duration`. 

Conflate, however, will consume both elements 1 and 4 because the first element has already been emitted and is being consumed. When the consumer is ready for the next element, button click 4 will be emitted since it is cached.

elizarov:
Thanks. I see the difference now. Borrowing example code from `conflate` docs:
```
val flow = (1..30).asFlow().onEach { delay(100) }
println(flow.conflate().onEach { delay(1000) }.toList())
```
Prints: `[1, 10, 20, 30]`, while:
```
println(flow.asFlowable().onBackpressureDrop().asFlow().onEach { delay(1000) }.toList())
```
Prints: `[1, 11, 21]`

Now, the question is: what are the use-cases for `onBackpressureDrop`? In what circumstances you'd prefer its behavior to `conflate()`?

kevincianfarini:
The first practical use case that comes to mind for me is a notification channel. 

It's pretty common that notifications sent via websockets, google play services, etc don't have a payload of the notification itself. Instead they indicate the client that updates are available and to make a call to get them.  

Using `discard` on a flow which represents a notification channel would be more correct than using conflate, since the client wouldn't finish processing the first notification update, and then process another conflated update request which was received while the client was processing the first.

kevincianfarini:
@elizarov awesome! I'll rework this soon. Thanks

kevincianfarini:
I'm working and pushing updates now. Since we're in different time zones, when this gets approved could we wait to merge so I have a chance to squash all of the commits?

kotlinx-coroutines-core/common/src/flow/operators/Context.kt
```diff
@@ -150,6 +150,39 @@ public fun <T> Flow<T>.buffer(capacity: Int = BUFFERED): Flow<T> {
...
+ * another element.
+ *
+ * ### Operator Fusion
+ *
+ * Adjacent applications of `dropWhileBusy`, [conflate]/[buffer], [channelFlow], [flowOn], [produceIn], and [broadcastIn] are
+ * always fused so that only one properly configured channel is used for execution.
+ * **Conflation takes precedence over `buffer()` calls with any other capacity.**
+ */
+@ExperimentalCoroutinesApi
+public fun <T> Flow<T>.dropWhileBusy(): Flow<T> = DiscardFlowOperator(this)
```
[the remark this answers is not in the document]
kevincianfarini: Follow up to this: 

What does it mean for us to fuse this operator with others? It seems like we could fuse this with `buffer`. Then, when the buffer of this channel gets full we would drop elements after. 

What would it mean for this to get fused with conflate? Seems like it shouldn't be a valid operation. Thoughts?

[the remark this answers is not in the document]
kevincianfarini: Follow, follow up. 

I'm implemented operator fusion with `buffer`. It seems to make logical sense. 

I throw an `IllegalArgumentException` is `conflate` is called downstream of `dropWhileBusy` because it doesn't seem like a valid operation. This has been documented.

kevincianfarini:
@elizarov I've made some more updates and left review comments that I'd like your guidance on. :)

kotlinx-coroutines-core/common/src/flow/internal/DiscardingCollector.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.flow.FlowCollector
+
+/**
+ * A [FlowCollector] specifically built to be used for `Flow.dropWhileBusy`
+ *
+ * This collector ensures that the fist element is always sent. It then offers subsequent elements,
+ * allowing them to be dropped.
+ */
+@InternalCoroutinesApi
+internal class DiscardingCollector<T>(private val channel: SendChannel<T>) : FlowCollector<T> {
```
elizarov: I'd suggest to rename it to `DropWhileBusyCollector`.

kotlinx-coroutines-core/common/src/flow/internal/DiscardingCollector.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    private var hasSentFirstValue: Boolean = false
+
+    override suspend fun emit(value: T) {
```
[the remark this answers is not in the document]
elizarov: The idea looks good to me. It seems very appropriate for the way collection from the flow conceptually works, where you write `flow.collect { value -> someCode(value) }` and `someCode` is called (and can become busy) only when the first element is there. However, it can be surprising for cases like `flow.onStart { delay(1000) }.collect { ... }` where it would still wait for the actual collection to start and would not consider the flow to be "busy" for this initial 1000 ms. So, it should be explicitly mentioned in the docs of `dropWhileBusy`, that is give there a more detailed explanation of what "busy" means for this operator.

kotlinx-coroutines-core/common/src/flow/operators/Context.kt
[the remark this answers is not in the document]
kevincianfarini: The minimal strategy seems simple and to make sense. I do have concerns about the full strategy though. 

> It would be more efficient to directly offer into the rendezvous channel and use only two coroutines. That is, it would be more efficient if dropWhileBusy could fuse with the preceding ChannelFlowBuilder. But that means that ChannelFlowBuilder has to provide to its the block the special implementation of ProducerScope interface that makes send to behave like offer.

Seeing that `ProducerScope` has both `offer` and `send`, changing the semantics of send when `dropWhileBusy` is fused seems alarming. Send could secretly be an offer in the right instance, which seems sketchy  to me. 

Thoughts? If we were to go this route, I would want it to be thoroughly documented on both `produce/broadcast` as well as `dropWhileBusy`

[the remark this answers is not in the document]
kevincianfarini: Likewise, to ensure first element delivery the overridden `send` (which is really offer) would need to ensure first element delivery. Should we apply those same constraint when a user calls `offer` within the scope?

elizarov:
Ping!

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -141,3 +152,20 @@ internal open class ProducerCoroutine<E>(
...
+
+    override suspend fun send(element: E) {
+        if (hasSentFirstValue) {
+            offer(element)
+        } else {
+            super.send(element)
+            hasSentFirstValue = true
+        }
+    }
+}
```
elizarov: Performance nit: Using `_channel.offer` and `_channel.send` would be slightly better here.

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -141,3 +152,20 @@ internal open class ProducerCoroutine<E>(
         if (!processed && !handled) handleCoroutineException(context, cause)
     }
 }
+
+private class DropWhileBusyProducerCoroutine<E>(
+    parentContext: CoroutineContext,
+    channel: Channel<E>
+) : ProducerCoroutine<E>(parentContext, channel) {
+
+    private var hasSentFirstValue: Boolean = false
```
elizarov: Now, since this is a channel, the `send` function can be called concurrently, so this variable has to be atomic (`private val hasSentFirstValue = atomic(false)`) and the check & set on it has to be atomic (`hasSentFirstValue.getAndSet(true)`) to make sure that when there are two concurrent calls to `send` it works correctly.

kotlinx-coroutines-core/common/src/flow/operators/Context.kt
```diff
@@ -109,10 +109,15 @@ public fun <T> Flow<T>.buffer(capacity: Int = BUFFERED): Flow<T> {
...
     }
-    return if (this is ChannelFlow)
-        update(capacity = capacity)
-    else
-        ChannelFlowOperatorImpl(this, capacity = capacity)
+
+    return when (this) {
+        is ChannelFlowOperatorImpl -> update(
+            capacity = capacity,
+            dropWhileBusy = if (capacity == CONFLATED) false else dropWhileBusy
```
elizarov: I would rather move the logic of merging `dropWhileBusy` flag into the implementation of `update` function and leave the code in the `buffer` operator as it was before. It will just call `update(capacity = capacity)` leaving a default value of `dropWhileBusy = false` and letting `update` take care of figuring it out.

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -141,3 +153,20 @@ internal open class ProducerCoroutine<E>(
...
+
+    private val hasSentFirstValue = atomic(false)
+
+    override suspend fun send(element: E) {
+        if (hasSentFirstValue.value) {
+            _channel.offer(element)
+        } else {
+            _channel.send(element)
+            hasSentFirstValue.getAndSet(true)
+        }
```
elizarov: There's still a race here. Two threads calling `send` concurrently will both do `send`. You should do `if (hasSentFirstValue.getAndSet(true)) _channel.offer(element) else _channel.send(element)`

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -27,6 +27,17 @@ public abstract class ChannelFlow<T>(
     // buffer capacity between upstream and downstream context
     @JvmField val capacity: Int
 ) : Flow<T> {
+
+    private var shouldDropWhileBusy: Boolean = false
```
elizarov: This should be just a constructor parameter.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -27,6 +27,17 @@ public abstract class ChannelFlow<T>(
...
+
+    internal open fun update(
```
elizarov: There's no need to have a separate 3-argument `update`. You can just 3rd parameter to a regular `update` for simplicity. It is an internal method, so you don't have to preserve it.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -64,11 +75,17 @@ public abstract class ChannelFlow<T>(
     private val produceCapacity: Int
         get() = if (capacity == Channel.OPTIONAL_CHANNEL) Channel.BUFFERED else capacity
 
-    open fun broadcastImpl(scope: CoroutineScope, start: CoroutineStart): BroadcastChannel<T> =
-        scope.broadcast(context, produceCapacity, start, block = collectToFun)
+    open fun broadcastImpl(scope: CoroutineScope, start: CoroutineStart): BroadcastChannel<T> {
+        return scope.broadcast(context, produceCapacity, start, block = collectToFun)
```
elizarov: No need to change these two lines. Can be reverted to the way it was before.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -64,11 +75,17 @@ public abstract class ChannelFlow<T>(
...
+    }
 
-    open fun produceImpl(scope: CoroutineScope): ReceiveChannel<T> =
-        scope.produce(context, produceCapacity, block = collectToFun)
+    open fun produceImpl(scope: CoroutineScope): ReceiveChannel<T> {
+        return if (shouldDropWhileBusy) {
```
elizarov: Style nit: Use `= expr` instead of `{ return expr }` when possible.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -126,13 +143,30 @@ internal abstract class ChannelFlowOperator<S, T>(
...
-    override fun create(context: CoroutineContext, capacity: Int): ChannelFlow<T> =
-        ChannelFlowOperatorImpl(flow, context, capacity)
 
-    override suspend fun flowCollect(collector: FlowCollector<T>) =
-        flow.collect(collector)
+    override fun create(context: CoroutineContext, capacity: Int): ChannelFlow<T> {
+        return ChannelFlowOperatorImpl(
+            flow,
+            context,
+            dropWhileBusy = capacity != Channel.CONFLATED && dropWhileBusy,
```
elizarov: This code properly belongs to `update` function. That is where all trickly "fusion" of parameters is performance. Here is should simply create implementation with the given params. To get there, `create` function shall have `dropWhileBusy` parameter.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -126,13 +143,30 @@ internal abstract class ChannelFlowOperator<S, T>(
 internal class ChannelFlowOperatorImpl<T>(
     flow: Flow<T>,
     context: CoroutineContext = EmptyCoroutineContext,
+    val dropWhileBusy: Boolean = false,
```
elizarov: Please, keep the order of `context`, `capacity`, `dropWhileBusy` parameters consistent between constructors, `create`, and `update` functions. Seems logical to add `dropWhileBusy` as the last one.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
kevincianfarini: The other update fun is not internal, it's public. Should I still coalesce these to a single method?

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
kevincianfarini: In order to do that, should I overload the create function or change the signature in `ChannelFlowOperator`?

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
elizarov: It is public in an internal class. It can be safely updated (put intented).

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
elizarov: You should just add `dropWhileBusy` boolean to `create`. It is all internal code that is not directly available to end-users of API.

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -76,8 +85,11 @@ public abstract class ChannelFlow<T>(
      * handlers, while the pipeline before does not, because it was cancelled during its dispatch.
      * Thus `onCompletion` and `finally` blocks won't be executed and it may lead to a different kinds of memory leaks.
      */
-    open fun produceImpl(scope: CoroutineScope): ReceiveChannel<T> =
+    open fun produceImpl(scope: CoroutineScope): ReceiveChannel<T> = if (shouldDropWhileBusy) {
+        scope.dropWhileBusyProduce(context, start = CoroutineStart.ATOMIC, block = collectToFun)
```
elizarov, line 89: `produceCapacity` parameter is missing in the call to `dropWhileBusyProduce`. It should be affecting the chain of operators like  `dropWhileBusy().buffer(10)` where, logically, elements shall be dropped when the buffer is full.

kotlinx-coroutines-core/common/test/flow/operators/DropWhileBusyTest.kt
```diff
@@ -0,0 +1,90 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.withVirtualTime
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class DropWhileBusyTest : TestBase() {
+
+    @Test
+    fun `elements are dropped during slow consumption`() = withVirtualTime {
```
elizarov, line 13: We don't use back-quotes test names in this project. Please use `testXxx` naming convention in all the tests here. Also, the tests here would be easier to read and review (less scrolling to do) without all the empty lines inside the corresponding test functions.

PR #1801, opened 2020-02-13
Fix a race in some tests for JavaRX integration

dkhalanskyjb:
See the commit message for the rationale.

I looked at the other uses of `withExceptionHandler`, and it seems that all the other executions are deterministic. I have slight doubts about `testExceptionAfterCancellation`, but, to my understanding, `blockingSubscribe` ensures that every execution is finished.

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/test/LeakedExceptionTest.kt
+++ b/reactive/kotlinx-coroutines-rx2/test/LeakedExceptionTest.kt
@@ -27,8 +27,8 @@ class LeakedExceptionTest : TestBase() {
             runBlocking {
                 repeat(10000) {
                     combine(flow, flow) { _, _ -> Unit }
-                            .catch {}
-                            .collect { }
+                        .catch {}
+                        .collect {}
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


reactive/kotlinx-coroutines-rx2/test/LeakedExceptionTest.kt
```diff
@@ -22,37 +22,84 @@ class LeakedExceptionTest : TestBase() {
...
-                combine(flow, flow) { _, _ -> Unit }
-                    .catch {}
-                    .collect { }
+        withFixedThreadPool(4) { dispatcher ->
+            val flow = rxSingle<Unit>(dispatcher) { throw TestException() }.toFlowable().asFlow()
+            runBlocking {
+                repeat(10000) {
+                    combine(flow, flow) { _, _ -> Unit }
+                            .catch {}
+                            .collect { }
```
elizarov: Style nit: Continuation lines should be formatted with 4 spaces of indentation (not 8).
Also, it as we are touching this code anyway, it would be create to fix `{ }` -> `{}` (remove space)

reactive/kotlinx-coroutines-rx2/test/LeakedExceptionTest.kt
```diff
@@ -22,37 +22,84 @@ class LeakedExceptionTest : TestBase() {
...
+
+    /**
+     * Run in a thread pool, then wait for all the tasks to finish.
+     */
+    private fun withFixedThreadPool(numberOfThreads: Int, block: (CoroutineDispatcher) -> Unit) {
+        val pool = Executors.newFixedThreadPool(numberOfThreads)
+        val dispatcher = pool.asCoroutineDispatcher()
+        block(dispatcher)
+        pool.shutdown()
+        while (!pool.awaitTermination(10, TimeUnit.SECONDS));
```
elizarov: when you write `while` with empty body write `{ /* deliberately empty */ }` (with comment!) as its body.

dkhalanskyjb: Fixed.

reactive/kotlinx-coroutines-rx2/test/LeakedExceptionTest.kt
dkhalanskyjb: Fixed.

PR #1803, opened 2020-02-13
Introduce ExecutorRule for stress-tests for simpler development process

qwwdfsad:
It is a non-goal to replace all existing executor usages in stress tests, replaced just a few for the demonstration

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/ExecutorRule.kt
+++ b/kotlinx-coroutines-core/jvm/test/ExecutorRule.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+import org.junit.rules.*
+import org.junit.runner.*
+import org.junit.runners.model.*
+import java.util.concurrent.*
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/jvm/test/ExecutorRule.kt
```diff
@@ -0,0 +1,45 @@
...
+    private var _executor: ExecutorCoroutineDispatcher? = null
+    override val executor: Executor
+        get() = _executor?.executor ?: error("Executor is not initialized")
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                val threadPrefix = description.className.substringAfterLast(".") + "#" + description.methodName
+                _executor = newFixedThreadPoolContext(numberOfThreads, threadPrefix)
+                ignoreLostThreads(threadPrefix)
```
elizarov: Why `ignoreLostThreads`? They should not be lost.

qwwdfsad: Because `TestBase.onCompletion` marked with `@After` is invoked during the rule evaluation, not after (=> first check for lost threads and the close executor in the rule) and there is no easy way to fix

PR #1811, opened 2020-02-16
Experimenta kotlin.time.Duration support

### LINKED ISSUE #1402: kotlin.time.Duration overloads for delay in common

fvasco:
Initial proposal, no test included.
Fixes #1402

### DIFF
--- a/kotlinx-coroutines-core/common/src/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/Delay.kt
@@ -99,9 +99,5 @@ internal val CoroutineContext.delay: Delay get() = get(ContinuationInterceptor) 
  */
 @ExperimentalTime
 internal fun Duration.toDelayMillis(): Long =
-        when {
-            this > Duration.ZERO -> toLongMilliseconds().coerceAtLeast(1)
-            this < Duration.ZERO -> -1
-            // 0, -0, NaN
-            else -> 0
-        }
+        if (this > Duration.ZERO) toLongMilliseconds().coerceAtLeast(1)
+        else 0
--- a/kotlinx-coroutines-core/common/src/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/Delay.kt
@@ -6,6 +6,8 @@ package kotlinx.coroutines
 
 import kotlinx.coroutines.selects.*
 import kotlin.coroutines.*
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
 
 /**
  * This dispatcher _feature_ is implemented by [CoroutineDispatcher] implementations that natively support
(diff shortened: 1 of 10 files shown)


elizarov:
Look good. Needs tests. Also, these functions should not be `inline`. There's no much gain from inlining here, plus it would limit further evolution. For example, we could make them nanosecond-precise in the future (if anyone needs it) without recompilation or deprecated long-accepting version when Duration becomes stable without having references to it in the bytecode.

fvasco:
Hi @elizarov,
I agree with you.
I marked these functions `inline` due to its experimental nature. I will remove the modifier.

Should I put all tests in the `DelayTest` class, should I duplicate the `DelayTest` class, should I create two different ones with a common abstract superclass? Any suggestion?

elizarov:
I'd duplicate `DelayTest` into something like `DelayDurationTest`.

fvasco:
Done.

EdwarDDay:
Do you also want to add `kotlin.time.Duration` support to the corresponding `Flow` operators (`sample` and `debounce`)?

fvasco:
Good idea

kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -75,5 +77,31 @@ public suspend fun delay(timeMillis: Long) {
...
+
+/**
+ * Convert this duration to its millisecond value.
+ * Positive durations are coerced at least `1`.
+ */
+@ExperimentalTime
+internal fun Duration.toDelayMillis(): Long =
+        when {
+            this > Duration.ZERO -> toLongMilliseconds().coerceAtLeast(1)
+            this < Duration.ZERO -> -1
```
elizarov: I've reviewed usages of `toDelayMillis()` and could not find a case which cares about the distinction between zero and a negative value (either both mean the same or both are not supported), so this branch of `when` seems redundant (negative values can be also mapped to zero).

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -71,6 +73,34 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+/**
+ * Returns a flow that mirrors the original flow, but filters out values
+ * that are followed by the newer values within the given [timeout].
+ * The latest value is always emitted.
+ *
+ * Example:
+ * ```
+ * flow {
+ *     emit(1)
+ *     delay(90.0.milliseconds)
```
whyoleg: why not to use extensions on `Int` instead of `Double`, like `90.milliseconds` instead of `90.0.milliseconds` ?

kotlinx-coroutines-core/common/src/Delay.kt
fvasco: Ok, test passes

elizarov:
Merged with #1864

PR #1820, opened 2020-02-20
Make publication validator part of the main gradle project

dkhalanskyjb:
Before, publication validator was just lying nearby, in these
files, and built and run separately. Now, it is compiled
alongside the main project (in the `test` gradle configuration), so
its breakage will be detected during normal usage. Still, it is
only run when a special gradle property, `DeployVersion`, is
defined.

Merging this requires changing the `Dependency validation` build configuration on TeamCity to only consist of a single step that calls `gradle -PdryRun=true testPublishing`.


publication-validator/build.gradle
```diff
@@ -2,34 +2,33 @@
...
 }
 
-compileKotlin {
-    kotlinOptions.jvmTarget = "1.8"
-}
 compileTestKotlin {
     kotlinOptions.jvmTarget = "1.8"
 }
+
+task testPublishing(type: Test) {
```
qwwdfsad: Could you please elaborate on why this task does not depend on the root `build` task?

dkhalanskyjb: The intention was that running `testPublishing` does the same as what is currently being done in `Dependency validation` configuration on TeamCity:
* `--info -PdryRun=true publishNpm publishToMavenLocal -x dokka` in the root project;
* `build` in the project of the validator.
So, if there was a reason to depend on `build`, it wasn't reflected in the build process before.

If we don't appeal to how it's currently done, this still seems reasonable to me: to test publication, we don't really need to know that the stress tests pass, for example; we only need to see the to-be-published packages themselves, and explicitly only depending on them being prepared for publishing conveys this intention more clearly.

dkhalanskyjb: I changed the configuration so that `build` depends on this validation.

PR #1832, opened 2020-02-27
Enable JS IR backend

skuzmich:


gradle/compile-js-multiplatform.gradle
```diff
@@ -27,26 +31,30 @@ compileKotlinJs {
     kotlinOptions.metaInfo = true
     kotlinOptions.sourceMap = true
     kotlinOptions.moduleKind = 'umd'
-
-    kotlinOptions {
-        // drop -js suffix from outputFile
-        def baseName = project.name - "-js"
-        outputFile = new File(outputFile.parent, baseName + ".js")
```
skuzmich: :warning: This could break compatibility with JS. Looking how to resolve this issue on plugin side.

skuzmich: Resolved. Used manual test runner for classic JS backend.

gradle.properties
```diff
@@ -43,3 +44,6 @@ kotlin.native.ignoreDisabledTargets=true
 
 # Site deneration
 jekyll_version=4.0
+
+# JS IR baceknd sometimes crashes with out-of-memory
```
qwwdfsad: Please create an issue for that and link it here so we can keep track of it and remove this option as soon as it's fixed

gradle/compile-js-multiplatform.gradle
```diff
@@ -6,12 +6,20 @@ apply from: rootProject.file('gradle/node-js.gradle')
 
 kotlin {
     targets {
-        fromPreset(presets.js, 'js')
+        fromPreset(presets.js, 'js') {
+            // Enable built-in test runner only for IR target.
+            // These runners don't support changing js module name change.
+            if (js.hasProperty("irTarget")) {
+                irTarget.browser()
+                irTarget.nodejs()
```
qwwdfsad: `mocha`/`jsdom` target is not supported?

gradle/compile-js-multiplatform.gradle
```diff
@@ -6,12 +6,20 @@ apply from: rootProject.file('gradle/node-js.gradle')
...
+            }
+        }
     }
 
     sourceSets {
         jsMain.dependencies {
             api "org.jetbrains.kotlin:kotlin-stdlib-js:$kotlin_version"
+            api "org.jetbrains.kotlinx:atomicfu-js:$atomicfu_version"
```
qwwdfsad: dependency should be conditional and present only for IR backend

gradle/compile-js-multiplatform.gradle
```diff
@@ -39,14 +47,25 @@ compileTestKotlinJs {
     kotlinOptions.metaInfo = true
     kotlinOptions.sourceMap = true
     kotlinOptions.moduleKind = 'umd'
+    kotlinOptions.freeCompilerArgs += "-XXLanguage:-NewInference"
+}
+
+if (project.tasks.findByName('compileKotlinJsIr')) {
+    compileKotlinJsIr {
+        kotlinOptions.freeCompilerArgs += "-XXLanguage:-NewInference"
```
qwwdfsad: Please elaborate why NI is disabled

js/example-frontend-js/build.gradle
```diff
@@ -5,6 +5,24 @@
 apply plugin: 'kotlin-dce-js'
 apply from: rootProject.file('gradle/node-js.gradle')
 
+// Workaround resolving new Gradle metadata with kotlin2js
+// TODO: Remove once plugin is fixed
```
qwwdfsad: ...and how do we know that the plugin is fixed?

gradle/compile-js-multiplatform.gradle
skuzmich: nodejs runs mocha. `jsdom` is not enabled yet, true. Would you need jsdom tests if browser tests will be enabled?

qwwdfsad: I wouldn't if it worked like a charm on both teamcity and my local machine.
Please elaborate why these tests are enabled conditionally, while in atomicfu they are enabled unconditionally

gradle/compile-js-multiplatform.gradle
qwwdfsad: Same question regarding atomicfu

gradle/publish-bintray.gradle
```diff
@@ -89,8 +89,8 @@ publishing {
                 break
         }
 
-        // disable metadata everywhere, but in native modules
-        if (type == 'maven' || type == 'metadata' || type == 'jvm' || type == 'js') {
+        // disable metadata everywhere, but in native and js modules
+        if (type == 'maven' || type == 'metadata' || type == 'jvm') {
```
qwwdfsad: Isn't it going to break our releases with Kotlin 1.3.70? Do we really have to publish metadata if IR BE is not here?

gradle/test-mocha-js.gradle
```diff
@@ -19,7 +19,9 @@ task testMochaNode(type: NodeTask, dependsOn: [compileTestKotlinJs, installDepen
     if (project.hasProperty("teamcity")) args += ['--reporter', 'mocha-teamcity-reporter']
 }
 
-jsTest.dependsOn testMochaNode
+def legacyjsTestTask = project.tasks.findByName('legacyjsTest') ? legacyjsTest : jsTest
+
+legacyjsTestTask.dependsOn testMochaNode
```
qwwdfsad: same question regarding inconsistency with atomicfu

gradle/publish-bintray.gradle
skuzmich: > Do we really have to publish metadata if IR BE is not here?
We don't

> Isn't it going to break our releases with Kotlin 1.3.70?
I assumed adding metadata wouldn't break stuff. @ilgonmic please take a look

ilgonmic: Since Gradle 5.3 it can consume metadata for dependency from repo. So for this version range it should not break release.
But this check can be covered with plus one check on 1.4 release, and not publish metadata for 1.3.70 releases

gradle/compile-js-multiplatform.gradle
skuzmich: Apparently its not easy to enable browser tests on TC. Local browser tests pass for both backends. Currently `jsdom` tests are is disabled in IR backend only.

gradle.properties
skuzmich: Done

gradle/compile-js-multiplatform.gradle
skuzmich: Reverted. It used to fail but now new inference is fixed for JS IR and disabled for classic backend.

js/example-frontend-js/build.gradle
skuzmich: Created issue

gradle/test-mocha-js.gradle
skuzmich: Built-in test runner does not support custom `outputFile` compiler options. These options are used only in legacy backend. So legacy backend tests are running as previously, using custom `testMochaNode` task.

gradle/compile-js-multiplatform.gradle
skuzmich: It might be hard to enable it conditionally in this build script. @ilgonmic can you help?

ilgonmic: It seems that it should be not dependency in js source set but specific ir target compilation
```
js().irTarget?.compilations['main']?.dependencies {
   api "org.jetbrains.kotlinx:atomicfu-js:$atomicfu_version"
}

```

PR #1834, opened 2020-02-28
java.time adapter for Flow.debounce and Flow.sample

fvasco:
I get trouble problem with tests

```
No tests found for given includes: [time.FlowDebounceTest](filter.includeTestsMatching)
```

However it should be ok (I hope :)


elizarov:
Fixed API dump and merged, see #1863

PR #1841, opened 2020-03-04
Reuse RxJava built-in disposed Disposable

JakeWharton:


reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
```diff
@@ -100,12 +100,7 @@ public fun <T: Any> ObservableSource<T>.asFlow(): Flow<T> = callbackFlow {
     }
 
     subscribe(observer)
-    awaitClose { disposableRef.getAndSet(Disposed)?.dispose() }
-}
-
-private object Disposed : Disposable {
-    override fun isDisposed() = true
-    override fun dispose() = Unit
+    awaitClose { disposableRef.getAndSet(Disposables.disposed())?.dispose() }
```
langara: I treated this `private object Disposed` as a private "token" similarly to `DisposableHelper.DISPOSED` in Rx itself (they have the warning there "don't leak it"), but here I didn't do `compareAndSet(Disposed, ...)` because comparing to initial `null` was enough, so I think your change is correct.

JakeWharton: Yep. Just trying to save APK size anywhere I can by reusing existing stuff.

PR #1848, opened 2020-03-06
Introduce repeatWhen operator

linean:
I've added new flow operator `repeatWhen` which behaves the same as one from RxJava
http://reactivex.io/RxJava/javadoc/rx/Observable.html#repeatWhen-rx.functions.Func1-

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
@@ -117,3 +117,24 @@ public fun <T> Flow<T>.scanReduce(operation: suspend (accumulator: T, value: T) 
         emit(accumulator as T)
     }
 }
+
+/**
+ * Re-subscribes to the current flow (`this`) each time the [other] flow emits a new item.
+ * The previous flow is immediately cancelled.
+ *
+ * For example:
+ * ```
+ * val repeatFlow = flow {
+ *      emit("Repeat")
+ * }
+ *
+ * flowOf(1, 2, 3, 4).repeatWhen(repeatFlow).toList()
+ * ```
+ * will produce `[1, 2, 3, 4, 1, 2, 3, 4]`
+ */
+public fun <T> Flow<T>.repeatWhen(other: Flow<*>) = flow {
+    emit(this@repeatWhen)
+    other.collect {
+        emit(this@repeatWhen)
+    }
+}.flatMapLatest { it }
--- a/kotlinx-coroutines-core/common/test/flow/operators/RepeatWhenTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/RepeatWhenTest.kt
@@ -1,0 +1,69 @@
+package kotlinx.coroutines.flow.operators
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class RepeatWhenTest : TestBase() {
+
+    @Test
+    fun testRepeatWhen() = runTest {
+        val repeatFlow = flow {
+            emit("Repeat")
+        }
+
+        val result = flowOf(1, 2, 3, 4).repeatWhen(repeatFlow).toList()
+        assertEquals(listOf(1, 2, 3, 4, 1, 2, 3, 4), result)
+    }
+
+    @Test
+    fun testEmptyFlow() = runTest {
+        val repeatFlow = flow {
+            emit("Repeat")
+        }
+
+        val result = emptyFlow<Unit>().repeatWhen(repeatFlow).toList()
+        assertEquals(emptyList(), result)
+    }
+
+    @Test
+    fun testCancelPreviousFlow() = runTest {
+        val repeatFlow = flow {
+            emit("Repeat")
+        }
+
+        val flow = flow {
+            emit(1)
+            emit(2)
+            yield()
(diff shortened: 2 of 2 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
```diff
@@ -117,3 +117,24 @@ public fun <T> Flow<T>.scanReduce(operation: suspend (accumulator: T, value: T)
...
+ *
+ * flowOf(1, 2, 3, 4).repeatWhen(repeatFlow).toList()
+ * ```
+ * will produce `[1, 2, 3, 4, 1, 2, 3, 4]`
+ */
+public fun <T> Flow<T>.repeatWhen(other: Flow<*>) = flow {
+    emit(this@repeatWhen)
+    other.collect {
+        emit(this@repeatWhen)
+    }
```
zach-klippenstein: This could just be `other.map { this@repeatWhen }`.

elizarov:
Please, explain what are your use-cases for this operator. In what situation have you encountered the need to have it? What problem were you solving?

linean:
**My use case:**
I have a repository which provides items from a local database. I perform a transformation which is based on the local date for each item.

**My problem:**
I need to emit new items each time there is a date change.

**Solution:**
```
fun getItems(): Flow<List<Item>>() {
    return database
        .getItems() // Flow from Room database
        .map { it.performDateBasedTransformation() }
        .repeatWhen(dateChanged)
}
```
For `dateChanged` flow I use `ConflatedBroadcastChannel` which properly bridges system events.

kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
linean: Which basically does the same under the hood. :see_no_evil:

fvasco:
It looks similar to:

```
dateChanged.flatMap {
  database.getItems()
    .map { it.performDateBasedTransformation() }
}
```

linean:
> It looks similar to:
> 
> ```
> dateChanged.flatMap {
>   database.getItems()
>     .map { it.performDateBasedTransformation() }
> }
> ```

It looks similar, but important difference is that here you need to wait for `dateChanged` to emit first value.

fvasco:
so this request permits to avoid a [onStrart](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/on-start.html) invocation in your use case

linean:
Basically yes. Also I think that adding `repeatWhen` operator should make migration from RxJava much easier.

elizarov:
Sorry, but I'm a bit lost here. According to Rx's documentation on `repeatWhen` it only checks the provided predicate when the source stream completes. So, let us take a look at the following code:
```
fun getItems(): Flow<List<Item>>() {
    return database
        .getItems() // Flow from Room database
        .map { it.performDateBasedTransformation() }
        .repeatWhen(dateChanged)
}
```

Accoriding to the docs it is going to work like this:

1. Get items from database
2. Transform them with `performDateBasedTransformation`
3. When the items from DB are over, it will check if date has changed and repeat. If the data has not changed by that time, it will just complete without ever trying again. 

I don't understand how it addresses your use-case. Please, elaborate.

linean:
Solution I've proposed works a little bit different than one from RxJava.
Here `repeatWhen` is not waiting for the flow to finish but instead it will cancel current one and resubscribe to the source again. 

In example I've posted emissions from DB flow never ends (DB is listening for changes).

elizarov:
That confuses me even more. I don't see any point in naming it like Rx's `repeatWhen` operator when it does quite a different thing.

Moreover, I think this kind of feature is already available via `flatMapLatest` operator, assuming that we have a flow of `currentDate` which emits new values as date changes:

```
currentDate.flatMapLatest { date -> // current date
    database
        .getItems() // Flow from Room database
        .map { it.performDateBasedTransformation(date) }
}
```

Here,  `getItems` request will be canceled & restated when date changes. We also explicitly pass date to `performDateBasedTransformation` which makes the code easier to understand since dependency on the date is now explicit.

linean:
You are right, but it'll work only if we can provide initial value for parent flow `(currentDate)`. If we want to listen for some other events, we would need to use onStart as @fvasco mentioned.

Then it would look like that:
```
someOtherEvent
    .onStart {
        database
            .getItems() // Flow from Room database
            .map { it.performDateBasedTransformation() }
    }
    .flatMapLatest {
        database
            .getItems() // Flow from Room database
            .map { it.performDateBasedTransformation() }
    }
```

elizarov:
I'm lost. If we don't have a flow of dates then how do we can know when date changes and react to this change?

linean:
This is getting more and more confusing :face_with_head_bandage: 
Code that I've initially posted was just an example, which as we can see was pretty bad, because you found a better way to solve it.

Goal of this operator was to have possibility to restart flow when some external event occurs.

elizarov:
`flatMapLatest` _is_ an operator that restarts a flow when an external event occurs. It does not seem we have a rationale so far to add more way to do it.

fvasco:
My suggestion was:

>  but important difference is that here you need to wait for dateChanged to emit first item

`dateChanged.onStart { emit(...someInitialValue...) }`

elizarov:
A flow of dates should always emit the current element when you collect it. That makes the rest of it clicks.

PR #1865, opened 2020-03-13
Readme improvement

coroutineDispatcher:
There was a small mistake in the Readme of kotlinx-coroutines-test. A lateinit variable evaluated to null at the beginning.


elizarov:
Thanks.

PR #1869, opened 2020-03-16
Flow firstOrNull support [fixes]

qwwdfsad:


kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +119,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
...
+        collect { value ->
+            if (predicate(value)) {
+                result = value
+                throw AbortFlowException(NopCollector)
+            }
+        }
+    } catch (e: AbortFlowException) {
+        // Do nothing
+    }
+    return result.takeUnless { it === null } as T?
```
bradynpoulsen: The `takeUnless` isn't needed here.

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +119,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
...
+    var result: Any? = null
+    try {
+        collect { value ->
+            result = value
+            throw AbortFlowException(NopCollector)
+        }
+    } catch (e: AbortFlowException) {
+        // Do nothing
+    }
+    return result.takeUnless { it === null } as T?
```
bradynpoulsen: `takeUnless` isn't needed here as well.

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -120,3 +119,39 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
     if (result === NULL) throw NoSuchElementException("Expected at least one element matching the predicate $predicate")
     return result as T
 }
+
+/**
+ * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
+ * Returns `null` if the flow was empty.
+ */
+public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
+    var result: Any? = null
```
elizarov: It can be declared as `var result: T? = null` and will not need unchecked cast on return. Same of version with predicate.

Issue #856, opened 2018-11-22
DefaultExecutor can never exit on IDLE


cy6erGn0m:
See `DefaultExecutor.run`. According to it's source, if there are no tasks then `processNextEvent()` returns `Long.MAX_VALUE`. Then we check `shutdownNanos` and if not yet initialized then we check it again for the same condition (that is always true) and compute `parkNanos` time. After parking for `KEEP_ALIVE_NANOS` we respin here again but we are falling into the second case where `parkNanos = parkNanos.coerceAtMost(KEEP_ALIVE_NANOS)` is executed and the same after one more respin so we will never exit but spin every second instead.

Also it looks like we need to reset `shutdownNanos` if there was a task scheduled but we don't do it.

mirjalal:
Any fix for this?

PR #1876, opened 2020-03-23
Fix DefaultExecutor not being able to exit

dkhalanskyjb:
Solves https://github.com/Kotlin/kotlinx.coroutines/issues/856.

### DIFF
--- a/kotlinx-coroutines-core/common/src/EventLoop.common.kt
+++ b/kotlinx-coroutines-core/common/src/EventLoop.common.kt
@@ -287,6 +287,11 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
         return dequeue()
     }
 
+    override fun processNextEvent(): Long {
+        dequeueNextTask()?.run()
+        return nextTime
+    }
+
     public final override fun dispatch(context: CoroutineContext, block: Runnable) = enqueue(block)
 
     public fun enqueue(task: Runnable) {
--- a/kotlinx-coroutines-core/common/src/EventLoop.common.kt
+++ b/kotlinx-coroutines-core/common/src/EventLoop.common.kt
@@ -23,8 +23,6 @@ import kotlin.native.concurrent.*
 internal abstract class EventLoop : CoroutineDispatcher() {
     /**
      * Counts the number of nested `runBlocking` and [Dispatchers.Unconfined] that use this event loop.
-     * There are two 32-bit counters encoded in this 64-bit value, allowing to count [Dispatchers.Unconfined]
-     * separately from `runBlocking`; see [delta] and its uses.
      */
     private var useCount = 0L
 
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -64,11 +67,17 @@ internal abstract class EventLoop : CoroutineDispatcher() {
...
         task.run()
         return true
     }
+
+    public fun popUnconfinedTask(): DispatchedTask<*>? =
+        unconfinedQueue?.removeFirstOrNull()
+
+    protected open fun popNextTask(onlyUnconfined: Boolean = false): Runnable? =
+        popUnconfinedTask()
+
```
elizarov: Naming nit: `pop` verb is primarily used with stacks (`push`/`pop`), rarely so with queues, especially in JVM & Kotlin ecosystem. I'd suggest using `poll` (ala Java's Queue interface) for a short verb these operations or you can go with long and explicit `removeFirstXxxOrNull`.

kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -249,9 +258,15 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
         }
     }
 
-    override fun processNextEvent(): Long {
+    override fun popNextTask(onlyUnconfined: Boolean): Runnable? {
```
elizarov: I cannot find a place where this method if called with `onlyUnconfined = true`.

dkhalanskyjb:
If we do this optimization, then the decomposition of `process` into (poorly named) `pop` and `run` is not needed: I introduced it only so that `DefaultExecutor` could discern between a situation where tasks arrive slower than they are executed (and `process` always returns `Long.MAX_VALUE`) and one where there are no tasks at all (and `process` always returns `Long.MAX_VALUE`).

Should I keep the decomposition or should I revert it and instead implement the small change to the logic of `process` that is also a performance optimization?

elizarov:
@dkhalanskyjb 
> If we do this optimization, then the decomposition of `process` into (poorly named) `pop` and `run` is not needed
It would be also great to avoid decomposition. It will make the code easier to understand.

kotlinx-coroutines-core/jvm/test/DefaultExecutorStressTest.kt
```diff
@@ -35,4 +36,30 @@ class DefaultExecutorStressTest : TestBase() {
         }
         finish(2 + iterations * 4)
     }
+
+    @Test
+    fun workerShutsDown() = withVirtualTimeSource {
```
elizarov: By convention in this project, test function shall start with the verb `test` like `testWorkerShutsDown`.

dkhalanskyjb: Yeah, sorry. I was already told that once, so I'll be sure to remember it.

PR #1877, opened 2020-03-23
Enable strict explicit mode

qwwdfsad:


gradle/compile-common.gradle
```diff
@@ -12,3 +12,8 @@ kotlin.sourceSets {
         api "org.jetbrains.kotlin:kotlin-test-annotations-common:$kotlin_version"
     }
 }
+
+
+kotlin.sourceSets.matching({ it.name.contains("Main") }).all { srcSet ->
+    project.ext.set("kotlin.mpp.freeCompilerArgsForSourceSet.${srcSet.name}", "-Xexplicit-api=strict")
+}
```
elizarov: Instead of `compile-common.gradle` and `compile-jvm.gradle` you can add it here together with all other Kotlin compilation options:
https://github.com/Kotlin/kotlinx.coroutines/blob/3bb3e55a53359ebe268c7def40c6b016ca0e5f9a/build.gradle#L179-L192

qwwdfsad: This is not going to work:
1) Explicit mode will be applied for test source sets
2) IDE integration will be broken for some reason

PR #1885, opened 2020-03-30
Fix leak of subscription reference in closed ArrayBroadcastChannel

elizarov:
When ArrayBroadcastChannel was closed it was still retaining a reference to its subscription (even if that subscription was cancelled) while, in fact, either closing a broadcast channel or cancelling subscription should remove the reference.

This is no problem with ConflatedBroadcastChannel, but it is added to the test for completeness.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/ArrayBroadcastChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/ArrayBroadcastChannel.kt
@@ -218,13 +218,15 @@ internal class ArrayBroadcastChannel<E>(
         override val isBufferAlwaysFull: Boolean get() = error("Should not be used")
         override val isBufferFull: Boolean get() = error("Should not be used")
 
-        override fun onCancelIdempotent(wasClosed: Boolean) {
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/channels/ArrayBroadcastChannel.kt
```diff
@@ -218,13 +218,15 @@ internal class ArrayBroadcastChannel<E>(
         override val isBufferAlwaysFull: Boolean get() = error("Should not be used")
         override val isBufferFull: Boolean get() = error("Should not be used")
 
-        override fun onCancelIdempotent(wasClosed: Boolean) {
+        override fun close(cause: Throwable?): Boolean {
```
qwwdfsad: It looks like `onCancelIdempotent` should be overridden (and do nothing) to avoid an extra bit of work.

elizarov: `ArrayBroadcastChannel.Subscriber` extends `AbstractChannel` and `onCancelIdemptotent` does not do anything there.

Issue #871, opened 2018-12-03
Fix inconsistencies in the documentation on exception handling and cancellation


mtopolnik:
These are a few quotes from [exception-handling.html](https://kotlinlang.org/docs/reference/coroutines/exception-handling.html) in the official docs:

(1)
>Coroutine builders come in two flavors: propagating exceptions automatically (`launch` and `actor`) or exposing them to users (`async` and `produce`). The former treat exceptions as unhandled, similar to Java's `Thread.uncaughtExceptionHandler`, while the latter are relying on the user to consume the final exception, for example via `await` or `receive`

(2)
>If a coroutine encounters exception other than `CancellationException`, it cancels its parent with that exception. This behaviour cannot be overridden and is used to provide stable coroutines hierarchies for structured concurrency which do not depend on `CoroutineExceptionHandler` implementation.

(3)
>`CoroutineExceptionHandler` is invoked only on exceptions which are not expected to be handled by the user, so registering it in `async` builder and the like of it has no effect.

These sentences convey a confusing message. 

(1) seems to claim that `launch` and `async` are different and `async` won't propagate the exceptions, relying on the user to receive them in an `await()`. 

(2) seems to contradict that and claim that exceptions always cancel their coroutine and propagate to the parent.

(3) seems to once again make a difference between `launch` and `async`.

The actual behavior is that all builders propagate the exception to their parent, so the claim in (1) is incorrect. That in turn means that all exceptions are equally treated as "unhandled", which defeats the rationale given in (3). Yet the actual behavior is as stated, `async` ignores exception handlers.

ylemoigne:
Ok, this is related to a question I asked today on coroutine slack channel.

The questions is : "why In 'uncaughtExceptionHandler' ...  is called ?"

```kotlin
import kotlinx.coroutines.*
import kotlinx.coroutines.channels.produce

// From : https://kotlinlang.org/docs/reference/coroutines/exception-handling.html
//
// Coroutine builders come in two flavors: propagating exceptions automatically (launch and actor) or
// exposing them to users (async and produce).
//
// The former treat exceptions as unhandled, similar to Java's Thread.uncaughtExceptionHandler,
// while the latter are relying on the user to consume the final exception,
// for example via await or receive (produce and receive are covered later in Channels section).
fun main() {
    val job = GlobalScope.launch(CoroutineExceptionHandler { _, t -> println("In 'uncaughtExceptionHandler' $t") }) {
        try {

[...]

PR #1886, opened 2020-03-30
Improve docs for CoroutineExceptionHandler

### LINKED ISSUE #1746: CoroutineExceptionHandler: Issues with Documentation, Possible Bug in runBlocking
Given this code:

```
val job = Job()

val scope = CoroutineScope(job)

val handler = CoroutineExceptionHandler { coroCtx, e ->
    println("Handler caught $e")
}

fun CoroutineScope.handlerTest() = this.launch(handler) {
    throw Exception("Launched coroutine failed")
}
```

When I run it like this (using the scope constructed above):

    scope.handlerTest()

I get the expected output:

```
Handler caught java.lang.Exception: Launched coroutine failed
```

However, when I run it like this (using the `runBlocking` scope):

    runBlocking { this.handlerTest() }

the handler doesn't observe the exception and I get the default handler's output instead:

```
Exception in thread "main" java.lang.Exception: Launched coroutine failed
	at org.mtopol.TestingKt$handlerTest$1.invokeSuspend(testing.kt:14)
    ...
	at org.mtopol.TestingKt.main(testing.kt:18)
	at org.mtopol.TestingKt.main(testing.kt)
```

Even when I add the handler to `runBlocking`:

    runBlocking(handler) { this.handlerTest() }

it still crashes without the handler observing the exception.

The behavior is not specific to `runBlocking`, I can write the code as follows:

    GlobalScope.launch { this.handlerTest() }

and I still get just a crash (now inside the `Default` thread pool).

Finally, if I write it as follows (adding the handler to the top-level `launch`):

    GlobalScope.launch(handler) { this.handlerTest() }

this time it gets handled.

After studying the documentation on `CoroutineExceptionHandler` in detail, I found the above behaviors surprising. The documentation explicitly specifies only what happens when a coroutine exception handler is _not_ installed, and I could only indirectly guess at what it was supposed to do when there is one. Yet my guesses were all wrong. Paraphrasing, the docs seem to claim the following:

1. Never handle a `CancellationException`.
2. If there is a `Job` in the context, invoke `Job.cancel`.
3. *Otherwise* pass the exception to the exception handler.

The actual behavior is more like the following:

1. Always invoke `coroutineContext[Job]!!.cancel()` (since there is always a `Job` in any coroutine's context)
2. To decide whether to pass the exception to the handler, follow these rules:
    a. Never handle a `CancellationException`.
    b. If the current coroutine has a parent coroutine, don't handle the exception. it was passed to the parent via `job.cancel()`.
    c. Otherwise pass the exception to the exception handler.

Furthermore, the behavior with `runBlocking` seems like a bug to fix.

elizarov:
Fixes #1746

### DIFF
--- a/docs/exception-handling.md
+++ b/docs/exception-handling.md
@@ -247,7 +247,7 @@ CoroutineExceptionHandler got java.lang.ArithmeticException
 ### Exceptions aggregation
 
 When multiple children of a coroutine fail with an exception the
-general rule is "the first exception wins", so the first exception gets handed.
+general rule is "the first exception wins", so the first exception gets handled.
 All additional exceptions that happen after the first one are attached to the first exception as suppressed ones. 
 
 <!--- INCLUDE
--- a/docs/exception-handling.md
+++ b/docs/exception-handling.md
@@ -80,8 +80,11 @@ Caught ArithmeticException
 
 What if one does not want to print all exceptions to the console?
 [CoroutineExceptionHandler] context element on a _root_ coroutine can be used as generic `catch` block for
-this root coroutine and all its children where custom logging or exception handling may take place.
+this root coroutine and all its children where custom exception handling may take place.
 It is similar to [`Thread.uncaughtExceptionHandler`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler)).
+You cannot recover from the exception in the `CoroutineExceptionHandler`. The coroutine had already completed
+with the corresponding exception when the handler is called. Normally, the handler is used to
+log the exception, show some kind of error message, terminate, and/or restart the application.
 
 On JVM it is possible to redefine global exception handler for all coroutines by registering [CoroutineExceptionHandler] via
 [`ServiceLoader`](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html).
--- a/docs/exception-handling.md
+++ b/docs/exception-handling.md
@@ -18,22 +18,22 @@
 
 ## Exception Handling
 
-
 This section covers exception handling and cancellation on exceptions.
-We already know that cancelled coroutine throws [CancellationException] in suspension points and that it 
-is ignored by coroutines machinery. But what happens if an exception is thrown during cancellation or multiple children of the same 
-coroutine throw an exception?
+We already know that cancelled coroutine throws [CancellationException] in suspension points and that it
+is ignored by the coroutines' machinery. Here we look at what happens if an exception is thrown during cancellation or multiple children of the same
+coroutine throw an exception.
 
 ### Exception propagation
 
-Coroutine builders come in two flavors: propagating exceptions automatically ([launch] and [actor]) or 
+Coroutine builders come in two flavors: propagating exceptions automatically ([launch] and [actor]) or
 exposing them to users ([async] and [produce]).
-The former treat exceptions as unhandled, similar to Java's `Thread.uncaughtExceptionHandler`, 
-while the latter are relying on the user to consume the final 
+When these builders are used to create a _root_ coroutine, that is not a _child_ of another coroutine,
+the former builder treat exceptions as **uncaught** exceptions, similar to Java's `Thread.uncaughtExceptionHandler`,
+while the latter are relying on the user to consume the final
 exception, for example via [await][Deferred.await] or [receive][ReceiveChannel.receive] 
 ([produce] and [receive][ReceiveChannel.receive] are covered later in [Channels](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/channels.md) section).
 
-It can be demonstrated by a simple example that creates coroutines in the [GlobalScope]:
+It can be demonstrated by a simple example that creates root coroutines using the [GlobalScope]:
 
 <div class="sample" markdown="1" theme="idea" data-highlight-only>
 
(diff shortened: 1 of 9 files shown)


kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -52,11 +52,32 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
+ * All _children_ coroutines (coroutines created in the context of another [Job]) delegate handling of their
+ * exceptions to their parent coroutine, which also delegates to the parent, and so on until the root,
+ * so the `CoroutineExceptionHandler` installed in their context is never used.
  * A coroutine that was created using [async][CoroutineScope.async] always catches all its exceptions and represents them
- * in the resulting [Deferred] object.
+ * in the resulting [Deferred] object, so it cannot result in uncaught exceptions either.
+ *
+ * ### Handling coroutine exceptions
+ *
+ * `CoroutineExceptionHandler` is a last-resort mechanism for global "catch all" behavior.
```
mtopolnik: One thing I would make clear here and in `exception-handling.md` is that the coroutine has already failed when this handler runs. There is often confusion about the terms "catch" and "handle" here, they usually imply recovery from the exception. There are many questions on Stack Overflow about this aspect.

elizarov: I've added additional clarification on that.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -52,11 +52,37 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
+ * You cannot recover from the exception in the `CoroutineExceptionHandler`. The coroutine had already completed
+ * with the corresponding exception when the handler is called. Normally, the handler is used to
+ * log the exception, show some kind of error message, terminate, and/or restart the application.
+ *
+ * If you need to handle exception in a specific part of the code, it is recommended to use `try`/`catch` around
+ * the corresponding code inside your coroutine. This way you can you prevent completion of the coroutine
```
EdwarDDay: Small Typo
```suggestion
 * the corresponding code inside your coroutine. This way you can prevent completion of the coroutine
```

elizarov:
Another go. Also fixes #871

mtopolnik:
I've read through these docs again, I couldn't find any statement about catching `CancellationException`. You can catch a `CancellationException` and if you don't take care to rethrow it, it prevents the coroutine from completing. This does conform to my expectation from Java, but I could also imagine Kotlin generating such bytecode that always catches and rethrows a `CancellationException`. So I had to go and try it to know for sure.

We have statements like

> it is ignored by the coroutines' machinery

>these exceptions are ignored by all handlers

Lacking an explicit statement about the catch-blocks, the user can be easily misled.

docs/exception-handling.md
```diff
@@ -227,22 +240,15 @@ The output of this code is:
...
-But that may cause lost exceptions, for example if coroutine throws an exception in its `finally` block.
-So, additional exceptions are suppressed. 
-
-> One of the solutions would have been to report each exception separately, 
-but then [Deferred.await] should have had the same mechanism to avoid behavioural inconsistency and this 
-would cause implementation details of a coroutines (whether it had delegated parts of its work to its children or not)
-to leak to its exception handler.
-
+When multiple children of a coroutine fail with an exception the
+general rule is "the first exception wins", so the first exception gets handed.
```
qwwdfsad, line 250: gets handed or gets handled?

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -52,11 +52,37 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
     }
 
 /**
- * An optional element in the coroutine context to handle uncaught exceptions.
+ * An optional element in the coroutine context to handle **uncaught** exceptions.
  *
- * Normally, uncaught exceptions can only result from coroutines created using the [launch][CoroutineScope.launch] builder.
+ * Normally, uncaught exceptions can only result from _root_ coroutines created using the [launch][CoroutineScope.launch] builder.
+ * All _children_ coroutines (coroutines created in the context of another [Job]) delegate handling of their
```
qwwdfsad, line 58: excepts ones whose parent is a supervisor job. I think it is a good idea to mention here, Kdoc will be read much more frequently than our guide

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -52,11 +52,37 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
+ * scope.launch { // launch child coroutine in a scope
+ *     try {
+ *          // do something
+ *     } catch (e: Throwable) {
+ *          // handle exception
+ *     }
+ * }
+ * ```
+ *
+ * ### Implementation details
```
qwwdfsad, line 85: While we are here, please mention that handler can be invoked from an arbitrary thread (opposed to "one the root coroutine was launched" as it may seem)

elizarov:
@mtopolnik 
> I've read through these docs again, I couldn't find any statement about catching `CancellationException`.

I'm not sure where would be the right place (one that users will notice) to add additional details on cancellation exception. It does not seem related to the docs for `CoroutineExceptionHandler`, though, so I'm merging this doc improvements as is.

PR #1892, opened 2020-03-31
Removed extra non working SharedImmutable annotations

LepilkinaElena:


elizarov:
Can you, please, comment on why the are non-working and can be removed?

LepilkinaElena:
Due to documentation https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.native.concurrent/-shared-immutable/ `@SharedImmutable` is applicable only to top level objects/variables. It was just ignored before, now we added FE checkers and compiler fails.

PR #1896, opened 2020-04-01
Implement `mapForEach` Flow operator

PHELAT:
This `mapForEach` operator helps coroutines users avoid nested mapping function, for mapping items of an Iterable.
#### Current situation:
```kotlin
flow {
    emit(listOf(1, 2, 3))
}.map { items ->
    items.map {
        it.toString()
    }
}
```
#### With `mapForEach` operator:
```kotlin
flow {
    emit(listOf(1, 2, 3))
}.mapForEach {
    it.toString()
}
```

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
@@ -57,6 +57,14 @@ public inline fun <T, R: Any> Flow<T>.mapNotNull(crossinline transform: suspend 
 }
 
 /**
+ * Returns a flow containing a list of the results of applying the given [transform] function to each value of the original flow
+ */
+public inline fun <T, R> Flow<Iterable<T>>.mapForEach(crossinline transform: suspend (value: T) -> R): Flow<List<R>> = transform { value ->
+    val transformed = value.map { transform(it) }
+    return@transform emit(transformed)
+}
+
+/**
  * Returns a flow that wraps each element into [IndexedValue], containing value and its index (starting from zero).
  */
 @ExperimentalCoroutinesApi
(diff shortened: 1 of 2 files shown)


fvasco:
Your "Current situation" example does not compile, the function `Int.map` does not exist.
What is your use case?

PHELAT:
@fvasco Sorry about that, I've updated the example.

elizarov:
Can you, please, elaborate what was the original problem you were trying to solve?

PHELAT:
@elizarov Sure, this operator is kinda inspired from `flatMapIterable` operator of RxJava but just for the use-case of mapping the items of that Iterable. To achieve this in RxJava we can use the combination of `flatMapIterable` + `map` + `toList`:
```kotlin
observable.flatMapIterable { it }
    .map { it.toString }
    .toList()
```
So for a specific use-case, let's say that we have a data source class with a single `read(): Flow<List<MyEntity>>` function which queries a [Room](https://developer.android.com/topic/libraries/architecture/room) database and then returns it's result, but the thing is that we don't want to expose our Room entity class through this function and to fix this, we can map our list:
```kotlin
roomDao.query()
    .map { roomEntities: List<RoomEntity> ->
        roomEntities.map {
             MyEntity(...)
        }
    }
```
The code above works fine but it's nested two levels and it's kinda hard to read. so that's why I've wrote this operator to help with this:
```kotlin
roomDao.query()
    .mapForEach { roomEntity: RoomEntity ->
        MyEntity(...)
    }
```

elizarov:
Thanks a lot for the explanation of your use-case. Our goal is to provide a minimal set of out-of-the-box operators that meet wide set of needs or are especially tricky to implement. None of that seems to be applicable here. I don't see any problem with implementing this operator in your project when you need it.

PR #1898, opened 2020-04-01
Add linearizability tests for Mutex and Semaphore and fix them

### LINKED ISSUE #1737: Test linearizability of Semaphore and Mutex
Now that lincheck property supports suspending functions please add test for linearizability of Semaphore and Mutex

ndkoval:
Fixes #1737


ndkoval:
After adding the corresponding tests, I figured out that the `Semaphore` implementation is non-linearizable

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -121,76 +121,80 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
     }
 
     override suspend fun acquire() {
-        val p = _availablePermits.getAndDecrement()
-        if (p > 0) return // permit acquired
-        addToQueueAndSuspend()
+        while (true) {
+            val p = _availablePermits.getAndDecrement()
+            if (p > 0) return // permit acquired
+            if (addToQueueAndSuspend()) return
```
elizarov, line 127: This is not a tail-call call anymore. It should be refactored using fast-path/slow-path patter (where fast-path does not suspend and calls slow-path suspending method as a tail-call). See how it is done in channels, for example.

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -210,19 +219,23 @@ private class SemaphoreSegment(id: Long, prev: SemaphoreSegment?, pointers: Int)
...
         onSlotCleaned()
-        return cancelled
     }
 
     override fun toString() = "SemaphoreSegment[id=$id, hashCode=${hashCode()}]"
 }
-
 @SharedImmutable
-private val RESUMED = Symbol("RESUMED")
+private val MAX_SPIN_CYCLES = systemProp("kotlinx.coroutines.semaphore.maxSpinCycles", 100_000)
```
elizarov, line 232: The dafault value looks very, very big here:
* If the caller lost time slot its too small
* If the caller just between incrementing the segment & casing, it too big.
I suggest a default of 100

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -123,74 +123,88 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
...
+        // it is important to make `suspend` function invocations in a way
+        // so that the tail-call optimization can be applied.
+        acquireSlowPath()
     }
 
-    override fun release() {
-        val p = incPermits()
-        if (p >= 0) return // no waiters
-        resumeNextFromQueue()
+    private suspend fun acquireSlowPath() {
```
elizarov, line 132: While this is a "slow path" it does not mean it should be inefficient. There is a way to improve a similar way to how channels are implemented. Please, call `suspendAtomicCancellableCoroutineReusable` in this function and pass the corresponding continuation to the `addToQueue` function (that would not be suspending anymore). This way, this Semaphore class wolud not allocate a state-machine object even in its slow-path.

PR #1905, opened 2020-04-06
Extract debugger

qwwdfsad:


kotlinx-coroutines-debug/src/internal/Attach.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import net.bytebuddy.*
+import net.bytebuddy.agent.*
+import net.bytebuddy.dynamic.loading.*
+
+/*
+ * These classes are used reflectively from kotlinx-coroutines-core when this module is present in the classpath.
+ * It is a substitute for service loading
+ */
+internal class ByteBuddyDynamicAttach : Function0<Unit> {
```
elizarov, line 15: Separate classes for attach&detach look quite excessive. It also forces an extra code in the `core` module (a separate `getFunction`, lots of string class names, etc). You can have a single `class DynamicAttach : (Boolean) -> Unit` with a `Boolean` parameter that means `attach` or `detach` operation. Should make the code smaller.

kotlinx-coroutines-debug/src/DebugProbes.kt
```diff
@@ -114,7 +121,7 @@ public object DebugProbes {
      * Returns all existing coroutines info.
      * The resulting collection represents a consistent snapshot of all existing coroutines at the moment of invocation.
      */
-    public fun dumpCoroutinesInfo(): List<CoroutineInfo> = DebugProbesImpl.dumpCoroutinesInfo()
+    public fun dumpCoroutinesInfo(): List<CoroutineInfo> = DebugProbesImpl.dumpCoroutinesInfo().map { CoroutineInfo(it) }
```
elizarov, line 124: The difference between:
* `dumpDebuggerInfo` -> returning `List<DebufferInfo>` and
* `dumpCoroutinesInfo` -> returning `List<DebugCoroutineInfo>`
it too subtle and the names do not help to figure out which does what.

Do we really need both?

kotlinx-coroutines-core/jvm/src/debug/AgentPremain.kt
```diff
@@ -0,0 +1,65 @@
...
+            val name = className.replace("/", ".")
+            if (name != "kotlin.coroutines.jvm.internal.DebugProbesKt") {
+               return null
+            }
+            /*
+             * DebugProbesKt.bin contains `kotlin.coroutines.jvm.internal.DebugProbesKt` class
+             * with method bodies that delegate all calls directly to their counterparts in
+             * kotlinx.coroutines.debug.DebugProbesImpl. This is done to avoid classfile patching
+             * on the fly (-> get rid of ASM dependency)
+             */
```
elizarov, line 45: The `DebugProbesKt.bin` file should be generated during build of the project (as opposed to being committed as a byte blob).

kotlinx-coroutines-debug/src/DebugProbes.kt
qwwdfsad: `dumpCoroutinesInfo` is publci a user-facing API.
`dumpDebuggerInfo` is an internal method for IDEA plugin that should return serializable list. I will properly document and test the latter as soon as I know that it fits IDEA needs

elizarov: Maybe we should make `dumpDebuggerInfo` internal then? (with clean `@JvmName`, of course)

kotlinx-coroutines-core/jvm/src/debug/AgentPremain.kt
qwwdfsad: Do we really want to? It's a lot of burden to maintain, more Gradle machinery, some risk of accidentally including this classfile in the JAR (not as a resource), using unofficial flag `-Xallow-kotlin-package` etc. And benefits are really unclear

elizarov: Two reasons:
* Transparency: Just an opaque `.bin` file looks spooky.
* Maintenance: What if we need to change it in the future (move Impl to different package or rename, etc) and the secret on how to produce this file is lost.

kotlinx-coroutines-debug/src/DebugProbes.kt
qwwdfsad: `dumpDebuggerInfo` lies in `DebugProbesImpl` which is already internal

elizarov: It is public in a  `public object DebugProbes`.

elizarov: I see. I was just confused by the names, then.

qwwdfsad:
JFYI: `javap -v DebugProbesKt.class | grep major` (javap can't process `.bin` files)

kotlinx-coroutines-core/jvm/src/debug/AgentPremain.kt
qwwdfsad: Let's compromise on the middle ground :)

integration-testing/build.gradle
```diff
@@ -62,19 +74,28 @@ task debugAgentTest(type: Test) {
...
+    coreAgentTestCompile project(':kotlinx-coroutines-core')
 }
 
 compileTestKotlin {
     kotlinOptions.jvmTarget = "1.8"
 }
 
 test {
-    dependsOn([npmTest, mavenTest, debugAgentTest])
+    dependsOn([npmTest, mavenTest, debugAgentTest, coreAgentTest])
```
elizarov, line 100: It has just occurred to me that we are wrong in adding all the dependencies to the `test` task as it makes it impossible to ask for `gradlew test` meaning "run all the regular tests only". We should be adding them all to the `check` task.

integration-testing/src/debugAgentTest/kotlin/PrecompiledDebugProbesTest.kt
```diff
@@ -0,0 +1,35 @@
...
+        val className: String = clz.getName()
+        val classFileResourcePath = className.replace(".", "/") + ".class"
+        val stream = clz.classLoader.getResourceAsStream(classFileResourcePath)!!
+        val array = stream.readBytes()
+        val binFile = clz.classLoader.getResourceAsStream("DebugProbesKt.bin")!!
+        val binContent = binFile.readBytes()
+        assertTrue(array.contentEquals(binContent))
+    }
+
+    private fun diffpos(a: ByteArray, b: ByteArray, typeLenght: Int): Int {
```
elizarov, line 25: `diffpos` is not used (`array.contentEquals(binContent)` is used).

integration-testing/src/debugAgentTest/kotlin/PrecompiledDebugProbesTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    @Test
+    fun testClassFileContent() {
+        val clz = Class.forName("kotlin.coroutines.jvm.internal.DebugProbesKt")
+        val className: String = clz.getName()
+        val classFileResourcePath = className.replace(".", "/") + ".class"
+        val stream = clz.classLoader.getResourceAsStream(classFileResourcePath)!!
+        val array = stream.readBytes()
+        val binFile = clz.classLoader.getResourceAsStream("DebugProbesKt.bin")!!
+        val binContent = binFile.readBytes()
+        assertTrue(array.contentEquals(binContent))
```
elizarov, line 22: It should include some helpful message like `Blah-blah binary does not match, please copy this file to there`. Note that this will have to be done with every update of kotlin compiler version, since the compiler version is stored in the compiled metadata, so this process needs some automation (some kind of "overwrite" switch).

kotlinx-coroutines-core/jvm/src/debug/AgentPremain.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    private fun installSignalHandler() {
+        try {
+            Signal.handle(Signal("TRAP")) { // kill -5
+                if (DebugProbesImpl.isInstalled) {
+                    // Case with 'isInstalled' changed between this check-and-act is not considered
+                    // a real debug probes use-case, thus is not guarded against.
+                    DebugProbesImpl.dumpCoroutines(System.out)
+                } else {
+                    println("""Cannot perform coroutines dump, debug probes are disabled""")
```
elizarov, line 59: Tripple quotes here seems to be some refactoring leftover.

kotlinx-coroutines-core/jvm/src/debug/AgentPremain.kt
```diff
@@ -0,0 +1,70 @@
...
+    internal object DebugProbesTransformer : ClassFileTransformer {
+        override fun transform(
+            loader: ClassLoader,
+            className: String,
+            classBeingRedefined: Class<*>?,
+            protectionDomain: ProtectionDomain,
+            classfileBuffer: ByteArray?
+        ): ByteArray? {
+            val name = className.replace("/", ".")
+            if (name != "kotlin.coroutines.jvm.internal.DebugProbesKt") {
```
elizarov, line 40: Performance nit: You can don't need to call `replace` and the whole `val name` here. You can directly compare `className` to `"kotlin/coroutines/jvm/internal/DebugProbesKt"`

qwwdfsad:
Asked Volodya to verify this branch against IDEA, will merge after his approve

Issue #540, opened 2018-09-09
Integrate ticker channels into structured concurrency and flows


elizarov:
We need to integrate ticker channels into a structured concurrency story to make sure they are not lost running forever. They should be somehow marked as "daemon" (?) children of their parent and should not prevent the parent completion, but shall be cancelled when all other (non-daemon?) children coroutines complete. This would make standard timing-related operations easy to write in a less error-prone way.

P.S. This is not critical for 1.0 release, since the API for channels is going to be kept experimental anyway.

wyaeld:
@elizarov is this coming? Currently trying to build coroutine-based background processing but having a time-keeping channel will simplify a number of things.

elizarov:
@wyaeld Yes. Ticker channels will become a part of `Flow` framework. Yet, this is not going to happen in the short-term future. Actually, flows makes it so easy to write this kind of things yourself, as to almost remove the need for any kind of library primitive. 

Can you, please, write more details on what kind of background processing you are doing and what are your use-cases for time-keeping channels.

wyaeld:
Sure.  I'm reading through all the Flow docs this week to understand how
that works, since that was my assumption on how you would do it.
In my case I'm writing at IoT control element, which is continually
processing commands that get delivered in batches, and for which I want an
observer/reporting type element that is able to report say every 10s, on
the status of everything in flight.  This command batches can be hundreds
of individual messages that have to go out.  The TickerChannel seemed like
a useful trigger for the reporting element.

elizarov:
@wyaeld For your use-case we plan something like a specialized time-based `chunked` primitive. See #1302

esdudnik:
Will ticker channels be multiplatform? I came from this issue
https://github.com/Kotlin/kotlinx.coroutines/issues/1186
Looking for something similar for multiplatform library.

elizarov:
Yes. The plan is to have them flow-based an multiplatform.

MarcelReiter:
Are there any updates?

We'd like to use this feature in an app displaying a one-time-password for two-step-verification similar to the [google authenticator](https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2)

Usually, OTPs are valid 30 seconds, and "start" on a full minute or on a half minute,
e.g. 12:00:00 or 12:00:30, but not 12:00:01

In the moment we're using rx-java and emit:
- on creation
[...]

PR #1908, opened 2020-04-06
Flow based Ticker implemented.

ibrahimyilmaz:
This pull request contains basic implementation of the ticker with Flow.

I have seen this issue(https://github.com/Kotlin/kotlinx.coroutines/issues/540) maybe this can contribute to that issue(plan)

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
+++ b/kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
@@ -16,7 +16,7 @@ import kotlin.concurrent.schedule
  * @param period period between each element in milliseconds.
  * @param initialDelay delay after which the first element will be produced (it is equal to [period] by default) in milliseconds.
  */
-public fun ticker(
+public fun tickerFlow(
         period: Long,
         initialDelay: Long = period
 ): Flow<Unit> = callbackFlow {
--- a/kotlinx-coroutines-core/jvm/src/flow/TickerFlows.kt
+++ b/kotlinx-coroutines-core/jvm/src/flow/TickerFlows.kt
@@ -1,0 +1,33 @@
+package kotlinx.coroutines.flow
+
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.channels.awaitClose
+import java.util.*
+import kotlin.concurrent.schedule
+
+/**
+ * Creates a flow that produces the first item after the given initial delay and subsequent items with the
+ * given delay between them.
+ *
+ * The resulting flow is a callback flow, which basically listens @see [Timer.schedule]
+ *
+ * This Flow stops producing elements immediately after [Job.cancel] invocation.
+ *
+ * @param period period between each element in milliseconds.
+ * @param initialDelay delay after which the first element will be produced (it is equal to [period] by default) in milliseconds.
+ */
+public fun ticker(
+        period: Long,
+        initialDelay: Long = period
+): Flow<Unit> = callbackFlow {
+    require(period > 0)
+    require(initialDelay > -1)
+
+    val timer = Timer()
+    timer.schedule(initialDelay, period) {
+        offer(Unit)
+    }
+
+    awaitClose { timer.cancel() }
+}
+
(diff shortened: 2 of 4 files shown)


kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
```diff
@@ -0,0 +1,33 @@
...
+ * The resulting flow is a callback flow, which basically listens @see [Timer.schedule]
+ *
+ * This Flow stops producing elements immediately after [Job.cancel] invocation.
+ *
+ * @param period period between each element in milliseconds.
+ * @param initialDelay delay after which the first element will be produced (it is equal to [period] by default) in milliseconds.
+ */
+public fun tickerFlow(
+        period: Long,
+        initialDelay: Long = period
```
joffrey-bion, line 21: To follow the example of functions like `withTimeout` and `debounce`, I think we should add the `Millis` suffix to these parameter names, and maybe provide an overload using `kotlin.time.Duration`?

kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
```diff
@@ -0,0 +1,33 @@
...
+): Flow<Unit> = callbackFlow {
+    require(period > 0)
+    require(initialDelay > -1)
+
+    val timer = Timer()
+    timer.schedule(initialDelay, period) {
+        offer(Unit)
+    }
```
joffrey-bion, line 29: This implementation uses a background thread and is JVM-specific.
Couldn't we simply use a loop with `delay()` and make it multiplatform? Or am I naive here?

kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
ibrahimyilmaz: Good point for naming and we have `.receiveAsFlow()` extension function.

Directly we can use kotlinx-coroutine's `ticker`. Also this can be better idea to have single implementation.
What do you think about it?

kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
ibrahimyilmaz: Thanks! very good point.
 Also `ticker` implementation can be moved to `commonMain`.

kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
joffrey-bion: It could indeed be done this way, but that would couple it to JVM platform 😥
I am not sure a channel would be necessary behind the scenes to implement it, but I will let the maintainers comment on this.

Also, I can see now why a concurrent behaviour is necessary and would like to backtrack a bit: we can't use a simple loop with delays because otherwise the ticker delay is impacted by the execution time of the collector code, which is probably not desirable in this case.

That's why we do need some concurrent process (a coroutine?) to send the ticks.

kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
joffrey-bion: I can see now why a concurrent behaviour is necessary and would like to backtrack a bit: we can't use a simple loop with delays because otherwise the ticker delay is impacted by the execution time of the collector code, which is probably not desirable in this case.

That's why we do need some concurrent process (a coroutine?) to send the ticks.
This still doesn't mean we have to use a thread and Java's timer.
There may be a multiplatform way to do this.

I'll let the maintainers of the lib comment on this, though

kotlinx-coroutines-core/jvm/src/flow/TickerFlow.kt
```diff
@@ -0,0 +1,30 @@
...
+ * @param mode specifies behavior when elements are not received ([FIXED_PERIOD][TickerMode.FIXED_PERIOD] by default).
+ */
+public fun tickerFlow(
+        delayMillis: Long,
+        initialDelayMillis: Long = delayMillis,
+        context: CoroutineContext = EmptyCoroutineContext,
+        mode: TickerMode = TickerMode.FIXED_PERIOD
+): Flow<Unit> {
+    require(delayMillis > 0)
+    return ticker(delayMillis, initialDelayMillis, context, mode).receiveAsFlow()
```
joffrey-bion, line 29: Because the channel is hidden here, we probably want to handle cancellation properly.
Using `consumeAsFlow()` instead of `receiveAsFlow()` should do the trick I think.

PR #1923, opened 2020-04-19
Add way to convert CoroutineDispatcher to Scheduler 

### LINKED ISSUE #548: Way to use a CoroutineDispatcher as a RxJava Scheduler
The rx2 artifact currently provides `Scheduler.asCoroutineDispatcher()` to use a Scheduler as a CoroutineDispatcher, but not the other way around.

Ideally I'd like to move over to the new `IO` and `BackgroundDispatcher`, and then move my Rx observables to use the same threads too.

### LINKED ISSUE #968: Allow converting CoroutineDispatcher to RxJava scheduler
Currently there is a way to create a `CoroutineDispatcher` from a `Scheduler` (RxJava) which is great but it would also be convenient to be able to do the reverse. RxJava allows you to set your own schedulers in the `Schedulers` singleton meaning it could be overridden for the whole system.

recheej:
Fixes #968 and fixes #548 (duplicates).

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/api/kotlinx-coroutines-rx2.api
+++ b/reactive/kotlinx-coroutines-rx2/api/kotlinx-coroutines-rx2.api
@@ -67,7 +67,8 @@ public final class kotlinx/coroutines/rx2/RxObservableKt {
 }
 
 public final class kotlinx/coroutines/rx2/RxSchedulerKt {
-	public static final fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/rx2/SchedulerCoroutineDispatcher;
+	public static final fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/CoroutineDispatcher;
+	public static final fun asScheduler (Lkotlinx/coroutines/CoroutineDispatcher;)Lio/reactivex/Scheduler;
 }
 
 public final class kotlinx/coroutines/rx2/RxSingleKt {
--- a/reactive/kotlinx-coroutines-rx2/api/kotlinx-coroutines-rx2.api
+++ b/reactive/kotlinx-coroutines-rx2/api/kotlinx-coroutines-rx2.api
@@ -68,6 +68,7 @@ public final class kotlinx/coroutines/rx2/RxObservableKt {
 
 public final class kotlinx/coroutines/rx2/RxSchedulerKt {
 	public static final fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/CoroutineDispatcher;
+	public static final synthetic fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/rx2/SchedulerCoroutineDispatcher;
 	public static final fun asScheduler (Lkotlinx/coroutines/CoroutineDispatcher;)Lio/reactivex/Scheduler;
 }
 
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -29,12 +29,9 @@ public fun CoroutineDispatcher.asScheduler(): Scheduler =
 
 private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) : Scheduler() {
 
-    private val scope = CoroutineScope(SupervisorJob())
-
     override fun scheduleDirect(run: java.lang.Runnable): Disposable {
         val decoratedRun = RxJavaPlugins.onSchedule(run)
         val worker = createWorker() as DispatcherWorker
-        worker.setJob(Job(scope.coroutineContext[Job]))
         worker.schedule(decoratedRun)
         return worker
     }
@@ -42,26 +39,18 @@ private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) :
     override fun scheduleDirect(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
         val decoratedRun = RxJavaPlugins.onSchedule(run)
         val worker = createWorker() as DispatcherWorker
-        val job = scope.launch {
-            delay(unit.toMillis(delay))
-            worker.schedule(decoratedRun)
-        }
-        worker.setJob(job)
+        worker.schedule(decoratedRun, delay, unit)
         return worker
     }
 
-    private inner class DispatcherWorker : Worker() {
+    private class DispatcherWorker(private val dispatcher: CoroutineDispatcher) : Worker() {
 
-        private lateinit var workerJob: Job
+        private val workerScope = CoroutineScope(Job())
 
-        fun setJob(job: Job) {
-            workerJob = job
-        }
+        override fun isDisposed(): Boolean = !workerScope.isActive
 
-        override fun isDisposed(): Boolean = !workerJob.isActive
-
-        override fun schedule(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
-            return if (workerJob.isActive) {
+        override fun schedule(run: java.lang.Runnable): Disposable {
+            return if (workerScope.isActive) {
                 dispatcher.dispatch(EmptyCoroutineContext, run)
                 return this
             } else {
@@ -69,16 +58,24 @@ private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) :
             }
         }
 
+        override fun schedule(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
+            return if (workerScope.isActive) {
+                workerScope.launch {
+                    delay(unit.toMillis(delay))
+                    dispatcher.dispatch(EmptyCoroutineContext, run)
+                }
+                return this
+            } else {
+                Disposables.disposed()
+            }
+        }
+
         override fun dispose() {
-            workerJob.cancel()
+            workerScope.cancel()
         }
     }
 
-    override fun createWorker(): Worker = DispatcherWorker()
-
-    override fun shutdown() {
-        scope.cancel()
-    }
+    override fun createWorker(): Worker = DispatcherWorker(dispatcher)
 }
 
 /**
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -69,10 +69,6 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
         private val blockChannel = Channel<SchedulerChannelTask>(Channel.UNLIMITED)
 
         init {
-            workerJob.invokeOnCompletion {
-                blockChannel.close()
-            }
-
             workerScope.launch {
                 while (isActive) {
                     val task = blockChannel.receive()
@@ -113,7 +109,6 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
 
         override fun dispose() {
             workerScope.cancel()
-            blockChannel.close()
         }
     }
 }
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -119,13 +119,17 @@ private class SchedulerChannelTask(
     }
 
     fun execute() {
-        if (!delayJob.isCompleted) {
+        if (job.isCancelled) {
+            return
+        }
+
+        if (delayJob.isCompleted) {
+            block.run()
+        } else {
             taskScope.launch {
                 delayJob.join()
                 block.run()
             }
-        } else {
-            block.run()
         }
     }
 }
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -174,4 +174,4 @@ private class JobDisposable(private val job: Job) : Disposable {
     }
 }
 
-public fun Job.asDisposable(): Disposable = JobDisposable(this)
+private fun Job.asDisposable(): Disposable = JobDisposable(this)
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -62,7 +62,7 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
         scope.cancel()
     }
 
-    private class DispatcherWorker(private val dispatcher: CoroutineDispatcher, parentJob: Job) : Worker() {
+    private class DispatcherWorker(dispatcher: CoroutineDispatcher, parentJob: Job) : Worker() {
 
         private val workerJob = SupervisorJob(parentJob)
         private val workerScope = CoroutineScope(workerJob + dispatcher)
@@ -90,9 +90,25 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
             if (!workerScope.isActive) return Disposables.disposed()
 
             val newBlock = RxJavaPlugins.onSchedule(block)
-            val task = SchedulerChannelTask(delay, newBlock, dispatcher, workerJob)
-            blockChannel.offer(task)
-            return task
+
+            fun addTaskToQueue(job: Job) {
+                val task = SchedulerChannelTask(newBlock, job)
+                blockChannel.offer(task)
+            }
+
+            val job = if(delay == 0L) {
+                Job(workerJob).also {
+                    addTaskToQueue(it)
+                }
+            } else {
+                workerScope.launch {
+                    delay(unit.toMillis(delay))
+                    val task = SchedulerChannelTask(newBlock, checkNotNull(coroutineContext[Job]))
+                    blockChannel.offer(task)
+                }
+            }
+
+            return job.asDisposable()
         }
 
         override fun dispose() {
@@ -108,38 +124,30 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
  * Delayed tasks do not block [Channel] from processing other tasks
  */
 private class SchedulerChannelTask(
-    private val delayMillis: Long,
     private val block: Runnable,
-    dispatcher: CoroutineDispatcher,
-    parentJob: Job
+    job: Job
 ) : Disposable {
-    private val taskScope = CoroutineScope(dispatcher + Job(parentJob))
-    private val delayResult: Deferred<Unit>
+    private var isActive = true
 
     init {
-        delayResult = taskScope.async {
-            delay(delayMillis)
-        }
-    }
-
-    fun execute() {
-        if (taskScope.isActive) {
-            if (delayResult.isCompleted) {
-                block.run()
-            } else {
-                taskScope.launch {
-                    delayResult.await()
-                    block.run()
-                }
+        job.invokeOnCompletion {
+            if (it is kotlinx.coroutines.CancellationException) {
+                isActive = false
             }
         }
     }
 
+    fun execute() {
+        if (isActive) {
+            block.run()
+        }
+    }
+
     override fun isDisposed(): Boolean =
-        !taskScope.isActive
+        !isActive
 
     override fun dispose() {
-        taskScope.cancel()
+        isActive = false
     }
 }
 
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -8,6 +8,7 @@ import io.reactivex.*
 import io.reactivex.disposables.*
 import io.reactivex.plugins.*
 import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
 
@@ -29,27 +30,43 @@ public fun CoroutineDispatcher.asScheduler(): Scheduler =
 
 private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) : Scheduler() {
 
+    val parentJob = SupervisorJob()
+    private val parentScope = CoroutineScope(parentJob)
+
     override fun scheduleDirect(run: java.lang.Runnable): Disposable {
         val decoratedRun = RxJavaPlugins.onSchedule(run)
         val worker = createWorker() as DispatcherWorker
+        worker.startProcessingQueue(parentJob)
         worker.schedule(decoratedRun)
         return worker
     }
 
+    override fun scheduleDirect(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
+        val decoratedRun = RxJavaPlugins.onSchedule(run)
+        val worker = createWorker() as DispatcherWorker
+        worker.startProcessingQueue(parentJob)
+        worker.schedule(decoratedRun, delay, unit)
+        return worker
+    }
+
+    override fun createWorker(): Worker =
+        DispatcherWorker(dispatcher)
+
+    override fun shutdown() {
+        parentScope.cancel()
+    }
+
     private class DispatcherWorker(private val dispatcher: CoroutineDispatcher) : Worker() {
 
-        val parentJob = SupervisorJob()
-        private val workerScope = CoroutineScope(parentJob)
-        private var previousNonDelayJob: Job? = null
+        private lateinit var workerScope: CoroutineScope
+        private val blockChannel = Channel<java.lang.Runnable>()
 
         override fun isDisposed(): Boolean = !workerScope.isActive
 
         override fun schedule(block: java.lang.Runnable): Disposable =
             if (workerScope.isActive) {
                 workerScope.launch(dispatcher) {
-                    previousNonDelayJob?.join()
-                    previousNonDelayJob = this.coroutineContext[Job]
-                    block.run()
+                    blockChannel.send(block)
                 }
                 this
             } else {
@@ -74,16 +91,20 @@ private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) :
         override fun dispose() {
             workerScope.cancel()
         }
-    }
 
-    override fun scheduleDirect(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
-        val decoratedRun = RxJavaPlugins.onSchedule(run)
-        val worker = createWorker() as DispatcherWorker
-        worker.schedule(decoratedRun, delay, unit)
-        return worker
+        fun startProcessingQueue(parentJob: Job) {
+            workerScope = CoroutineScope(SupervisorJob(parentJob))
+            workerScope.launch(dispatcher) {
+                while (true) {
+                    yield()
+                    if (!blockChannel.isEmpty) {
+                        val block = blockChannel.receive()
+                        block.run()
+                    }
+                }
+            }
+        }
     }
-
-    override fun createWorker(): Worker = DispatcherWorker(dispatcher)
 }
 
 /**
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -87,24 +87,29 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
 
             val newBlock = RxJavaPlugins.onSchedule(block)
 
-            fun addTaskToQueue(job: Job) {
-                val task = SchedulerChannelTask(newBlock, job)
+            val taskJob = Job(workerJob)
+
+            fun addTaskToQueue() {
+                val task = SchedulerChannelTask(newBlock, taskJob)
                 blockChannel.offer(task)
             }
 
-            val job = if(delay == 0L) {
-                Job(workerJob).also {
-                    addTaskToQueue(it)
-                }
+            if (delay == 0L) {
+                addTaskToQueue()
             } else {
-                workerScope.launch {
+                // Use `taskJob` as the parent here so the delay will also get cancelled if the Disposable
+                // is disposed.
+                workerScope.launch(taskJob) {
+                    // Delay *before* enqueuing the task, so other tasks (e.g. via schedule without delay)
+                    // aren't blocked by the delay.
                     delay(unit.toMillis(delay))
-                    val task = SchedulerChannelTask(newBlock, checkNotNull(coroutineContext[Job]))
-                    blockChannel.offer(task)
+                    // Once the task is ready to run, it still needs to be executed via the queue to comply
+                    // with the Scheduler contract of running all worker tasks in a non-overlapping manner.
+                    addTaskToQueue()
                 }
             }
 
-            return job.asDisposable()
+            return taskJob.asDisposable()
         }
 
         override fun dispose() {
@@ -120,29 +125,19 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
  */
 private class SchedulerChannelTask(
     private val block: Runnable,
-    job: Job
+    private val job: Job
 ) : Disposable {
-    private var isActive = true
-
-    init {
-        job.invokeOnCompletion {
-            if (it is kotlinx.coroutines.CancellationException) {
-                isActive = false
-            }
-        }
-    }
-
     fun execute() {
-        if (isActive) {
+        if (job.isActive) {
             block.run()
         }
     }
 
     override fun isDisposed(): Boolean =
-        !isActive
+        !job.isActive
 
     override fun dispose() {
-        isActive = false
+        job.cancel()
     }
 }
 
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -16,7 +16,12 @@ import kotlin.coroutines.*
  * Converts an instance of [Scheduler] to an implementation of [CoroutineDispatcher]
  * and provides native support of [delay] and [withTimeout].
  */
-public fun Scheduler.asCoroutineDispatcher(): SchedulerCoroutineDispatcher = SchedulerCoroutineDispatcher(this)
+public fun Scheduler.asCoroutineDispatcher(): CoroutineDispatcher =
+    if (this is DispatcherScheduler) {
+        dispatcher
+    } else {
+        SchedulerCoroutineDispatcher(this)
+    }
 
 /**
  * Converts an instance of [CoroutineDispatcher] to an implementation of [Scheduler].
@@ -28,7 +33,7 @@ public fun CoroutineDispatcher.asScheduler(): Scheduler =
         DispatcherScheduler(this)
     }
 
-private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) : Scheduler() {
+private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) : Scheduler() {
 
     val parentJob = SupervisorJob()
     private val parentScope = CoroutineScope(parentJob)
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -23,7 +23,7 @@ public fun Scheduler.asCoroutineDispatcher(): CoroutineDispatcher =
         SchedulerCoroutineDispatcher(this)
     }
 
-@Deprecated(level = DeprecationLevel.HIDDEN, message = "Since 1.4.0, binary compatibility with earlier versions")
+@Deprecated(level = DeprecationLevel.HIDDEN, message = "Since 1.4.2, binary compatibility with earlier versions")
 @JvmName("asCoroutineDispatcher")
 public fun Scheduler.asCoroutineDispatcher0(): SchedulerCoroutineDispatcher =
     SchedulerCoroutineDispatcher(this)
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -45,8 +45,8 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
 
     override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
         if (!scope.isActive) return Disposables.disposed()
+        val newBlock = RxJavaPlugins.onSchedule(block)
         return scope.launch {
-            val newBlock = RxJavaPlugins.onSchedule(block)
             if (delay > 0) {
                 delay(unit.toMillis(delay))
             }
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -49,17 +49,16 @@ private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) :
 
         override fun isDisposed(): Boolean = !workerScope.isActive
 
-        override fun schedule(run: java.lang.Runnable): Disposable {
-            return if (workerScope.isActive) {
+        override fun schedule(run: java.lang.Runnable): Disposable =
+            if (workerScope.isActive) {
                 dispatcher.dispatch(EmptyCoroutineContext, run)
-                return this
+                this
             } else {
                 Disposables.disposed()
             }
-        }
 
-        override fun schedule(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
-            return if (delay <= 0) {
+        override fun schedule(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable =
+            if (delay <= 0) {
                 schedule(run)
             } else {
                 if (workerScope.isActive) {
@@ -67,12 +66,11 @@ private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) :
                         delay(unit.toMillis(delay))
                         schedule(run)
                     }
-                    return this
+                    this
                 } else {
                     Disposables.disposed()
                 }
             }
-        }
 
         override fun dispose() {
             workerScope.cancel()
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -38,25 +38,17 @@ private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) 
     private val job = SupervisorJob()
     private val scope = CoroutineScope(job + dispatcher)
 
-    override fun scheduleDirect(block: Runnable): Disposable {
-        if (!scope.isActive) return Disposables.disposed()
-        return scope.launch {
-            block.runWithRx()
-        }.asDisposable()
-    }
+    override fun scheduleDirect(block: Runnable): Disposable =
+        scheduleDirect(block, 0, TimeUnit.MILLISECONDS)
 
     override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
-        if (delay <= 0) {
-            return scheduleDirect(block)
-        }
         if (!scope.isActive) return Disposables.disposed()
         return scope.launch {
+            val newBlock = RxJavaPlugins.onSchedule(block)
             delay(unit.toMillis(delay))
-            block.runWithRx()
+            newBlock.run()
         }.asDisposable()
     }
-
-    private fun Runnable.runWithRx() = RxJavaPlugins.onSchedule(this).run()
 
     override fun createWorker(): Worker =
         DispatcherWorker(dispatcher, job)
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -35,38 +35,35 @@ public fun CoroutineDispatcher.asScheduler(): Scheduler =
 
 private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) : Scheduler() {
 
-    val job = SupervisorJob()
-    private val scope = CoroutineScope(job)
+    private val job = SupervisorJob()
+    private val scope = CoroutineScope(job + dispatcher)
 
-    override fun scheduleDirect(run: Runnable): Disposable {
+    override fun scheduleDirect(block: Runnable): Disposable {
         if (scope.isActive) {
             return scope.launch {
-                dispatchBlock(run)
+                block.runWithRx()
             }.asDisposable()
         }
 
         return Disposables.disposed()
     }
 
-    override fun scheduleDirect(run: Runnable, delay: Long, unit: TimeUnit): Disposable {
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
         if (delay <= 0) {
-            return scheduleDirect(run)
+            return scheduleDirect(block)
         }
 
         if (scope.isActive) {
             return scope.launch {
                 delay(unit.toMillis(delay))
-                dispatchBlock(run)
+                block.runWithRx()
             }.asDisposable()
         }
 
         return Disposables.disposed()
     }
 
-    private fun dispatchBlock(block: Runnable) {
-        val decoratedRun = RxJavaPlugins.onSchedule(block)
-        dispatcher.dispatch(EmptyCoroutineContext, decoratedRun)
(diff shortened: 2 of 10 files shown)


reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -5,24 +5,87 @@
...
+public fun CoroutineDispatcher.asScheduler(): Scheduler =
+        if (this is SchedulerCoroutineDispatcher) {
+            scheduler
+        } else {
+            DispatcherScheduler(this)
+        }
+
+private class DispatcherScheduler(private val dispatcher: CoroutineDispatcher) : Scheduler() {
+
+    private val scope = CoroutineScope(SupervisorJob())
```
elizarov, line 34: There does not seem to be any need to create a `Job` for this parent scope. It is used to create child worker jobs, but it is not really needed. There's no code that would ever cancel this parent job.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -5,24 +5,87 @@
...
+        val decoratedRun = RxJavaPlugins.onSchedule(run)
+        val worker = createWorker()
+        worker.schedule(decoratedRun)
+        return worker
+    }
+
+    override fun scheduleDirect(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
+        val decoratedRun = RxJavaPlugins.onSchedule(run)
+        val worker = createWorker()
+        scope.launch {
```
elizarov, line 46: You must run this task in the scope of the worker's job that was just created. The way it is written right now, destroying this worker while it waits for the initial delay does not cancel this waiting job. Please, add a stress test for this problem -- schedule & immediately cancel millions of jobs. This code, before it is fixed, will get OutOfMemory error.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -5,24 +5,87 @@
 package kotlinx.coroutines.rx2
 
 import io.reactivex.Scheduler
+import io.reactivex.disposables.Disposable
+import io.reactivex.disposables.Disposables
+import io.reactivex.plugins.RxJavaPlugins
```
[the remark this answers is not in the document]
recheej: good to know. do you have the code style published anywhere?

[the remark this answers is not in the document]
elizarov: Nope. That's a standard official Kotlin style with start imports (which are not covered by style). Committing it to a Git seems like a good idea.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: @elizarov while it's not created with the job from ```launch```, the worker does create its own job upon instantiation. so it will get cancelled. Although it seems like a waste to create two Jobs so I'll change. Hopefully that makes sense.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: @elizarov the parent scope is cancelled in the ```shutdown``` method:

```kotlin
    override fun shutdown() {
        scope.cancel()
    }
```

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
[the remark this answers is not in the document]
recheej: @elizarov Yea I think for IDEA IDE you can publish the ```editorConfig.xml``` to version control to share. But in my experience it's limited so we've also had to mix it with ```ktlint```. I think IDEA change will be simple so I'll look into putting that up as a PR.

recheej:
@elizarov ready for review again. Indeed there were some OOM issues so I simplified the job handling. as per your recommendations.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,17 +4,84 @@
...
+            } else {
+                Disposables.disposed()
+            }
+        }
+
+        override fun schedule(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
+            return if (delay <= 0) {
+                schedule(run)
+            } else {
+                if (workerScope.isActive) {
```
elizarov, line 65: You don't need this `if`. `launch` on the next line will check it anyway.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,17 +4,84 @@
...
+    }
+
+    private class DispatcherWorker(private val dispatcher: CoroutineDispatcher) : Worker() {
+
+        private val workerScope = CoroutineScope(SupervisorJob())
+
+        override fun isDisposed(): Boolean = !workerScope.isActive
+
+        override fun schedule(run: java.lang.Runnable): Disposable {
+            return if (workerScope.isActive) {
```
elizarov, line 53: Please expression function form (`= if ...`) if the function is doing just `return`.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: @elizarov it's recommended to return ```Disposable.disposed``` if you know the task is already cancelled. That's why this if check is here.

elizarov: Makes sense. The corresponding comment in the code would be appropriate, then.
Also, stylistically this code has too much nesting. The "usual path" is burried too deep. It would be easier to read if it is rewritten in "early return" style:
```
if (delay <= 0) return schedule(run)
if (!workerScope.isActive) return Disposables.disposed() // comment here
// the usual path here without nesting
```

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,118 @@
...
+
+        private fun startProcessingQueue() {
+            if (queueProcessingJob == null) {
+                queueProcessingJob = workerScope.launch {
+                    while (true) {
+                        if (isActive && !blockChannel.isEmpty) {
+                            val block = blockChannel.receive()
+                            block.run()
+                        }
+                        yield()
```
zach-klippenstein, line 112: Won't this loop effectively busy-wait on the channel? It seems like it would be simpler and more efficient to just receive on the channel and let the coroutine suspend until another runnable is available.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,118 @@
...
+
+        override fun dispose() {
+            workerScope.cancel()
+            blockChannel.close()
+        }
+
+        private fun startProcessingQueue() {
+            if (queueProcessingJob == null) {
+                queueProcessingJob = workerScope.launch {
+                    while (true) {
```
zach-klippenstein, line 107: Isn't it more idiomatic to put the `isActive` check in the loop condition?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,118 @@
...
+    }
+
+private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) : Scheduler() {
+
+    val parentJob = SupervisorJob()
+    private val parentScope = CoroutineScope(parentJob)
+
+    override fun scheduleDirect(run: Runnable): Disposable {
+        val decoratedRun = RxJavaPlugins.onSchedule(run)
+        val worker = createWorker() as DispatcherWorker
```
zach-klippenstein, line 43: Can you just run the runnable directly on the dispatcher? It seems wasteful to create a worker with it's extra queuing machinery just to do this. The built-in rxjava schedulers seem to prefer running the task as directly as possible from these `scheduleDirect` methods, instead of delegating to their workers.

E.g.
```kotlin
val job = SupervisorJob(parentJob)
dispatcher.dispatch(job, decoratedRun)
// return the job as a Disposable
```

If that doesn't work, even just launching a coroutine to run this one runnable would probably be better than creating a whole worker.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,118 @@
...
+        private val workerScope = CoroutineScope(SupervisorJob(parentJob) + dispatcher)
+        private val blockChannel = Channel<Runnable>()
+        private var queueProcessingJob: Job? = null
+
+        override fun isDisposed(): Boolean = !workerScope.isActive
+
+        override fun schedule(block: Runnable): Disposable {
+            startProcessingQueue()
+            if (workerScope.isActive) {
+                workerScope.launch(dispatcher) {
```
zach-klippenstein, line 75: I think this could violate the Worker contract. If the underlying dispatcher uses a thread pool, and a bunch of tasks are scheduled on the worker at the same time, these launched coroutines could get distributed over multiple threads and the order in which the tasks are added to the queue and executed would be non-deterministic.

One solution would be to make the channel unlimited capacity and just call `blockChannel.offer` here, although you'd need to then explicitly associate some disposable with the Runnable.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: @zach-klippenstein that's not a bad idea, however it seems like it would break things for clients. ```createWorker``` is a public method so how should it behave if we don't use workers to run the tasks? Would we return null for ```createWorker```?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: Agreed.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: Good point. I actually found what I think is a good solution to this by offering a lazy ```Job```. Let me know what you think (about to push in a few)

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
zach-klippenstein: No, `createWorker` is part of the Scheduler API, and your implementation of it seems reasonable (although I've never implemented a custom scheduler, I'm just comparing with Rx source). Take a look at the built-in schedulers' implementations to see what I mean. In fact the [documentation on `Scheduler` even says](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/core/Scheduler.java): `Implementors of this class are encouraged to provide a more efficient direct scheduling implementation to avoid the time and memory overhead of creating such {@code Worker}s for every task.`

`scheduleDirect` is used to schedule a single task, independently of any others. `createWorker` is used when multiple tasks need to be scheduled to run in order. I believe this is used to satisfy a lot of Rx contracts that require order to be preserved - eg `onNext`s always need to be executed in the order they were invoked in, so they use a worker. 

You can find the implementations of the built in schedulers here: https://github.com/ReactiveX/RxJava/tree/3.x/src/main/java/io/reactivex/rxjava3/internal/schedulers

(I've referenced rx3 docs/code here, but the contract is the same for rx2, you can look at the rx2 source in that same repo as well.)

recheej: Ah I see! I read the docs on worker and scheduler but guess I misunderstood. Thanks for clearing that up to me and pointing to those examples! Yes i think I'll change my implementation.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: @zach-klippenstein done. let me know what you think of new implementation.

recheej:
@elizarov I created new API: ```Job.asDisposable()```. What steps should I follow to make sure the api is documented, published, etc?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,17 +4,82 @@
...
+    }
+
+    override fun scheduleDirect(run: java.lang.Runnable, delay: Long, unit: TimeUnit): Disposable {
+        val decoratedRun = RxJavaPlugins.onSchedule(run)
+        val worker = createWorker() as DispatcherWorker
+        worker.schedule(decoratedRun, delay, unit)
+        return worker
+    }
+
+    private class DispatcherWorker(private val dispatcher: CoroutineDispatcher) : Worker() {
```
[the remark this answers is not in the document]
recheej: @elizarov done. thank you for pointing this out. I added both tests + I am now using channel to guarantee sequential behavior.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,129 @@
...
+
+        override fun isDisposed(): Boolean = !workerScope.isActive
+
+        override fun schedule(block: Runnable): Disposable {
+            startProcessingQueue()
+            if (workerScope.isActive) {
+                val job = workerScope.launch(start = CoroutineStart.LAZY) {
+                    block.run()
+                }
+                blockChannel.offer(job)
```
zach-klippenstein, line 89: If the channel has been closed, this will throw an exception, which i don't think should be thrown from this method directly - maybe reported to `RxPlugins.onError`?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -45,8 +158,20 @@ public class SchedulerCoroutineDispatcher(
...
+
     /** @suppress */
     override fun equals(other: Any?): Boolean = other is SchedulerCoroutineDispatcher && other.scheduler === scheduler
+
     /** @suppress */
     override fun hashCode(): Int = System.identityHashCode(scheduler)
 }
+
+private class JobDisposable(private val job: Job) : Disposable {
+    override fun isDisposed(): Boolean = !job.isActive
```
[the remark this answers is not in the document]
elizarov: Indeed, but `isCompleted` is enough. It subsumes `isCancelled`.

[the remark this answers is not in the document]
zach-klippenstein: According to the table on the docs, `isCompleted` returns while the job is still in the cancelling state, so I would think it depends if `isDispose` should immediately return true after calling `dispose`. The docs for `Disposable` aren't clear on what exact behavior is expected, but the built-in implementations seem to suggest those two operations should be atomic.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,129 @@
...
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+        if (delay <= 0) {
+            return scheduleDirect(block)
+        }
+
+        if (scope.isActive) {
+            return scope.launch {
+                delay(unit.toMillis(delay))
+                block.runWithRx()
```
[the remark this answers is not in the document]
zach-klippenstein: You'll probably want to have a test for this then too.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,129 @@
...
+    if (this is SchedulerCoroutineDispatcher) {
+        scheduler
+    } else {
+        DispatcherScheduler(this)
+    }
+
+private class DispatcherScheduler(internal val dispatcher: CoroutineDispatcher) : Scheduler() {
+
+    private val job = SupervisorJob()
+    private val scope = CoroutineScope(job + dispatcher)
```
zach-klippenstein, line 39: I wonder if you need a `CoroutineExceptionHandler` here and in the worker's scope to report exceptions to `RxJavaPlugins`. Right now any exceptions that come from the Runnable or anything else will get dropped on the floor, which is bad.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,129 @@
...
+        }
+
+        override fun dispose() {
+            workerScope.cancel()
+            blockChannel.close()
+        }
+
+        private fun startProcessingQueue() {
+            if (queueProcessingJob == null) {
+                queueProcessingJob = workerScope.launch {
```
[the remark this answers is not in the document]
zach-klippenstein: For that matter, it wouldn't even have to be lazy, the first this this coroutine does is suspend itself so just keep it simple and start it normally.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,129 @@
...
+
+        override fun schedule(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+            startProcessingQueue()
+            if (delay <= 0) {
+                return schedule(block)
+            }
+            if (workerScope.isActive) {
+                return workerScope.launch {
+                    delay(unit.toMillis(delay))
+                    block.run()
```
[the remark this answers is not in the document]
recheej: @zach-klippenstein I'll also make this this on the channel.

As far as cancelling the task, wouldn't the solution I have to schedule without delay work? Right now I put the Job/Disposable itself on the queue. Then I check if the job is active once we dequeue from the channel. Wouldn't that handle what you're referring to? Let me know if I'm misunderstanding.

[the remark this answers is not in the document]
zach-klippenstein: Yea, I think it would work, but creating a new coroutine + job for that isn't necessary (the job isn't needed for structured concurrency, hierarchy, or cancelling a running coroutine – just checking if the task was disposed while waiting in the queue). But I think it should still be correct, I would just add some comments to explain why the launch is lazy and in what cases the job would be cancelled, since it all looks a bit strange at a first glance.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,129 @@
...
+            return scope.launch {
+                delay(unit.toMillis(delay))
+                block.runWithRx()
+            }.asDisposable()
+        }
+
+        return Disposables.disposed()
+    }
+
+    private fun Runnable.runWithRx() = RxJavaPlugins.onSchedule(this).run()
```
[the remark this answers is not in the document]
recheej: @zach-klippenstein from what I'm reading in the docs, the reason they use ```SchedulerRunnableIntrospection``` is because the ```Scheduler.schedule``` calls ```Scheduler.Worker.schedule```. This leads to multiple ```RxJavaPlugins.onSchedule(block)``` calls since you're supposed to call that method in both classes. 

However, since I'm not delegating to the worker in the scheduler in ```Scheduler.schedule```, I don't believe I need to use ``SchedulerRunnableIntrospection```. 

Let me know if that makes sense.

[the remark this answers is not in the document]
recheej: docs don't say explicitly that custom classes should use it.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: I expect the channel to be closed at the same time as ```workerScope``` is set to inactive. So it should not even reach here if the channel has been closed. If it does I think it's better to let the exception be thrown so we can immediately know and fix.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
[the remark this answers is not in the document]
zach-klippenstein: You're right, I misread the docs.

recheej:
@elizarov ready for review again.

recheej:
Before this gets merged I will squash + rebase

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+        yield()
+        finish(3)
+    }
+
+    @Test
+    fun testAsSchedulerWithDelay(): Unit = runTest {
+        expect(1)
+        val mainThread = Thread.currentThread()
+        val scheduler = (currentDispatcher() as CoroutineDispatcher).asScheduler()
+        val delayMillis = 300L
```
zach-klippenstein, line 64: This is going to slow down test runs. I think you could split this test into two: one with a minimal delay (1ms) that verifies the dispatcher is actually being used (like in the non-delay case), and one that uses a `TestCoroutineDispatcher` to fast-forward through a longer delay to verify that the delay is actually being respected.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+
+    @Test
+    fun testAsSchedulerWithNoDelay(): Unit = runTest {
+        expect(1)
+        val mainThread = Thread.currentThread()
+        val scheduler = (currentDispatcher() as CoroutineDispatcher).asScheduler()
+        suspendCancellableCoroutine<Unit> {
+            scheduler.scheduleDirect {
+                val t1 = Thread.currentThread()
+                assertSame(t1, mainThread)
```
zach-klippenstein, line 50: It's not clear these tests actually verify anything about threading. It would be better to use an explicit dispatcher that is guaranteed to use a different thread from the one the unit tests are being ran. Otherwise these tests will pass even if you remove the launch calls.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+                it.resume(Unit)
+            }, 0, TimeUnit.MILLISECONDS)
+        }
+
+        scheduler.shutdown()
+        finish(3)
+    }
+
+    @Test
+    fun testAsSchedulerWithNegativeDelay(): Unit = runTest {
```
zach-klippenstein, line 95: All the tests in this file are for `asScheduler`, these function names should indicate that these tests are testing `scheduleDirect` vs the worker methods.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+
+    @Test
+    fun testDisposeDuringDelay(): Unit = runTest {
+        expect(1)
+        val scheduler = (currentDispatcher() as CoroutineDispatcher).asScheduler()
+        val delayMillis = 300L
+        val disposable = scheduler.scheduleDirect({
+            expectUnreached()
+        }, delayMillis, TimeUnit.MILLISECONDS)
+        delay(100)
```
zach-klippenstein, line 119: These tests should be using a `DelayController` to advance time instead of actually sleeping.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+        val originalDispatcher = currentDispatcher() as CoroutineDispatcher
+        val scheduler = originalDispatcher.asScheduler()
+        val dispatcher = scheduler.asCoroutineDispatcher()
+        assertEquals(originalDispatcher, dispatcher)
+
+        finish(2)
+    }
+
+    @Test
+    fun testExpectRxPluginsCall(): Unit = runTest {
```
zach-klippenstein, line 179: With the change I mentioned above, the ordering this test should verify is:

1. scheduleDirect is called.
2. Plugin is invoked: `expect(2)`
3. Task has not been started: `expect(3)`
3. Start the task (`DelayController.runCurrent`)
4. Plugin's wrapper Runnable is executed: `expect(4)`
5. ScheduleDirect Runnable is executed: `expect(5)`
6. `finish(6)`

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+        suspendCancellableCoroutine<Unit> {
+            scheduler.scheduleDirect {
+                expect(3)
+                it.resume(Unit)
+            }
+        }
+
+        setScheduler(4)
+        suspendCancellableCoroutine<Unit> {
+            scheduler.scheduleDirect({
```
zach-klippenstein, line 205: This is testing two different functions, it should be two separate tests. This makes it more clear what's actually broken if it fails, and also ensures that you're not accidentally relying on (or being broken by) shared state between the two function calls.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+        scheduler.shutdown()
+        finish(6)
+    }
+
+    /**
+     * Let's test the [Scheduler.Worker] to make sure it satisfies the documented constraint of running all work
+     * sequentially
+     */
+    @Test
+    fun testSchedulerWorkerSequentialOrdering(): Unit = runTest {
```
zach-klippenstein, line 220: Should probably also have a test for the delayed version of this that ensures tasks scheduled at the same time don't overlap. 

Also a test that ensures that delays are actually respected (tasks scheduled sooner in the future run before tasks scheduled later, even when the later task is submitted before the earlier one). Such a test would have caught the misplaced delay call above.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +38,252 @@ class SchedulerTest : TestBase() {
...
+        val disposable = scheduler.scheduleDirect {
+            expectUnreached()
+        }
+        disposable.dispose()
+        yield()
+        finish(2)
+    }
+
+    @Test
+    fun testAsSchedulerWorksWithSchedulerCoroutineDispatcher(): Unit = runTest {
```
zach-klippenstein, line 140: All these tests for `scheduleDirect` should also be duplicated for tasks scheduled via workers.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,109 @@
...
+        RxJavaPlugins.onError(throwable)
+    })
+
+    override fun scheduleDirect(block: Runnable): Disposable =
+        scheduleDirect(block, 0, TimeUnit.MILLISECONDS)
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+        if (!scope.isActive) return Disposables.disposed()
+        return scope.launch {
+            val newBlock = RxJavaPlugins.onSchedule(block)
```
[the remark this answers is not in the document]
recheej: done. and used the coroutines-test to assert it.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,109 @@
...
+
+    override fun shutdown() {
+        scope.cancel()
+    }
+
+    private class DispatcherWorker(dispatcher: CoroutineDispatcher, parentJob: Job) : Worker() {
+
+        private val workerScope = CoroutineScope(SupervisorJob(parentJob) + dispatcher)
+        private val blockChannel = Channel<Job>(Channel.UNLIMITED)
+        private var queueProcessingJob: Job? = null
```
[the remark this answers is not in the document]
recheej: actually this job isn't needed anymore so removing.

reactive/kotlinx-coroutines-rx2/test/SchedulerStressTest.kt
```diff
@@ -0,0 +1,84 @@
...
+        coroutineScope {
+            repeat(n) { i ->
+                val a = ByteArray(1000000) //1MB
+                val disposable = scheduler.scheduleDirect {
+                    runBlocking {
+                        keepMe(a)
+                    }
+                }
+                disposable.dispose()
+                expect(i + 2)
```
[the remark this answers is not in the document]
recheej: hmm. sorry not quite following. This line is executed for n iterations.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
recheej: @zach-klippenstein indeed. updated.

recheej:
@zach-klippenstein @elizarov ready for review.

reactive/kotlinx-coroutines-rx2/test/SchedulerStressTest.kt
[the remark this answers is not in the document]
zach-klippenstein: All the expect/finally calls in this method happen in the same coroutine. These functions are intended to use to assert execution order across multiple concurrent coroutines. 

Eg this test doesn't do anything because there is no concurrency.:
```kotlin
fun testFoo() = runTest {
  expect(1)
  for (i in 0 until 100) {
    expect(i+1)
  }
  finish(101)
}
```

Compare with this test, which asserts that the underlying dispatcher starts executing coroutines in the order in which they were launched:
```kotlin
fun testFoo() = runTest {
  expect(1)
  for (i in 0 until 100) {
    launch { expect(i+1) }
  }
  finish(101)
}
```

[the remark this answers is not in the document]
recheej: Ah ok understood. Thanks for the explanation. I'll change

elizarov:
Please, ping me when you're done and its ready for review.

recheej:
@elizarov this is ready for review

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,144 @@
...
+    private class DispatcherWorker(private val dispatcher: CoroutineDispatcher, parentJob: Job) : Worker() {
+
+        private val workerJob = SupervisorJob(parentJob)
+        private val workerScope = CoroutineScope(workerJob + dispatcher)
+        private val blockChannel = Channel<SchedulerChannelTask>(Channel.UNLIMITED)
+
+        init {
+            workerJob.invokeOnCompletion {
+                blockChannel.close()
+            }
```
zach-klippenstein, line 74: I think this is unnecessary. The only reason to close the channel is to signal to consumers that the work is done. However, the only way `workerJob` will be completed is if it's been cancelled, and since the only consumer is the coroutine launched under `workerJob` anyway, this is redundant.

The only other reason I can think of to close the channel would be to ensure that `schedule` throws when called after the worker has been disposed, but I don't think that's required for the Scheduler contract (and if it is, there should probably be a comment here mentioning that intent).

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,144 @@
...
+
+            val newBlock = RxJavaPlugins.onSchedule(block)
+            val task = SchedulerChannelTask(delay, newBlock, dispatcher, workerJob)
+            blockChannel.offer(task)
+            return task
+        }
+
+        override fun dispose() {
+            workerScope.cancel()
+            blockChannel.close()
```
zach-klippenstein, line 100: This is also redundant, see above comment.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,144 @@
...
+        }
+    }
+
+    fun execute() {
+        if (taskScope.isActive) {
+            if (delayResult.isCompleted) {
+                block.run()
+            } else {
+                taskScope.launch {
+                    delayResult.await()
```
zach-klippenstein, line 131: This delay logic is still wrong I think. It will not start the delay until the task is processed, and then when the delay is over it the task will not necessarily be ran exclusively with other tasks from the same worker. The delay needs to happen _before_ submitting the task to the worker queue, and then the task needs to be submitted to the queue after the delay.

The distinction is that the _delay_ can (and should) happen concurrently with other delays, and other tasks, but the actual execution of the task itself must happen in a way that doesn't "overlap" with other tasks submitted to the worker (scheduled or not).

Here's a rough diagram of the sequence that needs to happen, as I understand it:

| t (ms) | submitter1    | submitter2           | worker                        |                                                                                                                                        |
|--------|---------------|----------------------|-------------------------------|----------------------------------------------------------------------------------------------------------------------------------------|
| 0      | schedule(foo) | schedule(bar, 1 sec) | offer(foo), delay(bar, 1 sec) | foo and bar are submitted concurrently. Worker enqueues foo immediately, but defers enqueuing bar.                                     |
| 1      |               |                      | run foo                       | Worker reads from queue, runs bar                                                                                                      |
| …      |               |                      |                               |                                                                                                                                        |
| 1000   | schedule(baz) |                      | offer(baz), offer(bar)        | After 1 sec delay, the delayed bar task is enqueued concurrently with a new task baz, but bar and baz must still be executed in order. |
| 1001   |               |                      | run baz                       |                                                                                                                                        |
| 1002   |               |                      | run bar                       |                                                                                                                                        |

reactive/kotlinx-coroutines-rx2/test/SchedulerStressTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        val n = 2000 * stressTestMultiplier
+        coroutineScope {
+            repeat(n) { i ->
+                launch {
+                    val a = ByteArray(1000000) //1MB
+                    val disposable = block(Runnable {
+                        expectUnreached()
+                        runBlocking {
+                            keepMe(a)
+                        }
```
zach-klippenstein, line 51: `runBlocking` with `delay` is effectively just a `Thread.sleep`, it would be less code and simpler to just do that here.

reactive/kotlinx-coroutines-rx2/test/SchedulerStressTest.kt
```diff
@@ -0,0 +1,115 @@
...
+                val a = ByteArray(1000000) //1MB
+                val delayMillis: Long = 10
+                val disposable = block(Runnable {
+                    runBlocking {
+                        keepMe(a)
+                    }
+                }, delayMillis, TimeUnit.MILLISECONDS)
+                disposable.dispose()
+                advanceTimeBy(delayMillis)
+                expect(i + 2)
```
zach-klippenstein, line 107: Again, this `expect` isn't doing anything. It's not running concurrently with any other coroutines, so there's no ambiguous ordering to verify. It would be testing something if it were inside the `Runnable` above.

reactive/kotlinx-coroutines-rx2/test/SchedulerStressTest.kt
recheej: While less code, it also requires an extra param for the sleep which I think makes it a little dirtier. i'd prefer to keep it as is.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: @zach-klippenstein thanks again for the explanation. sorry the expected execution order isn't clear to me from reading the docs but what you said I think makes sense.

recheej: Done

recheej:
@zach-klippenstein @elizarov ready for review

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,147 @@
...
+
+            val job = if(delay == 0L) {
+                Job(workerJob).also {
+                    addTaskToQueue(it)
+                }
+            } else {
+                workerScope.launch {
+                    delay(unit.toMillis(delay))
+                    val task = SchedulerChannelTask(newBlock, checkNotNull(coroutineContext[Job]))
+                    blockChannel.offer(task)
```
zach-klippenstein, line 103: These two lines should be a call to `addTaskToQueue`. I think you also can't use this `Job`, since the job will become completed as soon as this `launch` lambda returns. Instead, just pull the `Job(workerJob)` call in the first branch out into a val and pass that same instance to `addTaskToQueue` in both branches.

So just to be clear:
```kotlin
val taskJob = Job(workerJob)

fun addTaskToQueue() {
  val task = SchedulerChannelTask(newBlock, taskJob)
  blockChannel.offer(task)
}

if (delay == 0L) {
  addTaskToQueue()
} else {
  // Use `taskJob` as the parent here so the delay will also get cancelled if the Disposable
  // is disposed.
  workerScope.launch(taskJob) {
    // Delay *before* enqueuing the task, so other tasks (e.g. via schedule without delay)
    // aren't blocked by the delay.
    delay(unit.toMillis(delay))
    // Once the task is ready to run, it still needs to be executed via the queue to comply
    // with the Scheduler contract of running all worker tasks in a non-overlapping manner.
    addTaskToQueue()
  }
}

return taskJob.asDisposable()
```
Including those comments, since the implications of this code are evidently pretty subtle, and it would be helpful to future readers/maintainers to explain why it's done the way it is.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,147 @@
...
+/**
+ * Represents a task to be queued sequentially on a [Channel] for a [Scheduler.Worker].
+ *
+ * Delayed tasks do not block [Channel] from processing other tasks
+ */
+private class SchedulerChannelTask(
+    private val block: Runnable,
+    job: Job
+) : Disposable {
+    private var isActive = true
```
zach-klippenstein, line 125: This property is redundant with `job` and actually introduces a whole bunch of issues because of that. Just query/cancel `job` directly.

```kotlin
private class SchedulerChannelTask(
    private val block: Runnable,
    private val job: Job
) : Disposable {
    fun execute() {
        // The task may have been disposed while it was waiting in the queue, so the
        // job's status needs to be explicitly checked before executing it.
        if (job.isActive) {
            block.run()
        }
    }

    override fun isDisposed(): Boolean = !job.isActive

    override fun dispose() {
        job.cancel()
    }
}
```

recheej:
@zach-klippenstein thank you so much for all the time spent on this review. resolved your comments. @elizarov ready for review again.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,142 @@
...
+        private val blockChannel = Channel<SchedulerChannelTask>(Channel.UNLIMITED)
+
+        init {
+            workerScope.launch {
+                while (isActive) {
+                    val task = blockChannel.receive()
+                    task.execute()
+                }
+            }
+        }
```
elizarov, line 78: Running a (potentially) concurrent Job from the constructor is a bad style. It would be better to move the code from the `init {}` block in a separate `fun start()` and call `start()` from the `createWorker` after construction.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -50,3 +176,13 @@ public class SchedulerCoroutineDispatcher(
     /** @suppress */
     override fun hashCode(): Int = System.identityHashCode(scheduler)
 }
+
+private class JobDisposable(private val job: Job) : Disposable {
+    override fun isDisposed(): Boolean = job.isCancelled || job.isCompleted
```
elizarov, line 181: You can also use `!job.isActive` here -- it is shorter and better.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,142 @@
...
+            block.run()
+        }
+    }
+
+    override fun isDisposed(): Boolean =
+        !job.isActive
+
+    override fun dispose() {
+        job.cancel()
+    }
```
elizarov, line 141: You can reduce the amount of code (both source and bytecode) by making `JobDisposable` an `open class` and making `SchedulerChannelTask` to extend `JobDisposable`, thus reusing implementations of both `isDisposed()` and `dispose()` functions.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,142 @@
...
+            schedule(block, 0, TimeUnit.MILLISECONDS)
+
+        override fun schedule(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+            if (!workerScope.isActive) return Disposables.disposed()
+
+            val newBlock = RxJavaPlugins.onSchedule(block)
+
+            val taskJob = Job(workerJob)
+
+            fun addTaskToQueue() {
```
elizarov, line 92: It would be slightly more efficient is `addTaskToQueue` is made a private function in the `DispatcherWorker` class. Actually, even better. You can just crate Alternative you can just always create `val task = SchedulerChannelTask(newBlock, taskJob)` and them replace calls to `addTaskToQueue()` with `blockChannel.offer(task)`. You would not need this function at all.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +40,405 @@ class SchedulerTest : TestBase() {
...
+            scheduler.scheduleDirect(Runnable {
+                expect(2)
+                it.resume(Unit)
+            })
+        }
+        finish(3)
+    }
+
+    @Test
+    fun tesConvertDispatcherToOriginalScheduler(): Unit = runTest {
```
elizarov, line 180: Typo `tes` -> `test`

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -31,4 +40,405 @@ class SchedulerTest : TestBase() {
...
+        val originalScheduler = Schedulers.io()
+        val dispatcher = originalScheduler.asCoroutineDispatcher()
+        val scheduler = dispatcher.asScheduler()
+        assertEquals(originalScheduler, scheduler)
+
+        finish(2)
+    }
+
+    @Test
+    fun tesConvertSchedulerToOriginalDispatcher(): Unit = runTest {
```
elizarov, line 192: Typo `tes` -> `test`

elizarov (changes requested):
❤️ It looks very good now. I cannot find any functional problems anymore. All suggestions are stylistic and efficiency-related.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: good catch.

zach-klippenstein:
If this PR is merged soon, is there a chance it could be included in the 1.4.0 release?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,133 @@
...
+/**
+ * Converts an instance of [CoroutineDispatcher] to an implementation of [Scheduler].
+ */
+public fun CoroutineDispatcher.asScheduler(): Scheduler =
+    if (this is SchedulerCoroutineDispatcher) {
+        scheduler
+    } else {
+        DispatcherScheduler(this)
+    }
+private class DispatcherScheduler(val dispatcher: CoroutineDispatcher) : Scheduler() {
```
elizarov, line 35: code style nit: an empty line is needed between a function and a class

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,133 @@
...
+
+
+/**
+ * Represents a task to be queued sequentially on a [Channel] for a [Scheduler.Worker].
+ *
+ * Delayed tasks do not block [Channel] from processing other tasks
+ */
+private class SchedulerChannelTask(
+    private val block: Runnable,
+    private val job: Job
```
elizarov, line 126: You don't need a `private val` on this `job` parameter. It is already stored by the `JobDisposable`. Just make a `DisposableJob.job` a `protected` property.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,133 @@
...
+            return taskJob.asDisposable()
+        }
+
+        override fun dispose() {
+            workerScope.cancel()
+        }
+    }
+}
+
+
```
elizarov, line 118: code style nit: one empty line is enough here

elizarov (changes requested):

reactive/kotlinx-coroutines-rx2/api/kotlinx-coroutines-rx2.api
```diff
@@ -67,7 +67,9 @@ public final class kotlinx/coroutines/rx2/RxObservableKt {
 }
 
 public final class kotlinx/coroutines/rx2/RxSchedulerKt {
-	public static final fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/rx2/SchedulerCoroutineDispatcher;
+	public static final fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/CoroutineDispatcher;
+	public static final synthetic fun asCoroutineDispatcher (Lio/reactivex/Scheduler;)Lkotlinx/coroutines/rx2/SchedulerCoroutineDispatcher;
```
recheej, line 71: @elizarov is this "synthetic" ok?

elizarov: Yes. That is because it is HIDDEN. Old code that was compiled against the previous version of the library will still work.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,138 @@
...
+    })
+
+    override fun scheduleDirect(block: Runnable): Disposable =
+        scheduleDirect(block, 0, TimeUnit.MILLISECONDS)
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+        if (!scope.isActive) return Disposables.disposed()
+        val newBlock = RxJavaPlugins.onSchedule(block)
+        return scope.launch {
+            if (delay > 0) {
```
elizarov, line 56: Nit: in this code `if (delay > 0)` is really not needed, since that's the first thing `delay` implementation checks for.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,138 @@
...
+        override fun schedule(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+            if (!workerScope.isActive) return Disposables.disposed()
+
+            val newBlock = RxJavaPlugins.onSchedule(block)
+
+            val taskJob = Job(workerJob)
+            val task = SchedulerChannelTask(newBlock, taskJob)
+
+            if (delay == 0L) {
+                blockChannel.offer(task)
```
elizarov, line 101: The docs for `schedule` say

> delay -- time to "wait" before executing the action; non-positive values indicate an non-delayed schedule

It means, it should be checking for `delay <= 0L` here.

elizarov (changes requested):
Almost done! Just minor things left.

elizarov (approved):
🚀 Good to go

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,136 @@
...
+import kotlinx.coroutines.channels.*
+import java.util.concurrent.*
+import kotlin.coroutines.*
 
 /**
  * Converts an instance of [Scheduler] to an implementation of [CoroutineDispatcher]
  * and provides native support of [delay] and [withTimeout].
  */
-public fun Scheduler.asCoroutineDispatcher(): SchedulerCoroutineDispatcher = SchedulerCoroutineDispatcher(this)
+public fun Scheduler.asCoroutineDispatcher(): CoroutineDispatcher =
```
SuperCatDev, line 19: Can you add a similar extension for RxJava3?

recheej:
@elizarov I went ahead and copied the changes over for rx3. let me know what else you think can be done.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
recheej: done!

SuperCatDev (approved):

dkhalanskyjb:
Thanks! We will edit this a bit in https://github.com/Kotlin/kotlinx.coroutines/pull/3150 before merging.

PR #1931, opened 2020-04-22
Mention that the copied exception's message must be like the original

adamschoenemann:
Just a quick fly-by PR as a suggestion on including this information.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -61,6 +61,7 @@ Exception copy logic is straightforward:
   2) If the exception class has class-specific fields not inherited from Throwable, the exception is not copied.
   3) Otherwise, one of the public exception's constructor is invoked reflectively with an optional `initCause` call.  
 
+In any case, the copied exception *must have the same* `message` field as the original exception.
 ## Debug agent
 
 [kotlinx-coroutines-debug](../kotlinx-coroutines-debug) module provides one of the most powerful debug capabilities in `kotlinx.coroutines`.


elizarov:
I am not sure such a strong wording is really needed. Nothing is going to break inside coroutines machinery is the copied exception message is different from the original, so an emphasis on "must" does not seem to be warranted here. I think it is better to rephrase it as a recommendation.

adamschoenemann:
I'm not married to the wording at all here.
But the point I wanted to come across was that it's *not* gonna attempt to restore the stack if the message is not the same as copied exception.
As per this line https://github.com/Kotlin/kotlinx.coroutines/blob/d7de5f5ba66a8d005e5cbd03b18522112303fd54/kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt#L70

elizarov:
Good point. That should be added to the docs. That is, the explanation should mentation that the same message is a prerequisite for the stack recovery mechanism.

adamschoenemann:
Exactly :)

adamschoenemann:
Should I close this MR and open an issue instead or what would you prefer?

elizarov:
It just needs a bit rephrasing and clarification on why the same message is needed. No need to close PR. It can be amended.

qwwdfsad:
Thanks! This PR triggered me to investigate the exception message comparison and uncovered #2749.

This PR is superseded by #2749

PR #1932, opened 2020-04-22
Migrate to androidx.annotation from legacy android.support.annotion

elizarov:
Legacy -> new artifact mappings: https://developer.android.com/jetpack/androidx/migrate/artifact-mappings
The most recent version: https://developer.android.com/jetpack/androidx/releases/annotation


KaiyuanZhao:
Thanks @elizarov , when will a new version be released?

PR #1935, opened 2020-04-22
kotlin.time.Duration support for the test library

EdwarDDay:


qwwdfsad:
It's currently being reworked almost from scratch and Duration will be supported out-of-the-box there

PR #1937, opened 2020-04-22
Breaking: Get rid of atomic cancellation and provide a replacement

### LINKED ISSUE #1265: Flow.collects receives items after it was cancelled on a single thread

### LINKED ISSUE #1813: Breaking: Rethink atomicity of certain low-level primitives
There is a number of primitives in the library that provide "atomic cancellation":

* [Channel.send](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-send-channel/send.html), [Channel.receive](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-receive-channel/receive.html)
* [Mutex.lock](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/lock.html), [Semaphore.acquire](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-semaphore/acquire.html)

As their documentation states:

> *Cancellation of suspended invocation is atomic* &mdash; when this function throws `CancellationException` it means that the operation was not performed. As a side-effect of atomic cancellation, a thread-bound coroutine (to some UI thread, for example) may continue to execute even after it was canceled from the same thread in the case when this operation was already resumed and the continuation was posted for execution to the thread's queue.

This is a surprising behavior that needs to be accounted for when writing higher-level primitives like `Flow`. See #1265 as an example problem and discussion in #1730.

It all points to the need to rethink the atomic cancellation behavior of the underlying primitives.

### More background on the atomic cancellation

The concept of atomic cancellation has been introduced early in the design of `kotlinx.coroutines`, before `1.0` release. Originally it was used only in channels with the goal of enabling safe ownership transfer of closeable _resources_ between different coroutines. 

> Hereinafter we use the word _resource_ to denote a reference to any object that must be explicitly closed. Usually, but not always, it encapsulates some native resource that is not managed by garbage collection. Losing a reference to such a resource creates a leak.

For example, consider a coroutine that opens a file and gets an `InputStream`. `InputStream` is a resource that must be explicitly closed after use. Now, if the coroutine needs to send it to another coroutine for processing via `val channel = Channel<InputStream>()`, then it has the following code:

```
val input = file.inputStream() // opens an input stream to a file
channel.send(input) // send for processing
```

the other coroutine receives the input stream and continues to work with it, closing it afterward:

```
val input  = channel.receive()
input.use { process(it) } // closes at the end
```

If `send` and `receive` were simply _cancellable_, as most other suspending functions created with the help of [suspendCancellableCoroutine](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html) are, then the sender would have no way of knowing if the `send` had actually completed successfully or not. It could be canceled before completing send or it could send and then get canceled. The receiver can get canceled after receiving it, too. There would be no way to ensure safe ownership transfer from one coroutine to another. As a solution to this problem the concept of "atomic cancellation" was born. In `kotlinx.coroutines` up until now, the above code is given the following guarantees:

* **Atomic send cancellation**: if `send` completes successfully then the reference was put into the channel, otherwise it was not.
* **Atomic receive cancellation**: if `receive` completes successfully then the reference was retrieved from the channel, otherwise it was not.

Together, these two guarantees support the leak-free transfer of ownership between coroutines when the code on the send side modified in a specific way like this:

```
val input = file.inputStream() // opens file
try { 
    channel.send(input)
} catch (e: Throwable) 
    input.close() // close input on any exception, we know it was not delivered
    throw e // retrow
}
```

### Public API status of atomic cancellation

The "atomic cancellation" was never widely publicized. It was briefly documented by a single paragraph in both `send` and `receive` documentation, but it was not mentioned in any of the guides concerning coroutines. Moreover, as it was shown above, the atomic cancellation itself only enables safe ownership transfer. One still has to write the code with extreme caution to ensure that resources are transferred safely between coroutines and the corresponding code one has to write was never explained in the documentation before. 

The atomic cancellation was only employed by a small number of `kotlinx.coroutines` functions and it was not possible to write user-defined primitives with similar atomic-cancellation behavior. The corresponding low-level mechanisms are internal in the library.

### Problems with atomic cancellation

Atomic cancellation creates a major hurdle when using channels in UI code. In UI application cancellation is used to manage the lifecycle of UI elements. For example, consider a coroutine that is running in the scope of some UI view:

```
val data = dataChannel.receive() // receive some data
updateUI(data)
```

When the `data` is received from the channel the line with `updateUI(data)` is not executed immediately. The corresponding continuation is scheduled for execution onto the main thread and needs to wait until the main thread is available. However, while this continuation waits in the queue the corresponding UI view might get destroyed. Normally, with all the other cancellable suspending functions that use `suspendCancellableCoroutines`, when the main thread is finally ready to execute the continuation the cancellation state of the coroutine is checked and the `CancellationException` is thrown instead of executing `updateUI(data)`. However, this check for cancellation is not performed for `Channel.receive` continuation because there is an "atomic cancellation" guarantee for `receive`. The data element was already received and must be delivered, so `updateUI(data)` will get executed despite the fact that the coroutine is already canceled. On Android, in particular, an attempt to update UI view that was already destroyed would lead to an exception. 

In practice, it means that every time a `Channel` (or another atomically-cancellable primitive) is used in the main thread, one must not forget to manually add the check for cancellation of the current coroutine:

```
val data = dataChannel.receive() // receive some data
ensureActive() // check for cancellation manually
updateUI(data)
```

Forgetting to add this check leads to hard-to-find bug that only rarely manifests itself.

This design creates irregularities in the coroutines API surface. All suspending functions in `kotlinx.coroutines` are cancellable in a regular fashion and it is guaranteed that when they resume successfully the corresponding coroutine was not canceled. However, there are a few exceptional functions with "atomic cancellation" behavior that one must remember by heart. There is no consistent naming to make them stand apart. They look like all the other suspending functions and they are cancellable, too, but they _can_ resume successfully when the coroutine was already cancellated. Tricky and error-prone.

Moreover, atomic cancellation does not make resource transfer easy. Even with atomic cancellation, it is still a tricky and error-prone endeavor. In addition to the intricate code-dance, one has to perform (as shown above), there is a problem with channel cancellation. When the receiver on the channel does not plan to continue receiving data and [cancels](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-receive-channel/cancel.html) the whole channel to indicate that, all references that were stored in the channel buffer are simply dropped. It means that channel cancellation cannot be used when channel is used to transfer resources. You can work around it when you use a channel directly (avoid cancellation, manually retrieve and close all resources from the channel you no longer need), but this makes it impossible to design resource-leak-free high-level primitives like Kotlin `Flow` that must rely on channel cancellation for their own needs.

All in all, atomic cancellation fails to deliver on the promise of being a safe solution for resource transfer from one coroutine to another.

### Additional details on Mutex and Semaphore

The atomic cancellation of `Mutex.lock` and `Semaphore.acquire` does not bring any advantages as with channels, but only all the problems. The typical pattern of mutex/semaphore use is:

```
mutex.lock()
try {
    doSomething()
} finally { 
    mutex.unlock()
}
```

The correctness of this pattern does not require that `lock` is atomically cancellable in the same way as `send`. All it needs is that a canceled `lock` attempt does not keep the mutex locked. `CancellableContinuation` already provides `resume(value) { doOnCancel() }` API to enable a correct lock implementation. See [resume](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/resume.html).

### Proposed solution

The proposal is to completely drop the concept of "atomic cancellation" from `kotlinx.coroutines` and make all the suspending functions in the library to be consistently cancellable in a regular way with a guarantee that a suspending function does not resume normally in a canceled coroutine.

In order to address the rare use-case of actually sending resources between coroutines via channels a different easy-to-use mechanism will be introduced together with this change. See #1936 for a concrete proposal.

The key rationale for this particular and radical way to address the problem is that atomic cancellation was underdocumented and very error-prone to start with. Thus, it is extremely unlikely that there is a significant amount of code in the wild that correctly exploits those atomic cancellation guarantees to perform a leak-free transfer of resources between coroutines. At the same time, changing the behavior to regular cancellation will likely fix a lot of hard-to-find cancellation bugs in the existing Android applications that use coroutines.

### Impact of the change

**This is breaking change**. Some code might have been relying on the atomic cancellation guarantees. After this change this code might encounter two kinds of errors:

* Receiving `CancellationException` from `channel.send` the code might assume that item was not sent to the channel and will go to close the corresponding resource. However, in fact, without atomic cancellation guarantee on `send`, the resource might have been actually delivered to the receiver.
* The `receive` call for a resource what was previously successfully sent channel to the channel might get `CancellationException` without atomic cancellation guarantee on `channel.receive`. Thus, the reference is lost and and the resource leaks. 

Whether this is an acceptable behavior change for a major release is TBD. The key question is whether there are any widely used libraries that could be seriously impacted by this change. We plan to publish at least one `-Mx` (milestone) release with the proposed change so that the impact of this change could be evaluted while running the actual code with the updated `kotlinx.coroutines`.

### Possible alternatives

We can try to maintain a certain degree of backward compatibility while making this change:

**Option 1**: Maintain binary backward compatibility. It means that we'll retain the old behavior (with atomic cancellation) for previously compiled code but any code that gets compiled against new version of `kotlinx.compatibility` will get new behavior (without atomic cancellation).

* **Cons**: It will vastly complicate the code and this kind of backward compatibility guarantee will be very hard to test and to support in the future.

**Option 2**: Maintain source backward compatibility. It means that old `send` and `receive` will behave as before (with atomic cancellation) and some new methods to perform regularly cancellable send and receive are introduced. It can be done by either having a dedicated `Channel` constructor to customize cancellation behavior or dedicated `send`/`receive` methods with new names.

* **Cons**: It will make the public API surface of `kotlinx.coroutines` more complex. What's worse, as was shown in the preceding overview, the whole atomic cancellation behavior is hard-to-use and error-prone anyway. It does not deserve to be maintained in the long run, so the corresponding "old" methods that perform atomically-cancellable `send`/`receive` will have to be deprecated, meaning a transition to less natural names.

### LINKED ISSUE #1915: Tweak CancellableContinuation.invokeOnCancellation contract for consistency
Consider the following code that uses [suspendCancellableCoroutine](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html): 

```
val result = suspendCancellableCoroutiune<Unit> { cont: CancellableContinuation<Unit> ->
   // call later, asynchronously
   cont.resume(Unit)
}
```

This call to `suspendCancellableCoroutiune` will not necessarily produce `result == Unit`. It may also end up throwing `CancellationException`. While indistinguishable from the outside, internally there are two conceptually different cases when `CancellationExcption` could be produced:

* **Case A**: The coroutine running this code may get canceled before the call to `cont.resume(Unit)`.
* **Case B**: The coroutine running this code may get canceled after the call to `cont.resume(Unit)`, but before the continuation gets dispatched to the caller.

> Note, that in a general case the caller of `cont.resume(Unit)` cannot distinguish these two cancellation cases either, because coroutine cancellation is asynchronous and may race with the call to `cont.resume(Unit)`. The latter does not provide any indication to the caller whether cancellation or resumption came first. There's only an _internal API_ called `tryResume` that can be used to find it out.

Now there are two ways to install a handler that gets invoked on cancellation:

* [`cont.invokeOnCancellation { ... }`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/invoke-on-cancellation.html) (stable) &mdash; continuation cancellation handler.
* [`cont.resume(Unit) { ... }`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/resume.html) (experimental) &mdash; resume cancellation handler.

They are currently not consistent between themselves with respect as to which cancellation cases they react to:

*  `cont.invokeOnCancelltion { ... }` continuation handler only gets invoked in case A of cancellation. If `resume` happened before cancellation, but the coroutine was canceled later (case B), then the continuation cancellation handler is not called.
* `cont.resume(Unit) { ... }` cancellation handler gets invoked in both cases. 

The proposal is to make them consistent so that both are invoked in both cases. While this is technically a breaking change, it is not clear what kind of correctly working code this kind of change can break, since the difference in behavior would only manifest in cases when `cont.resume` was called after cancellation, but there is no way to figure out whether `cancel` or `resume` was first anyway. 

There is a motivation beyond consistency, too. This inconsistency turned out while working on #1813. When the prototype without atomic cancellation was implemented, the implementation of `Semaphore` stoped to work correctly in case of `acquire` cancellation. The root of the problem is that with atomic cancellation the case B was not possible and there was no need to worry about it. However, making the proposed change to `invokeOnCancellation` (so that it is called in case B) fixed `Semaphore`. This leads us to believe that this change may also eliminate some `cancel`/`resume` races that currently exist in 3rd party code, too, because 3rd party code did not have access to atomic cancellation so it did not have any means to distinguish two cases of cancellation and was likely incorrect in racing scenarios.

### LINKED ISSUE #1936: Support leak-free closeable resources transfer via channel
The change in #1813 removes support for "atomic cancellation", making it impossible to transfer a resource via a channel from one coroutine to another. When a closeable resource (like open file or a handle to another native resource) is transferred via channel from one coroutine to another it can be lost if either send or receive operation are cancelled in transit.

We introduce the following replacement. A `Channel()` constructor function gets and additional optional parameter: `onElementCancel: ((E) -> Unit)? = null` (the actual name is TBD, feedback is welcome). 

When `onElementCancel` option parameter is set, the corresponding function is called once for each element that was sent to the channel and is being lost due to cancellation, which can happen in the following cases:
 
* `send` operation was cancelled before it had a chance to actually send the element.
* `receive` operation retrieved the element from the channel but was cancelled when trying to return it the caller.
* The channel was cancelled, in which case `onElementCancel` is called on every remaining element in the channel's buffer.

Note, that `onElementCancel` is called synchronously in an arbitrary context. It should be fast, non-blocking, and should not throw exceptions. Any exception thrown by `onElementCancel` is wrapped into an internal runtime exception which is either rethrown or handed off to the exception handler in the current context (`CoroutineExceptionHandler`) when one is available.

elizarov:
This is problematic for Android when `Main` dispatcher is cancelled on destroyed activity. Atomic nature of channels was designed to prevent loss of elements, which is really not an issue for a typical application, but creates problem when used with channels. Atomic cancellation for channel is replaced with the "prompt cancellation guarantee".

Fixes #1813 (explains the problem in detail)
Fixes #1265
Fixes #1915
Fixes #1936


kotlinx-coroutines-core/common/src/internal/DispatchedContinuation.kt
```diff
@@ -194,29 +194,38 @@ internal class DispatchedContinuation<in T>(
...
+    }
+
     @Suppress("NOTHING_TO_INLINE")
-    inline fun resumeCancelled(): Boolean {
+    inline fun resumeCancelled(state: Any?): Boolean {
         val job = context[Job]
         if (job != null && !job.isActive) {
-            resumeWithException(job.getCancellationException())
+            val cause = job.getCancellationException()
+            cancelCompletedResult(state, cause)
```
elizarov, line 225: Need to figure out what to do if `onCancellation` throws exception here (this is not a safe call).

kotlinx-coroutines-core/common/src/internal/DispatchedTask.kt
```diff
@@ -22,7 +48,10 @@ internal abstract class DispatchedTask<in T>(
 
     internal abstract fun takeState(): Any?
 
-    internal open fun cancelResult(state: Any?, cause: Throwable) {}
+    /**
+     * Called when this task was cancelled while it was being dispatched.
+     */
+    internal open fun cancelCompletedResult(takenState: Any?, cause: Throwable) {}
```
elizarov, line 54: Need to figure out if `cancelCompletedResult` can throw an exception (and caller must catch) or it takes care about all the 3rd-party exceptions itself.

kotlinx-coroutines-core/common/src/internal/ElementCancel.kt
```diff
@@ -0,0 +1,33 @@
...
+}
+
+internal fun <E> OnElementCancel<E>.callElementCancel(resource: E, context: CoroutineContext) {
+    callElementCancelCatchingException(resource, null)?.let { ex ->
+        handleCoroutineException(context, ex)
+    }
+}
+
+internal fun <E> OnElementCancel<E>.bindCancellationFun(element: E, context: CoroutineContext): (Throwable) -> Unit =
+    { _: Throwable -> callElementCancel(element, context) }
```
elizarov, line 31: TBD: Can we avoid allocating this lambda every time channel with `onElementCancel` resumes a receive operation?

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -355,24 +349,28 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
...
+        mutex: MutexImpl,
         owner: Any?,
         @JvmField val cont: CancellableContinuation<Unit>
-    ) : LockWaiter(owner) {
-        override fun tryResumeLockWaiter() = cont.tryResume(Unit)
+    ) : LockWaiter(mutex, owner) {
+        override fun tryResumeLockWaiter() = cont.tryResume(Unit, idempotent = null) {
+            // if this continuation get's cancelled during dispatch to the caller, then release the lock
+            mutex.unlock(owner)
+        }
```
elizarov, line 359: We allocate an extra lambda on each resume of the lock waiter. This might be Ok (this is kind of slow path) but needs additional thought.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -307,30 +356,87 @@ internal open class CancellableContinuationImpl<in T>(
...
+    ): Any? = when {
+        proposedUpdate is CompletedExceptionally -> {
+            assert { idempotent == null } // there are no idempotent exceptional resumes
+            assert { onCancellation == null } // only successful results can be cancelled
+            proposedUpdate
+        }
+        !resumeMode.isCancellableMode && idempotent == null -> proposedUpdate // cannot be cancelled in process, all is fine
+        onCancellation != null || state is CancelHandler || idempotent != null ->
+            // mark as CompletedContinuation if special cases are present (cancellation handlers or idempotent resume)
+            CompletedContinuation(proposedUpdate, state as? CancelHandler, onCancellation, idempotent)
```
elizarov, line 374: TBD: WARNING: A `CancellableContinuation` with result and a `CancelHandler` (that was installed via `invokeOnCancellation`) now becomes more expensive, since now `CompletedContinuation` wrapper is created in this case. It might affect performance of channels, which use `invokeOnCancellation` and might need a special workaround here, since we know that channels do not need "extended" `invokeOnCancellation` contract.

kotlinx-coroutines-core/common/src/internal/DispatchedContinuation.kt
elizarov: Ok here. Added tests.

kotlinx-coroutines-core/common/src/internal/DispatchedTask.kt
elizarov: Figured out that it always takes care of them. Added tests.

kotlinx-coroutines-core/common/src/internal/ElementCancel.kt
elizarov: Don't care about this case for now.

kotlinx-coroutines-core/common/test/CancellableResumeTest.kt
```diff
@@ -118,25 +118,30 @@ class CancellableResumeTest : TestBase() {
             try {
                 suspendCancellableCoroutine<String> { cont ->
                     expect(3)
-                    // resumed first, then cancelled, so no invokeOnCancellation call
-                    cont.invokeOnCancellation { expectUnreached() }
+                    // resumed first, dispatched, then cancelled, bit still got invokeOnCancellation call
```
LouisCAD, line 121: Minor typo
```suggestion
                    // resumed first, dispatched, then cancelled, but still got invokeOnCancellation call
```

kotlinx-coroutines-core/common/src/sync/Mutex.kt
elizarov: Does not really make a big difference.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
elizarov: It does not seem to produce a visible impact on benchmarks

elizarov:
Ready for review.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -832,7 +850,8 @@ internal abstract class AbstractChannel<E> : AbstractSendChannel<E>(), Channel<E
                     return@sc
                 }
                 if (result !== POLL_FAILED) {
-                    cont.resume(true)
+                    @Suppress("UNCHECKED_CAST")
+                    cont.resume(true, channel.onElementCancel?.bindCancellationFun(result as E, cont.context))
```
qwwdfsad, line 854: Here is a general observation regarding `bindCancellationFun` and `onElementCancel`:

1) Both functions are invoked **rarely**, so the general performance cost doesn't matter here and allocation wrapper-lambda is ok
2) Most of the time `onElementCancel` is `null`.
3) Code with a lot of null-checks eats bytecode size here and there.
4) Maintaining code where `onElementCancel` always has to be wrapped into `bind` is not really pleasant


So my proposal is:
1) Create `bindCancellationFun` in the channel constructor. It will simplify our maintenance cost.
2) For regular channels, use singleton instance of `onElementCancel`. It will save our question marks and bytecode size.

After that, the code should become a bit more readable and smaller by the cost of additional allocation in a rare case of channel with resources

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -254,17 +259,23 @@ internal open class ArrayChannel<E>(
...
                 }
-                size.value = 0
+                buffer[head] = EMPTY
+                head = (head + 1) % buffer.size
             }
+            size.value = 0
         }
         // then clean all queued senders
         super.onCancelIdempotent(wasClosed)
+        elementCancelException?.let { throw it } // throw cancel exception at the end if there was one
```
qwwdfsad, line 278: This behaviour is inconsistent with `cancel` over a regular channel.
Why one rethrows an exception, while the other reports it to global exception handler?
I'd say both should throw an exception from the `cancel` method

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
```diff
@@ -254,17 +259,23 @@ internal open class ArrayChannel<E>(
...
-            lock.withLock {
-                repeat(size.value) {
-                    buffer[head] = 0
-                    head = (head + 1) % buffer.size
+        val onElementCancel = onElementCancel
+        var elementCancelException: ElementCancelException? = null // first cancel exception, others suppressed
+        lock.withLock {
+            repeat(size.value) {
+                val value = buffer[head]
+                if (value !== EMPTY && onElementCancel != null) {
```
qwwdfsad, line 267: [nit] `onElementCancel != null && value !== EMPTY` is a bit faster and reflects the intent better

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -120,12 +115,20 @@ internal open class ConflatedChannel<E> : AbstractChannel<E>() {
...
-            }
+        var elementCancelException: ElementCancelException? = null // resource cancel exception
+        lock.withLock {
+            elementCancelException = updateValueLocked(EMPTY)
         }
         super.onCancelIdempotent(wasClosed)
+        elementCancelException?.let { throw it } // throw exception at the end if there was one
+    }
+
+    private fun updateValueLocked(element: Any?): ElementCancelException? {
```
qwwdfsad, line 126: Honestly, it's hard to imagine that someone creates a channel of resources that are being conflated and thrown away. Such solution is probably a bad design and should be avoided (+potentially blocking close() under the lock that blocks even concurrent `offer`s to the same channel!)

Let's prohibit non-null `onElementCancel` in conflated channels

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
```diff
@@ -54,7 +49,7 @@ internal open class ConflatedChannel<E> : AbstractChannel<E>() {
                     }
                 }
             }
-            value = element
+            updateValueLocked(element)?.let { throw it }
```
qwwdfsad, line 52: Regular channels report these exceptions to global exception handler, while this one rethrows an exception.
Let's stick to a consistent approach

qwwdfsad (changes requested):
I've reviewed only channels, the second part of review will arrive later.

I've also found a ~20% perf regressions on `ChannelSinkBenchmark` that doesn't seem to be expected. Fast perf session has shown that new version allocates 30% more memory (in bytes) and has a serious dominator `kotlinx.coroutines.CompletedContinuation` in `tryResumeImpl`.
Let's fix that and see what else may be seriously affected

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -163,25 +181,126 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
+ *         // A callback provides a reference to some closeable resource
+ *         override fun onCompleted(resource: T) {
+ *             // Resume coroutine with a value provided by the callback and ensure the resource is closed in case
+ *             // when the coroutine is cancelled before the caller gets a reference to the resource.
+ *             continuation.resume(resource) {
+ *                 resource.close() // Close the resource on cancellation
+ *             }
+ *         }
+ *     // ...
+ * }
```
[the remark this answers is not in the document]
qwwdfsad: Also, a wild idea: hide & deprecate existing `resume {}` member and provide it as an extension to reduce interface API shape and have a less code to maintain

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -361,24 +358,28 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
     }
 
     private class LockCont(
+        mutex: MutexImpl,
```
qwwdfsad, line 361: can be plain `Mutex` here

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -361,24 +358,28 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
...
         owner: Any?,
         @JvmField val cont: CancellableContinuation<Unit>
-    ) : LockWaiter(owner) {
-        override fun tryResumeLockWaiter() = cont.tryResume(Unit)
+    ) : LockWaiter(mutex, owner) {
+        override fun tryResumeLockWaiter() = cont.tryResume(Unit, idempotent = null) {
```
qwwdfsad, line 365: You can avoid allocation here by making `LockCont` implement `(cause: Throwable) -> Unit` and passing `this` to the second parameter

kotlinx-coroutines-core/jvm/test/sync/MutexStressTest.kt
```diff
@@ -26,4 +26,34 @@ class MutexStressTest : TestBase() {
...
+        val mutex = Mutex(true) // create a locked mutex
+        newSingleThreadContext("SemaphoreStressTest").use { pool ->
+            repeat (n) {
+                // Initially, we hold the lock and no one else can `lock`,
+                // otherwise it's a bug.
+                assertTrue(mutex.isLocked)
+                var job1EnteredCriticalSection = false
+                val job1 = launch(start = CoroutineStart.UNDISPATCHED) {
+                    mutex.lock()
+                    job1EnteredCriticalSection = true
```
qwwdfsad, line 42: Replacing it with
```
mutex.onLock {
    job1EnteredCriticalSection = true
    mutex.unlock()
}
```

triggers test failure

qwwdfsad (changes requested):

kotlinx-coroutines-core/jvm/test/sync/MutexStressTest.kt
elizarov: Fixed

kotlinx-coroutines-core/common/src/sync/Mutex.kt
elizarov: Cannot do it. It will not work on JS.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -163,25 +181,126 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
      * There is no guarantee on the execution context of its invocation.
      */
     @ExperimentalCoroutinesApi // since 1.2.0
-    public fun resume(value: T, onCancellation: (cause: Throwable) -> Unit)
+    public fun resume(value: T, onCancellation: ((cause: Throwable) -> Unit)?)
 }
 
 /**
  * Suspends the coroutine like [suspendCoroutine], but providing a [CancellableContinuation] to
  * the [block]. This function throws a [CancellationException] if the coroutine is cancelled or completed while suspended.
```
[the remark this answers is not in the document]
elizarov: I've clarified the meaning of this doc and added a test: https://github.com/Kotlin/kotlinx.coroutines/commit/a2339f29bd47051a841153b380030277725b1aeb

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -163,25 +181,126 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
+ *
+ * A typical use of this function is to suspend a coroutine while waiting for a result
+ * from a single-shot callback API and to return the result to the caller.
+ * For multi-short callback APIs see [callbackFlow][kotlinx.coroutines.flow.callbackFlow].
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -163,25 +181,126 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
+ * > concurrently with the call of the callback.
+ *
+ * ### Prompt cancellation guarantee
+ *
+ * This function provides **prompt cancellation guarantee**.
+ * If the [Job] of the current coroutine was cancelled while this function was suspended it will not resume
+ * successfully.
+ *
+ * The cancellation of the coroutine's job is generally asynchronous with respect to the suspended coroutine.
+ * The suspended coroutine is resumed via the call it to its [Continuation.resumeWith] member function or to
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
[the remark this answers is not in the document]
elizarov: I've tried to get rid of `resume { ... }`. It is not easy. For example, Channels install `invokeOnCancellation {}` to remove a node from lists, there is not value at this point. When calling `resume {}` they already have a values, but the `invokeOnCancellation {}` is already set. We can get rid of `resume { ... }` if we either allow multiple `invokeOnCancellation {}` handler or say that a second `invokeOnCancellation` overwrites a previously installed one.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -199,40 +318,24 @@ public suspend inline fun <T> suspendCancellableCoroutine(
...
-
-/**
- *  Suspends coroutine similar to [suspendAtomicCancellableCoroutine], but an instance of [CancellableContinuationImpl] is reused if possible.
- */
-internal suspend inline fun <T> suspendAtomicCancellableCoroutineReusable(
+internal suspend inline fun <T> suspendCancellableCoroutineReusable(
     crossinline block: (CancellableContinuation<T>) -> Unit
 ): T = suspendCoroutineUninterceptedOrReturn { uCont ->
-    val cancellable = getOrCreateCancellableContinuation(uCont.intercepted())
+    val cancellable = getOrCreateCancellableContinuation(uCont.intercepted(), resumeMode = MODE_CANCELLABLE_REUSABLE)
```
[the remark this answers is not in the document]
elizarov: Done

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -199,40 +318,24 @@ public suspend inline fun <T> suspendCancellableCoroutine(
...
+    delegate: Continuation<T>, resumeMode: Int
+): CancellableContinuationImpl<T> {
+    assert { resumeMode.isReusableMode }
     // If used outside of our dispatcher
     if (delegate !is DispatchedContinuation<T>) {
-        return CancellableContinuationImpl(delegate, resumeMode = MODE_ATOMIC_DEFAULT)
+        return CancellableContinuationImpl(delegate, resumeMode)
     }
     /*
      * Attempt to claim reusable instance.
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/CompletionState.kt
```diff
@@ -21,6 +28,11 @@ internal fun <T> recoverResult(state: Any?, uCont: Continuation<T>): Result<T> =
     else
         Result.success(state as T)
 
+internal data class CompletedWithCancellation(
+    @JvmField val result: Any?,
+    @JvmField val onCancellation: ((cause: Throwable) -> Unit)
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -508,6 +493,44 @@ public interface ChannelIterator<out E> {
...
  *   [Sending][send] suspends only when the buffer is full, and [receiving][receive] suspends only when the buffer is empty.
+ *
+ * ### Prompt cancellation guarantee
+ *
+ * All suspending functions with channels provide **prompt cancellation guarantee**.
+ * If the job was cancelled while send or receive function was suspended, it will not resume successfully.
+ * If the [send][SendChannel.send] call threw [CancellationException] there is no way
+ * to tell if the element was already sent to the channel or not.
+ * If the [receive][ReceiveChannel.receive] call threw [CancellationException] there is no way
+ * to tell if some element was already received from the channel or not.
```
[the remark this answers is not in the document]
elizarov: Rephrased: https://github.com/Kotlin/kotlinx.coroutines/commit/f1ed47eeec7b41ca8ab04a8f8259e20c3115cf23

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -508,6 +493,44 @@ public interface ChannelIterator<out E> {
...
+ * When `onElementCancel` parameter is set, the corresponding function is called once for each element
+ * that was sent to the channel and is being lost due to cancellation, which can happen in the following cases:
+ *
+ * * [send][SendChannel.send] operation was cancelled before it had a chance to actually send the element.
+ * * [receive][ReceiveChannel.receive] operation retrieved the element from the channel but was cancelled when trying
+ *   to return it the caller.
+ * * The channel was [cancelled][ReceiveChannel.cancel], in which case `onElementCancel` is called on every
+ *   remaining element in the channel's buffer.
+ *
+ * Note, that `onElementCancel` function is called synchronously in an arbitrary context. It should be fast, non-blocking,
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -20,6 +20,9 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
     /** @suppress **This is unstable API and it is subject to change.** */
     protected val queue = LockFreeLinkedListHead()
 
+    open val onElementCancel: ((E) -> Unit)?
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/channels/LinkedListChannel.kt
```diff
@@ -17,7 +17,7 @@ import kotlinx.coroutines.selects.*
  *
  * @suppress **This an internal API and should not be used from general code.**
  */
-internal open class LinkedListChannel<E> : AbstractChannel<E>() {
+internal open class LinkedListChannel<E>(onElementCancel: ((E) -> Unit)?) : AbstractChannel<E>(onElementCancel) {
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -1010,8 +1044,9 @@ internal interface ReceiveOrClosed<in E> {
  * Represents sender for a specific element.
  */
 @Suppress("UNCHECKED_CAST")
-internal class SendElement(
-    override val pollResult: Any?,
+internal class SendElement<E>(
+    override val pollResult: E,
+    @JvmField val onElementCancel: ((E) -> Unit)?,
```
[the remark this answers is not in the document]
elizarov: Did not work out: `SendElement.cancelElement()` is called from `SendElement.remove()` which overrides a `remove()` method in LinkedList.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
elizarov: `bindCancellationFun` is doing an actual _bind_ -- it captures the resulting value, so that the `onUnderlivedElement` function can be called for it in case of cancellation.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -194,8 +190,8 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
         return lockSuspend(owner)
     }
 
-    private suspend fun lockSuspend(owner: Any?) = suspendAtomicCancellableCoroutineReusable<Unit> sc@ { cont ->
-        val waiter = LockCont(owner, cont)
+    private suspend fun lockSuspend(owner: Any?) = suspendCancellableCoroutineReusable<Unit> sc@ { cont ->
+        val waiter = LockCont(this, owner, cont)
```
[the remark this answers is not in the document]
elizarov: Done!

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
elizarov: That would be weird from a composability standpoint, e.g. when a certain combo on `Channel(...)` constructor params is prohibited.  Too many things to document. Also, when we start supporting `onUndeliveredElement` in flows (in a similar way to reactive `onDiscard`) it would immediately show itself up. Let's discuss this separately.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
elizarov: LockCont is an inner class now, not an issue anymore.

kotlinx-coroutines-core/common/src/channels/ConflatedChannel.kt
elizarov: There's no global exception handler to report the exception to here. Channel has no context on its own. In other places we have an inverse situation. E.g., during resume we cannot just rethow exception (since execution stack is arbitrary and unrelated to the issue), but we have a context with an exception handler on our hands.

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
elizarov: Fixed

kotlinx-coroutines-core/common/src/channels/ArrayChannel.kt
elizarov: There's no global exception handler to report the exception in this `cancel`. Channel has no context on its own. In a cancel method, though, the caller up the stack is unreleated to the channel (it is some code cancelling a job), so it makes not sense to force it to handle an exception, but we have a context with an exception handler on our hands.

kotlinx-coroutines-core/common/src/channels/ArrayBroadcastChannel.kt
```diff
@@ -180,6 +180,8 @@ internal class ArrayBroadcastChannel<E>(
                             this.tail = tail + 1
                             return@withLock // go out of lock to wakeup this sender
                         }
+                        // too late, already cancelled, but we removed it from the queue and need to cancel resource
+                        send!!.cancelElement()
```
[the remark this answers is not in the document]
elizarov: Fixed

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -431,7 +434,7 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
     // ------ private ------
 
     private class SendSelect<E, R>(
-        override val pollResult: Any?,
+        override val pollResult: E,
```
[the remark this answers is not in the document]
elizarov: It is actually "the result that pollInternal returns when it rendezvous with this send", so the naming is correct. Added a comment.

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -165,10 +168,10 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
         return closed.sendException
     }
 
-    private suspend fun sendSuspend(element: E): Unit = suspendAtomicCancellableCoroutineReusable sc@ { cont ->
+    private suspend fun sendSuspend(element: E): Unit = suspendCancellableCoroutineReusable sc@ { cont ->
         loop@ while (true) {
             if (isFullImpl) {
-                val send = SendElement(element, cont)
+                val send = SendElement(element, onElementCancel, cont)
```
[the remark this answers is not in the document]
elizarov: `offer` is fine -- it return `false` when it fails to send, it does not need to notify `onUndeliveredElement`. Added comment.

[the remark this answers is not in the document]
elizarov: 👍 Great catch on the case of cancelled/closed channel, though. Added this stress-test, code, and tests.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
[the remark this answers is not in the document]
qwwdfsad: Filed https://github.com/Kotlin/kotlinx.coroutines/issues/2277. Let's leave it as is in this review

qwwdfsad:
I still have 20% performance regression on this particular benchmark: ` ChannelSinkBenchmark.channelPipelineOneThreadLocal`. What's the source of it?

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -232,64 +272,77 @@ internal open class CancellableContinuationImpl<in T>(
...
+                is CompletedContinuation -> {
+                    /*
+                     * Continuation was already completed, and might already have cancel handler.
+                     */
+                    if (state.cancelHandler != null) multipleHandlersError(handler, state)
+                    // BeforeResumeCancelHandler does not need to be called on a completed continuation
+                    if (cancelHandler is BeforeResumeCancelHandler) return
+                    if (state.cancelled) {
+                        // todo: extra layer of protection against the second invokeOnCancellation
+                        // if (!state.makeHandled()) multipleHandlersError(handler, state)
```
qwwdfsad, line 329: This should be uncommented back, shouldn't it?

qwwdfsad (changes requested):
Please take a look at benchmarks one more time

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -151,32 +153,40 @@ internal abstract class AbstractSendChannel<E> : SendChannel<E> {
...
+            result is Closed<*> -> {
+                throw recoverStackTrace(helpCloseAndGetSendException(element, result))
             }
-            result is Closed<*> -> throw recoverStackTrace(helpCloseAndGetSendException(result))
             else -> error("offerInternal returned $result")
         }
     }
 
-    private fun helpCloseAndGetSendException(closed: Closed<*>): Throwable {
+    private inline fun helpCloseAndGetSendException(element: E, closed: Closed<*>): Throwable {
```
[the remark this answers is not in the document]
elizarov: Good catch. Refactoring leftover. Don't have to be inline in this version.

kotlinx-coroutines-core/common/src/internal/OnUndeliveredElement.kt
```diff
@@ -0,0 +1,41 @@
...
+
+internal fun <E> OnUndeliveredElement<E>.callUndeliveredElementCatchingException(
+    element: E,
+    undeliveredElementException: UndeliveredElementException? = null
+): UndeliveredElementException? {
+    try {
+        invoke(element)
+    } catch (ex: Throwable) {
+        if (undeliveredElementException != null) {
+            undeliveredElementException.addSuppressedThrowable(ex)
```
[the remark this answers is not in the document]
elizarov: Will not harm, tho. Added.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -557,17 +615,25 @@ public interface Channel<E> : SendChannel<E>, ReceiveChannel<E> {
...
+        UNLIMITED -> LinkedListChannel(onUndeliveredElement)
+        CONFLATED -> ConflatedChannel(onUndeliveredElement)
+        BUFFERED -> ArrayChannel(CHANNEL_DEFAULT_CAPACITY, onUndeliveredElement)
+        else -> ArrayChannel(capacity, onUndeliveredElement)
     }
 
+/**
+ * @suppress Binary compatibility only, should not be documented
+ */
+@Deprecated(level = DeprecationLevel.HIDDEN, message = "Binary compatibility")
```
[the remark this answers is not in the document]
elizarov: 👍

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -164,7 +167,7 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
         acquireSlowPath()
     }
 
-    private suspend fun acquireSlowPath() = suspendAtomicCancellableCoroutineReusable<Unit> sc@ { cont ->
+    private suspend fun acquireSlowPath() = suspendCancellableCoroutineReusable<Unit> sc@ { cont ->
```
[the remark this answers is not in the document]
elizarov: Which one in particular? `SemaphoreStressTest` has most of the same tests as `MutexStressTest` with the exception of ones supporting `select {}` (which Mutex does not support).

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -203,7 +206,8 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
         // On CAS failure -- the cell must be either PERMIT or BROKEN
         // If the cell already has PERMIT from tryResumeNextFromQueue, try to grab it
         if (segment.cas(i, PERMIT, TAKEN)) { // took permit thus eliminating acquire/release pair
-            cont.resume(Unit)
+            val result = cont.tryResumeAcquire()
```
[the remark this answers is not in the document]
elizarov: We don't have to! Refactoring leftover. Fixed, added comment.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -171,10 +189,10 @@ internal open class CancellableContinuationImpl<in T>(
             val update = CancelledContinuation(this, cause, handled = state is CancelHandler)
             if (!_state.compareAndSet(state, update)) return@loop // retry on cas failure
             // Invoke cancel handler if it was present
-            if (state is CancelHandler) invokeHandlerSafely { state.invoke(cause) }
+            (state as? CancelHandler)?.let { callCancelHandler(it, cause) }
             // Complete state update
             detachChildIfNonResuable()
-            dispatchResume(mode = MODE_ATOMIC_DEFAULT)
+            dispatchResume(mode = MODE_ATOMIC) // no need for additional cancellation checks
```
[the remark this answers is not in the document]
elizarov: 👍  Actually, it leads to extra hoops of copying the state with this exception to dispatched continuation instead directly scheduling this CancellableContinuationImpl instance (which is also s RunnableTask) and, accidentally, it was working around an inconsistent code in stack-trace recovery that was only calling stacktrace recovery in this specific case, but should have been doing it in advance. Fixed.

kotlinx-coroutines-core/common/src/internal/DispatchedContinuation.kt
```diff
@@ -19,7 +19,7 @@ internal val REUSABLE_CLAIMED = Symbol("REUSABLE_CLAIMED")
 internal class DispatchedContinuation<in T>(
     @JvmField val dispatcher: CoroutineDispatcher,
     @JvmField val continuation: Continuation<T>
-) : DispatchedTask<T>(MODE_ATOMIC_DEFAULT), CoroutineStackFrame, Continuation<T> by continuation {
+) : DispatchedTask<T>(MODE_ATOMIC), CoroutineStackFrame, Continuation<T> by continuation {
```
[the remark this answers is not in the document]
elizarov: Added invalid marker here to make the code clearer.

kotlinx-coroutines-core/common/src/internal/DispatchedContinuation.kt
```diff
@@ -180,10 +180,10 @@ internal class DispatchedContinuation<in T>(
         val state = result.toState()
         if (dispatcher.isDispatchNeeded(context)) {
             _state = state
-            resumeMode = MODE_ATOMIC_DEFAULT
+            resumeMode = MODE_ATOMIC
             dispatcher.dispatch(context, this)
         } else {
-            executeUnconfined(state, MODE_ATOMIC_DEFAULT) {
+            executeUnconfined(state, MODE_ATOMIC) {
```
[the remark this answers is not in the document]
elizarov: Yes. It is used for `launch(start = ATOMIC)` in `Unconfined` dispatcher. There is a failing test if it is changed.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -88,20 +88,22 @@ internal open class CancellableContinuationImpl<in T>(
...
         assert { state !is NotCompleted }
-        if (state is CompletedIdempotentResult) {
+        if (state is CompletedContinuation && state.idempotentResume != null) {
+            // Cannot reuse continuation that was resumed with idempotent marker
             detachChild()
             return false
         }
         _decision.value = UNDECIDED
         _state.value = Active
+        this.resumeMode = resumeMode
```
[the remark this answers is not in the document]
elizarov: Fixed. Also renamed it to `resetStateReusable` and added assertion that the continuation we use it on has `MODE_CANCELLABLE_REUSABLE`.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
elizarov: I've just removed it. It does not need this extra check anymore as "mulitipleHandlersError" check is already done a few lines above. I've added more tests on various "double invokeOnCancalltion call" cases, too.

elizarov:
**Update w.r.t benchmarks**. With this commit https://github.com/Kotlin/kotlinx.coroutines/commit/164d3e5f60da6ea3994a876b59f539eed3600638 the memory allocation rate (in bytes per op) on the `ChannelSinkBenchmark.channelPipelineOneThreadLocal` is now the same as it is in the main branch of the code. However, it is still slower.

qwwdfsad (approved):
Great! Good to go 🚀

Issue #1816, opened 2020-02-19
ConflatedBroadcastChannel should deliver last value


pacher:
For `ConflatedChannel` the problem was discussed in #332, addressed in #1235 and fixed by #1239,
but not for `ConflatedBroadcastChannel`

Slightly modified example from #1235:
```kotlin
fun main() = runBlocking {
    val channel = Channel<Int>(capacity = Channel.CONFLATED)
    with(channel) {
        send(1); send(2); send(3); send(4)
        close()
    }
    println(channel.receive())
}
```
Correctly prints `4`, but
```kotlin
fun main() = runBlocking {
    val channel = ConflatedBroadcastChannel<Int>()
    with(channel) {
        send(1); send(2); send(3); send(4)
        close()
    }
    println(channel.openSubscription().receive())
}
```
throws `ClosedReceiveChannelException`

Is it by design?

elizarov:
It is the design now, but we can change it. What is your use-case for `ConflatedBroadcastChannel` where this difference in behavior becomes important?

fluidsonic:
I just ran into the same issue.

I have a `ConflatedBroadcastChannel` for emitting `DownloadStatus` (initial, running[…], completed).

```kotlin
class DownloadCoordinator {
   private val stateChannel = ConflatedBroadcastChannel<DownloadState>(DownloadState.Initial)
   val state = stateChannel.asFlow()
   // etc.
}
```

Emitting looks like this:
- initial
- running (10%)
- …
- running (90%)
- complete
- **close**

Now a consumer that's interested in the latest state of the download simply does this:

```kotlin
downloadCoordinator.state.first()
```

It was totally unexpected to me that `.first()` doesn't work on a `ConflatedBroadcastChannel`-backed `Flow` once it's closed:
```
NoSuchElementException: Expected at least one element
```

Not even `val state = stateChannel.asFlow().conflated()` seems to help 🤔

qwwdfsad:
Thanks for the detailed explanation.
It seems like this particular problem will be resolved with stateful flow

elizarov:
It is being provided by implementation in #1974

pacher:
@elizarov Sorry for late reply in these difficult times
My use case is like in #332, that is why I did not specify it. More specifically second use case in your  [comment](https://github.com/Kotlin/kotlinx.coroutines/issues/332#issuecomment-405903512).
[...]

PR #1974, opened 2020-04-29
StateFlow implementation

### LINKED ISSUE #1973: Introduce StateFlow
We need to be able to conveniently use `Flow` to represent an updateable state in applications. This change introduces `StateFlow` -- an updateable value that represents a state and _is_ a `Flow`. The design is flexible to fit a variety of needs:

* `StateFlow<T>` interface is a _read-only_ view that gives access to the current `value` and implements a `Flow<T>` to collect updates to the values. 
* `MutabaleStateFlow<T>` interface adds value-modification operation.

A `MutableStateFlow(x)` constructor function is provided. It returns an implementation of `MutableStateFlow` with the given initial value. It can be exposed to the outside world as either `StateFlow<T>` if fast non-reactive access to the `value` is needed, or as `Flow<T>` if only reactive view of updates to the value is needed.

Core state flow API can be summarized like this:

```kotlin
package kotlinx.coroutines.flow

interface StateFlow<T> : Flow<T> {
    val value: T // always availabe, reading it never fails
}

interface MutableStateFlow<T> : StateFlow<T> {
    override var value: T // can read & write value
}

fun <T> MutableStateFlow(value: T): MutableStateFlow<T> // constructor fun
```

Implementation is available in PR #1974.

 ### StateFlow vs ConflatedBroadcastChannel

Conceptually state flow is similar to `ConflatedBroadcastChannel`  and is designed to completely replace `ConflatedBroadcastChannel` in the future. It has the following important improvements:

* `StateFlow` is simpler because it does not have to implement all the `Channel` APIs, which allows for faster, garbage-free implementation, unlike `ConflatedBroadcastChannel` implementation that allocates objects on each emitted value.
* `StateFlow` always has a value that can be safely read at any time via `value` property. Unlike `ConflatedBroadcastChannel`, there is no way to create a state flow without a value.
* `StateFlow` has a clear separation into a read-only `StateFlow` interface and a `MutableStateFlow`.
* `StateFlow` conflation is based on equality, unlike conflation in `ConflatedBroadcastChannel` that is based on reference identity. It is a stronger, more practical conflation policy, that avoids extra updates when data classes are emitted. You can consider it to have an embedded `distinctUntilChanged` out-of-the-box.
* `StateFlow` cannot be currently closed like `ConflatedBroadcastChannel` and can never represent a failure. This feature might be added in the future if enough compelling use-cases are found.

`StateFlow` is designed to better cover typical use-cases of keeping track of state changes in time, taking more pragmatic design choices for the sake of convenience.

### Example

For example, the following class encapsulates an integer state and increments its value on each call to `inc`:

```kotlin
class CounterModel {
    private val _counter = MutableStateFlow(0) // private mutable state flow
    val counter: StateFlow<Int> get() = _counter // publicly exposed as read-only state flow

    fun inc() {
        _counter.value++ // convenient state update
    }
}
```

### Experimental status

The initial version of this design is going to be introduced under `@ExperimentalCoroutinesApi`, but it is highly unlikely that the core of the design as described above, is going to change. It is expected to be stabilized quite fast. 

There are also some future possible enhancements (see below) that are not provided at this moment.

### Possible future enhancement: Closing state flow

A state flow could be optionally `closed` in a similar way to channels. When state flow is closed all its collectors complete normally or with the specified exception. Closing a state flow transitions it to the terminal state. Once the state flow is closed its value cannot be changed.  The most recent value of the closed state flow is still available via `value` property. Closing a state is appropriate in situations where the source of the state updates is permanently destroyed.

To support closing there would be a `MutableStateFlow.close(cause: Throwable? = null): Boolean` method for the owner of the state flow to close it via its writeable reference and `StateFlow.isClosed: Boolean` property on read-only interface. 

### Possible future enhancement: Turning Flow into a StateFlow
 
A regular `Flow` is cold. It does not have the concept of the last value and it only becomes active when collected. We could introduce a `stateIn` operator to turn any `Flow` into a hot `StateFlow` in the future as a part of the family of flow-sharing operators (see #1261). It could be designed to become a replacement for `broadcastIn` operator. It would subsume the need of having to have a separate "state flow builder" as you can simply write `flow { .... }.stateIn(scope)` to launch a coroutine that emits the values according to the code in curly braces.

elizarov:
StateFlow is a Flow analogue to ConflatedBroadcastChannel. Since Flow API is simpler than channels APIs, the implementation of StateFlow is simpler. It consumes and allocates less memory, while still providing full deadlock-freedom (even though it is not lock-free internally).

Fixes #1973 (see the issue for design and discussion)


zach-klippenstein:
It would be useful to have a version of `stateIn` that takes an initial value and does not suspend. This would make it easy to define a "loading" or default state while waiting for a resource to spin up.

```kotlin
fun <T> Flow<T>.stateIn(initialValue: T, scope: CoroutineScope): StateFlow<T>
```

Some example use cases:

- A streaming GRPC call that won't emit until it gets a value from the network, but you can provide a reasonable initial state locally.
- A settings store that will emit whenever a setting is changed, but allows you to provide a default value that will be used if the setting has never been set or until it's initially loaded from disk.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -0,0 +1,410 @@
...
+ *
+ * **Preview feature**: The design of state flow closing is tentative and could change.
+ *
+ * ### Turning Flow into a StateFlow
+ *
+ * A regular [Flow] is cold. It does not have the concept of the last value and it only becomes active when collected.
+ * It can be made hot by turning it into a [StateFlow] using [stateIn] operator. For example:
+ *
+ * ```
+ * coroutineScope {
```
yigit, line 67: I'm a bit worried about the usability of this one.

When I see state flow, i kind of want to create view models like this:

```
class ViewModel (db : Db) {
    val use = db.userDao().getUser() // returns Flow<User>
                   .stateIn(viewModelScope)
}
```
The problem is that, i cannot do it because `stateIn` is a suspend function which means i cannot assign my values directly like this in the constructor. On the other hand, it has to be a suspend function to inherit the nullability of the type `T` for the `value` field.

I'm not sure what the solution is to be honest. `value` could be always `nullable` (even when `T` is not nullable) but that would make it hard to distinguish whether it has emitted or not when `T` is `nullable`. 

I could also write:
```
class ViewModel(db : Db) {
    val user  = StateFlow<User?>(null)
   viewModelScope.launch {
        db.userDao().getUser().collect { user.value = it)
   }
}
```
but now my UI will always receive a `null` user when observing this even though it is not desired as i'd rather not dispatch a value until it is read.

A third option i can think of is:

```
class ViewModel(db : Db) {
    val user  = viewModelScope.async {
          db.userDao().getUser().stateIn(viewModelScope)
    }
}
```
May not be too bad, the observer just needs to do `user.await().collect`.

I'm actually not sure what the solution is, just wanted to highlight the use case where we want to have a StateFlow that does not have an initial value. Maybe it is not intended in the design and that is totally OK but it would be good to provide a recommended solution on how to handle these.

Any ideas?

LouisCAD: Idea from [a comment in the related issue](
https://github.com/Kotlin/kotlinx.coroutines/issues/1973#issuecomment-621410369):
Having this function to provide an initial state, that could be `null` or anything.
```kotlin
fun <T> Flow<T>.stateIn(initialValue: T, scope: CoroutineScope): StateFlow<T>
```

Otherwise, your third option seems the best to me. An extra optimization in case the given `StateFlow` is not always needed would be to pass `start = CoroutineStart.LAZY` to `async`.

elizarov: Great idea. I've added `stateIn(scope, value)` that is not suspending. The scope is the first parameter because it reads better this way.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -0,0 +1,411 @@
...
+ *   for faster, garbage-free implementation, unlike `ConflatedBroadcastChannel` implementation that
+ *   allocates objects on each emitted value.
+ * * `StateFlow` always has a value which can be safely read at any time via [value] property.
+ *    Unlike `ConflatedBroadcastChannel`, there is no way to create a state flow without a value.
+ * * `StateFlow` has a clear separation into a read-only `StateFlow` interface and a [MutableStateFlow].
+ * * `StateFlow` conflation works similarly to [Flow.conflate] operator in that the last emitted value
+ *    before close is received by collectors and is always available afterwards, unlike
+ *    `ConflatedBroadcastChannel` that does not keep the last value when closed.
+ * * `StateFlow` conflation is based on equality like [distinctUntilChanged] operator,
+ *    unlike conflation in `ConflatedBroadcastChannel` that is based on reference identity.
```
zach-klippenstein, line 126: This difference in how operators define equality seems arbitrary from the outside. Although even if you know that one uses a channel and one doesn't, it's still not obvious why the channel doesn't use `equals` (presumably performance?). I'm worried this inconsistency is going to be very confusing to people trying to learn this API for the first time, or even people who use it often (I am pretty sure I'm going to keep forgetting which operator does what). Especially coming from RxJava, where (I believe) all operators are consistent in this regard.

I don't think this should block this PR as it's a larger question, but is this something that could be addressed in the future?

elizarov: The rationale here is that `Channel` is a very low-level primitive, so it is Ok for it to be performance conscious, while `StateFlow` is a very high-level primitive, so it is Ok for it to rely on equality that might be broken, etc.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
elizarov: We've decided to postpone `stateIn` for a bit and include it into a separate enhancement that would include a whole family of flow-sharing operators as a cohesively designed group.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
zach-klippenstein: Agreed Channel is low-level, but that behavior is exposed by `Flow.conflate`, which doesn't seem as low-level. Thinking about this more though, it probably doesn't matter too much since the whole purpose of the `conflate()` operator is to affect backpressure, and for that it doesn't matter how equality is checked. If you want distinct elements, then you should still use `distinctUntilChanged()`, even if it's right after a `conflate()`.

kotlinx-coroutines-core/common/src/flow/operators/StateFlowFusion.kt
```diff
@@ -0,0 +1,47 @@
...
+ * Returns this.
+ * Applying [flowOn][Flow.flowOn] operator to [StateFlow] has no effect.
+ * See [StateFlow] documentation on Operator Fusion.
+ */
+@Deprecated(
+    level = DeprecationLevel.WARNING,
+    message = "Applying flowOn operator to StateFlow has no effect. See StateFlow documentation on Operator Fusion.",
+    replaceWith = ReplaceWith("this")
+)
+public fun <T> StateFlow<T>.flowOn(context: CoroutineContext): Flow<T> = this
```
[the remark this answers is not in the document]
elizarov: I've considered the `MigrationKt`, but then it kind of struck to me that the name is not very appropriate here. It's more about "warning on potential bugs in code". Could be something like `LintKt`.

kotlinx-coroutines-core/common/src/flow/operators/Distinct.kt
```diff
@@ -7,15 +7,22 @@
...
 import kotlinx.coroutines.flow.internal.*
 import kotlin.jvm.*
 import kotlinx.coroutines.flow.internal.unsafeFlow as flow
 
 /**
  * Returns flow where all subsequent repetitions of the same value are filtered out.
+ *
+ * Note that any instance of [StateFlow] already behaves as if `distinctUtilChanged` operator is
+ * applied to it, so applying `distinctUntilChanged` to a `StateFlow` has no effect.
+ * See [StateFlow] documentation on Operator Fusion.
```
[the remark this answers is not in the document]
elizarov: It is mostly designed for seasoned Rx users who sometimes tend to use `distinctUntilChanged` a lot in their code when they work with the application state.

kotlinx-coroutines-core/common/src/flow/operators/Context.kt
```diff
@@ -190,13 +194,16 @@ public fun <T> Flow<T>.conflate(): Flow<T> = buffer(CONFLATED)
  *     .flowOn(Dispatchers.Default)
  * ```
  *
+ * Note that an instance of [StateFlow] does not have an execution context by itself,
```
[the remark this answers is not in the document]
elizarov: The other reason is that that documentation should be exaustive. Otherwise, there is no way to figure out which operators behave in a specialized way for `StateFlow`, but to read documentation for all operations.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -0,0 +1,317 @@
...
+                        return // done
+                    }
+                    // reread everything for the next loop under the lock
+                    curSequence = sequence
+                    curSlots = slots
+                }
+            }
+        }
+
+    @Suppress("UNCHECKED_CAST")
```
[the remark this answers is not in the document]
elizarov: Refactoring leftover. Great catch!

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    private fun allocateSlot(): StateFlowSlot = synchronized(this) {
+        val size = slots.size
+        if (nSlots >= size) slots = slots.copyOf(2 * size)
+        var index = nextIndex
+        var slot: StateFlowSlot
+        while (true) {
+            slot = slots[index] ?: StateFlowSlot().also { slots[index] = it }
```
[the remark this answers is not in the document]
elizarov: This code is reusing allocated slot, too. This code was structured this way mostly for legacy reasons, but it is easier to think about the new slot as "unused one".

Slot reuse is a small optimization, but together with `suspendCancellableCoroutine` -> `suspendCancellableCoroutineReusable` change (that we will make after #1813 is merged) it turn this `StateFlow` implementation into a completely allocation-free in its steady-state even for cases of fast collector creation and disappearance (when someone is using `first()` operator a lot, for example). Although, we can also make a fast-path to `first()` operator for `StateFlow`, if needed.

kotlinx-coroutines-core/jvm/test/flow/StateFlowStressTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import kotlin.random.*
+
+class StateFlowStressTest : TestBase() {
+    private val nSeconds = 3 * stressTestMultiplier
+    private val state = MutableStateFlow<Long>(0)
+    private lateinit var pool: ExecutorCoroutineDispatcher
+
+    @After
+    fun tearDown() {
+        pool.close()
```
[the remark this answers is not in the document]
elizarov: Hm.... I cannot figure out how to make `ExecutorRule` work with dynamically initialized executor like this code here (see, it sets the number of threads depending on the run):
```
pool = newFixedThreadPoolContext(nEmitters + nCollectors, "StateFlowStressTest")
```

kotlinx-coroutines-core/common/src/flow/operators/StateFlowFusion.kt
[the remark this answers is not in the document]
qwwdfsad: `LintKt` works for me.
But let's make implementations deprecated with error and without implementation at all.

elizarov:
Note, that it also fixes #395 and it removes the need for `ConflatedBroadcastChannel` to be allocation-free. `StateFlow` implementation is allocation-free and covers the use-cases presented there.

elizarov:
Also fixes #1816

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -0,0 +1,316 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
qwwdfsad, line 2: Copyright year

proto-mvp:
How can we use **StateFlow**, in case we need to emit data classes, that have same values ? 
In other words with **distinctUntilChanged** disabled.

It makes sense when a sequence of actions needs to emit the same result through the StateFlow, or simply when retrying an action from a user interface has the same result.

LouisCAD:
@proto-mvp Adding a timestamp or index to your data class can do. Otherwise, `StateFlow` is not designed for what you're looking for.

proto-mvp:
@LouisCAD I did something to make my data classes unique, but wanted to know if we could override it.

dave08:
@proto-mvp 

If you have something like `val MutableStateFlow<T?>.valueNotDistinct set(newValue: T) { value = null; value = newValue }`, I think you could simulate that behavior without extra allocations.

elizarov:
@proto-mvp What's your actual use-case? What kind of data are you sending?

proto-mvp:
emitting states of UI on an MVI - reactive architecture.
Was using previously a MutableLiveData with the asFlow operator, and got into this after switching to StateFlow, when had to emit the same state again. 
e.g. There is an error, user presses retry , emitting second time a Retry data class wasn't captured.

Ended up having an extra Id field to distinguish  and get the behaviour I wanted with StateFlow.

I know it is by design **distinctUntilChanged** enabled, but would be handy to have a StateFlow without this behaviour, in cases we need to drop MutableLiveData, or PublishSubject from Rx.Java, and not having to use channels.

proto-mvp:
> @proto-mvp
> 
> If you have something like `val MutableStateFlow<T?>.valueNotDistinct set(newValue: T) { value = null; value = newValue }`, I think you could simulate that behavior without extra allocations. And then use `filterNotNull()` to listen to it.

that is a good tip. Changing the sequence of the values should world. Will give it a go. Cheers!

LouisCAD:
> would be handy to have a StateFlow without this behaviour

This is what `SharedFlow` will have.

AdamSHurwitz:
> How can we use StateFlow, in case we need to emit data classes, that have same values?
In other words with distinctUntilChanged disabled.
> 
> It makes sense when a sequence of actions needs to emit the same result through the StateFlow, or simply when retrying an action from a user interface has the same result.

Great question @proto-mvp. I experienced the same use case this week when building a Model-View-Intent (MVI) pattern proof-of-concept with StateFlow. 

The proposed solution of '_having an extra ID field to distinguish unique data_' will work. If no extra data is required for the intent, a new class or [data class](https://kotlinlang.org/docs/reference/data-classes.html) instance, when data is needed, can be set to the `MutableStateFlow` value which will have a unique hash code, and thus be considered as a unique value by `distinctUntilChanged`.

In the sample below every time `loadRequestIntent.value = Emit()` is implemented, _SomeViewModel_ will observe the emission of the unique intent/event.

_Emit.kt_
```Kotlin
class Emit
```

_SomeFragment.kt_

```Kotlin
@ExperimentalCoroutinesApi
class FeedFragment : Fragment(), FeedView {
    private var loadRequestIntent = MutableStateFlow<Emit?>(null)
    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
        super.onViewCreated(view, savedInstanceState)
        if (savedInstanceState == null)
            loadRequestIntent.value = Emit()
    }
}
```

_SomeView.kt_

```Kotlin
interface SomeView {
    /**
     * Load feed from the network
     *
     * @return A flow emitting the feed state returned from the network
     */
    @ExperimentalCoroutinesApi
    fun loadNetworkIntent(): Flow<Emit>
}
```

_SomeViewModel.kt_

```Kotlin
@ExperimentalCoroutinesApi
class FeedViewModel(private val repository: FeedRepository) : ViewModel() {
    fun bindIntents(view: FeedView) {
        view.loadNetworkIntent().onEach {
            loadNetwork(it.toRetry)
        }.launchIn(viewModelScope)
    }
```

Zhuinden:
@AdamSHurwitz

>        if (savedInstanceState == null)
>            loadRequestIntent.value = Emit()

Who's going to trigger a data reload after process death then? 🤔

AdamSHurwitz:
This is a great point @Zhuinden. In the [sample](https://github.com/Kotlin/kotlinx.coroutines/pull/1974#issuecomment-640108388) above, when a process death occurs `savedInstanceState != null`, and the ViewModel data will be cleared according to the documentation, [Options for preserving UI state](https://developer.android.com/topic/libraries/architecture/saving-states#options_for_preserving_ui_state). In summary, the ViewModel does not survive a process death, and the saved instance state does survive.

The result in the sample code is there will not be a new network request for data, however existing data persisted in Room database will populate the view state. In the full ViewModel implementation the above sample is derived from, if a new network data request is not made from `loadRequestIntent`, the existing view state will be initialized from Room's database and a user can swipe-to-refresh in order to initiate a new network request initiating a new `loadRequestIntent` emission. 

I need to think more in-depth about a clean way to refresh the network request on process death vs. relying on a manual swipe-to-refresh to this point. I'm open to suggestions as well! 🙏🏻 Also, I'm updating the open-source [Coinverse app](https://github.com/AdamSHurwitz/Coinverse) with a full sample of the above pattern I will share with you @Zhuinden once drafted.

AdamSHurwitz:
@Zhuinden, here is the full [Coinverse](https://proandroiddev.com/android-model-view-intent-with-kotlin-flow-ca5945316ec#bbd3) sample using StateFlow with a Model-View-Intent pattern.

_**Re: "Who's going to trigger a data reload after process death then?"**_

Thinking more about your WorkManager (WM) suggestion over DMs. An implementation could look like WM checking a SharedPref value to see when the network data was last updated, and emit `loadRequestIntent.value = true` when a given threshold of time has passed and when WM conditions have been met.

PR #1975, opened 2020-04-29
igushkin/gradle-6-3

h0tk3y:


turansky:
@h0tk3y Why gradle wrapper not updated fully? 
Automatic update [generate more changes](https://github.com/Kotlin/kotlinx.coroutines/pull/1971/commits/ec446e6e85eb656bfa23c66e0a8be5213eb21500), (not only in properties).

integration-testing/build.gradle
```diff
@@ -56,7 +56,6 @@ task npmTest(type: Test) {
 task mavenTest(type: Test) {
     def sourceSet = sourceSets.mavenTest
     dependsOn(project(':').getTasksByName("publishToMavenLocal", true))
-    dependsOn.remove(project(':').getTasksByName("dokka", true))
```
qwwdfsad, line 59: Please elaborate on why did you remove this line

turansky: No `remove` in 6.3 :(

elizarov: I do belive we still build all docs no CI now because the docs are needed for `knitCheck` task which we run in CI to ensure that we catch fast the cases where somebody committed updates to docs and forgot to run `knit` task.

Issue #1938, opened 2020-04-22
Kotlin DSL


turansky:
- [x] #1939 - Add version extension

### Convert root build script
- [x] #1969 - platform util
- [x] #1977 - plugin versions in `settings`
- [x] #1984 - Single JDK check

### New Kotlin plugins
- [x] #2018 - JVM, MPP
- [x] #1983 - JS
- [x] Remove old DCE
 
### Convert build scripts
- [x] #1952 - `benchmarks`
- [x] #2255 - `integration/kotlinx-coroutines-guava`
- [x] #1944 - `integration/kotlinx-coroutines-jdk8`
- [x] #2493 - `integration/kotlinx-coroutines-play-services`
- [x] #2256 - `integration/kotlinx-coroutines-slf4j`
- [ ] `integration-testing`
- [x] #2492 - `js/example-frontend-js`
- [x] #1959 - `js/js-stub`
- [ ] `kotlinx-coroutines-bom`
- [ ] `kotlinx-coroutines-core`
- [ ] `kotlinx-coroutines-debug`
- [x] #1945 - `kotlinx-coroutines-test`
- [x] #2257 - `reactive/kotlinx-coroutines-jdk9`
- [x] #1946 - `reactive/kotlinx-coroutines-reactive`
- [x] #2258 - `reactive/kotlinx-coroutines-reactor`
- [ ] `reactive/kotlinx-coroutines-rx2`
- [ ] `reactive/kotlinx-coroutines-rx3`
- [x] #1954 - `site`
- [x] #1955 - `stdlib-stubs`
- [x] #1956 - `ui/kotlinx-coroutines-android`
- [x] #1942 - `ui/kotlinx-coroutines-android/android-unit-tests`
- [x] #1941 - `ui/kotlinx-coroutines-android/animation-app`
- [x] #1941 - `ui/kotlinx-coroutines-android/animation-app/app`
- [x] #1940 - `ui/kotlinx-coroutines-android/example-app`
- [x] #1940 - `ui/kotlinx-coroutines-android/example-app/app`
- [x] #2017 - `ui/kotlinx-coroutines-javafx`
- [x] #1943 - `ui/kotlinx-coroutines-swing`

### Convert additional scripts to plugins
- [ ] `compile-common`
- [x] #2474 - `compile-js`
- [ ] `compile-js-multiplatform`
- [x] #2474 - `compile-jvm`
- [ ] `compile-jvm-multiplatform`
[...]

PR #1982, opened 2020-05-02
`stub-convention` plugin

turansky:
Part of #1938


buildSrc/build.gradle.kts
```diff
@@ -9,3 +11,14 @@ repositories {
 kotlinDslPluginOptions {
     experimentalWarning.set(false)
 }
+
+val props = Properties().apply {
```
qwwdfsad, line 15: please rename it to just `properties`

js/js-stub/build.gradle.kts
```diff
@@ -2,10 +2,6 @@
...
  */
 
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
-
-tasks.named<KotlinCompile>("compileKotlin") {
-    kotlinOptions {
-        freeCompilerArgs += "-Xallow-kotlin-package"
-    }
+plugins {
+    `stub-convention`
```
qwwdfsad, line 6: Please rename it to `stdlib-stub-convention`, just "stub" is too broad term

buildSrc/build.gradle.kts
turansky: `properties` already exists in `Project`

js/js-stub/build.gradle.kts
turansky: Done

PR #1983, opened 2020-05-02
Use new Kotlin/JS plugin

turansky:
Part of #1938

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -239,7 +239,7 @@ configure(subprojects.findAll { it.name != coreModule && it.name != rootModule }
 }
 
 // Redefine source sets because we are not using 'kotlin/main/fqn' folder convention
-configure(subprojects.findAll { !sourceless.contains(it.name) && it.name != "benchmarks" && it.name != coreModule }) {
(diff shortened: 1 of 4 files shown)


elizarov:
Look good to me. Note, that the new Kotlin/JS has build-in Webpack support and it should be possible to completely get rid of the custom node configuration in `gradle/node-js.gradle`.

turansky:
Next 2 steps after this PRs:
1. Use new MPP plugin for `core`
2. Use `build/distributions` in examples 
2.1. Remove old DCE

@skuzmich Could I skip your change request in this case (CR will be removed in step 2)?

turansky:
> Note, that the new Kotlin/JS has build-in Webpack support and it should be possible to completely get rid of the custom node configuration in `gradle/node-js.gradle`.

One of the goal of this PR - simplify migration to Kotlin DSL (no `gradle/node-js.gradle` == no migration required)

skuzmich:
If kotlinx.coroutines maintainers are OK with temporarily breaking 1.4 builds I can lift the change request.
cc @elizarov

PR #1984, opened 2020-05-02
Add project JDK convention 

turansky:
Part of #1938

* JDK 11
* Weak check for IDEA


elizarov:
Ops. The project now does not even import under JDK8, which is not good. The idea was that use should not be able to _build_ it under JDK8, but there should be minimal barriers for letting users open the project in their IDE.

turansky:
Could we exclude `javafx` subproject for those users?

PR #1988, opened 2020-05-03
Use standard random API

turansky:
* Deprecated API usage removed

### DIFF
--- a/js/example-frontend-js/src/ExampleMain.kt
+++ b/js/example-frontend-js/src/ExampleMain.kt
@@ -36,8 +36,6 @@ private fun HTMLElement.setPosition(x: Double, y: Double) {
     }
 }
 
-private fun random(): Double = Random.nextDouble()
-
 class Application : CoroutineScope {
     private val body get() = document.body!!
     private val scene get() = document.getElementById("scene") as HTMLElement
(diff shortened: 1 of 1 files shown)


js/example-frontend-js/src/ExampleMain.kt
```diff
@@ -35,8 +36,7 @@ private fun HTMLElement.setPosition(x: Double, y: Double) {
     }
 }
 
-@Suppress("DEPRECATION")
-private fun random() = kotlin.js.Math.random()
+private fun random():Double = Random.nextDouble()
```
elizarov, line 39: One empty line should also be removed.

turansky: Which one? I don't add empty lines :)

elizarov: Sorry. My bad. Disregard.

elizarov: I somehow thought I've removed the function completely. Actually, there's now no much sense in having this function in the first place -- it would be better if it is just inlined to all its use-sites.

turansky: Inlined

PR #2001, opened 2020-05-08
Version 1.3.6

qwwdfsad:


CHANGES.md
```diff
@@ -1,5 +1,26 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.3.6
+
+### Flow
+
+* `StateFlow`, new primitive for state handling (#1973, #1816, #395).
```
elizarov, line 7: Let's add the following note here:

> The `StateFlow` is designed to eventually replace `ConflatedBroadcastChannel` for _state_ publication scenarios. Please, try it and share your feedback. Note, that Flow-based primitives to publish _events_ will be added later. For _events_ you should continue to either use `BroadcastChannel(1)`, if you put events into the `StateFlow`, protect them from double-processing with flags.

PR #2006, opened 2020-05-09
Update README to Kotlin 1.3.72

NahroTo:


elizarov:
Sorry, but it is not enough to update readme to 1.3.72. This library is still build with 1.3.71 and readme is updated to match the actual Kotlin version it is built againts.

PR #2010, opened 2020-05-11
Mutex.withLock() added EXACTLY_ONCE contract

ZhouGongZaiShi:
`Mutex.withLock(owner: Any? = null, action: () -> T)` should be like `synchronized (lock: Any, block:()-> R)`, add `EXACTLY_ONCE` contract. 
This is very helpful when the compiler infers variable initialization.
```
val str : String
mMutex.withLock {
   str = "If no Exception is thrown, it will be assigned."
}
println("I should be able to use str field. str = $str")
```

### DIFF
--- a/kotlinx-coroutines-core/common/src/sync/Mutex.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Mutex.kt
@@ -107,7 +107,7 @@ public fun Mutex(locked: Boolean = false): Mutex =
  *
  * @return the return value of the action.
  */
-@ExperimentalContracts
+@OptIn(ExperimentalContracts::class)
 public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T): T {
     contract { 
         callsInPlace(action, InvocationKind.EXACTLY_ONCE)
--- a/kotlinx-coroutines-core/common/src/sync/Mutex.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Mutex.kt
@@ -9,9 +9,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.internal.*
 import kotlinx.coroutines.intrinsics.*
 import kotlinx.coroutines.selects.*
-import kotlin.contracts.ExperimentalContracts
-import kotlin.contracts.InvocationKind
-import kotlin.contracts.contract
+import kotlin.contracts.*
 import kotlin.coroutines.*
 import kotlin.jvm.*
 import kotlin.native.concurrent.*
--- a/kotlinx-coroutines-core/common/src/sync/Mutex.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Mutex.kt
@@ -107,6 +107,10 @@ public fun Mutex(locked: Boolean = false): Mutex =
  * @return the return value of the action.
  */
 public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T): T {
+    contract { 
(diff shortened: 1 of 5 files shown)


fvasco:
Same issue on `Semaphore.withPermit`

zach-klippenstein:
I think this needs to be `AT_MOST_ONCE` since the block won't be ran at all if `lock` throws (eg cancelled while suspended).

ZhouGongZaiShi:
> I think this needs to be `AT_MOST_ONCE` since the block won't be ran at all if `lock` throws (eg cancelled while suspended).

I think that if `mutex.withlock{}` throws a `CancellationException`(or other exception)  while suspended, then the coroutine will not run the code after `mutex.withlock{}`.

zach-klippenstein:
Yep, I forgot that the compiler will assume anything run inside a `try` wasn't executed if it throws.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -9,6 +9,9 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.internal.*
 import kotlinx.coroutines.intrinsics.*
 import kotlinx.coroutines.selects.*
+import kotlin.contracts.ExperimentalContracts
```
elizarov, line 12: Please, use star imports in all the files (you can take style guide settings from #2021 )

kotlinx-coroutines-core/common/test/sync/MutexTest.kt
```diff
@@ -5,8 +5,10 @@
 package kotlinx.coroutines.sync
 
 import kotlinx.coroutines.*
+import kotlin.contracts.ExperimentalContracts
 import kotlin.test.*
 
+@ExperimentalContracts
```
elizarov, line 11: Why is it needed in tests and examples?  Something is wrong here.

ZhouGongZaiShi: The `contract` is marked with the annotation `@ExperimentalContracts`.     
@ExperimentalContracts have propagatinging, so tests and examples also need to be marked @ExperimentalContracts. (See [Opt-in Requirements](https://kotlinlang.org/docs/reference/opt-in-requirements.htm)l) 
My idea is that `Mutex.withLock()` / `Semaphore.withPermit()` is an open API, and users have the right to decide whether to use this Experimental funcation, rather than the library developers directly blocking its propagate.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ZhouGongZaiShi: Updated. [a132210](https://github.com/Kotlin/kotlinx.coroutines/pull/2010/commits/a132210713c9278af60f38d5e80d1b5cf4e6c063)

kotlinx-coroutines-core/common/test/sync/MutexTest.kt
elizarov: That's a no-go. Both `withLock` and `withPermit` are public. They cannot force their users to opt-in. However, both `withLock` and `withPermit` can opt-in into experimental contracts without forcing users to.

ZhouGongZaiShi: Updated. [752e060](https://github.com/Kotlin/kotlinx.coroutines/pull/2010/commits/752e0604d882eec219831dfe91c3cf2db0b3535b)

PR #2012, opened 2020-05-12
Cancel current Job on RejectedExecutionException

### LINKED ISSUE #2003: Coroutines launch on default dispatcher when they should not
In one of our production services we had a fixed thread-pool for handling blocking DB calls, and we are using `withContext` and `asCoroutineDispatcher` to launch such calls in context of DB threads. These calls have to stick to these thread-pool we have created otherwise they fail:
```kotlin
          val executor = ThreadPoolExecutor(
                affinity,
                MAX_IO_THREADS,
                DEFAULT_TTL_SECONDS_THREAD,
                TimeUnit.SECONDS,
                SynchronousQueue()
            ) { runnable ->
                Thread(runnable).also {
                    it.name = "io-pool-${it.id}"
                    it.isDaemon = true
                    it.priority = Thread.NORM_PRIORITY
                }
            }
          
          val dispatcher = executor.asCoroutineDispatcher()
```

**Notice** `SynchronousQueue` to block spinning off any further threads/calls in when pool is full. Turns out looking at code of `ExecutorCoroutineDispatcherBase` https://github.com/Kotlin/kotlinx.coroutines/blob/d7de5f5ba66a8d005e5cbd03b18522112303fd54/kotlinx-coroutines-core/jvm/src/Executors.kt#L80-L87 turns out we launch/enqueue or coroutine on `DefaultExecutor`. This has multiple problems:

 - It breaks the foundational assumption of respecting dispatcher that was supplied, worst part of this is that it's undocumented caveat. 
 - It can result in a flurry of undesired threads; ultimately leading to OOM with it's existing behavior. Having an unbound queue or unbound set of threads is exactly what should be avoided at all costs.

Here is a simple [example to reproduce this issue locally](https://pl.kotl.in/3iPiVQjir):
```kotlin
import kotlinx.coroutines.asCoroutineDispatcher
import kotlinx.coroutines.async
import kotlinx.coroutines.awaitAll
import kotlinx.coroutines.withContext
import java.util.concurrent.Executors
import java.util.concurrent.SynchronousQueue
import java.util.concurrent.ThreadPoolExecutor
import java.util.concurrent.TimeUnit

val executor = ThreadPoolExecutor(1, 1, 60, TimeUnit.SECONDS, SynchronousQueue()) { runnable ->
    Thread(runnable).also {
        it.name = "io-pool-${it.id}"
        it.isDaemon = true
        it.priority = Thread.NORM_PRIORITY
    }
}

fun main() {
    runBlocking(executor.asCoroutineDispatcher()) {
        for (i in 1..5) {
            launch {
                println(Thread.currentThread().name)
                Thread.sleep(1000)
            }
        }
    }
}
```

I believe the subsequent co-routines should be rejected; launching by default on `DefaultExecutor` is an undesired behavior and breaks fundamental assumption.

elizarov:
When the Executor that was used with Executor.asCoroutineDispatcher() extension rejects submitted task, it means that it had reached its capacity and so the executing current Job should be cancelled to terminate it as soon as possible. This way RejectedExecutionException works as a rate-limiter just like it serves this purpose in executor-based Java code.

**NOTE**: This PR changes internal coroutines API for `Delay` interface which some 3rd part code might be using, so it should be merged only in the major release.

Fixes #2003

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -39,8 +39,9 @@ public abstract class ExecutorCoroutineDispatcher: CoroutineDispatcher(), Closea
 /**
  * Converts an instance of [ExecutorService] to an implementation of [ExecutorCoroutineDispatcher].
  *
- * Note, that if the underlying executor throws [RejectedExecutionException] on
- * attempt to submit a continuation task (it typically happens on executor shutdown or when it uses limited queues),
+ * If the underlying executor throws [RejectedExecutionException] on
+ * attempt to submit a continuation task (it happens when [closing][ExecutorCoroutineDispatcher.close] the
+ * resulting dispatcher, on underlying executor [shutdown][ExecutorService.shutdown], or when it uses limited queues),
  * then the [Job] of the affected task is [cancelled][Job.cancel] and the task is submitted to the
  * [Dispatchers.IO], so that the affected coroutine can cleanup its resources and promptly complete.
  */
@@ -51,8 +52,9 @@ public fun ExecutorService.asCoroutineDispatcher(): ExecutorCoroutineDispatcher 
 /**
  * Converts an instance of [Executor] to an implementation of [CoroutineDispatcher].
  *
- * Note, that if the underlying executor throws [RejectedExecutionException] on
- * attempt to submit a continuation task (it typically happens on executor shutdown or when it uses limited queues),
+ * If the underlying executor throws [RejectedExecutionException] on
+ * attempt to submit a continuation task (it happens when [closing][ExecutorCoroutineDispatcher.close] the
+ * resulting dispatcher, on underlying executor [shutdown][ExecutorService.shutdown], or when it uses limited queues),
  * then the [Job] of the affected task is [cancelled][Job.cancel] and the task is submitted to the
  * [Dispatchers.IO], so that the affected coroutine can cleanup its resources and promptly complete.
  */
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -38,6 +38,11 @@ public abstract class ExecutorCoroutineDispatcher: CoroutineDispatcher(), Closea
 
 /**
  * Converts an instance of [ExecutorService] to an implementation of [ExecutorCoroutineDispatcher].
+ *
+ * Note, that if the underlying executor uses limited queues and throws [RejectedExecutionException] on
+ * attempt to submit a task, then the [Job] of the affected task is [cancelled][Job.cancel] and the
+ * task is submitted to the default single-threaded executor, so that the affected coroutine can cleanup its
+ * resources and promptly complete.
  */
 @JvmName("from") // this is for a nice Java API, see issue #255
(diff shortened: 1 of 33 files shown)


kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -38,13 +38,23 @@ public abstract class ExecutorCoroutineDispatcher: CoroutineDispatcher(), Closea
 
 /**
  * Converts an instance of [ExecutorService] to an implementation of [ExecutorCoroutineDispatcher].
+ *
+ * Note, that if the underlying executor uses limited queues and throws [RejectedExecutionException] on
```
fvasco, line 42: Job is cancelled iff the executor refects a task, indipendently if is queue is limited or not (ie on shutdown). Same in `Executor.asCoroutineDispatcher()`.

kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -38,13 +38,23 @@ public abstract class ExecutorCoroutineDispatcher: CoroutineDispatcher(), Closea
...
+ * attempt to submit a task, then the [Job] of the affected task is [cancelled][Job.cancel] and the
```
fvasco, line 43: I don't know if this is the right point to explain that a cancellation can happen at coroutine start, cancellation and in any suspension point.

kotlinx-coroutines-core/jvm/src/Executors.kt
elizarov: I've clarified docs a bit. Check it out.

kotlinx-coroutines-core/jvm/src/Executors.kt
elizarov: That's probably going too much into the details and basics that are explained elsewhere.

kotlinx-coroutines-core/jvm/src/Executors.kt
fvasco: Looks good to me.

kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -38,13 +38,23 @@ public abstract class ExecutorCoroutineDispatcher: CoroutineDispatcher(), Closea
...
+ * [Dispatchers.IO], so that the affected coroutine can cleanup its resources and promptly complete.
  */
 @JvmName("from") // this is for a nice Java API, see issue #255
 public fun ExecutorService.asCoroutineDispatcher(): ExecutorCoroutineDispatcher =
     ExecutorCoroutineDispatcherImpl(this)
 
 /**
  * Converts an instance of [Executor] to an implementation of [CoroutineDispatcher].
+ *
+ * Note, that if the underlying executor throws [RejectedExecutionException] on
```
qwwdfsad, line 54: The same comment worth adding to `newSingleThreadContext` and `newFixedThreadPoolContext`

fvasco: ...and tests to avoid regression.

elizarov:
I've made better tests and docs. 

NOTE: It should be merged on the next major update (`1.4.0`) only, because it updates `Delay` interface APIs. Even though it is an `@InternalCoroutinesApi`, still extra caution will not harm.

qwwdfsad (approved):

qwwdfsad (approved):

PR #2043, opened 2020-05-21
Improve the docs and guide on flow cancellation

elizarov:
* Remove outdated phrase on "flow infrastructure does not introduce additional cancellation points". They were introduced by #2028
* Add a section on "Flow cancellation check" with examples on `Flow.cancellable()`` operator.
* Add a bit more detail in `flow` and `cancellable` docs with links to `ensureActive()`.

### DIFF
--- a/docs/flow.md
+++ b/docs/flow.md
@@ -10,7 +10,7 @@
     * [Suspending functions](#suspending-functions)
     * [Flows](#flows)
   * [Flows are cold](#flows-are-cold)
-  * [Flow cancellation](#flow-cancellation)
+  * [Flow cancellation basics](#flow-cancellation-basics)
   * [Flow builders](#flow-builders)
   * [Intermediate flow operators](#intermediate-flow-operators)
     * [Transform operator](#transform-operator)
@@ -42,6 +42,8 @@
     * [Successful completion](#successful-completion)
   * [Imperative versus declarative](#imperative-versus-declarative)
   * [Launching flow](#launching-flow)
+  * [Flow cancellation checks](#flow-cancellation-checks)
+    * [Making busy flow cancellable](#making-busy-flow-cancellable)
   * [Flow and Reactive Streams](#flow-and-reactive-streams)
 
 <!--- END -->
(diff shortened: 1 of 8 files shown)


docs/flow.md
```diff
@@ -1777,6 +1779,120 @@ as cancellation and structured concurrency serve this purpose.
...
+
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
+
+```kotlin
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+
+//sampleStart           
+fun main() = runBlocking<Unit> {
+    (1..5).asFlow().().collect { value ->
```
EdwarDDay, line 1873: It looks like a missing function
```suggestion
    (1..5).asFlow(). cancellable().collect { value -> 
```

elizarov: Thanks! Messed it up during commit. Does not pass tests this way either.

docs/flow.md
```diff
@@ -1777,6 +1779,120 @@ as cancellation and structured concurrency serve this purpose.
...
+ 
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
+
+```kotlin
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+
+//sampleStart           
+fun foo(): Flow<Int> = flow { 
+    for (i in 1..5) { emit(i) }
```
LouisCAD, line 1795: Might make sense to print before the call to emit to show evidence that the flow builder is cancelled, and not just dropped, WDYT?

elizarov: Great idea! Added.

docs/flow.md
```diff
@@ -1777,6 +1779,127 @@ as cancellation and structured concurrency serve this purpose.
...
+It means that a busy loop emitting from a `flow { ... }` is cancellable:
+ 
+<div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
+
+```kotlin
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+
+//sampleStart           
+fun foo(): Flow<Int> = flow {
```
qwwdfsad, line 1794: [Nit] For guides it would be nice to stick with a meaningful naming (e.g. `simpleFlow`) instead of just `foo`

LouisCAD: I agree, `foo` and `bar` are confusing to many people.
I like to use "some" as a prefix, and continue with what it is for extra clarity.
Examples: `someFlow`, `SomeClass`, `someFunction()`, `SomeType.someExtensionProperty`.

There's some alternatives to "some" like "aRandom", or suffixing with number literals when there's two or more similarly kinded symbols (better than foo and bar IMHO).

elizarov: I'll follow up with `foo` banishment in a separate change.

PR #2045, opened 2020-05-21
CancellableQueueSynchronizer and ReadWriteMutex

ndkoval:
Introduce `SegmentQueueSynchronizer` abstraction for synchronization primitives and `ReadWriteMutex`. Please follow the provided documentation and the test suite for details.

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/ConcurrentLinkedList.kt
+++ b/kotlinx-coroutines-core/common/src/internal/ConcurrentLinkedList.kt
@@ -188,7 +188,16 @@ internal abstract class ConcurrentLinkedListNode<N : ConcurrentLinkedListNode<N>
  * Essentially, this is a node in the Michael-Scott queue algorithm,
  * but with maintaining [prev] pointer for efficient [remove] implementation.
  */
-internal abstract class Segment<S : Segment<S>>(val id: Long, prev: S?, pointers: Int): ConcurrentLinkedListNode<S>(prev) {
+internal abstract class Segment<S : Segment<S>>(val id: Long, prev: S?, pointers: Int) :
+    ConcurrentLinkedListNode<S>(prev),
+    // Segments typically store waiting continuations. Thus, on cancellation, the corresponding
+    // slot should be cleaned and the segment should be removed if it becomes full of cancelled cells.
+    // To install such a handler efficiently, without creating an extra object, we allow storing
+    // segments as cancellation handlers in [CancellableContinuationImpl] state, putting the slot
+    // index in another field. The details are here: https://github.com/Kotlin/kotlinx.coroutines/pull/3084.
+    // For that, we need segments to implement this internal marker interface.
+    NotCompleted
+{
     /**
      * This property should return the number of slots in this segment,
      * it is used to define whether the segment is logically removed.
@@ -211,6 +220,13 @@ internal abstract class Segment<S : Segment<S>>(val id: Long, prev: S?, pointers
 
     // returns `true` if this segment is logically removed after the decrement.
     internal fun decPointers() = cleanedAndPointers.addAndGet(-(1 shl POINTERS_SHIFT)) == numberOfSlots && !isTail
+
+    /**
+     * This function is invoked on continuation cancellation when this segment
+     * with the specified [index] are installed as cancellation handler via
+     * `SegmentDisposable.disposeOnCancellation(Segment, Int)`.
+     */
+    abstract fun onCancellation(index: Int, cause: Throwable?)
 
     /**
      * Invoked on each slot clean-up; should not be invoked twice for the same slot.
--- a/kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
+++ b/kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
@@ -209,6 +209,7 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
         returnValue(value)
     }
 
+    @Suppress("INFERRED_TYPE_VARIABLE_INTO_POSSIBLE_EMPTY_INTERSECTION")
     internal fun suspendCancelled(): T? {
         // Increment `suspendIdx` and find the segment
         // with the corresponding id. It is guaranteed
@@ -238,6 +239,7 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
         if (value !== BROKEN && segment.cas(i, value, TAKEN)) {
             // The elimination is performed successfully,
             // complete with the value stored in the cell.
+            @Suppress("UNCHECKED_CAST")
             return value as T
         }
         // The cell is broken, this can happen only in the `SYNC` resumption mode.
@@ -245,7 +247,7 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
         return null
     }
 
-    @Suppress("UNCHECKED_CAST")
+    @Suppress("UNCHECKED_CAST", "INFERRED_TYPE_VARIABLE_INTO_POSSIBLE_EMPTY_INTERSECTION")
     internal fun suspend(waiter: Waiter): Boolean {
         // Increment `suspendIdx` and find the segment
         // with the corresponding id. It is guaranteed
@@ -335,7 +337,7 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
      * moves [resumeIdx] to the first possibly non-cancelled cell, i.e.,
      * to the first segment id multiplied by [SEGMENT_SIZE].
      */
-    @Suppress("UNCHECKED_CAST")
+    @Suppress("UNCHECKED_CAST", "INFERRED_TYPE_VARIABLE_INTO_POSSIBLE_EMPTY_INTERSECTION")
     private fun tryResumeImpl(value: T, adjustResumeIdx: Boolean): Int {
         // Check that `adjustResumeIdx` is `false` in the simple cancellation mode.
         assertNot { cancellationMode == SIMPLE && adjustResumeIdx }
@@ -561,12 +563,14 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
                 // provided by a concurrent `resume(..)`.
                 // The value could be put only in the asynchronous mode,
                 // so the `resume(..)` call above must not fail.
+                @Suppress("UNCHECKED_CAST")
                 resume(value as T)
             } else {
                 // The `resume(..)` that will come to this cell should be refused.
                 // Mark the cell correspondingly and help a concurrent
                 // `resume(..)` to process its value if needed.
                 val value = markRefuse(index) ?: return
+                @Suppress("UNCHECKED_CAST")
                 returnRefusedValue(value as T)
             }
         }
--- a/kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
+++ b/kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
@@ -6,6 +6,7 @@ package kotlinx.coroutines.internal
 
 import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
 import kotlinx.coroutines.internal.SegmentQueueSynchronizer.*
 import kotlinx.coroutines.internal.SegmentQueueSynchronizer.CancellationMode.*
 import kotlinx.coroutines.internal.SegmentQueueSynchronizer.ResumeMode.*
@@ -208,6 +209,42 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
         returnValue(value)
     }
 
+    internal fun suspendCancelled(): T? {
+        // Increment `suspendIdx` and find the segment
+        // with the corresponding id. It is guaranteed
+        // that this segment is not removed since at
+        // least the cell for this `suspend` invocation
+        // is not in the `CANCELLED` state.
+        val curSuspendSegm = this.suspendSegment.value
+        val suspendIdx = suspendIdx.getAndIncrement()
+        val segment = this.suspendSegment.findSegmentAndMoveForward(id = suspendIdx / SEGMENT_SIZE, startFrom = curSuspendSegm,
+            createNewSegment = ::createSegment).segment
+        assert { segment.id == suspendIdx / SEGMENT_SIZE }
+        // Try to install the waiter into the cell - this is the regular path.
+        val i = (suspendIdx % SEGMENT_SIZE).toInt()
+        if (segment.cas(i, null, CANCELLED)) {
+            // The continuation is successfully installed, and
+            // `resume` cannot break the cell now, so this
+            // suspension is successful.
+            // Add a cancellation handler if required and finish.
+            return null
+        }
+        // The continuation installation has failed. This happened because a concurrent
+        // `resume` came earlier to this cell and put its value into it. Remember that
+        // in the `SYNC` resumption mode this concurrent `resume` can mark the cell as broken.
+        //
+        // Try to grab the value if the cell is not in the `BROKEN` state.
+        val value = segment.get(i)
+        if (value !== BROKEN && segment.cas(i, value, TAKEN)) {
+            // The elimination is performed successfully,
+            // complete with the value stored in the cell.
+            return value as T
+        }
+        // The cell is broken, this can happen only in the `SYNC` resumption mode.
+        assert { resumeMode == SYNC && segment.get(i) === BROKEN }
+        return null
+    }
+
     @Suppress("UNCHECKED_CAST")
     internal fun suspend(waiter: Waiter): Boolean {
         // Increment `suspendIdx` and find the segment
@@ -359,17 +396,30 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
                     return TRY_RESUME_SUCCESS
                 }
                 // Does the cell store a cancellable continuation?
-                cellState is CancellableContinuation<*> -> {
+                cellState is Waiter -> {
                     // Change the cell state to `RESUMED`, so
                     // the cancellation handler cannot be invoked
                     // even if the continuation becomes cancelled.
                     if (!segment.cas(i, cellState, RESUMED)) continue@modify_cell
                     // Try to resume the continuation.
-                    val token = (cellState as CancellableContinuation<T>).tryResume(value, null, { returnValue(value) })
-                    if (token != null) {
-                        // Hooray, the continuation is successfully resumed!
-                        cellState.completeResume(token)
-                    } else {
+                    val resumed = when(cellState) {
+                        is CancellableContinuation<*> -> {
+                            (cellState as CancellableContinuation<T>)
+                            val token = cellState.tryResume(value, null, { returnValue(value) })
+                            if (token != null) {
+                                // Hooray, the continuation is successfully resumed!
+                                cellState.completeResume(token)
+                                true
+                            } else {
+                                false
+                            }
+                        }
+                        is SelectInstance<*> -> {
+                            cellState.trySelect(this@SegmentQueueSynchronizer, value)
+                        }
+                        else -> error("unexpected")
+                    }
+                    if (!resumed) {
                         // Unfortunately, the continuation resumption has failed.
                         // Fail the current `resume` if the simple cancellation mode is used.
                         if (cancellationMode === SIMPLE)
@@ -552,7 +602,7 @@ internal abstract class SegmentQueueSynchronizer<T : Any> {
                 val cellState = get(index)
                 when {
                     cellState === RESUMED -> return false
-                    cellState is CancellableContinuation<*> -> {
+                    cellState is Waiter -> {
                         if (cas(index, cellState, CANCELLING)) return true
                     }
                     else -> {
--- a/kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
+++ b/kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
@@ -1,0 +1,669 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.internal
+
+import kotlinx.atomicfu.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.internal.SegmentQueueSynchronizer.*
+import kotlinx.coroutines.internal.SegmentQueueSynchronizer.CancellationMode.*
+import kotlinx.coroutines.internal.SegmentQueueSynchronizer.ResumeMode.*
+import kotlinx.coroutines.selects.*
+import kotlinx.coroutines.sync.*
+import kotlin.coroutines.*
+import kotlin.math.*
+import kotlin.native.concurrent.*
+
+/**
+ * [SegmentQueueSynchronizer] (SQS) is an abstraction for implementing _fair_ synchronization and communication primitives.
+ * Essentially, It maintains a FIFO queue of waiting requests and provides two main functions:
+ *   - [suspend] that stores the specified waiter into the queue, and
+ *   - [resume] that tries to retrieve and resume the first waiter, passing the specified value to it.
+ * The key advantage of these semantics is that SQS allows to invoke [resume] before [suspend], and synchronization
+ * primitives can leverage such races. For example, our [Semaphore] implementation actively uses this property
+ * for better performance.
+ *
+ * One useful mental image of [SegmentQueueSynchronizer] is an infinite array with two counters: one references the
+ * next cell in which new waiter is enqueued as a part of [suspend] call, and another references the next cell
+ * for [resume]. The intuition is that [suspend] increments its counter and stores the continuation in the corresponding
+ * cell. Likewise, [resume] increments its counter, visits the corresponding cell, and completes the stored
+ * continuation with the specified value. However, [resume] may come to the cell before [suspend] and find the cell
+ * in the empty state. To solve this race, we introduce two [resumption modes][ResumeMode]: synchronous and asynchronous.
+ * In both case, [resume] puts the value into the empty cell, and then either finishes immediately in the [asynchronous][ASYNC]
+ * resumption mode, or waits until the value is taken by a concurrent [suspend] in the [synchronous][SYNC] one.
+ * In the latter case, if the value is not taken within a bounded time, [resume] marks the cell as _broken_.
+ * Thus, both this [resume] and the corresponding [suspend] fail. The intuition is that allowing for broken cells keeps
+ * the balance of pairwise operations, such as [acquire()][Semaphore.acquire] and [release()][Semaphore.release],
+ * so they should simply restart. This way, we can achieve wait-freedom with the [asynchronous][ASYNC] mode, and
+ * obstruction-freedom with the [synchronous][SYNC] mode.
+ *
+ * **CANCELLATION**
+ *
+ * Since [suspend] can store [CancellableContinuation]-s, it is possible for [resume] to fail if the completing continuation
+ * is already cancelled. In this case, most of the algorithms retry the whole operation. In [Semaphore], for example,
+ * when [Semaphore.release] fails on the next waiter resumption, it logically provides the permit to the waiter but takes it back
+ * after if the waiter is canceled. In the latter case, the permit should be released again so the operation restarts.
+ *
+ * However, the cancellation logic above requires to process all canceled cells. Thus, the operation that resumes waiters
+ * (e.g., [release][Semaphore.release] in [Semaphore]) works in linear time in the number of consecutive canceled waiters.
+ * This way, if N coroutines come to a semaphore, suspend, and cancel, the following [release][Semaphore.release] works
+ * in O(N) because it should process all these N cells (and increment the counter of available permits in our implementation).
+ * While the complexity is amortized by cancellations, it would be better to make such operations as [release][Semaphore.release]
+ * in [Semaphore] work predictably fast, independently on the number of cancelled requests.
+ *
+ * The main idea to improve cancellation is skipping these `CANCELLED` cells, so [resume] always succeeds
+ * if no elimination happens (remember that [resume] may fail in the [synchronous][SYNC] resumption mode if
+ * it finds the cell empty). Additionally, users should specify a cancellation handler that modifies the data structure
+ * after the continuation is cancelled (in [Semaphore], this handler increments the available permits counter back).
+ *
+ * As a result, we support two cancellation policies in [SegmentQueueSynchronizer]. In the [simple cancellation mode][SIMPLE],
+ * [resume] fails and returns `false` if it finds the cell in the `CANCELLED` state or if the waiter resumption
+ * (see [CancellableContinuation.tryResume]) fails. As we discussed, these failures are typically handled
+ * by restarting the whole operation from the beginning. With the [smart cancellation][SMART], [resume] skips `CANCELLED`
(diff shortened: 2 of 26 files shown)


ndkoval:
Here is a link to a successful nightly build with these changes: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStress/3009472

kotlinx-coroutines-core/jvm/test/linearizability/SegmentQueueSynchronizerLCStressTests.kt
```diff
@@ -0,0 +1,770 @@
...
+            else -> false
+        }
+    }
+
+    override fun onCancellation(): Boolean {
+        // Decrement the number of arrived parties if possible.
+        arrived.loop { cur ->
+            // Are we going to be resumed?
+            // The resumption permit should be refused in this case.
+            if (cur == parties.toLong()) return false
```
dkhalanskyjb, line 419: A small problem here. I believe this should be `cur <= parties.toLong()`, but I think it's interesting why so and, more importantly, why it didn't cause any problems.

`parties - 1` calls happen and go to sleep; then comes the call that is about to wake everyone up but doesn't yet do anything after incrementing the counter; then, 1000 new calls happen, and `a == parties + 1000`. Then, every thread is cancelled, performs `onCancellation` and, since `arrived` is much bigger than `parties.toLong()`, successfully decrements `arrived`. Then the resumer comes, sees `parties - 1` `CANCELLED` marks and proceeds to wake up `parties - 1` more cells.

This works only because this behavior, while not optimal, doesn't break anything: after the cells are all woken up, this SQS is never used again. However, if `resumeMode`  was set to `SYNC`, everything would hang in this scenario, but it seems that with the change above it would work fine in this case too.

It should be noted though that without the check for `cur == parties.toLong()` the code would not be linearizable: as is, it guarantees that once the barrier is broken, it can't be accidentally "un-broken" no matter how many cancellations happen, and so the subsequent calls to `arrive` all return `false` in any case.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * [resume] waits in a bounded spin-loop cycle until the put element is taken by a concurrent [suspend],
+ * marking the cell as broken if it is not taken after the spin-loop is finished. In this case, both the current
+ * [resume] and the [suspend] that comes to this broken cell fail and the corresponding operations on the data
+ * structure are typically restarted from the beginning.
+ *
+ * Since [suspend] can store [CancellableContinuation]-s, it is possible for [resume] to fail if the
+ * continuation is already cancelled. In this case, most of the algorithms retry the whole operation.
+ * However, if there are many consecutive cancelled waiters, it seems more efficient to skip them
+ * somehow. Thus, there are different [cancellation modes][cancellationMode] that can be used.
+ * In the simple mode ([SIMPLE]), which is used by default, [resume] simply fails on cancelled waiters as
```
dkhalanskyjb, line 46: Not only does [resume] simply fail, but it also is necessarily called once for each cell, or, in other words, only works with a single cell, even if it's cancelled.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * described above, while in smart modes ([SMART_SYNC] and [SMART_ASYNC]) [resume] skips cells in the
```
dkhalanskyjb, line 47: The specific semantics of skipping are fuzzy here. If a cell is cancelled in a segment that is not removed, for example, this cell is not skipped, so saying that [resume] skips cells is not strictly correct. Another difference in the implementation is that [resume] does not fail when encountering a cancelled cell but proceeds to wake another one up, but this is not "skipping", as the cancelled cell is still visited, but more like guaranteeing that each call to [resume] wakes at least one waiter up if it's possible at all.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * [cancelled][CANCELLED] state. However, if cancellation happens concurrently with [resume], it can be illegal
+ * to simply skip the cell and resume the next waiter, e.g., if this cancelled waiter is the last one.
```
dkhalanskyjb, line 49: Either we have to specify it in the contract that [resume] may only be called when it is known in advance that there is a waiter in the queue or this example is incorrect: there is nothing inherently wrong with trying to resume a cell without an imminent waiter, it would still work, just suboptimally.

As an aside, the existing formal proof of an earlier version of the thread queue *does* make the assumption that [resume] is only ever called if [suspend] was called before, and it did allow for some stronger guarantees, so I'm in favor of forbidding resumption in absence of suspension.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * Thus, it is possible for [SegmentQueueSynchronizer] to refuse this [resume]. In order to support this logic,
+ * users should implement  [onCancellation] function, which returns `true` if the cell can be
+ * moved to the [cancelled][CANCELLED] state, or `false` if the [resume] that comes to this cell
```
dkhalanskyjb, line 52: > moved

This process is usually called a "transition". "Transition table" etc.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * should be refused. In the last case, [tryReturnRefusedValue] is invoked, so that the value
```
dkhalanskyjb, line 53: "the last" -> "the latter"

dkhalanskyjb: Some redundant clarification would be helpful, I think: "In the latter case, [resume] no longer tries to retrieve a waiter to pass a value to. Instead, [...]"

Without this, "the value" is also somewhat invalid, as there is no "value" in the context of the speech. "The value that [resume] is trying to pass to a waiter in the queue" or something similar would have to be used.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * can be put back to the data structure. However, it is possible for [tryReturnRefusedValue] to
```
dkhalanskyjb, line 54: That [tryReturnRefusedValue] may fail is stated as a fact of life, whereas it is our design choice. I think that this should be framed as if we were providing an API for specifying both a fast path and a slow path for returning refused values.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * fail, and [returnValue] is called in this case. Typically, this [returnValue] function coincide
```
dkhalanskyjb, line 55: "coincide" -> "coincides"

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * with the one that resumes waiters (e.g., [release][Semaphore.release] in [Semaphore]). The difference
+ * between [SMART_SYNC] and [SMART_ASYNC] modes is that in the first one the [resume] that comes to the
```
dkhalanskyjb, line 57: "first" -> "former"

dkhalanskyjb: [resume] does not need an article here, and "to the cell" -> "to a cell" (there are potentially many cancelled cells).

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * cell with cancelled waiter waits in a spin-loop until the cell is marked either with [CANCELLED]
+ * or [REFUSE] mark, while in [SMART_ASYNC] mode it replaces the cancelled waiter with the value of
```
dkhalanskyjb, line 59: Either "first" -> "SMART_SYNC" above, or "SMART_ASYNC" -> "the second" here. Or "the former/the latter", which I prefer.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * the resumption and finishes, so that the cancellation handler completes this [resume] eventually.
```
dkhalanskyjb, line 60: There is little rationale for why the two modes are needed. The next statement does mention that [SMART_ASYNC] may be undesirable due to problems with linearization, but the design of these modes does not follow at all. Additional clarification is needed, like mentioning that [SMART_SYNC] does this so that by the end of the operation the element is always done being passed to the data structure.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *  +-------+                         +--------------+  successful.   +---------------+  avoid memory leaks.  +---------+  ( `resume` SUCCEEDS )
+ *     |                                     |
+ *     |                                     | The continuation
+ *     | `resume` comes to                   | is cancelled.
+ *     | the cell before                     |
+ *     | `suspend` and puts                  V                                                                      ( THE CORRESPONDING `resume` SHOULD BE   )
+ *     | the element into           +-----------------+    The concurrent `resume` should be refused,    +--------+ ( REFUSED AND `tryReturnRefusedValue` OR )
+ *     | the cell, waiting          | cont: cancelled | -----------------------------------------------> | REFUSE | ( `returnValue` IF IT FAILS IS USED TO   )
+ *     | for `suspend` in           +-----------------+      `onCancellation` has returned `false`.      +--------+ ( RETURN THE VALUE BACK TO THE ORIGINAL  )
+ *     | SYNC mode.                          |         \                                                     ^      ( SYNCHRONIZATION PRIMITIVE              )
```
dkhalanskyjb, line 77: > "waiting for `suspend` in SYNC mode

implies that `resume` always waits for `suspend`, while being in some SYNC mode.

", waiting for `suspend` if the resume mode in SYNC", I think, is more clear.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *  +-------+   `suspend` succeeds.   +--------------+  `resume` is   +---------------+  store `RESUMED` to   +---------+  ( `cont` HAS BEEN   )
+ *  |  NULL | ----------------------> | cont: active | -------------> | cont: resumed | --------------------> | RESUMED |  ( RESUMED AND THIS  )
+ *  +-------+                         +--------------+  successful.   +---------------+  avoid memory leaks.  +---------+  ( `resume` SUCCEEDS )
+ *     |                                     |
+ *     |                                     | The continuation
+ *     | `resume` comes to                   | is cancelled.
+ *     | the cell before                     |
+ *     | `suspend` and puts                  V                                                                      ( THE CORRESPONDING `resume` SHOULD BE   )
+ *     | the element into           +-----------------+    The concurrent `resume` should be refused,    +--------+ ( REFUSED AND `tryReturnRefusedValue` OR )
+ *     | the cell, waiting          | cont: cancelled | -----------------------------------------------> | REFUSE | ( `returnValue` IF IT FAILS IS USED TO   )
```
dkhalanskyjb, line 75: > IT FAILS

What is "IT" here? Looks like it is [resume].

I'm not a pro of English punctuation, but I think that the following has the needed connotation?
> THE CORRESPONDING `resume` SHOULD BE REFUSED AND `tryReturnRefusedValue`, OR `returnValue` IF IT FAILS, IS USED TO RETURN THE VALUE BACK TO THE ORIGINAL SYNCHRONIZATION PRIMITIVE

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *     | for `suspend` in           +-----------------+      `onCancellation` has returned `false`.      +--------+ ( RETURN THE VALUE BACK TO THE ORIGINAL  )
```
dkhalanskyjb, line 76: Why "has"? I don't think it's needed here.

dkhalanskyjb: "ORIGINAL" -> "OUTER"?

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *     | SYNC mode.                          |         \                                                     ^      ( SYNCHRONIZATION PRIMITIVE              )
+ *     |                                     |          \                                                    |
+ *     |        Mark the cell as `CANCELLED` |           \                                                   |
+ *     |         if the cancellation mode is |            \  `resume` delegates its completing to            | `onCancellation` returned `false,
+ *     |        `SIMPLE` or `onCancellation` |             \   the concurrent cancellation handler if        | mark the state accordingly and
+ *     |                has returned `true`. |              \   `SMART_ASYNC` cancellation mode is used.     | complete the hung `resume`.
```
dkhalanskyjb, line 82: Ditto: don't see a reason for "has" here.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *     | `suspend` and puts                  V                                                                      ( THE CORRESPONDING `resume` SHOULD BE   )
+ *     | the element into           +-----------------+    The concurrent `resume` should be refused,    +--------+ ( REFUSED AND `tryReturnRefusedValue` OR )
+ *     | the cell, waiting          | cont: cancelled | -----------------------------------------------> | REFUSE | ( `returnValue` IF IT FAILS IS USED TO   )
+ *     | for `suspend` in           +-----------------+      `onCancellation` has returned `false`.      +--------+ ( RETURN THE VALUE BACK TO THE ORIGINAL  )
+ *     | SYNC mode.                          |         \                                                     ^      ( SYNCHRONIZATION PRIMITIVE              )
+ *     |                                     |          \                                                    |
+ *     |        Mark the cell as `CANCELLED` |           \                                                   |
+ *     |         if the cancellation mode is |            \  `resume` delegates its completing to            | `onCancellation` returned `false,
```
dkhalanskyjb, line 80: "completing" -> "completion"?

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *     |        `SIMPLE` or `onCancellation` |             \   the concurrent cancellation handler if        | mark the state accordingly and
+ *     |                has returned `true`. |              \   `SMART_ASYNC` cancellation mode is used.     | complete the hung `resume`.
+ *     |                                     |               +------------------------------------------+    |
+ *     |                                     |                                                           \   |
+ *     |    ( THE CONTINUATION IS    )       V                                                            V  |
+ *     |    ( CANCELLED AND `resume` ) +-----------+                                                     +-------+
+ *     |    ( EITHER FAILS IN THE    ) | CANCELLED | <-------------------------------------------------- | value |
```
dkhalanskyjb, line 87: I think that "EITHER" implies that any of the two outcomes is likely, but here we know in advance which one is the one. Removing the word "EITHER" fixes this.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ *     |    ( SIMPLE CANCELLATION    ) +-----------+   Mark the cell as `CANCELLED` if `onCancellation`  +-------+
+ *     |    ( MODE OR SKIPS THIS     )              returned true, complete the hung `resume` accordingly.
+ *     |    ( CELL IN THE SMART ONE  )
+ *     |
+ *     |
+ *     |            `suspend` gets   +-------+  ( ELIMINATION HAPPENED, )
+ *     |         +-----------------> | TAKEN |  (  BOTH `resume` and    )
```
dkhalanskyjb, line 94: Capitalization of "and".

dkhalanskyjb (commented):
Comments about the first comment.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+        return false
+    }
+
+    /**
+     * Tries to resume the next waiter and returns `true` if
+     * the resumption succeeds. However, it can fail due to
+     * several reasons. First of all, if the [resumption mode][resumeMode]
+     * is [synchronous][SYNC], this [resume] invocation may come
+     * before [suspend] and mark the cell as [broken][BROKEN];
+     * `false` is returned in this case. At the same time, according
```
dkhalanskyjb, line 236: "At the same time" -> "Similarly"? Though I don't see much similarity here. "Another reason for [resume] to fail is..."?

dkhalanskyjb: "according to X" commonly means "if we are to believe X", so here it may be a better idea to just say "if the cancellation is in simple mode and the next waiter is cancelled".

dkhalanskyjb (changes requested):

ndkoval:
We should add a comment [here](https://github.com/CruGlobal/android-gto-support/pull/671) once the PR is merged

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+        // The cell is broken, this can happen only in `SYNC` resumption mode.
+        assert { resumeMode == SYNC && segment.get(i) === BROKEN }
+        return false
+    }
+
+    /**
+     * Tries to resume the next waiter and returns `true` if
+     * the resumption succeeds. However, it can fail due to
+     * several reasons. First of all, if the [resumption mode][resumeMode]
+     * is [synchronous][SYNC], this [resume] invocation may come
```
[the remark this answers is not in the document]
ndkoval: as well as arriving :)

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+        // that this segment is not removed since at
+        // least the cell for this [suspend] invocation
+        // is not in the `CANCELLED` state.
+        val curTail = this.tail.value
+        val enqIdx = enqIdx.getAndIncrement()
+        val segment = this.tail.findSegmentAndMoveForward(id = enqIdx / SEGMENT_SIZE, startFrom = curTail,
+            createNewSegment = ::createSegment).segment
+        assert { segment.id == enqIdx / SEGMENT_SIZE }
+        // Try to install the continuation in the cell,
+        // this is the regular path.
```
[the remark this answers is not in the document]
ndkoval: Opposite to the situation when a concurrent `resume` comes earlier to this cell.

kotlinx-coroutines-core/jvm/test/linearizability/SegmentQueueSynchronizerLCStressTests.kt
ndkoval: It should be either `>=` or `==`. In my opinion, it is always better to decrement the counter if possible, so that the number of `resume` invocations is decreased. We should also add a check that the barrier is completed in the beginning.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
ndkoval: "move" sounds better

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * One may consider this abstraction as an infinite array with two counters that reference the next cells
+ * for enqueueing a continuation in [suspend] and for retrieving it in [resume]. To be short, when
+ * [suspend] is invoked, it increments the corresponding counter via fast `Fetch-And-Add` and stores the
+ * continuation into the cell. At the same time, [resume] increments its own counter and comes to the
+ * corresponding cell.
+ *
+ * A typical implementation via [SegmentQueueSynchronizer] performs some synchronization at first,
+ * (e.g., [Semaphore] modifies the number of available permits), and invokes [suspend] or [resume]
+ * after that. Following this pattern, it is possible in a concurrent environment that [resume]
+ * is executed before [suspend] (similarly to the race between `park` and `unpark` for threads),
```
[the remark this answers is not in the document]
ndkoval: it is important to descrive the race here

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
ndkoval: We do not have segments in the abstraction, cells in the CANCELLED state are always skipped when a smart cancellation mode is used.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
ndkoval: This is not about fast and slow paths. We need `tryReturnRefusedValue` since the part of cancellation is already done in `onCancellation`.

kotlinx-coroutines-core/common/src/internal/SegmentQueueSynchronizer.kt
```diff
@@ -0,0 +1,606 @@
...
+ * the value into the cell if it is still empty and finishes immediately; this way, the following [suspend] comes
+ * to this cell and simply grabs the element without suspension. At the same time, in the [synchronous][SYNC] mode,
+ * [resume] waits in a bounded spin-loop cycle until the put element is taken by a concurrent [suspend],
+ * marking the cell as broken if it is not taken after the spin-loop is finished. In this case, both the current
+ * [resume] and the [suspend] that comes to this broken cell fail and the corresponding operations on the data
+ * structure are typically restarted from the beginning.
+ *
+ * Since [suspend] can store [CancellableContinuation]-s, it is possible for [resume] to fail if the
+ * continuation is already cancelled. In this case, most of the algorithms retry the whole operation.
+ * However, if there are many consecutive cancelled waiters, it seems more efficient to skip them
```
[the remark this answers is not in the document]
ndkoval: I re-wrote this part

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+            // resuming readers now, so that the `RWR` flag is set in this case.
+            if (state.compareAndSet(s, state(s.ar + 1, false, 0, s.rwr)))
+                return true
+            // CAS failed => the state has changed.
+            // Re-read it and try to acquire a reader lock again.
+        }
+    }
+
+    private suspend fun readLockSlowPath() {
+        // Increment the number of waiting readers at first.
```
dkhalanskyjb, line 246:
```suggestion
        // Increment the number of waiting readers first.
```

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+            // Read the current state.
+            val s = state.value
+            // Is there a writer holding the lock or waiting for it?
+            if (s.wla || s.ww > 0) {
+                // The number of waiting readers was incremented
+                // correctly, wait for a reader lock in `sqsReaders`.
+                suspendCancellableCoroutineReusable<Unit> { sqsReaders.suspend(it) }
+                return
+            } else {
+                // A race has detected! The number of waiting readers increment
```
dkhalanskyjb, line 262:
```suggestion
                // A race was detected! The number of waiting readers increment
```

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+                // was wrong, try to decrement it back. However, it could
+                // already become zero due to a concurrent `writeUnlock`
+                // which gets the number of waiting readers, puts `0`
```
dkhalanskyjb, line 265:
```suggestion
                // which reads the number of waiting readers, puts `0`
```

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+    //                          of coroutines waiting for the write lock in `sqsWriters`;
+    // - `RWR` (resuming waiting readers) is a flag which is `true` when waiting readers
+    //                                    resumption is in progress.
+    private val state = atomic(0L)
+
+    private val sqsReaders = ReadersSQS() // The place where readers should suspend and be resumed.
+    private val sqsWriters = WritersSQS() // The place where writers should suspend and be resumed.
+
+    override suspend fun readLock() {
+        // Try to acquire a reader lock without suspension at first,
```
dkhalanskyjb, line 219:
```suggestion
        // Try to acquire a reader lock without suspension first,
```

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+        // this can be considered as a performance optimization and
+        // does not affect correctness.
```
dkhalanskyjb, line 221: It does. If not for this line, `readLock()` would spin until a writer arrived.

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+        }
+    }
+
+    private suspend fun readLockSlowPath() {
+        // Increment the number of waiting readers at first.
+        // If the current invocation should not suspend,
+        // the counter will be decremented back later.
+        waitingReaders.incrementAndGet()
+        // Check whether this operation should suspend. If not,
+        // try to decrement the counter of waiting readers and re-start.
```
dkhalanskyjb, line 251:
```suggestion
        // try to decrement the counter of waiting readers and restart.
```

kotlinx-coroutines-core/common/src/internal/ConcurrentLinkedList.kt
```diff
@@ -203,7 +203,7 @@ internal abstract class Segment<S : Segment<S>>(val id: Long, prev: S?, pointers
      * There are no pointers to this segment from outside, and
      * it is not a physical tail in the linked list of segments.
      */
-    override val removed get() = cleanedAndPointers.value == maxSlots && !isTail
+    override val removed get() = !isTail && cleanedAndPointers.value == maxSlots
 
     // increments the number of pointers if this segment is not logically removed.
     internal fun tryIncPointers() = cleanedAndPointers.addConditionally(1 shl POINTERS_SHIFT) { it != maxSlots || isTail }
```
dkhalanskyjb, line 209: (Duplicating the bug report here for documentation purposes) The changes to `removed` were introduced to fix a bug that caused a segment that was already seen to be logically removed to become alive once again. However, this fix is not sufficient.

Consider the following execution: in a tail segment that has `cleanedAndPointers.value == max_slots`, `tryIncPointers()` is executed concurrently with adding a new segment to the tail's end. First, `tryIncPointers()` observes that this segment is, in fact, the tail. Second, the segment stops being the tail and becomes logically removed. Finally, `tryIncPointers()` finishes, adding a new pointer and raising the segment from the dead.

The suggested fix (which was employed in the formal proof):
```kotlin
override val removed get() = cleanedAndPointers.value == maxSlots
internal fun tryIncPointers() = cleanedAndPointers.addConditionally(1 shl POINTERS_SHIFT) { it != maxSlots }
internal fun decPointers() = cleanedAndPointers.addAndGet(-(1 shl POINTERS_SHIFT)) == maxSlots
fun onSlotCleaned() {
    if (cleanedAndPointers.incrementAndGet() == maxSlots) remove()
}
```
This way, a once-removed segment can't return to life.

Then, in `remove()`, don't fail but instead exit early if it turns out that the segment is the tail one.

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+                    // `writeUnlock` and a reader lock is going to be
+                    // passed via `sqsReaders`? Suspend in this case --
+                    // it is guaranteed that the lock will be provided
+                    // when this concurrent `writeUnlock` completes.
+                    if (wr == 0) {
+                        suspendCancellableCoroutineReusable<Unit> { sqsReaders.suspend(it) }
+                        return
+                    }
+                    // Otherwise, try to decrement the number of waiting
+                    // readers and re-try the operation from the beginning.
```
dkhalanskyjb, line 282:
```suggestion
                    // readers and retry the operation from the beginning.
```

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
```diff
@@ -0,0 +1,603 @@
...
+                    if (waitingReaders.compareAndSet(wr, wr - 1)) {
+                        // Try again starting from the fast path
+                        // since the state has changed.
```
dkhalanskyjb, line 285:
```suggestion
                        // as the state has changed.
```

dkhalanskyjb (changes requested):
A small batch of suggestions for now.

kotlinx-coroutines-core/common/src/internal/ConcurrentLinkedList.kt
ndkoval: Thanks for the duplication, I will fix it in a separate PR

kotlinx-coroutines-core/common/src/sync/ReadWriteMutex.kt
ndkoval: Agree, removed this note

GavinRay97:
Just out of curiosity, why did this die?

It seems well-written/documented (can't comment on the implementation)

qwwdfsad:
There were unresolved questions about public API shape and how to expose read/write parts.
We may return to it after #3103

Tmpod:
> We may return to it after #3103

That was merged earlier this year (congrats to all btw, big PR!). Has there been any internal progress on this since then?

qwwdfsad:
@Tmpod we are evaluating this change on IDEA codebase.
 Could you please elaborate on your interest in this primitive? E.g. what use-cases are you aiming to solve with it?

PR #2057, opened 2020-05-25
Properly check identity of caught AbortFlowException in Flow.first op…

### LINKED ISSUE #2051: Flow<T>.first operator throws NoSuchElementException where CancellationException must be thrown
Consider the following test

```kotlin
@Test
fun flowFirstOperatorTest() {
    assertDoesNotThrow {
            launch {
                (0..10000).asFlow().flatMapMerge(Int.MAX_VALUE) {
                    channelFlow {
                        val v = channelFlow { send(1) }.first()
                        send(v)
                    }
                }.take(100)
            }
        }
    }
}
```

It is expected to take N elements and correctly finish the coroutine but actually it throws `NoSuchElementException`.

It happens because `AbortFlowException`, that is used inside first operator, extends `CancellationException` and if upstream flow is cancelled before calling `collect`, it must throw `CancellationException` instead of `NoSuchElementException`.

Test passes when `AbortFlowException` is extended from `Throwable`

```kotlin
@Test
fun flowFirstOperatorTest() {
    assertDoesNotThrow {
            launch {
                (0..10000).asFlow().flatMapMerge(Int.MAX_VALUE) {
                    channelFlow {
                        val v = channelFlow { send(1) }.firstFixed()
                        send(v)
                    }
                }.take(100)
            }
        }
    }
}

object AbortFlowException : Throwable("", null, true, false)

suspend fun <T> Flow<T>.firstFixed(): T {
    var result: T? = null
    try {
        collect { value ->
            result = value
            throw AbortFlowException
        }
    } catch (e: AbortFlowException) {
        // Do nothing
    }

    if (result == null) throw NoSuchElementException("Expected at least one element")
    return result!!
}
```

This issue is relevant for firstOrNull operator and, perhaps, other similar ones.

qwwdfsad:
…erator.

It fixes two problems:
    * NoSuchElementException can be thrown during cancellation sequence (see FirstJvmTest that reproduces this problem with explanation)
    * Cancellation can be accidentally suppressed and flow activity can be prolonged

Fixes #2051

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
+++ b/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
@@ -84,15 +84,18 @@ public suspend fun <T: Any> Flow<T>.singleOrNull(): T? {
  */
 public suspend fun <T> Flow<T>.first(): T {
     var result: Any? = NULL
-    try {
-        collect { value ->
+    val collector = object : FlowCollector<T> {
+        override suspend fun emit(value: T) {
             result = value
-            throw AbortFlowException(NopCollector)
+            throw AbortFlowException(this)
         }
-    } catch (e: AbortFlowException) {
-        // Do nothing
     }
-
+    try {
+        collect(collector)
+    } catch (e: AbortFlowException) {
+        // Do not suppress other cancellation sources
+        e.checkOwnership(collector)
+    }
     if (result === NULL) throw NoSuchElementException("Expected at least one element")
     return result as T
 }
@@ -103,17 +106,20 @@ public suspend fun <T> Flow<T>.first(): T {
  */
 public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
     var result: Any? = NULL
-    try {
-        collect { value ->
+    val collector = object : FlowCollector<T> {
+        override suspend fun emit(value: T) {
             if (predicate(value)) {
                 result = value
-                throw AbortFlowException(NopCollector)
+                throw AbortFlowException(this)
             }
         }
-    } catch (e: AbortFlowException) {
-        // Do nothing
     }
-
+    try {
+        collect(collector)
+    } catch (e: AbortFlowException) {
+        // Do not suppress other cancellation sources
+        e.checkOwnership(collector)
+    }
     if (result === NULL) throw NoSuchElementException("Expected at least one element matching the predicate $predicate")
     return result as T
 }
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -265,8 +265,7 @@ internal fun <T> getOrCreateCancellableContinuation(delegate: Continuation<T>):
 public suspend inline fun <T> suspendAtomicCancellableCoroutine(
     holdCancellability: Boolean = false,
     crossinline block: (CancellableContinuation<T>) -> Unit
-): T =
-    suspendAtomicCancellableCoroutine(block)
+): T = suspendAtomicCancellableCoroutine(block)
```
elizarov, line 268: No need to change this line. Will cause unneeded conflicts with #1937

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -84,15 +84,18 @@ public suspend fun <T: Any> Flow<T>.singleOrNull(): T? {
...
             result = value
-            throw AbortFlowException(NopCollector)
+            val e = AbortFlowException(NopCollector)
+            expectedException = e
+            throw e
         }
     } catch (e: AbortFlowException) {
-        // Do nothing
+        // Do not suppress other cancellation sources
+        if (!e.isSameExceptionAs(expectedException)) throw e
```
elizarov, line 97: Now we have two way to check ownership of `AbortFlowException` -- its identity and its `owner` constructor parameter. It seems too much. Can we just use `owner` here instead, just like `combine` and `take` are doing?

qwwdfsad: We can't because at this point we do not have a reference to the "owner" (aka `FlowCollector`).
Alternatively, we can introduce another collect function with the following signature:
`internal suspend inline fun <T> Flow<T>.collectInternal(crossinline action: suspend FlowCollector<T>.(value: T) -> Unit): FlowCollector<T>`, but it may interfere with compiler optimization and obfuscate the code for external readers

elizarov: It still seems safer than checking exception identities. It complicated interaction with various debugging features that copy exceptions. You can do without `collectInternal`, but just create `val flowCollector = object : FlowCollector { ... }` and then call `collect(flowCollector)`. You can also introduce `FlowCollector { ... }` builder function that will go away in Kotlin 1.4 when we turn `FlowCollector` into a `fun interface`.

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -141,15 +127,28 @@ public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
...
+            result = it
+            true
+        } else {
+            false
+        }
+    }
+    return result
+}
+
+internal suspend inline fun <T> Flow<T>.collectWhile(crossinline block: suspend (value: T) -> Boolean) {
```
elizarov, line 141: This looks like `collectUntil`.

saurabhkpatel:
Are there any timelines to release this fix?

PR #2061, opened 2020-05-26
Remove keep rules mentioning kotlinx.coroutines.android

mkj-gram:
The rule -keepnames class kotlinx.coroutines.android.AndroidDispatcherFactory {} is present in kotlinx.coroutines.android/META-INF/proguard/coroutines.pro and should therefore not be needed. If the -keepnames class kotlinx.coroutines.android.AndroidExceptionPreHandler {} is needed, it should go into kotlinx.coroutines.android/META-INF/proguard/coroutines.pro

### DIFF
--- a/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
+++ b/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
@@ -1,8 +1,6 @@
 # ServiceLoader support
 -keepnames class kotlinx.coroutines.internal.MainDispatcherFactory {}
 -keepnames class kotlinx.coroutines.CoroutineExceptionHandler {}
--keepnames class kotlinx.coroutines.android.AndroidExceptionPreHandler {}
--keepnames class kotlinx.coroutines.android.AndroidDispatcherFactory {}
 
 # Most of volatile fields are updated with AFU and should not be mangled
 -keepclassmembernames class kotlinx.** {
@@ -13,3 +11,4 @@
 -keepclassmembernames class kotlin.coroutines.SafeContinuation {
     volatile <fields>;
 }
+


elizarov:
I'm not sure I understand why it should not be needed? Can you please elaborate?

mkj-gram:
The type kotlinx.coroutines.android.AndroidDispatcherFactory do not exist in the kotlinx-coroutines-core package, it is a type that is found in the kotlinx-coroutines-android package (as far as I can see). 

When proguard or R8 reads kotlinx-coroutines-core for projects where kotlinx-coroutines-android is not referenced, the shrinkers will see the rule -keepnames class kotlinx.coroutines.android.AndroidDispatcherFactory {} but since that type is missing it will have no effect - and R8 will print information about the rule being unused.

For android projects, where kotlinx-coroutines-android is referenced, it could have an effect. But kotlinx-coroutines-android defines its own META-INF/proguard/coroutines.pro file, that includes the rule: 
-keepnames class kotlinx.coroutines.android.AndroidDispatcherFactory {}

As a result, removing the above rule should not make any difference except that warnings are no longer printed from R8.

Now, the kotlinx-coroutines-android do not define:
-keepnames class kotlinx.coroutines.android.AndroidExceptionPreHandler {}

However, the type is only defined in the kotlinx-coroutines-android package, so if it is important for kotlinx-coroutines-android to function correctly, it should be defined there and not in kotlinx-coroutines-core.

elizarov:
Could you then, please, add the missing `-keepnames class kotlinx.coroutines.android.AndroidExceptionPreHandler` in  `kotlinx-coroutines-android` too, in the same PR and rebase your PR to `develop` branch of the library.

mkj-gram:
Just checked again and that rule should not be necessary - for one, the -keepnames will only keep the name if the class survives the shrinking phase and there seem to not be any reflective call in the codebase to kotlinx.coroutines.android.AndroidExceptionPreHandler.

Additionally, the class itself is marked with @keep: https://github.com/Kotlin/kotlinx.coroutines/blob/583ec6e862fb70c7fc0232af948a028bab3b20de/ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt#L13

and that is why the class stays in the output, and not because of the -keepnames rule.

Rebased on the developer branch.

OscarSpruit:
This change broke the minified builds of our Android app. As a workaround we added the rules to our own ProGuard config. Might be nice to either document this or add the rules to kotlinx-coroutines-android

elizarov:
> This change broke the minified builds of our Android app. As a workaround we added the rules to our own ProGuard config. Might be nice to either document this or add the rules to kotlinx-coroutines-android

Can you share the build files that would reproduce the problem? What version of AGP and what kind of minifier are you using?

OscarSpruit:
I'm not allowed to share the build files, but I will try to make a sample project that reproduces it later today.
We use AGP version 4.0.1 and R8 as minifier

OscarSpruit:
@elizarov I couldn't get a sample to reproduce working, but I think I found the cause and fixed it with #2154

PR #2074, opened 2020-06-01
Refactor mpp/native build, introduce "concurrent" source set, test launcher

elizarov:
Refactor mpp/native build, introduce "concurrent" source set, test launcher

New source sets:
* "concurrent" source set is shared between "jvm" and "native"
* "native" source set is subdivided into "nativeDarwin" (Apple) and "nativeOther" (Linux, etc)

Native tests are launched in two variants:
* A default "test" task runs tests with memory leak checker from "mainNoExit" entry point.
* A special "backgroundTest" task runs tests in a background worker from "mainBackground" entry point.

Other build improvement:
* Modernize old-style IDEA-active hacks to kts helper.
* Extract versions of JS test runner dependencies.
* Remove redundant google repo reference from android tests.


elizarov:
Fixed import to IDEA, so that it now resolves properly in IDEA.

kotlinx-coroutines-core/native/src/WorkerMain.native.kt
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+internal expect inline fun workerMain(block: () -> Unit)
```
qwwdfsad, line 7: Why does this function belong to the `main` source-set instead of `test` one?
It's used only from within tests

kotlinx-coroutines-core/nativeDarwin/test/Launcher.kt
```diff
@@ -0,0 +1,28 @@
...
+import platform.CoreFoundation.*
+import kotlin.native.concurrent.*
+import kotlin.native.internal.test.*
+import kotlin.system.*
+
+// This is a separate entry point for tests in background
+fun mainBackground(args: Array<String>) {
+    val worker = Worker.start(name = "main-background")
+    worker.execute(TransferMode.SAFE, { args.freeze() }) {
+        val result = testLauncherEntryPoint(it)
```
qwwdfsad, line 16: `testLauncherEntryPoint` is K/N internal API to run tests, right?

kotlinx-coroutines-core/build.gradle
```diff
@@ -31,12 +78,18 @@ kotlin {
...
-            // binary compatibility support
-            if (targetName.contains("jvm") && compilation.compilationName == "main") {
-                compileTask.kotlinOptions.freeCompilerArgs += ["-Xdump-declarations-to=${buildDir}/visibilities.json"]
+        // Configure additional binaries and test runs -- one for each OS
+        if (["macos", "linux", "mingw"].any { name.startsWith(it) }) {
+            binaries {
+                // Test for memory leaks using a special entry point that does not exit but returns from main
+                binaries.getTest("DEBUG").freeCompilerArgs += ["-e", "kotlinx.coroutines.mainNoExit"]
+                // Configure a separate test where code runs in background
+                test("background", [org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType.DEBUG]) {
```
qwwdfsad, line 87: Why it is not possible to use `-trw` flag instead of hand-rolled implementation?

kotlinx-coroutines-core/native/src/WorkerMain.native.kt
elizarov: It is used from the main code in `native-mt` branch and I'm trying to minimize the diff in `native-mt` branch.

kotlinx-coroutines-core/nativeDarwin/test/Launcher.kt
elizarov: Yes

kotlinx-coroutines-core/build.gradle
elizarov: It does not have `CFRunLoopRun` in the main thread, thus attempts to dispatch tasks to the main thread from `-trw` will hang.

kotlinx-coroutines-core/native/src/WorkerMain.native.kt
elizarov: Added comment.

elizarov:
I've also added additional workarounds to improve IDE experience under 1.4. It is not ideal under 1.4 (platform declaration like `CFRunLoopRun` are not resolved), but it is better, because stdlib resolves properly.

kotlinx-coroutines-core/build.gradle
```diff
@@ -110,12 +110,26 @@ compileKotlinMetadata {
...
+    // Kotlin 1.3.x: Look for platform libraries in "compile" configuration
     def compileConfiguration = configurations[hostMainCompilation.compileDependencyConfigurationName]
-    def hostNativePlatformLibs = files(provider {
-        compileConfiguration.findAll {
-            it.path.endsWith(".klib") || it.absolutePath.contains("klib${File.separator}platform")
+    // Kotlin 1.4.x: Look for platform libraries in "implementation" for default source set
+    def implementationConfiguration = configurations[hostMainCompilation.defaultSourceSet.implementationMetadataConfigurationName]
+    // Now find the libraries:
+    // Kotlin 1.3.x: Find all platform libs
+    // Kotlin 1.4.x: Finds platform libs & stdlib, but platform declarations are still not resolved due to IDE bugs
```
LouisCAD, line 119: Do we have the links to these bugs to allow future reviews to check if they are resolved? If so, it'd be nice to add extra comment lines for them IMO.

elizarov: They will be fixed before 1.4 release and affect only HMPP projects anyway.

elizarov:
I've dropped hacks for 1.3.x and enabled HPMM. Now need to figure out why tests with workers don't pass under K/N.

PR #2076, opened 2020-06-02
Improve grammar in docs/Basics

medavox:
This is just a few small corrections to some minor second-language grammar mistakes; also improving readability and punctuation in a few places.

I've also replaced the phrase 'get full code here' with 'get the full code here', because that sounds more natural to me. But Englishes do differ, and is a very minor thing anyway. I understand this is a pretty ubiquitous phrase, and I've only changed it in this one document, so I've kept those changes in a separate commit so you can reject them (and keep the others) if you don't think that's a worthwhile change.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -41,7 +41,7 @@ fun main() {
 
 </div>
 
-> You can get full code [here](../kotlinx-coroutines-core/jvm/test/guide/example-basic-01.kt).
+> You can get the full code [here](../kotlinx-coroutines-core/jvm/test/guide/example-basic-01.kt).
 
 You will see the following result:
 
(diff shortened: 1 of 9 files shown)


elizarov:
Thanks a lot. Indeed, that `the full code` reads better. Can you then, please, search & replace for it in all `*.md` files in docs folder for consistency and add to this PR, too.

medavox:
OK, will do.

medavox:
Done. I'm still reading the guide and learning how to use coroutines, so there may be further proofreading commits to come, but no promises.

elizarov:
Thanks. I've merged it into `master` with a couple of corrections (consistent _the_ `runBlocking` and added docs link to the `coroutineScope` function mention).

medavox:
thanks!

PR #2078, opened 2020-06-03
Rename 'foo' function to 'simple' in flow docs

elizarov:
The name 'foo' might be hard to get for novice developers. It is really hard to find a "meaningful" name for all examples in the docs, so this is simply a consistent solution. The alternative is to something like "simpleFlow", but adding a "Flow" suffix to the name implicitly condones Hungarian notation, which is not aligned with Kotlin coding style.

### DIFF
--- a/docs/flow.md
+++ b/docs/flow.md
@@ -54,16 +54,16 @@ multiple asynchronously computed values? This is where Kotlin Flows come in.
 ### Representing multiple values
 
 Multiple values can be represented in Kotlin using [collections]. 
-For example, we can have a function `foo()` that returns a [List] 
+For example, we can have a `simple` function that returns a [List] 
 of three numbers and then print them all using [forEach]:
 
 <div class="sample" markdown="1" theme="idea" data-min-compiler-version="1.3">
 
 ```kotlin
-fun foo(): List<Int> = listOf(1, 2, 3)
+fun simple(): List<Int> = listOf(1, 2, 3)
  
 fun main() {
-    foo().forEach { value -> println(value) } 
+    simple().forEach { value -> println(value) } 
 }
 ```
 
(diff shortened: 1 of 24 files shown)


elizarov:
cc @LouisCAD

LouisCAD:
> The alternative is to something like "simpleFlow", but adding a "Flow" suffix to the name implicitly condones Hungarian notation, which is not aligned with Kotlin coding style.

Suffixing a symbol for the sake of an example isn't endorsing Hungarian notation to me.

In application code, I often suffix the symbols of type `Flow` because I'll need to disambiguate with the parameter name of the lambda passed to `collect`.
**Example:**
```kotlin
val isOnFlow = whatever().distinctUntilChanged()
isOnFlow.collectLatest { isOn -> doSomeStuff() }
```
compared to:
```kotlin
val isOn = whatever().distinctUntilChanged()
isOn.collectLatest { isOn -> doSomeStuff() }
```

That's why I personally prefer `someFlow()` or `exampleFlow()` to `simple()`.

I'm also not a fan of `simple()` because of the meaning of the word "simple", that I find not relevant in this context, while "some" doesn't hint at anything special and directs attention to the rest (here, the operations performed on the flow itself).

Of course, for the examples where the fact that the flow is simple is relevant (e.g. showing how to make a simple flow with the most basic builder, `flow { }`), I think it makes sense to have the word "simple" there, although I'd lean towards `simpleFlow()` rather just `simple()`.


That's ony my point of view. That's alright if you disagree.
If you think some points I made are not right, I'm curious to knwo which ones and why.

Also, maybe we can ask other developers on Kotlin's Slack, or [developers in this Twitter thread for example](https://twitter.com/hinchman_amanda/status/1197073903262322688) (cc @ahinchman1) to get other angles and confidence in the chosen naming?

LouisCAD:
A few other namings that I (personally) find alright and less confusing than `simple`: `someNumbers()` & `numbers()`. Thought about this after [this short Slack thread](https://kotlinlang.slack.com/archives/C1CFAFJSK/p1591311717106900).

PR #2091, opened 2020-06-16
Fix missing unbox state in native runBlocking

qwwdfsad:


elizarov:
I've cherry-picked into an upcoming build of `native-mt` branch, too.

PR #2093, opened 2020-06-16
Minor improvements of debugging experience

qwwdfsad:
Recognizable toString is not only pretty and aligned with our own glossary, but also significantly affects "group by dispatcher" feature in the upcoming IDEA debugger, see the screenshot: https://imgur.com/a/2zqM5pn

The second commit is just nice to have while we are here

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
@@ -22,10 +22,10 @@ internal object DefaultScheduler : ExperimentalCoroutineDispatcher() {
     )
 
     override fun close() {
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
```diff
@@ -190,7 +196,7 @@ private class LimitingDispatcher(
     }
 
     override fun toString(): String {
-        return "${super.toString()}[dispatcher = $dispatcher]"
+        return name ?: "${super.toString()}[dispatcher = $dispatcher]"
```
qwwdfsad, line 199: This will be useful in upcoming "dispatcher view" feature

qwwdfsad:
Dispatchers.Main is coming a bit later

PR #2096, opened 2020-06-19
JobSupport: non-atomic adding to a Job's list of listeners

SokolovaMaria:
Getting rid of using DCSS while adding a `JobNode` to the list


kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -517,8 +523,12 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
                 ?: InvokeOnCompletion(this, handler)
     }
 
-    private fun addLastAtomic(expect: Any, list: NodeList, node: JobNode<*>) =
-        list.addLastIf(node) { this.state === expect }
+    private fun addLastAtomic(expect: Any, list: NodeList, node: JobNode<*>): Boolean {
+        list.addLast(node)
+        if (this.state === expect) return true
```
LouisCAD, line 528:
```suggestion
    private fun addLastAtomic(expectedState: Any, list: NodeList, node: JobNode<*>): Boolean {
        list.addLast(node)
        if (this.state === expectedState) return true
```

I think this naming is more explicit and avoids confusion with the `expect` soft keyword.

LouisCAD:
@arash1999 I think your pocket left an empty review…

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -517,8 +533,13 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
                 ?: InvokeOnCompletion(this, handler)
     }
 
-    private fun addLastAtomic(expect: Any, list: NodeList, node: JobNode<*>) =
-        list.addLastIf(node) { this.state === expect }
+    private fun addLast(expectedState: Any, list: NodeList, node: JobNode<*>): Boolean {
```
qwwdfsad, line 536: This method not only adds to the tail, but does something else, please properly documented it and describe why this machinery is required. Especially `ChildCompletion`

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -453,14 +452,17 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
-        var nodeCache: JobNode<*>? = null
+        var nodeAdded: JobNode<*>? = null
         loopOnState { state ->
             when (state) {
                 is Empty -> { // EMPTY_X state -- no completion handlers
                     if (state.isActive) {
                         // try move to SINGLE state
-                        val node = nodeCache ?: makeNode(handler, onCancelling).also { nodeCache = it }
-                        if (_state.compareAndSet(state, node)) return node
+                        val node = makeNode(handler, onCancelling)
```
qwwdfsad, line 461: why this code was changed? Why is it important not to cache node now?

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -1457,33 +1516,48 @@ internal abstract class JobCancellingNode<out J : Job>(job: J) : JobNode<J>(job)
...
-    override fun toString(): String = "ChildHandle[$childJob]"
+    override fun toString(): String =  "ChildHandle[$childJob]"
 }
 
 // Same as ChildHandleNode, but for cancellable continuation
 internal class ChildContinuation(
     parent: Job,
     @JvmField val child: CancellableContinuationImpl<*>
 ) : JobCancellingNode<Job>(parent) {
+    private val _invoked = atomic(false)
```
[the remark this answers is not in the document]
SokolovaMaria: `atomicfu` does not support access to the parent field from an inheritor right now. This is a good use case, I'll create a ticket on it.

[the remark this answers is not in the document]
qwwdfsad: You can make `_invoked` private in the parent and introduce a method for CAS'ing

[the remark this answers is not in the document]
SokolovaMaria: Yes, sure, done this way.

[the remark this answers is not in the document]
SokolovaMaria: Moved `isInvoked` field and CAS to the `JobNode`, as we should prevent all JobNodes from the repeated invocation.

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -453,14 +452,17 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
         loopOnState { state ->
             when (state) {
                 is Empty -> { // EMPTY_X state -- no completion handlers
                     if (state.isActive) {
                         // try move to SINGLE state
-                        val node = nodeCache ?: makeNode(handler, onCancelling).also { nodeCache = it }
-                        if (_state.compareAndSet(state, node)) return node
+                        val node = makeNode(handler, onCancelling)
+                        if (_state.compareAndSet(state, node)) {
+                            nodeAdded = node
```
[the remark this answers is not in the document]
SokolovaMaria: I've fixed that, there was no reason for changing this logic. `nodeAdded` is actually the same cache, that should store a node added to the list, no matter if CAS succeeded or not.

kotlinx-coroutines-core/common/src/JobSupport.kt
SokolovaMaria: I've removed this extra logic from `addLast` method, it now just tries to add a node to the list.

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -481,9 +478,14 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
                                 // or we are adding a child to a coroutine that is not completing yet
                                 if (rootCause == null || handler.isHandlerOf<ChildHandleNode>() && !state.isCompleting) {
                                     // Note: add node the list while holding lock on state (make sure it cannot change)
-                                    val node = makeNode(handler, onCancelling)
-                                    val res = addLast(state, list, node).also { nodeAdded = node }
-                                    if (!res) return@loopOnState // retry
+                                    val node = nodeAdded ?: makeNode(handler, onCancelling).also { nodeAdded = it }
+                                    if (!addLast(state, list, node)) {
+                                        // parent failed to add ChildCompletion node to the child, the child became Completed
```
SokolovaMaria, line 483: I've moved that `ChildCompletion` logic from `addLast` impl here.

Consider this scenario:
The Parent is `Finishing` and starts to wait for a child to complete. It tries to add `ChildCompletion` node to the child:
 
if it succeeds and `ChildCompletion` is added before the child changes its state to `Completed`, then the parent just waits for the child

if the child has already completed, and there are no other children left to wait for, then the parent starts it's final transition to `Completed` state.

But as adding to the list is not atomic with the state change now, the child may change it's state to `Completed` and see the list with `ChildCompletion` added before this node is disposed, invoke `ChildCompletion` and start parent's terminal state transition to `Completed` state (`finalizeFinishingState`). While the parent reports, that the child has already completed, no one to wait for and starts it's transition to the `Completed` state itself (concurrently with the child). 

So, I check here whether `ChildCompletion` was already invoked and if so, the parent should wait for the child completion.

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -480,7 +484,14 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
                                 // or we are adding a child to a coroutine that is not completing yet
                                 if (rootCause == null || handler.isHandlerOf<ChildHandleNode>() && !state.isCompleting) {
```
qwwdfsad, line 485: [nit] while we are here, please remove `isHandlerOf` function and replace it with `handler is ChildHandleNode`

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -456,12 +456,16 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
+        var nodeAdded: JobNode? = null // prevents handler from double invocation
         loopOnState { state ->
             when (state) {
                 is Empty -> { // EMPTY_X state -- no completion handlers
                     if (state.isActive) {
                         // try move to SINGLE state
-                        if (_state.compareAndSet(state, node)) return node
+                        if (_state.compareAndSet(state, node)) {
+                            nodeAdded = node
+                            return node
```
qwwdfsad, line 467: No need to update `nodeAdded` here as we immediately return from this function. These lines can be left intact

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -520,8 +547,12 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
         return node
     }
 
-    private fun addLastAtomic(expect: Any, list: NodeList, node: JobNode) =
-        list.addLastIf(node) { this.state === expect }
+    private fun addLast(expectedState: Any, list: NodeList, node: JobNode): Boolean {
+        list.addLast(node)
+        if (this.state === expectedState) return true
+        node.dispose()
```
qwwdfsad, line 553: Do we need to call `dispose` here? If the state was updated, then there is no references to the list -> no need to remove node from the list

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -490,17 +501,33 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
...
+                            }
                         }
                     }
                 }
                 else -> { // is complete
                     // :KLUDGE: We have to invoke a handler in platform-specific way via `invokeIt` extension,
                     // because we play type tricks on Kotlin/JS and handler is not necessarily a function there
-                    if (invokeImmediately) handler.invokeIt((state as? CompletedExceptionally)?.cause)
+                    if (invokeImmediately) {
+                        if (nodeAdded != null) {
```
qwwdfsad, line 525: Seems like it can be replaced with `(nodeAdded ?: node).invoke((state as? CompletedExceptionally)?.cause)`

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -490,17 +501,33 @@ public open class JobSupport constructor(active: Boolean) : Job, ChildJob, Paren
                         }
                         if (rootCause != null) {
                             // Note: attachChild uses invokeImmediately, so it gets invoked when adding to cancelled job
-                            if (invokeImmediately) handler.invokeIt(rootCause)
+                            if (invokeImmediately) {
+                                if (nodeAdded != null) {
+                                    nodeAdded!!.invoke(rootCause)
```
qwwdfsad, line 506: Seems like it can be replaced with `(nodeAdded ?: node).invoke(rootCause)`

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/JobSupport.kt
qwwdfsad: That's a really useful comment, thanks for the explanation!

I would appreciated if you could embed it into the code comment in a slightly more condensed form

kotlinx-coroutines-core/common/src/JobSupport.kt
mvicsokolova: I guess that's not correct. Consider transition to `Finishing` state where the list reference is captured or  transition from `Finishing` to `Completed` state that includes invocation of completion handlers from the list after the state cas (`completeStateFinalization`).

And skipping dispose may lead to duplicate nodes added to the list (when we retry to add one in `invokeOnCompletion`).

kotlinx-coroutines-core/common/src/JobSupport.kt
mvicsokolova: Kotlin/Js compilation will fail then with "Incompatible types" error.

qwwdfsad: Then please drop a comment and report it to JS team

mvicsokolova: Ok

qwwdfsad (approved):

elizarov:
https://github.com/Kotlin/kotlinx.coroutines/blob/8a220f3b1e49724c8158e7b605c831643d825b10/kotlinx-coroutines-core/common/src/JobSupport.kt#L1533

This comment is outdated (cannot be invoked multiple times anymore).

kotlinx-coroutines-core/common/src/JobSupport.kt
elizarov: Unfortunately, the `LinkedList` algorithm does not support removing and re-adding the same node. Consider the following scenario:
1. T1 calls `addLastIf` in `Incomplete` state and started executing `list.addLast(node)`, stalls (looses CPU) mid-progress inside addition.
2. T2 calls `addLastIf`, too, sees non-finished add by T1, starts "helping" it. but gets stalled too.
3. T3 transitions the state from `Incomplete` to `Finishing`.
4. T1 completes its `addLastIf`, sees that state has changed to `Finishing` and calls `node.dispose()`, which removes the node from the list. 
5. T1 now returns from `addLastIf` to `invokeOnCompletion` and retries. It discovers that the state is now `Finishing` and calls `addLastIf` again.
6. T1's calls to `addLastIf` calls `list.addLast(node)`. The first thing this call does is that it resets `prev` and `next` pointers in Node to point to itself. Assume it stalls right after that. 
7. T2 resumes and continues helping the original addition to the node. However, `prev` and `next` pointers are reset to point to the node itself, this helping algorithm can do all sorts of damage or hang.

I suggest the following change to fix it:

1. Inline both usages of `addLastIf` into their call sites. This will make reasoning about it easier. The first call to `addLastIf` from line 493 is trivial since the state is already `Finishing` there and it can only transition to `Complete`, so it does not suffer from "we need to remove it" problem at all. It is Ok if it stays added to the list. 

2. The only problematic call site of `addListIf` is on the line 526 when the state is `Incomplete`. When you detect that the state has changed, you can actually check _how_ its changed. If it becomes `Complete`, then it is trivial as above (you can safely retry) ,but if it became `Finishishing`, then you can actually use a boolean flag meaning "we've already added the node to the list", and check this flag before trying to add the node in the corresponding branches of `invokeOnCompletion` code to avoid adding the node the second time.


Instead of "removing the node to be readded later" I s

elizarov (changes requested):

kotlinx-coroutines-core/common/src/CompletionHandler.common.kt
```diff
@@ -44,4 +44,6 @@ internal expect val CancelHandlerBase.asHandler: CompletionHandler
 // because we play type tricks on Kotlin/JS and handler is not necessarily a function there
 internal expect fun CompletionHandler.invokeIt(cause: Throwable?)
 
+// :KLUDGE: We have to use `isHandlerOf` extension, because performing this type check directly in the code
+// causes Incompatible types error during Kotlin/JS compilation
```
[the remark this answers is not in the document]
LouisCAD: @elizarov Did you mean "Link" instead of "Like"?

LouisCAD:
🐣

The incubation period of this pull request is complete.

PR #2098, opened 2020-06-21
Fix typos

Inego:


docs/exception-handling.md
```diff
@@ -428,19 +428,19 @@ fun main() = runBlocking {
         supervisorScope {
             val child = launch {
                 try {
-                    println("Child is sleeping")
+                    println("The child is sleeping")
```
qwwdfsad, line 431: Please also run `./gradlew knit` to update actual `.kt` files

docs/coroutine-context-and-dispatchers.md
```diff
@@ -625,7 +625,7 @@ fun main() = runBlocking<Unit> {
 In this example we launch a new coroutine in a background thread pool using [Dispatchers.Default], so
 it works on a different thread from the thread pool, but it still has the value of the thread local variable
 that we specified using `threadLocal.asContextElement(value = "launch")`,
-no matter on what thread the coroutine is executed.
+no matter what thread the coroutine is executed on.
```
LouisCAD, line 628:
```suggestion
no matter which thread the coroutine is executed on.
```

Inego: Yes, thanks.

docs/exception-handling.md
Inego: It doesn't work, it says `JDK_16 environment variable is not defined.`

Inego: It's really, really sad that I would have to install JDK 1.6 just to fix some typos :(

LouisCAD: If the "Allow modifications from maintainers" checkbox is checked, maybe @qwwdfsad can do it?

elizarov:
Thanks. I've run knit and merged manually via f4fb204b220b224d5ffa9b7c00ff077b681213e9 Btw, for the purposes of knit you can simply set JDK_16 to JDK_HOME. I'll see if I can patch the code to make it happen automatically.

Inego:
Thank you, Roman!

Issue #1927, opened 2020-04-20
Add throttleLatest Flow operator


fluidsonic:
We have no operator that prevents subsequent emissions for a certain time after an emission.
In RxJava there is `throttleLatest` for that.

```kotlin
flow {
    repeat(10) {
        emit(it)
        delay(101)
    }
}
    .throttleLatest(500)
    .collect {
        println(it)
    }
```
The example above should print `0` and `5`. Both should be emitted without any added delay.

One **use-case** is button clicks where (accidental) repeated clicks should be filtered. If I click a button twice accidentally, only the first click event would make it downstream and the second click be ignored because it came in too fast after the first one.

`debounce(500)` is not suitable because it
a) delays the emission of the first value by least 500ms and
b) delays the emission of any value until the user stops clicking for at least 500ms.

`sample(500)` is not suitable because it
a) delays the emission of the first value by least 500ms and
b) emits nothing if the `Flow` is closed within 500ms after collection has started.

`throttleLatest` was also mentioned in #1107 but no reason was given why it didn't make it into `Flow` so far.

jxdabc:
@fluidsonic 
If I'm not wrong.
The above `throttleLatest` example should print `0`, '4'.
The operator you want seems being called `throttleFirst`.
Plese see https://github.com/ReactiveX/RxJava/wiki/Filtering-Observables#throttlefirst

fluidsonic:
@jxdabc you're right, `throttleFirst` is more appropriate for solving the problem. I'll update the issue.

davidmigloz:
Duplicate of https://github.com/Kotlin/kotlinx.coroutines/issues/1927

hoc081098:
please check my lib https://github.com/hoc081098/FlowExt https://hoc081098.github.io/FlowExt/docs/latest/-flow-ext/com.hoc081098.flowext/throttle-time.html

Issue #1446, opened 2019-08-18
throttleFirst Flow operator


fjna:
Is there a Flow equivalent for this Reactive stream operator? I guess it might be derived from the sample operator.

Is it possible to implement this operator using only the publicly exposed api? I tried to modify the sample operator implementation but since some parts of it are internal it isn't possible.

qwwdfsad:
>Is there a Flow equivalent for this Reactive stream operator?

No, but we have plans to implement its analogue (actually, the whole family of time-based sampling).

>Is it possible to implement this operator using only the publicly exposed api?

Modulo some cancellation details. For ad-hoc solution you can replace `scopedFlow` with `coroutineScope { flow { ... } }`

fjna:
I see, thanks. What are your plans regarding custom operator implementation?
I am still learning Flow API, but this makes me think that not every operator could be implemented because of internal API, is it?

elizarov:
Every operator should be functionally implementable via some public API. Internal API is only needed for highly-efficient (optimized) implementations. In particular, `throttleFirst` is relative easy  to implement. See here quite a straightforward implementation: https://pl.kotl.in/kKgI484X-

elizarov:
@fjna You'll need additional code to support `TestCoroutineScope`. Instead of using `System.currentTimeMillis()` you shall use the following snippet:

```
val time = (coroutineContext[ContinuationInterceptor] as? DelayController)?.currentTime ?: System.currentTimeMillis()
```

As for your implementation it looks fine apart from the code style (extra new later before `{`). It has different behavior which might be exactly what you wanted, though. 

> Btw, that's the key challenge in providing time-related operators out-of-the-box, since there are some many subtly different ways to implement them.

elizarov:
@zach-klippenstein Yes. Created a separate issue for it: #1499

pavlospt:
Is there any activity related to this issue going on 😄 ?

elizarov:
> Is there any activity related to this issue going on 😄 ?

Not yet.

lukas1:
@elizarov 
It looked a little bit difficult to prepare your code snipped for tests, or at least I didn't find any easy way, so I've tried to implement it a little bit differently. Maybe this can be useful for other folks too!

 I'm worried that maybe I have overlooked something, though. Do you think this implementation could work? Are there any inherent disadvantages to it? (such as leaking some scope or something similar).

```kotlin
[...]

PR #2128, opened 2020-07-08
Flow ThrottleFirst and ThrottleLatest operators

thibseisel:
This provides an implementation of  the `throttleFirst` Flow operator asked in #1446 and #1927.
This also adds an equivalent `throttleLatest` that conflates values until timeout is ellapsed.

The behavior of those operators is the same as their RxJava equivalents: [throttleFirst](http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Flowable.html#throttleFirst-long-java.util.concurrent.TimeUnit-), [throttleLatest](http://reactivex.io/RxJava/3.x/javadoc/io/reactivex/rxjava3/core/Flowable.html#throttleLatest-long-java.util.concurrent.TimeUnit-).

Both operators are used mostly for UI programming, for example:
- Reduce the pace at which a button is clicked,
- Reduce the rate a which display is updated (UI state, notifications, etc).

I feel like documentation of those operators and implementation of `throttleFirst` could be improved. Suggestions are more than welcome!

About the API, maybe we could rename `throttleLatest` to `throttleConflate` and `throttleFirst` to `throttle` or something like `throttleDrop`. I've found that `throttleLatest` works a bit like conflated channels, hence the name.

Also in RxJS, we have a single `throttle` operator that accepts 2 boolean parameters `leading` and `trailing`. We could follow the same approach by merging those 2 operators into a single `throttle(withTrailing: Boolean)` operator.

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/api/kotlinx-coroutines-jdk8.api
+++ b/integration/kotlinx-coroutines-jdk8/api/kotlinx-coroutines-jdk8.api
@@ -17,6 +17,7 @@ public final class kotlinx/coroutines/time/TimeKt {
 	public static final fun onTimeout (Lkotlinx/coroutines/selects/SelectBuilder;Ljava/time/Duration;Lkotlin/jvm/functions/Function1;)V
 	public static final fun sample (Lkotlinx/coroutines/flow/Flow;Ljava/time/Duration;)Lkotlinx/coroutines/flow/Flow;
 	public static final fun throttleFirst (Lkotlinx/coroutines/flow/Flow;Ljava/time/Duration;)Lkotlinx/coroutines/flow/Flow;
+	public static final fun throttleLatest (Lkotlinx/coroutines/flow/Flow;Ljava/time/Duration;)Lkotlinx/coroutines/flow/Flow;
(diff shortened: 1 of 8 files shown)


elizarov:
We've been postponing the implementation of these operators, because we have not yet designed a clean model of working with timers with Flows. Our goal is to minimize the set of built-in operators to avoid their explosive growth and rather focus on composable primitives. To this end, we planned to look if it is possible to design some small number of time-related core primitive to make the implementation of all sorts of throttling/sampling primitives trivial with them, as opposed to just adding all those variations that differ in hard-to-remember internal implementation details.

damianw:
@elizarov Is there a place/issue to follow along with that discussion?

elizarov:
> @elizarov Is there a place/issue to follow along with that discussion?

Not yet.

StanislavChumarin:
@elizarov Are there any updates on this issue or time-related core primitive?

qwwdfsad:
https://github.com/Kotlin/kotlinx.coroutines/pull/2745#pullrequestreview-681969186 here is the complete answer.

Unfortunately, we cannot provide timely update yet

PR #2145, opened 2020-07-17
Fix hanging RunInterruptibleStressTest on Windows with JDK 1.6

### LINKED ISSUE #2144: runInterruptibleStressTest hangs on JDK 1.6 on Windows
See https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStressWindows/3026232?buildTab=log&focusLine=3&linesState=41.34906

elizarov:
The test is improved so that it fails, not hangs, on a failure. However, it will not pass on Windows + JDK 1.6 due to a bug in JDK (which can be easily confirmed with this modification to a test), so it excluded when running on JDK 1.6.

Fixes #2144


elizarov:
With the changes in the test if you actually run it on Windows + JDK1.6 it fails with the following message:
```
java.lang.IllegalStateException: Sleep was not interrupted, Thread.isInterrupted=true
```

Issue #1216, opened 2019-05-22
Add debounce operator with selector to Flow


Thomas-Vos:
Could an operator like the following be added to Flow? The debounce operator is already added, but I need one with a different timeout depending on the item. It shouldn’t be too difficult to modify the current operator to add this.
```kotlin
fun <T> Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T>
```

My use case is the following:
when the app is opened, it connects to multiple devices simultaneously. This normally takes about a second. It shows `connecting...` and finally `connected`. What I want is that the "connecting" text is only shown if it takes more than one second. For example if the device is connected in 0.5 seconds, the "connecting" message must not be shown (only the "connected" message).

I need the timeout for "connecting, because otherwise the "connecting" message will be shown very shortly if the device is connected quickly.

Here's an example:
```kotlin
enum class ConnectionState {
    DISCONNECTED,
    CONNECTING,
    CONNECTED
}

fun appVisible() {
    // Multiple devices are being connected simultaneously. Every time one of the
    // connection states changes, a new list is sent to this flow.
    val statesFlow: Flow<List<ConnectionState>> = observeConnectionStates()
    
    statesFlow
        .debounce { states ->
            if (states.any { it == ConnectionState.CONNECTING }) {
                1000L // still connecting, delay by one second if nothing changes.
            } else {
                0L // all devices are connected, immediately show the message
            }
        }
        .collect { TODO("update message") }
}
```

Previously I was using [this RxJava operator](https://github.com/ReactiveX/RxJava/blob/8f51d2df7dfa4cfbf735a68085e6673aaa4a2856/src/main/java/io/reactivex/Observable.java#L7294-L7325) (line 7294-7325), but I am migrating my code to coroutines for multiplatform support.

(related discussion on Slack: https://kotlinlang.slack.com/archives/C1CFAFJSK/p1558109098082700)

Thomas-Vos:
I'm currently using the following implementation. It is a modified version of the already implemented debounce operator:
```kotlin
fun <T> Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T> = flow {
    coroutineScope {
        val values = Channel<Any?>(Channel.CONFLATED) // Actually Any, KT-30796
        // Channel is not closed deliberately as there is no close with value
        val collector = launch {
            try {
                collect { value -> values.send(value ?: NullSurrogate) }
            } catch (e: Throwable) {
[...]

PR #2148, opened 2020-07-19
Add debounce selector

mkano9:

Resolves https://github.com/Kotlin/kotlinx.coroutines/issues/1216.


kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -99,6 +105,39 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+                        lastValue = DONE
+                    } else {
+                        lastValue = it
+                    }
+                }
+
+                lastValue?.let { value ->
+                    val unboxedValue: T = NULL.unbox(value)
+                    val timeoutMillis = timeoutMillisSelector(unboxedValue)
+                    require(timeoutMillis > 0) { "Debounce timeout should be positive" }
```
Ezfero, line 130: Can we make 0 valid value as well? In case of using a selector and having different behaviour to different items in the steam, it is quite possible that some of the items would require to have no delay at all. Then we can move this line to `debounce(timeoutMillis: Long)` function to make sure that current behaviour won't change.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -37,39 +37,50 @@ import kotlin.time.*
...
+ * ```
+ * produces `3, 4, 6`.
+ *
+ * @param timeoutMillisSelector [T] is the emitted value and the return value is timeout in milliseconds. 0ms timeout is allowed.
+ */
+@FlowPreview
+public fun <T> Flow<T>.debounce(timeoutMillisSelector: (T) -> Long): Flow<T> =
+    debounceInternal { emittedItem ->
+        val timeoutMillis = timeoutMillisSelector(emittedItem)
+        require(timeoutMillis >= 0) { "Debounce timeout should not be negative" }
```
fvasco, line 81: Historically this library considers negative timeouts like a `0` (see `onTimeout`)

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -37,39 +37,50 @@ import kotlin.time.*
  *
  * Note that the resulting flow does not emit anything as long as the original flow emits
  * items faster than every [timeoutMillis] milliseconds.
+ * @param timeoutMillis must be positive
  */
 @FlowPreview
 public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
     require(timeoutMillis > 0) { "Debounce timeout should be positive" }
```
fvasco, line 44: Why `delay(0)` is good while `debounce(0)` isn't.
Does timeout "should be positive" or "must be positive"?

Please consider:

    if (timeoutMillis <= 0) return this

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
mkano9: Cool, I should've checked what `onTimeout` does internally. I just had blindly kept the original check in place. I will update.

kotlinx-coroutines-core/common/test/flow/operators/DebounceTest.kt
```diff
@@ -223,4 +221,68 @@ class DebounceTest : TestBase() {
...
+        finish(4)
+    }
+
+    @Test
+    fun testZeroDebounceTimeSelector() = withVirtualTime {
+        expect(1)
+        val flow = flow {
+            expect(2)
+            emit("A")
+            delay(1)
```
mkano9, line 278: When I emit two values immediately one after another, the first one never reaches the `select<Unit> {` block. I see the same behavior prior to my changes though. 

It looks like it's getting conflated, but I don't understand where exactly. I tried modifying `val values = produce<Any?>(capacity = Channel.CONFLATED) {` to use different capacity just to troubleshoot, but no luck.

Any suggestions are appreciated.

mkano9:
Hi @fvasco, any updates?

fvasco:
Hi, @mkano9,
I haven't updates for you, I'm sorry.

If you think there is an issue with the current implementation, please consider to write a consistent test case and file a new issue.

akhbulatov:
+1 to this feature. Any progress?

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -40,36 +40,44 @@ import kotlin.time.*
...
-                // Should be receiveOrClosed when boxing issues are fixed
-                values.onReceiveOrNull {
-                    if (it == null) {
-                        if (lastValue != null) downstream.emit(NULL.unbox(lastValue))
-                        lastValue = DONE
-                    } else {
-                        lastValue = it
-                    }
-                }
+    if (timeoutMillis <= 0) return this
```
elizarov, line 43: Can you please elaborate on the reasoning behind the change from `require(timeoutMillis > 0)` to `if (timeoutMillis <= 0) return this`? Why would you need to silently ignore zero or negative values here?

kotlinx-coroutines-core/common/test/flow/operators/DebounceTest.kt
elizarov: Changing from `produce<Any?>(capacity = Channel.CONFLATED)` to `produce<Any?>(capacity = 0)` should remove the conflation. The reason that `CONFLATED` is used here, though, is because the `debounce` is effectively a "stronger conflation" operator that conflates even more values.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
mkano9: I interpreted 0 `timeoutMillis` as no debounce and emit right away. As a user of this API, if you passed a 0 for whatever reason, I think I would expect it to just not debounce.

Also, in this variation `debounce(timeoutMillisSelector: (T) -> Long)`, the intention was allow 0ms to skip a debounce in certain conditions. It would match the behavior between the two variations.

There is a relevant comment in [here](https://github.com/Kotlin/kotlinx.coroutines/pull/2148#discussion_r457106584).

kotlinx-coroutines-core/common/test/flow/operators/DebounceTest.kt
mkano9: It still fails the `testZeroDebounceTimeSelector` test with `kotlin.AssertionError: Expected <[A, B]>, actual <[B]>` if I remove that `delay(1)` from my test code.

elizarov:
We've discussed the design aspects of this change in the team. Green light to the addition itself, but here are some minor adjustments that are needed:
* The function with selector shall throw an IllegalArgumentException if a negative timeout is returned by a selector at any time.
* The support for zero timeouts shall be added and it should work without conflation. A non-conflated channel should be used internally in all cases and that is fine. 
* The above changes to timeout behavior shall be consistent with the version without a selector, too (it shall start support zero timeouts returning `this`, but throw on negative ones).
* There needs to be an `@ExpertimentalTime` version that is overloaded for `(T) -> Duration`  selector, too.

mkano9:
Hi @elizarov, thanks for the feedback. I updated the PR.

kotlinx-coroutines-core/common/test/flow/operators/DebounceTest.kt
mkano9: I don't understand why `testZeroDebounceTimeSelector` fails without that `delay(1)`. Any theories?

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.time.*
 
 /**
  * Returns a flow that mirrors the original flow, but filters out values
- * that are followed by the newer values within the given [timeout][timeoutMillis].
+ * that are followed by the newer values within the given [timeoutMillis].
```
elizarov, line 19: Please, leave this part of the docs as before (it is consistent with how `sample` is documented, too)

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -40,36 +40,46 @@ import kotlin.time.*
...
+ *     } else {
+ *         300L
+ *     }
+ * }
+ * ```
+ * produces `3, 4, 6`.
+ *
+ * @param timeoutMillisSelector [T] is the emitted value and the return value is timeout in milliseconds.
+ */
+@ExperimentalTime
```
elizarov, line 77: This one should not have `@ExperimentalTime` annotation, but there should be another version with `(T) - Duration` selector that has one.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -99,6 +109,39 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+                        lastValue = it
+                    }
+                }
+
+                lastValue?.let { value ->
+                    val unboxedValue: T = NULL.unbox(value)
+                    val timeoutMillis = timeoutMillisSelector(unboxedValue)
+                    require(timeoutMillis >= 0L) { "Debounce timeout should not be negative" }
+                    // set timeout when lastValue != null
+                    onTimeout(timeoutMillis) {
```
elizarov, line 136: A case with `timeoutMillis == 0` should be explicitly optimized here -- emit it immediately instead of installing `onTimeout` case.

elizarov (changes requested):

mkano9:
I accidentally merged with master and made further mess. I opened a new PR https://github.com/Kotlin/kotlinx.coroutines/pull/2314 and closing this one.

Issue #2157, opened 2020-07-21
Support for Mingw86


danielgruber:
We're trying to build a 32 bit version for our Kotlin Native Project and would love to support 32 bit. Is this possible right now?

qwwdfsad:
Unfortunately, it is not really possible.
We neither have bare metal to actually _test_ mingw binaries, nor enough time to take care of this issue

JohnLCaron:
I see that kotlinx-coroutines-core has a mingwx64 target but not a mingwx86. Is there something about mingwx86 that makes it harder than mingwx64?

qwwdfsad:
[mingwX86 is deprecated](https://kotlinlang.org/docs/native-target-support.html#deprecated-targets) thus we are not adding support for that target

PR #2158, opened 2020-07-21
- add 32bit

danielgruber:
Add 32 bit with #2157

### DIFF
--- a/gradle/compile-native-multiplatform.gradle
+++ b/gradle/compile-native-multiplatform.gradle
@@ -22,7 +22,7 @@ kotlin {
             addTarget(presets.iosX64)
             addTarget(presets.macosX64)
             addTarget(presets.mingwX64)
-            addTarget(presets.mingwX84)
+            addTarget(presets.mingwX86)
             addTarget(presets.tvosArm64)
             addTarget(presets.tvosX64)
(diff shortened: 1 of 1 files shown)


elizarov:
Does it actually work? Would the tests run on a 64bit version of windows?

elizarov:
It does not build at our CI: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildWindows/3042932

danielgruber:
weird. Would you help me in one thing: Where do the presets come from? So how do I need to define a new Win32 preset? Though this comes from Multiplatform Library core, but seems like it doesn't as it does not find the minwx86 preset although it should support it? :)

danielgruber:
I fixed the Typo, but I think it's still depending on AtomicFus Pull Request first: https://github.com/Kotlin/kotlinx.atomicfu/pull/146

qwwdfsad (commented):
Alas, we are not _yet_ ready to enable new Native targets.
We are working on the list with full Kotlin/Native-wide list of various targets support (debugging support/unit testing support/whether we test our libraries on a specific target and so on) and as soon as it's ready, we'll enable _all_ existing targets, referencing a giving guarantees

PR #2164, opened 2020-07-25
feature: Add setup

AdamSHurwitz:
• Gradle implementation 
• Link to latest library release/version

### DIFF
--- a/integration/kotlinx-coroutines-play-services/README.md
+++ b/integration/kotlinx-coroutines-play-services/README.md
@@ -9,6 +9,15 @@ Extension functions:
 | [Task.await][await] | Awaits for completion of the Task (cancellable)
 | [Deferred.asTask][asTask] | Converts a deferred value to a Task
 
+## Setup
+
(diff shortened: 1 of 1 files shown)


elizarov:
Sorry, but this is not consistent with any of the other integration modules we have. We consistently have setup instructions only in the top-level README.md.

AdamSHurwitz:
That makes sense. Would the Play Services setup instructions be a better fit under the top-level README.md's _Using in your projects_ section as an H3?

qwwdfsad:
@AdamSHurwitz thanks for the input!
Unfortunately, we cannot add an instruction on how to include each particular module into dependencies as it will bloat all our documentation for no good reason. If one knows how to include one jar into dependencies, the other ones shouldn't be a big deal.

We will think about how to improve the situation here and probably will add a link with the list of all modules

AdamSHurwitz:
> A link with the list of all modules

This is a great idea! Thank you for the feedback.

PR #2165, opened 2020-07-27
Added a `mapIterable` operator

scrobot:
I suppose to add the new operator called `mapIterable`.  It's can flatten flow stream and give the ability to work with each entity of collection. I will be glad to hear any feedback. I'm ready to complete work on this operator by you code conducts. Best regards.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
@@ -49,6 +49,20 @@ public inline fun <T, R> Flow<T>.map(crossinline transform: suspend (value: T) -
 }
 
 /**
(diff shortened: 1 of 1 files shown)


elizarov:
Why it is needed? What problem does it solve that is not solved by existing operators?

zach-klippenstein:
Couldn't this also be done with: 
```
  .map { it.asFlow() }
  .flatMap { … }
```

elizarov:
I am a bit lost with this PR. First of all. The doc says that is takes `List<Entity>` and then you can do `mapIterable` on it. But the actual extension is defined on `Flow`. Let's move it to issues. Please, create a separate issue on the problem you're trying to solve, and I'll close this PR for now.

PR #2178, opened 2020-08-01
Handle undeliverable errors in rxObservable

### LINKED ISSUE #2173: Inconsistent handling of undeliverable errors in RX builders
While looking at the implementations of the RX coroutines builders in the `kotlinx-coroutines-rx*` modules, I noticed some inconsistencies in the handling of undeliverable errors between the `rxObservable` implementation on the one hand, and the `rxSingle`, `rxMaybe`, and `rxCompletable` implementations on the other.

In the `rxSingle`, `rxMaybe`, and `rxCompletable` implementations, when the coroutine is cancelled with an exception, it first tries to deliver it to the observer, and if that's not possible then it is reported to the RX global exception handler. This behaviour is the same regardless of whether the error is _fatal_ or not.

Whereas in the `rxObservable` implementation, it drops undeliverable non-fatal errors, but _always_ reports fatal errors to the RX global exception handler, regardless of whether the observable could handle it or not.

The `rxFlowable`/`publish` implementation seems to be a combination of both, it reports all undeliverable _or_ fatal exceptions to the global exception handler. This seems to be the best and safest approach.

1zaman:
If an error can't be delivered to the observer due to the observer already having disposed the subscription, then it should be reported to the RX global exception handler. This was already being done in the implementations for `rxSingle`, `rxMaybe`, and `rxCompletable`, but not in `rxObservable`.

Fixes one inconsistency from #2173. I'll look into fixing the other as well.

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
@@ -164,12 +164,10 @@ private class RxObservableCoroutine<T: Any>(
                          * Such behaviour is inconsistent, leads to silent failures and we can't possibly know whether
                          * the cause will be handled by onError (and moreover, it depends on whether a fatal exception was
                          * thrown by subscriber or upstream).
-                         * To make behaviour consistent and least surprising, we always handle fatal exceptions
-                         * by coroutines machinery, anyway, they should not be present in regular program flow,
-                         * thus our goal here is just to expose it as soon as possible.
+                         * To make behaviour consistent and least surprising, we always deliver fatal exceptions to the
+                         * RX global exception handler.
                          */
-                        subscriber.tryOnError(cause)
-                        if (!handled && cause.isFatal()) {
+                        if (!subscriber.tryOnError(cause) || (!handled && cause.isFatal())) {
                             handleUndeliverableException(cause, context)
                         }
                     }
--- a/reactive/kotlinx-coroutines-rx2/test/ObservableTest.kt
+++ b/reactive/kotlinx-coroutines-rx2/test/ObservableTest.kt
@@ -4,14 +4,10 @@
 
 package kotlinx.coroutines.rx2
 
-import io.reactivex.*
-import io.reactivex.exceptions.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.CancellationException
 import org.junit.*
 import org.junit.Test
-import java.util.concurrent.*
-import java.util.concurrent.atomic.*
 import kotlin.test.*
 
 class ObservableTest : TestBase() {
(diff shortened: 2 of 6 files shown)


reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
```diff
@@ -164,12 +164,10 @@ private class RxObservableCoroutine<T: Any>(
                          * Such behaviour is inconsistent, leads to silent failures and we can't possibly know whether
                          * the cause will be handled by onError (and moreover, it depends on whether a fatal exception was
                          * thrown by subscriber or upstream).
-                         * To make behaviour consistent and least surprising, we always handle fatal exceptions
-                         * by coroutines machinery, anyway, they should not be present in regular program flow,
-                         * thus our goal here is just to expose it as soon as possible.
+                         * To make behaviour consistent and least surprising, we always deliver fatal exceptions to the
+                         * RX global exception handler.
```
1zaman, line 168: This was changed in #1638, but seems like the comment was not updated.

reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
```diff
@@ -164,12 +164,10 @@ private class RxObservableCoroutine<T: Any>(
...
                          */
-                        subscriber.tryOnError(cause)
-                        if (!handled && cause.isFatal()) {
+                        if (!subscriber.tryOnError(cause) || (!handled && cause.isFatal())) {
```
1zaman, line 170: It seems that the `handled` flag is redundant here after the earlier check for `CancellationException`, since it could only be handled in the current implementation if it was a `CancellationException`. There doesn't seem to be any special handling of fatal exceptions by the coroutine. However, I have left that logic as it was here.

reactive/kotlinx-coroutines-rx2/test/ObservableTest.kt
```diff
@@ -137,28 +134,4 @@ class ObservableTest : TestBase() {
...
-                .interval(1, TimeUnit.MILLISECONDS)
-                .take(1000)
-                .switchMapSingle {
-                    rxSingle {
-                        timeBomb().await()
-                    }
-                }
-                .blockingSubscribe({}, {})
-        }
-    }
```
1zaman, line 161: This test seems to have been taken from the issue reported at #252. However it's complicated and not well implemented (I originally tried to improve it in commit 58ad0cdfca81ab2d5c5c410f889293c51beede75). Also, it was actually testing `rxSingle` instead of `rxObservable`, which is why it didn't catch the issue in the first place that this pull request is trying to solve. `rxSingle` (and the other RX builders) already have a simpler and deterministic test for undeliverable error handling (`testUnhandledException()`), which I have copied over in the `ObservableExceptionHandlingTest` test suite, so I deleted this test since should be no further need for it.

qwwdfsad:
Fixed it in 1.5.0, thanks for the report

PR #2187, opened 2020-08-08
Fix sporadic ConcurrentWeakMapTest failures

qwwdfsad:
Root cause: https://youtrack.jetbrains.com/issue/KT-40943


elizarov (approved):
Great catch!

PR #2201, opened 2020-08-16
Fix some sample outputs not being displayed in flow.md

MasoodFallahpoor:
For some reason, some sample outputs in section `Flow Exceptions` are not displayed in the documentation. This commit fixes this issue.

### DIFF
--- a/docs/flow.md
+++ b/docs/flow.md
@@ -1471,6 +1471,8 @@ Exception in thread "main" java.lang.IllegalStateException: Collected 2
 	at ...
 ```
 
+<!--- TEST EXCEPTION -->
+
 #### Catching declaratively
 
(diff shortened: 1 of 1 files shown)


elizarov:
Thanks for noticing it. Indeed, the prose indicates that these should an example, but it is mistakenly hidden inside the Knit directive. However, it is not enough to replace it with a markdown code block. You should add the corresponding single-line `<!--- TEST xxx -->` directive directly after the markdown text block to make sure the test case is still being generated (checkout how it is done in other examples).

MasoodFallahpoor:
@elizarov Based on other examples, I've added the missing directives.

elizarov:
The second directive was also `TEST EXCEPTION`. It does not pass the test anymore.

Issue #1442, opened 2019-08-14
ListenableFuture future() cancellation


yorickhenning:
[Issue 1346](https://github.com/Kotlin/kotlinx.coroutines/issues/1346) describes cancellation issues with ListenableFuture.kt.

The PR that should close that issue fixes up `asDeferred()` and `asListenableFuture()`, but _doesn't_ fix `future()`.

`future()` also needs modifying to implement bidirectional cancellation, and some tests validating it.

I'll TAL.

elizarov:
Ping?

yorickhenning:
I haven't had the time to work on this since I filed it. If someone's interested I can hand this off. Is there urgency from anywhere?

PR #2222, opened 2020-09-01
Repair some corner cases in cancellation propagation between coroutines and listenable futures 

vadimsemenov:
Implement bidirectional cancellation for `future` coroutine builder.

This also:

* Refactors JobListenableFuture infrastructure so it can be reused in CoroutineScope.future and Deferred.asListenableFuture;
* Provides more descriptive `toString` implementation for the returned Future;
* Fixes stack traces in thrown exception, so it includes call to get() that triggered the exception to be thrown;
* Hides ListenableFuture.asDeferred return type, so it can't be casted to CompletableDeferred;
* Adds more tests to cover fixed corner cases;
* Improves documentation;
* Suppresses annoying warnings in tests.

### DIFF
--- a/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
+++ b/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
@@ -167,7 +167,7 @@ public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
  * state - a serious fundamental bug.
  */
 private fun ExecutionException.nonNullCause(): Throwable {
-  return this.cause!!
+    return this.cause!!
 }
 
 /**
@@ -479,4 +479,4 @@ private class JobListenableFuture<T>(private val jobToCancel: Job): ListenableFu
  * [SettableFuture] can't store the reason of cancellation. To mitigate this, we wrap cancellation exception into this
  * class and pass it into [SettableFuture.complete]. See implementation of [JobListenableFuture].
  */
-private class Cancelled(val exception: CancellationException)
+private class Cancelled(@JvmField val exception: CancellationException)
--- a/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
+++ b/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
@@ -207,7 +207,7 @@ public fun <T> Deferred<T>.asListenableFuture(): ListenableFuture<T> {
         if (throwable == null) {
             listenableFuture.complete(getCompleted())
         } else {
-            listenableFuture.completeExceptionally(throwable)
+            listenableFuture.completeExceptionallyOrCancel(throwable)
         }
     }
     return listenableFuture
@@ -302,14 +302,14 @@ private class ListenableFutureCoroutine<T>(
 ) : AbstractCoroutine<T>(context) {
 
     // JobListenableFuture propagates external cancellation to `this` coroutine. See JobListenableFuture.
-    val future = JobListenableFuture<T>(this)
+    @JvmField val future = JobListenableFuture<T>(this)
 
     override fun onCompleted(value: T) {
         future.complete(value)
     }
 
     override fun onCancelled(cause: Throwable, handled: Boolean) {
-        if (!future.completeExceptionally(cause) && !handled) {
+        if (!future.completeExceptionallyOrCancel(cause) && !handled) {
             // prevents loss of exception that was not handled by parent & could not be set to JobListenableFuture
             handleCoroutineException(context, cause)
         }
@@ -342,8 +342,10 @@ private class JobListenableFuture<T>(private val jobToCancel: Job): ListenableFu
      * [AbstractFuture] has a highly-correct atomic implementation of `Future`'s completion and
      * cancellation semantics. By using that type, the [JobListenableFuture] can delegate its semantics to
      * `auxFuture.get()` the result in such a way that the `Deferred` is always complete when returned.
+     *
+     * To preserve Coroutine's [CancellationException], this future points to either `T` or [Cancelled].
      */
-    private val auxFuture = SettableFuture.create<Result<T>>()
+    private val auxFuture = SettableFuture.create<Any>()
 
     /**
      * When the attached coroutine [isCompleted][Job.isCompleted] successfully
@@ -351,7 +353,7 @@ private class JobListenableFuture<T>(private val jobToCancel: Job): ListenableFu
      *
      * This should succeed barring a race with external cancellation.
      */
-    fun complete(result: T) = auxFuture.set(Result.Success(result))
+    fun complete(result: T): Boolean = auxFuture.set(result)
 
     /**
      * When the attached coroutine [isCompleted][Job.isCompleted] [exceptionally][Job.isCancelled]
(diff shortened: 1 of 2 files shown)


vadimsemenov:
Fixes #1442

vadimsemenov:
Ping?

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -271,57 +278,46 @@ private class ToContinuation<T>(
...
-     * the [Deferred] and [ListenableFuture] types, while permitting both kinds of promise to point
-     * to the same running task.
-     */
-    override fun run() {
-        if (future.isCancelled) {
-            cancel()
-        }
-    }
+    // JobListenableFuture propagates external cancellation to `this` coroutine. See JobListenableFuture.
+    val future = JobListenableFuture<T>(this)
```
qwwdfsad, line 305: `@JvmField` will save an accessor here (which is redundant)

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -271,57 +278,46 @@ private class ToContinuation<T>(
...
     override fun onCompleted(value: T) {
-        future.set(value)
+        future.complete(value)
     }
 
-    // TODO: This doesn't actually cancel the Future. There doesn't seem to be bidi cancellation?
     override fun onCancelled(cause: Throwable, handled: Boolean) {
-        if (!future.setException(cause) && !handled) {
-            // prevents loss of exception that was not handled by parent & could not be set to SettableFuture
+        if (!future.completeExceptionally(cause) && !handled) {
```
qwwdfsad, line 312: It seems like (a long-existing) problem here: if `future {}` is cancelled **normally** (e.g. its parent got cancelled) and meanwhile someone completes the future from the outside, then `JobCancellationException` got reported into an exception handler, while we guarantee that no `CancellationException` will ever reach it.

This problem seems to exist in most of our `AbstractCoroutine` implementors, unfortunately. It would be nice if you could write a test that reproduces the problem and a fix for that (see: `MonoCoroutine.onCancelled`), I'll take care of the rest of the builders

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -333,130 +329,161 @@ private class ListenableFutureCoroutine<T>(
...
+ *   - To maintain sanity, this class implements [ListenableFuture] and uses an auxiliary [SettableFuture]
+ *     around coroutine's result as a state engine to establish happens-after-completion. This
+ *     could probably be compressed into one subclass of [AbstractFuture] to save an allocation, at the
  *     cost of the implementation's readability.
  */
-private class OuterFuture<T>(private val deferred: Deferred<T>): ListenableFuture<T> {
-    val innerFuture = DeferredListenableFuture(deferred)
+private class JobListenableFuture<T>(private val jobToCancel: Job): ListenableFuture<T> {
+    /**
+     * Serves as a state machine for [Future] cancellation.
```
qwwdfsad, line 340: Great job with auxulary comments 👍

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -333,130 +329,161 @@ private class ListenableFutureCoroutine<T>(
...
+     * When the attached coroutine [isCompleted][Job.isCompleted] [exceptionally][Job.isCancelled]
+     * its outcome should be passed to this method.
+     *
+     * This method will map coroutine's exception into corresponding Future's exception.
+     *
+     * This should succeed barring a race with external cancellation.
+     */
+    // CancellationException is wrapped into `Result.Cancellation` to preserve original cause and message.
+    // All the other exceptions are delegated to SettableFuture.setException.
+    fun completeExceptionally(t: Throwable) =
```
qwwdfsad, line 366: Please declare an explicit return type here for the sake of readability

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -333,130 +329,161 @@ private class ListenableFutureCoroutine<T>(
...
+ *
+ * - When coroutine [isCompleted][Job.isCompleted] successfully, its result is wrapped into [Success].
+ * - When coroutine [isCompleted][Job.isCompleted] exceptionally, meaning [isCancelled]
+ *   returns `true`. According to [Future]'s contract, the exception can be of two types:
+ *     - [CancellationException] if the coroutine was _cancelled normally_, meaning it threw [CancellationException].
+ *       In this case the exception is wrapped into [Result.Cancellation].
+ *     - [ExecutionException] with the original [Throwable] in its [cause][ExecutionException.cause] otherwise.
+ *       This case is handled separately. See [JobListenableFuture.completeExceptionally].
+ */
+private sealed class Result<out T> {
```
qwwdfsad, line 481: Thanks for the explanation!
Would suggest getting rid of the `Result` clas tho: `Result.success` means a different thing in Kotlin (name clash with a popular stdlib class is not a great thing to have) and we also have to allocate on successful completion.

We can use raw value for the successful results and a `Cancelled(exception)` wrapper for cancellation exception. `isCancelled` then can be replaced with `val Any.is[Future]Cancelled() get() = this is Cancelled`

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -195,13 +196,21 @@ private fun ExecutionException.nonNullCause(): Throwable {
...
-    outerFuture.afterInit()
-    return outerFuture
+    val listenableFuture = JobListenableFuture<T>(this)
+    // This invokeOnCompletion completes the JobListenableFuture with the same result as `this` Deferred.
+    // The JobListenableFuture may have completed earlier if it got cancelled! See JobListenableFuture.cancel().
+    invokeOnCompletion { throwable ->
+        if (throwable == null) {
+            listenableFuture.complete(getCompleted())
+        } else {
+            listenableFuture.completeExceptionally(throwable)
```
qwwdfsad, line 210: It seems like a bug if you don't know what's inside `completeExceptionally`. Please rename it to `completeExceptionallyOrCancel` or anything else that declares the intent

qwwdfsad (changes requested):
Great job! A decent number of comments, every change and corner case is tested, really valuable contribution!

Let's polish it a bit and it's ready to go.

vadimsemenov (commented):
Thanks for the review. I added another commit. I'll squash it into the previous one if it looks good to you.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
qwwdfsad: You are indeed right, I've checked the wrong source

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -152,7 +152,8 @@ public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
     deferred.invokeOnCompletion {
         cancel(false)
     }
-    return deferred
+    // Return hides the CompletableDeferred. This should prevent casting.
+    return object : Deferred<T> by deferred {}
```
qwwdfsad, line 156: While we are here, please fix indent on line 170 (should be 4 spaces instead of 2)

qwwdfsad (approved):
Great! Please fix a minor comment and it's good to go.

I'm ready to squash & merge it and make a 1.4.2 release as soon as it's ready

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -333,130 +329,154 @@ private class ListenableFutureCoroutine<T>(
...
 }
+
+/**
+ * A wrapper for `Coroutine`'s [CancellationException].
+ *
+ * If the coroutine is _cancelled normally_, we want to show the reason of cancellation to the user. Unfortunately,
+ * [SettableFuture] can't store the reason of cancellation. To mitigate this, we wrap cancellation exception into this
+ * class and pass it into [SettableFuture.complete]. See implementation of [JobListenableFuture].
+ */
+private class Cancelled(val exception: CancellationException)
```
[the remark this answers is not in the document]
vadimsemenov: Oops, the same mistake again, sorry.

vadimsemenov:
Done.
I also filled #2409 to address the issue with two ignored tests when `AbstractFuture.afterDone` is finalised.

qwwdfsad (approved):

Issue #2224, opened 2020-09-02
Stop requiring JDK 6 to run tests


LouisCAD:
Hello, I'm working on #2213 and I get this on a machine that runs macOS Catalina (10.15.6), where it's not possible to install JDK 6 as far as I know:

```
> Task :kotlinx-coroutines-core:cleanJvmTest UP-TO-DATE
> Task :kotlinx-coroutines-core:checkJdk16 FAILED
FAILURE: Build failed with an exception.
* Where:
Build file '/Users/louiscad/Documents/GitHub/LouisCAD/kotlinx.coroutines/kotlinx-coroutines-core/build.gradle' line: 81
* What went wrong:
Execution failed for task ':kotlinx-coroutines-core:checkJdk16'.
> JDK_16 environment variable is not defined. Can't build against JDK 1.6 runtime and run JDK 1.6 compatibility tests. Please ensure JDK 1.6 is installed and that JDK_16 points to it.
```

Can you please drop this requirement? I think JDK 7 can do. Even better would be to have instructions depending on our OS in the error message to get setup without having to search on Google and focus back on whatever we were doing.

elizarov:
We don't require it anymore. Contribution docs were updated to more explicitly state it.

LouisCAD:
Yep, I noticed, thank you for your action!

PR #2225, opened 2020-09-02
Add awaitCancellation()

### LINKED ISSUE #2213: [Feature request] Add awaitCancellation()
I can submit a PR.

### Use case examples

With `try`/`finally`:
```kotlin
suspend fun showStuffUntilCancelled(data: Stuff): Nothing {
    try {
        someSubView.text = data.title
        anotherSubView.text = data.description
        someView.isVisible = true
        awaitCancellation() // Suspends so the view stays visible
    } finally {
        someView.isVisible = false // Called on cancellation, cannot forget to hide the view!
    }
}
```

With an inline higher order function:
```kotlin
suspend fun showStuffUntilCancelled(data: Stuff): Nothing {
    someSubView.text = data.title
    anotherSubView.text = data.description
    someView.visibleInScope {
        awaitCancellation() // Suspends so the view stays visible
    }
}
```

### Implementation

```kotlin
import kotlinx.coroutines.CancellationException
import kotlinx.coroutines.suspendCancellableCoroutine

/**
 * Suspends until cancellation, in which case it will throw a [CancellationException].
 *
 * Handy because it returns [Nothing], allowing it to be used in any coroutine,
 * regardless of the required return type.
 */
suspend inline fun awaitCancellation(): Nothing = suspendCancellableCoroutine {}
```

LouisCAD:
I'm unsure about the relevance of the test and I couldn't run it on the JVM locally (because of #2224), but it succeeded in native (macosX64).

Feel free to make any suggestion regarding the files where the new code is, or suggest/perform edits to the doc.

Resolves #2213

### DIFF
--- a/kotlinx-coroutines-core/common/src/Await.kt
+++ b/kotlinx-coroutines-core/common/src/Await.kt
@@ -8,14 +8,6 @@ import kotlinx.atomicfu.*
 import kotlin.coroutines.*
 
 /**
- * Suspends until cancellation, in which case it will throw a [CancellationException].
- *
- * Handy because it returns [Nothing], allowing it to be used in any coroutine,
- * regardless of the required return type.
- */
-public suspend inline fun awaitCancellation(): Nothing = suspendCancellableCoroutine {}
-
-/**
  * Awaits for completion of given deferred values without blocking a thread and resumes normally with the list of values
  * when all deferred computations are complete or resumes with the first thrown exception if any of computations
  * complete exceptionally including cancellation.
--- a/kotlinx-coroutines-core/common/src/Await.kt
+++ b/kotlinx-coroutines-core/common/src/Await.kt
@@ -8,6 +8,14 @@ import kotlinx.atomicfu.*
 import kotlin.coroutines.*
 
 /**
+ * Suspends until cancellation, in which case it will throw a [CancellationException].
+ *
+ * Handy because it returns [Nothing], allowing it to be used in any coroutine,
+ * regardless of the required return type.
+ */
+public suspend inline fun awaitCancellation(): Nothing = suspendCancellableCoroutine {}
+
+/**
  * Awaits for completion of given deferred values without blocking a thread and resumes normally with the list of values
  * when all deferred computations are complete or resumes with the first thrown exception if any of computations
  * complete exceptionally including cancellation.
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/src/Await.kt
```diff
@@ -7,6 +7,14 @@ package kotlinx.coroutines
 import kotlinx.atomicfu.*
 import kotlin.coroutines.*
 
+/**
+ * Suspends until cancellation, in which case it will throw a [CancellationException].
+ *
+ * Handy because it returns [Nothing], allowing it to be used in any coroutine,
+ * regardless of the required return type.
+ */
+public suspend inline fun awaitCancellation(): Nothing = suspendCancellableCoroutine {}
```
elizarov, line 16: It seems more efficient in terms of the resulting bytecode to make it not-inline. Note, that you'll need to update API dump after that. See https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#updating-the-public-api-dump

Also, it is very hard to pick the "right" file to place this declaration in, but I think it might be better to add to `Delay.kt` and actually mention it in the docs of the `delay` function. Since the current workaround is to write `delay(Long.MAX_VALUE)`, it could say that this is equivalent to `awaitCancellation()`.

kotlinx-coroutines-core/common/src/Await.kt
```diff
@@ -7,6 +7,14 @@ package kotlinx.coroutines
 import kotlinx.atomicfu.*
 import kotlin.coroutines.*
 
+/**
+ * Suspends until cancellation, in which case it will throw a [CancellationException].
+ *
+ * Handy because it returns [Nothing], allowing it to be used in any coroutine,
```
elizarov, line 13: Actually, it would be great to show some small practical example here, if you have one in mind.

elizarov (changes requested):

kotlinx-coroutines-core/common/src/Await.kt
LouisCAD: Added two examples. I failed to find simpler examples for now, but it's not public API so it can be changed in the future.

Also, I know `trySend` doesn't exist yet, but I thought it would look nicer than `runCatching { offer(…) }` or a try/catch.

kotlinx-coroutines-core/common/src/Await.kt
LouisCAD: Done. BTW, thanks for updating the contributing guide.

elizarov (approved):
Good! I'm merge it manually with few changes.

elizarov:
Merged.

PR #2230, opened 2020-09-04
Allow nullable types in Flow's firstOrNull extension

### LINKED ISSUE #2229: Flow.firstOrNull should allow nullable types
`Flow.firstOrNull` does not allow nullable types, presumably so that you can differentiate between the flow being empty or not. However, every other `firstOrNull` function such as `List.firstOrNull` does allow nullable types and I see no technical reason why the flow version should not allow nullable types.

This change would be binary compatible since the type argument's nullability isn't even reflected in the bytecode. It would also be source compatible since it only loosens the restrictions.

ansman:
This closes #2229

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
+++ b/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
@@ -124,7 +124,7 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
  * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
  * Returns `null` if the flow was empty.
  */
-public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
+public suspend fun <T> Flow<T>.firstOrNull(): T? {
     var result: T? = null
     try {
         collect { value ->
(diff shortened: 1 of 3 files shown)


qwwdfsad (commented):
Thanks!

Could you please also remove `Any` upper bound from `singleOrNull` operator and add a corresponding unit test?

ansman:
Done

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -138,10 +138,10 @@ public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
 }
 
 /**
- *  The terminal operator that returns the first element emitted by the flow matching the given [predicate] and then cancels flow's collection.
+ * The terminal operator that returns the first element emitted by the flow matching the given [predicate] and then cancels flow's collection.
  * Returns `null` if the flow did not contain an element matching the [predicate].
```
ansman, line 142: Do we want to comment on what happens if the flow contains a single `null` value or let that be implied?

qwwdfsad (changes requested):
`assertFailsWith<NoSuchElementException> { flowOf<Int?>().singleOrNull() }` doesn't actually fail but returns `null` instead. Added test doesn't pass

ansman:
My bad. I'm unable to run the tests locally because java 6 is needed.

LouisCAD:
@ansman Read the updated `CONTRIBUTING.md` on the main/master branch, it's specified you can make the env variable point to a newer JDK, including JDK11+

qwwdfsad (approved):
Multiple `null` values are handled incorrectly.
I've created #2287 with your changes and fixed it in https://github.com/Kotlin/kotlinx.coroutines/pull/2287/commits/7d006bbf028e3575469f35ddc50411afc4bb122e to reduce a rountrip time and merge it.

Thanks for raising the issue and your contribution, it's always great to evolve our API based on our users feedback!

PR #2231, opened 2020-09-07
Provide way to dedup function calls

### LINKED ISSUE #1097: Add way to de-duplicate function calls
I currently have a number of places where I request a specific URL. It would be nice if any concurrent requests would be de-duplicated so that any requested which are started while one is in progress, joins the other and receives the same result.

I imagine that this would be similar to Go's [singleflight](https://godoc.org/golang.org/x/sync/singleflight): a collection wrapper around a Mutex, keyed on a string.

recheej:
Resolves #1097.

I think naming can def change. Open to suggestions. Let me know what you think for API. Btw, how do you update ```kotlinx-coroutines-core.api```?

### DIFF
--- a/kotlinx-coroutines-core/common/src/Group.kt
+++ b/kotlinx-coroutines-core/common/src/Group.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+public class Group<T, R> internal constructor(
+    private val scope: CoroutineScope,
+    private val groupMap: MutableMap<T, Deferred<R>>,
+    private val forgottenSet: MutableSet<T>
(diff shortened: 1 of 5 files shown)


elizarov:
You are proposing to add some quite massing public API without any docs. It does not work this way. Can you, please, start by explaining the design goals of this PR. What exact problem does it solve? How generic it is? Why it has to add some new public classes to the API?

recheej:
@elizarov sure I can provide more documentation. First, it should be known that I'm trying to replicate [api from go](https://godoc.org/golang.org/x/sync/singleflight#Group) as mentioned in the issue. But sounds like you maybe have issue with new public classes? Mind giving me your thoughts there before I go adding all this documentation to this public class that may be remove in the end?

recheej:
Closing issue as it's required by maintainers that things like this come with a design write up. While I understand the need it's too much investment for me right now.

PR #2232, opened 2020-09-08
Contributing Guidelines

elizarov:
Also moving there building instructions from README.md as only contributors shall care about them.


qwwdfsad (approved):
Great job!

PR #2245, opened 2020-09-12
Add getCurrentJob and getCurrentScopeJob functions

### LINKED ISSUE #2159: Coroutine equivalent of Thread.currentThread() requires a !! cast
The closest equivalent to the JVM's `Thread.currentThread()` for coroutines is

```kotlin
coroutineContext[Job]!!
```

This `!!` cast always succeeds, but it draws the developer's attention and brings in uncertainty about the correctness of the code.  [Here's](https://stackoverflow.com/questions/58972786/how-to-efficiently-show-loading-dialog-if-kotlin-coroutine-job-takes-quite-some/58974475#comment104210622_58974475) an example where this concern was raised on Stack Overflow.

Kotlin could provide a function that hides this check, for example:

```kotlin
suspend fun currentJob(): Job = coroutineContext[Job]!!
```

Originally reported [here](https://youtrack.jetbrains.com/issue/KT-40326)

recheej:
Resolves #2159 .

Adding different names for the suspend and `CoroutineScope` extension versions since there can be some confusion. Added docs and did api dump. Also replaced places in the code base (mostly tests) that accessed jobs with `!!`


elizarov:
@recheej Can you, please, elaborate what's your use-case for these functions? What kind of code needs them?

recheej:
@elizarov I personally would use these functions for debugging coroutines code (see state of job, children, etc). But I could also see them being used for lower level coroutines work or for logging state of things.

In the end, I think the `Job` is such an important context element that it deserves its own function(s).

qwwdfsad (changes requested):
Thanks to your contribution!
We've discussed it internally and decided that it's not the addition we are ready to make. It's unclear who and how is going to use it, its naming is misaligned with existing memebrs (e.g. `CoroutineScope.coroutineContext`, but `currentCoroutineContext()`).

We'd like to introduce the only `coroutineContext.job: Job?` extension property for the sake of brevity where it is necessary. If you want to do this extra-mile and replace your PR with that, we'd be happy to help you with review and integration. If not, it's completely okay to tell us, then we are going to implement it ourselves

recheej:
Sure I can do that change

PR #2247, opened 2020-09-14
Support JetBrains cache-redirector service for stable CI builds

elizarov:
On CI set CACHE_REDIRECTOR=true environment property.

### DIFF
--- a/buildSrc/src/main/kotlin/CacheRedirector.kt
+++ b/buildSrc/src/main/kotlin/CacheRedirector.kt
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import org.gradle.api.*
+import org.gradle.api.artifacts.dsl.*
+import org.gradle.api.artifacts.repositories.*
+import org.gradle.api.initialization.dsl.*
+import java.net.*
+
+/**
+ * Enabled via environment variable, so that it can be reliably accessed from any piece of the build script,
+ * including buildSrc within TeamCity CI.
+ */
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/CacheRedirector.kt
```diff
@@ -0,0 +1,152 @@
...
+            host == "teamcity.jetbrains.com" ||
+            host == "buildserver.labs.intellij.net"
+
+private fun Project.checkRedirectUrl(url: URI, containerName: String): URI {
+    val redirected = url.maybeRedirect()
+    if (redirected == null && !url.isCachedOrLocal()) {
+        val msg = "Repository $url in $containerName should be cached with cache-redirector"
+        val details = "Using non cached repository may lead to download failures in CI builds." +
+            " Check buildSrc/src/main/kotlin/CacheRedirector.kt for details."
+        logger.warn("WARNING - $msg\n$details")
```
qwwdfsad, line 118: Nobody will see this warning until builds start to fail.
Maybe a better strategy is to fail-fast, so the maintainer (this script is going to be copy-pasted :)) will see the failure, file an issue and mute the problem only then.

qwwdfsad (approved):
Otherwise good

buildSrc/src/main/kotlin/CacheRedirector.kt
elizarov: It is not a test, so the only thing I can do is to hard-fail here, which is too much. I've tried doing it as a test (that's what `kotlin` project does), but that turned out to be quite complicated due to various technical issues, since I want to verify repositories in all projects and buildscript.

PR #2252, opened 2020-09-18
Added docs on withTimeout asynchrony and its use with resources

### LINKED ISSUE #2233: Acquiring a semaphore inside withTimeout block doesn't guarantee to release a semaphore when timed out
Hello! Today I have discovered a strange behavior the work of withTimeout method.
I've need a simple function like _tryAcquire(long timeout, TimeUnit unit)_ from _java.util.concurrent.Semaphore_ but working in a non-blocking manner. I've implemented it as an extension function to kotlinx.coroutines.sync.Semaphore class and it looks this:

```kotlin
suspend fun Semaphore.acquirePermitToRequest(pendingTimeMs: Long): Boolean {
  return try {
    withTimeout(pendingTimeMs) {
      acquire()
      true
    }
  } catch (ce: Exception) {
    false
  }
}
```

When I was testing this function by code block below I encountered non-deterministic results:

```kotlin
fun main() = runBlocking {
  val semaphore = Semaphore(10)
  for (i in 1..1) {
    coroutineScope {
      repeat(5000) {
        launch {
          coroutineScope {
            if (semaphore.acquirePermitToRequest(25)) {
              delay(50)
              semaphore.release()
            }
          }
        }
      }
    }

    println(semaphore.availablePermits)
  }
}
```

Sometimes it prints **0** to my console! In other words, there're no available permits in my semaphore but 10 are expected!

I've tried to add some counters that could be useful to research this behavior, all code placed here: https://gist.github.com/akrikheli/121171b3fbca50ddad93e8b8d5b09a48

The output log when this problem is reproduced:
```
0
cancellation counter=4990
acquire counter=20
release counter=10
after with timeout = 10
```

So, the real acquisition of the semaphore was 20 times instead of the expected 10 despite the cancellation was 4990 times (5000-10) and it's a bit confusing.

As a workaround, I can rewrite my extension function using _tryAcquire_ method in the loop. 
May be anybody can explain the current behavior when _acquire_ method used in withTimeout block?

elizarov:
This is a tricky gotcha that needs additional explanation. There are two examples added, one showing the bad code and explaining why it does not work, and the other showing the correct way to write it.

Fixes #2233

### DIFF
--- a/docs/cancellation-and-timeouts.md
+++ b/docs/cancellation-and-timeouts.md
@@ -11,6 +11,7 @@
   * [Closing resources with `finally`](#closing-resources-with-finally)
   * [Run non-cancellable block](#run-non-cancellable-block)
   * [Timeout](#timeout)
+  * [Asynchronous timeout and resources](#asynchronous-timeout-and-resources)
 
 <!--- END -->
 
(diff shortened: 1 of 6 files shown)


docs/cancellation-and-timeouts.md
```diff
@@ -355,6 +356,114 @@ Result is null
...
+}
+
+fun main() {
+//sampleStart
+    runBlocking {
+        repeat(100_000) { // Launch 100K coroutines
+            launch { 
+                var resource: Resource? = null // Not acquired yet
+                try {
+                    withTimeout(60) { // Timeout of 60 ms
```
qwwdfsad, line 440: This example could be simplified with `withTimeoutOrNull` to avoid `try`-`finally`

qwwdfsad (approved):
LGTM

docs/cancellation-and-timeouts.md
elizarov: try-finally here is used here because we can crash when "doing something else" in this code and we still need to close the resource after that.

PR #2253, opened 2020-09-19
Remove redundant stdlib dependencies

turansky:


qwwdfsad (commented):
Could you please ensure that stdlib is still present in POM dependencies in Gradle metadata?

turansky:
Check results:
in POM dependencies - **present** 
in Gradle metadata - **present**

PR #2254, opened 2020-09-19
Plugin repo management in 'pluginManagement' block

turansky:


qwwdfsad:
Merged into `kts-transition` branch

PR #2259, opened 2020-09-22
Update Flow.sample KDoc example timings, add tests

### LINKED ISSUE #2243: Flow.sample() - incorrect documentation example or bug
The example provided in the docs for `Flow.sample(...)` does not seem to behave as the documentation states:

``` Kotin
    flow {
        repeat(10) {
            emit(it)
            delay(50)
        }
    }
    .sample(100)
    .collect { print("$it ")}
```

According to documentation, expected output: `1 3 5 7 9`
Actual output : `2 4 6 8` (sometimes even `2 4 6 8 9`)

However, `debounce`'s example works as documented when tested side-by-side here: https://pl.kotl.in/NS_M7Sx01

This could even be a bug with the playground, but it's weird that `debounce` works as expected in the same environment 🤔

elizarov:
* All Flow.debounce/sample KDoc example code snippets are automatically tested with Knit.
* Flow.sample timings are made larger, so that they produce an expected output when run under the real time, too.

Fixes #2243

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -14,13 +14,30 @@ import kotlinx.coroutines.selects.*
 import kotlin.jvm.*
 import kotlin.time.*
 
+/* Scaffolding for Knit code examples
+<!--- TEST_NAME FlowDelayTest -->
+<!--- PREFIX .*-duration-.*
+@file:OptIn(ExperimentalTime::class)
+----- INCLUDE .*-duration-.*
+import kotlin.time.*
+----- INCLUDE .*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+
+fun main() = runBlocking {
+----- SUFFIX .*
+.toList().joinToString().let { println(it) } }
+-->
+*/
+
 /**
  * Returns a flow that mirrors the original flow, but filters out values
  * that are followed by the newer values within the given [timeout][timeoutMillis].
  * The latest value is always emitted.
  *
  * Example:
- * ```
+ *
+ * ```kotlin
  * flow {
  *     emit(1)
  *     delay(90)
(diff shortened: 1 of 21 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -33,7 +50,14 @@ import kotlin.time.*
  *     emit(5)
  * }.debounce(1000)
  * ```
- * produces `3, 4, 5`.
+ * <!--- KNIT example-delay-01.kt -->
```
qwwdfsad, line 53: It would be nice if these tests could be numbered automatically without interfering with the source.
In general, I do no mind having Knit-specific directives in our markdown, but this documentation will likely be read from the IDEA rather than from a webpage with processed markdown, so reducing the number of MD tags would be nice

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -14,13 +14,30 @@ import kotlinx.coroutines.selects.*
 import kotlin.jvm.*
 import kotlin.time.*
 
+/* Scaffolding for Knit code examples
+<!--- TEST_NAME FlowDelayTest -->
```
qwwdfsad, line 18: Same here, would be nice to hide it in an external file

qwwdfsad (approved):
Nice addition to the Knit!

My comments regarding Knit tag in .kt files are not blockers and can be done later, it's up to you

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
elizarov: They _are_ automatically numbered. When writing those directives you can use any number or even just a placeholder like `<!--- KNIT example-delay-num.kt -->` and when you run `gradlew knit` it will renumber them sequentially and will update the corresponding source files.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
elizarov: Yes. That would be a nice addition. Created https://github.com/Kotlin/kotlinx-knit/issues/18 and https://github.com/Kotlin/kotlinx-knit/issues/18

Issue #2266, opened 2020-09-23
Missing class: java.lang.instrument.ClassFileTransformer


Anigif:
When building our Android app in release mode with gradle (and thereby also R8), the task `minifyReleaseWithR8` gives the following warning: `Missing class: java.lang.instrument.ClassFileTransformer`

Removing the following line from the dependency list the warning dissapers, so I assume it's related to that module: `implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.9"`

Here's stats I hope help identifying the issue:
- targetSdkVersion: 30
- minSdkVersion: 23
- com.android.tools.build:gradle: 4.0.1
- Kotlin version: 1.4.10

I'm not sure if this is related to #2046 since the same class is mentioned. That issue is however for ProGuard and mentions other issues too.

qwwdfsad:
We'll add a rule to ignore this problem in 1.4 release

Issue #2046, opened 2020-05-21
Proguard issues with 1.3.6/1.3.7


ZakTaccardi:
Seems that 1.3.6/1.3.7 bring in the `-debug` artifact, which causes compilation issues when minifying with proguard (using `6.2.2`).

Output for `1.3.7`:
```
Note: there were 17 duplicate class definitions.
      (http://proguard.sourceforge.net/manual/troubleshooting.html#duplicateclass)
Warning: kotlinx.coroutines.debug.AgentPremain$DebugProbesTransformer: can't find superclass or interface java.lang.instrument.ClassFileTransformer
Warning: kotlinx.coroutines.debug.AgentPremain$installSignalHandler$1: can't find superclass or interface sun.misc.SignalHandler
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class java.lang.instrument.ClassFileTransformer
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class java.lang.instrument.Instrumentation
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class java.lang.instrument.Instrumentation
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class sun.misc.Signal
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class sun.misc.Signal
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class sun.misc.SignalHandler
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class sun.misc.Signal
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class java.lang.instrument.Instrumentation
Warning: kotlinx.coroutines.debug.AgentPremain: can't find referenced class java.lang.instrument.Instrumentation
Warning: kotlinx.coroutines.debug.AgentPremain$DebugProbesTransformer: can't find referenced class java.lang.instrument.ClassFileTransformer
Warning: kotlinx.coroutines.debug.AgentPremain$installSignalHandler$1: can't find referenced class sun.misc.SignalHandlerNote: there were 17 duplicate class definitions.
      (http://proguard.sourceforge.net/manual/troubleshooting.html#duplicateclass)
Warning: kotlinx.coroutines.debug.AgentPremain$DebugProbesTransformer: can't find superclass or interface java.lang.instrument.ClassFileTransformer
[...]

PR #2273, opened 2020-09-29
Fix proguard rules

dkhalanskyjb:
Solves https://github.com/Kotlin/kotlinx.coroutines/issues/2046 and
https://github.com/Kotlin/kotlinx.coroutines/issues/2266.

I tested these issues separately, reproduced both problems and checked that this edit solves them, but I'm not sure whether we should also test for this. Maybe make the example Android projects part of the build conditionally?


qwwdfsad (approved):
LGTM

Issue #1490, opened 2019-08-29
Convert `ReceiveChannel<T>` to `Flow<T>` without consuming


ZakTaccardi:
My `ViewModel` exposes a stream of events as a `ReceiveChannel<T>`

```kotlin
// in `ViewModel`
private val events = Channel(capacity = Channel.UNLIMITED) // these are "side effects"

fun events(): ReceiveChannel<T> = events
```

My activity/fragment observes this stream of events, without consuming on cancellation. 

```kotlin
// in an Activity's onCreate

launch {
  for (event in viewModel.events()) {
    ui.handleEvent(event)
  }
}
```

By not consuming on cancellation (unsubscription), this allows my UI to re-use the same `ReceiveChannel<T>` instance across configuration changes. 

```kotlin
// this exists already
fun <T> ReceiveChannel<T>.consumeAsFlow(): Flow<T>
// this does not
fun <T> ReceiveChannel<T>.asFlowWithoutConsuming: Flow<T>
```

I think the latter function would be useful. Can it be added?

elizarov:
Why would you need to have it exposed as `Flow` type? What is wrong with the code that you currently have that explicitly declares your `events` with `ReceiveChannel<T>` type? It looks easier to understand for me this way (the way you have it now).

ZakTaccardi:
@elizarov because a `ReceiveChannel<T>` is mutable - it can be closed by the caller.

```kotlin
// this is fine
launch {
  for (event in viewModel.events()) {
    ui.handleEvent(event)
  }
}
```

```kotlin
// this is not, because on rotation when you subscribe again to the same channel instance
// crash because the channel is already closed
launch {
  viewModel.events()
    .consumeEach { event -> ui.handleEvent(event) }
}
```

elizarov:
Thanks. That's clear. How about `receiveAsFlow` name for this conversion?

ZakTaccardi:
perfect name!

My original thinking is that `receive` indicates a single value, but then I realized there is `consume` for a single value and for multiples there is `consumeEach`. Now I 
am just realizing I originally felt this way because there is no `ReceiveChannel.receiveEach`. Should there be a `.receiveEach { }` too? (maybe could make this a separate issue)

archibishop:
Hey @ZakTaccardi , @elizarov. I have had a look at this issue , no one has implemented it yet and would love to take it up and raise a pr on this. To implement this i was thinking of such an approach . 

[...]

PR #2278, opened 2020-10-02
Workaround for receiveOrClosed is an abstract method issue

stevefan1999-personal:
Majority of the context: https://github.com/Kotlin/kotlinx.coroutines/issues/1490#issuecomment-695959926


elizarov:
That's too an intrusive change. We cannot have this kind of a work-around right in the core method that is optimized for performance. I'd suggest to send PR to vertx to fix their implementation.

PR #2285, opened 2020-10-07
Provide asFlowable and asObservable by their names in binary instead …

### LINKED ISSUE #2182: Ambiguous method error when using RxConvert from Java
While using the extension functions from [RxConvert](https://github.com/Kotlin/kotlinx.coroutines/blob/master/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt) to convert `Flow` to `Observable` or `Flowable` in Java, we fail to build from an ambiguous method error.
`asFlowable()` and `asObservable()` both have a `JVMName` of `from`. The following test case in Java is a way to reproduce this issue:

```java

public class FlowToRxJavaTest {
    @Test
    void observableFromFlowTest() {
        Flow<String> flow = stringFlow();
        Observable<String> observable = RxConvertKt.from(flow);
    }
    @Test
    void flowableFromFlowTest() {
        Flow<String> flow = stringFlow();
        Flowable<String> flowable = RxConvertKt.from(flow);
    }
}
```
A simple fix could be removing the `JVMName` entirely. Another option is changing the `JVMName` to be more specific to the function, but an appropriate name doesn't come to mind.

After a solution is decided I would be more than happy to create a PR for a fix.

qwwdfsad:
…of 'from' function to prevent naming clash for Java users.

    * Do not provide @JvmOverloads for convenience of Java interop
    * Deprecate ReceiveChannel.asObservable by the way

Fixes #2182


elizarov (approved):
👍 Great stuff.

PR #2286, opened 2020-10-07
Clarify documentation for IO dispatcher

### LINKED ISSUE #2272: IO Dispatcher creates more than 64 threads
Annotation describes IO Dispatcher like this,

_It defaults to the limit of 64 threads or the number of cores (whichever is larger)._

As I understood, Maximum number of threads that can be created by Dispatchers.IO is up to 64 because I use only 6core and 12 threads CPU, but my result is a little bit tricky.

This is my code.

```
repeat(500) {
    CoroutineScope(Dispatchers.IO).launch {
        println(Thread.currentThread().name)
    }
}
```
and this is the result.

```
...

DefaultDispatcher-worker-18
DefaultDispatcher-worker-46
DefaultDispatcher-worker-17
DefaultDispatcher-worker-47
DefaultDispatcher-worker-69
DefaultDispatcher-worker-64
DefaultDispatcher-worker-66
DefaultDispatcher-worker-67
DefaultDispatcher-worker-68
DefaultDispatcher-worker-41

...
```
Why is threadpool count bigger than 64? 
Does that mean actual threads was created more than 64?

qwwdfsad:
Fixes #2272

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/Dispatchers.kt
+++ b/kotlinx-coroutines-core/jvm/src/Dispatchers.kt
@@ -97,7 +97,7 @@ public actual object Dispatchers {
      * The [CoroutineDispatcher] that is designed for offloading blocking IO tasks to a shared pool of threads.
      *
      * Additional threads in this pool are created and are shutdown on demand.
-     * The number of threads used by this dispatcher is limited by the value of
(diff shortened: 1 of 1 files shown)


elizarov (approved):
Good. I really hope it helps.

PR #2287, opened 2020-10-07
Pr/2230

qwwdfsad:
Editable version of #2230

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
+++ b/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
@@ -112,7 +112,7 @@ public suspend fun <T> Flow<T>.first(predicate: suspend (T) -> Boolean): T {
  * The terminal operator that returns the first element emitted by the flow and then cancels flow's collection.
  * Returns `null` if the flow was empty.
  */
-public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
+public suspend fun <T> Flow<T>.firstOrNull(): T? {
     var result: T? = null
     collectWhile {
         result = it
@@ -122,10 +122,10 @@ public suspend fun <T : Any> Flow<T>.firstOrNull(): T? {
 }
 
 /**
- *  The terminal operator that returns the first element emitted by the flow matching the given [predicate] and then cancels flow's collection.
+ * The terminal operator that returns the first element emitted by the flow matching the given [predicate] and then cancels flow's collection.
  * Returns `null` if the flow did not contain an element matching the [predicate].
  */
-public suspend fun <T : Any> Flow<T>.firstOrNull(predicate: suspend (T) -> Boolean): T? {
+public suspend fun <T> Flow<T>.firstOrNull(predicate: suspend (T) -> Boolean): T? {
     var result: T? = null
     collectWhile {
         if (predicate(it)) {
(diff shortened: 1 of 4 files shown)


ansman:
Now that I think about it it feels weird that `singleOrNull` ever throws. It feels like it should do what `single()` does but instead of throwing it returns `null`.

That is of course a pretty major change in behavior.

qwwdfsad:
That's actually my bad, it should return `null` and your original change was correct.
Just checked the wrong operator

ansman (approved):

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -47,33 +48,43 @@ public suspend inline fun <T, R> Flow<T>.fold(
...
- * The terminal operator, that awaits for one and only one value to be published.
+ * The terminal operator that awaits for one and only one value to be emitted.
  * Throws [NoSuchElementException] for empty flow and [IllegalStateException] for flow
  * that contains more than one element.
  */
 public suspend fun <T> Flow<T>.single(): T {
     var result: Any? = NULL
     collect { value ->
-        if (result !== NULL) error("Expected only one element")
+        require(result == NULL) { "Flow has more than one element" }
```
elizarov, line 58: Must use `===` to avoid problems with broken-equals types (tests can be added for this kind of class, too).

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -47,33 +48,43 @@ public suspend inline fun <T, R> Flow<T>.fold(
...
+ * Returns the single value or `null`, if the flow was empty or emitted more than one value.
  */
-public suspend fun <T: Any> Flow<T>.singleOrNull(): T? {
-    var result: T? = null
+public suspend fun <T> Flow<T>.singleOrNull(): T? {
+    var result: Any? = NULL
     collect { value ->
-        if (result != null) error("Expected only one element")
-        result = value
+        /*
```
elizarov, line 73: The `singleOrNull` operator shall be written via `collectWhile` to terminate yearly with a `null` result when the second value is encountered. This way, it will be consistent with stdlib `Sequence.singleOrNull` and with `Flow.single()` that also terminates early on the second item.

elizarov (changes requested):

kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt
```diff
@@ -47,33 +48,43 @@ public suspend inline fun <T, R> Flow<T>.fold(
...
+         * result === NULL -> first value
+         * result === user value -> we already had first value and second one arrived
+         * result === DONE -> we've seen more than one value, time to return it
+         * as well.
+         */
+        result = if (result === NULL) {
+            value
+        } else {
+            // Indicator that more than one value was observed
+            DONE
```
EdwarDDay, line 83: It looks like `singleOrNull` checks every element in the `Flow`. This shouldn't be necessary. After the 2nd element it could return `null`.
```kotlin
@Test
fun testDoNotCheckThirdElement() = runTest {
    val flow = flow {
        emit(1)
        emit(2)
        throw TestException() // should never be executed
    }
    assertFailsWith<IllegalArgumentException> { flow.single() }
    assertNull(flow.singleOrNull())
}
```
This could be fixed by using `collectWhile`.

elizarov (approved):
🚀 Good to go

Issue #2274, opened 2020-09-29
DebugProbesKt.bin gets shipped in release apk


G00fY2:
Not a big deal, but I am wondering if the Android build process should really add this to every release APK (gets placed into the root folder).

Tested with Android Studio 4.0.1 and 4.2.0 Canary 12 using the latest Kotlin and Coroutines Android plugins/libraries.

As a workaorund it should be safe to use 
```
packagingOptions {
  exclude 'DebugProbesKt.bin'
}
```
for release builds, right?

elizarov:
Yes. It is safe to exclude. I wonder if we can also add the corresponding exclusion options directly to the library. Is it even possible?

qwwdfsad:
Fixed by providing how-to documentation in 1.4.0

G00fY2:
For everyone reading this, [this ](https://github.com/Kotlin/kotlinx.coroutines/commit/b82439e6b570249cfbd25622dc06cb0ec156c006#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)was added to the README.

bubenheimer:
@qwwdfsad Can you clarify the how-to instructions, please? They are hard to actually understand and follow. They say "no loss of functionality", but it would appear that I will lose some mystery debug capabilities. The instructions appear to remove these capabilities from both release and debug builds, and not being a Gradle wiz, like most devs, I'm not sure how to limit this to only release builds. My guess, after some meddling, would be to put it in `android { buildTypes { release`, but I can't say whether that actually does the right thing.

Or maybe the Gradle snippet is usable exactly as written and meant in a "just trust us" sense? Does "DebugProbesKt.bin" get introduced only in release build dependencies? "Just trust us" is generally unwise for a dev. Please add additional explanation.

qwwdfsad:
>but it would appear that I will lose some mystery debug capabilities.

Could you please elaborate on what exact part of debug functionality disappears when you are applying `packagingOptions
`? The instruction tells nothing about debugging capabilities of `kotlinx.coroutines` and explains that "a resource file that is not required for the coroutines to operate
normally and is only used by the debugger" (meaning IDEA debugger, not an application-level capabilities)

bubenheimer:
[...]

Issue #2023, opened 2020-05-13
Compilation error on the androidTest configuration after updating to 1.3.6


dmarin:
The compilation error occurs after updating to `1.3.6`, and it gets fixed if I downgrade to `1.3.5`.

Kotlin Version 1.3.72

The coroutines dependencies I am using are:
``` 
implementation platform("org.jetbrains.kotlinx:kotlinx-coroutines-bom:${Versions.COROUTINES_VERSION}")
implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android"
implementation "org.jetbrains.kotlinx:kotlinx-coroutines-play-services"

testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test"
androidTestImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test"
``` 

Stacktrace:
```
* What went wrong:
Execution failed for task ':app:mergeStagingDebugAndroidTestJavaResource'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
   > More than one file was found with OS independent path 'win32-x86/attach_hotspot_windows.dll'
* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:mergeStagingDebugAndroidTestJavaResource'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:207)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:205)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:186)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:114)
	at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
	at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
[...]

PR #2288, opened 2020-10-07
Document problems with building coroutines-using Android projects

dkhalanskyjb:
Added instructions to work around #2023.
Also, the answer to #2274 is now documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,17 @@ threads are handled by Android runtime.
 R8 and ProGuard rules are bundled into the [`kotlinx-coroutines-android`](ui/kotlinx-coroutines-android) module.
 For more details see ["Optimization" section for Android](ui/kotlinx-coroutines-android/README.md#optimization). 
 
+#### Avoiding including the debug infrastructure in the resulting APK
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -176,6 +176,17 @@ threads are handled by Android runtime.
 R8 and ProGuard rules are bundled into the [`kotlinx-coroutines-android`](ui/kotlinx-coroutines-android) module.
 For more details see ["Optimization" section for Android](ui/kotlinx-coroutines-android/README.md#optimization). 
 
+#### Avoiding including the debug infrastructure in the resulting APK
+
+`kotlinx-coroutines-core` provides resources that are not required for the applications to operate and are only needed
+for debugging coroutine-using code. To exclude them at no loss of functionality, add the following to the `android`
```
qwwdfsad, line 182:
```suggestion
`kotlinx-coroutines-core` artifact contains a resource file that is not required for the coroutines to operate normally and is only used by the debugger. To exclude it at no loss of functionality, add the following snippet to the `android`
```

qwwdfsad (approved):
Otherwise looks good, thanks!

PR #2292, opened 2020-10-09
Fix doc references that differ only in case (function vs interface)

### LINKED ISSUE #2279: Coroutines basics Docs
While reading the coroutines basics docs https://kotlinlang.org/docs/reference/coroutines/basics.html 

I noticed that a reference link to [" CoroutineScope"](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/) interface leads to the ["coroutineScope"]( https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html) function.  

On the docs it reads like this, "Essentially, coroutines are light-weight threads. They are launched with launch coroutine builder in a context of some CoroutineScope." but the link on the "CoroutineScope" here leads to the [coroutineScope function](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html), rather than ["CoroutineScope" interface](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/) . 

Personally, I feel referencing the interface should be the right thing but I could be wrong.

elizarov:
Fixes #2279

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ suspend fun main() = coroutineScope {
   * [delay] and [yield] top-level suspending functions;
   * [Flow] &mdash; cold asynchronous stream with [flow][_flow] builder and comprehensive operator set ([filter], [map], etc);
   * [Channel], [Mutex], and [Semaphore] communication and synchronization primitives;
-  * [coroutineScope], [supervisorScope], [withContext], and [withTimeout] scope builders;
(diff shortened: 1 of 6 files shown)


qwwdfsad (approved):
Nice!

PR #2308, opened 2020-10-15
Zip and combine performance

qwwdfsad:


kotlinx-coroutines-core/common/test/flow/operators/ZipTest.kt
```diff
@@ -67,14 +68,35 @@ class ZipTest : TestBase() {
         val f1 = flow<String> {
             emit("1")
             emit("2")
-            expectUnreached() // the above emit will get cancelled because f2 ends
         }
 
-        val f2 = flowOf("a", "b")
+        val f2 =flow<String> {
```
LouisCAD, line 73:
```suggestion
        val f2 = flow<String> {
```

kotlinx-coroutines-core/common/test/flow/operators/ZipTest.kt
```diff
@@ -174,19 +201,18 @@ class ZipTest : TestBase() {
...
         val f2 = flow {
             expect(2)
             emit(1)
-            expect(5)
+            expect(3)
             hang { expect(6) }
         }
 
-        val flow = f1.zip(f2, { _, _ -> 1 }).onEach { expect(3) }
+        val flow = f1.zip(f2, { _, _ -> 1 }).onEach { expect(4) }
```
LouisCAD, line 215:
```suggestion
        val flow = f1.zip(f2) { _, _ -> 1 }.onEach { expect(4) }
```

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,129 +9,141 @@ import kotlinx.coroutines.*
...
+                // Optimization to avoid AFE allocation when the other flow is done
+                if (collectJob.isActive) collectJob.cancel(AbortFlowException(this@unsafeFlow))
+            }
 
-        val otherIterator = second.iterator()
-        try {
-            first.consumeEach { value ->
-                if (!otherIterator.hasNext()) {
-                    return@consumeEach
+            val newContext = coroutineContext + scopeJob
```
elizarov, line 104: `newContext` is the same as `coroutineContext`, since `scopeJob` is actually a `Job` from the `coroutineContext`. It would be easier to understand if `newContext` is simply renamed to `scopeContext`. For the same reason you don't need `val scopeJob` at all.

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,129 +9,141 @@ import kotlinx.coroutines.*
...
+                 *    so we emit to downstream from coroutineScope job.
+                 * Typically, such hierarchy requires coroutine for collector that communicates
+                 * with coroutines scope via a channel, but it's way too expensive, so
+                 * we are using this trick instead.
+                 */
+                withContextUndispatched( coroutineContext + collectJob) {
+                    flow.collect { value ->
+                        withContextUndispatched(newContext, cnt) {
+                            val otherValue = second.receiveOrNull() ?: throw AbortFlowException(this@unsafeFlow)
+                            emit(transform(NULL.unbox(value), NULL.unbox(otherValue)))
```
elizarov, line 123: `NULL.unbox(value)` is not needed. It is not boxed.

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,129 +9,141 @@ import kotlinx.coroutines.*
...
-         *
-         * Invariant: this clause is invoked only when all elements from the channel were processed (=> rendezvous restriction).
-         */
-        (second as SendChannel<*>).invokeOnClose {
-            if (!first.isClosedForReceive) first.cancel(AbortFlowException(this@unsafeFlow))
-        }
+internal fun <T1, T2, R> zipImpl(flow: Flow<T1>, flow2: Flow<T2>, transform: suspend (T1, T2) -> R): Flow<R> =
+    unsafeFlow {
+        coroutineScope {
+            val second = asChannel(flow2)
```
elizarov, line 85: Btw, it is the only usage of `asChannel` now. It would read better if `asChannel` is simply inlined here (since you can see right here how it boxes values that are sent to the channel).

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,129 +9,141 @@ import kotlinx.coroutines.*
...
+                            val otherValue = second.receiveOrNull() ?: throw AbortFlowException(this@unsafeFlow)
+                            emit(transform(NULL.unbox(value), NULL.unbox(otherValue)))
+                        }
+                    }
                 }
-                emit(transform(NULL.unbox(value), NULL.unbox(otherIterator.next())))
+            } catch (e: AbortFlowException) {
+                e.checkOwnership(owner = this@unsafeFlow)
+            } finally {
+                if (!second.isClosedForReceive) second.cancel(AbortFlowException(this@unsafeFlow))
```
elizarov, line 130: You don't need to pass `AbortFlowException` here. `second` scope is outside this `try { ... } catch { ... }` anyway and you don't actually need to isolate second channel cancellation. Simply calling `cancel()` here works just as well.

elizarov (changes requested):

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,129 +9,141 @@ import kotlinx.coroutines.*
...
-                    if (latestValues[i] == null) --remainingNulls
-                    latestValues[i] = value
-                    if (remainingNulls != 0) return@onReceive
-                    val arguments = arrayFactory()
-                    for (index in 0 until size) {
-                        arguments[index] = NULL.unbox(latestValues[index])
-                    }
-                    transform(arguments as Array<T>)
+    latestValues.fill(UNINITIALIZED) // Smaller bytecode & faster that Array(size) { UNINITIALIZED }
+    val isClosed = BooleanArray(size)
```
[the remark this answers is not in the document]
qwwdfsad: Yep, good catch

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,129 +9,141 @@ import kotlinx.coroutines.*
...
+             * in the following scenario:
+             * ```
+             * val f1 = flow { emit(1); delay(Long.MAX_VALUE) }
+             * val f2 = flowOf(1)
+             * f1.zip(f2) { ... }
+             * ```
+             *
+             * Invariant: this clause is invoked only when all elements from the channel were processed (=> rendezvous restriction).
+             */
+            val collectJob = Job()
```
elizarov, line 97: The way this is written now would not properly display in coroutines debugger when the first flow is being collector. It should be, at least, changed to `val collectJob = Job(coroutineContext[Job])`, so it is a child of the current job. But, in fact, it should use an implementation of the `Job` that inherits `ScopeCoroutine`.

qwwdfsad: Good observation, tho it works perfectly fine in the debugger right now because `withContextUndispatched` creates no coroutine and frame walker successfully finds the last suspension point.

Adding `Job(coroutineContext[Job])` will make things worse as standalone and irrelevant `JobImpl` will be printed by `DebugProbes.printJob`

elizarov: Currently it does not really work if you put breakpoint inside `collect`. Checkout this test in debugger:

```
package kotlinx.coroutines

import kotlin.test.*

class ScratchTest : TestBase() {
    @Test
    fun test() = runTest {
        val f1 = flowOf("a")
        val f2 = flowOf(1)
        f1.zip(f2) { i, j -> i + j }.collect {
            foo()
        }
    }
}

suspend fun foo() {
    yield() // suspend
    println("!!!") // breakpoint
}
```

You see the 1st coroutine in the debugger, but it does not show the correct suspention point inside `foo`, because the `Job` that it is suspended inside is detached.

qwwdfsad: Aha! I was thinking about the other job.

>But, in fact, it should use an implementation of the Job that inherits ScopeCoroutine

Won't really help because stackwalking doesn't leverage jobs in any way (otherwise `withContext(Job())` wouldn't work)
Neither won't help `Job(coroutineContext[Job])`.

Another observation is that we have broken stackwalking in multiple places: in `zip` and, for the same reason, in `flowOn`. Also, while writing tests for "the same as zip, but flowOn", I've uncovered a bug in `SafeCollector` with the same cause.
Luckily they all have straightforward fix

elizarov (approved):
🚀 Great!
(however, `LocalAtomicRef` can be removed)

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -9,133 +9,135 @@ import kotlinx.coroutines.*
...
-                    if (latestValues[i] == null) --remainingNulls
-                    latestValues[i] = value
-                    if (remainingNulls != 0) return@onReceive
-                    val arguments = arrayFactory()
-                    for (index in 0 until size) {
-                        arguments[index] = NULL.unbox(latestValues[index])
-                    }
-                    transform(arguments as Array<T>)
+    latestValues.fill(UNINITIALIZED) // Smaller bytecode & faster that Array(size) { UNINITIALIZED }
+    val resultChannel = Channel<Update>(flows.size)
```
denis-bezrukov, line 27: can be
```suggestion
    val resultChannel = Channel<Update>(size)
```
for consistensy with the following lines. 

Is this size important for fairness? Will something breaks if one of the flows end early (e.g. there will be N-1 active flows, but N slots in the channel)? as I see `testFairnessInVariousConfigurations` takes only frolws with the same size

qwwdfsad: Yes, it is. Channel has a slot for each flow and each flow gives a chance to the rest of flows to emit (using `yield`).
Most of the single-threaded tests that have their flows in `runBlocking` will fail with this change

denis-bezrukov: @qwwdfsad I was mostly asking about the case when `slot count` > `active flows`. But it also seems ok 🤔

PR #2311, opened 2020-10-16
Remove workaround for KT-42671 that triggers test failures from IDEA …

qwwdfsad:
…and access properties in a safe manner from agent premain instead


elizarov (approved):
💯 That's the actual fix.

PR #2312, opened 2020-10-16
Introduce CoroutineContext.job extension

### LINKED ISSUE #2159: Coroutine equivalent of Thread.currentThread() requires a !! cast

qwwdfsad:
Fixes #2159


elizarov (approved):
🚀 Good to go.

PR #2314, opened 2020-10-18
Add debounce selector

### LINKED ISSUE #1216: Add debounce operator with selector to Flow

mkano9:

Resolves #1216.

Reopening [my PR](https://github.com/Kotlin/kotlinx.coroutines/pull/2148) as I accidentally messed up my branch.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -64,36 +64,58 @@ fun main() = runBlocking {
  */
 @FlowPreview
 public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
-    require(timeoutMillis > 0) { "Debounce timeout should be positive" }
-    return scopedFlow { downstream ->
-        // Actually Any, KT-30796
-        val values = produce<Any?>(capacity = Channel.CONFLATED) {
-            collect { value -> send(value ?: NULL) }
-        }
-        var lastValue: Any? = null
-        while (lastValue !== DONE) {
-            select<Unit> {
-                // Should be receiveOrClosed when boxing issues are fixed
-                values.onReceiveOrNull {
-                    if (it == null) {
-                        if (lastValue != null) downstream.emit(NULL.unbox(lastValue))
-                        lastValue = DONE
-                    } else {
-                        lastValue = it
-                    }
-                }
-
-                lastValue?.let { value ->
-                    // set timeout when lastValue != null
-                    onTimeout(timeoutMillis) {
-                        lastValue = null // Consume the value
-                        downstream.emit(NULL.unbox(value))
-                    }
-                }
-            }
-        }
-    }
+    require(timeoutMillis >= 0L) { "Debounce timeout should not be negative" }
+    if (timeoutMillis == 0L) return this
+    return debounceInternal { timeoutMillis }
 }
+
+/**
+ * Returns a flow that mirrors the original flow, but filters out values
+ * that are followed by the newer values within the given [timeout][timeoutMillis].
+ * The latest value is always emitted.
+ *
+ * A variation of [debounce] that allows specifying the timeout value dynamically.
+ *
+ * Example:
+ *
+ * ```kotlin
+ * flow {
+ *     emit(1)
+ *     delay(90)
+ *     emit(2)
+ *     delay(90)
+ *     emit(3)
+ *     delay(1010)
+ *     emit(4)
+ *     delay(1010)
+ *     emit(5)
+ * }.debounce {
+ *     if (it == 1) {
+ *         0L
+ *     } else {
+ *         1000L
+ *     }
+ * }
+ * ```
+ * <!--- KNIT example-delay-02.kt -->
+ *
+ * produces the following emissions
+ *
+ * ```text
+ * 1, 3, 4, 5
+ * ```
+ * <!--- TEST -->
+ *
+ * Note that the resulting flow does not emit anything as long as the original flow emits
+ * items faster than every [timeoutMillis] milliseconds.
+ *
+ * @param timeoutMillis [T] is the emitted value and the return value is timeout in milliseconds.
+ */
+@FlowPreview
+public fun <T> Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T> =
+    debounceInternal { emittedItem ->
+        timeoutMillis(emittedItem)
+    }
 
 /**
  * Returns a flow that mirrors the original flow, but filters out values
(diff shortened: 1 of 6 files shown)


mkano9:
@elizarov I opened this as a new PR as I messed up my branch in https://github.com/Kotlin/kotlinx.coroutines/pull/2148. 

I made the changes requested and it's ready for another review. Thanks in advance.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,98 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
  */
 @ExperimentalTime
 @FlowPreview
-public fun <T> Flow<T>.debounce(timeout: Duration): Flow<T> = debounce(timeout.toDelayMillis())
+public fun <T> Flow<T>.debounceWithDuration(timeout: Duration): Flow<T> = debounce(timeout.toDelayMillis())
```
mkano9, line 154: It's not ideal that I am renaming this method from `Flow<T>.debounce(timeout: Duration): Flow<T>`. 

The selector variation method name `Flow<T>.debounce(timeout: (T) -> Duration): Flow<T>` below was conflicting with `Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T>`. I could just rename only the duration selector variation to e.g. `Flow<T>.debounceWithDurationSelector(timeout: (T) -> Duration): Flow<T>`, but I felt this would be a better grouping to keep the word `duration` in the method names for both instead of just one of them. 

Please let me know if renaming an existing method concerns.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,98 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+    scopedFlow { downstream ->
+        // Actually Any, KT-30796
+        val values = produce<Any?>(capacity = 0) {
+            collect { value -> send(value ?: NULL) }
+        }
+        var lastValue: Any? = null
+        while (lastValue !== DONE) {
+            select<Unit> {
+                // Give a chance to consume lastValue first before onReceiveOrNull receives a new value
+                lastValue?.let { value ->
```
mkano9, line 215: I reversed the order of `values.onReceiveOrNull` and `lastValue?.let` because when the flow emits `"A", "B"` at once like the `testZeroDebounceTimeSelector` test, `values.onReceiveOrNull` kept picking it up first and get selected by `select<Unit>`; therefore, skipping the first value `"A"`.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
elizarov: The right way to do it is this:
* Keep the Kotlin name `debounce`
* Add `@JvmName("debounceDuration")` to avoid platform declaration clash (we don't use `With` or other prepositions in these kind of names).

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,98 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+                    val unboxedValue: T = NULL.unbox(value)
+                    val timeoutMillis = timeoutMillisSelector(unboxedValue)
+                    require(timeoutMillis >= 0L) { "Debounce timeout should not be negative" }
+
+                    if (timeoutMillis == 0L) {
+                        lastValue = null
+                        runBlocking {
```
elizarov, line 222: What's the purpose of `runBlocking/launch` here? It should be just `downstream.emit`.

elizarov (changes requested):
Also needs to fix the name of `debounchWithDelay` (see comment) and update `apiDump`.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
mkano9: I get the `Suspension functions can be called only within coroutine body` error. Any suggestions? I used `runBlocking` to make sure it gets executed before moving on.

elizarov: That's exactly what `emit` is doing.

mkano9: Yes, I meant to say that the `select {}` block does not let me call a suspend function `downstream.emit(unboxedValue)`.

elizarov: You should move this code outside of `select {}` block, that is, you should check conditions and `emit` if needed first, then do select.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
mkano9: `@JvmName` does not seem to work when a function argument is used. I see the issue mentioned in https://youtrack.jetbrains.com/issue/KT-22119. I pushed with `debounceDuration(...)`.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
mkano9: Thanks, I pushed a fix for this.

mkano9:
@elizarov It's ready for another review.

elizarov:
I've restored binary compatibility with the previous version, cleaned up implementation a bit, and reopened it as #2336 (see the last commit for the proper combination of annotation)

mkano9:
@elizarov Thank you for the reviews and for merging it! :)

PR #2315, opened 2020-10-19
Cleanup lazy coroutines that have been cancelled but not yet garbage …

### LINKED ISSUE #2294: Lazy coroutine stays in DebugProbes output if never started
When a coroutine with `start = CoroutineStart.LAZY` is cancelled before ever starting, `DebugProbes` seems to keep them around forever.

Example:

```
import kotlinx.coroutines.*
import kotlinx.coroutines.debug.DebugProbes

object Main {

    @JvmStatic
    fun main(args: Array<String>): Unit = runBlocking {
        DebugProbes.enableCreationStackTraces = false
        DebugProbes.install()

        val job = launch(start = CoroutineStart.LAZY) {
            delay(Long.MAX_VALUE)
        }
        job.invokeOnCompletion { println("job stopped: ${it?.message}") }

        job.cancel("foo")
        job.join()

        DebugProbes.dumpCoroutines(System.out)
    }
}
```

Output:
```
job stopped: foo
Coroutines dump 2020/10/10 13:52:45

Coroutine LazyStandaloneCoroutine{Cancelled}@26b3fd41, state: CREATED
	(Coroutine creation stacktrace)
Process finished with exit code 0
```

Note that if that lazy job is ever started, it gets correctly removed.

Example with `start = CoroutineStart.DEFAULT`:

```
import kotlinx.coroutines.*

object Main {

    @JvmStatic
    fun main(args: Array<String>): Unit = runBlocking {
        DebugProbes.enableCreationStackTraces = false
        DebugProbes.install()

        val job = launch {
            delay(Long.MAX_VALUE)
        }
        job.invokeOnCompletion { println("job stopped: ${it?.message}") }

        job.cancel("foo")
        job.join()

        DebugProbes.dumpCoroutines(System.out)
    }
}
```

Output:
```
job stopped: foo
Coroutines dump 2020/10/10 13:52:29
Process finished with exit code 0
```

qwwdfsad:
…collected

Fixes #2294

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
@@ -158,8 +158,8 @@ internal object DebugProbesImpl {
                 // Leave in the dump only the coroutines that were not collected while we were dumping them
                 .mapNotNull { owner ->
                     // Fuse map and filter into one operation to save an inline
-                    if (owner.notFinished()) owner.info.context?.let { context -> create(owner, context) }
-                    else null
+                    if (owner.isFinished()) null
+                    else owner.info.context?.let { context -> create(owner, context) }
                 }
         }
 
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
```diff
@@ -183,10 +187,27 @@ internal object DebugProbesImpl {
...
+    /*
+     * Filters out coroutines that do not call probeCoroutineCompleted,
+     * are completed, but not yet garbage collected.
+     *
+     * Typically, we intercept completion of the coroutine so it invokes "probeCoroutineCompleted",
+     * but it's not the case for lazy coroutines that get cancelled before start.
+     */
+    private fun CoroutineOwner<*>.notFinished(): Boolean {
+        // Guarded by lock
+        val job = info.context?.get(Job) ?: return true
```
qwwdfsad, line 199: Seems okay to skip coroutines with garbage-collected context (will be filtered one call later) or without a job

elizarov: It looks Ok. However, to my taste double-negation in the logic makes it hard to understand. I'd rather call it `isFinished` an negate its return value.

elizarov (approved):
👍 Very good

Issue #974, opened 2019-02-07
SendChannel.offer should never throw


LouisCAD:
As of version 1.1.1, if you call `offer` from the non blocking world (which doesn't handle `CancellationException` by default, on a closed `Channel`, an exception is thrown, which by default will crash the program.

This is a problem in several ways:
1. `offer` which returns a `Boolean` per its signature actually is kind of a "tri-lean" as if the channel is closed, it will not return `false` but throw.
2. `offer` is possibly called from a thread that is not the one where the callback that offers the values is unregistered, so race conditions may imply `offer` being called once or a few times after the channel is closed, and usually, the intention is to ignore those. Simply returning `false` instead of throwing would likely be a better behavior.
3. There's no way to `offer` without having this function possibly throw, because by the time `isClosedForSend` returns `false`, it may have been closed due to concurrent execution, so a subsequent call to `offer` would throw.

My current workaround is the following:
```kotlin
fun <E> SendChannel<E>.offerCatching(element: E): Boolean {
    return runCatching { offer(element) }.getOrDefault(false)
}
```

elizarov:
We can actually have a slightly different proposal in mind:

```
suspend fun <E> ReceiveChannel<E>.receiveOrClosed(): ValueOrClosed<E>
suspend fun <E> SendChannel<E>.sendOrClosed(element: E): ValueOrClosed<Unit>
fun <E> SendChannel<E>.offerOrClosed(element: E): ValueOrClosed<Boolean>
```

where a `ValueOrClosed<T>` is a special class we are going to introduce that contains either a value of type `T` or a closed token.

See also #330

LouisCAD:
I think the "OrClosed" suffixed functions would greatly fit the use cases where you need to know the channel was closed.
However, there are some cases where handling the closed case is not desirable and ignoring it is preferable (think interop with callback hell based APIs that can call back on an unspecified thread).

If the "OrClosed" suffixed functions make it to a release, would the `offer` function (and maybe `poll` although less useful IMHO) change its behavior to stop throwing when the channel is closed and return false instead?

elizarov:
I don't have an answer on the last question yet. Indeed, we are running into this and some other issues that might require us tweaking semantics of channels with respect to the close operations, however it is not clear how we could introduce those changing without breaking a lot of code.

LouisCAD:
I wonder if someone ever caught that closed exception satisfying a use case with this behavior.

elizarov:
[...]

PR #2316, opened 2020-10-19
1.4.0 experimentality

qwwdfsad:
Matters to discuss/TODO list:

- [x] `StateFlow`, `ShareFlow` and co. Let's lift experimentality right now?
- [ ] `callbackFlow`/`channelFlow`. TBD, but seems like it's time to stabilize it
- [x] Reactive integrations experimentality. A lot of open issues regarding lifecycle management, let's leave it as is
- [x] ~`(Conflated)BroadcasChannel`. It's definitely time to mark them as `Obsolete`, but if we agree on p. 1, let's deprecate it with a warning for future removal (hiding, at least).~ **Resolution:** left as is, deprecate with `WARNING` in 1.5.0 when shared/state flows are well-used and widespread


kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -95,7 +95,6 @@ public interface Delay {
  * }
  * ```
  */
-@ExperimentalCoroutinesApi
 public suspend fun awaitCancellation(): Nothing = suspendCancellableCoroutine {}
```
qwwdfsad, line 98: I feel uneasy about this one.

On the one hand, it's okay the way it is. On the other hand, we have `awaitClose` that receives `block` to invoke on close (a bit inconsistent) and `hang` function in our tests that I use rather extensively.

WDYT about changing it to
```
public suspend [inline] fun awaitCancellation(afterCancellation: () -> Unit = {}): Nothing {
    try {
        suspendCancellableCoroutine<Nothing> {}
    } finally {
        afterCancellation()
    }
}
```
?

pacher:
W.r.t SharedFlow, I would suggest to keep `resetReplayCache` experimental. I did not provide a compelling use-case yet, but I do believe that it should return current `replayCache` instead of Unit and might convince you in the future.

kotlinx-coroutines-core/common/src/Delay.kt
LouisCAD: When I submitted `awaitCancellation()`, @elizarov told me the bytecode seemed to be more efficient without the `inline` keyword, so I dropped it (though I don't know how to check and compare these matters yet).

Assuming this is right, shouldn't we have an overload of `awaitCancellation` that takes no default value for its lambda parameter?

LouisCAD:
Please, don't stabilize `callbackFlow`/`channelFlow` before #974 is addressed.

kotlinx-coroutines-core/common/src/CoroutineStart.kt
```diff
@@ -71,7 +71,7 @@ public enum class CoroutineStart {
      *
      * **Note: This is an experimental api.** Execution semantics of coroutines may change in the future when this mode is used.
      */
-    @ExperimentalCoroutinesApi
+    @ExperimentalCoroutinesApi  // Since 1.0.0, no ETA on stability
```
elizarov, line 74: Unlike ATOMIC, this mode does not seem problematic to stabilize. It has pretty well-defined behavior. 

However, it was originally envisioned as a performance optimization for UI thread callbacks which wound start coroutines in `UNDISPATCHED` mode to avoid extra dispatch, but this use-case is exceedingly rare nowadays since we have a stable `immediate` dispatcher which gives the same kind of optimization, so it does no seem necessary anymore to hurry up with stabilization of this mode.

kotlinx-coroutines-core/common/src/Delay.kt
elizarov: I'd vote for stabilizing `awaitCancellation()` as is. If we find real-life use-cases for adding `awaitClose`-like block to it, then we can always add the corresponding inline overload later in the future. Moreover, it would be more efficient to implement a variant with a block as an inline function that is calling this non-inline `awaitCancellation()` without a block, wrapping it into `try/finally` (`suspendCancellableCoroutine<Nothing> {}` produces much more bytecode).

elizarov (approved):

elizarov (changes requested):
+1 to make StateFlow/SharedFlow stable. There've been quite extensive community testing already, bugs were found and fixed.

elizarov (approved):
👍

kotlinx-coroutines-core/common/src/Debug.common.kt
```diff
@@ -27,7 +27,7 @@ internal expect fun assert(value: () -> Boolean)
  * Copy mechanism is used only on JVM, but it might be convenient to implement it in common exceptions,
  * so on JVM their stacktraces will be properly recovered.
  */
-@ExperimentalCoroutinesApi
+@ExperimentalCoroutinesApi // Since 1.2.0, no ETA on stability
```
ghus-raba, line 30: What is needed for `CopyableThrowable` to become stable? Is there any issue to watch? Can we help somehow?

This came to my attention, because Retrofit now produces not very helpful stacktraces from suspending calls and the corresponding fix is blocked because this is experimental: https://github.com/square/retrofit/issues/3474

qwwdfsad: The best way is to let us know and file an issue. Filed https://github.com/Kotlin/kotlinx.coroutines/issues/2367

PR #2319, opened 2020-10-19
Fix potential crashes in asFlow extensions for RxJava 2&3 integrations

### LINKED ISSUE #2299: ObservableSource.asFlow() extension should at least surround sendBlocking() call with try/catch block
I work out of an Android code base that is mostly Rx-based. However, we have started writing new code using Coroutines. The `kotlinx-coroutines-rx2` library is crucial for interfacing between the two libraries. However, `ObservableSource.asFlow()` has proven to be dangerous and has introduced crashes for us in production.

Please see https://github.com/Kotlin/kotlinx.coroutines/issues/2104. I have shared several stack traces in that issue we have been seeing. The root cause seems to correspond with https://github.com/Kotlin/kotlinx.coroutines/issues/974. However, it seems that effort is going to take some careful redesign. In the meantime, can we consider wrapping the call to `sendBlocking()` with a `try/catch` block? I'm thinking the `catch` block can specifically catch `CancellationException` or the appropriate subtypes, e.g., `JobCancellationException` and `ChildCancelledException` to avoid catching all `Exceptions`.

LouisCAD:
Fixes #2299 

Related to issue #974

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
@@ -84,7 +84,7 @@ public fun <T: Any> ObservableSource<T>.asFlow(): Flow<T> = callbackFlow {
         override fun onNext(t: T) {
             try {
                 sendBlocking(t)
-            } catch (ignored: Exception) { //TODO: Replace when this issue is fixed: https://github.com/Kotlin/kotlinx.coroutines/issues/974
+            } catch (ignored: Throwable) { //TODO: Replace when this issue is fixed: https://github.com/Kotlin/kotlinx.coroutines/issues/974
             }
         }
         override fun onError(e: Throwable) { close(e) }
--- a/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
@@ -81,7 +81,12 @@ public fun <T: Any> ObservableSource<T>.asFlow(): Flow<T> = callbackFlow {
     val observer = object : Observer<T> {
         override fun onComplete() { close() }
         override fun onSubscribe(d: Disposable) { if (!disposableRef.compareAndSet(null, d)) d.dispose() }
-        override fun onNext(t: T) { sendBlocking(t) }
+        override fun onNext(t: T) {
+            try {
+                sendBlocking(t)
+            } catch (ignored: Exception) { //TODO: Replace when this issue is fixed: https://github.com/Kotlin/kotlinx.coroutines/issues/974
+            }
+        }
         override fun onError(e: Throwable) { close(e) }
     }
 
--- a/reactive/kotlinx-coroutines-rx2/test/ObservableSourceAsFlowStressTest.kt
+++ b/reactive/kotlinx-coroutines-rx2/test/ObservableSourceAsFlowStressTest.kt
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.rx2
+
+import io.reactivex.*
+import io.reactivex.exceptions.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+import org.junit.*
+import java.util.concurrent.*
+
+class ObservableSourceAsFlowStressTest : TestBase() {
+
+    private val maxIterations = 25
+    private val collectorJobCancelDelay = 1_000L
+    private val nextIterationDelay = collectorJobCancelDelay * 2
+
+    private var jobCancelledException: Throwable? = null
+    private val exceptionHandler = { throwable: Throwable ->
+        jobCancelledException = extractJobCancelledException(throwable)
+    }
+
+    @Before
+    fun setup() {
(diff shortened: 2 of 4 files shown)


mhernand40:
Thanks for taking this. May I request that the fix also be applied to Rx2 as well please? 🙂

LouisCAD:
Don't you want to do it now that you've seen my example, @mhernand40?

mhernand40:
@LouisCAD Typically I've seen bug fixes for Coroutines <-> Reactive APIs applied all throughout where applicable in the same PR. Hence why I am asking it be addressed within this PR. But I am also okay with tackling it in a separate PR provided this one gets approved first. Anyhow, FWIW I personally would be grateful if it would be addressed in this PR. 🙂

LouisCAD:
Done.

reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
```diff
@@ -81,7 +81,12 @@ public fun <T: Any> ObservableSource<T>.asFlow(): Flow<T> = callbackFlow {
     val observer = object : Observer<T> {
         override fun onComplete() { close() }
         override fun onSubscribe(d: Disposable) { if (!disposableRef.compareAndSet(null, d)) d.dispose() }
-        override fun onNext(t: T) { sendBlocking(t) }
+        override fun onNext(t: T) {
+            try {
+                sendBlocking(t)
+            } catch (ignored: Exception) { //TODO: Replace when this issue is fixed: https://github.com/Kotlin/kotlinx.coroutines/issues/974
```
mhernand40, line 87: Do we want to catch all `Exceptions` or should we be more specific, e.g., `CancellationException`, `ClosedSendChannelException`?

LouisCAD: I'm following the example in the KDoc of `callbackFlow`. It's unlikely (impossible?) this call will throw anything else anyway.

qwwdfsad: It can throw `Throwable` or `Error` if the channel was closed with it as a cause. It should generally catch `Throwable`

qwwdfsad, line 87: Please add a corresponding unit-test (both to rx2 and rx3) that fails if this change is reverted.

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
LouisCAD: Adding a test for this kind of race condition is not trivial… I mean, I have no clue of a way to implement it reliably.

If you know a way, please commit it to this branch, the "Allow edit from maintainers" checkbox is checked.

LouisCAD: Ah yes, missed the `close(e)` call that has the `Throwable` type.

elizarov: For race conditions like this, we are writing stress tests that fail at least sometimes and run them nightly with increased duration. We name them ending with `StressTest`. You can check out examples.

drinkthestars: Been able to repro the consequence of the throwing `sendBlocking` via this stress test: https://gist.github.com/drinkthestars/4fa82b744a623612b481c05efeacd812

The crux of the test is this:
``` Kotlin
val outerObservable = Observable.interval(1L, TimeUnit.MILLISECONDS)
val innerObservable = Observable.interval(1L, TimeUnit.MILLISECONDS)

val job = launch(newSingleThreadContext("Collector Thread")) {
   outerObservable
      .asFlow()
      .flatMapLatest { // The collect cancellations caused by this trigger the race condition
         innerObservable.asFlow()
      }
      .collect { delay(200) }
}
```

LouisCAD: @drinkthestars Would you send a PR with that stress test at the right location against the branch of this PR in my fork?

I'll then merge it so it lands in this PR, and keeps the attribution to you.

drinkthestars: Will do.

drinkthestars: @LouisCAD Check out https://github.com/LouisCAD/kotlinx.coroutines/pull/1. Let me know if there are any issues, thanks!

LouisCAD: LGTM but I didn't test locally.

Forwarding that without the fix of this PR, these stress tests I just merged would fail somewhere between the iterations 1 and 7.

mhernand40:
I'm really hoping this makes it into the final Coroutines `1.4` release. We had a major incident recently due to this issue where app crashes surged in production and we ultimately had to push out a hot fix. I was planning in the meantime to provide our own internal implementation of `ObservableSource.asFlow()` which would pretty much be a copy/paste of this implementation but I am hoping that updating to `1.4` will be all we need. 🙂

qwwdfsad:
I've opened https://github.com/Kotlin/kotlinx.coroutines/pull/2333 where I squashed all the changes into a single commit and added stable and small reproducer.

Thanks for taking care of this issue!

PR #2323, opened 2020-10-21
Release intercepted SafeCollector when onCompletion block is done

qwwdfsad:
Unfortunately, I wasn't able to find any meaningful scenario (=> write a test) when this actually leads to a memory leak for the existing set of API.  But still, this is a time bomb waiting to explode


elizarov (changes requested):
Btw, I've also added a test (fails without a fix), please check it out

elizarov (approved):
🚀 Good to go

PR #2326, opened 2020-10-21
Add the model checking mode in Lincheck tests

ndkoval:


ndkoval:
Here is a nightly build on Linux: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStress/3179735

kotlinx-coroutines-core/build.gradle
```diff
@@ -205,12 +206,26 @@ task jvmStressTest(type: Test, dependsOn: compileTestKotlinJvm) {
     systemProperty 'kotlinx.coroutines.semaphore.maxSpinCycles', '10'
 }
 
+task jvmLincheckTest(type: Test, dependsOn: compileTestKotlinJvm) {
+    classpath = files { jvmTest.classpath }
+    testClassesDirs = files { jvmTest.testClassesDirs }
+    minHeapSize = '1g'
+    maxHeapSize = '10g' // we may need more space for building an interleaving tree in the model checking mode
```
elizarov, line 213: Min memory requirements for our teamcity agents are currently set to 6Gb. We need to investigate if it can be raised to > 10Gb without the risk of losing agents in the pool.

ndkoval: 6Gb is enough, change the `maxHeapSize` correspondingly and checked on [a Linux nightly build](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStress/3180685?buildTab=tests).

kotlinx-coroutines-core/common/src/internal/Symbol.kt
```diff
@@ -10,7 +10,7 @@ package kotlinx.coroutines.internal
  * @suppress **This is unstable API and it is subject to change.**
  */
 internal class Symbol(val symbol: String) {
-    override fun toString(): String = symbol
+    override fun toString(): String = "<$symbol>"
```
elizarov, line 13: Why this is needed?

kotlinx-coroutines-core/build.gradle
```diff
@@ -205,12 +219,26 @@ task jvmStressTest(type: Test, dependsOn: compileTestKotlinJvm) {
...
+    systemProperty 'kotlinx.coroutines.semaphore.maxSpinCycles', '1' // better for the model checking mode
+}
+
 task jdk16Test(type: Test, dependsOn: [compileTestKotlinJvm, checkJdk16]) {
     classpath = files { jvmTest.classpath }
     testClassesDirs = files { jvmTest.testClassesDirs }
     executable = "$System.env.JDK_16/bin/java"
     exclude '**/*LFStressTest.*' // lock-freedom tests use LockFreedomTestEnvironment which needs JDK8
-    exclude '**/*LCStressTest.*' // lin-check tests use LinChecker which needs JDK8
+    exclude '**/*LincheckTest.*' // Lincheck tests use LinChecker which needs JDK8
```
elizarov, line 241: That would also need renaming all the `xxxLCTest.kt` files to `xxxLinckeckTest.kt`. Actually, I suggest to revert the name change and leave all the classes named as `xxxLCTest`, using `LC` as an abbreviation for `Linckeck`. Otherwise, it is somewhat inconvenient, since the first time you open the source code you get `"Typo: In word 'Lincheck'` warning.

kotlinx-coroutines-core/build.gradle
```diff
@@ -177,12 +177,26 @@ jvmTest {
...
+        // system properties for Lincheck and stress tests.
+        systemProperty 'java.security.manager', 'kotlinx.coroutines.TestSecurityManager'
+    }
     // 'stress' is required to be able to run all subpackage tests like ":jvmTests --tests "*channels*" -Pstress=true"
     if (!Idea.active && rootProject.properties['stress'] == null) {
+        exclude '**/*LincheckTest.*'
         exclude '**/*StressTest.*'
     }
-    systemProperty 'kotlinx.coroutines.scheduler.keep.alive.sec', '100000' // any unpark problem hangs test
+    if (Idea.active) {
```
elizarov, line 191: There is a separate `jvmLincheckTest` task that one can also use from IDEA. Such extensive tweaking of options for the regular `jvmTest` is not a good idea, as makes figuring out what's wrong with regular tests harder. I suggest to simply exclude lincheck tests from `jvmTest` configuration.

elizarov (changes requested):

kotlinx-coroutines-core/build.gradle
ndkoval: I would prefer not to revert this change since the names were `xxxLCStressTest.kt` and we have stress testing and model checking modes in Lincheck -- this overlap in the "stress" word usages confuses a bit.

kotlinx-coroutines-core/build.gradle
ndkoval: It is too inconvenient if you cannot simply press the "run" button on your lincheck test. Moreover, using this special `jvmLincheckTest` from IDEA does not seem obvious for me.

ndkoval: The only real problem is that we have to switch off our custom security manager for testing. Another option would be allowing it to set the required for Lincheck system properties.

kotlinx-coroutines-core/common/src/internal/Symbol.kt
ndkoval: This is just more readable, to distinguish symbols from strings/enums

.idea/dictionaries/ndkoval.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="ndkoval">
```
elizarov, line 2: Please, give it some neutral name (like `shared`).

kotlinx-coroutines-core/build.gradle
```diff
@@ -184,13 +184,30 @@ jvmTest {
...
+        // Configure the IDEA runner for Lincheck and stress tests
+        // TODO this can be removed once IDEA is allowed to use `jvmLincheckTest` and
+        // TODO `jvmStressTest` tasks for the corresponding tests instead of using `jvmTest`.
+        jvmArgs = ['--add-opens', 'java.base/jdk.internal.misc=ALL-UNNAMED',   // required for transformation
+                   '--add-exports', 'java.base/jdk.internal.util=ALL-UNNAMED'] // in the model checking mode
+        maxHeapSize = '6g' // we need more memory for model checking
+        systemProperty 'kotlinx.coroutines.semaphore.segmentSize', '2' // better for testing in general
+        systemProperty 'kotlinx.coroutines.semaphore.maxSpinCycles', '1' // better for model checking
+        systemProperty 'kotlinx.coroutines.scheduler.keep.alive.sec', '100000' // any unpark problem hangs test
```
elizarov, line 209: Please, extract this configuration into a separate method. There's too much duplication.

elizarov (changes requested):

.idea/dictionaries/ndkoval.xml
ndkoval: In IDEA each user has its own dictionary to reduce merge conflicts with the global dictionary, but all these per-user dictionaries are used by IDEA independently on the current user.

kotlinx-coroutines-core/build.gradle
ndkoval: Fixed.

.idea/dictionaries/ndkoval.xml
elizarov: That's exactly my point. The dictionary that is committed to the report should not be your personal dictionary that is tied to your user-name, it shall be a common dictionary that is shared across the project.

ndkoval: That is not how IntelliJ IDEA works (yes, the behavior is counterintuitive). You cannot create a dictionary with a custom name -- it will be automatically cleaned up by IntelliJ IDEA. Instead, there are per-user dictionaries which can be updated by each user personally -- when you use IntelliJ IDEA **all** these per-user dictionaries are used, so that your IDEA will know about the words listed in `ndkoval` dictionary. The original motivation (provided by the IntelliJ IDEA team) is that in this case you will not have merge conflicts on the dictionary since it is split by users. I, personally, would not agree with this solution (at least, I spent an hour trying to understand the behavior), but that is how it works.

ndkoval: In conclusion,
1) you cannot create dictionaries with custom names, and, in particular, a single global dictionary
2) all per-user dictionaries are used in IDEA, independently of the current user

elizarov: Of course, you can. I've committed a fix.

elizarov (approved):
Looks good now (I've made a few extra tweaks). I'll wait for CI and merge.

PR #2329, opened 2020-10-22
Propagate exception from stateIn to the caller when the upstream fail…

qwwdfsad:
…ed to produce initial value


elizarov (approved):
👍 Good catch. It makes total sense.

PR #2331, opened 2020-10-23
Fix BlockHound false positives

dkhalanskyjb:


kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
```diff
@@ -1,16 +1,53 @@
...
 
 @Suppress("UNUSED")
-public class CoroutinesBlockHoundIntegration: BlockHoundIntegration {
+public class CoroutinesBlockHoundIntegration : BlockHoundIntegration {
 
-    override fun applyTo(builder: BlockHound.Builder) {
-        builder.addDynamicThreadPredicate { isSchedulerWorker(it) }
-        builder.nonBlockingThreadPredicate { p -> p.or { mayNotBlock(it) } }
+    override fun applyTo(builder: BlockHound.Builder): Unit = with(builder) {
+        // These classes use a lock internally, but should be safe to use.
```
qwwdfsad, line 13: I would rather write something a bit more encouraging for external users such as "These classes use a lock internally, but only for well-known, short and isolated state transitions that are known to be short and fast"

kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
```diff
@@ -1,16 +1,53 @@
...
+        for (method in listOf(
```
qwwdfsad, line 14: Currently, it's a bit hard to read and get through. Could you please split it to a few methods?
Like
```
allowBlockingCallsInsideChannelImpl()
allowBlockingCallsInsideReflection()
... etc. ...
```

qwwdfsad (changes requested):
Please revisit a few more places on our code that rely on fast locking:
- `StateFlow` and `SharedFlow`
- `JobSupport` (`finalizeFinishingState` & co.)

dkhalanskyjb:
Yeah, I completely missed that some code is protected using `synchronized`.

kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
```diff
@@ -1,16 +1,145 @@
...
+    /** Allows blocking inside [kotlinx.coroutines.channels.ConflatedChannel]. */
+    private fun BlockHound.Builder.allowBlockingCallsInConflatedChannel() {
+        for (method in listOf("offerInternal", "offerSelectInternal", "pollInternal", "pollSelectInternal",
+            "onCancelIdempotent"))
+        {
+            allowBlockingCallsInside("kotlinx.coroutines.channels.ConflatedChannel", method)
+        }
+    }
+
+    /** Allows blocking when enqueuing tasks into a thread pool.
```
qwwdfsad, line 106: In our codestyle, we typically start comment on the next line after `/**`

qwwdfsad (approved):
Otherwise looks good, great job

PR #2332, opened 2020-10-23
Propagate kotlin_snapshot_version to buildSrc and kts files

qwwdfsad:


settings.gradle
```diff
@@ -12,6 +12,12 @@ pluginManagement {
     }
 }
 
+// Propagate this property to buildSrc via the only possible way: system properties
```
h0tk3y, line 15: Wait, how is `kotlin_snapshot_version` propagated to `buildSrc`, then?

h0tk3y (approved):

PR #2336, opened 2020-10-26
Add debounce selector

### LINKED ISSUE #1216: Add debounce operator with selector to Flow

elizarov:

Resolves #1216.

Based on PR #2314

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -112,10 +112,10 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
  * @param timeoutMillis [T] is the emitted value and the return value is timeout in milliseconds.
  */
 @FlowPreview
+@OptIn(kotlin.experimental.ExperimentalTypeInference::class)
+@OverloadResolutionByLambdaReturnType
 public fun <T> Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T> =
-    debounceInternal { emittedItem ->
-        timeoutMillis(emittedItem)
-    }
+    debounceInternal(timeoutMillis)
 
 /**
  * Returns a flow that mirrors the original flow, but filters out values
@@ -151,14 +151,15 @@ public fun <T> Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T> =
  */
 @ExperimentalTime
 @FlowPreview
-public fun <T> Flow<T>.debounceDuration(timeout: Duration): Flow<T> = debounce(timeout.toDelayMillis())
+public fun <T> Flow<T>.debounce(timeout: Duration): Flow<T> =
+    debounce(timeout.toDelayMillis())
 
 /**
  * Returns a flow that mirrors the original flow, but filters out values
  * that are followed by the newer values within the given [timeout].
  * The latest value is always emitted.
  *
- * A variation of [debounceDuration] that allows specifying the timeout value dynamically.
+ * A variation of [debounce] that allows specifying the timeout value dynamically.
  *
  * Example:
  *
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-core/common/test/flow/operators/DebounceTest.kt
```diff
@@ -223,4 +221,102 @@ class DebounceTest : TestBase() {
...
+        }
+
+        expect(2)
+        val result = flow.debounce { 0 }.toList()
+
+        assertEquals(listOf("A", "B"), result)
+        finish(5)
+    }
+
+    @ExperimentalTime
```
qwwdfsad, line 292: [code style] either file-level opt-in or project-level (for test source sets only) opt-in should be used

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,104 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+ *
+ * produces the following emissions
+ *
+ * ```text
+ * 1, 3, 4, 5
+ * ```
+ * <!--- TEST -->
+ *
+ * Note that the resulting flow does not emit anything as long as the original flow emits
+ * items faster than every [timeout] milliseconds.
```
qwwdfsad, line 195:
```suggestion
 * items faster than every [timeout] unit.
```

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,104 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+ *
+ * @param timeout [T] is the emitted value and the return value is timeout in [Duration].
```
qwwdfsad, line 197: Should mention that it has millisecond resolution and that positive durations are coerced to at least one millisecond

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,104 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+public fun <T> Flow<T>.debounce(timeout: (T) -> Duration): Flow<T> =
+    debounceInternal { emittedItem ->
+        timeout(emittedItem).toDelayMillis()
+    }
+
+private fun <T> Flow<T>.debounceInternal(timeoutMillisSelector: (T) -> Long) : Flow<T> =
+    scopedFlow { downstream ->
+        // Produce the values using the default (rendezvous) channel
+        // Note: the actual type is Any, KT-30796
+        val values = produce<Any?> {
```
qwwdfsad, line 213: While we are here, should be `Any`, KT-30796 is fixed

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -129,7 +151,104 @@ public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
...
+                // Set timeout when lastValue exists and is not consumed yet
+                if (lastValue != null) {
+                    onTimeout(timeoutMillis) {
+                        downstream.emit(NULL.unbox(lastValue))
+                        lastValue = null // Consume the value
+                    }
+                }
+                // Should be receiveOrClosed when boxing issues are fixed
+                values.onReceiveOrNull { value ->
+                    if (value == null) {
```
qwwdfsad, line 242: Assignment can be lifted out `lastValue = if (...)`

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -64,37 +64,59 @@ fun main() = runBlocking {
...
+ *
+ * Note that the resulting flow does not emit anything as long as the original flow emits
+ * items faster than every [timeoutMillis] milliseconds.
+ *
+ * @param timeoutMillis [T] is the emitted value and the return value is timeout in milliseconds.
+ */
+@FlowPreview
+@OptIn(kotlin.experimental.ExperimentalTypeInference::class)
+@OverloadResolutionByLambdaReturnType
+public fun <T> Flow<T>.debounce(timeoutMillis: (T) -> Long): Flow<T> =
```
qwwdfsad, line 117: What happens when the timeout is negative should be documented.
Also, lambda parameter can be named: `(value: T) -> Long`

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -64,37 +64,59 @@ fun main() = runBlocking {
  */
 @FlowPreview
 public fun <T> Flow<T>.debounce(timeoutMillis: Long): Flow<T> {
```
qwwdfsad, line 66: Behaviour for zero timeout should be documented as well, especially because we are changing it

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
elizarov: Added docs. However, we don't usually add `value` name to such lambdas in Flow operators (see `filter`, `map`, etc)

kotlinx-coroutines-core/common/test/flow/operators/DebounceTest.kt
elizarov: In fact, in other tests we use test-class-level annotation.

qwwdfsad (approved):
Nice!

PR #2339, opened 2020-10-27
Fix in Global coroutines are like daemon threads doc

Husseinhj:
The `delay` with 1300 millisecond time print [following results](https://pl.kotl.in/tpgnfGR2f):

```
I'm sleeping 3 ... I'm sleeping 0 ...
I'm sleeping 1 ...
I'm sleeping 2 ...
``` 

When changing this to the 1000 millisecond it will show the [correct output](https://pl.kotl.in/dJ437pRYI) like the documentation:

```
I'm sleeping 0 ...
I'm sleeping 1 ...
I'm sleeping 2 ...
```

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -373,7 +373,7 @@ fun main() = runBlocking {
             delay(500L)
         }
     }
-    delay(1300L) // just quit after delay
+    delay(1000L) // just quit after delay
 //sampleEnd    
 }
 ```


elizarov:
The code is correct. There is some problem in Kotlin playground. Thanks for noticing it. I've reported it here: https://youtrack.jetbrains.com/issue/KTL-182

PR #2342, opened 2020-10-27
Fixes for breaking changes in 1.4.20

qwwdfsad:


reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
```diff
@@ -41,7 +39,10 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
  * immediately resumes with [CancellationException].
  */
 @Suppress("UNCHECKED_CAST")
-public suspend fun <T> MaybeSource<T>.await(): T? = (this as MaybeSource<T?>).awaitOrDefault(null)
+public suspend fun <T> MaybeSource<T>.await(): T? {
+    val result = (this as MaybeSource<Any>).awaitOrDefault(MISSING_VALUE)
+    return if (result === MISSING_VALUE) null else result as T
```
elizarov, line 44: What's the purpose of this change? `MaybeSource` is supposed to be null-hostile and cannot return produce `null` value.

reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
```diff
@@ -52,12 +53,23 @@ public suspend fun <T> MaybeSource<T>.await(): T? = (this as MaybeSource<T?>).aw
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * immediately resumes with [CancellationException].
  */
-public suspend fun <T> MaybeSource<T>.awaitOrDefault(default: T): T = suspendCancellableCoroutine { cont ->
+public suspend fun <T : Any> MaybeSource<T>.awaitOrDefault(default: T): T = suspendCancellableCoroutine { cont ->
```
elizarov, line 56: Haven't we decide to leave them as extensions on `T : Any?` for the sake of everybody's convenience? With the changes in this file all the generic usages for these extensions will have to be updated.

elizarov (changes requested):

reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
qwwdfsad: Yes, but because of its nature, it can emit nothing and just complete. In such a situation, we return `null`.

PR #2358, opened 2020-11-02
Add JVM IR target in build_snapshot_train mode

ALikhachev:


udalov:
As in https://github.com/ktorio/ktor/pull/2175#issuecomment-721058147, I suggest not to commit API dumps for JVM IR, but rather mute the tests if possible (with the option to unmute locally).

build.gradle
```diff
@@ -262,11 +302,18 @@ configure(subprojects.findAll {
 }) {
     // Pure JS and pure MPP doesn't have this notion and are configured separately
     // TODO detect it via platformOf and migrate benchmarks to the same scheme
-    sourceSets {
-        main.kotlin.srcDirs = ['src']
-        test.kotlin.srcDirs = ['test']
-        main.resources.srcDirs = ['resources']
-        test.resources.srcDirs = ['test-resources']
+    kotlin {
```
qwwdfsad, line 305: Why this change is require?

qwwdfsad (changes requested):
I've skimmed through PR and have the following comments that should be applied to all PRs in libraries:

1) I see a lot of irrelevant changes in gradle configurations. Either they should be avoided (-> reverted or amended), or a comment explaining why this change was done is needed. Now it's far from being clear, neither from the change nor from commit messages.

2) Api dumps should not be touched. They should be left as-is for JVM and be ignored for JVM IR

3) Subtle places (e.g.: `attributesSchema`, `jvm-multiplatform` etc.) should be added with a comment

3) Can we avoid migration to MPP plugin in JVM-only project? This change is not really desirable

ALikhachev:
During private discussion we've decided to try another way. I'll open new PR.

Issue #1679, opened 2019-11-28
Error with infinite awaiting of execution of runBlocking


evgzakharov:
I have find an infinite awaiting of execution of `runBlocking`.
Minimal reproducible code:

```
import kotlinx.coroutines.CompletableDeferred
import kotlinx.coroutines.async
import kotlinx.coroutines.delay
import kotlinx.coroutines.isActive
import kotlinx.coroutines.launch
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.withTimeoutOrNull
import org.junit.jupiter.api.Test
import kotlin.coroutines.coroutineContext

class BugCheck {
    @Test
    fun `check new player is created successfully`() = runBlocking {
        val deferred = CompletableDeferred<Unit>()
        val completed = CompletableDeferred<Boolean>()
        val leagueTask = async { leagueCheck(deferred) }

        deferred.await()
        launch { playerCheck(completed) }

        leagueTask.await()
        completed.complete(true)
        println("stop")
        Unit
    }

    private suspend fun playerCheck(completed: CompletableDeferred<Boolean>) {
        executeOnceTest {
            while (coroutineContext.isActive && !completed.isCompleted) {
                println("...")
                delay(100)
            }

            true
        }
    }

    private suspend fun leagueCheck(deferred: CompletableDeferred<Unit>) {
        deferred.complete(Unit)
//        delay(1) //uncomment and all works fine

        executeOnceTest {
            println("inner done")
            true
        }
        println("555")

        delay(1000)
    }

    private fun executeOnceTest(scenario: suspend () -> Boolean): Unit = runBlocking {
        withTimeoutOrNull(10000) {  //or make timeout 1000
            scenario()
        }

        runBlocking {  // freeze on this
            println("test")
        }
    }
}
```

If you running test, code will never complete. Execution freeze on runBlocking right after `withTimeout`. But if uncomment `delay(1)` all will works fine.  Make timeout smaller to 1000, works too. 

kotlin 1.3.61
kotlin-coroutines 1.3.2
gradle 6.0.1

qwwdfsad:
Could you please verify your example one more time? I've copypasted it as is and it seems to work properly with Kotlin 1.3.61 and coroutine 1.3.2

evgzakharov:
Yes, it reproduce in jvm 8, 12, 13. But sometimes version without `delay(1)` complete after 5+ seconds. But if uncomment line with delay code will complete very fast. 
[...]

PR #2374, opened 2020-11-08
Fix forever lock of main thread in runBlocking

slavonnet:
In havy concurect UI load where in views have many small runBlocking (and withContext) blocks some time we can get parkNanos == Long.MAX_VALUE.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -68,6 +68,15 @@ import kotlin.native.concurrent.*
  * the `onBufferOverflow` parameter, which is equal to one of the entries of the [BufferOverflow] enum. When a strategy other
  * than [SUSPENDED][BufferOverflow.SUSPEND] is configured, emissions to the shared flow never suspend.
  *
+ * ### Unbuffered shared flow
+ *
+ * A default implementation of a shared flow that is created with `MutableSharedFlow()` constructor function
+ * without parameters has no replay cache nor additional buffer.
+ * [emit][MutableSharedFlow.emit] call to such a shared flow suspends until all subscribers receive the emitted value
+ * and returns immediately if there are no subscribers.
+ * Thus, [tryEmit][MutableSharedFlow.tryEmit] call succeeds and returns `true` only if
+ * there are no subscribers (in which case the emitted value is immediately lost).
+ *
  * ### SharedFlow vs BroadcastChannel
  *
  * Conceptually shared flow is similar to [BroadcastChannel][BroadcastChannel]
--- a/kotlinx-coroutines-core/jvm/src/Builders.kt
+++ b/kotlinx-coroutines-core/jvm/src/Builders.kt
@@ -8,7 +8,6 @@
 
 package kotlinx.coroutines
 
-import java.util.concurrent.locks.*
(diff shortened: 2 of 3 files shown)


slavonnet:
https://youtrack.jetbrains.com/issue/KT-43244

slavonnet:
@elizarov can you review please? :)

slavonnet:
maybe fix #1679

elizarov (changes requested):
First and foremost, this PR needs a test that it supposedly fixes. Also, looking at the code I see (wihtout any test) that this code will be consuming 100% CPU time when there are no tasks for `runBlocking { ... }` to execute, since, in this case, `nanos == Long.MAX_VALUE` and this change makes `parkNanos` a noop that immediately returns instead of waiting (without consuming CPU time) as it should be.

slavonnet:
There is a difficulty ... I cannot write a test. I know the result of a long blocking, but how quickly to get an empty queue, but with a working coroutine there are only guesses ..

I repeat (rarely) when creating hundreds of custom views (form fields) where there is runBlocking in the init

slavonnet:
fixed

slavonnet (commented):
Code more clean

elizarov (changes requested):
All I see from this code is that it is attempting to shuffle some problems under the rug by replacing a proper "sleep until woken up" with "sleep at most 10ms", instead of figuring out what's going on and fixing the problem. Without a test, it is not even possible to very that it helps in any sense.

slavonnet:
@elizarov Я попробую, но ты не прав. Ограничения на главном потоке более важны и критичны чем текущий бfu. Это ограничение должно проверяться независимо от текущего вопросы. Главный поток должен просыпаться внутри фрейма чтобы на главном Лупере отрисовать фрейм..  Если по какой-то причине след событие далеко - не забываем  запускать фрейм. ю 

И что плохого если глобальное у и логичное условие не дает права на такую ошибку в будущем? Почему рельзчя глобально исправить кейс?

slavonnet:
I wrote a draft that shows deadlocks in different places. Results vary from system to system.
Please manual run file from 
https://github.com/Kotlin/kotlinx.coroutines/pull/2428

slavonnet:
Looks like queue overload somewhere

image
image

qwwdfsad (commented):
I've studied tests from the https://github.com/Kotlin/kotlinx.coroutines/pull/2428.

The very first hang test can be boiled down to a classic deadlock:
```
val executor = newFixedThreadPoolContext(1, "")
runBlocking(executor) {
    withContext(executor) {
}
```

It's not a bug in `kotlinx.coroutines`, but rather a classic deadlock when trying to submit two tasks into single-threaded executor that depend on each other.
The rest of the test timeouts because the executors from the previous tests were not reset and still hang on coroutines from the previous test runs.

Neither PR nor tests seem to be constructive or respectful, so I'm closing this PR.
You can create a separate issue with a stable reproducer that has the self-contained snippet of code, expected behaviour and actual behaviour description. Please note that just giving us 625 different test configurations that fail for some reason without further analysis is not helpful and just waste maintainers time. I also suggest getting familiar with [contribution guide](https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md)

The official language of the Kotlin org is English, any further comments using non-English language will be removed

PR #2376, opened 2020-11-09
Add lint warnings on SharedFlow operations that never complete

### LINKED ISSUE #2340: Flow toList could cause infinite suspension
In the rather heated discussion in a Telegram chat, there were complaints that calling toList() function on an infinite Flow (like the StateFlow for example) could cause an infinite suspension, therefore, producing a rather hard to catch bug.

For example, this code will block indefinitely:
```kotlin
import kotlinx.coroutines.*
fun main() {
    var flow = MutableStateFlow(0)
    runBlocking{
      val list = flow.toList()
    }
}
```

There are several solutions to that problem:
* Introduce a default large limit to this operation and throw an error if the limit is exceeded.
* Throw error immediately for specific types of Flow which are known to be infinite.
* Update documentation to warn people that using terminal operations on infinite flows could lead to infinite suspensions.

All those solutions could be applied together.

The positive idea from the same discussion. It is possible to introduce something like

```kotlin
inline class LimitedFlow(val flow: Flow): Flow by flow
```
And define dangerous terminal operations on it. In this case operations like limit would provide type-safe wrappers. And the developer could explicitly declare the Flow as limited manually via something like `Flow::asLimited`.

### LINKED ISSUE #2368: SharedFlow didn't cancel or throw exception with callbackFlow
The exception didn't throw to parent scope or cancel itself while using SharedFlow with callbackFlow. I tried to cancel sharedflow in case of failure. Here is my code snippet  

```
fun adapt(call: Call<T>): SharedFlow<T> {
        val scope = CoroutineScope(dispatcher)
        return callbackFlow<T> {
            call.enqueue(object : Callback<T> {
                override fun onResponse(call: Call<T>, response: Response<T>) {
                    sendBlocking(response.body())
                    close()
                }

                override fun onFailure(call: Call<T>, t: Throwable) {
                    close(t)
                }
            })
            awaitClose()
        }.catch {
            /**
             *Here caught exception that threw in Failure case
             * Try to cancel Job so SharedFlow can terminate but It didn't work However catching exception here won't crash my app
             */
            scope.cancel()
        }.shareIn(scope, SharingStarted.Lazily, 1)
    }
```
I used it
```
runBlocking{

//this will get first output in case of success and then completed this coroutine
//But In case of Failure, this will always wait to get the first item and never completed

    val result = adapt(someCall).first()
}
```
Here is another way tried to accomplish the same thing

```
fun adapt(call: Call<T>): SharedFlow<T> {
        val scope = CoroutineScope(dispatcher)
        return callbackFlow<T> {
            call.enqueue(object : Callback<T> {
                override fun onResponse(call: Call<T>, response: Response<T>) {
                    sendBlocking(response.body())
                    close()
                }

                override fun onFailure(call: Call<T>, t: Throwable) {
                    close(t)
                }
            })
            awaitClose()
        }.shareIn(scope, SharingStarted.Lazily, 1).catch {
            /**
             *This catch block never invoked in case of failure
             * Try to cancel Job so sharedFlow can terminate its coroutine
             */
              currentCoroutineContext().cancel()
        }
    }
```
I used this as previous but this time
```
runBlocking{
//this will get first output in case of success and then completed this coroutine
//But In case of Failure App crashes
    val result = adapt(someCall).first()

}
```
sharedflow.collect|first()|take(1){}  stays always in resume state I want to cancel it in case of failure.

elizarov:
Fixes #2340
Fixes #2368

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
@@ -2,12 +2,13 @@
  * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-@file:Suppress("unused")
+@file:Suppress("unused", "INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 
 package kotlinx.coroutines.flow
 
 import kotlinx.coroutines.*
 import kotlin.coroutines.*
+import kotlin.internal.InlineOnly
 
 /**
  * Applying [cancellable][Flow.cancellable] to a [SharedFlow] has no effect.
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -79,4 +80,61 @@ public fun FlowCollector<*>.cancel(cause: CancellationException? = null): Unit =
...
+    message = "SharedFlow never completes, so this operator has no effect.",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("this")
+)
+@InlineOnly
+public inline fun <T> SharedFlow<T>.catch(noinline action: suspend FlowCollector<T>.(cause: Throwable) -> Unit): Flow<T> =
+    (this as Flow<T>).catch(action)
+
+@Deprecated(
+    message = "SharedFlow never completes, so this operator has no effect.",
```
slavonnet, line 95: maybe text ressource move to const? Any optimize  f for equal annotation dubles?

elizarov: What do you mean?

slavonnet: No way to reduce copy-paste / duplicate code / messages?

elizarov: There is no reason to. The code is easier to read this way, with messages right here.

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -79,4 +80,61 @@ public fun FlowCollector<*>.cancel(cause: CancellationException? = null): Unit =
...
+    replaceWith = ReplaceWith("this")
+)
+@InlineOnly
+public inline fun <T> SharedFlow<T>.retryWhen(noinline predicate: suspend FlowCollector<T>.(cause: Throwable, attempt: Long) -> Boolean): Flow<T> =
+    (this as Flow<T>).retryWhen(predicate)
+
+@Deprecated(
+    message = "SharedFlow never completes, so this terminal operation never completes.",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("awaitCancellation()")
```
qwwdfsad, line 118: `awaitCancellation` seems to be unintended

elizarov: I really cannot figure out a better replacement. We can leave it without a replacement, though.

qwwdfsad (approved):
Nice precaution with `InlineOnly`.
Could drop a comment `// Error in 1.5.0` so we won't forget to do it

Issue #1302, opened 2019-06-27
Flow buffer operator that buffers elements within specified time range


gmarchelos:
Currently Flow support only buffer operator with capacity.
It would be useful to buffer elements within specified time range.
```flow.buffer(Duration.ofSeconds(5)).collect {...}```

qwwdfsad:
Could you please explain a bit your use-case? E.g. what kind of domain-specific problem you would like to solve with such operator

fvasco:
I need to batch live events to reduce synchronization cost and limit refresh rate.

Currently I implemented this feature in my project: events are sent to a channel and batched to 10Hz in an event bus.

gmarchelos:
I have flow of trades and I need to collect trades within specified timespan (it's high probability for many trades to occur within 1-5 sec) and then do some processing of collected trades.
It's costly for me to process trades one by one so I would like to have buffer operator that collects events within specified timespan.

elizarov:
Thanks. This kind of operator would be called `chunked` in Kotlin lingo. Thanks for spelling out a use-case with a time-limit chunking. See #1290 for a size-limited use-case.

PaulWoitaschek:
I'm in need of this too. My use case is as following:

I implemented a barcode scanner which scans the camera input live. That means that I get barcode results every few milliseconds.
However the barcode implementation is not perfect and sometimes detects a wrong barcode.
For that case I want consume the barcode events chunked in 2 second windows and pick the barcode that was detected most frequently.

ps-feng:
@PaulWoitaschek I had a requirement for a similar (though not exactly the same operator) and ended up implementing [my own](https://dev.to/psfeng/a-story-of-building-a-custom-flow-operator-buffertimeout-4d95) (final implementation at the end), if it helps.

circusmagnus:
I have a need for a time based chunked too. I have a stream of analytics events originating in an mobile app. I would like to chunk them in time-based windows, so that I can send them to the backend periodically without initiating too many network connections. Say, chunk events for a minute, send them over the network and repeat for the next package. 

For this task I care for:
- chunks that are not empty. No events = no chunks, despite time window getting closed.
- not losing events
- perhaps ability to limit max size of a chunk, to not blow up memory in some corner cases

I do not care for:
- exact timing of chunks emissions.

I had a discussion earlier with @elizarov about size-based chunk operator:
[...]

Issue #1290, opened 2019-06-24
Flow.chunked operator with size limit


azulkarnyaev:
It would be useful to have an optional transformation in `Flow.buffer` method to aggregate the buffered items like in `kotlin.sequences.Sequence.chunked`.
I mean,
```fun buffer(capacity: Int = BUFFERED, suspend transform: (List<T>) -> R) : Flow<T>``` 

Then we can write
```
runBlocking {
        (1..100).asFlow().buffer(capacity = 10) { it.sum() }.collect { println(it) }
    }
```
with result 55, 155, 255, ... , 955

**Comment by elizarov** [comment]
That would be a separate operator(s) that we'll call `chunked`, `window` or something like that (and will generally follow stdlib naming convention). These operators will be totally unrelated to the `buffer` operator. Unlike `buffer` all those operators will be fully sequential.

P.S. Rx has a whole set of `bufferXxx` operators that actually correspond to `chunked`/`windowed` in Kotlin.   On the other hand, `buffer`/`conflate` operators in Kotlin flow somewhat correspond to Rx `onBackpressureBuffer` operators.

**Comment by elizarov** [comment]
Also, I've forgot to ask what be your use-case for such an operator?

**Comment by azulkarnyaev** [comment]
Thank you very much for the response!
About the use case: I need to handle stream of vectors (csv row) from a tcp socket and write aggregated statistics for every chunk of `n` messages to files.

**Comment by elizarov** [comment]
Are you sure you need `Flow` for this? Wouldn't a `Sequence` from Kotlin standard library work for you?

**Comment by azulkarnyaev** [comment]
Well, I'm using Ktor as a server for socket connection and using coroutines to write data to a file. Pseudo code for my task:  
```
launch {
    val socket = server.accept()
    val input = socket.openReadChannel()
    flow {
          while (true) {
                val line = input.readUTF8Line()
                emit(line)
           }
    }.map { 
          convertToDomainObject(it) 
    }. chunked(1000) {
          aggregateToDomainObjects(it)
    }.collect {
          writeToFile(it)
    }
}
```
Yes, I can use just sync `Sequence`. But then I need to provide a back pressure mechanism manually: what if I receive messages faster than I store them? Hopefully, `chunked()` method will be the same as `buffer()` and provide a way for back pressure out of the box.

elizarov:
@azulkarnyaev Thanks for explanation. It does make sense.

circusmagnus:
[...]

PR #2378, opened 2020-11-10
Flow size- and time-based chunked

circusmagnus:
Add two operators for size- and time-based chunking of Flows. Targets issues: https://github.com/Kotlin/kotlinx.coroutines/issues/1290
https://github.com/Kotlin/kotlinx.coroutines/issues/1302

New operators:
`public fun <T> Flow<T>.chunked(maxSize: Int, minSize: Int = 1)`

`public fun <T> Flow<T>.chunked(
    chunkDuration: Duration,
    minSize: Int = 1,
    maxSize: Int = NO_MAXIMUM
): Flow<List<T>>`

`public fun <T> Flow<T>.chunked(
    chunkDurationMs: Long,
    minSize: Int = 1,
    maxSize: Int = NO_MAXIMUM
): Flow<List<T>>`

Time based impl:
- By default starts chunking values with first emission since start of collection or last chunked emission (non-continuous time windows)
- With minSize = 0 switches to continuous time-windows: Starts chunking right away. New chunking window starts right after previous
- Can specify max size of a chunk. If specified, upon reaching max size, it emits chunked values and resets timer.

To be added:
- docs
- Annotations (@Experimental, etc.)


elizarov:
Please, please, read contributing guidelines first: https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md

If you introduce any new public APIs:
* All new APIs must come with documentation and tests.
* All new APIs are initially released with @ExperimentalCoroutineApi annotation and are graduated later.
* Update the public API dumps and commit the resulting changes as well. It will not pass the tests otherwise.
* **If you plan large API additions, then please start by submitting an issue with the proposed API design
to gather community feedback.** 
* Contact the maintainers to coordinate any big piece of work in advance.

elizarov (changes requested):
This PR does not follow contributing guidelines: it has no docs and it does not have a community-discussed design.

circusmagnus:
Sure. I have initiated discussion in https://github.com/Kotlin/kotlinx.coroutines/issues/1302#issuecomment-725721317

Commits with lacking docs, annotations and api dump will follow shortly.

circusmagnus:
I am converting this PR to Draft until we reach some outcome in #1302 discussion.

README.md
```diff
@@ -7,7 +7,7 @@
 [![Slack channel](https://img.shields.io/badge/chat-slack-green.svg?logo=slack)](https://kotlinlang.slack.com/messages/coroutines/)
```
circusmagnus, line 7: This file has been apparently changed on master but not on develop - with PR: Improve readability. #2563

circusmagnus:
I have:
- Updated the solution
- Added @Experimental annotations
- Updated ApiDump

nikiJava:
Hi :) Are there any updates? The operator looks very handy and useful.

gregsh:
1.
It looks to me that using `onTime` select clause in the provided implementation [chunked(ByTime)](https://github.com/Kotlin/kotlinx.coroutines/pull/2378/files#diff-97093ca1b59244a9a9796d9e3a5ceee18b5dc0dbcc4a87904f5e47b0c49452b8R129) is not efficient. It keeps executing even when the source flow is really idle.

For example, if one tries to "chunk" mouse clicks, the code in `onTimeout(3.seconds)` keeps running each 3 seconds regardless of whether user really clicks the mouse or just left the app running for a night.

2.
Also, in the scenario above we probably want to buffer events while user keeps clicking and emit that buffer only when user stops. I see 2 options how to "chunk" here. We may want to a) emit something right away and cancel later as `collectLatest` does (responsive UI) or b) just buffer until there is duration-long silence. There are no such options in this PR.

kotlinx-coroutines-core/common/src/flow/operators/Chunk.kt
```diff
@@ -0,0 +1,211 @@
...
+        requirePositive(intervalMs)
+        requirePositive(maxSize)
+    }
+
+    override fun <T> Flow<T>.chunk(): Flow<List<T>> = scopedFlow { downstream ->
+        val emitNowAndMaybeContinue = Channel<Boolean>(capacity = Channel.RENDEZVOUS)
+        val elements = produce<T>(capacity = maxSize) {
+            collect { element ->
+                val hasCapacity = channel.trySend(element).isSuccess
+                if (!hasCapacity) {
```
oguzhaneren, line 166: when maxSize is 1 , hasCapacity returns true so  emitNowAndMaybeContinue cannot trigger

oguzhaneren, line 166: what do you think about size instead of maxSize like SizedBased? when channel reached to size it should be emit

PR #2381, opened 2020-11-12
Prevent potential re-park when the unparking thread is too slow

qwwdfsad:
I've stressed my machine with our scheduler stress tests, both with and without a fix and it seems like the fix eliminates most of the hangs we observe on our CI from time to time.


elizarov (approved):
🎉 Very good.

PR #2382, opened 2020-11-12
Fix replacement for deprecated broadcastIn operator

### LINKED ISSUE #2344: broadcastIn's ReplaceWith is incorrect
I tried using the `ReplaceWith` for `broadcastIn` in 1.4.0 and it has several issues.

Firstly it doesn't handle `CoroutineStart` other than lazily which seems a bit dangerous as it might change behavior.

Secondly the order of the parameters is incorrect. In the `ReplaceWith` it's listed as `shareIn(scope, 0, SharingStarted.Lazily)` while the final version of `shareIn` has `scope, started, replay`. 

Lastly it was just broken:
```kotlin
someFlow
  .broadcastIn(viewModelScope)
// became
run {
  flow
  shareIn<Type>(viewModelScope, 0, SharingStarted.Lazily)
}
```

One solution would be to add a deprecated, single argument, overload for `broadcastIn` which is replaced with `shareIn(scope)` and the 2 argument overload can have its `ReplaceWith` removed since it cannot be done reliably.

qwwdfsad:
Fixes #2344

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Channels.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Channels.kt
@@ -201,7 +201,7 @@ public fun <T> BroadcastChannel<T>.asFlow(): Flow<T> = flow {
  */
 @Deprecated(
     message = "Use shareIn operator and the resulting SharedFlow as a replacement for BroadcastChannel",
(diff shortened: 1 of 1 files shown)


elizarov (approved):
👍

PR #2384, opened 2020-11-13
Conditionally create an instance of CancellationException in Channel.…

qwwdfsad:
…cancel()

Avoid creating costly exception when the channel is cancelled to save a few cycles when it's not necessary. Cancellation is heavy-enough when the channel is open, so the single check won't worsen it.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
@@ -635,6 +635,7 @@ internal abstract class AbstractChannel<E>(
         cancelInternal(cause)
 
     final override fun cancel(cause: CancellationException?) {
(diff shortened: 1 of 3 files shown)


elizarov (approved):
👍 Good stuff

paulferaud:
This changed the behavior for completed `callbackFlow` that are still blocked. Example:
```
  @Test
  fun p2384() = runBlocking {
    val myFlow = callbackFlow<Any> {
      close()

      // Should throw?
      delay(Duration.INFINITE)
    }

    // Should complete?
    myFlow.collect()
  }
  ```
  Here, `delay` used to throw a cancellation exception, and the flow would complete.
  Now, it blocks on `delay`, and `collect()` never completes.

qwwdfsad:
Thanks for the report, that's an unexpected side-effect, filed https://github.com/Kotlin/kotlinx.coroutines/issues/2506

Issue #2386, opened 2020-11-13
java.lang.ClassCastException: to `CoroutineStackFrame` at kotlinx.coroutines.internal.ScopeCoroutine.getCallerFrame(Scopes.kt:19)


willbuck:
For library authors wishing to implement a custom `kotlin.coroutines.Continuation` class, casting to an internal class like `CoroutineStackFrame` is problematic. 

See, for example, this issue filed against micronaut:

https://github.com/micronaut-projects/micronaut-core/issues/4438

The error comes on this line, where micronaut has implemented `Continuation` with the `CompletableFutureContinuation` class

https://github.com/micronaut-projects/micronaut-core/blob/0afdef29d0a6b38b55dc39e3ef2a72bea85be45f/aop/src/main/java/io/micronaut/aop/internal/intercepted/KotlinInterceptedMethod.java#L123

I think the solve is to have `ScopeCoroutine` optionally cast `uCont` to `CoroutineStackFrame`, as `callerFrame` is nullable anyway (and it also seems like what other implementations of `CoroutineStackFrame`, like `SelectBuilderImpl` and `StackFrameContinuation` are doing).

I can work on adding a test to reproduce this issue and create a PR since I'm already in the weeds on it, will update once I do!

qwwdfsad:
Thanks for the report! Filed https://youtrack.jetbrains.com/issue/KT-43395

This is indeed bug on our side worth fixing, but while we are here, I would strongly suggest implementing `CoroutineStackFrame` in `CompletableFutureContinuation` in the following way:

```
class CompletableFutureContinuation(private val continuation: Continuation<Any>) : Continuation<Any>, CoroutineStackFrame {

    override val callerFrame: CoroutineStackFrame?
            get() = continuation as? CoroutineStackFrame

    override fun getStackTraceElement(): StackTraceElement? = null
}
```

willbuck:
@qwwdfsad thanks for the feedback! As far as we could tell, `CoroutineStackFrame` is an internal class, and thus would not be accessible outside of the `kotlinx.coroutines` module for Micronaut to extend it. Is that mistaken?-

qwwdfsad:
`CoroutineStackFrame` is actually a public interface from kotlin stdlib, in the package `kotlin.coroutines.jvm.internal`.

[...]

PR #2388, opened 2020-11-14
fixes ScopeCoroutine cast to internal CoroutineStackFrame (#2386)

willbuck:
I attempted (unsuccessfully) to orchestrate the proper conditions to reproduce the issue, but I am afraid my mastery of coroutineScope, job scheduling, interception and the like was such that I was unable to do so.

What I was trying was adding a second test to `DebugAgentTest` using a barebones custom continuation implementation, something like this:

```
class CustomContinuationImpl<T>(continuation: Continuation<T>): Continuation<T> {
    override val context = continuation.context
    override fun resumeWith(result: Result<T>) {
        println("CustomContinuationImpl for result $result")
        // no-op, this is just a sample class to reproduce issue #2386
    }
}
```

But figuring out how to stitch that into a test s.t. that it would own something that was a ScopeCoroutine s.t. it would attempt to be cast to `CoroutineStackFrame` within the `DebugProbesImpl` wasn't working out for me. I would certainly be open to guidance on how better to set up such a test, but perhaps this is a simple enough fix to just merge sans test replicating it? Feedback welcome!

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/Scopes.kt
+++ b/kotlinx-coroutines-core/common/src/internal/Scopes.kt
@@ -16,7 +16,7 @@ internal open class ScopeCoroutine<in T>(
     context: CoroutineContext,
     @JvmField val uCont: Continuation<T> // unintercepted continuation
 ) : AbstractCoroutine<T>(context, true), CoroutineStackFrame {
-    final override val callerFrame: CoroutineStackFrame? get() = uCont as CoroutineStackFrame?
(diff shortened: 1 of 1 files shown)


qwwdfsad (approved):
Thanks!

Will merge manually with minor change: 4a099ff21d6a7d53fd0801deeb9ccf967c168ea7

qwwdfsad:
Merged manually, thanks!

PR #2389, opened 2020-11-15
Add build parameter to build coroutines with JVM IR compiler

ALikhachev:
To be used in specific kotlinx-train CI configuration. It's needed for faster adoption of JVM IR compiler.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,9 @@ buildscript {
             throw new IllegalArgumentException("'kotlin_snapshot_version' should be defined when building with snapshot compiler")
         }
     }
+    // These two flags are enabled in train builds for JVM IR compiler testing
+    ext.jvm_ir_enabled = rootProject.properties['enable_jvm_ir'] != null
+    ext.native_targets_enabled = rootProject.properties['disable_native_targets'] == null
 
     // Determine if any project dependency is using a snapshot version
     ext.using_snapshot_version = build_snapshot_train
--- a/build.gradle
+++ b/build.gradle
@@ -33,8 +33,9 @@ buildscript {
             throw new IllegalArgumentException("'kotlin_snapshot_version' should be defined when building with snapshot compiler")
         }
     }
-    // These two flags are enabled in train builds for JVM IR compiler testing
+    // These three flags are enabled in train builds for JVM IR compiler testing
     ext.jvm_ir_enabled = rootProject.properties['enable_jvm_ir'] != null
+    ext.jvm_ir_api_check_enabled = rootProject.properties['enable_jvm_ir_api_check'] != null
     ext.native_targets_enabled = rootProject.properties['disable_native_targets'] == null
 
     // Determine if any project dependency is using a snapshot version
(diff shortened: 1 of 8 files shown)


kotlinx-coroutines-core/build.gradle
```diff
@@ -183,6 +190,14 @@ jvmTest {
         exclude '**/*StressTest.*'
     }
     systemProperty 'kotlinx.coroutines.scheduler.keep.alive.sec', '100000' // any unpark problem hangs test
+
+    // TODO: JVM IR generates different stacktrace so temporary disable stacktrace tests
```
qwwdfsad, line 194: Do I have to set up a reminder for that? When is it safe to remove this filter?

build.gradle
```diff
@@ -33,6 +33,9 @@ buildscript {
             throw new IllegalArgumentException("'kotlin_snapshot_version' should be defined when building with snapshot compiler")
         }
     }
+    ext.jvm_ir_enabled = rootProject.properties['enable_jvm_ir'] != null
+    ext.jvm_ir_api_check_enabled = rootProject.properties['enable_jvm_ir_api_check'] != null
+    ext.native_targets_enabled = rootProject.properties['disable_native_targets'] == null
```
qwwdfsad, line 39: Why are you using `-Xallow-jvm-ir-dependencies` in atomicfu, but not here?

qwwdfsad (changes requested):

build.gradle
ALikhachev: This flag was was used in atomicfu only for test project for testing gradle plugin. Here we will fairly use IR dependency with IR compiled project

kotlinx-coroutines-core/build.gradle
ALikhachev: @udalov

udalov: I suppose when we switch kotlinx-coroutines to JVM IR, we'll just update these stacktraces with the new expectations and unmute the tests. Stacktraces differ slightly in JVM IR because of slightly different logic in the tail call optimization. Here's the commit that makes the tests pass for me on JVM IR, @qwwdfsad please verify that it's OK (to be applied later when we enable JVM IR) or if there's something that needs to be fixed in the backend:
https://github.com/udalov/kotlinx.coroutines/commit/352a173dd160de3ceaa1266c4eeb45cb7d94fbb6

BTW @ALikhachev `CoroutinesDumpTest` passes for me with JVM_IR, so probably it can be unmuted (both here and in -debug).

qwwdfsad: Verified 👍 

@ALikhachev please unmute `CoroutinesDumpTest` then

qwwdfsad (approved):

kotlinx-coroutines-debug/build.gradle
```diff
@@ -28,6 +28,16 @@ dependencies {
     api "net.java.dev.jna:jna-platform:$jna_version"
 }
 
+// TODO: JVM IR generates different stacktrace so temporary disable stacktrace tests
+if (rootProject.ext.jvm_ir_enabled) {
+    tasks.named('test', Test) {
+        filter {
+            excludeTest('kotlinx.coroutines.debug.CoroutinesDumpTest', 'testCreationStackTrace')
```
udalov, line 35: This one can also be unmuted

udalov (approved):

PR #2390, opened 2020-11-16
Fix cancellation support for Mutex when lock becomes available between tryLock and suspending.

### LINKED ISSUE #2380: mutex.withLock does not reliably release lock on cancellation.
Playground here with close to minimal test case code.
https://pl.kotl.in/AGAyMv-l5

Running with Kotlin 1.4.10 it will output something like:
37,31,6
java.lang.IllegalStateException: Already locked by java.lang.Object@7c23b6a

With 1.3.72 it outputs:
26,26,0
java.lang.Exception

So it appears to be a regression.

The use of owner is not required, it just allows for the throwing of exception and counting failure frequency rather than hanging the entire test.

Tilps:
A synchronous call to continuation.resume in suspendLock doesn't preclude a cancellation race, so it needs the same care as during the passing of ownership that takes place in unlock. 
Fixes #2380


qwwdfsad:
Thanks! Opened https://github.com/Kotlin/kotlinx.coroutines/pull/2396 with semaphore fix and better solution, will merge it instead

PR #2393, opened 2020-11-16
update: update docs for the new WebHelp engine

koshachy:


docs/topics/debug-flow-with-idea.md
```diff
@@ -0,0 +1,126 @@
...
+    The **Debug** tool window appears: 
+    * The **Frames** tab contains the call stack.
+    * The **Variables** tab contains variables in the current context. It tells us that the flow is emitting the first value.
+    * The **Coroutines** tab contains information on running or suspended coroutines.
+
+    ![Debug the coroutine](flow-debug-1.png)
+
+3. Resume the debugger session by clicking **Resume program** in the **Debug** tool window. The program stops at the same breakpoint.
+
+    ![Debug the coroutine](flow-resume-debug.png)
```
qwwdfsad, line 79: These links are broken for MD files. Is it expected?

qwwdfsad (changes requested):
Three major points:

1) Please update `CONTRIBUTING.md` so it's clear how to edit the documentation (stardust plugin etc.) and see the result. Currently, it's not clear either to maintainers or external contributors

2) We have references to the guide in our main Github page (README section) and these guides are now not so readable. I'd suggest replacing links in README from .md files to the actual website. In addition, it would be nice if each .md section will start with a paragraph like "...this is an internal documentation, please use ${THIS_LINK_TO_THE_WEBSITE_WITH_THIS_SECTION} instead"

3) We moved a lot of `.md` files, would be nice to have basic redirect (even if it's a short text "file was moved") for people who have been referencing these docs from their blogposts and code.

docs/topics/debug-flow-with-idea.md
koshachy: Yep, the WebHelp file structure ignores hierarchy, it's flat.

katia-energizer: We discussed this in Slack last year. It's expected for now. We'll think what we can do with this in the future if we see that this is a real issue for users. 
Based on current information, it doesn't seem to be an issue for contributors. KMM docs have much more images than coroutines docs and we didn't get any feedback from users complaining that images were not rendered in MD.

If we get such feedback, we'll reprioritize this task.

koshachy:
> Three major points:
> 
> 1. Please update `CONTRIBUTING.md` so it's clear how to edit the documentation (stardust plugin etc.) and see the result. Currently, it's not clear either to maintainers or external contributors

@katia-energizer will prepare it soon.

> 2. We have references to the guide in our main Github page (`README` section) and these guides are now not so readable. I'd suggest replacing links in `README` from `.md` files to the actual website. In addition, it would be nice if each `.md` section started with a paragraph like "...this is an internal documentation, please use ${THIS_LINK_TO_THE_WEBSITE_WITH_THIS_SECTION} instead"

I've updated the README.md with new links. Some files (`debugging.md` and `compatibility.md`) are not published on our website, so I didn't change their links. These files are in old MD format, so there are no issues to render them with standard GH renderer.

> 3. We moved a lot of `.md` files, would be nice to have basic redirect (even if it's a short text "file was moved") for people who have been referencing these docs from their blogposts and code.

I've added stubs for all moved *.md files.

qwwdfsad (approved):

PR #2396, opened 2020-11-17
Pr/2390

qwwdfsad:


kotlinx-coroutines-core/jvm/test/sync/SemaphoreStressTest.kt
```diff
@@ -90,4 +90,22 @@ class SemaphoreStressTest : TestBase() {
...
+        val semaphore = Semaphore(1)
+        val n = 1000 * stressTestMultiplier
+        repeat(n) {
+            val job = launch(Dispatchers.Default) {
+                semaphore.acquire()
+                semaphore.release()
+            }
+            semaphore.withPermit {
+                job.cancel()
+
```
elizarov, line 105: Nit: extra empty line

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -201,7 +201,8 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
                         // try lock
                         val update = if (owner == null) EMPTY_LOCKED else Empty(owner)
                         if (_state.compareAndSet(state, update)) { // locked
-                            cont.resume(Unit)
+                            // TODO implement functional type in LockCont as soon as we get rid of legacy JS
```
elizarov, line 204: Btw, if that helps, we can optimize it for `onwer==null` case by caching a reference to a preallocated `onCancellationUnlock` reference similar to the way it is done in Semaphore.

elizarov (approved):
💯 Awesome. That's the right way to fix it. Tests are also great. See some comments.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -201,7 +201,8 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
...
+                            cont.resume(Unit) { unlock(owner) }
```
yschimke, line 205: I love Kotlin's last param is lambda logic generally. I think in this API, this callsite is an example of where it hides the intent.

```
cont.resume(Unit, onCancellation = { unlock(owner) })
```

PR #2397, opened 2020-11-17
Make SharingStarted a fun interface

elizarov:
It was a part of the original design, but was forgotten because prototype was developer before Kotlin 1.4.0.
It makes implementing custom SharingStarted strategies more concise.


qwwdfsad (approved):
👍

PR #2399, opened 2020-11-18
Clarify thread-safety of SharedFlow methods in docs

elizarov:
* Override MutableSharedFlow.emit to attach a more appropriate docs than the one inherited from FlowCollector.
* Clarify thread-safety of all the MutableSharedFlow & MutableState "mutating" methods.

The latter is needed, because Flows, in general, are sequential, but shared flows provide all the necessarily synchronization themselves, so, to avoid confusion it makes sense to additionally mention thread-safety of shared flows in all the relevant mutating functions.

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -1064,6 +1064,7 @@ public final class kotlinx/coroutines/flow/LintKt {
 }
 
 public abstract interface class kotlinx/coroutines/flow/MutableSharedFlow : kotlinx/coroutines/flow/FlowCollector, kotlinx/coroutines/flow/SharedFlow {
+	public abstract fun emit (Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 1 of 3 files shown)


qwwdfsad (approved):
A bit weird, because we already have an explicit mention of this behaviour in class documentation https://github.com/Kotlin/kotlinx.coroutines/blob/d4c3c9cb3826cf2cb762a5521a50740a2ca5c0c6/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt#L100

E.g. we do not expect to have each method of `ConcurrentHashMap` to be documented as thread-safe, do we?

I don't really mind, it just a bit excessive

Tolriq:
@qwwdfsad HashMap method does not explicitly have methods with docs that says it's not thread safe. Here IDE on emit says in big not thread safe making people doubt.

elizarov:
I believe that excessiveness is warranted here, because a regular `Flow` is sequential and, in general, a `FlowCollector.emit` cannot be used concurrently. `SharedFlow` is an exception and stressing this difference is important (not everyone will read load `SharedFlow` doc in full).

PR #2402, opened 2020-11-20
CoroutinesTimeout for JUnit5

### LINKED ISSUE #2197: CoroutinesTimeout replacement for JUnit 5
https://github.com/asarkar/coroutines-test

dkhalanskyjb:
Fixes #2197.


kotlinx-coroutines-debug/build.gradle
```diff
@@ -20,6 +20,10 @@ configurations {
 
 dependencies {
     compileOnly "junit:junit:$junit_version"
+    compileOnly "org.junit.jupiter:junit-jupiter-api:$junit5_version"
+    testCompile "org.junit.jupiter:junit-jupiter-api:$junit5_version"
```
qwwdfsad, line 24: should `testCompile` duplicate `compileOnly`?

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeout.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.api.extension.*
+import org.junit.jupiter.api.parallel.*
+import java.lang.annotation.*
+
+/**
+ * Coroutines timeout annotation that is similar to JUnit5's [Timeout] annotation. It allows running test methods in a
+ * separate thread, failing them after the provided time limit and interrupting the thread.
+ *
+ * Additionally, it installs [DebugProbes] and dumps all coroutines at the moment of the timeout. It also cancels
+ * coroutines on timeout if [cancelOnTimeout] set to `true`. The dumps contain the coroutine creation stack traces. If
```
qwwdfsad, line 18:
```suggestion
 * coroutines on timeout if [cancelOnTimeout] set to `true`. The dump contains the coroutine creation stack traces. If
```

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeout.kt
```diff
@@ -0,0 +1,70 @@
...
+ * whether the methods that it produces obey the timeout as well.
+ *
+ * Beware that registering the extension via this annotation conflicts with manually registering the extension on the
+ * same tests via other methods (most notably, [RegisterExtension]) and is prohibited.
+ *
+ * Example usage:
+ * ```
+ * @CoroutinesTimeout(100)
+ * class CoroutinesTimeoutSimpleTest {
+ *      // times out in 150 ms
```
qwwdfsad, line 39: It timeouts in 1000, isn't it?

Minor optional nit: annotations sorted by the length of the names are typically more readable (tho it's quite opinionated, so feel free to ignore it)

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeout.kt
```diff
@@ -0,0 +1,70 @@
...
+ *     @CoroutinesTimeout(1000)
+ *     @Test
+ *     fun classTimeoutIsOverridden() {
+ *         runBlocking {
+ *             delay(150)
+ *         }
+ *     }
+ *
+ *     // times out in 100 ms
```
qwwdfsad, line 48:
```suggestion
 *     // times out in 100 ms, timeout value is taken from the class-level annotation
```

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeoutExtension.kt
```diff
@@ -0,0 +1,282 @@
...
+
+import kotlinx.coroutines.debug.*
+import kotlinx.coroutines.debug.runWithTimeoutDumpingCoroutines
+import org.junit.jupiter.api.extension.*
+import org.junit.platform.commons.support.AnnotationSupport
+import java.lang.reflect.*
+import java.util.*
+import java.util.concurrent.atomic.*
+
+public class CoroutinesTimeoutException(public val timeoutMs: Long): Exception("test timed out ofter $timeoutMs ms")
```
qwwdfsad, line 15: This exception definitely shouldn't be public.

Moreover, it seems that throwing `java.util.concurrent.TimeoutException` is enough

kotlinx-coroutines-debug/test/junit5/CoroutinesTimeoutSimpleTest.kt
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package junit5
```
qwwdfsad, line 5: all tests have a wrong package directive, should be `kotlinx.coroutines.debug.junit5`

qwwdfsad (changes requested):
Skimmed through KDoc and public API, will get back to the engine impl later.

Could you please rebase this branch on the recent `develop`?

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeoutExtension.kt
dkhalanskyjb: I don't like `java.util.concurrent.TimeoutException` here, as its documentation says the following on the first line:
> Exception thrown when a blocking operation times out.

I don't feel that this describes our case. So, I think that a separate class is warranted after all. I agree that it shouldn't be public though, as there aren't many opportunities to catch it.

kotlinx-coroutines-debug/build.gradle
dkhalanskyjb: Not at this point: other dependencies in `testCompile` depend on `junit-jupiter-api` transitively. I removed the duplicate.

kotlinx-coroutines-debug/build.gradle
```diff
@@ -38,6 +41,12 @@ if (rootProject.ext.jvm_ir_enabled) {
     }
 }
 
+java {
```
qwwdfsad, line 44: Doesn't seem to be necessary for the latest `develop` where we have 1.8 target

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeoutExtension.kt
```diff
@@ -0,0 +1,282 @@
...
+                           enableCoroutineCreationStackTraces: Boolean = true): CoroutinesTimeoutExtension =
+            CoroutinesTimeoutExtension(enableCoroutineCreationStackTraces, timeout.toLong() * 1000, cancelOnTimeout)
+    }
+
+    /** @see [initialize] */
+    private val debugProbesOwnershipPassed = AtomicBoolean(false)
+
+    private fun tryPassDebugProbesOwnership() = debugProbesOwnershipPassed.compareAndSet(false, true)
+
+    private val isDebugProbesOwnershipPassed
```
qwwdfsad, line 84: unused val

kotlinx-coroutines-debug/src/junit/junit5/CoroutinesTimeoutExtension.kt
```diff
@@ -0,0 +1,282 @@
...
+ *          delay(Long.MAX_VALUE) // somewhere deep in the stack
+ *          ...
+ *     }
+ * }
+ * ```
+ *
+ * @see [CoroutinesTimeout]
+ * */
+// NB: the constructor is not private so that JUnit is able to call it via reflection.
+public class CoroutinesTimeoutExtension internal constructor(
```
qwwdfsad, line 57: We do not expect extension to be widely used manually, let's make it experimental or even `internal`?

qwwdfsad (commented):
Let's address one final question and ship it 🚀

kotlinx-coroutines-debug/build.gradle
dkhalanskyjb: I attempted to remove this block, but then the build was failing. I managed to fix the build of this subproject by adding
```gradle
tasks.withType(JavaCompile) {
    targetCompatibility = JavaVersion.VERSION_1_8
}
```
However, then kotlinx-coroutines-test also required `targetCompatibility` set. At that point, I stopped. Do you think this should be pursued further?

qwwdfsad: I think that's enough, thanks for checking it

qwwdfsad (approved):
Great!

PR #2418, opened 2020-11-30
Remove outdated Android examples

qwwdfsad:
These examples are way outdated, unmaintained and cannot be built as is in the most recent Android Studio. There is plenty of high-quality examples in the Web right now, no need to keep (and update Kotlin version) these two


elizarov (approved):
👍 Good riddance.

PR #2420, opened 2020-11-30
Use combineUnsafe for 4 parameter combine

### LINKED ISSUE #2419: 4-parameter version of `combine` does not use `combineUnsafe`
This is a strange discrepancy I noticed in [`Zip.kt`](https://github.com/Kotlin/kotlinx.coroutines/blob/0e67d0a89bc006305e71dbc905cc0a6c4cf562b6/kotlinx-coroutines-core/common/src/flow/operators/Zip.kt#L156), after the performance improvements of #2308. It looks like the 4 parameter version of `combine` wasn't updated to use `combineUnsafe`, unlike the 3 parameter and 5 parameter versions.

alexvanyo:
This fixes #2419 by updating the 4-parameter version of combine to use the optimized `combineUnsafe`.


qwwdfsad (approved):
Thanks!

Issue #2423, opened 2020-12-01
Implementation by delegation violates internal invariant of JobSupport


vadimsemenov:
```
    @Test
    fun testAwaitAllDelegates() = runTest {
        expect(1)
        val deferred = CompletableDeferred<String>()
        deferred.invokeOnCompletion {
            expect(4)
        }

        val delegate = object : Deferred<String> by deferred {}  // This is failing
//        val delegate = deferred  // This works fine

        launch {
            expect(3)
            deferred.complete("OK")
        }
        expect(2)
        awaitAll(delegate)
        finish(5)
    }
```

I managed to reproduce it with `awaitAll`, not sure if the problem is in JobSupport or in AwaitAll.

Full stack trace:
```
java.lang.AssertionError
	at kotlinx.coroutines.JobSupport.makeNode(JobSupport.kt:516)
	at kotlinx.coroutines.JobSupport.invokeOnCompletion(JobSupport.kt:496)
	at kotlinx.coroutines.JobSupport.invokeOnCompletion(JobSupport.kt:449)
	at kotlinx.coroutines.AwaitTest$testAwaitAllDelegates$1$delegate$1.invokeOnCompletion(AwaitTest.kt)
	at kotlinx.coroutines.AwaitAll.await(Await.kt:79)
	at kotlinx.coroutines.AwaitKt.awaitAll(Await.kt:26)
	at kotlinx.coroutines.AwaitTest$testAwaitAllDelegates$1.invokeSuspend(AwaitTest.kt:269)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:274)
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:84)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
	at kotlinx.coroutines.TestBase.runTest(TestBase.kt:194)
	at kotlinx.coroutines.TestBase.runTest$default(TestBase.kt:188)
	at kotlinx.coroutines.AwaitTest.testAwaitAllDelegates(AwaitTest.kt:254)
[...]

PR #2425, opened 2020-12-01
Remove protection from downcasting in asDeferred

vadimsemenov:
To work around #2423


qwwdfsad (commented):
I'll take a look at #2423 and see if it can be easily fixed first. Maybe this PR is not really needed

qwwdfsad:
Fixed by #2427, closing

PR #2431, opened 2020-12-04
Gradle stewardship

qwwdfsad:
* Remove conditional publication that is compatible with Kotlin 1.3.70 plugin
    * Get rid of workaround in old MPP plugin with dependency renaming

### DIFF
--- a/gradle/publish-bintray.gradle
+++ b/gradle/publish-bintray.gradle
@@ -13,7 +13,6 @@ apply plugin: 'maven-publish'
 
 def isMultiplatform = project.name == "kotlinx-coroutines-core"
 def isBom = project.name == "kotlinx-coroutines-bom"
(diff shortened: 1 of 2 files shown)


gradle/publish-bintray.gradle
```diff
@@ -78,33 +77,23 @@ publishing {
...
+                it.artifactId = "$project.name-metadata"
                 break
             case 'jvm':
-                it.artifactId = isKotlin137x ? project.name : "$project.name-jvm"
+                it.artifactId = "$project.name-jvm"
                 break
             case 'js':
             case 'native':
                 it.artifactId = "$project.name-$type"
                 break
```
whyoleg, line 95: seems now, all those switch cases are the same :)

qwwdfsad: thanks!

elizarov (approved):
👍 Looks good to me.

PR #2436, opened 2020-12-07
Atomicfu delegated properties

SokolovaMaria:
Atomicfu updated to `0.15.0`


elizarov (changes requested):
Something is not right -> fails tests.

PR #2437, opened 2020-12-07
Update SharedFlow documentation

qwwdfsad:


ivakub:
https://youtrack.jetbrains.com/issue/KT-43770

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -148,8 +148,8 @@ public interface SharedFlow<out T> : Flow<T> {
  */
 public interface MutableSharedFlow<T> : SharedFlow<T>, FlowCollector<T> {
     /**
-     * Emits a [value] to this shared flow, suspending on buffer overflow if the shared flow was created
-     * with the default [BufferOverflow.SUSPEND] strategy.
+     * Emits a [value] to this shared flow, suspending on buffer overflow until all subscribers receive the value
```
elizarov, line 151: That's a correct statement only about a shared flow without a buffer. A buffered flow only suspends when there's no room in the buffer. So, without a buffer it translates into suspension until every subscribers receives (because there's no buffer to store it).

elizarov (changes requested):

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
qwwdfsad: Reverted this line

elizarov (approved):
Nice.

PR #2438, opened 2020-12-07
Update Lincheck version to 2.11

ndkoval:
In Lincheck 2.11 we 

1. fixed prompt cancellation support,
2. fixed a bug where cancellation handlers could not be aborted (does not really matter for current implementations), and
3. slightly improved the trace representation. 

In addition, it is now possible to show all occurring events in the trace (without collapsing function invocations that were not interrupted) via `ModelCheckingOptions.verboseTrace`.


ndkoval:
Don't merge now, please. There is a bug with prompt cancellation emulation in Lincheck, let's fix it and release a new Lincheck version at first.

PR #2440, opened 2020-12-08
Cleanup core module after update to Kotlin 1.4

qwwdfsad:
* Remove explicit API mode suppressions
    * Use list build in ReceiveChannel.toList to prevent casting to mutable list
    * Migrate to maxByOrNull
    * Remove redundant nullabilities


elizarov (approved):
👍 Very good

PR #2442, opened 2020-12-09
Update deprecate symbols

qwwdfsad:


elizarov:
It should be labeled for "1.5", isn't it?

elizarov (approved):
Great job.

PR #2443, opened 2020-12-11
Bug fix and better LockFreeLinkedList verification with Lincheck

elizarov:
Fixes a rare crash of LockFreeLinkedListAtomicLFStressTest, e.g:
https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStressWindows/3215239
with AssertionError at kotlinx.coroutines.internal.LockFreeLinkedListNode.validateNode

This was happening because during concurrent execution of a complex atomic operation a thread that adds a node to the list could be helped to complete by other threads and the newly added node could be already removed. Now, the original adder thread will execute its add operation too, adding the removed element back to the list, leaving the list in the state that is logically correct, but which is not correct physically (with a removed node still in the list).

The Lincheck test that consistently reproduces this problem is added and the fix is introduced.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
@@ -437,6 +437,12 @@ public actual open class LockFreeLinkedListNode {
                 else -> next // roll back if consensus if failure
             }
             affected._next.compareAndSet(this, update)
+            if (consensus == null) {
+                // If the consensus was on a successful completion of operation the above CAS could have reinstalled
+                // the node that was already removed into the list. This call to finishOnSuccess properly completes
+                // the ongoing operation.
+                desc.finishOnSuccess(affected, next)
+            }
             return null
         }
 
@@ -451,9 +457,9 @@ public actual open class LockFreeLinkedListNode {
         protected open fun takeAffectedNode(op: OpDescriptor): Node? = affectedNode!! // null for RETRY_ATOMIC
         protected open fun failure(affected: Node): Any? = null // next: Node | Removed
         protected open fun retry(affected: Node, next: Any): Boolean = false // next: Node | Removed
-        protected abstract fun finishOnSuccess(affected: Node, next: Node)
 
         public abstract fun updatedNext(affected: Node, next: Node): Any
+        public abstract fun finishOnSuccess(affected: Node, next: Node)
 
         public abstract fun finishPrepare(prepareOp: PrepareOp)
 
@@ -540,7 +546,15 @@ public actual open class LockFreeLinkedListNode {
      */
     private fun finishAdd(next: Node) {
         next._prev.loop { nextPrev ->
-            if (this.next !== next) return // this or next was removed or another node added, remover/adder fixes up links
+            val thisNext = this.next
+            if (thisNext !== next) { // this or next was removed or another node added
+                if (thisNext is Removed) {
+                    // If this node was removed, let's help the removal operation complete
+                    next.correctPrev(null)
+                }
+                // Otherwise another node was added -> adder fixes up links
+                return
+            }
             if (next._prev.compareAndSet(nextPrev, this)) {
                 // This newly added node could have been removed, and the above CAS would have added it physically again.
                 // Let us double-check for this situation and correct if needed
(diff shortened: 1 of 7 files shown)


kotlinx-coroutines-core/jvm/test/internal/LockFreeLinkedListAtomicLFStressTest.kt
```diff
@@ -18,6 +18,7 @@ import kotlin.test.*
  */
 class LockFreeLinkedListAtomicLFStressTest {
```
ndkoval, line 19: Do we still need this test if we require obstruction-freedom in `LockFreeListMultiLincheckTest`?

elizarov: Extra testing will never harm, IMO.

ndkoval: The only advantage I see in this obstruction-freedom test (guess, you also test for obstruction-freedom, not for lock-freedom like the test name says) is that it can find bugs caused by weak memory model effects. Should we add this functionality to the stress testing in Lincheck and remove all such tests? I personally find this test quite complicated and would be happy to replace it with a straightforward one someday 🙂 .

kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
```diff
@@ -611,11 +625,21 @@ public actual open class LockFreeLinkedListNode {
...
+
     internal fun validateNode(prev: Node, next: Node) {
-        assert { prev === this._prev.value }
-        assert { next === this._next.value }
+        val result = validateNodeResult(prev, next)
+        check(result.isEmpty()) { result }
+    }
+
+    internal fun validateNodeResult(prev: Node, next: Node): String = buildString {
+        append(validateSameResult("prev", prev, _prev.value))
```
ndkoval, line 636: It seems that `next` property completes the descriptor stored in it, and this is the only way `prev` and `_prev.value` to be different. However, reading `prev` at first always completes the descriptor. Thus, it seems that this check is useless. I would avoid modifications in such validation functions, it is error-prone.

kotlinx-coroutines-core/jvm/test/lincheck/LockFreeListMultiLincheckTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    @Operation
+    fun addLast(@Param(name = "id") id: Int, @Param(name = "value") value: Int) {
+        q[id].addLast(Node(value))
+    }
+
+    @Operation
+    fun addTwo(@Param(name = "id") id: Int, @Param(name = "value") value: Int) {
+        var result: Any?
+        do {
+            val add1 = q[id].describeAddLast(Node(value))
```
ndkoval, line 52: You always follow the same order here ("0, 1", "1, 2", "2, 3") and never the opposite. However, everything is fine with two lists. I do not think that you should fix this, but want to draw some attention.

ndkoval (changes requested):

kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
elizarov: It seems that I've fixed it -- I'm checking raw `_next.value`.

kotlinx-coroutines-core/jvm/test/lincheck/LockFreeListMultiLincheckTest.kt
elizarov: The order is "0,1", "1, 2", and "2,0" (mind the modulo operation), so there is no hierarchy and it properly stresses the deadlock scenario. Moreover, there is variant of the test with n=2, with does both "0,1" and "1,0".  You can comment out `override val opSequence` to disable deadlock-avoidance and check that this test immediately produces `StackOverflow`.

kotlinx-coroutines-core/jvm/src/internal/LockFreeLinkedList.kt
ndkoval: Aaah, I see -- you use the same names for the argument and the field, it confused me a bit. Then it is correct, you are right.

kotlinx-coroutines-core/jvm/test/lincheck/LockFreeListMultiLincheckTest.kt
ndkoval: 👍

ndkoval (approved):

elizarov:
Note: this PR does not pass stress-tests. See: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_NightlyStress/3249233?buildTab=tests&status=failed&expandedTest=827483333453888335

PR #2444, opened 2020-12-11
Fix LinkedListChannel.onUndeliveredElement call on channel cancel

### LINKED ISSUE #2435: Unlimited channel doesn't correctly invoke onUndeliveredElement on cancellation
Reproducer:
```
 val channelBuffered = Channel<Int>(10) { println("RELEASE[BUFFER]: $it") }
 val channelUnlimited = Channel<Int>(Channel.UNLIMITED) { println("RELEASE[UNLIMITED]: $it") }
 repeat(5) {
     channelBuffered.offer(it)
     channelUnlimited.offer(it)
 }
 channelBuffered.cancel()
 channelUnlimited.cancel()
```

elizarov:
Fixes #2435

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
@@ -680,7 +680,7 @@ internal abstract class AbstractChannel<E>(
     }
 
     /**
-     * This method is overriden by [LinkedListChannel] to handle cancellation of [SendBuffered] elements from the list.
+     * This method is overridden by [LinkedListChannel] to handle cancellation of [SendBuffered] elements from the list.
      */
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -669,6 +676,13 @@ internal abstract class AbstractChannel<E>(
             // Add to the list only **after** successful removal
             list += previous as Send
         }
+        onCancelIdempotentList(list, closed)
+    }
+
+    /**
+     * This method is overriden by [LinkedListChannel] to handle cancellation of [SendBuffered] elements from the list.
```
qwwdfsad, line 683:
```suggestion
     * This method is overridden by [LinkedListChannel] to handle cancellation of [SendBuffered] elements from the list.
```

qwwdfsad (approved):

Issue #1437, opened 2019-08-12
Take into account synthetic declarations when sanitizing a stacktrace


qwwdfsad:
The following test
```
@Test
fun foo() = runBlocking {
    DebugProbes.dumpCoroutines()
}
```

will print these lines:
```
(Coroutine creation stacktrace)
	at kotlin.coroutines.intrinsics.IntrinsicsKt__IntrinsicsJvmKt.createCoroutineUnintercepted(IntrinsicsJvm.kt:116)
	at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:26)
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
	at definitely.not.kotlinx.coroutines.SanitizedProbesTest.foo(SanitizedProbesTest.kt:25)
```

Probably it is worth to make sanitizing heuristic aware of such methods and provide at least one frame with actual source, e.g.

```
(Coroutine creation stacktrace)
	at kotlin.coroutines.intrinsics.IntrinsicsKt__IntrinsicsJvmKt.createCoroutineUnintercepted(IntrinsicsJvm.kt:116)
	at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:26)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:36)
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
	at definitely.not.kotlinx.coroutines.SanitizedProbesTest.foo(SanitizedProbesTest.kt:25)
```

PR #2452, opened 2020-12-17
Preserve a frame with source code location when sanitizing traces

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/1437

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
@@ -492,7 +492,7 @@ internal object DebugProbesImpl {
                 // first index past the end of the span of internal methods that starts from `i`
                 var j = i + 1
                 while (j < size && stackTrace[j].isInternalMethod) {
-                    j += 1
+                    ++j
                 }
                 // index of the last non-synthetic internal methods in this span, or `i` if there are no such methods
                 var k = j - 1
@@ -508,7 +508,7 @@ internal object DebugProbesImpl {
                 i = j
             } else {
                 result += stackTrace[i]
-                i += 1
+                ++i
             }
         }
         return result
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-debug/test/SanitizedProbesTest.kt
```diff
@@ -63,6 +63,7 @@ class SanitizedProbesTest : DebugTestBase() {
                     "\t(Coroutine creation stacktrace)\n" +
                     "\tat kotlin.coroutines.intrinsics.IntrinsicsKt__IntrinsicsJvmKt.createCoroutineUnintercepted(IntrinsicsJvm.kt:116)\n" +
                     "\tat kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:23)\n" +
+                    "\tat kotlinx.coroutines.BuildersKt__Builders_commonKt.async\$default(Builders.common.kt)\n" +
```
qwwdfsad, line 66: It seems like the actual frame (one with line numbers) points to the wrong line.
Typically, it's worth reporting it to our bugtracker (subsystem: `Backend. JVM. Coroutines`, assignee -- Ilmir)

kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
```diff
@@ -477,33 +477,40 @@ internal object DebugProbesImpl {
...
+                    after `i`, so we include it. */
+                    result += stackTrace[k]
+                }
+                result += stackTrace[j - 1] // we include the other boundary of this span in any case, too
+                i = j
             } else {
-                result += element
-                includeInternalFrame = true
+                result += stackTrace[i]
+                i += 1
```
qwwdfsad, line 511: [nitpicking] `++i` is a single BCI, so it's better to use it.

qwwdfsad (approved):
Great job!

PR #2453, opened 2020-12-17
Fix incorrect ProGuard rules that allowed shrinking volatile felds

### LINKED ISSUE #1564: Insufficient Proguard configuration rules
If a field is never written, R8 will replace all reads of that field by the default value of the field. Therefore it is important that the Proguard configuration for kotlinx.coroutines correctly specifies which fields are written using reflection.

In https://issuetracker.google.com/issues/141416122, a small project is shared that breaks due to missing Proguard configuration rules in kotlinx.coroutines. I found that adding the following two keep rules to the project solves the issue.
```
-keep,allowobfuscation class kotlinx.coroutines.io.ByteBufferChannel { kotlin.coroutines.Continuation writeOp; }
-keep,allowobfuscation class kotlinx.io.pool.DefaultPool { long top; }
```

I then found that the Proguard configuration rules for kotlinx.coroutines contain the following rule (https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro#L8):
```
-keepclassmembernames class kotlinx.** {
    volatile <fields>;
}
```

This should be updated to:
```
-keepclassmembers class kotlinx.** {
    volatile <fields>;
}
```

In addition to specifying that the names of these fields cannot be changed, this also specifies that these fields are assigned using reflection (in this case, AtomicReferenceFieldUpdater).

Originally filed at https://issuetracker.google.com/issues/141416122 and https://github.com/ktorio/ktor/issues/1354.

dkhalanskyjb:
Fixes #1564

### DIFF
--- a/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
+++ b/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
@@ -3,7 +3,7 @@
 -keepnames class kotlinx.coroutines.CoroutineExceptionHandler {}
 
 # Most of volatile fields are updated with AFU and should not be mangled
--keepclassmembers class kotlinx.** {
+-keepclassmembers class kotlinx.coroutines.** {
     volatile <fields>;
 }
 
--- a/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
+++ b/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
@@ -3,12 +3,12 @@
 -keepnames class kotlinx.coroutines.CoroutineExceptionHandler {}
 
 # Most of volatile fields are updated with AFU and should not be mangled
--keepclassmembernames class kotlinx.** {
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
```diff
@@ -3,12 +3,12 @@
 -keepnames class kotlinx.coroutines.CoroutineExceptionHandler {}
 
 # Most of volatile fields are updated with AFU and should not be mangled
--keepclassmembernames class kotlinx.** {
+-keepclassmembers class kotlinx.** {
```
JakeWharton, line 6: While we're here it would be nice to scope these to the correct package rather than such a top-level one which encompasses many libraries.

dkhalanskyjb: Good call, thanks!

qwwdfsad (approved):

PR #2457, opened 2020-12-18
Fix await/asDeferred for MinimalState implementations

### LINKED ISSUE #2456: `CompletionStage<T>.await()` is broken for `MinimalStage`
Many JDK methods that return `CompletionStage` are internally implemented using a class called `MinimalStage`, which inherits implementation from `CompletableFuture` (and therefore extends `Future` at runtime) but throws `UnsupportedOperationException` at runtime for most `Future` methods e.g. `isDone()`. Provided the caller uses the declared return type of `CompletionStage` this doesn't matter.
Unfortunately `CompletionStage<T>.await()` checks reflectively whether the passed `CompletionStage` implements `Future` and then calls `Future#isDone()` if it does, leading to an exception like:

````
java.lang.UnsupportedOperationException
	at java.base/java.util.concurrent.CompletableFuture$MinimalStage.isDone(CompletableFuture.java:2844)
	at kotlinx.coroutines.future.FutureKt.await(Future.kt:150)
````

Minimal reproduction (tested on JDK11):

````
import java.util.concurrent.CompletableFuture
import kotlinx.coroutines.future.await

suspend fun main() {
    CompletableFuture.completedStage(Unit).await()
    println("Hello World") // never called
}
````

### PR DESCRIPTION
**Pull request by elizarov**
Fixes #2456

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
@@ -165,14 +165,9 @@ public suspend fun <T> CompletionStage<T>.await(): T {
     return suspendCancellableCoroutine { cont: CancellableContinuation<T> ->
         val consumer = ContinuationConsumer(cont)
         whenComplete(consumer)
-        if (cont.isActive) {
-            // avoid creation of a lambda when continuation was already resumed during whenComplete call
-            // TODO: In a major release this lambda can be made to extend CancelFutureOnCompletion class from core module
-            // This will further save one allocated object here.
-            cont.invokeOnCancellation {
-                future.cancel(false)
-                consumer.cont = null // shall clear reference to continuation to aid GC
-            }
+        cont.invokeOnCancellation {
+            future.cancel(false)
+            consumer.cont = null // shall clear reference to continuation to aid GC
         }
     }
 }
--- a/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
@@ -105,22 +105,23 @@ private fun Job.setupCancellation(future: CompletableFuture<*>) {
 }
 
 /**
- * Converts this completion stage to an instance of [Deferred].
+ * Converts this [CompletionStage] to an instance of [Deferred].
  * When this completion stage is an instance of [Future], then it is cancelled when
  * the resulting deferred is cancelled.
+ *
+ * ### Implementation details
+ *
+ * [CompletionStage] does not extend a [Future] and does not provide future-like methods to check for completion and
+ * to retrieve a resulting value, so this implementation always takes a slow path of installing a callback with
+ * [CompletionStage.whenComplete]. For cases when [CompletionStage] is statically known to be an instance
+ * of [CompletableFuture] there is an overload of this `asDeferred` function with a [CompletableFuture] receiver that
+ * has a fast-path for a future that is already complete. Note, that it is not safe to dynamically check if an instance
+ * of a [CompletionStage] is an instance of a [CompletableFuture], because JDK functions return instances that do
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: integration/kotlinx-coroutines-jdk8/src/future/Future.kt [hunk 1]
```diff
@@ -137,34 +138,98 @@ public fun <T> CompletionStage<T>.asDeferred(): Deferred<T> {
...
+ * Awaits for completion of [CompletionStage] without blocking a thread.
  *
  * This suspending function is cancellable.
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * stops waiting for the completion stage and immediately resumes with [CancellationException][kotlinx.coroutines.CancellationException].
- * This method is intended to be used with one-shot futures, so on coroutine cancellation completion stage is cancelled as well if it is instance of [CompletableFuture].
+ * This method is intended to be used with one-shot futures, so on coroutine cancellation completion stage is cancelled as well if it is instance of a [Future].
  * If cancelling given stage is undesired, `stage.asDeferred().await()` should be used instead.
+ *
+ * ### Implementation details
```
qwwdfsad, line 173: Actually, there is far more elegant and less-error prone solution: `CompletableStage.toCompletableFuture()`.
It returns `this` for actual instances of CF and does the corresponding wrapping for `MinimalStage` including proper `cancel` propagation.

This also will allow us to leave the single method and users can safely upcast the type without `UOE`  from our side

qwwdfsad (changes requested):

integration/kotlinx-coroutines-jdk8/src/future/Future.kt
elizarov: Very good 👍

integration/kotlinx-coroutines-jdk8/src/future/Future.kt
```diff
@@ -132,25 +135,28 @@ public fun <T> CompletionStage<T>.asDeferred(): Deferred<T> {
...
  */
 public suspend fun <T> CompletionStage<T>.await(): T {
+    val future = toCompletableFuture() // retrieve the future
     // fast path when CompletableFuture is already done (does not suspend)
-    if (this is Future<*> && isDone()) {
+    if (future.isDone) {
         try {
-            @Suppress("UNCHECKED_CAST")
-            return get() as T
+            @Suppress("UNCHECKED_CAST", "BlockingMethodInNonBlockingContext")
```
qwwdfsad, line 158: Filed https://youtrack.jetbrains.com/issue/KT-44092

integration/kotlinx-coroutines-jdk8/src/future/Future.kt
```diff
@@ -159,10 +165,14 @@ public suspend fun <T> CompletionStage<T>.await(): T {
     return suspendCancellableCoroutine { cont: CancellableContinuation<T> ->
         val consumer = ContinuationConsumer(cont)
         whenComplete(consumer)
-        cont.invokeOnCancellation {
-            // mayInterruptIfRunning is not used
-            (this as? CompletableFuture<T>)?.cancel(false)
-            consumer.cont = null // shall clear reference to continuation to aid GC
+        if (cont.isActive) {
```
qwwdfsad, line 168: Not a big fan of optimizations that work 0.01% of the time (when future completes after `isDone` check but after `whenComplete`): hard to test and reason about.
I may be too opinionated here tho

qwwdfsad (approved):
Good fix 👍

integration/kotlinx-coroutines-jdk8/src/future/Future.kt
elizarov: Yep. It is indeed not needed anymore, where we (again) have fast-path via `future.isDone` here.

Issue #2291, opened 2020-10-08
Get rid of backspace symbol in artificial stackframe for stacktrace recovery


qwwdfsad:
For the sake of IDE-friendliness, we are using backspace symbol `\b` in our [stacktrace recovery](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/debugging.md#stacktrace-recovery) machinery.

This trick is helpful to render a `StackTraceElement` as a raw string without line-numbers and parentheses that are obligatory for `StackTraceElement.toString` implementation.

While initially it was an acceptable solution, time moves on and debug mode is now frequently used in production environments, even by IntelliJ IDEA and its derivatives. 
For that, `\b` solution is no longer desirable: it clutters log files, obfuscates programmatic exception handling and even triggers crashes of third-party tools that do not escape special symbols (that have to be escaped) in exception frames.

### Solution

The solution is to get rid of `\b`, introduce properly documented artificial function to `kotlinx.coroutines` and make intermediate stackframe reference this function instead.
While it may be inconvenient for first-time users, it should be okay to once navigate to it in IDEA or to read its documentation to get used to it.

### Known issues

https://github.com/Flank/flank/issues/764
https://youtrack.jetbrains.com/issue/RIDER-52026

dkhalanskyjb:
It turns out that the coroutine debugger in IDEA would be slightly broken were we to release this change: https://github.com/Kotlin/kotlinx.coroutines/pull/2694#issuecomment-838341384. Because of this, we will revert https://github.com/Kotlin/kotlinx.coroutines/pull/2695 for the 1.5 release.

After the coroutine debugger is modified to accommodate such a change, we will need to wait for long enough to ensure that the supported versions of the Kotlin plugin are not affected by this. Then, we will be able to safely introduce the change.

qwwdfsad:
Here are the commits:

https://github.com/JetBrains/intellij-community/commit/fd29488cce387808d564d1e3c2ed830c7724c22f
https://jetbrains.team/p/ij/repositories/intellij/revision/657936acb69fbbc59a88e1b7e799f49e81d9f162

They cover all supported IDEA versions, so it's good to go.

PR #2461, opened 2020-12-22
Display artificial stack frames as calls to functions instead of text

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2291


kotlinx-coroutines-core/jvm/src/debug/ArtificialStackFrames.kt
```diff
@@ -0,0 +1,38 @@
...
+package kotlinx.coroutines.debug
+import kotlinx.coroutines.*
+
+/**
+ * A collection of artificial stack trace elements to be included in stack traces by the coroutines machinery.
+ */
+@InternalCoroutinesApi
+internal class ArtificialStackFrames {
+    /**
+     * Return an artificial stack trace element denoting the boundary between coroutine creation and its execution.
```
qwwdfsad, line 14:
```suggestion
     * Returns an artificial stack trace element denoting the boundary between coroutine creation and its execution.
```

kotlinx-coroutines-core/jvm/src/debug/ArtificialStackFrames.kt
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * In earlier versions of kotlinx-coroutines, this was displayed as "(Coroutine creation stacktrace)", which caused
+     * problems for tooling that processes stack traces: https://github.com/Kotlin/kotlinx.coroutines/issues/2291
+     *
+     * Note that presence of this marker in a stack trace implies that coroutine creation stack traces were enabled.
+     */
+    fun coroutineCreation(): StackTraceElement = Exception().stackTrace[0]
+
+    /**
+     * Return an artificial stack trace element denoting a coroutine boundary.
```
qwwdfsad, line 28:
```suggestion
     * Returns an artificial stack trace element denoting a coroutine boundary.
```

kotlinx-coroutines-core/jvm/src/debug/ArtificialStackFrames.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.debug
+import kotlinx.coroutines.*
+
+/**
+ * A collection of artificial stack trace elements to be included in stack traces by the coroutines machinery.
```
qwwdfsad, line 9: We can slightly improve this documentation: mention that it is for debugging purposes only and leave links to to the debug mode document and the corresponding property

kotlinx-coroutines-core/jvm/src/debug/ArtificialStackFrames.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+@InternalCoroutinesApi
+internal class ArtificialStackFrames {
```
qwwdfsad, line 12: `ArtificialStackFrames` is indeed not very precise name.

I'd suggest preparing a small list of potential names and discuss it for 15 minutes during our regular DM.
Or it can even be done asynchronously in Slack.
A few potential names: `CoroutinesDebugging`, `DebugStackTraces` and `CoroutinesKnit`.

Another solution could be to somehow highlight these stackframes visually in user stack traces, e.g. by getting rid of package FQN. It won't clutter IDEA experience, because this class is internal anyway, but will provide a clear visual separator:
image

qwwdfsad (changes requested):

dkhalanskyjb:
I tested IDE navigation in this program: https://github.com/dkhalanskyjb/testRepo/tree/check-stacktrace-navigation. I couldn't make it break and can't imagine a reason for the current solution to fail.

Though the namespace pollution does worry me, it is the case that at the moment when the sources for a dependency are not yet downloaded, IDE is unable to determine where `CoroutineDebugging.kt` is, so relying on actually having classfiles with correct names seems inevitable.

kotlinx-coroutines-core/jvm/src/debug/CoroutineDebugging.kt
```diff
@@ -0,0 +1,64 @@
...
+1) the artificial stack frames look pretty, and
+2) the IDE reliably navigates to this file. */
+package _COROUTINE
+
+/**
+ * A collection of artificial stack trace elements to be included in stack traces by the coroutines machinery.
+ *
+ * There are typically two ways in which one can encounter an artificial stack frame:
+ * 1. By using the debug mode, via the stacktrace recovery mechanism; see
+ * https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/topics/debugging.md#stacktrace-recovery. The usual way
```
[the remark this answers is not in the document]
qwwdfsad: It would be nice if you could update the sample in this documentation as well

qwwdfsad (changes requested):

qwwdfsad (approved):
Great job! 👍

PR #2463, opened 2020-12-26
Using the same logic as packNanos in the code below

slavonnet:
Signed-off-by: slavon <v.badalyan@open-bs.ru>


slavonnet:
@qwwdfsad please rewiew

qwwdfsad (approved):
Thanks for spotting!

PR #2466, opened 2020-12-29
Clarify yield documentation

### LINKED ISSUE #2464: yield function doc
The `yield` function doc states at the first line:
> Yields the thread (or thread pool) of the current coroutine dispatcher to other coroutines to run if possible.

The problem is that it makes an impression that it's applied globally regardless of the current dispatcher. In spite of the last sentence points to the dispatcher-sensitivity, people get confused anyway. 

We need to clarify it so people see that the yield function only affects coroutines running on the same dispatcher.

qwwdfsad:
Fixes #2464


elizarov:
👍 Nice clarification. Thanks.

PR #2467, opened 2020-12-29
Remove internal 'checkCompletion' function and replace it with the id…

qwwdfsad:
…entical 'ensureActive' extension to reduce code duplication


elizarov (approved):
Nice catch!

PR #2472, opened 2021-01-07
Update flow.md

jerryyanmj:
I think it worth mentioning that both flowOn and buffer both launch a different conroutine for emission.


qwwdfsad (commented):
Thanks for the suggestion! Unfortunately, I'd prefer not to re-duplicate the very first sentence of their respective KDocs and avoid additional not-really-useful lines of output in an already big enough guide.

Issue #2283, opened 2020-10-07
[native] `Dispatchers.Main.immediate` behavior on main thread


andersio:
Kotlin 1.4.10, Coroutine 1.3.9-native-mt-2

Not sure if this is intentional for `Dispatchers.Main.immediate` — `isDispatchNeeded()` returns `true` when called from main thread.

Seems like it should be `!invokedImmediately || !isMainThread()`?

https://github.com/Kotlin/kotlinx.coroutines/blob/1b0e196bc557f03fd9e13b81edb6997bddeb38e7/kotlinx-coroutines-core/nativeDarwin/src/Dispatchers.kt#L31

Code snippet:
```kotlin
check(NSThread.isMainThread)
println("@@@ isDispatchNeeded ${Dispatchers.Main.immediate.isDispatchNeeded(EmptyCoroutineContext)}")
```

elizarov:
That looks like an oversight. Thanks.

Thomas-Vos:
Does this bug cause any issues when using `Dispatchers.Main.immediate`? Is it safe to use without a fix?

andersio:
@Thomas-Vos I don't think it isn't safe to use, since the bugged logic can lead main thread bound resumptions/dispatches to be called on a background thread.

I am using a replica of `Dispatchers.Main.immediate` until it is fixed.

```kotlin
@SharedImmutable
actual val ImmediateMainDispatcher: CoroutineDispatcher = _ImmediateMainDispatcher()

@OptIn(InternalCoroutinesApi::class)
open class _ImmediateMainDispatcher(
    private val delegate: MainCoroutineDispatcher = Dispatchers.Main
): CoroutineDispatcher(), Delay by (delegate as Delay) {
    init { freeze() }

    override fun dispatch(context: CoroutineContext, block: Runnable)
        = delegate.dispatch(context, block)

    override fun dispatchYield(context: CoroutineContext, block: Runnable)
        = delegate.dispatchYield(context, block)

    override fun scheduleResumeAfterDelay(timeMillis: Long, continuation: CancellableContinuation<Unit>)
        = (delegate as Delay).scheduleResumeAfterDelay(timeMillis, continuation)

    override fun invokeOnTimeout(timeMillis: Long, block: Runnable, context: CoroutineContext): DisposableHandle
        = (delegate as Delay).invokeOnTimeout(timeMillis, block, context)

[...]

PR #2476, opened 2021-01-12
[native-mt] Fix the errornous isDispatchNeeded check in Dispatchers.Main.immediate on Darwin platforms.

andersio:
Resolves https://github.com/Kotlin/kotlinx.coroutines/issues/2283

First commit includes a test case that fails in both macosx64 background & main tests.


qwwdfsad (approved):
Thanks!

PR #2477, opened 2021-01-12
[native-mt] Fix ObjC autorelease object leaks with native-mt dispatchers

### LINKED ISSUE #2322: [native-mt] Darwin single thread dispatchers leak ObjC autorelease references
Kotlin: 1.4.10
Coroutines: 1.3.9-native-mt-2

Also filed as https://youtrack.jetbrains.com/issue/KT-42822 for the K/N runtime.

KT-42822 is about K/N `Worker`s not wrapping its work items with an autorelease pool. So when Kotlin code calls into ObjC/Swift on a `Worker`, any `autorelease` reference leaks unconditionally, e.g., any `id` or `NSObject*` return values.

However, even if KT-42822 is fixed, a fix at the coroutines library is still necessary, since kotlinx-coroutines-core runs its own EventLoop over the one in `Worker`.

The native single thread `CoroutineDispatcher` wraps the entire `EventLoop` with an autorelease pool, but not individual blocks as they get executed. This means autorelease references in these blocks would be pushed to a pool that is never drained until the `EventLoop` is closed.

So in other words, autorelease references are effectively leaked on any long living dispatchers created by `newSingleThreadContext()`.

https://github.com/Kotlin/kotlinx.coroutines/blob/1b0e196bc557f03fd9e13b81edb6997bddeb38e7/kotlinx-coroutines-core/native/src/Workers.kt#L48-L50

This could be solved by either:

* `native-mt` single-thread dispatchers wrap individual blocks with `autoreleasepool` before executing them; or
* wrap `processNextEvent()` with `autoreleasepool {}` maybe?

   https://github.com/Kotlin/kotlinx.coroutines/blob/1b0e196bc557f03fd9e13b81edb6997bddeb38e7/kotlinx-coroutines-core/native/src/Builders.kt#L77-L83

andersio:
Resolves #2322

While 1.4.30-M1 fixed `Worker.execute()` leaking ObjC autoreleasing objects when running native code on a K/N Worker thread ([KT-42822](https://youtrack.jetbrains.com/issue/KT-42822)), it does not address the issue at KoltinX Coroutines framework level.

More specifically, when tasks on an EventLoop call into native code, any autoreleasing object — especially many ObjC method return values — is effectively leaked <sup>[1]</sup>. This is because [the first available autoreleasepool in the call stack is in `workerMain()`, and it wraps around a most likely never-returning <sup>[2]</sup> `runEventLoop()`](https://github.com/Kotlin/kotlinx.coroutines/blob/89a9f7d53c7add3492d0846b5dbd623ab5b5d81d/kotlinx-coroutines-core/native/src/Workers.kt#L46-L52). 

This PR introduced a failing test case of this bug, and fixed it by wrapping all `Runnable.run()` calls in `EventLoopImplBase` with an autoreleasepool (or nothing for non-Darwin targets). This is equivalent to the [DISPATCH_AUTORELEASE_FREQUENCY_WORK_ITEM](https://developer.apple.com/documentation/dispatch/dispatch_autorelease_frequency_t/dispatch_autorelease_frequency_work_item)<sup>[3]</sup> in GCD.

An alternative to the per-item wrapping is integrating pool push/pop into `runEventLoop()`, with heruistics based on the park time returned by `processNextEvent()`. Not familiar enough about the particulars to try this out though.

_<sup>[1]</sup> ... unless an autoreleasepool is declared in user code explicitly, which is a rare practice except for memory optimization._
_<sup>[2]</sup> `Dispatchers.Default`, for example._
_<sup>[3]</sup> ... which is not the default behaviour. GCD threads inherit autorelease pool management from `NSRunLoop`, which recreates pools for each runloop iteration._

### DIFF
--- a/kotlinx-coroutines-core/common/src/EventLoop.common.kt
+++ b/kotlinx-coroutines-core/common/src/EventLoop.common.kt
@@ -530,6 +530,15 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
 
 internal expect fun createEventLoop(): EventLoop
 
+/**
+ * Used by Darwin targets to wrap a [Runnable.run] call in an Objective-C Autorelease Pool. It is a no-op on JVM, JS and
+ * non-Darwin native targets.
+ *
+ * Coroutines on Darwin targets can call into the Objective-C world, where a callee may push a to-be-returned object to
+ * the Autorelease Pool, so as to avoid a premature ARC release before it reaches the caller. This means the pool must
+ * be eventually drained to avoid leaks. Since Kotlin Coroutines does not use [NSRunLoop], which provides automatic
+ * pool management, it must manage the pool creation and pool drainage manually.
+ */
 internal expect inline fun platformAutoreleasePool(crossinline block: () -> Unit)
 
 internal expect fun nanoTime(): Long
--- a/kotlinx-coroutines-core/common/src/EventLoop.common.kt
+++ b/kotlinx-coroutines-core/common/src/EventLoop.common.kt
@@ -66,7 +66,9 @@ internal abstract class EventLoop : CoroutineDispatcher() {
     public fun processUnconfinedEvent(): Boolean {
         val queue = unconfinedQueue ?: return false
         val task = queue.removeFirstOrNull() ?: return false
-        task.run()
+        platformAutoreleasePool {
+            task.run()
+        }
         return true
     }
     /**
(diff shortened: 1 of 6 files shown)


elizarov (approved):
Looks great to me! Thanks a lot.

elizarov:
Does not compile: 
`Expected function 'platformAutoreleasePool' has no actual declaration in module <kotlinx-coroutines-core> for JVM`
It needs to be implemented as a NOP both for JVM and JS.

qwwdfsad (approved):
Great job! 👍  Thanks for a detailed write-up and isolated unit tests, it really made the review process easy and pleasant

PR #2487, opened 2021-01-21
Make internal functions public in already internal CoroutineScheduler…

qwwdfsad:
… in order to get prettier stacktraces without "$kotlinx_coroutines_core" in function names


elizarov (approved):
👍

PR #2489, opened 2021-01-21
Update Merge.kt

yunjoopark:
- flatMapConcat is a shortcut for `map().flattenConcat()` and flatMapMerge is a shorcut for `map().flattenMerge()`
- fix typo in Merge.kt

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Merge.kt
@@ -34,7 +34,7 @@ public val DEFAULT_CONCURRENCY: Int = systemProp(DEFAULT_CONCURRENCY_PROPERTY_NA
  * Transforms elements emitted by the original flow by applying [transform], that returns another flow,
  * and then concatenating and flattening these flows.
  *
- * This method is is a shortcut for `map(transform).flattenConcat()`. See [flattenConcat].
+ * This method is a shortcut for `map(transform).flattenConcat()`. See [flattenConcat].
  *
  * Note that even though this operator looks very familiar, we discourage its usage in a regular application-specific flows.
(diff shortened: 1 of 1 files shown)


docs/flow.md
```diff
@@ -1117,7 +1117,7 @@ as such, there is a family of flattening operators on flows.
 
 #### flatMapConcat
 
-Concatenating mode is implemented by [flatMapConcat] and [flattenConcat] operators. They are the most direct
+Concatenating mode is implemented by [map] and [flattenConcat] operators. They are the most direct
```
elizarov, line 1120: The original text was correct. Why do you think it needs to be changed?

docs/flow.md
```diff
@@ -1166,7 +1166,7 @@ The sequential nature of [flatMapConcat] is clearly seen in the output:
 
 Another flattening mode is to concurrently collect all the incoming flows and merge their values into
 a single flow so that values are emitted as soon as possible.
-It is implemented by [flatMapMerge] and [flattenMerge] operators. They both accept an optional 
+It is implemented by [map] and [flattenMerge] operators. They both accept an optional
```
elizarov, line 1169: Same here.

elizarov (changes requested):

docs/flow.md
yunjoopark: I think it is a description of `flatMapConcat`. According to the description of `flatMapConcat`, it is a shortcut for `map().flattenConcat(). 
So, I suggested the PR.
If not, what is the original context for?

elizarov: It is a description of the "concatenating mode", which is available when you use `flatMapConcat` and `flattenConcat`.

yunjoopark: Thanks for your explanation. Let me exclude the wrong changes from the pr.

elizarov (approved):
Thanks!

PR #2492, opened 2021-01-25
Kotlin DSL - 'example-frontend-js'

turansky:
Part of #1938


js/example-frontend-js/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
EdwarDDay, line 2: Should this be changed to 2021?

turansky: Practice, used in Kotlin - update year in all files in separate commit

qwwdfsad:
Great! Superseded by #2543

PR #2497, opened 2021-01-26
Fix toString representation of JobListenableFuture

vadimsemenov:
Also add tests for toString.
I've created a new test file but happy to move tests into the main test file if that's preferable.


qwwdfsad (approved):
Good job!
Merged manually into PR 2422

PR #2501, opened 2021-01-27
Clarify docs about restriction on MutableSharedFlow function's onBufferOverflow parameter.

zach-klippenstein:
The existing docs are confusing to me – it's not clear that "supported" in this case means "passing a value other than the default is supported". I've also included an explanation as to why this restriction makes sense.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -221,9 +221,11 @@ public interface MutableSharedFlow<T> : SharedFlow<T>, FlowCollector<T> {
  * @param replay the number of values replayed to new subscribers (cannot be negative, defaults to zero).
  * @param extraBufferCapacity the number of values buffered in addition to `replay`.
  *   [emit][MutableSharedFlow.emit] does not suspend while there is a buffer space remaining (optional, cannot be negative, defaults to zero).
- * @param onBufferOverflow configures an action on buffer overflow (optional, defaults to
- *   [suspending][BufferOverflow.SUSPEND] attempts to [emit][MutableSharedFlow.emit] a value,
- *   supported only when `replay > 0` or `extraBufferCapacity > 0`).
+ * @param onBufferOverflow configures an action on buffer overflow. Optional, defaults to
+ *   [suspending][BufferOverflow.SUSPEND] attempts to [emit][MutableSharedFlow.emit] a value.
+ *   Values other than [BufferOverflow.SUSPEND] are supported only when `replay > 0` or `extraBufferCapacity > 0` - in other words, when the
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -221,9 +221,11 @@ public interface MutableSharedFlow<T> : SharedFlow<T>, FlowCollector<T> {
  * @param replay the number of values replayed to new subscribers (cannot be negative, defaults to zero).
  * @param extraBufferCapacity the number of values buffered in addition to `replay`.
  *   [emit][MutableSharedFlow.emit] does not suspend while there is a buffer space remaining (optional, cannot be negative, defaults to zero).
- * @param onBufferOverflow configures an action on buffer overflow (optional, defaults to
- *   [suspending][BufferOverflow.SUSPEND] attempts to [emit][MutableSharedFlow.emit] a value,
- *   supported only when `replay > 0` or `extraBufferCapacity > 0`).
+ * @param onBufferOverflow configures an action on buffer overflow. Optional, defaults to
+ *   [suspending][BufferOverflow.SUSPEND] attempts to [emit][MutableSharedFlow.emit] a value.
+ *   Values other than [BufferOverflow.SUSPEND] are supported only when `replay > 0` or `extraBufferCapacity > 0` - in other words, when the
+ *   flow has no buffer, then [emit][MutableSharedFlow.emit] will always suspend unless there is at least one subscriber ready to accept the
```
elizarov, line 227: > when the flow has no buffer, then emit will always suspend **unless** there is at least one subscriber ready to accept the

This should be "will suspend **if** there is at least one subscriber". If there are no subscribes, then emit never suspends. It just drops the emitted value.

elizarov (changes requested):

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
zach-klippenstein: How are these?
1. "then [emit][MutableSharedFlow.emit] will always suspend when there is at least one subscriber and at least one subscriber is not ready to accept the new value. It will never suspend when there are no subscribers."
1. "then [emit][MutableSharedFlow.emit] will always suspend when any subscriber is not ready to accept the new value. It will never suspend when there are no subscribers."
1. "then [emit][MutableSharedFlow.emit] will always suspend when any subscriber is exerting backpressure. It will never suspend when there are no subscribers."

zach-klippenstein: @elizarov Let me know what you think of the above wordings and I'll update the PR.

elizarov:
It turned to be a challenge to rephrase. I've added my own clarifications on top of yours in #2560. Closing this PR.

PR #2505, opened 2021-01-28
Promote CoroutineStart.UNDISPATCHED to non experimental API

### LINKED ISSUE #1393: Promote CoroutineStart.UNDISPATCHED to non experimental API
I use it quite often in places where I know it's safe to use, with the following import alias:
```kotlin
import kotlinx.coroutines.CoroutineStart.UNDISPATCHED as Undispatched
```

It'd be great to not have to have to add `@UseExperimental(ExperimentalCoroutinesApi::class)` to each usage.

I'm interested to know of any considerations that could justify leaving it experimental, postponing that decision or replacing it with something else (like a `startDispatched: Boolean = true` parameter).

qwwdfsad:
Fixes #1393


elizarov (approved):
🚀 Good to go.

PR #2511, opened 2021-02-01
Cleanup continuation field in lazily-started coroutines after the start to reduce potential memory pressure

qwwdfsad:
Also, as long as onStart is our own private API, replace checkNotNull with !! operator to save few bytecodes


qwwdfsad:
Also, I've found that lazy broadcast is broken, but decided not to touch it because we are planning to deprecate it pretty soon anyway

PR #2512, opened 2021-02-01
Simplify internal machinery

qwwdfsad:


kotlinx-coroutines-core/common/src/AbstractCoroutine.kt
```diff
@@ -46,35 +44,21 @@ public abstract class AbstractCoroutine<in T>(
     @Suppress("LeakingThis")
     public final override val context: CoroutineContext = parentContext + this
 
+    init {
+        /*
+         * Setup parent-child relationship between the parent in the context
+         * and the current coroutine.
+         * It may cause this coroutine to become cancelled if the parent is already cancelled
+         */
+        initParentJobInternal(parentContext[Job])
```
elizarov, line 53: This is a very scary change. `AbstractCoroutine` has lots of inheriting classes. This change means that this AbstractCoroutine may get cancelled while the fields in the derived classes have not initialized yet. For example, `RxCompletableCoroutine` may not have its `subscriber` field initialized yet, when `onCancelled` is invoked during cancellation.

elizarov (changes requested):

qwwdfsad:
Instead of copy-pasting it as discussed, I've stick with a slightly better approach where the constructor demands to choose one or the other option

elizarov (approved):
Ok

PR #2513, opened 2021-02-01
Do not initialize CoroutineExceptionHandler in Job <clinit>

qwwdfsad:
* It slightly cuts down startup time
    * It simplifies the code in general
    * It do not serve its purpose, coroutines are already protected from StackOverflowError with nested event-loops where applicable


elizarov (approved):
Looks good.

PR #2517, opened 2021-02-02
Properly nest ThreadContextElement

### LINKED ISSUE #2195: ThreadContextElements are not nested properly
Hello!

For some of the implementations of the ThreadContextElements the order of invocation is important. Here is one example that we are struggling with and don't know how to solve cleanly:
- Kotlin gRPC library has [GrpcContextElement](https://github.com/grpc/grpc-kotlin/blob/master/stub/src/main/java/io/grpc/kotlin/GrpcContextElement.kt) that propagates gRPC context.
- The element above attaches and detaches gRPC context
- We have additional gRPC context elements that we want to propagate. This is to support opencensus trace contexts which uses gRPC context for the implementation.
- Kotlin coroutines library invokes `restoreThreadContext` in exactly the same order as `updateThreadContext`.
- But the order in which detach calls happen need to be the opposite of attach calls (required by gRPC).

Is there a reason why the order for `restoreThreadContext` is not the opposite of `updateThreadContext`?

qwwdfsad:
* Restore the context in the reverse order of update, so they are properly nested into each other
    * Also, do a minor cleanup

Fixes #2195

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/ThreadContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ThreadContext.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package kotlinx.coroutines.internal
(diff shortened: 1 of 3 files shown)


qwwdfsad:
I've also measured the potential overhead and it turned out that even tho we allocate slightly more, the average throughput of context restoration (`ChannelSinkBenchmark.channelPipelineTwoThreadLocals`) has been improved by 20%!

For single-shot coroutines without hot loops, the performance is roughly the same (few percents better, actually) with the cost of an additional small allocation

kotlinx-coroutines-core/jvm/src/internal/ThreadContext.kt
```diff
@@ -11,13 +11,22 @@ import kotlin.coroutines.*
...
-    fun take() = a[i++]
-    fun start() { i = 0 }
+    fun append(element: ThreadContextElement<*>, value: Any?) {
+        values[i] = value
+        elements[i++] = element as ThreadContextElement<Any?>
+    }
+
+    fun restore(context: CoroutineContext) {
+        for (i in elements.indices.reversed()) {
+            elements[i]?.restoreThreadContext(context, values[i])
```
elizarov, line 27: Minor nit: writing `elements[i]!!` will make it slightly easier to understand (it can never be `null`).

elizarov (approved):
Great improvement.

PR #2525, opened 2021-02-05
Test invariant that IDEA evaluator depends on

### LINKED ISSUE #2522: Coroutine scope is not always captured in coroutine stack frame
For coroutine stack frame's `this` object sometimes you can't see `L$0` in its field list.
Screen Shot 2021-01-29 at 12 07 12 PM
Screen Shot 2021-01-29 at 12 07 33 PM

# Note
You won't be able to reproduce this example in the debugger, because synthetic `this` objects are not shown in the variables view. I tweaked the code to illustrate the problem.

# Why is it important
In Kotlin debugger we need to remap `this` object from `Continuation` to `CoroutineScope` for coroutine stack frames. This remap is needed to keep information in the variables view consistent with type hints, and also without it, evaluation in some cases won't work properly. To solve the problem we try to get `CoroutineScope` from the `completion` field, however, I assume that it is better to use `L$0` for this purpose.

qwwdfsad:
Fixes #2522


elizarov (approved):
Great write-up!

PR #2528, opened 2021-02-05
Add bi-directional cancellation for Task.await and Task.asDeferred

### LINKED ISSUE #2527: Proposal: Add await for play-services integration for Task constructed with CancellationToken
The current `play-services` integration provides `Task.await` that allows waiting for the result of an already created `Task`. This function supports cancellation, but not in both directions: If `await` is cancelled, then the underlying `Task` is not cancelled.

Some methods of creating `Task`s don't support cancellation in this manner, but some do by creating the task with a `CancellationToken`. For an example: https://developers.google.com/places/android-sdk/reference/com/google/android/libraries/places/api/net/FindAutocompletePredictionsRequest.Builder#setCancellationToken(com.google.android.gms.tasks.CancellationToken)

My proposal is to add another `await` method, that operates on a lambda that constructs a `Task` given a `CancellationToken`.

alexvanyo:
This PR includes my proposed addition to the `play-services` integration to allow bi-directional cancellation for `Task.await` and `Task.asDeferred`, which in particular provides the hook to cancel a task via a `CancellationToken` (which fixes #2527)

Both `Task.asDeferred` and `Task.await` now each have a new overload that takes a passed in `CancellationTokenSource`. If the `Deferred`/`await` is cancelled, or if the receiving `Task` is cancelled, then the `CancellationTokenSource` will also be cancelled. In particular, this means that if the `Task` will be cancelled when the `CancellationTokenSource` is cancelled, cancelling the `Deferred`/`await` will cancel the underlying `Task`.

The existing implementations of `Task.asDeferred` and `Task.await` call into the new implementations, with a newly constructed `CancellationTokenSource` that will result in a no-op upon being cancelled.

### DIFF
--- a/integration/kotlinx-coroutines-play-services/src/Tasks.kt
+++ b/integration/kotlinx-coroutines-play-services/src/Tasks.kt
@@ -8,7 +8,6 @@ package kotlinx.coroutines.tasks
 
 import com.google.android.gms.tasks.*
 import kotlinx.coroutines.*
-import kotlin.contracts.*
 import kotlin.coroutines.*
 
 /**
@@ -38,83 +37,113 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
 
 /**
  * Converts this task to an instance of [Deferred].
+ *
+ * Prefer passing the corresponding [CancellationTokenSource] if the [Task] can be created with a [CancellationToken]
+ * to support bi-directional cancellation.
+ *
  * If task is cancelled then resulting deferred will be cancelled as well.
  */
-public fun <T> Task<T>.asDeferred(): Deferred<T> {
+public fun <T> Task<T>.asDeferred(): Deferred<T> = asDeferred(CancellationTokenSource())
+
+/**
+ * Converts this task to an instance of [Deferred] with a [CancellationTokenSource] to control cancellation.
+ *
+ * If the task is cancelled, then the [cancellationTokenSource] and the resulting deferred will be cancelled.
+ * If the deferred is cancelled, then the [cancellationTokenSource] will be cancelled.
+ * If the [cancellationTokenSource] is cancelled, then the deferred will be cancelled.
+ */
+@ExperimentalCoroutinesApi
+public fun <T> Task<T>.asDeferred(cancellationTokenSource: CancellationTokenSource): Deferred<T> {
+    val deferred = CompletableDeferred<T>()
+
     if (isComplete) {
         val e = exception
-        return if (e == null) {
-            @Suppress("UNCHECKED_CAST")
-            CompletableDeferred<T>().apply { if (isCanceled) cancel() else complete(result as T) }
+        if (e == null) {
+            if (isCanceled) {
+                deferred.cancel()
+            } else {
+                @Suppress("UNCHECKED_CAST")
+                deferred.complete(this.result as T)
+            }
         } else {
-            CompletableDeferred<T>().apply { completeExceptionally(e) }
+            deferred.completeExceptionally(e)
+        }
+    } else if (cancellationTokenSource.token.isCancellationRequested) {
+        // The task hasn't completed, yet cancellation was already requested.
+        // Interpret this by cancelling immediately (no way to cancel the task)
+        deferred.cancel()
+    } else {
+        addOnCompleteListener {
+            val e = it.exception
+            if (e == null) {
+                @Suppress("UNCHECKED_CAST")
+                if (it.isCanceled) deferred.cancel() else deferred.complete(it.result as T)
+            } else {
+                deferred.completeExceptionally(e)
+            }
+        }
+
+        cancellationTokenSource.token.onCanceledRequested {
+            deferred.cancel()
         }
     }
 
-    val result = CompletableDeferred<T>()
-    addOnCompleteListener {
-        val e = it.exception
-        if (e == null) {
-            @Suppress("UNCHECKED_CAST")
-            if (isCanceled) result.cancel() else result.complete(it.result as T)
-        } else {
-            result.completeExceptionally(e)
+    deferred.invokeOnCompletion {
+        if (it is CancellationException) {
+            cancellationTokenSource.cancel()
         }
     }
-    return result
+
+    return deferred
 }
 
 /**
  * Awaits for completion of the task without blocking a thread.
  *
- * Prefer [awaitTask] if a [Task] can be constructed with a [CancellationToken], to allow propagating cancellation to
- * that [Task].
+ * Prefer passing the corresponding [CancellationTokenSource] if the [Task] can be created with a [CancellationToken]
+ * to support bi-directional cancellation.
  *
  * This suspending function is cancellable.
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * stops waiting for the completion stage and immediately resumes with [CancellationException].
- *
- * @see [awaitTask]
  */
-public suspend fun <T> Task<T>.await(): T = awaitTask { this }
+public suspend fun <T> Task<T>.await(): T = await(CancellationTokenSource())
 
 /**
- * Awaits for completion of the created task via [block] without blocking a thread.
- *
- * Prefer this method over [Task.await] if a [Task] can be constructed with a [CancellationToken], to cancel the task
- * if this function is cancelled.
+ * Awaits for completion of the task with a [CancellationTokenSource] to control cancellation.
  *
  * This suspending function is cancellable.
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
- * stops waiting for the completion stage and immediately resumes with [CancellationException].
+ * cancels the [cancellationTokenSource] and throws a [CancellationException].
  *
- * @see [Task.await]
+ * If the task is cancelled, then [cancellationTokenSource] will be canceled and this function will throw a
+ * [CancellationException].
+ * If the [cancellationTokenSource] is cancelled, then this function will throw a [CancellationException].
  */
-@OptIn(ExperimentalContracts::class)
 @ExperimentalCoroutinesApi
-public suspend inline fun <T> awaitTask(block: (CancellationToken) -> Task<T>): T {
-    contract { callsInPlace(block, InvocationKind.EXACTLY_ONCE) }
-
-    val cancellation = CancellationTokenSource()
-    val task = block(cancellation.token)
-
+public suspend fun <T> Task<T>.await(cancellationTokenSource: CancellationTokenSource): T {
     // fast path
-    if (task.isComplete) {
-        val e = task.exception
+    if (isComplete) {
+        val e = exception
         return if (e == null) {
-            if (task.isCanceled) {
-                throw CancellationException("Task $task was cancelled normally.")
+            if (isCanceled) {
+                cancellationTokenSource.cancel()
+                throw CancellationException("Task $this was cancelled normally.")
             } else {
                 @Suppress("UNCHECKED_CAST")
-                task.result as T
+                result as T
             }
         } else {
             throw e
         }
+    } else if (cancellationTokenSource.token.isCancellationRequested) {
+        // The task hasn't completed, yet cancellation was already requested.
+        // Interpret this by throwing immediately (no way to cancel the task)
+        throw CancellationException("Cancellation was already requested")
     }
 
     return suspendCancellableCoroutine { cont ->
-        task.addOnCompleteListener {
+        addOnCompleteListener {
             val e = it.exception
             if (e == null) {
                 @Suppress("UNCHECKED_CAST")
@@ -123,8 +152,12 @@ public suspend inline fun <T> awaitTask(block: (CancellationToken) -> Task<T>): 
                 cont.resumeWithException(e)
             }
         }
+        cancellationTokenSource.token.onCanceledRequested {
+            cont.cancel()
+        }
+
         cont.invokeOnCancellation {
-            cancellation.cancel()
+            cancellationTokenSource.cancel()
         }
     }
 }
--- a/integration/kotlinx-coroutines-play-services/src/Tasks.kt
+++ b/integration/kotlinx-coroutines-play-services/src/Tasks.kt
@@ -67,41 +67,19 @@ public fun <T> Task<T>.asDeferred(): Deferred<T> {
 /**
  * Awaits for completion of the task without blocking a thread.
  *
+ * Prefer [awaitTask] if a [Task] can be constructed with a [CancellationToken], to allow propagating cancellation to
+ * that [Task].
+ *
  * This suspending function is cancellable.
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * stops waiting for the completion stage and immediately resumes with [CancellationException].
+ *
+ * @see [awaitTask]
  */
-public suspend fun <T> Task<T>.await(): T {
-    // fast path
-    if (isComplete) {
-        val e = exception
-        return if (e == null) {
-            if (isCanceled) {
-                throw CancellationException("Task $this was cancelled normally.")
-            } else {
-                @Suppress("UNCHECKED_CAST")
-                result as T
-            }
-        } else {
-            throw e
-        }
-    }
-
-    return suspendCancellableCoroutine { cont ->
-        addOnCompleteListener {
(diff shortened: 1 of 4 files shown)


integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -105,3 +99,31 @@ public suspend fun <T> Task<T>.await(): T {
...
+ * Prefer this method over [Task.await] if a [Task] can be constructed with a [CancellationToken], to cancel the task
+ * if this function is cancelled.
+ *
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
+ * stops waiting for the completion stage and immediately resumes with [CancellationException].
+ *
+ * @see [Task.await]
+ */
+@OptIn(ExperimentalContracts::class)
```
qwwdfsad, line 115: The general idea "let's support bi-directional cancellation" looks good and worth supporting.

The proposed method though is quite verbose and non-idiomatic (submitted tests show it clearly), as well as hardly discoverable for newcomers.

I would like to keep things consistent and instead provide an overload of `await` with `CancellationToken`.
Also, then it's possible to simplify a code and use `CancellableContinuation.invokeOnCancellation` similarly to how it's done in other integrations, e.g. `Future.await()`

qwwdfsad (changes requested):
Also, please don't forget to run `./gradlew apiDump` in order to updated our binary compatibility data (you can read more [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs)).

If you are not ready to walk this extra mile, that's completely okay, just notify us

integration/kotlinx-coroutines-play-services/src/Tasks.kt
alexvanyo: I definitely agree that the extension on a lambda is pretty awkward, although I don't believe it will be possible to provide an overload of `Task.await` that takes a `CancellationToken` unless I'm missing something.

In cases like [`FusedLocationProviderClient.getCurrentLocation()`](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderClient#getCurrentLocation(int,%20com.google.android.gms.tasks.CancellationToken)) and [`FindAutocompletePredictionsRequest.Builder.setCancellationToken`](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderClient#public-tasklocation-getcurrentlocation-int-priority,-cancellationtoken-token), the `CancellationToken` is required in order to create the `Task`, so we need to setup the coroutine-managed `CancellationTokenSource` to give its `CancellationToken` to the caller.

I think the ideal solution would be to define a static extension on `Tasks` that takes the `block` to construct a `Task` with the `CancellationToken`, but that's blocked by https://youtrack.jetbrains.com/issue/KT-11968, so instead we could have a top-level function along the lines of `awaitTask` or `awaitConstructedTask`. I just pushed up a second commit with that alternative, which also changes `Task.await()` to call into `awaitTask`, since `Task.await` is the special case where the `Task` doesn't support being cancelled.

Let me know what you think!

alexvanyo:
Thanks for the review @qwwdfsad , this should be ready for another look!

The `apiDump` should also be good to go as well!

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: >the CancellationToken is required in order to create the Task, so we need to setup the coroutine-managed CancellationTokenSource to give its CancellationToken to the caller.

It can be created in advance, e.g.:
```
val token = CancellationTokenSource()
val task = createTask(token)
task.await(token)
```

The proposed solution has a significant restriction -- it confines the creation and cancalltion management into a single encapsulated function.
Meaning that if somebody _already_ has a cancellable task (and maybe invokes `.cancel()` manually from a few places) in their API, then it won't be possible for them to leverage cancellable await at all.

With the `await(token)`, on the other hand, it's easy to emulate the proposed API by introducing a small project-specific helper that doesn't even have to use any coroutines API beyond `await(task)`.

My vote is for providing a flexible and usable building block, rather than a specific-purpose non-extensible function.

qwwdfsad (changes requested):

integration/kotlinx-coroutines-play-services/src/Tasks.kt
alexvanyo: Ah, I see now, thank you for the insight! We can also do a very similar approach to support bi-directional cancellation for `asDeferred()` as well, via a passed in `CancellationTokenSource` which I think makes sense to tackle here as well.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -44,45 +37,97 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
+ *
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
+ * cancels the [cancellationTokenSource] and throws a [CancellationException].
+ *
+ * If the task is cancelled, then [cancellationTokenSource] will be canceled and this function will throw a
+ * [CancellationException].
+ * If the [cancellationTokenSource] is cancelled, then this function will throw a [CancellationException].
+ */
+@ExperimentalCoroutinesApi
```
alexvanyo, line 123: Both of these could be combined to have `cancellationTokenSource` be a default argument, but then I couldn't apply `@ExperimentalCoroutinesApi` to just the new overload.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -91,17 +136,28 @@ public suspend fun <T> Task<T>.await(): T {
         } else {
             throw e
         }
+    } else if (cancellationTokenSource.token.isCancellationRequested) {
+        // The task hasn't completed, yet cancellation was already requested.
+        // Interpret this by throwing immediately (no way to cancel the task)
+        throw CancellationException("Cancellation was already requested")
```
alexvanyo, line 142: I added handling here (and for `asDeferred`) for the case where the `CancellationTokenSource` is unrelated to the `Task`, in which case we have 3 things to coordinate:
- The `Task`'s cancellation, which we can listen for but can't directly cause
- The `CancellationTokenSource`'s cancellation, which we can listen for and cause
- The coroutine's cancellation, which we can listen for and cause

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -91,17 +136,28 @@ public suspend fun <T> Task<T>.await(): T {
         } else {
             throw e
         }
+    } else if (cancellationTokenSource.token.isCancellationRequested) {
```
alexvanyo, line 139: I decided here to check for the cancellation of the `CancellationTokenSource` after checking for if the `Task` was already completed. I could also see the argument for going the other way instead, where we throw `CancellationException` if the `cancellationTokenSource` is cancelled, even if the `Task` is already completed.

alexvanyo:
Thank you for the review and insights!

I switched the implementation to an overload of `Task.await` that takes a `CancellationTokenSource` argument, and also added a mirroring implementation for `Task.asDeferred` to also support bi-directional cancellation there.

I'm curious to hear your thoughts on the handling for when the passed-in `CancellationTokenSource` isn't directly related to the `Task`, since we can't guarantee that will be the case.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -44,45 +37,97 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
+ * If the [cancellationTokenSource] is cancelled, then this function will throw a [CancellationException].
+ */
+@ExperimentalCoroutinesApi
+public suspend fun <T> Task<T>.await(cancellationTokenSource: CancellationTokenSource): T {
     // fast path
     if (isComplete) {
         val e = exception
         return if (e == null) {
             if (isCanceled) {
+                cancellationTokenSource.cancel()
```
qwwdfsad, line 130: This line is not needed, if the original task is cancelled, then their source should be cancelled as well

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: Doesn't feel like this check should be present at all.

Consider the following scenario:
1) `isComplete` returned false
2) Task failed with an application-specific exception
3) Concurrently the cancellation has been requested.
4) `isCancellationRequested` returned `true

In such scenario, we will throw `CancellationException` instead of application-specific exception and it will effectively became lost

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -91,17 +136,28 @@ public suspend fun <T> Task<T>.await(): T {
...
             if (e == null) {
                 @Suppress("UNCHECKED_CAST")
-                if (isCanceled) cont.cancel() else cont.resume(result as T)
+                if (it.isCanceled) cont.cancel() else cont.resume(it.result as T)
             } else {
                 cont.resumeWithException(e)
             }
         }
+        cancellationTokenSource.token.onCanceledRequested {
+            cont.cancel()
```
qwwdfsad, line 156: Same reasoning here: we should wait until task is completed, so we are sure it hasn't thrown an exception after cancellation

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -44,45 +37,97 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
-        } else {
-            result.completeExceptionally(e)
+    deferred.invokeOnCompletion {
+        if (it is CancellationException) {
+            cancellationTokenSource.cancel()
         }
     }
-    return result
+
+    return deferred
```
qwwdfsad, line 97:
```suggestion
    return object : Deferred<T> by deferred {}
```

so no casting will occur

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -44,45 +37,97 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
 
-    val result = CompletableDeferred<T>()
-    addOnCompleteListener {
-        val e = it.exception
-        if (e == null) {
-            @Suppress("UNCHECKED_CAST")
-            if (isCanceled) result.cancel() else result.complete(it.result as T)
-        } else {
-            result.completeExceptionally(e)
+    deferred.invokeOnCompletion {
```
qwwdfsad, line 91: This `invokeOnCompletion` is never required, only the task can move it to the final state

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -44,45 +37,97 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
+            val e = it.exception
+            if (e == null) {
+                @Suppress("UNCHECKED_CAST")
+                if (it.isCanceled) deferred.cancel() else deferred.complete(it.result as T)
+            } else {
+                deferred.completeExceptionally(e)
+            }
+        }
+
+        cancellationTokenSource.token.onCanceledRequested {
```
qwwdfsad, line 86: Same reasoning about the fail-cancel race can be applied here

qwwdfsad (changes requested):

integration/kotlinx-coroutines-play-services/src/Tasks.kt
alexvanyo: Ah, that makes sense for the race

The reason why I was checking for the token being cancelled directly was to guard against a scenario where the `cancellationTokenSource` is cancelled but the `Deferred` will wait forever, if the `cancellationTokenSource` doesn't control the `Task`:

```kotlin
val cancellationTokenSource = CancellationTokenSource()
// Construct a task without the cancellation token source
val taskCompletionSource = TaskCompletionSource<Int>()

val deferred: Deferred<Int> = async(start = CoroutineStart.LAZY) {
    taskCompletionSource.task.await(cancellationTokenSource)
}

assertFalse(deferred.isCompleted)
cancellationTokenSource.cancel()

try {
    deferred.await() // will suspend forever
    fail("deferred.await() should be cancelled")
} catch (e: Exception) {
    assertTrue(e is CancellationException)
}
```

I will remove those checks to avoid the race condition, and explicitly note that the `CancellationTokenSource` should only be used to create the task, and not an external method of cancellation.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
alexvanyo: I agree that the `source` should be cancelled as well, but just want to note that there's nothing enforcing that the user doesn't just pass in a different `CancellationTokenSource` in which case the slow path and fast path behave slightly differently.

I'll remove this line and the tests I had for it, and make sure the docs specify that it's not intended to be used in that way.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
alexvanyo: Couldn't the `Deferred` also be cancelled via `Deferred.cancel()`?

integration/kotlinx-coroutines-play-services/src/Tasks.kt
alexvanyo: In the latest commit, I removed this and documented that an unrelated `CancellationTokenSource` is not supported.

integration/kotlinx-coroutines-play-services/test/TaskTest.kt
```diff
@@ -149,5 +149,277 @@ class TaskTest : TestBase() {
...
+        }
+
+        assertFalse(deferred.isCompleted)
+        taskCompletionSource.setException(TestException("something went wrong"))
+
+        deferred.await()
+    }
+
+    @Test
+    fun testCancelledAwaitCancellableTask() = runTest {
```
qwwdfsad, line 312: This test seems to fail consistently on our CI: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildAggregated/3519632?expandedTest=build%3A%28id%3A3519630%29%2Cid%3A44777

PTAL

qwwdfsad (commented):
Almost there! Please take care of the test and I'll handle the rest minors and will ensure it makes it into the next release

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: Right, good catch!
Moreover, it can fail with any exception due to structured concurrency, so the check `it is CancellationException` is also redundant.

Example of such failure:
```
val d = CompletableDeferred<Unit>()
d.invokeOnCompletion {
    println("Completed with: $it")
}
val child = GlobalScope.launch(d, CoroutineStart.UNDISPATCHED) {
    throw IllegalStateException()
}
```

alexvanyo: Updated!

integration/kotlinx-coroutines-play-services/test/TaskTest.kt
alexvanyo: I haven't been able to reproduce locally, but I think I narrowed down what's happening. The latest changes should pass!

The `Task` internals have some locks and also `post` something onto the main looper. For these tests, it looks like the looper is faked in `FakeAndroid`, which ends up doing a `GlobalScope.launch { ... }`, which I believe implies that the task won't synchronously know if it has been cancelled or not.

My solution right now is to just remove those assertions, as arguably checking that the task is actually cancelled is the `Task` library's responsibility. Since the main looper is faked, we could probably control the race, but that's probably not very helpful either since real usage wouldn't use a fake looper.

alexvanyo:
Just saw that the tests failed again, I went ahead and removed a similar racing assertion from the `Deferred` tests as well in d9d886b.

qwwdfsad (approved):
Great work, thanks!

I've opened #2786 with minor edits, will merge it and it's going to be the part of 1.5.1.

alexvanyo:
Awesome, thank you @qwwdfsad for your help and your time to review!

PR #2537, opened 2021-02-15
Rename TimeSource to SchedulerTimeSource due to KT-42625

qwwdfsad:


elizarov (approved):
👍

gmk57:
This error happens again with coroutines 1.5.0-RC. Will this fix be merged into 1.5.* branches?

elizarov:
As far as I can see, it has been renamed in both 1.4.3 and 1.5.0-RC. Are you sure you still experience this problem with those versions?

gmk57:
Yes, for example [this code](https://gist.github.com/gmk57/67591e0c878cedc2a318c10b9d9f4c0c) with wildcard imports (`import kotlinx.coroutines.*` + `import kotlin.time.*`) does not compile: `Unresolved reference: TimeSource`

I've double-checked and this happens with both 1.4.3 and 1.5.0-RC. So [the 1.4.3 release notes](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.4.3) probably relate to something else: they mention "internal class `TimeSource`" while I see an internal interface. Here it is in the source code [for 1.4.3](https://github.com/Kotlin/kotlinx.coroutines/blob/1.4.3/kotlinx-coroutines-core/jvm/src/TimeSource.kt) and [1.5.0-RC](https://github.com/Kotlin/kotlinx.coroutines/blob/1.5.0-RC/kotlinx-coroutines-core/jvm/src/TimeSource.kt).

elizarov:
Thanks. Created https://github.com/Kotlin/kotlinx.coroutines/issues/2691

PR #2562, opened 2021-03-03
Post-release cleanup

qwwdfsad:
* Properly handle files with spaces in deploy.sh
    * Update version shield


elizarov (approved):
Good stuff

PR #2567, opened 2021-03-08
Add Log4J2 MDC integration for coroutines

delitescere:
As Log4J2 is superior ;-) I thought it would be nice to start having Kotlin play nicer with it, particularly with coroutines.

This patch is a copy of the SLF4J integration. It only considers `ThreadContext` map (aka MDC) at the moment. The `ThreadContext` stack (aka NDC) is for later consideration.

Another future consideration includes Log4J2's [asynchronous logger](https://logging.apache.org/log4j/log4j-2.3/manual/async.html) capability in the context of coroutines or other non-blocking code.

NB: I did _not_ test against a JDK 1.6, sorry.


delitescere:
Apologies, will based against `develop`.

PR #2570, opened 2021-03-08
Make it more Informative

SahilHemnani777:
Added some information- why we did { Thread.sleep(2000L) }


qwwdfsad:
Thanks, but we already have this very line right next to `Thread.sleep(2000)`: `block main thread for 2 seconds to keep JVM alive`

PR #2574, opened 2021-03-09
Fix build script compilation issues

Tapchicoma:
They were deprecated long time ago and were removed in Gradle 7.

### DIFF
--- a/integration-testing/build.gradle
+++ b/integration-testing/build.gradle
@@ -92,10 +92,10 @@ task coreAgentTest(type: Test) {
 }
 
 dependencies {
-    testCompile "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
-    testCompile 'junit:junit:4.12'
-    npmTestCompile 'org.apache.commons:commons-compress:1.18'
-    npmTestCompile 'com.google.code.gson:gson:2.8.5'
+    testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+    testImplementation 'junit:junit:4.12'
+    npmTestImplementation 'org.apache.commons:commons-compress:1.18'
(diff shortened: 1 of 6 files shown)


integration-testing/build.gradle
```diff
@@ -92,8 +92,8 @@ task coreAgentTest(type: Test) {
 }
 
 dependencies {
-    testCompile "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
-    testCompile 'junit:junit:4.12'
+    testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+    testImplementation 'junit:junit:4.12'
     npmTestCompile 'org.apache.commons:commons-compress:1.18'
```
ALikhachev, line 97:
```suggestion
    npmTestImplementation 'org.apache.commons:commons-compress:1.18'
```
I think it's okay to also replace this (and dependencies below)

ALikhachev (approved):

integration-testing/build.gradle
Tapchicoma: I was not sure, as it seems custom configuration here

qwwdfsad (changes requested):
The build does not pass, Gradle couldn't resolve dependencies now, please address it and ensure that the project is buildable

integration-testing/build.gradle
ALikhachev: There are source sets matching these configurations (https://github.com/Kotlin/kotlinx.coroutines/blob/6af3161dbc0972b0b6e68372b9abdae722b02a4b/integration-testing/build.gradle#L16)

qwwdfsad (approved):
Thanks for taking care of it

PR #2581, opened 2021-03-13
Rework reusability control in cancellable contination

### LINKED ISSUE #2564: Memory leak in coroutines 1.4.x
I've encountered a memory leak which I believe is caused by a race condition in the coroutine library. 

I've simplified the leak down to the following repro code:

```
private suspend fun <T : Any> ReceiveChannel<T>.receiveBatch(size: Int): List<T> {
    return List(size) { receive() }
}

fun main() = runBlocking {
    class Thing(val counter: Long /* for debugging on the heap dump */)

    val channel1 = produce(capacity = 64) { // from the main thread
        (0 until Long.MAX_VALUE).forEach {
            send(Thing(it))
        }
    }

    launch(newSingleThreadContext("Channel2")) { // to a single thread (This thread is important! no leak if everything is single threaded)
        flow {
            while (true) {
                emit(channel1.receiveBatch(64))
            }
        }.collect { }
    }

    Unit
}
```

This sample code leaks `Thing`s very rapidly

Yourkit screenshot

I haven't dug too deep, but one interesting thing is if I inline the code for receive batch, there is no leak. Example:

```
fun main() = runBlocking {
    class Thing(val counter: Long /* for debugging on the heap dump */)

    val channel1 = produce(capacity = 64) { // from the main thread
        (0 until Long.MAX_VALUE).forEach {
            send(Thing(it))
        }
    }

    launch(newSingleThreadContext("Channel2")) { // to a single thread (This thread is important! no leak if everything is single threaded)
        flow {
            while (true) {
                emit(List(64) { channel1.receive() })
            }
        }.collect { }
    }

    Unit
}
```

Tested with kotlin 1.4.30 and coroutines-core 1.4.3

qwwdfsad:
* Update initCancellability documentation and implementation to be aligned with current invariants
    * Make parentHandle non-volatile and ensure there are no races around it
    * Establish new reusability invariants
      - Reusable continuation can be used _only_ if it's state is not REUSABLE_CLAIMED
      - If it is, spin-loop and wait for release
      - Now the parent is attached to reusable continuation only if it was suspended at least once. Otherwise, the state machine can return via fast-path and noone will be able to release intercepted continuation (-> detach from parent)
      - It implies that the parent is attached after trySuspend call and can be concurrently reused, this is where new invariant comes into play

Fixes #2564

### DIFF
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -107,7 +107,7 @@ public interface CancellableContinuation<in T> : Continuation<T> {
      * Internal function that setups cancellation behavior in [suspendCancellableCoroutine].
      * It's illegal to call this function in any non-`kotlinx.coroutines` code and
      * such calls lead to undefined behaviour.
-     * Exposes in our ABI since 1.0.0 withing `suspendCancellableCoroutine` body.
+     * Exposed in our ABI since 1.0.0 withing `suspendCancellableCoroutine` body.
      *
      * @suppress **This is unstable API and it is subject to change.**
      */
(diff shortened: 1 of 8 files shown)


elizarov (approved):
Looks great. While studying the code I've made some minor refactoring (mostly to remove duplicated code) and added some more comments.

elizarov (approved):

PR #2583, opened 2021-03-15
Mention default buffering in flattenMerge documentation

### LINKED ISSUE #2546: Should flatMapLatest be buffered by default?
For me it was really unexpected, because first it's not clear from the operator name that it's buffered, and second there's no other operators, that are buffered by default (not counting `flatMapLatest` and others, as they use `transformLatest` themselves).

```
flow {
    emit((1..9).toList())
    delay(1500)
    emit(listOf(10))
}
.flatMapLatest { it.asFlow() }
// .buffer(0) // we can remove buffer from flatMapLatest
.distinctUntilChanged()
.collectLatest { int ->
    doSomeResourceConsumingWork(int) // for simplicity can be replaced with delay(1000)
}
```
Here I was expecting it to collect 1, then 2, then cancel and receive the last 10 emit. But because `flatMapLatest` is buffered, there was 10 collects without cancellation.

So the main question is - should `transformLatest` be buffered by default, or is it better to change it to be non-buffered to avoid confusion?

qwwdfsad:
Fixes #2546


elizarov (approved):
Great.

Issue #2079, opened 2020-06-03
Calling awaitOne on nonconforming Publishers gives nondescriptive errors 


vemilyus:
Hi, I've noticed that `Publisher.awaitOne` allows further calls to `onComplete` after receiving an `onError`  call. This can potentially hide the exception that was supplied to `onError` and instead throws this exception: 

```
java.lang.IllegalStateException: Already resumed, but proposed with update null
	at kotlinx.coroutines.CancellableContinuationImpl.alreadyResumedError(CancellableContinuationImpl.kt:335)
	at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl(CancellableContinuationImpl.kt:330)
	at kotlinx.coroutines.CancellableContinuationImpl.resumeWith(CancellableContinuationImpl.kt:250)
	at kotlinx.coroutines.reactive.AwaitKt$awaitOne$$inlined$suspendCancellableCoroutine$lambda$1.onComplete(Await.kt:138)
	at org.jooq.impl.AbstractResultQuery$1.doComplete(AbstractResultQuery.java:415)
	at org.jooq.impl.AbstractResultQuery$1.request(AbstractResultQuery.java:409)
	at kotlinx.coroutines.reactive.AwaitKt$awaitOne$$inlined$suspendCancellableCoroutine$lambda$1.onSubscribe(Await.kt:105)
	at org.jooq.impl.AbstractResultQuery.subscribe(AbstractResultQuery.java:381)
	at org.jooq.impl.SelectImpl.subscribe(SelectImpl.java:2716)
	at kotlinx.coroutines.reactive.AwaitKt.awaitOne(Await.kt:97)
	at kotlinx.coroutines.reactive.AwaitKt.awaitOne$default(Await.kt:95)
	at kotlinx.coroutines.reactive.AwaitKt.awaitFirstOrNull(Await.kt:46)
	[REDACTED]
	at kotlinx.coroutines.stream.StreamFlow.collect(Stream.kt:27)
	at kotlinx.coroutines.stream.StreamFlow$collect$1.invokeSuspend(Stream.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute$$$capture(AbstractEventExecutor.java:164)
[...]

PR #2586, opened 2021-03-16
Make the subscriber in awaitOne less permissive

dkhalanskyjb:
Reactive Streams' Subscriber's implementation for `await*` operations was assuming that the publisher is correct. Now, the implementation detects some instances of problematic behavior for publishers and reports them.

Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2079

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/Await.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Await.kt
@@ -10,6 +10,7 @@ import kotlinx.coroutines.suspendCancellableCoroutine
 import org.reactivestreams.Publisher
 import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
+import java.lang.IllegalStateException
 import java.util.*
 import kotlin.coroutines.*
 
@@ -134,29 +135,47 @@ private suspend fun <T> Publisher<T>.awaitOne(
     mode: Mode,
     default: T? = null
 ): T = suspendCancellableCoroutine { cont ->
+    /* This implementation must obey
+    https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#2-subscriber-code
+    The numbers of rules are taken from there. */
     injectCoroutineContext(cont.context).subscribe(object : Subscriber<T> {
-        private lateinit var subscription: Subscription
+        // It is unclear whether 2.13 implies (T: Any), but if so, it seems that we don't break anything by not adhering
+        private var subscription: Subscription? = null
         private var value: T? = null
         private var seenValue = false
+        private var inTerminalState = false
 
         override fun onSubscribe(sub: Subscription) {
+            /** cancelling the existing subscription due to rule 2.5, though the publisher would either have to
+             * subscribe more than once, which would break 2.12, or leak this [Subscriber]. */
+            subscription?.let {
+                value = null
+                seenValue = false
+                inTerminalState = false
+                it.cancel()
+            }
             subscription = sub
             cont.invokeOnCancellation { sub.cancel() }
-            sub.request(if (mode == Mode.FIRST) 1 else Long.MAX_VALUE)
+            sub.request(if (mode == Mode.FIRST || mode == Mode.FIRST_OR_DEFAULT) 1 else Long.MAX_VALUE)
         }
 
         override fun onNext(t: T) {
+            val sub = subscription.checkInitialized("onNext")
+            subscription.checkInitialized("onNext")
+            if (inTerminalState)
+                gotSignalInTerminalStateException("onNext")
             when (mode) {
                 Mode.FIRST, Mode.FIRST_OR_DEFAULT -> {
-                    if (!seenValue) {
-                        seenValue = true
-                        subscription.cancel()
-                        cont.resume(t)
-                    }
+                    if (seenValue)
+                        // TODO: decide if we want to be lenient here: after all, nothing breaks if this isn't true
+                        moreThanOneValueProvidedException(mode)
+                    seenValue = true
+                    sub.cancel()
+                    cont.resume(t)
                 }
                 Mode.LAST, Mode.SINGLE, Mode.SINGLE_OR_DEFAULT -> {
                     if ((mode == Mode.SINGLE || mode == Mode.SINGLE_OR_DEFAULT) && seenValue) {
-                        subscription.cancel()
+                        sub.cancel()
                         if (cont.isActive)
                             cont.resumeWithException(IllegalArgumentException("More than one onNext value for $mode"))
                     } else {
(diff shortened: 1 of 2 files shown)


reactive/kotlinx-coroutines-reactive/test/IntegrationTest.kt
```diff
@@ -130,6 +132,73 @@ class IntegrationTest(
...
+                    override fun cancel() {
+                    }
+                })
+            }
+        suspend fun<T> assertDetectsBadPublisher(operation: suspend Publisher<T>.() -> T,
+                                                 block: Subscriber<in T>.(n: Long) -> Unit) =
+            assertFailsWith<IllegalStateException> { publisher(block).operation() }
+
+        // Rule 1.1 broken: the publisher produces more values than requested.
+        assertDetectsBadPublisher<Int>({ awaitFirst() }) {
```
dkhalanskyjb, line 156: Is it expected that `::awaitFirst()` doesn't work here?

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -134,29 +135,47 @@ private suspend fun <T> Publisher<T>.awaitOne(
...
+            subscription?.let {
+                value = null
+                seenValue = false
+                inTerminalState = false
+                it.cancel()
+            }
             subscription = sub
             cont.invokeOnCancellation { sub.cancel() }
-            sub.request(if (mode == Mode.FIRST) 1 else Long.MAX_VALUE)
+            sub.request(if (mode == Mode.FIRST || mode == Mode.FIRST_OR_DEFAULT) 1 else Long.MAX_VALUE)
```
qwwdfsad, line 159: Shouldn't it be anything but `Mode.LAST`?

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -134,29 +135,47 @@ private suspend fun <T> Publisher<T>.awaitOne(
...
         }
 
         override fun onNext(t: T) {
+            val sub = subscription.checkInitialized("onNext")
+            subscription.checkInitialized("onNext")
```
qwwdfsad, line 164: Double `checkInitialized` seems to be redundant

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -134,29 +135,47 @@ private suspend fun <T> Publisher<T>.awaitOne(
...
             when (mode) {
                 Mode.FIRST, Mode.FIRST_OR_DEFAULT -> {
-                    if (!seenValue) {
-                        seenValue = true
-                        subscription.cancel()
-                        cont.resume(t)
-                    }
+                    if (seenValue)
+                        // TODO: decide if we want to be lenient here: after all, nothing breaks if this isn't true
+                        moreThanOneValueProvidedException(mode)
```
qwwdfsad, line 171: When writing a comment along with a statement, please wrap it in curly braces, otherwise it's a bit hard to read

qwwdfsad: Throwing here breaks `#2.13`. This is expected, but still worth properly documenting to know why it is here even tho it violates the spec.
Also, removing this condition does not trigger any test failures, because double-resume also throws `IllegalStateException`. It's worth verifying the exception message at least once

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -169,6 +188,8 @@ private suspend fun <T> Publisher<T>.awaitOne(
 
         @Suppress("UNCHECKED_CAST")
         override fun onComplete() {
+            subscription.checkInitialized("onComplete") // TODO: maybe don't enforce?
```
qwwdfsad, line 191: It seems like other implementations from reactive libraries do not enforce such rule, maybe we could stick with that as well

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-reactive/src/Await.kt
dkhalanskyjb: I don't think so, no: correct publishers don't produce more values than requested, and if we only requested one value for `SINGLE`, there wouldn't be any difference between `FIRST` and `SINGLE`.

reactive/kotlinx-coroutines-reactive/src/Await.kt
dkhalanskyjb: I am now verifying the exception messages. I don't like what I did in order to achieve it, but I like copy-pasting exception messages into tests even less, given how brittle that approach is. Maybe you have some better suggestions?

reactive/kotlinx-coroutines-reactive/test/IntegrationTest.kt
```diff
@@ -148,35 +148,41 @@ class IntegrationTest(
...
-        suspend fun<T> assertDetectsBadPublisher(operation: suspend Publisher<T>.() -> T,
-                                                 block: Subscriber<in T>.(n: Long) -> Unit) =
-            assertFailsWith<IllegalStateException> { publisher(block).operation() }
+        suspend fun <T> assertDetectsBadPublisher(
+            operation: suspend Publisher<T>.() -> T,
+            message: String? = null,
+            block: Subscriber<in T>.(n: Long) -> Unit,
+        ) =
+            assertFailsWith<IllegalStateException> { publisher(block).operation() }.let {
+                message == null || it.message == message
```
qwwdfsad, line 157: This line doesn't actually test exception messages, the resulting boolean is ignored

reactive/kotlinx-coroutines-reactive/src/Await.kt
qwwdfsad: The approach we're consistently using is the following:

- Do not expose exception messages to test in any way (e.g. via `internal` function)
- Instead of strictly comparing the error message as a whole, typically it's enough to do something like `e.message!!.contains("some_key_word_or_meaningful_part_of_the_message")`. 
 
 It simplifies the maintenance a bit, reduces the number of functions/string consts, but it's still brittle. Yet exception messages are rarely changed and the purpose of the test is rather fix the current behaviour

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-reactive/test/IntegrationTest.kt
dkhalanskyjb: Yeah, sorry. Fixed in the commit that changes the whole error-passing mechanism here.

qwwdfsad (approved):

dkhalanskyjb:
@qwwdfsad could you please check the latest commit? I spotted several errors during the final review, fixed them, and added a test to verify a new statement in a comment.

qwwdfsad (approved):

qwwdfsad (approved):

PR #2597, opened 2021-03-22
Add timeout operator to Flow

pablobaxter:
This adds functionality similar to the [`timeout()`](http://reactivex.io/documentation/operators/timeout.html) in Rx. This invokes a timeout if the upstream flow doesn't emit after the set period of time, with the option to perform a custom action on timeout.

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -1035,6 +1035,10 @@ public final class kotlinx/coroutines/flow/FlowKt {
 	public static final fun switchMap (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/Flow;
 	public static final fun take (Lkotlinx/coroutines/flow/Flow;I)Lkotlinx/coroutines/flow/Flow;
 	public static final fun takeWhile (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/Flow;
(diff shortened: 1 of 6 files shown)


qwwdfsad:
https://github.com/Kotlin/kotlinx.coroutines/pull/2745#pullrequestreview-681969186 

Though this operator is kind of different from the other time-related operators though, so there are chances we can accept it sooner.
But first, as explained in our [contribution guide,](https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md) please start with the issue that describes the use-case pattern and the problem it solves. Please note that "similar operator exists in Rx" is not the use-case and is not how we are implementing/designing/deciding on the operators

pablobaxter:
@qwwdfsad Understood. I did create an issue (#2624) that talked about use case for this PR, but that was after-the-fact. I probably should have added the content of that issue to this PR as well. Please advise if the use case from that issue is enough for this PR. Thanks!

qwwdfsad:
Oh, thanks, I've completely missed that as it wasn't linked and was quite old (and we're not handling PRs in a timely manner 😔) so I forgot about it.

That's great (especially now when it has linked PR), I will thoughtfully read and comment it tomorrow, the use-case with GPS looks quite compelling

pablobaxter:
> 
> 
> Oh, thanks, I've completely missed that as it wasn't linked and was quite old (and we're not handling PRs in a timely manner 😔) so I forgot about it.
> 
> That's great (especially now when it has linked PR), I will thoughtfully read and comment on it tomorrow, the use case with GPS looks quite compelling, thanks for detailed write up and your patience!

No worries. Given all the great things you and team are doing with Kotlin, it's completely understandable. Thank you coming back and looking at this!

pablobaxter:
Question regarding this PR, should I re-open a new one or is the JB team going to handle this one?

Issue #2262, opened 2020-09-22
mono { } with timeout never completes


frost13it:
**Expected behaviour:**
When a coroutine inside `mono { }` fails with a `TimeoutCancellationException`, the returned `Mono` terminates with an error.

**Actual behaviour:**
The returned `Mono` never completes.

Example:
```kotlin
import kotlinx.coroutines.delay
import kotlinx.coroutines.reactor.mono
import kotlinx.coroutines.withTimeout

fun main() {
    val mono = mono {
        withTimeout(1) { delay(100) }
    }
    mono.doOnSubscribe { println("Subscribed") }
        .doOnNext { println("Emitted: $it") }
        .doOnSuccess { println("Completed") }
        .doOnError { println("Error: $it") }
        .doOnCancel { println("Cancelled") }
        .block()
    println("Done")
}
```

The code above prints `Subscribed` and hangs forever.

It seems that such behaviour is introduced in #1638.

PR #2606, opened 2021-03-24
Complete mono { } on cancellation

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2262

### DIFF
--- a/reactive/kotlinx-coroutines-reactor/src/Mono.kt
+++ b/reactive/kotlinx-coroutines-reactor/src/Mono.kt
@@ -7,6 +7,7 @@
 package kotlinx.coroutines.reactor
 
 import kotlinx.coroutines.*
+import org.reactivestreams.*
 import reactor.core.*
 import reactor.core.publisher.*
 import kotlin.coroutines.*
@@ -30,6 +31,22 @@ public fun <T> mono(
     require(context[Job] === null) { "Mono context cannot contain job in it." +
             "Its lifecycle should be managed via Disposable handle. Had $context" }
     return monoInternal(GlobalScope, context, block)
+}
+
+@Suppress("UNCHECKED_CAST")
+internal suspend fun <T> Mono<T>.await(): T? = (this as Mono<T?>).awaitOrDefault(null)
+
+internal suspend fun <T> Mono<T>.awaitOrDefault(default: T): T = suspendCancellableCoroutine { cont ->
+    subscribe(object: CoreSubscriber<T> {
+        override fun onSubscribe(s: Subscription) {
+            cont.invokeOnCancellation { s.cancel() }
+            s.request(1)
+        }
+        override fun onNext(t: T) { cont.resume(t) }
+        override fun onError(t: Throwable) { cont.resumeWithException(t) }
+        override fun onComplete() { cont.resume(default) }
+
+    })
 }
 
 @Deprecated(
@@ -68,11 +85,8 @@ private class MonoCoroutine<in T>(
 
     override fun onCancelled(cause: Throwable, handled: Boolean) {
         try {
-            if (getCancellationException() === cause) {
-                /** Cancellation exceptions are meaningless to the user, so we present them as absences of a value. If
-                 * [sink] is already in a terminal state, this call will be ignored altogether, which is good. */
-                sink.success()
-            } else {
+            /** Cancellation exceptions that were caused by [dispose], that is, came from downstream, are not errors. */
+            if (getCancellationException() !== cause || !disposed) {
                 /** If [sink] turns out to already be in a terminal state, this exception will be passed through the
                  * [Hooks.onErrorDropped] hook, which is the way to signal undeliverable exceptions in Reactor. */
                 sink.error(cause)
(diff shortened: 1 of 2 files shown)


reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -68,11 +68,13 @@ private class MonoCoroutine<in T>(
 
     override fun onCancelled(cause: Throwable, handled: Boolean) {
         try {
-            /*
-             * sink.error handles exceptions on its own and, by default, handling of undeliverable exceptions is a no-op.
-             * Guard potentially non-empty handlers against meaningless cancellation exceptions
-             */
-            if (getCancellationException() !== cause) {
+            if (getCancellationException() === cause) {
```
qwwdfsad, line 71: Both current documentation and test seem to be really suspicious.

I think I got the idea why the code was written this way (also, `blame` was kinda useful): we, in fact, have _two_ cancellation modes: 
1) The coroutines was cancelled because the subscriber cancelled the subscription.
2) The coroutine itself has been cancelled due to timeouts or any other unhandled cancellation exceptions.

I've asked @sdeleuze about this, but current behaviour seems to be error-prone.
I'd propose the following instead:

1) If the subscriber cancelled the coroutine, then such cancellation is "expected" and CE should not be reported anywhere
2) If the coroutine was cancelled on its own, then it's a programmatic error that should be delivered.

Also, that way it will be consistent with `rxSingle` (its worth checking its implementation as well)

qwwdfsad (changes requested):

qwwdfsad:
Also, we've discussed that such behaviour creates unwanted inconsistency between `rxMaybe` and `rxSingle` even if the former always completes with a value

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -32,6 +33,22 @@ public fun <T> mono(
     return monoInternal(GlobalScope, context, block)
 }
 
+@Suppress("UNCHECKED_CAST")
+internal suspend fun <T> Mono<T>.await(): T? = (this as Mono<T?>).awaitOrDefault(null)
```
qwwdfsad, line 37: These functions seem to be misplaced and should belong to test source-set

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -67,17 +69,16 @@ private class MonoCoroutine<in T>(
...
 
     override fun onCancelled(cause: Throwable, handled: Boolean) {
-        try {
-            /*
-             * sink.error handles exceptions on its own and, by default, handling of undeliverable exceptions is a no-op.
-             * Guard potentially non-empty handlers against meaningless cancellation exceptions
-             */
-            if (getCancellationException() !== cause) {
+        /** Cancellation exceptions that were caused by [dispose], that is, came from downstream, are not errors. */
+        if (getCancellationException() !== cause || !disposed) {
```
qwwdfsad, line 73: With such a check, we have a scenario when an exception is swallowed.
E.g. the following test:
```
@Test
fun testSwallowedException() = runTest {
    val started = CountDownLatch(1)
    val disposed = CountDownLatch(1)
    val m = mono {
        started.countDown()
        disposed.await()
        println("Running")
        throw ArithmeticException()
    }
    val d = m.subscribe()
    started.await()
    d.dispose()
    disposed.countDown()
}
```

will print `Running` and then will completely ignore the arithmetic exception

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-reactor/src/Mono.kt
qwwdfsad: PTAL on discussion here https://github.com/Kotlin/kotlinx.coroutines/issues/252 and the corresponding commit that fixes the problem

dkhalanskyjb: I have read the discussion and am not sure how it is applicable. Could you please clarify what behavior would you expect instead?

What happens here is that `onError` gets called after `m` is disposed of, and Reactor deals with error from such objects by passing them to the `onOperatorError` hook, which is unset in this test. Setting the hook leads to the exception being reported:
```kotlin
    @Test
    fun testSwallowedException() = runTest {
        Hooks.onOperatorError("testSwallowedException") { t, a ->
            expectUnreached()
            t
        }
        val started = CountDownLatch(1)
        val disposed = CountDownLatch(1)
        val m = mono {
            started.countDown()
            disposed.await()
            println("Running")
            throw ArithmeticException()
        }
        val d = m.subscribe()
        started.await()
        d.dispose()
        disposed.countDown()
    }
```
successfully fails.

Moreover, it looks like using `onOperatorError` is the approach we take for undeliverable exceptions in Reactor, so… this exception is already dealt with the way we usually deal with the undeliverable exceptions. What am I missing here?

qwwdfsad: Oh, sorry, my bad, I've misinterpreted the code. Thanks for the clarification!

qwwdfsad (approved):

PR #2612, opened 2021-03-26
Deprecation and migration of receiveOrNull and onReceiveOrNull.

### LINKED ISSUE #1676: onReceiveOrNul extension convenience
Message from Thibault Seisel in public Slack: https://kotlinlang.slack.com/archives/C1CFAFJSK/p1574860187023800

Since `ReceiveChannel<T>.onReceiveOrNul`l has been deprecated, it is suggested to replace it with the `onReceiveOrNull` extension function.But I find that it's surprisingly hard to use that 
 function: Since there is a member property `onReceiveOrNull` it takes precedence over the extension. Since `onReceiveOrNull` is a function, the block passed as select clause is incorrectly interpreted as an extra lambda parameter. While the following workaround code works:
```
import kotlinx.coroutines.channels.onReceiveOrNull as onReceiveOrNullExt
select {
    channel.onReceiveOrNullExt().invoke { foo ->
        [...]
    }
    anotherChannel.onReceive { bar ->
        [...]
    }
}
```
It is not as convenient as the original Select API. Wouldn't it be better to convert that extension function to an extension property with a different name ?

qwwdfsad:
* Raise deprecation level for members, introduce deprecation for extensions
    * Explain rationale behind deprecation
    * Provide default implementation for deprecated members in Channel interface
    * Get rid of the internal implementation, leverage receiveCatching
    * Introduce new extensions for ChannelResult and use them as replacement in our own operators

Fixes #1676

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
+++ b/kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
@@ -54,7 +54,8 @@ internal suspend fun <R, T> FlowCollector<R>.combineInternal(
         ++currentEpoch
         // Start batch
         // The very first receive in epoch should be suspending
-        var element = resultChannel.receiveOrNull() ?: break // Channel is closed, nothing to do here
+        // TODO closed with an exception ?
+        var element = resultChannel.receiveCatching().getOrNull() ?: break // Channel is closed, nothing to do here
         while (true) {
             val index = element.index
             // Update values
@@ -129,7 +130,9 @@ internal fun <T1, T2, R> zipImpl(flow: Flow<T1>, flow2: Flow<T2>, transform: sus
                 withContextUndispatched(coroutineContext + collectJob, Unit) {
                     flow.collect { value ->
                         withContextUndispatched(scopeContext, Unit, cnt) {
-                            val otherValue = second.receiveOrNull() ?: throw AbortFlowException(this@unsafeFlow)
+                            val otherValue = second.receiveCatching().getOrElse {
+                                throw it ?:AbortFlowException(this@unsafeFlow)
+                            }
                             emit(transform(value, NULL.unbox(otherValue)))
                         }
                     }
--- a/kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
+++ b/kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
@@ -54,7 +54,6 @@ internal suspend fun <R, T> FlowCollector<R>.combineInternal(
         ++currentEpoch
         // Start batch
         // The very first receive in epoch should be suspending
-        // TODO closed with an exception ?
         var element = resultChannel.receiveCatching().getOrNull() ?: break // Channel is closed, nothing to do here
         while (true) {
             val index = element.index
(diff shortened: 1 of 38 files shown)


kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
```diff
@@ -54,7 +54,8 @@ internal suspend fun <R, T> FlowCollector<R>.combineInternal(
         ++currentEpoch
         // Start batch
         // The very first receive in epoch should be suspending
-        var element = resultChannel.receiveOrNull() ?: break // Channel is closed, nothing to do here
+        // TODO closed with an exception ?
```
elizarov, line 57: `resultChannel` cannot be closed with an exception

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -177,16 +177,6 @@ public interface SelectInstance<in R> {
...
- * | [Job]            | [join][Job.join]                              | [onJoin][Job.onJoin]                             | [isCompleted][Job.isCompleted]
- * | [Deferred]       | [await][Deferred.await]                       | [onAwait][Deferred.onAwait]                      | [isCompleted][Job.isCompleted]
- * | [SendChannel]    | [send][SendChannel.send]                      | [onSend][SendChannel.onSend]                     | [offer][SendChannel.offer]
- * | [ReceiveChannel] | [receive][ReceiveChannel.receive]             | [onReceive][ReceiveChannel.onReceive]            | [poll][ReceiveChannel.poll]
- * | [ReceiveChannel] | [receiveOrNull][ReceiveChannel.receiveOrNull] | [onReceiveOrNull][ReceiveChannel.onReceiveOrNull]| [poll][ReceiveChannel.poll]
- * | [Mutex]          | [lock][Mutex.lock]                            | [onLock][Mutex.onLock]                           | [tryLock][Mutex.tryLock]
- * | none             | [delay]                                       | [onTimeout][SelectBuilder.onTimeout]             | none
- *
  * This suspending function is cancellable. If the [Job] of the current coroutine is cancelled or completed while this
```
elizarov, line 180: Without a list of select clauses in this doc, a causal documentation reader who stumbles up the the `select` function would have zero clues on what are those "select clauses" that this doc talks about and how `select` exception can be actually used. It is non-discoverable without a list of various supported select clauses.

kotlinx-coroutines-core/common/src/flow/internal/Combine.kt
qwwdfsad: Thanks, checked it, but forgot to remove my internal TODO

kotlinx-coroutines-core/common/src/selects/Select.kt
qwwdfsad: I've returned the list back but removed "non-suspending functions" column. It's not really informative, but with the current names (`[onReceiveCatching][ReceiveChannel.onReceiveCatching]`) this table became as wide as 200 symbols, rendering it unreadable from IDE.

This documentation still needs some love, e.g. the table in its current state is not properly rendered in IDEA quick-KDoc, but it seems to be out of the scope

elizarov (approved):

PR #2613, opened 2021-03-26
Remove deprecated SendChannel.isFull 

qwwdfsad:
Also, give it default stub implementation and remove it from all the subclasses. This implementation previously was partially correct, was used as a performance optimization and can be safely introduced for a marginally popular method

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -31,14 +31,12 @@ public interface SendChannel<in E> {
     public val isClosedForSend: Boolean
 
     /**
-     * Returns `true` if the channel is full (out of capacity), which means that an attempt to [send] will suspend.
-     * This function returns `false`  if the channel [is closed for `send`][isClosedForSend].
+     * Warning in 1.2, error in 1.3, hidden in 1.5
      *
      * @suppress **Will be removed in next releases, no replacement.**
      */
-    @ExperimentalCoroutinesApi
-    @Deprecated(level = DeprecationLevel.ERROR, message = "Will be removed in next releases without replacement")
-    public val isFull: Boolean
+    @Deprecated(level = DeprecationLevel.HIDDEN, message = "Will be removed in next releases without replacement")
+    public val isFull: Boolean get() = true
 
     /**
      * Sends the specified [element] to this channel, suspending the caller while the buffer of this channel is full
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -31,14 +31,6 @@ public interface SendChannel<in E> {
     public val isClosedForSend: Boolean
 
     /**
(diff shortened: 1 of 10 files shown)


kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -31,14 +31,12 @@ public interface SendChannel<in E> {
...
-     * This function returns `false`  if the channel [is closed for `send`][isClosedForSend].
+     * Warning in 1.2, error in 1.3, hidden in 1.5
      *
      * @suppress **Will be removed in next releases, no replacement.**
      */
-    @ExperimentalCoroutinesApi
-    @Deprecated(level = DeprecationLevel.ERROR, message = "Will be removed in next releases without replacement")
-    public val isFull: Boolean
+    @Deprecated(level = DeprecationLevel.HIDDEN, message = "Will be removed in next releases without replacement")
+    public val isFull: Boolean get() = true
```
elizarov, line 39: Doesn't it defeat the purpose of making it hidden? What kind of previously compiled code might continue to work normally after such a change? Maybe it is better for such a code to simply break?

qwwdfsad: From our previous study of `isFull`, I remember that it was primarily used as a performance optimization to avoid slow callbacks and/or blocking sends/coroutines launches. Such existing code will continue to work.

Taking this into account, WDYT, is it worth to keep this implementation?

qwwdfsad: For the reference, I'm all in for removing it, rather was trying to minimize the potential damage

elizarov: I'm failing to find any example of a code with `.isFull`, but they way I see `isFull` could have been used is this:
```
while(true) {
    sleep(100) // or somehow wait
    if (channel.isFull) continue
    // not full -- send
    channel.send(expensiveObject())
}
```

This could would be totally broken with `isFull: Boolean get() = true`, but using `isFull: Boolean get() = false` would somewhat preserved. 

That was I was curious about. What kind of code might continue to be working correctly with  `isFull: Boolean get() = true`?

qwwdfsad: >What kind of code might continue to be working correctly with isFull: Boolean get() = true?

Ktor in its Netty integration had a heuristic check to flush all outbound content right from the [Netty] channel handler if the [coroutines] channel with incoming messages was full. Kind of "check how much in-progress message we have and process them in place", but as a performance optimization.

And back then, from our study of other usages, they all could've been boiled down to this

elizarov: Let's either drop `isFull` completely or keep its correct implementation for the hidden method. Since I cannot find any code using `isFull` in the wild (and it's been an ERROR for a while), dropping it completely seems Ok for me, too.

qwwdfsad: Removed it completely then

elizarov (approved):
Very good.

PR #2616, opened 2021-03-30
Fix wrong docs on Job.join

### LINKED ISSUE #2615: Job join method documentation
I have found very strange documentation for join method:

`In particular, it means that a parent coroutine invoking join on a child coroutine that was started using launch(coroutineContext) { ... } builder throws CancellationException if the child had crashed, unless a non-standard CoroutineExceptionHandler is installed in the context.`

I'm not sure that CoroutineExceptionHandler will have effect for CancellationException.
Example:

```
fun main() = runBlocking {
    val handler = CoroutineExceptionHandler { _, exception ->
        println("CoroutineExceptionHandler got $exception")
    }
    val job = GlobalScope.launch(handler) {
        val inner = launch { // all this stack of coroutines will get cancelled
            throw IOException() // the original exception
        }
        try {
            inner.join()
            println("handle join")
        }
    }
    job.join()
}
```

Output:
`handle join
CoroutineExceptionHandler got java.io.IOException`

So basically CancellationException will still be thrown regardless any installed handlers.

elizarov:
Fixes #2615

### DIFF
--- a/kotlinx-coroutines-core/common/src/Job.kt
+++ b/kotlinx-coroutines-core/common/src/Job.kt
@@ -252,9 +252,9 @@ public interface Job : CoroutineContext.Element {
      * suspending function is invoked or while it is suspended, this function
      * throws [CancellationException].
      *
-     * In particular, it means that a parent coroutine invoking `join` on a child coroutine that was started using
-     * `launch(coroutineContext) { ... }` builder throws [CancellationException] if the child
-     * had crashed, unless a non-standard [CoroutineExceptionHandler] is installed in the context.
+     * In particular, it means that a parent coroutine invoking `join` on a child coroutine throws
+     * [CancellationException] if the child had failed, since a failure of a child coroutine cancels parent by default,
+     * unless the child was launched from within [supervisorScope].
      *
      * This function can be used in [select] invocation with [onJoin] clause.
      * Use [isCompleted] to check for a completion of this job without waiting.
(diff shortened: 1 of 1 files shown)


serddmit:
@elizarov, Don't forget to change documentation for cancelAndJoin, it has the same statement.
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/cancel-and-join.html

Thanks

serddmit:
Don't forget to change documentation for cancelAndJoin, it has the same statement.
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/cancel-and-join.html

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -252,9 +252,9 @@ public interface Job : CoroutineContext.Element {
      * suspending function is invoked or while it is suspended, this function
      * throws [CancellationException].
      *
-     * In particular, it means that a parent coroutine invoking `join` on a child coroutine that was started using
-     * `launch(coroutineContext) { ... }` builder throws [CancellationException] if the child
-     * had crashed, unless a non-standard [CoroutineExceptionHandler] is installed in the context.
+     * In particular, it means that a parent coroutine invoking `join` on a child coroutine throws
```
qwwdfsad, line 255: I'd suggest removing the whole paragraph, now it seems to be redundant. 
Also, please fix the same phrasing for `cancelAndJoin`

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/Job.kt
elizarov: I think this explanation still makes sense as it is quite counter-intuitive for people that you can crash an attempt to `join` another coroutine.

qwwdfsad (approved):

PR #2617, opened 2021-03-30
Fixes for the reactive integrations

### LINKED ISSUE #2591: awaitSingleOrNull returns not nullable value
https://github.com/Kotlin/kotlinx.coroutines/blob/edd584acb852d4b403173d168ea036510d1e1109/reactive/kotlinx-coroutines-reactive/src/Await.kt#L107

awaitSingleOrNull returns not nullable value and
>  @throws NoSuchElementException if publisher does not emit any value

Also, behaviour differs from the std function:
>  @throws IllegalArgumentException if publisher emits more than one value

```
/**
 * Returns single element, or `null` if the list is empty or has more than one element.
 */
public fun <T> List<T>.singleOrNull()
```

dkhalanskyjb:
Improves documentation, adds a couple of new tests.


qwwdfsad (approved):
Great work!

PR #2627, opened 2021-04-05
Fix version shield. Bintray -> Maven Central

turansky:


qwwdfsad:
Thanks, already fixed in #2562

Issue #1587, opened 2019-10-01
Add await() extension on Rector Mono type


elizarov:
It should simply be an inline alias to `Publisher.awaitSingle()`. But special care shall be taken about `Mono<Void>` types that never emit any value. They need `awaitCompletion()` which is missing altogether now.

jorge-recio:
I would like to work on this issue.
If `suspend fun Mono<T>.await(): T` is implemented as an inline alias of `suspend fun Publisher.awaitSingle(): T` and Mono<Void>.await() is called the result will be a **TypeCastException: null cannot be cast to non-null type**. What should be the expected behavior?

About the implementation of suspend fun Mono<T>.awaitCompletion(): Unit can be implemented using a new function Publisher<T>.awaitCompletion(single: Boolean) with single = true. But can this feature be implemented as part of this issue or needs to be done in a different one?

elizarov:
They both need to be implemented together, but they need careful design on their behaviour in various edge-cases. It is not clear at this moment what should happen when `Mono` completes without emitting a value and how common this behavior is in the real code.

wickedev:
Since this issue hasn't progressed for a year, I've written PoC-level code. Need more edge-cases then this?

```kt
suspend inline fun <reified T> Mono<T>.await(): T {
    if (null is T) {
        // for Mono<T?>
        return awaitFirstOrNull() as T
    }

    return when (T::class) {
        // for Mono<Unit> or Mono<Void>
        Unit::class, Void::class -> {
            @Suppress("UNCHECKED_CAST", "ReactiveStreamsSubscriberImplementation")
            suspendCancellableCoroutine { cont ->
                cont as CancellableContinuation<T?>
                subscribe(object : Subscriber<T?> {
                    override fun onSubscribe(s: Subscription) {
                        cont.invokeOnCancellation {
                            s.cancel()
                        }
                        s.request(1)
                    }

                    override fun onNext(t: T?) {
                        error("Mono<Void> or Mono<Unit> cannot emit value")
                    }

                    override fun onError(t: Throwable) {
                        cont.resumeWithException(t)
                    }

                    override fun onComplete() {
                        cont.resume(null)
                    }
                })
            }
        }
        // for Mono<T>
        else -> awaitSingle() as T
    }
}

object MonoAwaitTest : Spek({
    test("mono await") {
        runBlocking {
            val m1: Mono<Int> = Mono.just(100)
            val r1: Int = m1.await()
[...]

Issue #2607, opened 2021-03-24
Provide an API to recover stacktrace of the exception when it does not cross coroutine boundary


qwwdfsad:
`kotlinx.coroutines` provides a [stacktrace recovery](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/topics/debugging.md#stacktrace-recovery) machinery in order to provide complete stacktraces that preserve information about all suspended functions in the stracktrace.

But the exception recovery only works through coroutine boundaries.

For example, the following code:
```
suspend fun foo() {
    yield() // Suspend
    bar()
    yield()
}

suspend fun bar() {
    yield() // Suspend
    throw IllegalStateException()
}

runBlocking { // Could've been 'launch' or any other coroutine builder
    foo()
}
```

will be missing `foo` function from the stacktrace because `IllegalStateException` does not cross any suspendable coroutine boundaries and stacktrace recovery machinery don't have a chance to kick in.

Instead, we could've provided a new top-level suspend function that recovers the stacktrace. Library authors and application developers could've been using that along with stacktrace-sensitive exceptions and/or "expected" failures.

There is a lot of shape for such API, e.g. `throw IllegalStateException().recoverStacktrace()` or 
```
recovering {
    require(myArg >= 1, "Expected myArg to be >= 1")
}
```

but the biggest question of this issue is whether it is useful for other developers.
Opinions regarding usefulness of such API are welcomed

dovchinnikov:
In IJ we often log an error trace and proceed without throwing, this would be a huge help to have the ability to recover full trace.

We'd like to obtain “trace when caught” at the point of “trace when thrown” in the following example:
```
package com.foo.bar

import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.delay
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.withContext
import kotlin.time.Duration.Companion.seconds

fun main() {
  runBlocking {
    try {
      withContext(Dispatchers.Default) {
        delay(1.seconds)
        doWork()
      }
    }
    catch (t: Throwable) {
      println("trace when caught:")
      t.printStackTrace(System.out)
    }
  }
}

private fun doWork(): Int {
  val ise = IllegalStateException("Internal invariant failed")
  println("trace when thrown:")
  ise.printStackTrace(System.out)
  throw ise
}
```

```
trace when thrown:
java.lang.IllegalStateException: Internal invariant failed
[...]

PR #2628, opened 2021-04-05
Deprecate awaitSingleOr*, specialize some await* functions for Mono and Maybe

dkhalanskyjb:
This PR
* Deprecates `awaitSingleOr*` on arbitrary Publishers
* Adds specialized `awaitSingle` and `awaitSingleOrNull` methods on `Maybe<T>` and `Mono<T>`
* Deprecates `Maybe<T>.await()` in favor of `Maybe<T>.awaitSingleOrNull()`
* Adds specializations of most of the `await*` methods for `Mono<T>` and deprecates them, as the only useful methods on `Mono<T>` are `awaitSingle` and `awaitSingleOrNull`
* Rewords some documentation for `await*` methods

Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2591
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/1587

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/Await.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Await.kt
@@ -9,7 +9,6 @@ import org.reactivestreams.Publisher
 import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
 import java.lang.IllegalStateException
-import kotlin.NoSuchElementException
 import kotlin.coroutines.*
 
 /**
@@ -90,6 +89,15 @@ public suspend fun <T> Publisher<T>.awaitSingle(): T = awaitOne(Mode.SINGLE)
  * If the [Job] of the current coroutine is cancelled or completed while the suspending function is waiting, this
  * function immediately cancels its [Subscription] and resumes with [CancellationException].
  *
+ * ### Deprecation
+ *
+ * This method is deprecated because the conventions established in Kotlin mandate that an operation with the name
+ * `awaitSingleOrDefault` returns the default value instead of throwing in case there is an error; however, this would
+ * also mean that this method would return the default value if there are *too many* values. This could be confusing to
+ * those who expect this function to validate that there is a single element or none at all emitted, and cases where
+ * there are no elements are indistinguishable from those where there are too many, though these cases have different
+ * meaning.
+ *
  * @throws NoSuchElementException if the publisher does not emit any value
  * @throws IllegalArgumentException if the publisher emits more than one value
  */
@@ -97,7 +105,7 @@ public suspend fun <T> Publisher<T>.awaitSingle(): T = awaitOne(Mode.SINGLE)
     message = "Deprecated without a replacement due to its name incorrectly conveying the behavior. " +
         "Please consider using awaitFirstOrDefault().",
     level = DeprecationLevel.WARNING
-)
+) // Warning since 1.5, error in 1.6, hidden in 1.7
 public suspend fun <T> Publisher<T>.awaitSingleOrDefault(default: T): T = awaitOne(Mode.SINGLE_OR_DEFAULT, default)
 
 /**
@@ -109,6 +117,15 @@ public suspend fun <T> Publisher<T>.awaitSingleOrDefault(default: T): T = awaitO
  * If the [Job] of the current coroutine is cancelled or completed while the suspending function is waiting, this
  * function immediately cancels its [Subscription] and resumes with [CancellationException].
  *
+ * ### Deprecation
+ *
+ * This method is deprecated because the conventions established in Kotlin mandate that an operation with the name
+ * `awaitSingleOrNull` returns `null` instead of throwing in case there is an error; however, this would
+ * also mean that this method would return `null` if there are *too many* values. This could be confusing to
+ * those who expect this function to validate that there is a single element or none at all emitted, and cases where
+ * there are no elements are indistinguishable from those where there are too many, though these cases have different
+ * meaning.
+ *
  * @throws IllegalArgumentException if the publisher emits more than one value
  */
 @Deprecated(
@@ -117,7 +134,7 @@ public suspend fun <T> Publisher<T>.awaitSingleOrDefault(default: T): T = awaitO
         "Alternatively, please consider using awaitFirstOrNull().",
     level = DeprecationLevel.WARNING,
     replaceWith = ReplaceWith("this.awaitSingleOrNull()", "kotlinx.coroutines.reactor")
-)
+) // Warning since 1.5, error in 1.6, hidden in 1.7
 public suspend fun <T> Publisher<T>.awaitSingleOrNull(): T? = awaitOne(Mode.SINGLE_OR_DEFAULT)
 
 /**
@@ -129,13 +146,22 @@ public suspend fun <T> Publisher<T>.awaitSingleOrNull(): T? = awaitOne(Mode.SING
  * If the [Job] of the current coroutine is cancelled or completed while the suspending function is waiting, this
  * function immediately cancels its [Subscription] and resumes with [CancellationException].
  *
+ * ### Deprecation
+ *
+ * This method is deprecated because the conventions established in Kotlin mandate that an operation with the name
+ * `awaitSingleOrElse` returns the calculated value instead of throwing in case there is an error; however, this would
+ * also mean that this method would return the calculated value if there are *too many* values. This could be confusing
+ * to those who expect this function to validate that there is a single element or none at all emitted, and cases where
+ * there are no elements are indistinguishable from those where there are too many, though these cases have different
+ * meaning.
+ *
  * @throws IllegalArgumentException if the publisher emits more than one value
  */
 @Deprecated(
     message = "Deprecated without a replacement due to its name incorrectly conveying the behavior. " +
         "Please consider using awaitFirstOrElse().",
     level = DeprecationLevel.WARNING
-)
+) // Warning since 1.5, error in 1.6, hidden in 1.7
 public suspend fun <T> Publisher<T>.awaitSingleOrElse(defaultValue: () -> T): T =
     awaitOne(Mode.SINGLE_OR_DEFAULT) ?: defaultValue()
 
--- a/reactive/kotlinx-coroutines-reactive/src/Await.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/Await.kt
@@ -93,17 +93,19 @@ public suspend fun <T> Publisher<T>.awaitSingle(): T = awaitOne(Mode.SINGLE)
 public suspend fun <T> Publisher<T>.awaitSingleOrDefault(default: T): T = awaitOne(Mode.SINGLE_OR_DEFAULT, default)
 
 /**
- * Awaits for the single value from the given publisher or `null` value if none is emitted without blocking a thread and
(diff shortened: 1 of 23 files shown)


reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -94,17 +94,19 @@ public suspend fun <T> Publisher<T>.awaitSingle(): T = awaitOne(Mode.SINGLE)
...
+ * produced by the publisher, `null` is returned.
  *
  * This suspending function is cancellable.
- * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
- * immediately resumes with [CancellationException].
- *
- * @throws NoSuchElementException if publisher does not emit any value
- * @throws IllegalArgumentException if publisher emits more than one value
+ * If the [Job] of the current coroutine is cancelled or completed while the suspending function is waiting, this
+ * function immediately cancels its [Subscription] and resumes with [CancellationException].
```
qwwdfsad, line 103: Documentation and tests of `awaitSingleOrElse`and `awaitSingleOrDefault` also should be adjusted

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -158,7 +162,7 @@ private suspend fun <T> Publisher<T>.awaitOne(
                     if ((mode == Mode.SINGLE || mode == Mode.SINGLE_OR_DEFAULT) && seenValue) {
                         subscription.cancel()
                         if (cont.isActive)
```
qwwdfsad, line 164: While we are here, please wrap this statement in curly braces

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-reactive/src/Await.kt
dkhalanskyjb: There's a huge commit pending with just the documentation adjustments: https://github.com/Kotlin/kotlinx.coroutines/commit/c22351c5819de286e77493f7fb5eb9cd9928f836 Would you prefer to instead deal with the documentation for all of `await*` here?

qwwdfsad: It's much easier when the relevant changes eventually (via squash) appear as a single commit, especially when reading history/blame and investigating previous decisions.

dkhalanskyjb: Sure, I just thought that it is the sweeping changes to the documentation that would be better as a single commit. Ok though, I don't mind. A question then: should the `AwaitTest` from https://github.com/Kotlin/kotlinx.coroutines/pull/2617/commits/87b6ab20171651280d09bd458ad9f2a3bd38e3a6 also be included in this PR, given that it checks the validity of the statement that is being made here?

qwwdfsad: Yes, tests are pretty much necessary for such changes

reactive/kotlinx-coroutines-reactive/src/Await.kt
dkhalanskyjb: I think adding https://github.com/Kotlin/kotlinx.coroutines/pull/2586/commits/4baba2bc2aa5fd21ca2ef287cb9bd642debeff13 to https://github.com/Kotlin/kotlinx.coroutines/pull/2586 would make for a better place for such a change.

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -9,114 +9,135 @@ import org.reactivestreams.Publisher
 import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
 import java.lang.IllegalStateException
-import java.util.*
+import kotlin.NoSuchElementException
```
qwwdfsad, line 12: We use star imports everywhere in our code style (and we're aware it is not aligned with an official recommended code style)

reactive/kotlinx-coroutines-reactive/src/Await.kt
```diff
@@ -9,114 +9,135 @@ import org.reactivestreams.Publisher
...
- * @throws IllegalArgumentException if publisher emits more than one value
+ * @throws NoSuchElementException if the publisher does not emit any value
+ * @throws IllegalArgumentException if the publisher emits more than one value
  */
+@Deprecated(
+    message = "Deprecated without a replacement due to its name incorrectly conveying the behavior. " +
+        "Please consider using awaitFirstOrDefault().",
+    level = DeprecationLevel.WARNING
+)
 public suspend fun <T> Publisher<T>.awaitSingleOrDefault(default: T): T = awaitOne(Mode.SINGLE_OR_DEFAULT, default)
```
qwwdfsad, line 101: Two things regarding deprecation:

1) It's worth adding `### Deprecation` section to KDoc with an explanation of why this function is deprecated and, if necessary, why there is no built-in replacement. The rest of KDoc can be removed, but not hard opinion here.
2) To simplify maintenance and user's view of deprecation timeline, we add a short note about it:
```
@Deprecated(
    message = "Deprecated without a replacement due to its name incorrectly conveying the behavior. " +
        "Please consider using awaitFirstOrDefault().",
    level = DeprecationLevel.WARNING
) // Warning since 1.5, error in 1.6, hidden in 1.7
```

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -34,6 +33,50 @@ public fun <T> mono(
...
+        private var seenValue = false
+
+        override fun onSubscribe(s: Subscription) {
+            cont.invokeOnCancellation { s.cancel() }
+            s.request(Long.MAX_VALUE)
+        }
+
+        override fun onComplete() {
+            if (!seenValue)
+                cont.resume(null)
```
qwwdfsad, line 56: [nit] I'd suggest sticking to either one-liners or using explicit scoping with `{}`

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -34,6 +33,50 @@ public fun <T> mono(
...
+ * Awaits the single value from the given [Mono] without blocking the thread and returns the resulting value, or,
+ * if this Mono has produced an error, throws the corresponding exception.
+ *
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is cancelled or completed while the suspending function is waiting, this
+ * function immediately cancels its [Subscription] and resumes with [CancellationException].
+ *
+ * @throws NoSuchElementException if the Mono does not emit any value
+ */
+public suspend fun <T> Mono<T>.awaitSingle(): T = awaitSingleOrNull() ?: throw NoSuchElementException()
```
qwwdfsad, line 78: Seems to be the candidate for https://github.com/Kotlin/kotlinx.coroutines/issues/2607?

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -89,3 +132,88 @@ public fun <T> CoroutineScope.mono(
...
+ *
+ * @throws NoSuchElementException if the publisher does not emit any value
+ */
+@Deprecated(
+    message = "Mono produces at most one value, so the semantics of dropping the remaining elements are not useful. " +
+        "Please use awaitSingle() instead.",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("this.awaitSingle()")
+)
+public suspend fun <T> Mono<T>.awaitFirst(): T = awaitSingle()
```
qwwdfsad, line 151: It's a "lint" signature, so we typically expect them to be similar to ones from `kotlinx.coroutines.flow.Migration` -- `ERROR` level, no body, `KDoc` with an explanation of our reasoning.

But this one is slightly different: it potentially shadows existing extensions from `Publisher`. So I suggest leaving it as `WARNING` with a body for now, but get rid of the current `KDoc` and only explain reasoning of deprecation.

Also, worth dropping `// For migration, added with warning in 1.5, should be error in 1.6 and never hidden` after the annotation

reactive/kotlinx-coroutines-reactive/src/Await.kt
dkhalanskyjb: Yep, I have that enabled in the config. Seems like Idea is being smart here and detecting that another code style is used in the file already.

reactive/kotlinx-coroutines-reactor/src/Mono.kt
dkhalanskyjb: Looks like it. Should I do something about it in this PR?

qwwdfsad: Dropping a comment is usually enough

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -133,87 +132,66 @@ public fun <T> CoroutineScope.mono(
...
+ * Both functions await the first value, or throw [NoSuchElementException] if there is none, but the name
+ * [Mono.awaitSingle] better reflects the semantics of [Mono].
  */
 @Deprecated(
     message = "Mono produces at most one value, so the semantics of dropping the remaining elements are not useful. " +
         "Please use awaitSingle() instead.",
     level = DeprecationLevel.WARNING,
     replaceWith = ReplaceWith("this.awaitSingle()")
-)
+) // Warning since 1.5, error in 1.6, hidden in 1.7
```
qwwdfsad, line 145: It won't be hidden because this is added as a lint hint, otherwise, it wouldn't  be added in the first place

reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -133,87 +132,66 @@ public fun <T> CoroutineScope.mono(
...
 /**
- * Awaits the first value from the given publisher without blocking the thread and returns the resulting value, or, if
- * the publisher has produced an error, throws the corresponding exception.
- *
- * This suspending function is cancellable.
- * If the [Job] of the current coroutine is cancelled or completed while the suspending function is waiting, this
- * function immediately cancels its [Subscription] and resumes with [CancellationException].
- *
- * @throws NoSuchElementException if the publisher does not emit any value
+ * This function is deprecated in favor of [Mono.awaitSingle].
```
qwwdfsad, line 136: This phrasing implicitly implies that this function was non-deprecated prior to this version.
I suggest rephrasing it more explicitly, e.g.: "This is a deprecated lint function to improve user experience and guard against ambiguous Mono usages." and adding example to the end of KDoc:

```
Example of ambiguous usage:
\```
// The signature of the call suggests that one or zero values can be returned from 'findById', but 'awaitFirstOrNull' instead of 'awaitSingleOrNull' may trick the reader into believing that more than one value can be returned
myDbClient.findById(uniqueId).awaitFirstOrNull()
\```
```

qwwdfsad (approved):
Great work!

Please address minor doc issues and it's good to go

reactive/kotlinx-coroutines-reactor/src/Mono.kt
dkhalanskyjb: Changed. What do you think?

qwwdfsad: 👍

PR #2630, opened 2021-04-06
Add the missing type boundaries to the reactive integrations

dkhalanskyjb:


qwwdfsad (commented):
Sorry, we've discussed it before and decided not to do that (the end of DM by 2020-09-23).

It's a source-incompatible change to stable API that will require all authors of generic code to introduce `Any` constraint. 
And any code that actually tries to emit `null` will fail at the first run.

PR #2632, opened 2021-04-06
Add debugger-related manifest to root metadata artifact of kotlinx-co…

### LINKED ISSUE #2619: Premain-Class missing in 1.4.3 causing debug failures in IntelliJ
Related to [KTIJ-5659](https://youtrack.jetbrains.com/issue/KTIJ-5659), the 1.4.3 build of `kotlinx-coroutines-core` does not include `Premain-Class` in the `MANIFEST.MF`, nor the `AgentPremain.class` file. Those do exist in `kotlinx-coroutines-core-jvm-1.4.3.jar`.

This is causing failures when debugging unit tests in IntelliJ, which attempts to reference `-javaagent:.../kotlinx-coroutines-core-1.4.3.jar` if you don't include a dependency on `kotlinx-coroutines-core-jvm-1.4.3.jar` (or if `kotlinx-coroutines-core-1.4.3.jar` is before `kotlinx-coroutines-core-jvm` in your effective `pom.xml` for a maven project).

qwwdfsad:
…routines-core

    * It fixes IDEA issue that was introduced in coroutines 1.4.3 where we switched to Kotlin 1.4.30 and stopped producing root artifact with JVM classes
    * It simplifies manual -javaagent specification for end-users who shouldn't even be aware of kotlinx-coroutines-core-jvm.jar

Fixes #2619

### DIFF
--- a/kotlinx-coroutines-core/build.gradle
+++ b/kotlinx-coroutines-core/build.gradle
@@ -208,8 +208,20 @@ jvmTest {
     }
 }
 
-jvmJar {
-    manifest {
+// Setup manifest for kotlinx-coroutines-core-jvm.jar
+jvmJar { setupManifest(it) }
+
+/*
+ * Setup manifest for kotlinx-coroutines-core.jar
+ * This is convenient for users that pass -javaagent arg manually and also is a workaround #2619 and KTIJ-5659.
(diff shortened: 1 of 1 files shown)


elizarov (approved):
Nice trick

PR #2638, opened 2021-04-08
Use cancellation exception from standard library in Kotlin/JS and K/N

qwwdfsad:
* On JVM it uses java.util.concurrent.CancellationException, same as stdlib declaration
    * Fully backward compatible for JVM, Native-friendly for Obj-C interop

### DIFF
--- a/kotlinx-coroutines-core/common/test/flow/internal/FlowScopeTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/internal/FlowScopeTest.kt
@@ -68,7 +68,7 @@ class FlowScopeTest : TestBase() {
             flowScope {
                 flowScope {
                     launch {
-                       throw CancellationException(null)
+                       throw CancellationException("")
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/common/test/flow/internal/FlowScopeTest.kt
```diff
@@ -68,10 +68,10 @@ class FlowScopeTest : TestBase() {
             flowScope {
                 flowScope {
                     launch {
-                       throw CancellationException(null)
+                       throw CancellationException("")
```
qwwdfsad, line 71: This is an intentional change, now common CE has both constructors from `Throwable?` cause and from `String?`, so `null` is ambiguous now

kotlinx-coroutines-core/native/src/Exceptions.kt
```diff
@@ -10,13 +10,7 @@ package kotlinx.coroutines
...
  * (see [CoroutineExceptionHandler]).
  */
-public actual open class CancellationException(
-    message: String?,
-    cause: Throwable?
-) : IllegalStateException(message, cause) {
-    public actual constructor(message: String?) : this(message, null)
-}
-
+public actual typealias CancellationException = kotlin.coroutines.cancellation.CancellationException
```
elizarov, line 13: Style nit: empty line after

elizarov (approved):
See comments

PR #2640, opened 2021-04-09
Add Mono<T>.await

### LINKED ISSUE #1587: Add await() extension on Rector Mono type

dkhalanskyjb:
Fixes #1587

### DIFF
--- a/reactive/kotlinx-coroutines-reactor/src/Mono.kt
+++ b/reactive/kotlinx-coroutines-reactor/src/Mono.kt
@@ -8,11 +8,9 @@ package kotlinx.coroutines.reactor
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.reactive.*
-import org.reactivestreams.*
 import reactor.core.*
 import reactor.core.publisher.*
 import kotlin.coroutines.*
-import kotlin.internal.*
 
 /**
  * Creates a cold [mono][Mono] that runs a given [block] in a coroutine and emits its result.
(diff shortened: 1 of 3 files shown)


reactive/kotlinx-coroutines-reactor/src/Mono.kt
```diff
@@ -34,6 +32,17 @@ public fun <T> mono(
...
+/**
+ * Awaits the completion of the [Mono] without blocking the thread.
+ * Returns the resulting value, or `null` if no value is produced, or throws the corresponding exception if this
+ * [Mono] has produced an error.
+ *
+ * This suspending function is cancellable.
+ * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this
+ * function immediately resumes with [CancellationException] and cancels its subscription.
+ */
+public suspend fun <T> Mono<T>.await(): T? = awaitFirstOrNull()
```
qwwdfsad, line 44: It seems like the `awaitOrNull` can be a more appropriate name.
We can discuss it along with `singleOr*` deprecation if you don't mind

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-reactor/src/Mono.kt
qwwdfsad: Or, alternatively, `awaitCompletion`?

dkhalanskyjb: Wouldn't we have to rename the existing `MaybeSource<T>.await()` then?

qwwdfsad: Good point, I'll revisit them

dkhalanskyjb:
Closed in favor of https://github.com/Kotlin/kotlinx.coroutines/pull/2628

PR #2643, opened 2021-04-12
Embrace new channel API

qwwdfsad:
* Introduce trySendBlocking and deprecate sendBlocking
    * Use it in callbackFlow example
    * Stabilize callbackFlow and channelFlow as they finally have error-safe API
    * Irrelevant: migrate from deprecated stdlib API to be able to build with Kotlin 1.5

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/channels/Channels.kt
+++ b/kotlinx-coroutines-core/jvm/src/channels/Channels.kt
@@ -11,28 +11,28 @@ import kotlinx.coroutines.*
 
 /**
  * **Deprecated** blocking variant of send.
- *
  * This method is deprecated in the favour of [trySendBlocking].
  *
- * `sendBlocking` is considered to be dangerous primitive -- it throws
+ * `sendBlocking` is a dangerous primitive &mdash; it throws an exception
  * if the channel was closed or, more commonly, cancelled.
- * Cancellation exceptions are not ignored by non-blocking code and frequently
+ * Cancellation exceptions in non-blocking code are unexpected and frequently
  * trigger internal failures.
  *
- * These bugs were hard-to-spot during code review and also forced users to write
- * their own wrappers around `sendBlocking`, so it was decided to deprecate
- * this function and provide a more explicit primitive instead.
+ * These bugs are hard-to-spot during code review and they forced users to write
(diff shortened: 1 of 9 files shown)


qwwdfsad:
Investigating stress test failure

kotlinx-coroutines-core/jvm/src/channels/Channels.kt
```diff
@@ -25,3 +38,38 @@ public fun <E> SendChannel<E>.sendBlocking(element: E) {
...
+ * channel.trySendBlocking(element)
+ *     .onSuccess { /* request next element or debug log */ }
+ *     .onFailure { t: Throwable? -> /* throw or log */ }
+ *
+ * ```
+ * For this operation it is guaranteed that [failure][ChannelResult.failed] always contains an exception in it.
+ *
+ * @throws [InterruptedException] if the current thread is interrupted during the blocking send operation.
+ */
+public fun <E> SendChannel<E>.trySendBlocking(element: E): ChannelResult<Unit> {
```
elizarov, line 62: Also, should add `@Throws(InterruptedException::class)` to trigger "inappropriate blocking call" inspection

elizarov (changes requested):

kotlinx-coroutines-core/jvm/src/channels/Channels.kt
```diff
@@ -10,12 +10,25 @@ package kotlinx.coroutines.channels
...
+ * This method is deprecated in the favour of [trySendBlocking].
+ *
+ * `sendBlocking` is considered to be dangerous primitive -- it throws
+ * if the channel was closed or, more commonly, cancelled.
+ * Cancellation exceptions are not ignored by non-blocking code and frequently
+ * trigger internal failures.
+ *
+ * These bugs were hard-to-spot during code review and also forced users to write
+ * their own wrappers around `sendBlocking`, so it was decided to deprecate
+ * this function and provide a more explicit primitive instead.
```
elizarov, line 24: It would be great to add an example of broken code here, too.

qwwdfsad: Good suggested, added one from the recently reported bug

elizarov (approved):
Good riddance!

PR #2645, opened 2021-04-12
Introduce runningFold operator, make scan alias to runningFold

### LINKED ISSUE #2641: Add runningFold function to Flow
Kotlin supports the following functions on Collections
- `reduce` and `runningReduce`
- `fold` and `runningFold`

whereas Flows only support
- `reduce` and `runningReduce`
- `fold` and **???**

Turns out the function is called `scan`, which is not consistent with the known naming scheme and hard to find. The `scan` function also exists for Collections but only acts as an alias for `runningFold` there.

It would be helpful to also add the `runningFold` function to Flows.

qwwdfsad:
* Otherwise Kotlin users with non-reactive background get confused by flow/stdlib inconsistency
    * Make it experimental to delay the final decision about the name

Fixes #2641


elizarov (approved):
🚀 Let's go

Issue #2551, opened 2021-02-24
Provide an API to check whether debug mode is enabled


qwwdfsad:
We already have a notion of debug mode in our documentation, but we do not have an API to check whether the mode is enabled.

It can be useful in multiple scenarios:

* To write unit tests that work both with and without debug mode
    * We already do that in our MPP tests
* To verify and debug program behaviour (see #2550)
* To write your own extensions to coroutines that may exhibit an additional behaviour when debug mode is enabled (stacktrace for cancellation exception subclasses, context markers and so on)

qwwdfsad:
Also, it's worth adding `unwrap` function and separate stacktrace recovery check.

We already bump into this in our own modules: https://github.com/Kotlin/kotlinx.coroutines/pull/2646/commits/89fd2c179efcd5a80d5dd2f12ada5a75734e9326

PR #2646, opened 2021-04-13
Consistently handle exceptions in reactive streams

dkhalanskyjb:
* Fixed `PublisherCoroutine`, `rxObservable`, and
  `Flow.toPublisher` ignoring cancellations.
* Fatal exceptions are not treated in a special manner by us
  anymore. Instead, we follow the requirement in the reactive
  streams specification that, in case some method of `Subscriber`
  throws, that subscriber MUST be considered canceled, and the
  exception MUST be reported in someplace other than `onError`.
* Fixed `trySend` sometimes throwing in `PublisherCoroutine` and
  `rxObservable`.
* When an exception happens inside a cancellation handler, we now
  consistently throw the original exception passed to the handler,
  with the new exception added as suppressed.
* Fixed `PublisherCoroutine` and `rxObservable` claiming that the
  channel is not closed for send for some time after `close()` has
  finished.
* Fixed publishers sometimes signalling `onComplete()` after
  cancellation even though their streams are not finite.

Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2173

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt
@@ -208,11 +208,9 @@ public class FlowSubscription<T>(
             consumeFlow()
             true
         } catch (cause: Throwable) {
-            if (cancellationRequested && cause === getCancellationException()) {
-                return
+            if (cancellationRequested && !isActive && (cause === getCancellationException() || cause.cause === getCancellationException() && cause.message == getCancellationException().message)) {
+                subscriber.onComplete()
             } else {
-                // TODO: this branch gets entered even when `cause` looks identical to `getCancellationException()`.
-                // Is stack sanitization to blame?
                 try {
                     subscriber.onError(cause)
                 } catch (e: Throwable) {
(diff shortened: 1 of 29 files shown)


dkhalanskyjb:
I will be adding some tests to this PR but would appreciate some feedback on the implementation in the meantime.

qwwdfsad (commented):
That's amazing and thoughtful work from you, thanks!

reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt
```diff
@@ -204,17 +205,12 @@ public class FlowSubscription<T>(
...
         } catch (cause: Throwable) {
-            /* TODO: The part after "||" is a hack needed due to [cause] having travelled over a coroutine boundary and
-                being changed from the result of [getCancellationException()]. */
-            if (cancellationRequested && !isActive && (cause === getCancellationException() || cause.cause === getCancellationException() && cause.message == getCancellationException().message)) {
-                /* TODO: This is incorrect, as [Subscriber.onComplete] denotes the end of the stream and not just any
-                    non-erroneous terminal state. */
-                subscriber.onComplete()
-            } else {
+            @Suppress("INVISIBLE_MEMBER")
+            val unwrappedCause = unwrap(cause)
```
qwwdfsad, line 212: https://github.com/Kotlin/kotlinx.coroutines/issues/2551 filed a use-case

qwwdfsad (approved):
Good to go

Issue #2558, opened 2021-03-01
Provide an API to close Dispatchers.Default and Dispatchers.IO


qwwdfsad:
In containerized environments (Gradle build scripts, application containers, IDEA plugins), it is important to be able to unload the bundle of classes when the job is done.

Unfortunately, any code that touches our dispatchers (e.g. transitively, or via `GlobalScope`) cannot be unloaded: few indefinitely parked scheduler threads retain a strong reference to its class loader and cannot be terminated in any public way. For that, we should provide a public API to shutdown our dispatchers, destroy all their threads and make the outer classloader free.

The open question is how to establish a method contract in a way that nicely interacts with the fact that `IO` and `Default` are closely tightened and also take into account that depending on the system property, `Dispatchers.Default` can be backed by `ForkJoinPool.commonPool` that cannot be shut down (it doesn't suffer from this problem tho, because it's a JDK class that is always loaded by a bootstrap classloader).

PR #2652, opened 2021-04-15
Public experimental shutdown method on DefaultDispatcher

### LINKED ISSUE #2491: DefaultDispatcher needs public shutdown API for native-mt
[`DefaultDispatcher` global object](https://github.com/Kotlin/kotlinx.coroutines/blob/native-mt/kotlinx-coroutines-core/native/src/Dispatchers.native.kt#L12) creates [a `Worker`](https://github.com/Kotlin/kotlinx.coroutines/blob/native-mt/kotlinx-coroutines-core/native/src/Workers.kt#L40) in its `SingleThreadDispatcher` delegate and never shuts it down. This means that there's an active `Worker` left at the program shutdown, which is catched by Kotlin/Native memory leak checker.

To make memory leak checker do its job, `DefaultDispatcher` would need some public API to shut itself down, which can then be called in user code before program (or test) termination.

mvicsokolova:
Fixes #2491


kotlinx-coroutines-core/native/src/Dispatchers.native.kt
```diff
@@ -12,6 +12,11 @@ public actual object Dispatchers {
     public actual val Default: CoroutineDispatcher get() = DefaultDispatcher
     public actual val Main: MainCoroutineDispatcher = createMainDispatcher(Default)
     public actual val Unconfined: CoroutineDispatcher get() = kotlinx.coroutines.Unconfined // Avoid freezing
+
+    @ExperimentalCoroutinesApi
+    public fun CoroutineDispatcher.shutdown() {
```
qwwdfsad, line 17: Please add documentation to this method.

Also, it's probably worth adding it right to common `Dispatchers` (but with effect only on Native problem) to simplify adoption for our MPP users

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -78,4 +78,14 @@ public expect object Dispatchers {
      * the value of [CoroutineStart.UNDISPATCHED].
      */
     public val Unconfined: CoroutineDispatcher
+
+    /**
+     * Shuts down the dispatcher.
```
qwwdfsad, line 83: Please explain the reasoning of adding the dispatcher (K/N memory management) here and also mention that this signature is present only in `native-mt` versions and not in regular coroutines to simplify KMPP users adoption

qwwdfsad: Also, you can mention that it will be available in regular versions when #2558 is implemented

kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -78,4 +78,14 @@ public expect object Dispatchers {
...
+     *
+     * Its implementation depends on the platform:
+     * - On Kotlin/JVM and Kotlin/JS takes no effect.
+     * - On Kotlin/Native requests termination of the worker created by the [Default] dispatcher.
+     */
+    @ExperimentalCoroutinesApi
+    public fun CoroutineDispatcher.shutdown()
```
qwwdfsad, line 90: The signature seems to be not really usable because it has two receivers and also has effect _only_ on the default dispatcher in K/N which is a bit weird.

I'd suggest sticking with `Dispatchers.shutdownDefaultDispatcher` while we don't have a separate type for `Dispatchers.Default` and cannot do `Dispatchers.Default.shutdown()`.

qwwdfsad (changes requested):

Sifat1998 (approved):

Sifat1998 (approved):

kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -78,4 +78,22 @@ public expect object Dispatchers {
      * the value of [CoroutineStart.UNDISPATCHED].
      */
     public val Unconfined: CoroutineDispatcher
+
+    /**
+     * Shuts down the [Default] dispatcher.
+     *
+     * Its implementation depends on the platform:
+     * - On Kotlin/JVM and Kotlin/JS takes no effect.
```
qwwdfsad, line 86:
```suggestion
     * - On Kotlin/JVM and Kotlin/JS has no effect.
```

kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -78,4 +78,22 @@ public expect object Dispatchers {
...
+     * - On Kotlin/Native requests termination of the Worker created by the [Default] dispatcher.
+     *   [Default] dispatcher is backed by a [SingleThreadDispatcher] that creates a Worker and never shuts it down.
+     *   According to the current implementation, Kotlin/Native memory leak checker can not do it’s job while
+     *   there are active workers left at the program shutdown. Thus, if the memory leak checker is activated
+     *   (`Platform.isMemoryLeakCheckerActive = true`) this method should be called before program termination
+     *   to avoid the leaking worker.
+     *
+     * **Note**: For now this declaration is only present in `native-mt` versions of `kotlinx.coroutines`.
+     *           It will be available in regular versions when #2558 is implemeted.
```
qwwdfsad, line 95:
```suggestion
     *           It will be available in regular versions when #2558 is implemented, potentially under different name
```

kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -78,4 +78,22 @@ public expect object Dispatchers {
...
+     */
+    @ExperimentalCoroutinesApi
+    public fun Dispatchers.shutdownDefaultDispatcher()
```
qwwdfsad, line 98: Why is it both extension on `Dispatchers` and the member of it? It should be only one of those

qwwdfsad (changes requested):

qwwdfsad:
Closing as won't fix

PR #2653, opened 2021-04-15
Kotlin version update

qwwdfsad:


elizarov (approved):
Ok. Need to forget about Android tests. Also need to add verification that we don't accidentally use Android-unsupported APis.

PR #2659, opened 2021-04-19
Properly implement replacement for Flow<T>.concatWith(flow) to avoid …

qwwdfsad:
…deadlocks when the downstream throws, but other flow has no elements in it

Fixes KT-46013


elizarov (approved):
Great

PR #2662, opened 2021-04-20
Introduce Flow.last and Flow.lastOrNull operators

### LINKED ISSUE #2246: Flow.last() operator
I found myself often doing this `flow.toList().last()` usually for code that works with Flow which represents progress, when I want to receive the final item, it would be complimentary to first()/single() item

It also can be replaced with something like 
```
flow
   .filterInInstance<SomeTerminalItemType>()
   .single()
```

but it's not always so simple, for example, a sealed hierarchy may have a few terminal types for operation

Reactive libraries have such operator http://reactivex.io/documentation/operators/last.html

qwwdfsad:
Fixes #2246


qwwdfsad:
Overloads with predicate are not added as we do not know compelling use-cases for them

elizarov (approved):
👍 Now, it is complete!

PR #2665, opened 2021-04-21
Add ChannelResult.onClosed

qwwdfsad:
* Establish clear contract on ChannelResult.isClosed
    * This method provides a **clear** migration from correct 'offer' usages to 'trySend'


elizarov (approved):
Nice

PR #2676, opened 2021-04-27
fix: remove coroutines-basic-jvm.md from TOC

p7nov:


elizarov (approved):
Too bad IDEA does not support those `.tree` files and does not understand the references when refactoring code.

koshachy:
> Too bad IDEA does not support those `.tree` files and does not understand the references when refactoring code.

Yeah, we know that.
We are working on a separate doc builder, and after that, we can integrate it into the coroutines build, like `knit`. 
Sorry for the inconvenience, I hope we will provide that builder till July.

Issue #2690, opened 2021-05-09
Coroutine debugger does not work with 1.5.0-RC


elizarov:
See https://youtrack.jetbrains.com/issue/KTIJ-11620
The error is: 

```
Exception is thrown by calling dumpCoroutines.

java.lang.IllegalStateException: Unable to find loaded class _COROUTINE._CREATION
	at org.jetbrains.kotlin.idea.debugger.coroutine.proxy.LocationCache.createLocation(LocationCache.kt:17)
...
```

dkhalanskyjb:
`_COROUTINE._CREATION` was introduced in 1.5.0-RC: https://github.com/Kotlin/kotlinx.coroutines/pull/2461, and it is now transitively used during the initialization of `DebugProbesImpl`. This is probably the root cause for this issue, as, looking at https://github.com/JetBrains/kotlin/blob/master/idea/jvm-debugger/jvm-debugger-coroutine/src/org/jetbrains/kotlin/idea/debugger/coroutine/proxy/mirror/coroutinesDebugMirror.kt, my surface understanding is that the plugin proxies every class that could be needed for `DebugProbesImpl` to function.

Looking at the jar file for 1.5.0-RC, it does contain the `_COROUTINE._CREATION` class, so there should be no problems with loading it; it's unclear though why `ArtificialStackFrames`, a class that is also needed now for `DebugProbesImpl` to function, does not provoke the same reaction.

Also, while I am somewhat able to reproduce this, I get a different stack trace:
```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'refType' of com/intellij/debugger/engine/DebuggerUtils.findMethod must not be null
	at com.intellij.debugger.engine.DebuggerUtils.$$$reportNull$$$0(DebuggerUtils.java)
	at com.intellij.debugger.engine.DebuggerUtils.findMethod(DebuggerUtils.java)
	at com.intellij.debugger.jdi.GeneratedLocation.<init>(GeneratedLocation.java:21)
	at org.jetbrains.kotlin.idea.debugger.coroutine.proxy.LocationCache.createLocation(LocationCache.kt:36)
	at org.jetbrains.kotlin.idea.debugger.coroutine.proxy.LocationCache.createLocation(LocationCache.kt:17)
	at org.jetbrains.kotlin.idea.debugger.coroutine.proxy.CoroutineLibraryAgent2Proxy.findStackFrames(CoroutineLibraryAgent2Proxy.kt:86)
[...]

PR #2695, opened 2021-05-11
Revert displaying fake stack frames as method calls

dkhalanskyjb:
A simple commit that mechanically reverts https://github.com/Kotlin/kotlinx.coroutines/pull/2461. Conflicts with https://github.com/Kotlin/kotlinx.coroutines/pull/2694.

Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2690


qwwdfsad (approved):
Please reference the original commit in the squash commit message

Issue #2669, opened 2021-04-22
Deadlock within Flow.onCompletion? 


peterfigure:
```
Coroutine "coroutine#30":StandaloneCoroutine{Active}@11501093, state: SUSPENDED
        at kotlinx.coroutines.channels.AbstractChannel.receiveOrClosed-WVj179g(AbstractChannel.kt:631)
        at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:50)
        at kotlinx.coroutines.flow.internal.ChannelFlow$collect$2.invokeSuspend(ChannelFlow.kt:135)
        at kotlinx.coroutines.flow.FlowKt__ErrorsKt.catchImpl(Errors.kt:230)
        at kotlinx.coroutines.flow.FlowKt__ErrorsKt$catch$$inlined$unsafeFlow$1.collect(Errors.kt:113)
        at kotlinx.coroutines.flow.FlowKt__EmittersKt$onCompletion$$inlined$unsafeFlow$1.collect(Emitters.kt:114)
```

The coroutine is suspended there and never makes any progress. This happens in a backend application after a period of sustained load. I'm trying to distill a simplified example to illustrate the situation but no success yet.

peterfigure:
Just to add more colour, the 8 threads in the scope in which this flow was launch are in waiting:

```
stackTrace:
java.lang.Thread.State: WAITING (parking)
at jdk.internal.misc.Unsafe.park(java.base@11.0.11/Native Method)
- parking to wait for <0x0000000749396fa8> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
at java.util.concurrent.locks.LockSupport.park(java.base@11.0.11/LockSupport.java:194)
at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(java.base@11.0.11/AbstractQueuedSynchronizer.java:2081)
at java.util.concurrent.LinkedBlockingQueue.take(java.base@11.0.11/LinkedBlockingQueue.java:433)
at java.util.concurrent.ThreadPoolExecutor.getTask(java.base@11.0.11/ThreadPoolExecutor.java:1054)
at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@11.0.11/ThreadPoolExecutor.java:1114)
[...]

PR #2700, opened 2021-05-13
Fail-fast in emitAll implementation from onCompletion

qwwdfsad:
It is helpful to prevent bugs like KT-46013 and potential deadlocks or delayed cancellations


qwwdfsad:
Also likely fixes #2669

elizarov (approved):
🚀 Great solution!

PR #2701, opened 2021-05-13
Flows returned by stateIn/shareIn keep strong reference to sharing job

### LINKED ISSUE #2557: Return value of sharIn&stateIn don't hold a strong reference to the sharing coroutine, letting them be garbage collected
Hello,

The `shareIn` and `stateIn` operators for `Flow` are unfortunately victims of that problem: https://youtrack.jetbrains.com/issue/KT-30588

Here's a self-contained commented snippet that you can run to see the issue:
```kotlin
@file:OptIn(ExperimentalTime::class, ExperimentalCoroutinesApi::class)

import kotlinx.coroutines.*
import java.lang.ref.WeakReference
import kotlin.coroutines.EmptyCoroutineContext
import kotlin.time.Duration
import kotlin.time.ExperimentalTime
import kotlin.time.milliseconds
import kotlin.time.seconds

fun main(): Unit = runBlocking {
    println("Hello World!")
    val jobToCancel = launch {
        someFlow.collect()
    }
    delay(100.milliseconds) // Ensure the flow has a chance to be activated before GC happens.
    repeat(5) {
        System.gc() // Keep only strong references
        System.gc() // Really GC!
        delay(1.seconds)
    }
    jobToCancel.cancel()
}

// Strongly references a Job holding scope that will keep child coroutines strongly referenced as well.
val theScope = CoroutineScope(EmptyCoroutineContext)

val someFlow = flow<Int> {
    iterationsFlow(1.seconds).collect {
        println("Got $it")
        currentCoroutineContext().ensureActive()
    }
}.stateIn(
    scope = CoroutineScope(EmptyCoroutineContext), // Replace it with `theScope` and the issue goes away
    started = SharingStarted.Eagerly, // Or use `WhileSubscribed()` and the issue somehow goes away too.
    initialValue = 0
)

fun iterationsFlow(interval: Duration): Flow<Int> = callbackFlow {
    val listener = Listener {
        runCatching { offer(it) } // runCatching because of https://github.com/Kotlin/kotlinx.coroutines/issues/974
    }
    val cb = CallbackRepeatCaller(interval, listener)
    awaitClose {
        cb.forget()
        listener.call(-1)
    }
}


fun interface Listener {
    fun call(iterations: Int)
}

class CallbackRepeatCaller(interval: Duration, listener: Listener) {
    private val ref = WeakReference(listener) // Mimics actual APIs like Android's MediaPlayer or SharedPreferences.

    //private val l = listener // Uncommenting this code would "fix" the issue,
    // but it only works if you control the code.

    private val caller = CoroutineScope(EmptyCoroutineContext).launch {
        var iterations = 0
        while (true) {
            ensureActive()
            ref.get()?.call(++iterations)
            delay(interval)
        }
    }

    fun forget() {
        caller.cancel()
        ref.clear()
    }
}

```

### Actual output:
```
Hello World!
Got 1
```

### Expected output:
```
Hello World!
Got 1
Got 2
Got 3
Got 4
Got 5
Got 6
```

You can see this is what we get when we use a scope holding a `Job` and keep a top level strong reference to it.

## What happens

The coroutine is suspended and nothing holds a strong reference to it because the returned `StateFlow`/`SharedFlow` doesn't have any internal reference to the parent `Job` of the coroutine collecting the flow to share, and the suspended call uses a `WeakReference` to the listener, which is the last holder of a strong reference to the continuation.

## Why it's a problem

It affected customers of my company as something was stuck, but didn't reproduce if GC hadn't triggered.
It was quite hard to find the root cause, and the solution needed some research.

A bunch of JVM APIs use `WeakReference` to help Java callers avoid memory leaks, so kotlinx.coroutines relying on every suspend function to keep a strong reference to the continuation while suspended is problematic and will lead to further occurrences until it's changed.

## Possible solutions I know

There's two problems:

### Problem 1

It's possible and very easy to get or instantiate a `CoroutineScope` that has no `Job()`.
These scopes (like `GlobalScope`) will not keep a strong reference to running coroutines and allow this issue to happen.

To prevent this problem from happening, the `coroutineContext` property of `CoroutineScope` could start requiring opt-in with `@InternalCoroutinesApi`, so users must either use implementations provided by kotlinx.coroutines, or acknowledge the risks and opt-in.
Then, the `GlobalScope` object implementation could be changed so its `coroutineContext` has a `SupervisorJob` that doesn't respond to `cancel()`, so it keeps references to the child coroutines.
The same could be done for `NonCancellable` so it doesn't leave the continuation unreferenced.

Introducing these changes would also prevent future recurrence of issues like the one that led to this same report: https://youtrack.jetbrains.com/issue/KT-30588

### Problem 2

The returned `SharedFlow`/`StateFlow` doesn't hold any reference to the coroutine if you're using `Eagerly` or `Lazily` for `started`, allowing this issue to happen as well.

To prevent this issue, the instance of `SharedFlow`/`StateFlow` returned by `shareIn` and `stateIn` should have a private field that keeps a reference to the result of the `launch` call done in the `launchSharing` internal function.
That way, regardless of the `SharingStarted` "strategy", this bug can't happen anymore.

elizarov:
Fixes #2557

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
@@ -145,7 +145,7 @@ public fun <T> Flow<T>.shareIn(
     )
     @Suppress("UNCHECKED_CAST")
     val job = scope.launchSharing(config.context, config.upstream, shared, started, NO_VALUE as T)
-    return ReadonlySharedFlowWithJob(shared, job)
+    return ReadonlySharedFlow(shared, job)
 }
 
 private class SharingConfig<T>(
@@ -303,7 +303,7 @@ public fun <T> Flow<T>.stateIn(
     val config = configureSharing(1)
     val state = MutableStateFlow(initialValue)
     val job = scope.launchSharing(config.context, config.upstream, state, started, initialValue)
-    return ReadonlyStateFlowWithJob(state, job)
+    return ReadonlyStateFlow(state, job)
 }
 
 /**
@@ -331,7 +331,7 @@ private fun <T> CoroutineScope.launchSharingDeferred(
             upstream.collect { value ->
                 state?.let { it.value = value } ?: run {
                     state = MutableStateFlow(value).also {
-                        result.complete(ReadonlyStateFlowWithJob(it, coroutineContext.job))
+                        result.complete(ReadonlyStateFlow(it, coroutineContext.job))
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Share.kt
```diff
@@ -359,20 +358,30 @@ public fun <T> MutableSharedFlow<T>.asSharedFlow(): SharedFlow<T> =
...
+private open class ReadonlyStateFlow<T>(
     flow: StateFlow<T>
 ) : StateFlow<T> by flow, CancellableFlow<T>, FusibleFlow<T> {
     override fun fuse(context: CoroutineContext, capacity: Int, onBufferOverflow: BufferOverflow) =
         fuseStateFlow(context, capacity, onBufferOverflow)
 }
 
+private class ReadonlyStateFlowWithJob<T>(
+    flow: StateFlow<T>,
+    private val job: Job // keeps a strong reference to the job
```
qwwdfsad, line 382: Please add suppress in order to avoid warnings and elaborate a bit on the purpose of this in the comment for us from the future :)

kotlinx-coroutines-core/common/src/flow/operators/Share.kt
```diff
@@ -303,8 +302,8 @@ public fun <T> Flow<T>.stateIn(
 ): StateFlow<T> {
     val config = configureSharing(1)
     val state = MutableStateFlow(initialValue)
-    scope.launchSharing(config.context, config.upstream, state, started, initialValue)
-    return state.asStateFlow()
+    val job = scope.launchSharing(config.context, config.upstream, state, started, initialValue)
+    return ReadonlyStateFlowWithJob(state, job)
```
qwwdfsad, line 306: I would prefer having a single class with a nullable job in it just to keep things simple and concise (also, dex size) by the cost of 4 more bytes in a long-living object.

No strong opinion here though

qwwdfsad (approved):

kotlinx-coroutines-core/common/src/flow/operators/Share.kt
elizarov: Done

kotlinx-coroutines-core/common/src/flow/operators/Share.kt
elizarov: Done

qwwdfsad (approved):

LouisCAD:
Vsevolod raised the exacts concerns I had in the original form of this PR. 👍
The `FieldWalker.assertReachableCount` trick is nice, I didn't know about it.

Would there be any value in a stress test that works in common code, not just JVM, but takes some time to execute (a few seconds at most)?
I had been working on it so I could demonstrate the issue of #2557 along with the fix, and was middle-refining when I learned about this PR.

elizarov:
We use stress tests as a last-resort measure when we cannot do it in another way. The test with `FieldWalker` is much better since it is fast and reliable. We've created this infrastructure exactly for the purpose of writing tests like that so that we don't have to resort to stress-testing.

PR #2702, opened 2021-05-13
Add repeat and repeatUntil as flow operators

jeziellago:
This pull request adds an implementation of repeat and RepeatUntil as Flow operators.
### repeat
```kotlin
flow { emit(1) }
	.repeat(3) 
```
Expected behavior:
```text
1, 1, 1
```

### repeatUntil
```kotlin
flow { emit(1) }
	.repeatUntil { myCondition() } 
```
In this case, the flow will `emit` until the `myCondition()` returns `true`.


qwwdfsad:
Please consider getting familiar with our [contribution guide].(https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs)

>PRs with new API, but without a corresponding issue with a positive feedback about the proposed implementation are unlikely to be approved or reviewed.


Describing a use-case with the problem this API is trying to solve is a good first step. Please note that "the similar operator exists in RxJava" **is not** a compelling argument for adding new API

PR #2707, opened 2021-05-14
Move DebugProbesKt from debug to core module

PierfrancescoSoffritti:
This change will allow Android Studio to instrument DebugProbesKt from kotlin stdlib, without relying on DebugProbesKt.bin.


qwwdfsad:
Thanks! Will merge it as #2716

PR #2709, opened 2021-05-14
Fix flaky SharingReferenceTest

qwwdfsad:
Previously assert has been racing with the coroutine state spilling, triggering spurious wakeups


elizarov (approved):
👍 Thanks! That's a nice catch.

PR #2710, opened 2021-05-14
User identityHashCode instead of deprecated `Kotlin_Any_hashCode` for…

qwwdfsad:
… Native debugging capabilities

### DIFF
--- a/kotlinx-coroutines-core/native/src/Debug.kt
+++ b/kotlinx-coroutines-core/native/src/Debug.kt
@@ -5,14 +5,12 @@
 package kotlinx.coroutines
 
 import kotlin.math.*
+import kotlin.native.*
 
 internal actual val DEBUG: Boolean = false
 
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/native/src/Debug.kt
```diff
@@ -5,14 +5,12 @@
 package kotlinx.coroutines
 
 import kotlin.math.*
+import kotlin.native.*
 
 internal actual val DEBUG: Boolean = false
 
-internal actual val Any.hexAddress: String get() = abs(id().let { if (it == Int.MIN_VALUE) 0 else it }).toString(16)
+internal actual val Any.hexAddress: String get() = abs(identityHashCode().let { if (it == Int.MIN_VALUE) 0 else it }).toString(16)
```
elizarov, line 12: We have UInt now. We don't need `abs` anymore. Can do `identityHashCode().toUInt().toString(16)`

elizarov (approved):
🚀 Good to go

PR #2727, opened 2021-05-24
Document and tweak the contract of Executor.asCoroutineDispatcher and ExecutorService.asCoroutineDispatcher

qwwdfsad:
* Document it properly
    * Make it more robust to signature changes and/or delegation (e.g. see the implementation of java.util.concurrent.Executors.newScheduledThreadPool)
    * Give a public way to reduce the memory pressure via ScheduledFuture.cancel


elizarov (approved):
👍

PR #2731, opened 2021-05-26
Prevent StackOverflowError in CompletableFuture.asDeferred and proper…

### LINKED ISSUE #2730: Silent StackOverflowError in CompleteableFuture.asDeferred
```
@Test
fun testStackOverflow() = runTest {
    val future = CompletableFuture<Int>()
    val completed = AtomicLong()
    val count = 10000L
    for (i in 0 until count) {
        launch(Dispatchers.Default) {
            future.asDeferred().await()
            completed.incrementAndGet()
        }
    }
    future.complete(1)
    withTimeout(60_000) {
        coroutineContext.job.children.toList().joinAll()
        assertEquals(count, completed.get())
    }
}
```

qwwdfsad:
…ly report exceptions from completion handlers

    * It turned out that 'cancel' on completed future tries to help and invoke 'whenComplete' handlers that also invoke 'cancel' on the very same future
    * Use top-level exception handler as a last resort to deliver an exception

Fixes #2730

### DIFF
--- a/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
+++ b/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
@@ -136,11 +136,13 @@ public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
         override fun onSuccess(result: T?) {
             // Here we work with flexible types, so we unchecked cast to trick the type system
             @Suppress("UNCHECKED_CAST")
-            deferred.complete(result as T)
+            runCatching { deferred.complete(result as T) }
+                .onFailure { handleCoroutineException(EmptyCoroutineContext, it) }
         }
 
         override fun onFailure(t: Throwable) {
-            deferred.completeExceptionally(t)
+            runCatching { deferred.completeExceptionally(t) }
+                .onFailure { handleCoroutineException(EmptyCoroutineContext, it) }
         }
     }, MoreExecutors.directExecutor())
 
--- a/integration/kotlinx-coroutines-jdk8/test/future/FutureAsDeferredUnhandledCompletionExceptionTest.kt
+++ b/integration/kotlinx-coroutines-jdk8/test/future/FutureAsDeferredUnhandledCompletionExceptionTest.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package future
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.future.*
+import org.junit.*
+import org.junit.Test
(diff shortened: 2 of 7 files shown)


integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -136,11 +136,13 @@ public fun <T> ListenableFuture<T>.asDeferred(): Deferred<T> {
...
             @Suppress("UNCHECKED_CAST")
-            deferred.complete(result as T)
+            runCatching { deferred.complete(result as T) }
+                .onFailure { handleCoroutineException(EmptyCoroutineContext, it) }
         }
 
         override fun onFailure(t: Throwable) {
-            deferred.completeExceptionally(t)
+            runCatching { deferred.completeExceptionally(t) }
+                .onFailure { handleCoroutineException(EmptyCoroutineContext, it) }
```
dkhalanskyjb, line 145: Not sure if it's intentional, but this is inconsistent with the way we decided to handle undeliverable exceptions in the reactive integrations: https://github.com/Kotlin/kotlinx.coroutines/blob/master/reactive/kotlinx-coroutines-reactor/src/Flux.kt#L58-L67 The discrepancy here is that the suppressed exception, which could potentially be useful, is not being set.

The same is true about the other implementations.

qwwdfsad: I believe that this a different use case.
Ones in RX is `use`-like, where we fail during (potentially user-supplied) exception handling on cancellation.

This is a completely different story, we're delivering the exception (not handling it, just doing the regular processing) and something unexpected and/or off-the-track happened: internal errors which we have no other means to deliver or exception from `invokeOnCancellation`, which is basically the same. Here this exception is the dominant one, not the one that was going to be deferred result

dkhalanskyjb (approved):

integration/kotlinx-coroutines-jdk8/test/future/FutureAsDeferredUnhandledCompletionExceptionTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import kotlinx.coroutines.future.*
+import org.junit.*
+import org.junit.Test
+import java.util.concurrent.*
+import kotlin.test.*
+
+class FutureAsDeferredUnhandledCompletionExceptionTest : TestBase() {
+
+    // This is a separate test in order to avoid interference with uncaught exception handlers in other tests
+    private val exceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
```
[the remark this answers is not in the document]
qwwdfsad: I find it more concise to initialize in-place everything that could be initialized in-place, [opinionated] it reads better and is easier to reason about

PR #2734, opened 2021-05-28
Use correct wording

shubham08gupta:
Using Dispatchers.Default we can make the coroutines run *parallely*.

### DIFF
--- a/docs/topics/shared-mutable-state-and-concurrency.md
+++ b/docs/topics/shared-mutable-state-and-concurrency.md
@@ -2,7 +2,7 @@
 
 [//]: # (title: Shared mutable state and concurrency)
 
-Coroutines can be executed concurrently using a multi-threaded dispatcher like the [Dispatchers.Default]. It presents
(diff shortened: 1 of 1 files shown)


docs/topics/shared-mutable-state-and-concurrency.md
```diff
@@ -2,7 +2,7 @@
 
 [//]: # (title: Shared mutable state and concurrency)
 
-Coroutines can be executed concurrently using a multi-threaded dispatcher like the [Dispatchers.Default]. It presents
+Coroutines can be executed parallely using a multi-threaded dispatcher like the [Dispatchers.Default]. It presents
 all the usual concurrency problems. The main problem being synchronization of access to **shared mutable state**.
```
dkhalanskyjb, line 6:
```suggestion
Coroutines can be executed parallelly using a multi-threaded dispatcher like the [Dispatchers.Default]. It presents
all the usual parallelism problems. The main problem being synchronization of access to **shared mutable state**. 
```

PR #2738, opened 2021-06-02
Introduce the throttle operator for rate-limiting a flow without losing trailing items

ryanpbrewster:
The `.sample()` operator for flows is useful for rate-limiting a flow, but there are use-cases where it is undesireable to potentially lose trailing items.

The motivating use-case here is something like a async progress bar. It is acceptable to drop intermediate progress, but it is very important that the last event does not get dropped.

I approached this somewhat differently than the existing implementation for `.sample()`. In particular, the "throttling" window is only started when an element is actually emitted downstream. I saw a `todo` in the existing implementation

>                 // todo: shall be start sampling only when an element arrives or sample aways as here?

so it seems that there was some consideration about whether to have this functionality for `sample()` as well.


qwwdfsad:
See the reasoning here: https://github.com/Kotlin/kotlinx.coroutines/pull/2745

It would be nice to add signatures that we are sure will stay and that are quite thought out and consistent with each other

PR #2739, opened 2021-06-02
Update R8 for Kotlin 1.5.0, enable DEX size metrics back

qwwdfsad:


dkhalanskyjb:
While you're here, maybe you could also do https://github.com/Kotlin/kotlinx.coroutines/blob/30f90aefe59373dcfdcc333434cdb87c8f51a055/buildSrc/src/main/kotlin/RunR8.kt#L17

qwwdfsad:
@dkhalanskyjb done

dkhalanskyjb (approved):

PR #2745, opened 2021-06-04
Add a preserveLatest option to Flow::sample()

ryanpbrewster:
This is a somewhat lighter-weight approach to https://github.com/Kotlin/kotlinx.coroutines/pull/2738

The .sample() operator for flows is useful for rate-limiting a flow, but there are use-cases where it is undesirable to potentially lose trailing items.

The motivating use-case here is something like a async progress bar. It is acceptable to drop intermediate progress, but it is very important that the last event does not get dropped.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -275,7 +275,7 @@ private fun <T> Flow<T>.debounceInternal(timeoutMillisSelector: (T) -> Long) : F
  * Note that the latest element is not emitted if it does not fit into the sampling window.
  */
 @FlowPreview
-public fun <T> Flow<T>.sample(periodMillis: Long): Flow<T> {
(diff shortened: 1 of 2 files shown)


qwwdfsad (commented):
Unfortunately, we cannot accept this API.

The whole idea behind the Flow API is that it's being composable and orthogonal:

E.g. instead of Rx-style `bufferSize` and `Scheduler` parameters for each possible method that may benefit from it, we have orthogonal and composable `.buffer()` and `.flowOn()` -- they are easy to learn, documented once, can be applied anywhere and fused with operators that can leverage it. The mentioned operators in the meantime become more simpler and less overloaded with irrelevant context/buffer details and are easier to learn as well.

The whole story with time-related operators is that a lot of them differs in whether the last element is emitted or not. We neither want to add a boolean flag for that, nor introduce a separate name for "operator with almost the same semantics". But alternatives are not yet quite thought out as we are busy with other things, so this API is not receiving all the love and attention it should have.

The best way to contibute to it is to garther various use-cases to know what patterns of sample/throttleFirst/throttleLast etc. are widespread and what are not and propose the potential API shape (or the vector of thought) that is composable.

PR #2747, opened 2021-06-07
Improve ':kotlinx-coroutines-core' substitution for all modules.

Tapchicoma:
Ensure "org.jetbrains.kotlinx:kotlinx-coroutines-core" transitive
dependencies are substituted with project itself to avoid test errors.


qwwdfsad (approved):
I'm okay with the final change with dependency substitution, thanks!

Regarding the PR in general -- it seems that most of the commits already exists in `kotlinx.train-1.5.20`, so it's likely that something went wrong with a rebase

PR #2755, opened 2021-06-10
Improve `supervisorScope` KDoc

erikhuizinga:
Previously, the KDoc wasn't explicit about the behaviour of `supervisorScope`, which is similar to the behaviour of the `coroutineScope` function.
This was confusing (at least for me), for example here:
https://kotlinlang.slack.com/archives/C1CFAFJSK/p1623338894383500?thread_ts=1623318861.372800&cid=C1CFAFJSK

This commit intends to redirect the reader to the KDoc of `coroutineScope`, where all behaviour details are explained very well.
The difference in behaviour is preserved, including the important reference to the `SupervisorJob` documentation.

---

I ran `./gradlew knit` successfully locally, but there were no additional changes to commit.


qwwdfsad:
Thanks for pointing it out!

Rephrased our docs and opened https://github.com/Kotlin/kotlinx.coroutines/pull/2757

PR #2761, opened 2021-06-14
Update code result

Jeff-Eu:
Update code result on the "Processing the latest value" section. The total time should be around 600 ms.


qwwdfsad:
600ms is what the summary of all `delay`s should be, but the program also spends its time on class loading, initialization and so on.

In fact, the time in the sample is just one of the runs we've made. I've played with it a few times -- most of the time it shows around ~700ms and never the value less than 690. I'd stick with a more realistic timeline in such case

PR #2765, opened 2021-06-15
Replace InternalCoroutinesApi with deprecation in NonCancellable

qwwdfsad:
* ICA is misused here as lint, for which we consistently use @Deprecated in other places
    * The current state is incompatible with 1.5.30 where KT-45844 is implemented


qwwdfsad:
Will be promoted to `ERROR` in the next major release

dkhalanskyjb (approved):

PR #2780, opened 2021-06-21
Clarify Dispatchers.IO's implementation note

joffrey-bion:
This implementation note mentions that withContext(IO) does not lead to a
context switch, which may be misunderstood as a general statement, while
it in fact only applies to switches between Default and IO dispatchers.

This is one example of misunderstanding:
https://stackoverflow.com/questions/68069529/is-kotlinx-coroutines-withcontext-safe-to-use-with-spring-webflux


qwwdfsad (approved):
Nice addition, thanks!

PR #2782, opened 2021-06-22
Add integration with Huawei Mobile Service Tasks API

burnoo:
Huawei Mobile Services library provides very similar Tasks API to Google Mobile Services. It helps developers easily switch between this two libraries.

This PR contains HMS integration - with the same Coroutines API as GMS. Implementation was copied from `play-services` module, and then adapted for Huawei Services. Differences are visible in this commit: https://github.com/Kotlin/kotlinx.coroutines/commit/d3ce28a8d0ad8daeef743124826581a3d0b1e249

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@ suspend fun main() = coroutineScope {
 * [ui](ui/README.md) &mdash; modules that provide coroutine dispatchers for various single-threaded UI libraries:
   * Android, JavaFX, and Swing.
 * [integration](integration/README.md) &mdash; modules that provide integration with various asynchronous callback- and future-based libraries:
-  * JDK8 [CompletionStage.await], Guava [ListenableFuture.await], Google Play Services [Task.await][gms.Task.await] and Huawei Mobile Services [Task.await][hms.Task.await];
+  * JDK8 [CompletionStage.await], Guava [ListenableFuture.await], Google Play Services [Task.await] and Huawei Mobile Services [Task.await](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-huawei-services/kotlinx.coroutines.tasks/com.google.android.gms.tasks.-task/await.html
+    );
   * SLF4J MDC integration via [MDCContext].
 
 ## Documentation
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -55,7 +55,8 @@ suspend fun main() = coroutineScope {
 * [ui](ui/README.md) &mdash; modules that provide coroutine dispatchers for various single-threaded UI libraries:
   * Android, JavaFX, and Swing.
 * [integration](integration/README.md) &mdash; modules that provide integration with various asynchronous callback- and future-based libraries:
-  * JDK8 [CompletionStage.await], Guava [ListenableFuture.await], and Google Play Services [Task.await];
+  * JDK8 [CompletionStage.await], Guava [ListenableFuture.await], Google Play Services [Task.await] and Huawei Mobile Services [Task.await](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-huawei-services/kotlinx.coroutines.tasks/com.google.android.gms.tasks.-task/await.html
```
burnoo, line 58: I tried to include docs link by a reference (the same way as others), but `knit` tool was removing it. Do you know why?

integration/kotlinx-coroutines-huawei-services/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+
+val tasksVersion = "1.5.0.300"
+
+val artifactType = Attribute.of("artifactType", String::class.java)
+val unpackedAar = Attribute.of("unpackedAar", Boolean::class.javaObjectType)
+
+configurations.configureEach {
+    afterEvaluate {
+        if (isCanBeResolved) {
+            attributes.attribute(unpackedAar, true) // request all AARs to be unpacked
```
burnoo, line 13: I found `unpackAar` task in `google-services` and I copied it as is, but is it needed?

qwwdfsad (commented):
Hi, thanks for the effort!

Unfortunately, I cannot accept this PR and suggest maintaining a separate self-contained library instead, at least for a while.

Maintaining nearly-identical copy-pasted modules is quite a pain in a long term (thinking of our reactive integrations) and before committing to that, we would like to know whether there exists any demand on it at all.

It's kind of a red flag when the first time we're hearing about HMS API is from PR that supports it, not from our users' feedback/issues/slack channel.
So instead I suggest creating a corresponding issue and see if anybody is interested in it, so later on we can decide on its support in kotlinx.coroutines

burnoo:
Hi, thanks for the fast response!

As you suggest I will publish these changes as a 3rd party library.

PR #2784, opened 2021-06-22
Cancel dispatched coroutine on Dispatchers.IO when the underlying Han…

### LINKED ISSUE #2778: Handler.asCoroutineDispatcher() can block runBlocking() forever
# Summary

[CoroutineDispatcher.dispatch()](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/dispatch.html) kdoc says:

> This method should guarantee that the given block will be eventually invoked, otherwise the system may reach a deadlock state and never leave it.

The `CoroutineDispatcher` instance returned by [Handler.asCoroutineDispatcher()](https://github.com/Kotlin/kotlinx.coroutines/blob/b852b20131fff5c0d605d7cb5f8001b715f02c1e/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt#L69) does not provide that guarantee: if the backing Handler thread has quit, `Handler.post()` will return false, but the dispatcher does not check for that and the coroutine does not run ([source](https://github.com/Kotlin/kotlinx.coroutines/blob/b852b20131fff5c0d605d7cb5f8001b715f02c1e/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt#L131-L133)).

```kotlin
  handler.post(block)
}
```

`runBlocking()` is an example of method that expects the coroutine dispatcher to run or cancel the coroutine, and otherwise will block forever.

`Executor.asCoroutineDispatcher()` implements this guarantee properly ([source](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/Executors.kt#L92-L100)):

```kotlin
        try {
            executor.execute(wrapTask(block))
        } catch (e: RejectedExecutionException) {
            unTrackTask()
            cancelJobOnRejection(context, e)
            Dispatchers.IO.dispatch(context, block)
        }
    }
```

# Example code

```kotlin
// setup
val thread = HandlerThread("stop stopping")
thread.start()
val handler = Handler(thread.looper)
val dispatcher = handler.asCoroutineDispatcher()

// Quit the thread, after this Handler.post() now returns false.
thread.quit()
// runBlocking waits for the dispatcher to run or cancel the coroutine. The current implementation does nothing.
runBlocking(dispatcher) {
  // This  never runs
}
```

Note: on Android, the main thread (which is a HandlerThread) never quits. However it's super common to create HandlerThread instances for single threaded work, or reuse existing instances when progressively migrating to coroutines. And it's common to quit such threads when running UI tests in CI and resetting the app in between every thread.

We ran into this edge case at Square, I posted the investigation on twitter: https://twitter.com/Piwai/status/1406301224295710720

qwwdfsad:
…dler is closed in Handler.asCoroutineDispatcher()

Fixes #2778

### DIFF
--- a/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
+++ b/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
@@ -161,7 +161,7 @@ internal class HandlerContext private constructor(
     }
 
     private fun cancelOnRejection(context: CoroutineContext, block: Runnable) {
-        context.cancel(CancellationException("The task was rejected"))
+        context.cancel(CancellationException("The task was rejected, the underlying handler '$handler' was closed"))
         Dispatchers.IO.dispatch(context, block)
     }
 
--- a/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
+++ b/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
@@ -7,8 +7,8 @@
 package kotlinx.coroutines.android
 
 import android.os.*
-import androidx.annotation.*
 import android.view.*
+import androidx.annotation.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.internal.*
 import java.lang.reflect.*
@@ -54,15 +54,22 @@ internal class AndroidDispatcherFactory : MainDispatcherFactory {
     override fun createDispatcher(allFactories: List<MainDispatcherFactory>) =
         HandlerContext(Looper.getMainLooper().asHandler(async = true))
 
-    override fun hintOnError(): String? = "For tests Dispatchers.setMain from kotlinx-coroutines-test module can be used"
+    override fun hintOnError(): String = "For tests Dispatchers.setMain from kotlinx-coroutines-test module can be used"
 
     override val loadPriority: Int
         get() = Int.MAX_VALUE / 2
 }
 
 /**
- * Represents an arbitrary [Handler] as a implementation of [CoroutineDispatcher]
+ * Represents an arbitrary [Handler] as an implementation of [CoroutineDispatcher]
  * with an optional [name] for nicer debugging
+ *
+ * ## Rejected execution
+ *
+ * If the underlying handler is closed and its post-related methods start to return `false` on
+ * an attempt to submit a continuation task to the resulting dispatcher,
+ * then the [Job] of the affected task is [cancelled][Job.cancel] and the task is submitted to the
+ * [Dispatchers.IO], so that the affected coroutine can cleanup its resources and promptly complete.
  */
 @JvmName("from") // this is for a nice Java API, see issue #255
 @JvmOverloads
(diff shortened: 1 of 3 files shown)


ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -129,24 +136,33 @@ internal class HandlerContext private constructor(
...
-            }
+        if (handler.postDelayed(block, timeMillis.coerceAtMost(MAX_DELAY))) {
+            return DisposableHandle { handler.removeCallbacks(block) }
         }
+        cancelOnRejection(context, block)
+        return NonDisposableHandle
+    }
+
+    private fun cancelOnRejection(context: CoroutineContext, block: Runnable) {
+        context.cancel(CancellationException("The task was rejected"))
```
dkhalanskyjb, line 164: This could be made more helpful by including `this.toString()`. By the way, why doesn't Android's `Dispatcher.Main` have the `name` set?

ui/kotlinx-coroutines-android/test/DisabledHandlerTest.kt
```diff
@@ -0,0 +1,64 @@
...
+        finish(3)
+    }
+
+    @Test
+    fun testWithTimeout() = runTest {
+        delegateToSuper = true
+        try {
+            withContext(disabledDispatcher) {
+                expect(1)
+                delegateToSuper = false
```
dkhalanskyjb, line 55: Maybe set this in a `finally` so that, if this test ever crashes, the others are not affected and don't increase the confusion?

ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -54,15 +54,22 @@ internal class AndroidDispatcherFactory : MainDispatcherFactory {
...
 }
 
 /**
- * Represents an arbitrary [Handler] as a implementation of [CoroutineDispatcher]
+ * Represents an arbitrary [Handler] as an implementation of [CoroutineDispatcher]
  * with an optional [name] for nicer debugging
+ *
+ * ## Rejected execution
+ *
+ * If the underlying handler is closed and its post-related methods start to return `false` on
```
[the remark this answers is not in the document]
qwwdfsad: Framed as `messages-scheduling methods`

ui/kotlinx-coroutines-android/test/DisabledHandlerTest.kt
qwwdfsad: The object of the test class is instantiated separately for each individual test and each new test is invoked on an isolated instance.

ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
dkhalanskyjb: > This could be made more helpful by including this.toString().

Doesn't this part still stand?

qwwdfsad: My bad, added more contextual information

ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -129,24 +136,33 @@ internal class HandlerContext private constructor(
...
-            }
+        if (handler.postDelayed(block, timeMillis.coerceAtMost(MAX_DELAY))) {
+            return DisposableHandle { handler.removeCallbacks(block) }
         }
+        cancelOnRejection(context, block)
+        return NonDisposableHandle
+    }
+
+    private fun cancelOnRejection(context: CoroutineContext, block: Runnable) {
+        context.cancel(CancellationException("The task was rejected, the underlying handler '$handler' was closed"))
```
dkhalanskyjb, line 164:
```suggestion
        context.cancel(CancellationException("The task was rejected, the handler '$handler' underlying the dispatcher was closed"))
```
Still think that employing `name` (for example, by calling `toString`) better reflects the user intent of supplying the `name`, but it's your call.

dkhalanskyjb (approved):

PR #2786, opened 2021-06-23
Pr/2528

qwwdfsad:
This is a redacted mirror of #2528.

I've reviewed and accepted it, but it would be nice to have a second pair of eyes here in case I missed something during the course of multiple review iterations

### DIFF
--- a/integration/kotlinx-coroutines-play-services/src/Tasks.kt
+++ b/integration/kotlinx-coroutines-play-services/src/Tasks.kt
@@ -8,7 +8,6 @@ package kotlinx.coroutines.tasks
 
 import com.google.android.gms.tasks.*
 import kotlinx.coroutines.*
-import kotlin.contracts.*
 import kotlin.coroutines.*
 
 /**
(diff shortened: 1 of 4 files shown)


integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -45,29 +38,56 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
+ * For bi-directional cancellation, an overload that accepts [CancellationTokenSource] can be used.
  */
-public fun <T> Task<T>.asDeferred(): Deferred<T> {
+public fun <T> Task<T>.asDeferred(): Deferred<T> = asDeferredImpl(null)
+
+/**
+ * Converts this task to an instance of [Deferred] with a [CancellationTokenSource] to control cancellation.
+ * The cancellation of this function is bi-directional:
+ * If the given task is cancelled, the resulting deferred will be cancelled.
+ * If the resulting deferred is cancelled, the provided [cancellationTokenSource] will be cancelled.
```
dkhalanskyjb, line 49:
```suggestion
 * * If the given task is cancelled, the resulting deferred will be cancelled.
 * * If the resulting deferred is cancelled, the provided [cancellationTokenSource] will be cancelled.
```
The IDE ignores newlines when formatting doc comments, so the layout looks confusing otherwise.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -45,29 +38,56 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
         val e = exception
-        return if (e == null) {
-            @Suppress("UNCHECKED_CAST")
-            CompletableDeferred<T>().apply { if (isCanceled) cancel() else complete(result as T) }
+        if (e == null) {
+            if (isCanceled) {
+                deferred.cancel()
+            } else {
+                @Suppress("UNCHECKED_CAST")
+                deferred.complete(this.result as T)
```
dkhalanskyjb, line 67: Why is `this.` needed?

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -45,29 +38,56 @@ public fun <T> Deferred<T>.asTask(): Task<T> {
...
-        } else {
-            result.completeExceptionally(e)
+    if (cancellationTokenSource != null) {
+        deferred.invokeOnCompletion {
+            cancellationTokenSource.cancel()
         }
     }
-    return result
+
+    return object : Deferred<T> by deferred {}
```
dkhalanskyjb, line 90:
```suggestion
    /** This is not just `deferred` so that it's impossible for the user to cast the result to `CompletableDeferred` and
      * complete the deferred on their own, which would break the invariants of this method. */
    return object : Deferred<T> by deferred {}
```
I'm not sure though if it's *impossible*, as I don't know if the objects that are `by`-delegated to are stored in some field internally.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -76,8 +96,26 @@ public fun <T> Task<T>.asDeferred(): Deferred<T> {
...
  * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
  * stops waiting for the completion stage and immediately resumes with [CancellationException].
+ *
+ * For bi-directional cancellation, an overload that accepts [CancellationTokenSource] can be used.
  */
-public suspend fun <T> Task<T>.await(): T {
+public suspend fun <T> Task<T>.await(): T = awaitImpl(null)
+
+/**
+ * Awaits for completion of the task that is linked to the given [CancellationTokenSource] to control cancellation.
```
dkhalanskyjb, line 105:
```suggestion
 * Awaits the completion of the task that is linked to the given [CancellationTokenSource] to control cancellation.
```
The same goes for the `await` above.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -76,8 +96,26 @@ public fun <T> Task<T>.asDeferred(): Deferred<T> {
...
+ *
+ * This suspending function is cancellable and cancellation is bi-directional:
+ * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
+ * cancels the [cancellationTokenSource] and throws a [CancellationException].
+ * If the task is cancelled, then this function will throw a [CancellationException].
```
dkhalanskyjb, line 110:
```suggestion
 * * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
 * cancels the [cancellationTokenSource] and throws a [CancellationException].
 * * If the task is cancelled, then this function will throw a [CancellationException].
```

integration/kotlinx-coroutines-play-services/src/Tasks.kt
```diff
@@ -95,13 +133,18 @@ public suspend fun <T> Task<T>.await(): T {
...
+                if (it.isCanceled) cont.cancel() else cont.resume(it.result as T)
             } else {
                 cont.resumeWithException(e)
             }
         }
+
+        if (cancellationTokenSource == null) return@suspendCancellableCoroutine
+        cont.invokeOnCancellation {
+            cancellationTokenSource.cancel()
+        }
```
dkhalanskyjb, line 148: Minor, opinion: I know that you like things de-indented, but to me, this looks just unnatural.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: Good catch!

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: Not really needed, seems to be a leftover

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: It is stored in the private field indeed, tho we never account malicious intents to break the API via reflection.

integration/kotlinx-coroutines-play-services/src/Tasks.kt
qwwdfsad: Sure, reverted the condition

dkhalanskyjb (approved):

PR #2790, opened 2021-06-25
Make DisposableHandle fun interface, remove our own adapter

qwwdfsad:
It was internal and was not part of the ABI as it was internal.
So no existing code should break (well, except for those who called internal API via reflection) and source compatibility is preserved


dkhalanskyjb (approved):
I checked to the best of my ability, and it does look like source-level compatibility is preserved.

qwwdfsad:
Seems that `::DisposableHandle` is not supported for fun interfaces, so let's postpone it for a while

PR #2795, opened 2021-06-29
Make helpRemove actually lock-free and address multiple chained remov…

### LINKED ISSUE #2590: Investigate failure of MutexLincheckTest.modelCheckingTest after updating to new Lincheck

qwwdfsad:
…ed nodes

    * The previous version invoked correctPrev on the next node and bailed out fi it was removed, effectively blocking the progress of any attempts to remove previous node
    * Use helpRemovePrev in helpRemove that was actually implemented to avoid such problem

Fixes #2590


elizarov (approved):
👏 Amazing discovery!

PR #2803, opened 2021-07-01
Introduce SharedFlow collect overload and override that return Nothing

### LINKED ISSUE #2502: Proposal: Provide inline overloads of terminal operators returning Nothing for SharedFlow
That would make it easier to mark functions that never return (responding only to cancellation after being called) because they collect an infinite flow.

### LINKED ISSUE #2789: Consider "overriding" SharedFlow.collect to return Nothing
Since a `SharedFlow` never ends and collecting it "never completes normally," it'd be possible, semantically correct, and possibly illustrative to add a "fauxveride" of `SharedFlow.collect` returning `Nothing`.

Since it's an extension method, it can't actually be an override, but adding it as an overload should have effectively the same result.

Inspired by https://stackoverflow.com/q/68121398/869736.

qwwdfsad:
* Override will ensure the proper implementation of the interface
    * collect extension is added as the very basic lint helper

Fixes #2789
Fixes #2502


LouisCAD:
> I've decided not to add Nothing overload for each terminal extension as it seems to be too much of a hassle.

Too much of a hassle to do, or from another standpoint like maintaining?

I'm asking to know if further PRs from the community to add them for terminal operators have a chance of getting merged.

hfhbd:
Hey, 
If I apply this patch and use a local object implementing `SharedFlow`:
```kotlin
interface Clock: SharedFlow<Int> { ... }
fun Clock() = object: Clock { }
```
the execution fails with `KotlinNothingValueException` before the second `emit`. Using an extra class `class ClockImpl: Clock` works without failing. Did you noticed this behavior too?

https://youtrack.jetbrains.com/issue/KT-48982

PR #2804, opened 2021-07-02
Terminate the process in case of an unhandled exception

### LINKED ISSUE #2398: Throw an exception inside coroutine launched with Dispatchers.Default doesn't crash the app
I throw an exception inside coroutine launched with Dispatchers.Default and it doesn't crash the app (just print stack trace in console). Works fine if I use Dispatchers.Main instead of Dispatchers.Default.

Here is an example project: https://drive.google.com/file/d/1NzwMU7QStgQuKBNgty1HZei1ZZvQG5Ys/view?usp=sharing

Kotlin: 1.4.20-RC
Coroutines: 1.3.9-native-mt
iOS: 14.1 (iPhone 11, simulator)

mvicsokolova:
Fixes #2398


kotlinx-coroutines-core/native/src/CoroutineExceptionHandlerImpl.kt
```diff
@@ -5,11 +5,14 @@
...
+import kotlin.native.concurrent.*
 
 internal actual fun handleCoroutineExceptionImpl(context: CoroutineContext, exception: Throwable) {
-    // log exception
-//    println("Exception in \"${Worker.current}\"")
-//    exception.printStackTrace()
-// todo: printing exception does not make it easy to debug (no source location), so let it crash instead
-    throw exception
+    // Use unhandled exception hook if it is set, otherwise print the stacktrace
+    val hook = setUnhandledExceptionHook({ _: Throwable -> }.freeze())
```
qwwdfsad, line 14: Apart from being hacky (which definitely requires additional explanation), this code is not thread-safe.

kotlinx-coroutines-core/native/src/CoroutineExceptionHandlerImpl.kt
```diff
@@ -5,11 +5,14 @@
...
+    if (hook != null) hook(exception) else exception.printStackTrace()
```
qwwdfsad, line 15: It worth prepending the coroutine context and the fact that an app is terminated due to unhandled exceptions before calling `exitProcess(-1)`

qwwdfsad (changes requested):

kotlinx-coroutines-core/native/src/CoroutineExceptionHandlerImpl.kt
pacher: Hmmm.... I see that this PR is related to native and I am not concerned at the moment. Also I don't do mobile an don't know the details of setup, so my question might look stupid. But can we be sure that this change is confined to iOS? In other words, are we sure that kotlin native with coroutines will never be used for server side or desktop development? Because `exitProcess(-1)` seems too radical for me on the server, I would definitely hate it. Again sorry for stupid question.

qwwdfsad: Your question raised an important issue that we're going to address, thanks for that!

It is indeed important to have an ability _not_ to terminate an application on unhandled exceptions, but it also important to be able to do so.
KMM and iOS is our primary focus for now and there is only an API to do the latter, not the former.

So, in the `1.5.1-native-mt` the default behaviour will be to crash (in the end, an unhandled exception in a coroutine is no different from the unhandled exception from `main` or a worker) for the sake of iOS apps. The workaround to avoid application crashes is not to leave exceptions unattended (e.g. by using `CoroutinesExceptionHandler`).

But in 1.5.30 (or in 1.6.0, it's being discussed right now) we'll provide a new K/N stdlib API that allows to invoke unhandled exception hook without termination (now this API is supposed to be used _only_ with the consecutive call to `abort` or `exitProcess`) and will provide an opt-in mechanism to set your own exception hook without crashing.

qwwdfsad:
Merged manually

PR #2809, opened 2021-07-06
General cleanup

qwwdfsad:
* Do not generate Dokka documentation for deprecated members
* Get rid of inline classes notation 
* Opt-in into DelicateCoroutineApi in our own project
* Get rid of JVM IR checks that are no longer relevant

### DIFF
--- a/buildSrc/src/main/kotlin/Publishing.kt
+++ b/buildSrc/src/main/kotlin/Publishing.kt
@@ -40,12 +40,8 @@ fun MavenPom.configureMavenCentralMetadata(project: Project) {
 }
 
 fun mavenRepositoryUri(): URI {
-    // TODO -SNAPSHOT detection can be made here as well
     val repositoryId: String? = System.getenv("libs.repository.id")
     return if (repositoryId == null) {
-        // Using implicitly created staging, for MPP it's likely to be a mistake because
-        // publication on TeamCity will create 3 independent staging repositories
-        System.err.println("Warning: using an implicitly created staging for coroutines")
         URI("https://oss.sonatype.org/service/local/staging/deploy/maven2/")
     } else {
         URI("https://oss.sonatype.org/service/local/staging/deployByRepositoryId/$repositoryId")
--- a/gradle/opt-in.gradle
+++ b/gradle/opt-in.gradle
@@ -3,10 +3,10 @@
  */
 
 // For new mpp
-ext.experimentalAnnotations = [
-        "kotlin.Experimental",
+ext.optInAnnotations = [
         "kotlin.experimental.ExperimentalTypeInference",
         "kotlin.ExperimentalMultiplatform",
+        "kotlinx.coroutines.DelicateCoroutinesApi",
         "kotlinx.coroutines.ExperimentalCoroutinesApi",
         "kotlinx.coroutines.ObsoleteCoroutinesApi",
         "kotlinx.coroutines.InternalCoroutinesApi",
--- a/gradle/opt-in.gradle
+++ b/gradle/opt-in.gradle
@@ -2,7 +2,6 @@
  * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
-// For new mpp
 ext.optInAnnotations = [
         "kotlin.experimental.ExperimentalTypeInference",
         "kotlin.ExperimentalMultiplatform",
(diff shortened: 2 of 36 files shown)


reactive/kotlinx-coroutines-rx3/src/RxConvert.kt
```diff
@@ -139,13 +139,15 @@ public fun <T: Any> Flow<T>.asObservable(context: CoroutineContext = EmptyCorout
...
 
+/** @suppress */
 @Suppress("UNUSED") // KT-42513
 @JvmOverloads // binary compatibility
 @JvmName("from")
 @Deprecated(level = DeprecationLevel.HIDDEN, message = "") // Since 1.4, was experimental prior to that
 public fun <T: Any> Flow<T>._asFlowable(context: CoroutineContext = EmptyCoroutineContext): Flowable<T> =
     asFlowable(context)
 
+/* @suppress */
```
dkhalanskyjb, line 150: Not a KDoc?

kotlinx-coroutines-core/common/src/flow/operators/Context.kt
```diff
@@ -278,37 +278,7 @@ private class CancellableFlowImpl<T>(private val flow: Flow<T>) : CancellableFlo
...
- *     ```
- * 2) To avoid such leaks, new primitive should be introduced to `kotlinx.coroutines` -- the subtraction of contexts.
- *    And this will become a new concept to learn, maintain and explain.
- * 3) It defers the execution of declarative [builder] until the moment of [collection][Flow.collect] similarly
- *    to `Observable.defer`. But it is unexpected because nothing in the name `flowWith` reflects this fact.
- * 4) It can be confused with [flowOn] operator, though [flowWith] is much rarer.
+ * @suppress
  */
 @FlowPreview
 @Deprecated(message = "flowWith is deprecated without replacement, please refer to its KDoc for an explanation", level = DeprecationLevel.ERROR) // Error in beta release, removal in 1.4
```
dkhalanskyjb, line 284: > please refer to its KDoc for an explanation

Won't be possible.

gradle/opt-in.gradle
```diff
@@ -3,10 +3,10 @@
  */
 
 // For new mpp
```
dkhalanskyjb, line 5: This comment doesn't say much to me.

buildSrc/src/main/kotlin/Publishing.kt
```diff
@@ -40,12 +40,8 @@ fun MavenPom.configureMavenCentralMetadata(project: Project) {
 }
 
 fun mavenRepositoryUri(): URI {
-    // TODO -SNAPSHOT detection can be made here as well
     val repositoryId: String? = System.getenv("libs.repository.id")
     return if (repositoryId == null) {
-        // Using implicitly created staging, for MPP it's likely to be a mistake because
-        // publication on TeamCity will create 3 independent staging repositories
-        System.err.println("Warning: using an implicitly created staging for coroutines")
```
dkhalanskyjb, line 48: I don't understand the changes to this file. They don't affect anything, but still, could you explain them?

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/flow/operators/Context.kt
qwwdfsad: KDoc is still part of the lib sources, we suppress it only out of generated site

qwwdfsad: Returned it

gradle/opt-in.gradle
qwwdfsad: It's obsolete, removed it

buildSrc/src/main/kotlin/Publishing.kt
qwwdfsad: Our library is multiplatform and thus should be published from multiple machines (one per supported K/N architecture).

When we're configuring publication, we have to specify the repository we're going to publish to.
The default recommended/implicitly used address is `/staging/deploy/maven2/`. When it's used, a new staging repository is created by the Maven plugin. Essentially it means that, when we're publishing from multiple agents, we'll create multiple staging repositories (=> multiple releases with the same version and partial artifacts set).
To avoid that, we preconfigure our builds with a pre-created staging repository and pass its id via `libs.repository.id` for publishing tasks.

We do not do that for regular builds, but Gradle still configures publishing tasks, so this warning has been printed as many times as we got modules on each Gradle configuration, obfuscating our build output

dkhalanskyjb (approved):

PR #2814, opened 2021-07-09
Update coroutine-context-and-dispatchers.md

BattleShipPark:
I found an uncompleted sentence.

### DIFF
--- a/docs/topics/coroutine-context-and-dispatchers.md
+++ b/docs/topics/coroutine-context-and-dispatchers.md
@@ -65,8 +65,7 @@ context of the main `runBlocking` coroutine which runs in the `main` thread.
 [Dispatchers.Unconfined] is a special dispatcher that also appears to run in the `main` thread, but it is,
 in fact, a different mechanism that is explained later.
 
-The default dispatcher that is used when no other dispatcher is explicitly specified in the scope.
-It is represented by [Dispatchers.Default] and uses a shared background pool of threads.
+[Dispatchers.Default] is a dispatcher that uses a shared background pool of threads.
   
 [newSingleThreadContext] creates a thread for the coroutine to run. 
 A dedicated thread is a very expensive resource. 


docs/topics/coroutine-context-and-dispatchers.md
```diff
@@ -65,8 +65,7 @@ context of the main `runBlocking` coroutine which runs in the `main` thread.
 [Dispatchers.Unconfined] is a special dispatcher that also appears to run in the `main` thread, but it is,
 in fact, a different mechanism that is explained later.
 
-The default dispatcher that is used when no other dispatcher is explicitly specified in the scope.
```
dkhalanskyjb, line 68: I think the correct thing to do here is to remove the word `that`. If we remove the sentence, the (as is, incorrectly stated) piece of information that there exists a default fallback dispatcher will be lost.

BattleShipPark: We have two terms: a default dispatcher and Dispatchers.Default.

For default dispatcher, its explanation is done before this line like this:
> When launch { ... } is used without parameters, it inherits the context (and thus dispatcher) from the CoroutineScope it is being launched from. In this case, it inherits the context of the main runBlocking coroutine which runs in the main thread.

In the line, the document need to explain Dispatchers.Default.

dkhalanskyjb: > For default dispatcher, its explanation is done before this line like this

This is not the case. The explanation that you cited refers to inherited contexts. The default dispatcher is `Dispatchers.Default`. Please see the docs for `Dispatchers.Default` or, if you're still not convinced, this code that is referenced throughout the codebase: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt#L32-L37

BattleShipPark: Good. I misunderstood the default dispatcher. Thanks 👍 
Do I need to send PR again removing only the word "that"?

qwwdfsad: Adjusting this PR should be enough

PR #2824, opened 2021-07-15
Fix js tests

qwwdfsad:


kotlinx-coroutines-core/common/test/CancelledParentAttachTest.kt
```diff
@@ -11,75 +11,105 @@ import kotlin.test.*
...
-    private fun testLaunchCancelledParent(start: CoroutineStart) =
-        runTest({ it is CancellationException }) {
-            cancel()
-            expect(1)
-            val d = launch(start = start) { }
-            expect(2)
-            d.invokeOnCompletion {
-                finish(3)
-                reset()
+    private suspend fun testLaunchCancelledParent(start: CoroutineStart) {
```
qwwdfsad, line 41: I've actually tried a less intrusive solution with chaining multiple `runTest` into a single promise, but it turned out to be almost impossible due to `GlobalScope.*` and undefined order between resolving of `runTest` promise and invocation of `invokeOnCompletion` in children bound to the resulting promise.

kotlinx-coroutines-core/js/test/TestBase.kt
```diff
@@ -79,7 +81,10 @@ public actual open class TestBase actual constructor() {
     ): dynamic {
         var exCount = 0
         var ex: Throwable? = null
-        return GlobalScope.promise(block = block, context = CoroutineExceptionHandler { context, e ->
+        if (lastTestPromise != null) {
+            error("Attempt to run multiple asynchronous test within one @Test method")
```
dkhalanskyjb, line 85: I think that this code looks unintuitive without additional comments surrounding it. What exactly happens when a promise is created but isn't returned? When this gets merged, `git blame` will show this PR, showing that this causes some errors  to not be reported, which makes it possible to establish the context, but I think a comment would not be too much effort but would go a long way.

kotlinx-coroutines-core/js/test/TestBase.kt
```diff
@@ -10,9 +10,11 @@ public actual val isStressTest: Boolean = false
 public actual val stressTestMultiplier: Int = 1
 
 public actual open class TestBase actual constructor() {
+    public actual val isJs = true
```
dkhalanskyjb, line 13: I strongly dislike the name of this flag.

Consider the following scenario: we learn that there's no longer a two-second timeout on JS (also, where does the timeout come from?). However, it doesn't give us a clear indication to change anything about these tests: `isJs` should still be true. Instead, we're supposed to go looking through the tests to find those that are affected by the timeout no longer being there.

As an alternative, there's a much worse-sounding but, in my opinion, the more accurate name `timesOutAfterTwoSeconds` or something like this. This is what distinguishes the JS platform currently, this is what tests check for, the documentation for it is the place that could explain where the timeout comes from.

kotlinx-coroutines-core/js/test/TestBase.kt
qwwdfsad: Sure thing, added a detailed explanation

kotlinx-coroutines-core/js/test/TestBase.kt
qwwdfsad: >where does the timeout come from?

The underlying limitation of the Mocha test framework.

The rationale of having `js` in the name is mostly due to the fact that JS is tremendously slow for this kind of tests.
Renamed to `boundByJsTestTimeout` which seems to be a good middle-ground

kotlinx-coroutines-core/common/test/TestBase.common.kt
```diff
@@ -14,6 +14,7 @@ public expect val isStressTest: Boolean
 public expect val stressTestMultiplier: Int
 
 public expect open class TestBase constructor() {
+    public val isBoundByJsTestTimeout: Boolean
```
dkhalanskyjb, line 17: A line about the Mocha limitation would be good.

kotlinx-coroutines-core/js/test/TestBase.kt
dkhalanskyjb: Great! Much clearer this way.

kotlinx-coroutines-core/js/test/TestBase.kt
dkhalanskyjb: This is a good name, I agree.

dkhalanskyjb (approved):

PR #2827, opened 2021-07-16
Introduce TestResult into TestBase as the first preparation step to m…

qwwdfsad:
…ultiplatform tests

    * The goal of this change is to test our IDE and tooling tolerance to hack-ish TestResult


dkhalanskyjb (approved):
Yeah, I got the exact same implementation.

PR #2829, opened 2021-07-16
Remove workarounds for obsolete compiler bugs

qwwdfsad:


qwwdfsad:
For 1.6.0 as covariant override maybe ABI incompatible for JS IR and K/N

qwwdfsad:
It is

Issue #2797, opened 2021-06-29
Prepare kotlinx.coroutines for new Native memory model


qwwdfsad:
Since [`1.5.30-dev-2892`](https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev/org/jetbrains/kotlin/kotlin-gradle-plugin/1.5.30-dev-2892/), the prototype of the new memory model is available in Kotlin/Native (including mark-and-sweep GC, object sharing and cycles collection) and it can be enabled via `-memory-model experimental` flag (UPD: also requires `kotlin.native.cacheKind=none` in `gradle.properties`).

Current `develop` already passes all tests under the new model, but yet coroutines in their current state can only be used from a single thread.

Here is the list of things that we'd like to get done, first in a separate branch, and later merge it in the mainline, in order to make coroutines work in a multithreaded environment with a new memory model:

- [x] Reuse `concurrent` source-set and share our multithreaded primitives (notably, linked list) between JVM and K/N. As opposed to `native-mt`, no platform specific-parts are required here, it should be enough to copy JVM implementation based on atomics and just share it with Native
- [x] Make dispatchers actually multithreaded: introduce `newSingleThreadContext` backed by worker and make `Dispatchers.Default` backed by a worker as well
- [x] Copy most of our JVM tests (ones that test actual concurrency, not JVM-specific tweaks) into `concurrent` source-set. This work is already done as part of the `native-mt` work, it can be mirrored here. 
- [x] Make all tests pass :) 
- [x] Make coroutines prepared to KMM: ensure that `Dispatchers.Main` uses main thread on Darwin.
- [x] Optional: Make `Dispatchers.Default` to use [global dispatch queue](https://developer.apple.com/documentation/dispatch/dispatchqueue/2300077-global) and 

No other code from `native-mt` (freezes, cycle breakers, shareable primitives) should be used in this implementation in order to ensure the complete correctness and real-world applicability of new memory model

andersio:
A couple of random thoughts since libdispatch is mentioned:

* `dispatch_get_global_queue(qos, int)` by default gives a non-overcommit queue (bound by # of logical processor), which seems an ideal execution environment for Kotlin coroutines. [Relevant notes from Swift Concurrency](https://github.com/apple/swift/blob/914928e954ee02f21d086c863dd3b5e514911fd5/stdlib/public/Concurrency/GlobalExecutor.cpp#L319-L328).

[...]

PR #2831, opened 2021-07-16
Concurrent source-set for sharing between JVM and K/N

mvicsokolova:
Sharing concurrent parts between JVM and K/N 
#2797


qwwdfsad:
`Actor.kt` and `TickerChannels.kt` belong to JVM and should not be shared with Native. Otherwise looks good

PR #2833, opened 2021-07-16
Incorporate new Native memory model into kotlinx-coroutines mainline

### LINKED ISSUE #2914: Merge new Native memory model support into the mainline
We have announced our new experimental memory manager: https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview/ and expect our users to check it out.

This experience cannot be full without `kotlinx.coroutines`, but unfortunately [special support](https://github.com/Kotlin/kotlinx.coroutines/pull/2833) is required. It forces us to support "one more fork" of coroutines, but what's more important -- it forces users to depend on [dev](https://github.com/Kotlin/kotlinx.coroutines/issues/2797#issuecomment-895077197) builds of unspecified quality, and also somehow to be aware of their existense.

My proposal is to leverage [isExperimentalMM](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/main/kotlin/kotlin/native/Platform.kt#L139) stdlib API (available only in 1.6.0+) and merge the support of the new MM in the mainline, enabling new "sharing" behaviour (and also implementations of `Dispatchers.Main`, `Dispatchers.Default` and `new*Context()`) conditionally. If the new MM is disabled, we can fallback to our single-threaded mode.

qwwdfsad:
Changeset:

* When the new memory model is disabled, behave the same way as before

For the new memory model:
* Concurrent tests shared with JVM (disabled for old MM)
* `Dispatchers.Default` backed by multiple workers on Linux/Windows and global_queue on OS X
* `Dispatchers.Main` on OS X and iOS
* `newSingleThreadContext` and `newFixedThreadPoolContext` on Native, shared declaration with JVM

Fixes #2914


kotlinx-coroutines-core/native/src/CoroutineContext.kt
```diff
@@ -8,33 +8,29 @@ import kotlinx.coroutines.internal.*
 import kotlin.coroutines.*
 import kotlin.native.concurrent.*
 
-private fun takeEventLoop(): EventLoopImpl =
-    ThreadLocalEventLoop.currentOrNull() as? EventLoopImpl ?:
-        error("There is no event loop. Use runBlocking { ... } to start one.")
-
 internal actual object DefaultExecutor : CoroutineDispatcher(), Delay {
```
dkhalanskyjb, line 11: I checked the uses of `DefaultExecutor`. It seems there are two places where it can be accessed in Native outside of `DefaultDelay`:
* `kotlinx-coroutines-core/common/src/EventLoop.common.kt`: `EventLoopImplBase.enqueue`.
* `kotlinx-coroutines-core/native/src/EventLoop.kt`: `reschedule`.

The only way these could be triggered, it seems, is if the event loop is closed, but new tasks arrive.

The only way for the event loop to be closed on Native is via a call to `decrementUseCount` if the event loop is shared, which it is when `runBlocking` is used.

So, it looks like it's possible to access this object even when `multithreadingSupported` is false.

dkhalanskyjb (commented):
This is just the first batch of comments so there's something to discuss while I check the rest of this.

kotlinx-coroutines-core/jvm/test/ConcurrentTestUtilities.kt
```diff
@@ -0,0 +1,30 @@
...
+    @Volatile
+    var sink = 1
+}
+
+
+@Suppress("ACTUAL_WITHOUT_EXPECT")
+internal actual typealias SuppressSupportingThrowable = Throwable
+
+@Suppress("EXTENSION_SHADOWED_BY_MEMBER")
+actual fun Throwable.printStackTrace() = printStackTrace()
```
dkhalanskyjb, line 28:
```suggestion
actual fun Throwable.printStackTrace() {
    val printer: Throwable.() -> Unit = Throwable::printStackTrace
    printer()
}
```
Found this form that doesn't look like recursion and doesn't become one if the method is not resolved.

kotlinx-coroutines-core/native/src/CoroutineContext.kt
```diff
@@ -8,33 +8,29 @@ import kotlinx.coroutines.internal.*
...
+
+    private val delegate = WorkerDispatcher(name = "Dispatchers.Default")
```
dkhalanskyjb, line 13: Maybe the following would work?
```suggestion
    private val delegate by lazy { newSingleThreadContext(name = "Dispatchers.Default") }
```
Then we could make `WorkerDispatcher` private, and also be loudly notified if we accessed this in a single-threaded configuration.

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -0,0 +1,76 @@
...
+internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(), Delay {
+    private val worker = Worker.start(name = name)
+
+    override fun dispatch(context: CoroutineContext, block: Runnable) {
+        worker.executeAfter(0L) { block.run() }
+    }
+
+    override fun scheduleResumeAfterDelay(timeMillis: Long, continuation: CancellableContinuation<Unit>) {
+        worker.executeAfter(timeMillis.toMicrosSafe())
+        { with(continuation) { resumeUndispatched(Unit) } }
```
dkhalanskyjb, line 33: Does the outer pair of braces do anything?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -0,0 +1,76 @@
...
+        return NonDisposableHandle
+    }
+
+    override fun close() {
+        worker.requestTermination().result // Note: calling "result" blocks
+    }
+
+    private fun Long.toMicrosSafe(): Long {
+        val result = this * 1000
+        return if (result > this) result else Long.MAX_VALUE
```
dkhalanskyjb, line 48: Are we 100% sure that `invokeOnTimeout` is always called with a strictly positive value?
(If not, this is a +1 for the overflow-aware operations in the stdlib)

kotlinx-coroutines-core/native/src/EventLoop.kt
```diff
@@ -4,18 +4,35 @@
...
+    private val current = Worker.current
+
+    protected actual fun unpark() {
+        current.executeAfter(0L, EMPTY_BLOCK)// send an empty task to unpark the waiting event loop
+    }
+
+    protected actual fun reschedule(now: Long, delayedTask: EventLoopImplBase.DelayedTask): Unit {
+        DefaultExecutor.invokeOnTimeout(now, delayedTask, EmptyCoroutineContext)
+        Unit
+    }
```
dkhalanskyjb, line 26:
```suggestion
    protected actual fun reschedule(now: Long, delayedTask: EventLoopImplBase.DelayedTask) {
        DefaultExecutor.invokeOnTimeout(now, delayedTask, EmptyCoroutineContext)
    }
```

kotlinx-coroutines-core/native/src/Dispatchers.kt
```diff
@@ -4,18 +4,21 @@
 
 package kotlinx.coroutines
 
+import kotlinx.coroutines.internal.multithreadingSupported
 import kotlin.coroutines.*
+import kotlin.native.concurrent.*
 
 /** Not inside [Dispatchers], as otherwise mutating this throws an `InvalidMutabilityException`. */
+@ThreadLocal
```
dkhalanskyjb, line 12: Do I understand correctly that, if `Dispatchers.setMain` was used, then `withContext (Dispatchers.Main)` will behave differently depending on from which thread it's called? I don't think such behavior is desirable.

kotlinx-coroutines-core/concurrent/test/sync/SemaphoreStressTest.kt
```diff
@@ -41,12 +47,38 @@ class SemaphoreStressTest : TestBase() {
     }
 
     @Test
-    fun stressCancellation() = runBlocking(Dispatchers.Default) {
-        val n = 10_000 * stressTestMultiplier
+    fun testNativeSpecificStress() = runMtTest {
+        /*
+         * Dispatchers.Default is still single-threaded, so here we're doing actual concurrency
```
dkhalanskyjb, line 52: Could you explain this comment?

kotlinx-coroutines-core/concurrent/test/sync/SemaphoreStressTest.kt
```diff
@@ -1,18 +1,24 @@
-package kotlinx.coroutines.sync
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package sync
```
dkhalanskyjb, line 5: The IDE messed up here?

kotlinx-coroutines-core/concurrent/test/sync/MutexStressTest.kt
```diff
@@ -1,22 +1,49 @@
...
     @Test
-    fun testStress() = runBlocking(Dispatchers.Default) {
-        val n = 1000 * stressTestMultiplier
+    fun testDefaultDispatcher() = runMtTest { testBody(Dispatchers.Default) }
+
+    @Test
+    fun testSingleThreadContext() = runMtTest {
+        val ctx = newSingleThreadContext("testSingleThreadContext")
+        testBody(ctx)
+        ctx.close()
```
dkhalanskyjb, line 23: `use` here and below?

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -59,6 +59,7 @@ public abstract class CoroutineDispatcher :
      * This method should generally be exception-safe. An exception thrown from this method
      * may leave the coroutines that use this dispatcher in the inconsistent and hard to debug state.
      */
+    @ExperimentalCoroutinesApi
```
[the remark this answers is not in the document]
qwwdfsad: That's some weird rebase artifact, thanks for spotting

kotlinx-coroutines-core/common/test/TestBase.common.kt
```diff
@@ -20,6 +22,10 @@ public expect val stressTestMultiplier: Int
 @Suppress("NO_ACTUAL_FOR_EXPECT")
 public expect class TestResult
 
+public expect val isNative: Boolean
+// "Speedup" native stress tests
+public val stressTestNativeDivisor = if (isNative) 10 else 1
```
dkhalanskyjb, line 27: I get that some tests should have fewer iterations on Native. However, the existence of this constant seems to imply that *all* tests will run with fewer iterations. And this is true: when `isStressTest` is true, `stressTestMultiplier` is *also* lower than on the JVM. So, we have three disparate mechanisms for dealing with the number of stress test iterations:
* `stressTestMultiplier` when `isStressTest` is true;
* For some tests, `stressTestNativeDivisor`;
* For some other tests, manual checks and overrides.

Also, I see that `stressTestNativeDivisor` is only used in a few places. Can we maybe simplify this system a bit by getting rid of this constant and instead, doing manual checks in the few tests that it is needed in?

dkhalanskyjb (commented):
Another batch of comments.

gradle.properties
```diff
@@ -57,3 +57,4 @@ org.gradle.jvmargs=-Xmx4g
 
 kotlin.mpp.enableCompatibilityMetadataVariant=true
 kotlin.mpp.stability.nowarn=true
+kotlin.native.cacheKind=none
```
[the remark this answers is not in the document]
qwwdfsad: Removed this one after verifying (by asking the responsible person in K/N) it is no longer necessary

kotlinx-coroutines-core/concurrent/test/ConcurrentExceptionsStressTest.kt
```diff
@@ -0,0 +1,66 @@
...
+            (e as? StressException)?.index ?: unexpectedException("suppressed $index", e)
+        }
+        repeat(nWorkers) { index ->
+            assertTrue(index in indices, "Exception $index is missing: $indices")
+        }
+        assertEquals(nWorkers, indices.size, "Duplicated exceptions in list: $indices")
+    }
+
+    private fun unexpectedException(msg: String, e: Throwable?): Nothing {
+        e?.printStackTrace()
```
dkhalanskyjb, line 60: Why? Won't the cause just also be printed? Is this also a workaround for Native?

kotlinx-coroutines-core/concurrent/test/AtomicCancellationTest.kt
```diff
@@ -85,7 +85,7 @@ class AtomicCancellationTest : TestBase() {
     }
 
     @Test
-    fun testSelectDeferredAwaitCancellable() = runBlocking {
+    fun testSelectDeferredAwaitCancellable() = runMtTest {
```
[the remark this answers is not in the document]
qwwdfsad: Seems like all these tests should be indeed `runBlocking`

kotlinx-coroutines-core/native/test/WorkerTest.kt
```diff
@@ -23,12 +23,12 @@ class WorkerTest : TestBase() {
...
 
     @Test
-    fun testLaunchInWorkerTroughGlobalScope() {
+    fun testLaunchInWorkerThroughGlobalScope() {
         val worker = Worker.start()
         worker.execute(TransferMode.SAFE, { }) {
             runBlocking {
                 CoroutineScope(EmptyCoroutineContext).launch {
-                    delay(1)
+                    delay(10)
```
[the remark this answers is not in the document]
qwwdfsad: The former.

kotlinx-coroutines-core/concurrent/test/StateFlowUpdateCommonTest.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow
+
+import kotlinx.coroutines.*
+import kotlin.test.*
+import kotlin.test.Test
```
dkhalanskyjb, line 9:
```suggestion
import kotlin.test.*
```

kotlinx-coroutines-core/concurrent/test/StateFlowUpdateCommonTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
```
dkhalanskyjb, line 10:
```suggestion

// A simplified version of StateFlowUpdateStressTest 
```

dkhalanskyjb (commented):
Second-to-last dump of the initial comments while I brace myself to read the Darwin `Dispatchers.Main` implementation and tests.

kotlinx-coroutines-core/nativeDarwin/test/MainDispatcherTest.kt
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }
+            assertFalse(isMainThread())
+            finish(4)
+        }
+    }
+
+    @Test
+    fun testWithContextTimeoutDelayTimeout() {
+        if (!canTestMainDispatcher()) return // skip if already on the main thread, run blocking doesn't really work well with that
```
dkhalanskyjb, line 128: Style: maybe this comment belongs to `canTestMainDispatcher` itself. Or even better: why not make a wrapper around `runTest` that does this check, like `runTestNotOnMainDispatcher`?

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/jvm/test/ConcurrentTestUtilities.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import kotlin.random.*
+
+actual fun randomWait() {
+    val n = Random.nextInt(1000)
+    if (n < 500) return // no wait 50% of time
+    repeat(n) {
+        BlackHole.sink *= 3
+    }
+    if (n > 900) Thread.yield()
```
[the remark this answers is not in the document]
qwwdfsad: The JVM is not allowed to due to its lazy open worldness -- it cannot prove that there is no classes in the classpath, that are not loaded by will be and will read this field

kotlinx-coroutines-core/native/test/ConcurrentTestUtilities.kt
```diff
@@ -0,0 +1,35 @@
...
+    var sink = 1
+}
+
+internal actual typealias SuppressSupportingThrowable = SuppressSupportingThrowableImpl
+
+actual val Throwable.suppressed: Array<Throwable>
+    get() = (this as? SuppressSupportingThrowableImpl)?.suppressed ?: emptyArray()
+
+@Suppress("EXTENSION_SHADOWED_BY_MEMBER")
+actual fun Throwable.printStackTrace() = printStackTrace()
```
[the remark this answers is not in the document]
qwwdfsad: The code in Native source sets will be red for a while unless all known problems in HMPP are fixed, sorry for that.
The rule is simple -- member always wins resolve against an extension (thus `EXTENSION_SHADOWED_BY_MEMBER`).

>Could you replace this with a qualified name 

Alas, I can't do that for member function

kotlinx-coroutines-core/jvm/test/exceptions/Exceptions.kt
```diff
@@ -15,7 +15,7 @@ import kotlin.test.*
  * but run only under JDK 1.8
  */
 @Suppress("ConflictingExtensionProperty")
-val Throwable.suppressed: Array<Throwable> get() {
+actual val Throwable.suppressed: Array<Throwable> get() {
     val method = this::class.java.getMethod("getSuppressed") ?: error("This test can only be run using JDK 1.7")
```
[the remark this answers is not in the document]
qwwdfsad: We're in limbo here. Let's discuss it separately

kotlinx-coroutines-core/native/src/Builders.kt
```diff
@@ -50,31 +56,47 @@ public fun <T> runBlocking(context: CoroutineContext = EmptyCoroutineContext, bl
     return coroutine.joinBlocking()
 }
 
+@SharedImmutable
+internal val EMPTY_BLOCK: () -> Unit = {}
```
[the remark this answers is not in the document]
qwwdfsad: This particular case wasn't optimized by K/N for some reason. Performance impact is likely to be negligible, I've extracted it during performance investigation

kotlinx-coroutines-core/jvm/test/ConcurrentTestUtilities.kt
qwwdfsad: It is the same very recursion if the method is not here:
```
class Foo 

fun Foo.foo() {
    val t = Foo::foo
}
```

kotlinx-coroutines-core/native/src/CoroutineContext.kt
qwwdfsad: Fixed

kotlinx-coroutines-core/jvm/src/Builders.kt
```diff
@@ -35,7 +35,7 @@ import kotlin.coroutines.*
  * @param block the coroutine code.
  */
 @Throws(InterruptedException::class)
-public fun <T> runBlocking(context: CoroutineContext = EmptyCoroutineContext, block: suspend CoroutineScope.() -> T): T {
+public actual fun <T> runBlocking(context: CoroutineContext, block: suspend CoroutineScope.() -> T): T {
     contract {
         callsInPlace(block, InvocationKind.EXACTLY_ONCE)
     }
```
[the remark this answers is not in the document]
qwwdfsad: Right, but then we won't be able to document them differently

kotlinx-coroutines-core/common/test/TestBase.common.kt
```diff
@@ -43,6 +49,16 @@ public expect open class TestBase constructor() {
     ): TestResult
 }
 
+public fun TestBase.runMtTest(
+    expected: ((Throwable) -> Boolean)? = null,
+    unhandled: List<(Throwable) -> Boolean> = emptyList(),
+    block: suspend CoroutineScope.() -> Unit
+): TestResult {
+    @Suppress("CAST_NEVER_SUCCEEDS")
+    if (!multithreadingSupported) return Unit as TestResult
```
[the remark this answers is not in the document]
qwwdfsad: Added

```
    // Never invoked on JS, not extracted to separate 'concurrent' source set
    // as it doesn't help with the suppress anyway and requires an additional ceremony
    // and spreads TestBase extensions over the codebase
```

kotlinx-coroutines-core/native/src/CoroutineContext.kt
qwwdfsad: `delegate` has to expose `Delay` supertype, so it won't work

kotlinx-coroutines-core/common/test/TestBase.common.kt
qwwdfsad: 8 places (potentially there will be more) looks like quite a lot.
Maybe we'll rename it to something more explicit? E.g. `nativeSpeciticSlowTestDivisor` with a comment?

kotlinx-coroutines-core/jvm/test/channels/RendezvousChannelStressTest.kt
```diff
@@ -1,29 +0,0 @@
...
-        val receiver = launch {
-            for (i in 1..n) check(q.receive() == i)
-            expect(3)
-        }
-        expect(1)
-        sender.join()
-        receiver.join()
-        finish(4)
-    }
-}
```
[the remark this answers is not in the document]
qwwdfsad: This test just tests really nothing and is superseded by our lincheck tests

kotlinx-coroutines-core/concurrent/test/RunBlockingTest.kt
```diff
@@ -52,26 +52,26 @@ class RunBlockingTest : TestBase() {
...
         }
         finish(4)
         thread.close()
     }
 
-
     @Test
-    fun testCancellation() = newFixedThreadPoolContext(2, "testCancellation").use {
-        val job = GlobalScope.launch(it) {
-            runBlocking(coroutineContext) {
```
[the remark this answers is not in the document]
qwwdfsad: There isn't any, this change precedes `newFixedThreadPoolContext` in Native. I've reverted it

kotlinx-coroutines-core/nativeDarwin/src/Dispatchers.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val invokeImmediately: Boolean
+) : MainCoroutineDispatcher(), Delay {
+    
+    override val immediate: MainCoroutineDispatcher =
+        if (invokeImmediately) this else DarwinMainDispatcher(true)
+
+    override fun isDispatchNeeded(context: CoroutineContext): Boolean = !(invokeImmediately && isMainThread())
+
+    override fun dispatch(context: CoroutineContext, block: Runnable) {
+        platformAutoreleasePool {
```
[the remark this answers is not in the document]
qwwdfsad: Could you please elaborate on the point of this change? 
`platformAutoreleasePool` is at least supported in IDE

kotlinx-coroutines-core/concurrent/test/ConcurrentExceptionsStressTest.kt
qwwdfsad: It's a leftover, removed

kotlinx-coroutines-core/concurrent/test/sync/SemaphoreStressTest.kt
qwwdfsad: Yes, I'll report it

kotlinx-coroutines-core/concurrent/test/sync/SemaphoreStressTest.kt
qwwdfsad: This is a leftover from times `Dispatchers.Default` was backed by a single worker

kotlinx-coroutines-core/native/src/Dispatchers.kt
qwwdfsad: Right, this is a good catch. I've moved it back to the object

kotlinx-coroutines-core/jvm/src/internal/Concurrent.kt
```diff
@@ -36,3 +36,5 @@ internal fun removeFutureOnCancel(executor: Executor): Boolean {
         return false // failed to setRemoveOnCancelPolicy, assume it does not removes future on cancel
     }
 }
+
+internal actual val multithreadingSupported: Boolean = true
```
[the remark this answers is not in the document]
qwwdfsad: I'd suggest doing so when we are going to evolve the code; it seems a bit premature for now

kotlinx-coroutines-core/jvm/test/ConcurrentTestUtilities.kt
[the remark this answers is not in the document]
dkhalanskyjb: Do you mean that Kotlin gives the guarantee that this object, though private, won't be optimized away on platforms that support runtime reflection that allows detecting its presence? Didn't know that compatibility with Java geos this far.

kotlinx-coroutines-core/native/src/Builders.kt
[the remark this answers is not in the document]
dkhalanskyjb: Do you mean, unlikely to be negligible? If so, could you drop a comment then that states that this is a performance optimization (not a workaround for a bug, not a functionally significant decision, etc.)?

kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -289,7 +289,11 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
             // todo: we should unpark only when this delayed task became first in the queue
             unpark()
         } else {
-            DefaultExecutor.enqueue(task)
+            if (multithreadingSupported) {
```
dkhalanskyjb, line 292: This looks very non-obvious to me: is this the behavior that we want on JS?

kotlinx-coroutines-core/jvm/src/internal/Concurrent.kt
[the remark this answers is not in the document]
dkhalanskyjb: Having searched my feelings for why this bugs me, I think I have a more compelling explanation.

This variable is present in the non-test common code, even though it wasn't used anywhere other than on the Native. This, like not making a variable `private` when it can be, is, I think, an anti-pattern: making the scope in which something is present large is easy, reducing the scope is not.

What happened just now, in this PR, is an example of just that: this variable only concerns the behavior on Native, but, due to it being available everywhere, it is used in the common code. If we later make Native multithreaded by default, we may not notice that the variable outlived its usefulness and can be removed everywhere but in tests (and possibly even there).

> it seems a bit premature for now

I'd argue the opposite: it's premature to have a broad scope of availability before it's actually needed.

kotlinx-coroutines-core/native/src/CoroutineContext.kt
dkhalanskyjb: Sorry, I think I was right initially: `reschedule` is dangerous after all.

kotlinx-coroutines-core/native/src/CoroutineContext.kt
dkhalanskyjb: Whereas just doing `as Delay` where needed is undesirable because of the runtime checks?

kotlinx-coroutines-core/nativeDarwin/src/Dispatchers.kt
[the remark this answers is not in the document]
dkhalanskyjb: It's purely a stylistic preference, not something functional. I was somehow convinced that this is obviously more pleasing and didn't require an explanation, but since it does, feel free to ignore it.

* If we look not just at this line, but around it, we see non-Kotlin-idiomatic code, with names like `dispatch_async`. Using a name like `platformAutoreleasePool` in the block of code where everything else looks like Swift is somewhat staggering, as is any place where constructs from different ecosystems are freely mixed. `Timer` below is nice: it's internally consistent in being "purely Foundations in Kotlin syntax".
* More succinct.

So, less letters, more local stylistic consistency.

kotlinx-coroutines-core/native/src/internal/CopyOnWriteList.kt
```diff
@@ -4,101 +4,76 @@
...
     override fun isEmpty(): Boolean = size == 0
-
     override fun set(index: Int, element: E): E = throw UnsupportedOperationException("Operation is not supported")
+    override fun get(index: Int): E = array[rangeCheck(index)] as E
 
-    override fun get(index: Int): E = array[rangeCheck(index)]!! as E
-
-    private class IteratorImpl<E>(private var array: Array<E>, private val size: Int) : MutableIterator<E> {
-
+    private class IteratorImpl<E>(private val array: Array<E>) : MutableIterator<E> {
```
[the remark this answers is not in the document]
dkhalanskyjb: "Historical note" only refers to the first paragraph. Probably should have split into several comments.

kotlinx-coroutines-core/native/src/Dispatchers.kt
```diff
@@ -4,18 +4,18 @@
 
 package kotlinx.coroutines
 
+import kotlinx.coroutines.internal.multithreadingSupported
 import kotlin.coroutines.*
 
-/** Not inside [Dispatchers], as otherwise mutating this throws an `InvalidMutabilityException`. */
```
dkhalanskyjb, line 9: Is this no longer true for some reason, or do we not want to support `Dispatchers.setMain` on Native without the new memory model?

kotlinx-coroutines-core/concurrent/test/RunBlockingTest.kt
```diff
@@ -52,34 +52,35 @@ class RunBlockingTest : TestBase() {
...
+                    while (true) {
+                        yield()
+                    }
                 }
             }
-        }
 
-        runBlocking {
+            runBlocking {
             job.cancelAndJoin()
```
dkhalanskyjb, line 82:
```suggestion
                job.cancelAndJoin()
```

kotlinx-coroutines-core/common/test/TestBase.common.kt
dkhalanskyjb: Explicitness would help, but I just don't understand what we achieve by introducing this constant (which, I assume, doesn't have inherent meaning and is just a heuristic) in the first place.
```kotlin
val n = 10_000 * stressTestMultiplier / stressTestNativeDivisor
val n = (if (isNative) 1000 else 10_000) * stressTestMultiplier
```
To me, the second version is easier to grasp and is also easier to tweak on a per-test basis.

kotlinx-coroutines-core/common/test/TestBase.common.kt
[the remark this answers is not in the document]
dkhalanskyjb: I see. I do think that this ceremony would be warranted, given that it would at least prevent `runMtTest` autocompletion from popping up anywhere other than inside `concurrent`. On the other hand, we may hope that `runMtTest` will only exist for a short while anyway, so it shouldn't be a huge annoyance.

kotlinx-coroutines-core/concurrent/test/flow/StateFlowCommonStressTest.kt
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow
+
+import kotlinx.coroutines.*
+import kotlin.random.*
+import kotlin.test.*
+
+// A simplified version of StateFlowUpdateStressTest
```
dkhalanskyjb, line 11: I think this goes to `StateFlowUpdateCommonTest`, whereas this is a simplified version of `StateFlowStressTest`.

kotlinx-coroutines-core/common/src/EventLoop.common.kt
dkhalanskyjb, line 292: I guess we don't have enough time for a more structured approach, one that doesn't require this conditional, like ensuring that we can override `enqueue`?

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/concurrent/test/RunBlockingTest.kt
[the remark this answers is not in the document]
dkhalanskyjb: Then it's also possible to remove `testCancellation` from `RunBlockingJvmTest`, right?

kotlinx-coroutines-core/concurrent/test/AtomicCancellationTest.kt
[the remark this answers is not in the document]
dkhalanskyjb: Do you mean, the one below should be `runBlocking`, but the other ones are fine with `runMtTest`?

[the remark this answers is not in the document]
qwwdfsad: All has to be tested with `runBlocking` as neither of tests implies actual multithreading

kotlinx-coroutines-core/native/src/Builders.kt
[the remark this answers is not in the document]
qwwdfsad: No, I've meant exactly that -- it's unlikely that this change affects anything. I've reverted it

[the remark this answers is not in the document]
dkhalanskyjb: Ah, ok.
> I've extracted it during performance investigation

This left the impression that you extracted it because of the results of the investigation, so I got confused. Got it.

kotlinx-coroutines-core/concurrent/test/RunBlockingTest.kt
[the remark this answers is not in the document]
qwwdfsad: Right

kotlinx-coroutines-core/native/src/CoroutineContext.kt
qwwdfsad: In K/N runtime performance is not really something we keep in mind for now, `WorkerDispatcher` approach just looks more straightforward

kotlinx-coroutines-core/common/src/EventLoop.common.kt
qwwdfsad: I'd rather hope for the deprecation of the old memory model and fast cleanup of every code path guarded by `multithreadingSupported`. That way it will be much simpler

kotlinx-coroutines-core/jvm/test/ConcurrentTestUtilities.kt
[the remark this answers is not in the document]
qwwdfsad: This is a JVM source-set, isn't it? I was only talking about JVM guarantees here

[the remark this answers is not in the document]
dkhalanskyjb: It is a JVM source set, but it doesn't matter in this discussion, as dead code elimination could as well happen in compile-time, even before the target platform is known. The only thing potentially preventing this is the language semantics.

kotlinx-coroutines-core/jvm/src/internal/Concurrent.kt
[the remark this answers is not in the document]
qwwdfsad: These are good points!

I've made the following: made `runMtTest` expect function only in `concurrent` source-set, made `multithreadingSupported` native-only declaration and moved the only use of `multithreadingSupported` in `common` to Native source-set

kotlinx-coroutines-core/common/test/TestBase.common.kt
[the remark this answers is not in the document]
qwwdfsad: I've addressed your concern along with `multithreadingSupport` below

kotlinx-coroutines-core/jvm/test/ConcurrentTestUtilities.kt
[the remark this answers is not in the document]
qwwdfsad: This (volatile, MM and, thus, the allowed set of optimizations) part of semantics is not really specified in Kotlin and is left for the target platform to deal with, so the compiler has to be really conservative here (esp. taking into account Java interop => reflection_

dkhalanskyjb (approved):
Great job ensuring that `DefaultExecutor` is safe!

Issue #2791, opened 2021-06-25
`CoroutineScope.future` breaks contract of `cancel`


jachor:
This issue is tricky as `CoroutineScope.future` ktdoc mentions this behavior:

> Note that the error and cancellation semantics of future are subtly different than asListenableFuture’s. 
> In particular, any exception that happens in the coroutine after returned future is successfully cancelled
> will be passed to the CoroutineExceptionHandler from the context. See ListenableFutureCoroutine for details.

I'd like to argue here it is not a good one and actually it means interface of `ListenableFuture` is incorrectly implemented by `CoroutineScope.future` -- `ListenableFuture` contract has `cancel()` method and it doesn't mention it can sometimes crash whole app. Due to that it fails "Liskov substitution principle" -- one can not use this specific impl of `ListenableFuture` instead of other `ListenableFuture` implementations.

To substantiate my observation this trivial code already sporadically will crash due to `future` behavior:

```kotlin
  @Test
  fun cancellation_withFailureReportingRace_crashesCoroutineScope(): Unit = runBlocking(Dispatchers.IO) {
    val innerFuture = SettableFuture.create<Unit>()
    val outerFuture = future { innerFuture.await() }

    launch { innerFuture.setException(IOException()) }
    launch { outerFuture.cancel(false) }

    // NOTE: This will sporadically crash, I'd expect the failure to be only present
    // in 'outterFuture'/'innerFuture'. I think unhandled exception in CoroutineScope
    // should not happen here.
  }
```

I'm aware of https://github.com/Kotlin/kotlinx.coroutines/issues/2774. I decided to create new issue as it is not about cancel/cancel race, although making cancel atomic might fix both. I don't think making cancel atomic will be fully possible due to constructs such as `NonCancellable`, but it would fix _some_ issues.

PR #2840, opened 2021-07-23
Breaking change: Guava future coroutine builder shouldn't report to CoroutineExceptionHandler

### LINKED ISSUE #2774: `CoroutineScope.future {}` behaves differently to `FutureTask` when cancelled
I think we've found another cancellation corner case in `CoroutineScope.future {}`.

In a race between asynchronous cancellation from `Future.cancel()` and cancellation from failure of the executing `Job`, `ListenableFutureCoroutine.onCancelled()` may execute after the `Future` has already been completed as cancelled.

When `onCancelled()` runs after the `Future` is complete, it invokes the `CoroutineExceptionHandler` and escalates the failure:

```
        if (!future.completeExceptionallyOrCancel(cause) && !handled) {
            // prevents loss of exception that was not handled by parent & could not be set to JobListenableFuture
            handleCoroutineException(context, cause)
        }
    }
```

This is different from how `FutureTask` treats failure after cancellation.

I haven't minimized a test case, but my reading is:

1. `Future.cancel()` is called on the returned `Future`
2. The cancelling thread of control executes [`JobListenableFuture.cancel()`](https://github.com/Kotlin/kotlinx.coroutines/blob/65e1b84f742a4fff14c67a11a873a4977a0ba539/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt#L460), succeeds in cancelling the `Future`, then calls `jobToCancel.cancel()`
3. Concurrently, the thread of control executing a continuation in that `Job` completes in failure, which also calls `cancel()` on that same `Job`
4. If the thread of control executing the continuation wins the race, `onCancelled()` gets called with a failure, fails to complete the already-complete-as-cancelled `Future` with that failure, and so invokes the `CoroutineExceptionHandler`

In contrast, `FutureTask` uses CAS from NEW -> COMPLETING when the submitted function finishes for any reason. If that CAS fails, the completion is dropped, whether or not the completion is a failure. Cancellation makes the same CAS away from `NEW`, and there's no `else` case if completion of the running function with an `Exception` doesn't win the race.

Since `future {}` calls the `CoroutineExceptionHandler`, errors that would be ignored in a function run with an `Executor` instead execute a failure handler. So at the moment, under cancellation, futures from `myCoroutineScope.future(::myFunc)` are incompatible with those returned from `myExecutor.submit(::myFunc)`.

I think to maintain compatibility, `future {}` would have to attempt to complete its `Future`, but return from the stack and drop the completion if the future is already complete, for any reason? 

Incidentally, I think the documentation on `ListenableFutureCoroutine` is out of date. That class is only used by `future {}`, but the docstring refers to its use in `asListenableFuture {}`.

vadimsemenov:
Fixes #2774 and #2791.

This change makes `future` coroutine builder consistent with `java.util.concurrent.FutureTask` which also drops exceptions that happen after successful cancellation.

Note: `future` will still pass the error to its parent in the tree if there's one. But if there's no parent, the exception is dropped.

### DIFF
--- a/integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
+++ b/integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
@@ -555,11 +555,7 @@ class ListenableFutureTest : TestBase() {
     }
 
     @Test
-    fun testUnhandledExceptionOnExternalCancellation() = runTest(
-        unhandled = listOf(
-            { it -> it is TestException } // exception is unhandled because there is no parent
-        )
-    ) {
+    fun testUnhandledExceptionOnExternalCancellation() = runTest {
         expect(1)
         // No parent here (NonCancellable), so nowhere to propagate exception
         val result = future(NonCancellable + Dispatchers.Unconfined) {
@@ -567,7 +563,7 @@ class ListenableFutureTest : TestBase() {
                 delay(Long.MAX_VALUE)
             } finally {
                 expect(2)
-                throw TestException() // this exception cannot be handled
+                throw TestException() // this exception cannot be handled and is set to be lost.
             }
         }
         result.cancel(true)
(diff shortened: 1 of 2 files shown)


vadimsemenov:
Gentle ping :)

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
```diff
@@ -775,4 +771,39 @@ class ListenableFutureTest : TestBase() {
             assertEquals(count, completed.get())
         }
     }
+
+    @Test
+    fun stressTestFutureDoesNotReportToCoroutineExceptionHandler() = runTest {
```
qwwdfsad, line 776: [nit] all tests are started with `test` prefix: https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs

qwwdfsad (approved):
Thanks for the detailed documentation and fix! The changes look good to me.

Is it okay to ask you to do the same in `CompletableFuture` builder within this PR?
I suspect it is prone to the same issue.
It's completely okay if you are short on time, your contribution is valuable and appreciated as is.

integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
vadimsemenov: Renamed.

vadimsemenov:
I can look into `CompletableFuture` next week. 
For some reason `testFutureDoesNotReportToCoroutineExceptionHandler` doesn't fail when ported to `CompletableFuture`. 
If you don't mind, I'd send a separate PR for `CompletableFuture`.

qwwdfsad:
Sure, thanks for taking care of it!

PR #2841, opened 2021-07-25
Add linuxArm32Hfp and linuxArm64 build targets

ObsidianX:
Note: depends on kotlinx.atomicfu ARM builds (PR for support here: https://github.com/Kotlin/kotlinx.atomicfu/pull/193 )

### DIFF
--- a/gradle/compile-native-multiplatform.gradle
+++ b/gradle/compile-native-multiplatform.gradle
@@ -14,6 +14,8 @@ kotlin {
 
     targets {
         addTarget(presets.linuxX64)
+        addTarget(presets.linuxArm32Hfp)
+        addTarget(presets.linuxArm64)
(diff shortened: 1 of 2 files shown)


qwwdfsad:
Thanks! We are working internally on guarantees that we are ready to provide for each individual target (as well as the overall understanding which targets are we going to support) and as soon as the public guarantees are published, we'll add each supported target to all our libraries at once

frichez:
Hi @qwwdfsad Please support/build all Kotlin multiplatform available targets. Many projects are using coroutines and are intent to run on embedded devices (arm32 in our case) and can't be build due to this limitation.

hfhbd:
Another use-case: linuxArm64 for arm64 docker images consumed on Linux as well as on macOS (Apple Silicon)

PR #2846, opened 2021-07-27
fix default dispatcher for react native

### LINKED ISSUE #2843: Default JS Dispatcher for React Native is wrong
As part of ecc68f13141d1353aa2bb1d027a077e0fd7be4fb the Default dispatcher for React Native was set to `NodeDispatcher`. However, there is no `process.nextTick` in React Native environments. There is only a fake `process.env` object but nothing else. As a result the Default dispatcher does not work in React Native.

So I tested the other dispatchers. In React Native environments the `SetTimeoutDispatcher` works well. If React Native is executed with a Debugger (Chrome) the JS code is executed in a Chrome worker. In this mode `SetTimeoutDispatcher` and `window.asCoroutineDispatcher()` can be used as dispatchers, `NodeDispatcher` does not work.

To resolve this issue, `createDefaultDispatcher` must be adapted, so it properly detects the fake `process` object without the `nextTick` function.

If needed I can provide a PR and test changes.

Legion2:
Fixes #2843 

I tested the changes with React Native on an Android device and in Debug mode with Chrome

### DIFF
--- a/kotlinx-coroutines-core/js/src/CoroutineContext.kt
+++ b/kotlinx-coroutines-core/js/src/CoroutineContext.kt
@@ -13,12 +13,6 @@ private const val UNDEFINED = "undefined"
 internal external val process: dynamic
 
 internal actual fun createDefaultDispatcher(): CoroutineDispatcher = when {
-    // Check if we are running under ReactNative. We have to use NodeDispatcher under it.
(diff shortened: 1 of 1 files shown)


qwwdfsad:
Also, please target your changes to `develop` branch.

qwwdfsad (approved):
Thanks!

PR #2851, opened 2021-07-30
Optimize mutex implementation

qwwdfsad:
* Get rid of addLastIf and DCSS primitive during contention
    * Leverage constants returned by tryResume* and simplify signatures

### DIFF
--- a/kotlinx-coroutines-core/common/src/sync/Mutex.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Mutex.kt
@@ -10,7 +10,6 @@ import kotlinx.coroutines.internal.*
 import kotlinx.coroutines.intrinsics.*
 import kotlinx.coroutines.selects.*
 import kotlin.contracts.*
-import kotlin.coroutines.*
 import kotlin.jvm.*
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -352,31 +369,37 @@ internal class MutexImpl(locked: Boolean) : Mutex, SelectClause2<Any?, Mutex> {
     private abstract inner class LockWaiter(
         @JvmField val owner: Any?
     ) : LockFreeLinkedListNode(), DisposableHandle {
+        private val isTaken = atomic<Boolean>(false)
```
qwwdfsad, line 372: it's possible to avoid this field and use an atomic wrapper over the owner (then it has to be passed between `tryResumeLockWaiter` and `completeResumeLockWaiter`), but I found it way too harsh

elizarov (approved):
👍 Very good. Indeed, the "add / recheck / invalidate if something changed" trick works fine for locks.

PR #2853, opened 2021-08-02
Dispatchers.Default backed by worker

qwwdfsad:


kotlinx-coroutines-core/native/src/EventLoop.kt
```diff
@@ -18,7 +18,7 @@ internal actual abstract class EventLoopImplPlatform : EventLoop() {
         current.execute(TransferMode.SAFE, {}) {} // send an empty task to unpark the waiting event loop
     }
 
-    // TODO verify loop was shut down is an okay behaviour
+    // TODO actually reschedule
     protected actual fun reschedule(now: Long, delayedTask: EventLoopImplBase.DelayedTask): Unit =
         loopWasShutDown()
```
mvicsokolova, line 23: Should we reschedule a task to DefaultExecutor here, like this:
`DefaultExecutor.invokeOnTimeout(now, delayedTask, EmptyCoroutineContext)`?

PR #2872, opened 2021-08-11
Non-conflating subscription count in SharedFlow and StateFlow

### LINKED ISSUE #2488: StateFlow with SharingStarted.Lazily parameter does not start background coroutine in specific cases
Consider the following test:

```Kotlin
@Test
fun lazyStateFlowTest() = runBlocking(Dispatchers.Default) {
    val scope = CoroutineScope(Dispatchers.Default + SupervisorJob() + CoroutineName("Test"))

    val flow0 = flow {
        var i = 1
        while (true) {
            delay(100)
            emit(i++)
        }
    }.stateIn(scope, SharingStarted.Lazily, initialValue = 1)

    val flow1 = flow0.stateIn(scope, SharingStarted.Lazily, initialValue = 0)

    assertEquals(expected = 0, actual = flow1.first()) // actual is 0
    delay(3000)
    assertNotEquals(illegal = 0, actual = flow1.first()) // actual is 0 but must be non-zero
}
```

The reference for the `SharingStarted.Lazily` says: "Sharing is started when the first subscriber appears and never stops".
When `flow1.first()` statement is executed, based on the reference, it is expected that `flow1` stream starts on the background and never stops.
The test above shows that the actual behavior differs from the expected: second call of `flow1.first()` statement sends `0` value but it shouldn't because `flow0` has already populated `flow1` state with new values.

It happens, perhaps, because `first()` operator gets the default value and cancels subscription from `flow1` stream before it subscribes to `flow0`. In some cases the test above is passed.

Another example that can reproduce race condition:

```kotlin
@Test
fun lazyStateFlowTest2() = runBlocking(Dispatchers.Default) {
    coroutineScope {
        repeat(1000) {
            launch {
                val x = MutableStateFlow(value = 1)
                val y = MutableStateFlow(value = 1)

                val z = x.combine(y) { a, b -> a + b }
                    .stateIn(CoroutineScope(Dispatchers.Default), SharingStarted.Lazily, initialValue = 0)

                assertEquals(expected = 0, actual = z.first())
                delay(2000)
                assertEquals(expected = 2, actual = z.first())
            }
        }
    }
}
```

Kotlin: 1.4.21
Coroutines: 1.4.2
Runtime: JVM

### LINKED ISSUE #2863: Unexpected behavior WhileSubscribed
From the documentation:

> stopTimeout - configures a delay between the disappearance of the last subscriber and the stopping of the sharing coroutine.

However, `stopTimeout` is not refreshed on new subscriptions. I also asked this question [here](https://stackoverflow.com/questions/68699222/kotlin-flow-whilesubscribed).
Another ephemeral example on Android:
```
    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
        val viewBinding = FragmentHomeBinding.bind(view)

        val instantFlow = flow {
            while (true) {
                emit(Instant.now())
                delay(Duration.ofMinutes(1).toMillis())
            }
        }
            .onStart { Log.d(TAG, "start") }
            .onCompletion { Log.d(TAG, "completion") }
            .shareIn(
                lifecycleScope,
                SharingStarted.WhileSubscribed(
                    Duration.ofSeconds(10).toMillis(),
                    Duration.ZERO.toMillis(),
                ),
                1,
            )
        val eventFlow = MutableSharedFlow<Unit>(
            0,
            Int.MAX_VALUE,
            BufferOverflow.SUSPEND,
        )
        viewBinding.button.setOnClickListener {
            check(eventFlow.tryEmit(Unit))
        }
        eventFlow
            .onEach { Log.d(TAG, "$it ${instantFlow.first()}") }
            .launchIn(lifecycleScope)
    }
```
Each click on the button should delay the completion of `instantFlow`, but this does not happen, the flow will end after a time equal to `stopTimeout`, ignoring the subscriptions that occur in `eventFlow`.

```
2021-08-09 17:32:06.835 : start
2021-08-09 17:32:06.839 : kotlin.Unit 2021-08-09T14:32:06.835Z
2021-08-09 17:32:08.411 : kotlin.Unit 2021-08-09T14:32:06.835Z
2021-08-09 17:32:09.256 : kotlin.Unit 2021-08-09T14:32:06.835Z
2021-08-09 17:32:10.978 : kotlin.Unit 2021-08-09T14:32:06.835Z
2021-08-09 17:32:12.119 : kotlin.Unit 2021-08-09T14:32:06.835Z
2021-08-09 17:32:13.964 : kotlin.Unit 2021-08-09T14:32:06.835Z
2021-08-09 17:32:16.862 : completion
```
I have little experience with kotlin coroutines and I may be doing something wrong. Is this the expected behavior?

### LINKED ISSUE #2871: Non-conflating subscription count in MutableStateFlow and MutableSharedFlow
### Problem statement

Currently, the infrastructure of `Flow` sharing is written with complete asynchrony and implementation unification in mind. While it benefits its usages in general, the implementation of sharing leaks `StateFlow` conflation into seemingly orthogonal concepts such as sharing strategies.

It can be demonstrated by the example from #2488:

```
val flow = flow {
    println("Flow is started") // Never printed
    emit(1)
}.stateIn(this, SharingStarted.Lazily, initialValue = 0)

assertEquals(expected = 0, actual = flow.first()) // Passes
yield()
assertNotEquals(illegal = 0, actual = flow.first()) // Fails
```

The root cause is sharing implementation -- any sharing strategy subscribes to the flow `subscriptionCount` and observes the subscriptions. When the collector subscribes and immediately (whether in the same dispatcher or in a different one, but the one that is lagging behind) unsubscribed, the original value is conflated to zero.

A more complex example that involves timings can be found in #2863.

### Proposed solution

During the investigation, it was found that _local_ changes cannot solve the asynchrony problem. While the targeted tweaks can address originally reported issues, the problem will be still unresolved and it will be possible to observe it in less trivial scenarios. It doesn't matter whether the sharing happens in the same thread (e.g. `Dispatchers.Main`) or in a separate one, 
the timings and conflations can still be observed in a way that breaks the mental model of sharing.

Due to compatibility reasons, it's close to impossible to change the type of `subscriptionCount` and it's not the goal to replace trivially-readable (`subscriptionCount.value`) primitive with a brand new type.

Instead, we decided to make `subscriptionCount`  non-conflating -- any subscriber will be able to observe _all_ changes in subscribers number. While it is not aligned with how the rest of the state flows behave, it is the smallest potentially-breaking change that only affects implementors of sharing strategies and introduces small inconsistency, but provides clear mental model to the library users and does not introduce new concepts to the library

qwwdfsad:
Sharing strategies are too sensitive to conflation around extrema and may miss the necessity to start or not to stop the sharing. For more particular examples see #2863 and #2488

Fixes #2488
Fixes #2863
Fixes #2871

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -277,7 +277,7 @@ internal class SharedFlowSlot : AbstractSharedFlowSlot<SharedFlowImpl<*>>() {
     }
 }
 
-internal class SharedFlowImpl<T>(
+internal open class SharedFlowImpl<T>(
     private val replay: Int,
     private val bufferCapacity: Int,
     private val onBufferOverflow: BufferOverflow
@@ -340,7 +340,7 @@ internal class SharedFlowImpl<T>(
      * A tweak for SubscriptionCountStateFlow to get the latest value.
      */
     @Suppress("UNCHECKED_CAST")
-    val lastReplayedLocked: T
+    protected val lastReplayedLocked: T
         get() = buffer!!.getBufferAt(replayIndex + replaySize - 1) as T
 
     @Suppress("UNCHECKED_CAST")
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -198,6 +198,8 @@ public interface MutableSharedFlow<T> : SharedFlow<T>, FlowCollector<T> {
      *     }
      *     .launchIn(scope) // launch it
      * ```
+     *
+     * Implementation note: the resulting flow **does not** conflate subscription count.
      */
     public val subscriptionCount: StateFlow<Int>
 
@@ -253,7 +255,7 @@ public fun <T> MutableSharedFlow(
 
 // ------------------------------------ Implementation ------------------------------------
 
-private class SharedFlowSlot : AbstractSharedFlowSlot<SharedFlowImpl<*>>() {
+internal class SharedFlowSlot : AbstractSharedFlowSlot<SharedFlowImpl<*>>() {
     @JvmField
     var index = -1L // current "to-be-emitted" index, -1 means the slot is free now
 
@@ -275,7 +277,7 @@ private class SharedFlowSlot : AbstractSharedFlowSlot<SharedFlowImpl<*>>() {
     }
 }
 
-private class SharedFlowImpl<T>(
+internal class SharedFlowImpl<T>(
     private val replay: Int,
     private val bufferCapacity: Int,
     private val onBufferOverflow: BufferOverflow
@@ -333,6 +335,13 @@ private class SharedFlowImpl<T>(
             for (i in 0 until replaySize) result += buffer.getBufferAt(replayIndex + i) as T
             result
         }
+
+    /*
+     * A tweak for SubscriptionCountStateFlow to get the latest value.
+     */
+    @Suppress("UNCHECKED_CAST")
+    val lastReplayedLocked: T
+        get() = buffer!!.getBufferAt(replayIndex + replaySize - 1) as T
 
     @Suppress("UNCHECKED_CAST")
     override suspend fun collect(collector: FlowCollector<T>) {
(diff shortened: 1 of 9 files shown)


elizarov (approved):
Great job! It seems to me that using delegation here for subscription count impl is a bit wasteful, so I've added a commit with inheritance-based implementation. Please, check it out.

qwwdfsad:
Looks good to me. I'll merge it shortly after Kotlin 1.5.30

PR #2881, opened 2021-08-13
Fix BlockHound false-positive in ConflatedChannel

### LINKED ISSUE #2866: BlockHound false positives in ConflatedChannel
I have seen BlockHound complaining on a blocking call on `jdk.internal.misc.Unsafe#park` apparently caused by:
```text
kotlinx.coroutines.channels.ConflatedChannel.enqueueReceiveInternal
```

Seems like a false positive. Unfortunately, I have not been able to reproduce it.

Probably it is worth adding `"isEmpty"` and `"enqueueReceiveInternal"` to the list in `allowBlockingCallsInConflatedChannel()`, as both of them use the same lock as the other members, but are not covered yet:

https://github.com/Kotlin/kotlinx.coroutines/blob/7e762d324aeeb34b53390dee931569013809660e/kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt#L135-L136

qwwdfsad:
Fixes #2866


dkhalanskyjb (approved):
LGTM. I filed https://github.com/reactor/BlockHound/issues/221 so we can avoid the whole class of these mistakes.

PR #2889, opened 2021-08-20
Update DebugProbesKt.bin exclusion instruction

ebraminio:
raw exclude call is now deprecated and gives the following warning,
```
  This property is deprecated. Use resources.excludes or jniLibs.excludes instead. Use
  jniLibs.excludes for .so file patterns, and use resources.excludes for all
  other file patterns.
```

Tested and works as expected here https://github.com/persian-calendar/DroidPersianCalendar/commit/b4c21e0d1eb4ea8f0a93e56de2f56793fa442608


qwwdfsad (approved):
Thanks!

PR #2892, opened 2021-08-23
Update CoroutineScope() factory documentation

### LINKED ISSUE #2891: Clarification on CoroutineScope doc
So I've noticed that the [doc](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/CoroutineScope.kt#L272) says
> If the given [context] does not contain a [Job] element, then a default `Job()` is created. 
This way, cancellation or failure of any child coroutine in this scope cancels all the other children, 
just like inside [coroutineScope] block.

The key confusion here is that _**cancellation** of any child coroutine in this scope cancels all the other children_. From my testing, cancellation of a child of a Job() doesn't really cancels all other children. Maybe I'm missing anything but the testing code is:
```
val scope = CoroutineScope(Dispatchers.Default)

val job1 = scope.launch {
    delay(99999)
}
val job2 = scope.launch {
    delay(500)
}

job2.cancel()

println("job1 state: ${job1.isCancelled}") // false
println("job2 state: ${job2.isCancelled}") // true
```

qwwdfsad:
Fixes #2891

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineScope.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineScope.kt
@@ -12,7 +12,7 @@ import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
 
 /**
- * Defines a scope for new coroutines. Every **coroutine builder** (like [launch], [async], etc)
+ * Defines a scope for new coroutines. Every **coroutine builder** (like [launch], [async], etc.)
  * is an extension on [CoroutineScope] and inherits its [coroutineContext][CoroutineScope.coroutineContext]
  * to automatically propagate all its elements and cancellation.
  *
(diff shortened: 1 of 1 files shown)


dkhalanskyjb (commented):
Sure.
> just like inside [coroutineScope] block.

Maybe it's also worth it to explicitly mention in the `coroutineScope` documentation that cancelling the scope also cancels the children? The only mention of this behaviour that I see is fairly indirect:
>  * 3) If the outer scope of `showSomeData` is cancelled, both started `async` and `withContext` blocks are cancelled.

qwwdfsad:
Honestly, I do not see how it can be fit organically. 

We have quite extensive documentation in `CoroutineScope` and `Job`, while the rest of the builders (whether it is `async` or `withContext`) point to it. Not sure it's worth repeating in every declaration

dkhalanskyjb (approved):

PR #2895, opened 2021-08-25
Fix BlockHound false positive in stack trace recovery

### LINKED ISSUE #2894: BlockHound false positive in tryCopyException
org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.1

Relevant stack trace excerpt:
```
Caused by: reactor.blockhound.BlockingOperationError: Blocking call! jdk.internal.misc.Unsafe#park
[...]
at kotlinx.coroutines.internal.ExceptionsConstuctorKt.tryCopyException(ExceptionsConstuctor.kt:47)
```

qwwdfsad:
Fixes #2894


qwwdfsad:
I feel like it doesn't worth adding a stress test for that. Tested manually by adding `Thread.sleep` to line 47 in `ExceptionConstructor`

dkhalanskyjb (approved):
:+1:

Issue #2237, opened 2020-09-10
Support the Java Module System


TheMrMilchmann:
I'm quoting directly from https://github.com/Kotlin/kotlinx.serialization/issues/940 because the same reasoning applies:

> **What is your use-case and why do you need this feature?**
> As of Kotlin 1.4-M3, Java 9 module descriptors have been added to the standard library artifacts ([KT-21266](https://youtrack.jetbrains.com/issue/KT-21266)). It would be nice to see this also happen to libraries in the Kotlin ecosystem going forward as this would enable Kotlin JVM applications to benefit from modern JDK tools (such as jlink).
> 
> **Describe the solution you'd like**
> Ideally, an explicit descriptor would be added using the Multi-Release JAR mechanism ([JEP 238](https://openjdk.java.net/jeps/238)) added in JDK 9 (just as it was done for the standard library).

Additionally,
1. I'm currently working on a desktop application for which using the module system (with `jlink` and `jpackage`) is extremely handy and I would love to use `kotlinx.coroutines` without additional non-trivial build-system hacks.
2. Publishing a library with an explicit module descriptor and with a dependency on `kotlinx.coroutines` is not possible because the latter does not have a stable module name.

A temporary (and fairly simple) solution for `2.` would be providing stable [`Automatic-Module-Name`](https://dzone.com/articles/automatic-module-name-calling-all-java-library-maintainers) entries.

MakHoCheung:
```
requires kotlinx.coroutines.core.jvm;
```
```
/Projects/java-demo/src/main/java/module-info.java:3: 错误: 找不到模块: kotlinx.coroutines.core.jvm
    requires kotlinx.coroutines.core.jvm;
```
I use kotlinx coroutines with jpms in my project. But I got error "cannot find module: kotlinx.coroutines.core.jvm". Does kotlinx coroutines not support jpms even using automatic module ?

jobobby04:
As a stop gap, for now add
`
addExtraDependencies("kotlinx.coroutines")
`
to your JLink gradle settings, it allows it to find coroutines

bjonnh:
This stopgap doesn't work for me. First there is no "kotlinx.coroutines" I can add to module-info. I can get IntelliJ to accept kotlinx.coroutines.core.jvm  … But trying to add any with addExtraDependencies doesn't help, it still says:
````
src/main/kotlin/module-info.java:6: error: module not found: kotlinx.coroutines.core.jvm
[...]

PR #2901, opened 2021-08-26
Support Java Platform Module System (JPMS)

santoszv:
Support the Java Module System through Automatic-Module-Name manifest entry (#2237).


TheMrMilchmann:
Note that automatic module names should probably be added to all JVM artifacts instead of just a select few. While an explicit descriptor would require all dependencies to be modularized, this is not the case for `Automatic-Module-Name` entries.

qwwdfsad:
Superseded by #3297

PR #2903, opened 2021-08-26
Introduce CloseableCoroutineDispatcher abstract class

qwwdfsad:
* Alias it to ExecutorCoroutineDispatcher on JVM as the most appropriate candidate
    * Add abstract classes to JVM and Native
    * This class will be implemented by new dispathers for K/N new memory model

### DIFF
--- a/kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+/**
+ * [CoroutineDispatcher] that provides method to close it,
+ * causing the rejection of any new tasks and cleanup of all underlying resources
+ * associated with the current dispatcher.
+ * Examples of closeable dispatchers are dispatchers backed by `java.lang.Executor` and
+ * by `kotlin.native.Worker`.
+ *
+ * **The `CloseableCoroutineDispatcher` class is not stable for inheritance in 3rd party libraries**, as new methods
+ * might be added to this interface in the future, but is stable for use.
+ */
+public expect abstract class CloseableCoroutineDispatcher() : CoroutineDispatcher {
+
+    /**
+     * Initiate the closing sequence of the coroutine dispatcher.
+     * After the successful call to [close], no new tasks will
+     * be accepted to be [dispatched][dispatch], but previously dispatched tasks will be run.
+     *
+     * Invocations of `close` on already closed dispatcher have no effect.
+     */
+    public abstract fun close()
+}
--- a/kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
@@ -21,7 +21,7 @@ public expect abstract class CloseableCoroutineDispatcher() : CoroutineDispatche
      * After the successful call to [close], no new tasks will
      * be accepted to be [dispatched][dispatch], but previously dispatched tasks will be run.
      *
-     * Invocations of `close` on already closed dispatcher have no effect.
+     * Invocations of `close` are idempotent and thread-safe.
      */
     public abstract fun close()
 }
--- a/kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
@@ -14,6 +14,7 @@ package kotlinx.coroutines
  * **The `CloseableCoroutineDispatcher` class is not stable for inheritance in 3rd party libraries**, as new methods
  * might be added to this interface in the future, but is stable for use.
  */
+@ExperimentalCoroutinesApi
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+ * by `kotlin.native.Worker`.
+ *
+ * **The `CloseableCoroutineDispatcher` class is not stable for inheritance in 3rd party libraries**, as new methods
+ * might be added to this interface in the future, but is stable for use.
+ */
+public expect abstract class CloseableCoroutineDispatcher() : CoroutineDispatcher {
+
+    /**
+     * Initiate the closing sequence of the coroutine dispatcher.
+     * After the successful call to [close], no new tasks will
```
dkhalanskyjb, line 21:
```suggestion
     * After a successful call to [close], no new tasks will
```
To my ear, "the" implies that the call will always happen eventually.

kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+     * be accepted to be [dispatched][dispatch], but previously dispatched tasks will be run.
+     *
+     * Invocations of `close` on already closed dispatcher have no effect.
```
dkhalanskyjb, line 24:
```suggestion
     * Invocations of `close` on an already closed dispatcher have no effect.
```

dkhalanskyjb: There's ambiguity: it is stated above that `close` *initiates* the closing sequence, but doesn't necessarily *close* the dispatcher. I assume the two are considered equivalent here? If so, two ways to rephrase this that I see:
* Talk here not about closed dispatchers, but only about those "that already entered the closing sequence". Cumbersome, in my opinion.
* Clarify via a more formal additional sentence that calls to `close` are *idempotent*.

kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * **The `CloseableCoroutineDispatcher` class is not stable for inheritance in 3rd party libraries**, as new methods
+ * might be added to this interface in the future, but is stable for use.
+ */
+public expect abstract class CloseableCoroutineDispatcher() : CoroutineDispatcher {
+
+    /**
+     * Initiate the closing sequence of the coroutine dispatcher.
+     * After the successful call to [close], no new tasks will
+     * be accepted to be [dispatched][dispatch], but previously dispatched tasks will be run.
```
dkhalanskyjb, line 22:
```suggestion
     * be accepted to be [dispatched][dispatch], but the previously dispatched tasks will be run.
```

kotlinx-coroutines-core/common/src/CloseableCoroutineDispatcher.kt
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines
+
+/**
+ * [CoroutineDispatcher] that provides method to close it,
```
dkhalanskyjb, line 8:
```suggestion
 * [CoroutineDispatcher] that provides a method to close it,
```

dkhalanskyjb (changes requested):
(Take my words with a grain of salt, as I wasn't on the design meeting for this and it's possible that these questions were already covered)
The main use case for this is https://github.com/Kotlin/kotlinx.coroutines/issues/2558, right?

As implemented here, `close()` is like asking the dispatcher "hey, unload when you're done, okay?" and then just leaving. Is this really the behavior that we want? My assumption is as follows: `close()` will always be called *after* the users think they're completely finished working with coroutines. If so, it is unexpected that some jobs are still running or pending at the moment of call to `close()`. Maybe we could do `awaitTermination` here. Maybe we could also call `shutdownNow` and loudly complain if there were some pending tasks. In any case, I think that someone who calls `close()` on coroutine dispatchers will not be satisfied with just a "best-effort" solution to stop the dispatchers and would want to be notified in some way if there is some rogue job that prevents the code from being unloaded.

qwwdfsad:
>The main use case for this is #2558, right?

It is not, for that we have another solution, #2915

The main use-case is #2914: we want to introduce `newFixedThreadPoolContext()` and `newSingleThreadContext` to `concurrent` source set (shared between Native and JVM). They represent closeable resources on both platforms, thus we need an abstraction over the dispatcher that represents closeable dispatcher and, well, can be closed.

Apart from that, we already have `ExecutorCoroutineDispatcher.close` (`CloseableCoroutineDispatcher` is an alias to that) since the 1.0.0 release and it seems to be accepted quite well, I'm not sure we want to break things here.
`awaitTermination` cannot be implemented on top of the `Executor` (only on top of `ExecutorService`) and it also would be nice to avoid one more public hierarchy for "dispatchers that can only be closed" and "dispatchers that can also be awaited".

dkhalanskyjb (approved):
Ah, ok. All my points are invalid if unloading the dispatchers is not the point of this.

qwwdfsad:
PTAL one more time just to double-check experimentality

dkhalanskyjb (approved):
Makes sense.

PR #2912, opened 2021-09-01
Move `isInstalledStatically` field to a separate object

PierfrancescoSoffritti:
By doing this DebugProbesImpl can check for static installation without having to depend on AgentPremain, which is not compatible with Android.


qwwdfsad (approved):
Thanks for taking care of it!

Superseded by #2913

PR #2915, opened 2021-09-03
Introduce Dispatchers.shutdown

### LINKED ISSUE #2558: Provide an API to close Dispatchers.Default and Dispatchers.IO

qwwdfsad:
Fixes #2558


dkhalanskyjb (changes requested):
The comments I made here https://github.com/Kotlin/kotlinx.coroutines/pull/2903#pullrequestreview-746080303 are somewhat applicable here as well: I think that we need to establish and document some specific behavior for cases where there are some pending jobs at the moment of shutting down the dispatchers.

As is, I managed to get a highly confusing and uninformative error with this code:
```kotlin
        val j = GlobalScope.launch {
            delay(1000)
            println("Hey-hey")
        }
        GlobalScope.launch {
            Dispatchers.shutdown()
            j.join()
        }
```
The error:
```
Exception in thread Thread[DefaultDispatcher-worker-1 @coroutine#1,5,main]: java.lang.AssertionError
java.lang.AssertionError
	at kotlinx.coroutines.EventLoopImplPlatform.reschedule(EventLoop.kt:17)
	at kotlinx.coroutines.EventLoopImplBase.schedule(EventLoop.common.kt:362)
	at kotlinx.coroutines.EventLoopImplBase.scheduleResumeAfterDelay(EventLoop.common.kt:235)
	at kotlinx.coroutines.DelayKt.delay(Delay.kt:121)
	at kotlinx.coroutines.RunBlockingTest$dispatcherShutdownTest$j$1.invokeSuspend(RunBlockingTest.kt:15)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

qwwdfsad:
I've improved the exception message and documentation.

The API is marked as delicate and has a pretty narrow use case, so I'm not sure it's worth (both in terms of contract and maintainability) giving more guarantees than now.

kotlinx-coroutines-core/jvm/src/Dispatchers.kt
```diff
@@ -116,4 +116,29 @@ public actual object Dispatchers {
...
+     * stopping all the threads associated with them and making them reject all new tasks.
+     * Dispatcher used as a fallback for time-related operations (`delay`, `withTimeout`)
+     * and to handle rejected tasks from other dispatchers is also shut down.
+     *
+     * This is a **delicate** API. It is not supposed to be called from a general
+     * application-level code and its invocation is irreversible.
+     * The invocation of shutdown affects most of the coroutines machinery and
+     * leaves the coroutines framework in an inoperable state.
+     * An invocation of shutdown method while having pending tasks or active coroutines
+     * will leave them in a permanent dormant state, preventing them from completion or execution.
```
dkhalanskyjb, line 131: I agree with your reasoning: maybe we shouldn't give some specific guarantees or thoroughly discuss the API for less than 0.001% of our users.

That said, I think that the line of documentation that I'm commenting on is still misleading. It implies strongly that `shutdown` will always succeed, and the tasks will just be silently dropped. The example above still crashes, though with a better message.

Maybe we should document the lack of guarantees then? Something like
```suggestion
     * The shutdown method should only be invoked when there are no pending tasks or active coroutines.
     * Otherwise, the behavior is unspecified: the call to `shutdown` may throw an exception without completing
     * the shutdown, or it may finish successfully, but the remaining jobs will be in a permanent dormant state,
     * never completing nor executing.
```
I like this version more because it clarifies two important points:
* One can't rely on `shutdown` finishing successfully if there are active jobs;
* One can't rely on `shutdown` throwing an exception to indicate the presence of active jobs (see the example below):
```kotlin
        val j = GlobalScope.launch {
            println("Before the delay")
            Thread.sleep(1000)
            println("This was reached")
        }
        println("Shutdown pending")
        Dispatchers.shutdown()
        println("Shutdown successful")
        runBlocking {
            j.join()
        }
```
This successfully performs the rescheduling :
```
Shutdown pending
Before the delay
This was reached
Shutdown successful
```
Replacing `Thread.sleep` with `delay` makes this crash after printing "Before the delay". So, the exception-throwing behavior is inconsistent.

kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
```diff
@@ -37,15 +37,39 @@ internal actual object DefaultExecutor : EventLoopImplBase(), Runnable {
...
+
+     override fun reschedule(now: Long, delayedTask: DelayedTask) {
+         // Reschedule on default executor can only be invoked after Dispatchers.shutdown
+         shutdownError()
+    }
+
+    private fun shutdownError() {
+        throw RejectedExecutionException("DefaultExecutor was shut down. " +
+            "This error indicates that Dispatchers.shutdown() was invoked prior to completion of exiting coroutines, leaving coroutines in incomplete state. " +
+            "Please refer to Dispatchers.shutdown documentation for more details")
```
dkhalanskyjb, line 65: > more details

The `Dispatchers.shutdown` documentation doesn't currently state that this crash may happen, I think? So, even after reading the documentation, the crash may still be confusing.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
qwwdfsad: I've committed the proposed documentation improvement, should be better now

dkhalanskyjb (approved):

PR #2916, opened 2021-09-03
Remove obsolete CommonPool support

qwwdfsad:
* We state that it's deprecated since the initial release of CoroutineScheduler
    * It is an additional DEX load for Android


kotlinx-coroutines-core/native/src/CoroutineContext.kt
```diff
@@ -25,7 +25,7 @@ internal actual object DefaultExecutor : CoroutineDispatcher(), Delay {
 
 internal fun loopWasShutDown(): Nothing = error("Cannot execute task because event loop was shut down")
 
-internal actual fun createDefaultDispatcher(): CoroutineDispatcher =
+internal fun createDefaultDispatcher(): CoroutineDispatcher =
```
dkhalanskyjb, line 28: Can't this be inlined, like on the JVM?

dkhalanskyjb (approved):

kotlinx-coroutines-core/native/src/CoroutineContext.kt
qwwdfsad: It will complicate the maintenance of `native-mt` branch during rebases

PR #2917, opened 2021-09-04
Fix the doc of 'Coroutine Exceptions Handling'

MasoodFallahpoor:
The reason to remove 'later' is that section "Channels" comes before section "Coroutine Exceptions Handling" not after it.


dkhalanskyjb:
Thanks!

PR #2927, opened 2021-09-12
Make 100k coroutine sample runnable

mhansen:
This should fix a few problems with this example:
1. The `//sampleStart` comments were showing up, when I think they were supposed to not
2. This code section is the only one that's not runnable. If this is cheap, feels like this should be runnable in browser?

image

### DIFF
--- a/docs/topics/coroutines-basics.md
+++ b/docs/topics/coroutines-basics.md
@@ -263,6 +263,7 @@ fun main() = runBlocking {
 }
 //sampleEnd
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 > You can get the full code [here](../../kotlinx-coroutines-core/jvm/test/guide/example-basic-06.kt).
 >


mhansen:
It's possible that this sample is too expensive to be runnable. I notice that this example https://kotlinlang.org/docs/cancellation-and-timeouts.html#asynchronous-timeout-and-resources always returns `Evaluation stopped while it's taking too long️` if you try to hit the 'play' button.

qwwdfsad (changes requested):
Thanks for pointing it out! `//sampleStart` and `//sampleEnd` comments are indeed can be safely removed.

>If so, perhaps a smaller-scale example would be better to illustrate the point that coroutines are cheap?

I would like to keep the things they are. The sample is not runnable because Kotlin playground is *really* tight on resources (to avoid DDoS, provide decent service for thousands of users, and so on), and the guide implies that the reader can copy the code and run it.

It's okay to add an explicit comment about it though

mhansen:
No worries. I’ll revert this.

PR #2928, opened 2021-09-12
Fix chapter pointer: it's the second next chapter

mhansen:
This probably used to be the next chapter, but not any more.


dkhalanskyjb:
Thank you!

PR #2935, opened 2021-09-14
Fix -Poverwrite.probes not being propagated to the test

dkhalanskyjb:
This affected `PrecompiledDebugProbesTest#testClassFileContent`.


qwwdfsad (approved):
I've always redefined the constant in tests as the fastest way :)

PR #2950, opened 2021-09-24
Add version file to each module resources

qwwdfsad:


build.gradle
```diff
@@ -245,6 +245,40 @@ configure(subprojects.findAll { !unpublished.contains(it.name) }) {
             }
         }
     }
+
+    def project = it
```
dkhalanskyjb, line 249: `project` → `subproject`, or `subproj` if the length is an issue? Encountering "`project`" in the middle of a gradle script could be confusing.

qwwdfsad: Renamed to `thisProject`. I avoid `project` semi-automatically to avoid capturing outer context into the closure where I do not expect it. Alas IDE is not helpful here

qwwdfsad:
Also changed the format of the file

dkhalanskyjb (approved):

PR #2952, opened 2021-09-24
Introduce coroutines platform to have all the dependencies aligned

### LINKED ISSUE #2865: Binary incompatibitlity between different kotlinx-coroutines modules
Given a following set of dependencies:
```
dependencies {
    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.1"
    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.4.3"
}
```

This simple function fails with `java.lang.NoSuchMethodError: 'void kotlinx.coroutines.AbstractCoroutine.<init>(kotlin.coroutines.CoroutineContext, boolean, int, kotlin.jvm.internal.DefaultConstructorMarker)'`:
````
suspend fun main(): Unit = coroutineScope {
    future {  }
}
````

It is caused by binary incompatibitlity of `AbstractCoroutine` from `kotlinx-coroutines-core` module. Despite it is an "internal" API, it is used in another module, whose version might not be the same.

Dependency graph:
<pre>
runtimeClasspath - Runtime classpath of compilation 'main' (target  (jvm)).
+--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.20
|    +--- org.jetbrains.kotlin:kotlin-stdlib:1.5.20
|    |    +--- org.jetbrains:annotations:13.0
|    |    \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.5.20
|    \--- org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.5.20
|         \--- org.jetbrains.kotlin:kotlin-stdlib:1.5.20 (*)
+--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.1
|    \--- org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.5.1
|         +--- org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.5.20 (*)
|         \--- org.jetbrains.kotlin:kotlin-stdlib-common:1.5.20
\--- <b>org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.4.3</b>
     +--- <b>org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.3 -> 1.5.1</b> (*)
     \--- org.jetbrains.kotlin:kotlin-stdlib:1.4.30 -> 1.5.20 (*)
</pre>

To prevent such situation, both modules should depend on `kotlinx-coroutines-bom` of the corresponding version, as described in [Gradle documentation](https://docs.gradle.org/current/userguide/dependency_version_alignment.html#aligning_versions_natively_with_gradle).
This way, `kotlinx-coroutines-core:1.5.1` would upgrade `kotlinx-corotuines-jdk8:1.4.3` to the same version, resolving the issue.

qwwdfsad:
Fixes #2865

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,11 +2,10 @@
  * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
+
+import org.jetbrains.dokka.gradle.DokkaTaskPartial
 import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
 import org.jetbrains.kotlin.konan.target.HostManager
-import org.gradle.util.VersionNumber
-import org.jetbrains.dokka.gradle.DokkaTaskPartial
-import org.jetbrains.dokka.gradle.DokkaMultiModuleTask
 
 apply plugin: 'jdk-convention'
 apply from: rootProject.file("gradle/opt-in.gradle")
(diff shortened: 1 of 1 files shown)


qwwdfsad:
Unfortunately, it cannot be built (=> merged) prior to 1.6.0: unless https://github.com/jetbrains/kotlin/commit/0cd29adcc7f0cf3fa6f394d7b3742c064e0ed366 is merged, `kotlin-compiler-embeddable` depends on `kotlinx-coroutines-core`, and after dependency alignment, in order to build
`kotlinx-coroutines-core` of our `x-SNAPSHOT` version we have to first resolve `kotlin-compiler-embeddable` dependency to the very same version, making a dependency cycle.

It's still possible to test it locally though: first one has to do `publishToMavenLocal` the original project, then add `mavenLocal` to the list of repositories, apply the proposed patch and then `publishToMavenLocal`.
Then I've verified the problem in #2865 is gone

qwwdfsad:
Verified it works on `1.6.0-RC-512`

PR #2954, opened 2021-09-27
Leverage definitely non-nullable types

qwwdfsad:


qwwdfsad:
Postponed due to IDE issues

PR #2956, opened 2021-09-28
currentTimeMillis is and internal inline

jeffreyg-sh:
The docs show the the correct implementation


qwwdfsad (commented):
Thanks!

This is intentional as we do track time in these tests. `System.currentTimeMillis` is not suitable for that

PR #2964, opened 2021-10-01
Non-propagating cancellation of @FlowPreview operators

qwwdfsad:


qwwdfsad:
I've updated the doc. 

`flatMapMerge` and `flattenMerge` indeed now have an observable difference in cancellation behaviour for concurrency equal `1` and `> 1`. This is a deliberate design decision as the fact that `concurrency == 1` is optimized into sequential transformation. Sequential -> no child coroutines -> in place invocations -> in-place cancellation is propagated

PR #2965, opened 2021-10-01
coroutines-test as an MPP

dkhalanskyjb:
This PR shouldn't change any behavior and instead only ensures that the test module can be built for all platforms. As is, it already requires introducing incompatibility, but, luckily, mostly in the interfaces that we decided to deprecate anyway.

`Dispatchers.setMain` isn't implemented for any targets other than JVM, all the other functionality seems to have been ported fine.


kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
```diff
@@ -56,12 +56,11 @@ public final class kotlinx/coroutines/test/TestDispatchers {
...
 
 public abstract interface class kotlinx/coroutines/test/UncaughtExceptionCaptor {
-	public abstract fun cleanupTestCoroutines ()V
+	public abstract fun cleanupTestCoroutinesCaptor ()V
 	public abstract fun getUncaughtExceptions ()Ljava/util/List;
 }
 
 public final class kotlinx/coroutines/test/UncompletedCoroutinesError : java/lang/AssertionError {
-	public fun <init> (Ljava/lang/String;Ljava/lang/Throwable;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/Throwable;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
dkhalanskyjb, line 65: `AssertionError` in common code doesn't have a corresponding constructor.

kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
```diff
@@ -56,12 +56,11 @@ public final class kotlinx/coroutines/test/TestDispatchers {
 }
 
 public abstract interface class kotlinx/coroutines/test/UncaughtExceptionCaptor {
-	public abstract fun cleanupTestCoroutines ()V
+	public abstract fun cleanupTestCoroutinesCaptor ()V
```
dkhalanskyjb, line 59: Without this rename, JS IR compiler just fails, and the native compiler complains
```
Member inherits different @Throws filters from UncaughtExceptionCaptor, DelayController
```

kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
qwwdfsad: Sure. 

If it was the stable API, I would suggest providing constructor that invokes `initCause`, but here it is not necessary

kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
qwwdfsad: Worth reporting, because JS IR should ignore `@Throws`

kotlinx-coroutines-test/build.gradle.kts
```diff
@@ -2,6 +2,24 @@
...
+        val srcDir = if (name.endsWith("Main")) "src" else "test"
+        val platform = name.dropLast(4)
+        kotlin.srcDir("$platform/$srcDir")
+        if (name == "jvmMain") {
+            resources.srcDir("$platform/resources")
+        } else if (name == "jvmTest") {
+            resources.srcDir("$platform/test-resources")
+        }
+        languageSettings {
+            experimentalAnnotations.forEach { optIn(it) }
```
qwwdfsad, line 22: `progressiveMode` is missing

kotlinx-coroutines-test/build.gradle.kts
```diff
@@ -2,6 +2,24 @@
...
 
-dependencies {
-    implementation(project(":kotlinx-coroutines-debug"))
+val experimentalAnnotations = listOf(
+    "kotlin.Experimental",
+    "kotlinx.coroutines.ExperimentalCoroutinesApi",
+    "kotlinx.coroutines.InternalCoroutinesApi"
+)
+
+kotlin {
```
qwwdfsad, line 11: I guess it's time to actually maintain (instead of duct-taping) our build scripts.
This part is the copy-paste from core-module, please introduce .kts with common utility

qwwdfsad (approved):
LGTM

PR #2966, opened 2021-10-01
Removed experimental annotation from old APIs

prithivraj:
Removes the need to add the ExperimentalCoroutinesApi annotation when testing coroutine based code.

### DIFF
--- a/kotlinx-coroutines-test/src/DelayController.kt
+++ b/kotlinx-coroutines-test/src/DelayController.kt
@@ -12,14 +12,12 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
  *
  * Testing libraries may expose this interface to tests instead of [TestCoroutineDispatcher].
  */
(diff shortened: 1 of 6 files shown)


dkhalanskyjb:
Sorry, won't accept. `ExperimentalCoroutinesApi` is there for a reason: we're not satisfied with this API and are working on changing it. As is, the test module is, just as the annotation says, experimental, and its users should be prepared to face some breaking changes at some point.

prithivraj:
Great to know that there is a plan to improve these APIs. I had removed it because the last commits made to most of the files are from 3 years ago and the tentative dates on the annotation was minor 2 releases ago. 

The Android community is pushing for using StateFlow as the recommended way of building apps. That means every Android app that needs Unit tests has to invariably run into using these APIs that are currently marked as experimental. Thanks again for your reply, I look forward to these great APIs hitting stable soon!

PR #2974, opened 2021-10-11
Use Dispatchers.Main as default delay source where applicable

### LINKED ISSUE #2972: Default `delay` behavior on Android creates additional threads when using external dispatchers
When `delay` is called using an externally-defined dispatcher that does not implement the internal `Delay` interface, it uses a default fallback implementation that can cause an additional thread to be created. Some Android apps strictly audit threads created by their library dependencies and have raised this as an issue when using Jetpack Compose, which defines its own UI dispatcher and both promotes using the `delay` API and uses it internally.

On Android, apps always have a main thread running an event loop that could be used in place of creating an additional thread to time the `delay`, which could make `Dispatchers.Main` a suitable choice for performing the default delay on Android when a compatible/known dispatcher is not found.

qwwdfsad:
It reduces the number of redundant threads in the system and makes time source predictable across Android/JavaFx applications

Fixes #2972

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
+++ b/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
@@ -4,10 +4,25 @@
 
 package kotlinx.coroutines
 
+import kotlinx.coroutines.internal.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
 
-internal actual val DefaultDelay: Delay = DefaultExecutor
+internal actual val DefaultDelay: Delay = initializeDefaultDelay()
+
+private val defaultMainDelayOptIn = systemProp("kotlinx.coroutines.main.delay", true)
+
+private fun initializeDefaultDelay(): Delay {
+    // Opt-out flag
+    if (!defaultMainDelayOptIn) return DefaultExecutor
+    val main = Dispatchers.Main
+    /*
+     * When we already are working with UI and Main threads, it makes
+     * no sense to create a separate thread with timer that cannot be controller
+     * by the UI runtime.
+     */
+    return if (main.isMissing() || main !is Delay) DefaultExecutor else main
+}
 
 @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
 internal actual object DefaultExecutor : EventLoopImplBase(), Runnable {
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
```diff
@@ -4,10 +4,25 @@
...
 package kotlinx.coroutines
 
+import kotlinx.coroutines.internal.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
 
-internal actual val DefaultDelay: Delay = DefaultExecutor
+internal actual val DefaultDelay: Delay = initializeDefaultDelay()
+
+private val defaultMainDelayOptIn = systemProp("kotlinx.coroutines.main.delay", true)
```
dkhalanskyjb, line 13: I think this should be documented at least somewhere.

qwwdfsad: We can mention it in the original issue and in the changelog

qwwdfsad: Also, we can document `Delay` itself, but it doesn't make it to the Dokka and is internal interface, so it's unlikely someone will read it

dkhalanskyjb: That depends. Do we plan to remove this opt-in flag in a release or two if no one complains or is it likely that we'll keep it? If it's the latter, then I don't think only documenting a configuration parameter in a changelog (or a closed issue) is sufficient.

qwwdfsad: I do not see any urge in removing it (e.g. we're still preserving flag for fast service loader), its maintenance has basically zero cost

dkhalanskyjb: A suggestion then: maybe later, outside of this pull request, we should document in the README (either the top-level one or for the core module) all the flags that our code non-transitively recognizes (or even transitively, if viable).

qwwdfsad: I would advise against it. All the options are intended to be user-facing features, but rather emergency brakes when everything went terribly wrong. Some of the flags (e.g. in the scheduler) can only be used for specific performance scenarios, and the overall behaviour with an arbitrary set of options is not well-defined.

dkhalanskyjb (approved):

PR #2982, opened 2021-10-14
Implemented `CopyableThreadContextElement`

### LINKED ISSUE #2839: Provide facilities for non-threadsafe mutable coroutine context element
Consider the use-case of keeping a trace of operations. In thread-based programming, the data structure that keeps the current trace is stored in a `ThreadLocal` variable and gets updated by the corresponding tracing calls in the thread. With coroutines, one can use `ThreadContextElement` to keep a reference to this structure. However, this element will get automatically inherited by children of the coroutine which might execute concurrently. This concurrency poses a problem when the data structure keeping a trace is not thread-safe and making it thread-safe might be impractical for performance reasons. The proposed solution is to add some mechanism, that would let the coroutine context element to copy itself during the creation of concurrent coroutines so that there can be a guarantee that each copy is accessed only sequentially.

yorickhenning:
This is PR #2936 after an explosive rewrite of git history and with feedback addressed. Sorry for the PR churn, my git is a bit rusty.

Fixes #2839.

This revision has reflects the `CopyableThreadContextElement` version of the change.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
@@ -17,10 +17,29 @@ import kotlin.coroutines.jvm.internal.CoroutineStackFrame
  */
 @ExperimentalCoroutinesApi
 public actual fun CoroutineScope.newCoroutineContext(context: CoroutineContext): CoroutineContext {
-    val combined = coroutineContext + context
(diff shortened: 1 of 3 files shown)


qwwdfsad (approved):
Great work, thanks!

Merged manually: https://github.com/Kotlin/kotlinx.coroutines/pull/3025

PR #3007, opened 2021-11-01
Update code to prepare for nullness annotations in rxjava3 and Guava.

cpovirk:
(In addition to keeping the code compiling in the future, this change
should make some runtime nullness errors impossible.)

rxjava3 nullness annotations don't yet trigger Kotlin compile errors,
but that will be changing in Kotlin 1.7:
https://github.com/JetBrains/kotlin/blob/05822c59b516b6d252bd6d27e9032e660e15b625/core/compiler.common.jvm/src/org/jetbrains/kotlin/load/java/JavaNullabilityAnnotationSettings.kt#L42-L46

We can preview the behavior by passing
-Xnullability-annotations=@io.reactivex.rxjava3.annotations:strict:
https://kotlinlang.org/docs/java-interop.html#nullability-annotations

We additionally set -Xtype-enhancement-improvements-strict-mode so that
the Kotlin compiler looks at type-use annotations on type arguments,
type parameters, etc.:
https://kotlinlang.org/docs/java-interop.html#annotating-type-arguments-and-type-parameters

Usually, the required update is to restrict a type parameter to
non-nullable types, since most rxjava types do not support null type
arguments. In a few cases, the update is to change an unnecessarily
nullable type to be non-nullable. Finally, I removed a `value == null`
check in `RxMaybeCoroutine.onCompleted` that the compiler now identifies
as unnecessary. (I can keep it if you'd prefer.)

Guava's current nullness annotations likewise don't yet trigger Kotlin
compile errors. However, Guava did recently remove the misleading
`@Nullable` annotation from the parameter of `FutureCallback.onSuccess`:

Before:
https://github.com/google/guava/blob/v28.0/guava/src/com/google/common/util/concurrent/FutureCallback.java#L34

After:
https://github.com/google/guava/blob/v31.0.1/guava/src/com/google/common/util/concurrent/FutureCallback.java#L35

That means that a `FutureCallback<T>` can now implement `onSuccess(T)`
instead of `onSuccess(T?)`. And with a future change to Guava, it will
_have to_.

When I updated this project from Guava 28.0 to Guava 31.0.1, I found
that that pulled in a newer version of the Checker Framework nullness
annotations. That version was build with a newer version of Gradle, so
it generates Gradle module metadata:
https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html

That metadata declares that the Checker Framework annotations require
Java 1.8. Even the old version had in fact required Java 1.8 -- and so
had even the old version of Guava -- just without having that encoded in
its metadata. So I fixed this by setting targetCompatibility and
sourceCompatibility to 1.8 for kotlinx-coroutines-guava, too.

I should warn you that I have very little understanding of coroutines
and of this library. I'm here because we're seeing compile errors inside
Google as we work to improve how we handle Kotlin-Java interoperability,
and these changes looked like they might be the right fixes. Sorry for
any mistakes.


reactive/kotlinx-coroutines-rx3/src/RxMaybe.kt
```diff
@@ -17,33 +17,33 @@ import kotlin.coroutines.*
...
     val coroutine = RxMaybeCoroutine(newContext, subscriber)
     subscriber.setCancellable(RxCancellable(coroutine))
     coroutine.start(CoroutineStart.DEFAULT, coroutine, block)
 }
 
-private class RxMaybeCoroutine<T>(
+private class RxMaybeCoroutine<T : Any>(
     parentContext: CoroutineContext,
     private val subscriber: MaybeEmitter<T>
 ) : AbstractCoroutine<T>(parentContext, false, true) {
```
dkhalanskyjb, line 43:
```suggestion
) : AbstractCoroutine<T?>(parentContext, false, true) {
```
With this, the incorrect change below is not needed.

reactive/kotlinx-coroutines-rx3/src/RxAwait.kt
```diff
@@ -42,7 +42,7 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
  * function immediately resumes with [CancellationException] and disposes of its subscription.
  */
 @Suppress("UNCHECKED_CAST")
-public suspend fun <T> MaybeSource<T>.awaitSingleOrNull(): T? = suspendCancellableCoroutine { cont ->
+public suspend fun <T : Any> MaybeSource<T>.awaitSingleOrNull(): T? = suspendCancellableCoroutine { cont ->
```
dkhalanskyjb, line 45: tl;dr: we'll probably just make this compile somehow, and only make the proposed change *after* 1.7.

Adding `T : Any` boundaries is a good change. In fact, it's so good that it's the third time it's suggested! The reason our earlier attempts failed is that this would break a lot of code generic in `T` that depends on this library: for example, https://github.com/Kotlin/kotlinx.coroutines/pull/2630. It looks like the change to the compiler will force our hand, but the general concern still stands: we don't want to introduce breaking changes that aren't that useful.

Due to this, I think the proper thing to do here is to make the code compile *without* introducing any incompatibility. When people migrate their code to 1.7, they will in any case have to adapt their code a lot, including the generic functions that we don't want to break. After they fix this on their side, we'll be able to remove the hacks that we'll need for the code to compile, and without breaking any code.

What do you think? I didn't understand from your description whether you actually need this change already for some reason.

dkhalanskyjb (changes requested):
Hi! Thanks for the PR. Do I understand correctly that the changes to the RxJava3 module are completely disjoint from the changes to the Guava integration? If so, it would probably be better if these were separated into different pull requests so we could merge them independently.

In any case, please change the base of this PR to the `develop` branch (see https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs). Also, if you agree with the reasoning that the fix to the RxJava3 module is premature, you could only rebase the changes to the Guava module.

cpovirk:
Thanks for the quick response. I should have thought to look for history, and sorry for getting distracted by some build issues and not reading the documentation about the branch to work on, etc. :(

I will rebase, split the Guava change out, read some more docs, see if I can find some more background on #2630, and get back to you -- hopefully later this week, but I have overcommitted myself a bit.

dkhalanskyjb:
Hey, no worries. The history for this is not that easy to find publically, I was simply describing the current state of things. Also, where I'm from, there are public holidays at the end of this week, so I wouldn't be able to respond until Monday anyway.

reactive/kotlinx-coroutines-rx3/src/RxAwait.kt
qwwdfsad: By the way, there is a way to workaround that: https://github.com/Kotlin/kotlinx.coroutines/pull/2954 
It's not currently merged due to IDE issues

reactive/kotlinx-coroutines-rx3/src/RxMaybe.kt
```diff
@@ -17,33 +17,33 @@ import kotlin.coroutines.*
  * If the context does not have any dispatcher nor any other [ContinuationInterceptor], then [Dispatchers.Default] is used.
  * Method throws [IllegalArgumentException] if provided [context] contains a [Job] instance.
  */
-public fun <T> rxMaybe(
+public fun <T : Any> rxMaybe(
     context: CoroutineContext = EmptyCoroutineContext,
-    block: suspend CoroutineScope.() -> T?
+    block: suspend CoroutineScope.() -> T
```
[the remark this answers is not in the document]
cshubham: works

reactive/kotlinx-coroutines-rx3/src/RxAwait.kt
cpovirk: Essentially, my _goal_ is to break a bunch of Google's code that depends on this library :)

Of course, I don't really _want_ to break our code. But as you say, I expect for us to _have_ to break it eventually. So I would rather break (well, fix) a small amount of code now than I would break (fix) a large amount of code later.

To be fair, this is a particularly good time for a breaking change _for us_: We don't have a ton of RxJava3 code _yet_, but we're expecting to start seeing a lot more soon. So it was pretty easy to update our existing users of these APIs a couple weeks ago, but already I'm seeing backsliding.

I understand that, in the wider community, these APIs already have far more users. Still, I think the principle stands: Updating users of the APIs is only going to get harder over time.

It would be great if we could avoid the breaking change entirely with #2954. However, I suspect that that's not going to work in the long term: Even an _API declaration_ like [`public suspend fun <T> SingleSource<T>.await()`](https://github.com/Kotlin/kotlinx.coroutines/pull/2954/files#diff-7c46154c0f1691c6a9d7804e3e19769f0f383b1894ee96b9e716889e8c2f68c5R117) is going to become an error eventually, since [`SingleSource` requires a non-null type argument](https://github.com/ReactiveX/RxJava/blob/e1e35b05030ee0b8bc5e7b1b5c7e47b8d7617019/src/main/java/io/reactivex/rxjava3/core/SingleSource.java#L29). So I think that even that API has to change to `public suspend fun <T> SingleSource<T`**` & Any`**`>.await()`. And I think that's as much a breaking change as this PR is.

---

I _may_ be able to mostly accomplish my goal another way: If I can set the `@io.reactivex.rxjava3.annotations:strict` flag for the rest of our codebase, then I _may_ be able to prevent callers from ever using these APIs with a nullable type argument. As it turns out, though, that's tricky because, if I set that flag for our whole codebase, we can no longer compile this project :) And our build setup / kotlinc doesn't appear to provide a good way to turn a flag "off" if it's on by default, so I can't easily turn it off just for this project. I can continue exploring ways to make that work, too; I'm just trying as many different approaches as I can.

reactive/kotlinx-coroutines-rx3/src/RxMaybe.kt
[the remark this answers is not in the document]
cpovirk: Resolved by your suggested edit below, as you said.

reactive/kotlinx-coroutines-rx3/src/RxConvert.kt
```diff
@@ -42,7 +42,7 @@ public fun Job.asCompletable(context: CoroutineContext): Completable = rxComplet
  *
  * @param context -- the coroutine context from which the resulting maybe is going to be signalled
  */
-public fun <T> Deferred<T?>.asMaybe(context: CoroutineContext): Maybe<T> = rxMaybe(context) {
+public fun <T : Any> Deferred<T>.asMaybe(context: CoroutineContext): Maybe<T> = rxMaybe(context) {
```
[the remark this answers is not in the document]
cpovirk: Resolved by your suggested edit below, as you said.

reactive/kotlinx-coroutines-rx3/src/RxAwait.kt
cpovirk: Two small updates on the breaking-changes front:
- First, if you ultimately find this PR to be too disruptive, it looks like we'll probably maintain it as a small patch for Google's copy. That unblocks my goal of flipping [-Xnullability-annotations=@io.reactivex.rxjava3.annotations:strict](https://kotlinlang.org/docs/java-interop.html#nullability-annotations) on for our depot. (I very much appreciate your catching my earlier errors so that I didn't break something, even if it would only have been in Google's code and not for _all_ rxjava3 users :))
- We've found that another future Kotlin flag flip, -Xenhance-type-parameter-types-to-def-not-null=true, forces our rxjava3 users to make some (but not all) of the same changes as this PR does. I don't know if that flag and the rxjava3.annotations flag will end up getting flipped by default in the same release of Kotlin or different releases. Nor do I know whether users would prefer to take all their rxjava3-related breakages at once or have multiple smaller breaking releases. But I wanted to pass along that that's another Kotlin future flag flip that may break some kotlinx-coroutines users.

cpovirk:
OK, 6 months later, I finally made some time to try to understand things instead of blindly changing `?`s :) And yes, all I needed to do was make the `RxMaybe` change you'd suggested....

Bookkeeping:
- This is now rebased on `develop`.
- The Guava-related changes were submitted separately in #3026, leaving only the rxjava3-specific changes in this PR.
- This still passes `JDK_16=$HOME/jdk1.6.0_45 ./gradlew assemble jvmTest` (and also the tests of our rxjava3 users inside Google).

That leaves only(?) the whole question of whether to make breaking changes now at all, as discussed in [the other thread](https://github.com/Kotlin/kotlinx.coroutines/pull/3007#discussion_r751444689).

cpovirk:
I realized that I never hit "Re-request review."

Again, though, I know you'd already indicated that you probably wouldn't merge this PR anytime soon (possibly never, in fact :)), so I'm hitting the button just in case you were waiting on me to do that. I think I've addressed the smaller comments, leaving only the big "Should we do this at all?" question. And as I said above, I expect to be able to patch Google's local copy of kotlinx.coroutines if you don't want to merge this.

qwwdfsad:
Hi! 

We are not yet decided what to do. As soon as we merge #3324 (which is blocked by 1.7.0 regression, so we are waiting for 1.7.10) where definitely-non-nullable types are available, we'll start investigating the compatibility story, that mostly lies within Java interop and flexible types. We'll figure the final shape out *before* 1.8.0 where Rx3 annotations are strict.
We do not have any preferences between generic parameter upper bound and `& Any` on receivers, we'll just pick the one that implies the least number of potentially source-breaking changes

PR #3011, opened 2021-11-08
Eagerly enter `launch` and `async` blocks with unconfined dispatcher.

dkhalanskyjb:
@qwwdfsad found a convenient migration path for most tests that rely on `launch` and `async` blocks being entered eagerly. Implemented it here.

Also, I fixed `Dispatchers.Main` not delegating `Delay` methods and discovered that, on JS, `Dispatchers.Main` gets reset during the test if it is reset in `AfterTest`.

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestCoroutineDispatchers.kt
+++ b/kotlinx-coroutines-test/common/src/TestCoroutineDispatchers.kt
@@ -7,6 +7,8 @@ package kotlinx.coroutines.test
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
 import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.test.internal.*
+import kotlinx.coroutines.test.internal.TestMainDispatcher
 import kotlin.coroutines.*
 
 /**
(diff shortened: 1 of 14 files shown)


kotlinx-coroutines-test/common/test/TestDispatchersTest.kt
```diff
@@ -4,16 +4,56 @@
...
+        Dispatchers.setMain(StandardTestDispatcher())
+    }
+
+    @AfterTest
+    fun tearDown() {
         Dispatchers.resetMain()
     }
 
+    /** Tests that asynchronous execution of tests does not happen concurrently with [AfterTest] (in fact, it does). */
+    @Ignore // fails on JS.
```
qwwdfsad, line 24: For such kludges, we typically introduce helpers like [`noJs`](https://github.com/Kotlin/kotlinx.serialization/blob/5ce8e2cbd8d324ed5e0447f133137ccacc9f6614/formats/json/commonTest/src/kotlinx/serialization/test/TestHelpers.kt#L27)

qwwdfsad (approved):
Neat!

kotlinx-coroutines-test/common/src/TestCoroutineDispatchers.kt
```diff
@@ -15,10 +17,32 @@ import kotlin.coroutines.*
  * This dispatcher is similar to [Dispatchers.Unconfined]: the tasks that it executes are not confined to any particular
  * thread and form an event loop; it's different in that it skips delays, as all [TestDispatcher]s do.
  *
+ * Like [Dispatchers.Unconfined], this one does not provide guarantees about the execution order when several coroutines
+ * are queued in this dispatcher. However, we ensure that the [launch] and [async] blocks at the top level of [runTest]
+ * are entered eagerly. This allows launching child coroutines and not calling [runCurrent] for them to start executing.
+ *
+ * ```
+ * @Test
+ * fun testEagerlyEnteringChildCoroutines() = runTest(UnconfinedTestDispatcher()) {
```
qwwdfsad, line 26: By the way, what do you think about providing `runTestUnconfined` for a better migration path?

dkhalanskyjb: I don't like it because of this confusing behavior:
```kotlin
val ctx = StandardTestDispatcher()

@Test
fun test() = runTestUnconfined(ctx) {
  // not unconfined
}
```

dkhalanskyjb:
I added separate `FailingTests` files that document and check the known problematic behaviors. I haven't seen such things in our codebase before—is there an established way to do this?

kotlinx-coroutines-test/common/test/TestDispatchersTest.kt
dkhalanskyjb: Good idea, using it now.

qwwdfsad (approved):
LGTM

What an elegant idea with `@NoPlatform` annotation!

qwwdfsad:
`FailingTests` is fine, we do not have en established way for this

PR #3015, opened 2021-11-11
Implement TestScope

dkhalanskyjb:
Is a draft because the new migration helpers need some tests.


kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,219 @@
...
+ *   paused by default, like [StandardTestDispatcher].
+ * * No access to the list of unhandled exceptions.
+ */
+@ExperimentalCoroutinesApi
+public sealed interface TestScope : CoroutineScope {
+    /**
+     * The delay-skipping scheduler used by the test dispatchers running the code in this scope.
+     */
+    @ExperimentalCoroutinesApi
+    public val testScheduler: TestCoroutineScheduler
```
qwwdfsad, line 47: Shouldn't it also provide typed access to `TestDispatcher`?

The benefits are the same as for creating `TestScope` standalone -- capability to configure entities that depend on either scope or dispatcher.

It's possible to argue that one can use `coroutineContext[CoroutineDispatcher]!!`, but it is both verbose and required `ExperimentalStdlibApi`, while having an explicit property with type `TestDispatcher` clearly indicates that scope is always bound to the instance of test dispatcher

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,219 @@
...
+
+internal class TestScopeImpl(context: CoroutineContext) :
+    AbstractCoroutine<Unit>(context, initParentJob = true, active = true), TestScope {
+
+    override val testScheduler get() = context[TestCoroutineScheduler]!!
+
+    var entered = false
+    var finished = false
+    val uncaughtExceptions = mutableListOf<Throwable>()
+    val lock = SynchronizedObject()
```
qwwdfsad, line 163: All these properties can be private?

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,219 @@
...
+                for (e in exceptions)
+                    addSuppressed(e)
+            }
+        }
+    }
+
+    /** Called at the end of the test. May only be called once. */
+    fun finish(): List<Throwable> {
+        val exceptions = synchronized(lock) {
+            check(entered && !finished)
```
qwwdfsad, line 185: If all these `check` calls can be triggered, then a (potentially actionable) diagnostic message should be added. Otherwise, it should be an assert

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,219 @@
...
+            }
+            dispatcher
+        }
+        null -> StandardTestDispatcher(get(TestCoroutineScheduler))
+        else -> throw IllegalArgumentException("Dispatcher must implement TestDispatcher: $dispatcher")
+    }
+    return this + dispatcher + dispatcher.scheduler
+}
+
+internal class TestScopeImpl(context: CoroutineContext) :
```
qwwdfsad, line 155: It's going to be used in tests (-> debugger), so pretty `toString` can be useful

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,219 @@
...
+    AbstractCoroutine<Unit>(context, initParentJob = true, active = true), TestScope {
```
qwwdfsad, line 156: Is making `TestScopeImpl` an `AbstractCoroutine` a deliberate decision?

Because now we have [potentially unexpected] broken invariant, e.g.:
```
val root = Job()
val scope = TestScope(root)
println(scope.coroutineContext.job == root) // False
```

Attempt to attach a child to such scope is also visible side-effect of this change

qwwdfsad (changes requested):

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: > `coroutineContext[CoroutineDispatcher]!!`

Or `StandardTestDispatcher(testCoroutineScheduler)`. I didn't want to prioritize specific dispatchers.

I believe that the reason to pass the test dispatcher somewhere explicitly is to ensure that a coroutine launched somewhere operates using the source of virtual time (from the sound of it I already get the feeling that it's the source of virtual time that should be passed), not a specific dispatching strategy: if it was the question of the strategy, then this coroutine would just be launched in this scope and inherit the dispatcher automatically.

From this point of view, both `StandardTestDispatcher(testCoroutineScheduler)` and `UnconfinedTestDispatcher(testCoroutineScheduler)` are valid, symmetric choices. However, introducing a `testDispatcher` field promotes one dispatching strategy (the one used for the scope) over the other.

Finally, if one does for some reason need to make sure that the scope and the other parts of the system use the same dispatcher, then this (I think, very specific) requirement can be fulfilled using
```kotlin
val dispatcher = StandardTestDispatcher()
val scope = TestScope(dispatcher)
```

Maybe I'm missing something and the need to pass the specific dispatcher would be much more frequent than I think?

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: This is deliberate, though maybe I'm missing the downsides.

My reasoning: with `TestCoroutineScope`, we had this hierarchy:
```
/ manually-created TestCoroutineScope
| / the child coroutines spawned off before the test
| \ 
| / the coroutine that is a new scope
| | | child coroutines spawned inside the test
| \ cleanupTestCoroutines after runTest
\ cleanupTestCoroutines in afterTest
```
This structure ensured that there were no uncompleted child jobs.

`TestScope` does not provide a user-visible cleanup procedure, so if we want to detect that the child coroutines spawned off before the test are not leaked, we need to perform their cleanup as well at the end of the test:
```
/ manually-created TestScope (if it is not a coroutine)
| / the child coroutines spawned off before the test
| \ 
| / a coroutine TestScope
| | | child coroutines spawned inside the test
| \ ...
\ cleanup after runTest
```
Cleaning up the parent job doesn't look natural in this case.

With the proposed scheme, we have the following:

```
/ potential root job
| / TestScope
| | /
| | | the child coroutines spawned off before the test
| | | child coroutines spawned inside the test
| | \ 
\ \ cleanup after runTest
```

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: `assert` is internal. Is it worth it to make it `PublishedApi`?

kotlinx-coroutines-test/common/src/TestScope.kt
qwwdfsad: That makes sense, thanks.

The downsides are kind of negligible -- the overall semantics is just not aligned with other scope factories. But it seems to be perfectly normal, taking into account all the limitations/machinery `TestScope` already has.

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,222 @@
...
+ * * If [context] doesn't have a [TestDispatcher], a [StandardTestDispatcher] is created.
+ * * A [CoroutineExceptionHandler] is created that makes [TestCoroutineScope.cleanupTestCoroutines] throw if there were
+ *   any uncaught exceptions, or forwards the exceptions further in a platform-specific manner if the cleanup was
+ *   already performed when an exception happened. Passing a [CoroutineExceptionHandler] is illegal, unless it's an
+ *   [UncaughtExceptionCaptor], in which case the behavior is preserved for the time being for backward compatibility.
+ *   If you need to have a specific [CoroutineExceptionHandler], please pass it to [launch] on an already-created
+ *   [TestCoroutineScope] and share your use case at
+ *   [our issue tracker](https://github.com/Kotlin/kotlinx.coroutines/issues).
+ * * If [context] provides a [Job], that job is used as a parent for the new scope;
+ *   otherwise, a [CompletableJob] is created.
```
qwwdfsad, line 105: The note about `CompletableJob` seems to be misleading, implicitly implying that one can complete it

qwwdfsad (commented):
Overall, the changes look good, looking forward to integrating that!

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -0,0 +1,222 @@
...
+    /** Called at the end of the test. May only be called once. */
+    fun leave(): List<Throwable> {
+        val exceptions = synchronized(lock) {
+            check(entered && !finished)
+            finished = true
+            uncaughtExceptions
+        }
+        if (exceptions.isEmpty() && (children.any { it.isActive } || !testScheduler.isIdle()))
+            throw UncompletedCoroutinesError(
+                "Unfinished coroutines during teardown. Ensure all coroutines are" +
```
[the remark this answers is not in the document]
dkhalanskyjb: Done.

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: Changed this.

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -186,11 +186,20 @@ internal class TestScopeImpl(context: CoroutineContext) :
             finished = true
             uncaughtExceptions
         }
-        if (exceptions.isEmpty() && (children.any { it.isActive } || !testScheduler.isIdle()))
-            throw UncompletedCoroutinesError(
-                "Unfinished coroutines during teardown. Ensure all coroutines are" +
-                    " completed or cancelled by your test."
-            )
+        val activeJobs = children.filter { it.isActive }.toList()
```
qwwdfsad, line 189: It seems like `it.isCompleted` should be used instead, as `isActive` being `false` may imply that the coroutine is in its cancellation sequence and still is executing.

Though I can't see how it is possible to trigger this check at all, could you please ensure it's still needed?

qwwdfsad: Probably it can be triggered by the timeout, but I neither see any tests nor sure that this exception will be actually the final one

dkhalanskyjb: If a timeout happened, this wouldn't be the final exception, no, it throws a more informative message instead.

I added this for use with `runBlockingTest`, which now supports using `TestScope` but doesn't properly await completion. Though looking at it, I think that it would just throw `IllegalStateException` in this case instead. Fixed it and added comments.

qwwdfsad (approved):

PR #3016, opened 2021-11-11
Base Mutex implementation on Semaphore and improve the latter

ndkoval:


kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -93,7 +110,7 @@ public interface Mutex {
  */
 @Suppress("FunctionName")
 public fun Mutex(locked: Boolean = false): Mutex =
-    MutexImpl(locked)
+    MutexImpl(locked) // TODO: locked with owner?
```
ndkoval, line 113: Should we add a function that specifies the owner?

ndkoval:
Only those tests that require `Mutex.onLock()` fail -- this will be fixed with new `select` implementation.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
qwwdfsad: Nope, it exists for ages and there wasn't a single request for that

qwwdfsad (commented):
By the way, you can reuse nom-intrusive implementation of `onLock`: https://github.com/Kotlin/kotlinx.coroutines/commit/7755edb43da2fbc9181e1ff4316ad34ffb1b5d7e#diff-86e8b6879026f347540a4b86b2c594fe25128a0fb168cc0689ef48008c8b4d5fR99

ndkoval:
@qwwdfsad I already have a proper implementation with new `select` API

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,76 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-    ) : AtomicOp<MutexImpl>() {
-        override fun prepare(affected: MutexImpl): Any? =
-            if (queue.isEmpty) null else UNLOCK_FAIL
-
-        override fun complete(affected: MutexImpl, failure: Any?) {
-            val update: Any = if (failure == null) EMPTY_UNLOCKED else queue
-            affected._state.compareAndSet(this, update)
+        override fun resume(value: Unit, onCancellation: ((cause: Throwable) -> Unit)?) {
+            this@MutexImpl.owner.value = owner
+            cont.resume(value, onCancellation)
```
qwwdfsad, line 203: What will happen if this resume will have no effect, e.g. because the continuation was concurrently cancelled?

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -52,24 +54,35 @@ public interface Mutex {
      * Note that this function does not check for cancellation when it is not suspended.
      * Use [yield] or [CoroutineScope.isActive] to periodically check for cancellation in tight loops if needed.
      *
-     * Use [tryLock] to try acquiring a lock without waiting.
+     * This function can be used in [select] invocation with [onLock] clause.
```
qwwdfsad, line 57: This part is irrelevant as `onLock` is being deprecated for removal

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,76 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-                // if this continuation gets cancelled during dispatch to the caller, then release the lock
-                unlock(owner)
-            } != null
+    private inner class CancellableContinuationWithOwner(
+        val cont: CancellableContinuation<Unit>,
+        val owner: Any?
+    ) : CancellableContinuation<Unit> by cont {
+        override fun tryResume(value: Unit, idempotent: Any?, onCancellation: ((cause: Throwable) -> Unit)?): Any? {
+            val token = cont.tryResume(value, idempotent, onCancellation)
+            if (token !== null) this@MutexImpl.owner.value = owner
```
qwwdfsad, line 197: Please add a comment why it is safe to have a race here -- the continuation can be already executing and even invoking `unlock` while owner is not set

qwwdfsad (changes requested):
I also see that the build is failing, please make it pass. You can temporary ignore `onLock` tests for that sake.

My biggest concern here is the relevance of the PR -- could you please elaborate on the purpose of the PR? What benefits does it bring?

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: It cannot be executing before `completeResume(..)`

ndkoval:
Thanks for the review -- I have fixed and clarified everything in #3020

PR #3018, opened 2021-11-11
Make ReceiveChannel.iterator() extension function

ndkoval:


ndkoval:
To show the performance impact I have run the existing benchmarks on an Intel machine that features four Intel Xeon Gold 5218 processors with 16 cores each and 256 Gb RAM. The results below show insignificant change.

Before the change:
```
```

After the change:
```
```

ndkoval:
Also, `ReusableCancellableContinuationTest` fails. @qwwdfsad, could you please help with understanding the problem?

qwwdfsad:
One doesn't need such a huge machine to verify this is a degrading change.

For the following benchmark:
```
@Benchmark
fun channelPipeline(bh: Blackhole): Unit = runBlocking {
    val channel = Channel.range(1, 10_000, Dispatchers.Unconfined)
        .filter(Dispatchers.Unconfined) { it % 4 == 0 }
    for (i in channel) {
        bh.consume(i)
    }
}
```

I see 66% decline in the performance:
```
// With the last commit applied
ChannelSinkBenchmark.channelPipeline  avgt    5  3.723 ± 0.069  ms/op
// Baseline
ChannelSinkBenchmark.channelPipeline  avgt    5  2.235 ± 0.098  ms/op
```

It would be nice to have a serious rationale prior to accepting the concept of the change and starting a review

ndkoval:
@qwwdfsad 
1. I just haven't marked the PR as a draft one -- it contained a non-optimized `hasNext` implementation. Now, the performance is similar -- see the experimental results above.
2. I used the remote machine I had access to. This PR does not change the iterator logic and only switches the implementation to use the standard channel operations instead of the internal API. Therefore, the performance stays the same (according to the benchmark results).
3. Could you please help to identify why the `ReusableCancellableContinuationTest` test fails?

qwwdfsad (changes requested):
The problem is still here.

The benchmarks you ran are mostly flow-related and do not really stress channel iterator in any way. 

[Here is my benchmark](https://gist.github.com/qwwdfsad/84f67d0bdf5117563a2bb28fc7975551), though the original ChannelSinkBenchmark should suffice as well and results are roughly the same:

```
// Current version
ChannelSinkBenchmark.channelPipeline  avgt    5  3.539 ± 0.071  ms/op
// Prior to the commit
ChannelSinkBenchmark.channelPipeline  avgt    5  2.323 ± 0.071  ms/op
```

50% is still way too much. Please bisect and analyze the performance impact prior to further discussion

PR #3019, opened 2021-11-11
Invoke `jmhJar` as a part of build procedure to ensure that benchmarks always compile

ndkoval:


ndkoval:
An example of failing build: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildAggregated/3692853?

qwwdfsad (approved):

PR #3020, opened 2021-11-11
New `select` and `Mutex` algorithms

ndkoval:
New `select` and `Mutex` algorithms


kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,120 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
+            val token = cont.tryResume(value, idempotent, onCancellation)
+            if (token !== null) this@MutexImpl.owner.value = owner
+            return token
         }
 
-        override fun completeResumeLockWaiter() = cont.completeResume(RESUME_TOKEN)
-        override fun toString(): String = "LockCont[$owner, ${cont}] for ${this@MutexImpl}"
+        override fun resume(value: Unit, onCancellation: ((cause: Throwable) -> Unit)?) {
+            this@MutexImpl.owner.value = owner
+            cont.resume(value, onCancellation)
```
qwwdfsad, line 229: Here it seems like it is possible to have an unlocked mutex with an owner different from `NO_OWNER` in case the cancellation of the outer job (that is parent of current `CancellableContinuationWithOwner`) happens right between setting owner and calling `resume`.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,120 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
+            select.selectInRegistrationPhase(internalResult)
+            this@MutexImpl.owner.value = owner
         }
     }
+
+    override fun toString() = "Mutex@${hexAddress}"
 }
+
+@SharedImmutable
+private val NO_OWNER = Symbol("NO_OWNER")
```
qwwdfsad, line 253: Please set a new line at the end of the file (it can be configured in IDEA)

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,120 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-        override fun complete(affected: MutexImpl, failure: Any?) {
-            val update: Any = if (failure == null) EMPTY_UNLOCKED else queue
-            affected._state.compareAndSet(this, update)
+        override fun selectInRegistrationPhase(internalResult: Any?) {
+            select.selectInRegistrationPhase(internalResult)
+            this@MutexImpl.owner.value = owner
         }
     }
+
+    override fun toString() = "Mutex@${hexAddress}"
```
qwwdfsad, line 249: It would be nice to have meaningful `toString` -- e.g. adding whether it's locked and the owner to the string

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,120 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-    ) : AtomicOp<MutexImpl>() {
-        override fun prepare(affected: MutexImpl): Any? =
-            if (queue.isEmpty) null else UNLOCK_FAIL
 
-        override fun complete(affected: MutexImpl, failure: Any?) {
-            val update: Any = if (failure == null) EMPTY_UNLOCKED else queue
-            affected._state.compareAndSet(this, update)
+        override fun selectInRegistrationPhase(internalResult: Any?) {
+            select.selectInRegistrationPhase(internalResult)
+            this@MutexImpl.owner.value = owner
```
qwwdfsad, line 245: I've not yet dived into selects yet, but can be there a race too? E.g. when select is selected concurrently, no one is going to reset this owner

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,120 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
+        }
 
-        override fun tryResumeLockWaiter(): Boolean {
-            if (!take()) return false
-            return cont.tryResume(Unit, idempotent = null) {
-                // if this continuation gets cancelled during dispatch to the caller, then release the lock
-                unlock(owner)
-            } != null
+    private inner class CancellableContinuationWithOwner(
+        val cont: CancellableContinuation<Unit>,
```
qwwdfsad, line 218: Here and below please mark these `val`-s with `@JvmField` to save a few bytes in dex/bytecode/classfile size

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +134,120 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
+    protected open fun onLockProcessResult(owner: Any?, result: Any?): Any? {
+        onAcquire.processResFunc(this, null, result)
+        return this
     }
 
-    private inner class LockCont(
-        owner: Any?,
-        private val cont: CancellableContinuation<Unit>
-    ) : LockWaiter(owner) {
+    protected val onCancellationUnlockConstructor: OnCancellationConstructor =
```
qwwdfsad, line 212: could be private and also please rename the second parameter to `_` (this is a built-in name for "unused parameter" in Kotlin)

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -140,74 +142,189 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
...
     override fun tryAcquire(): Boolean {
-        _availablePermits.loop { p ->
+        while (true) {
+            // Get the current number of available permits.
+            val p = _availablePermits.value
+            // Is the number of available permits greater
+            // than the maximal one because of an incorrect
+            // `release()` call without a preceding `acquire()`?
+            // Change it to `permits` and start from the beginning.
+            if (p > permits) {
```
qwwdfsad, line 164: Is this check added exclusively for mutex?

If so, it's never executed, because incorrect `unlock` never reaches `release` code-path and gets stuck in owner check

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: it cannot be selected concurrently in the registration phase

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: it is sometimes better to keep the name of the ignoring parameter for readability -- in my opinion, this is the case

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
ndkoval: No, this is a part of small optimization that leverages Fetch-And-Add in `release`.

ndkoval: As mentioned in the comments, this situation may happen due to an incorrect `release()` call, without a preceding `acquire()`.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: Nice catch! Indeed, an `onCancellation` action should be provided to this `resume` call.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -52,33 +54,48 @@ public interface Mutex {
...
-     * Deprecated for removal without built-in replacement.
+     * Clause for [select] expression of [lock] suspending function that selects when the mutex is locked.
+     * Additional parameter for the clause in the `owner` (see [lock]) and when the clause is selected
+     * the reference to this mutex is passed into the corresponding block.
+     *
+     * It is recommended to use [withLock] for safety reasons, so that the acquired lock is always
+     * released at the end of the critical section, and [unlock] is never invoked before a successful
+     * lock acquisition.
      */
     @Deprecated(level = DeprecationLevel.WARNING, message = "Mutex.onLock deprecated without replacement. " +
```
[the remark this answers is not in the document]
ndkoval: I would prefer to keep it as we use `onLock` internally.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
qwwdfsad: Because that's our common code-style and we prefer keeping our codebases consistent, instead of loads of diffs that add and remove trailing newline

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: Surprisingly, the current implementation works _almost_ correctly. After this buggy situation, the following lock acquisition that successfully acquires the semaphore permit _blindly_ overwrites owner. I have fixed this behavior, adding more invariant `assert`-s to the algorithm and a new stress test.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -84,6 +81,39 @@ public class PublisherCoroutine<in T>(
...
+     * and implement [onSend], we use a modified version of [Mutex.onLock], which
+     * invokes [doLockedNext] at the end and returns this [PublisherCoroutine] as result
+     * (see [PublisherCoroutineMutex.onLockProcessResult]).
+     *
+     * We use this dirty hack as we need to wait for a lock, and, therefore, the clause
+     * object should be a [Mutex] instance. Thus, the [doLockedNext] call must be a part
+     * of [PublisherCoroutineMutex.onLockProcessResult] as well. A possible alternative
+     * would be to inherit [PublisherCoroutine] from [Mutex], but it already extends [AbstractCoroutine].
+     */
+    @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE", "CANNOT_OVERRIDE_INVISIBLE_MEMBER")
```
qwwdfsad, line 99: I finally see what's going on here.

I suggest reverting this and fallback to previous behaviour: `onLock` is deprecated **for removal**. It means that eventually it at least will be hidden and no amount of suppresses will help us, forcing us to fallback to the previous hack with a coroutine anyways.

The earlier we start using such code, the better (so any potential regression will be found, the current workaround won't be forgotten and so on). Bonus points -- we won't have updating this code when raising deprecation levels.

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -140,74 +142,179 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
...
         // the regular (fast) path -- if the cell is empty, try to install continuation
-        if (segment.cas(i, null, cont)) { // installed continuation successfully
-            cont.invokeOnCancellation(CancelSemaphoreAcquisitionHandler(segment, i).asHandler)
+        if (segment.cas(i, null, waiter)) { // installed continuation successfully
+            when (waiter) {
+                is CancellableContinuation<*> -> {
+                    waiter.invokeOnCancellation(CancelSemaphoreAcquisitionHandler(segment, i).asHandler)
+                }
+                is SelectInstance<*> -> {
+                    waiter.disposeOnCompletion(CancelSemaphoreAcquisitionHandler(segment, i))
```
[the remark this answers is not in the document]
ndkoval: Could you please explain why may I need it here?

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +135,138 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
+            if (token != null) {
+                assert { this@MutexImpl.owner.value === NO_OWNER }
+                this@MutexImpl.owner.value = owner
+            }
+            return token
         }
 
-        override fun completeResumeLockWaiter() = cont.completeResume(RESUME_TOKEN)
-        override fun toString(): String = "LockCont[$owner, ${cont}] for ${this@MutexImpl}"
+        override fun resume(value: Unit, onCancellation: ((cause: Throwable) -> Unit)?) {
```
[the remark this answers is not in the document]
ndkoval: It's not, it should be `{ release() }`.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +135,138 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-        abstract fun completeResumeLockWaiter()
+    protected open fun onLockProcessResult(owner: Any?, result: Any?): Any? {
+        return this
     }
 
-    private inner class LockCont(
-        owner: Any?,
-        private val cont: CancellableContinuation<Unit>
-    ) : LockWaiter(owner) {
+    private val onCancellationUnlockConstructor: OnCancellationConstructor =
```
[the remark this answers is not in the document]
ndkoval: It is easier to read the code when a logical part is not split into pieces. 

The pattern is to allocate `OnCancellationConstructor` at once, reusing it on each `onLock` invocation. I would prefer to keep the code consistent in all implementations instead of trying to save a couple of bytes here.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
-     * * [RETRY_ATOMIC] on deadlock (needs retry, it is only possible when [otherOp] is not `null`)
-     * * `null` on failure to select (already selected).
-     * [otherOp] is not null when trying to rendezvous with this select from inside of another select.
-     * In this case, [PrepareOp.finishPrepare] must be called before deciding on any value other than [RETRY_ATOMIC].
-     *
-     * Note, that this method's actual return type is `Symbol?` but we cannot declare it as such, because this
-     * member is public, but [Symbol] is internal. When [SelectInstance] becomes a `sealed interface`
-     * (see KT-222860) we can declare this method as internal.
+     * When this `select` instance is stored as a waiter, the specified [handle][disposableHandle]
+     * defines how the stored `select` should be removed in case of cancellation or another clause selection.
```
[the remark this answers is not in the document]
ndkoval: That is not the case anymore. As I see, we usually do not mention that handlers can be invoked at most once -- this is the default behavior.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
ndkoval: Why do you want to completely remove `onLock` when its support for binary compatibility is very cheap with this new `select` implementation?

qwwdfsad: It's always cheaper to remove things when possible, because eventually `select` implementation (or any other underlying primitive) will be changed or tweaked, but as long as `onLock` is deprecated with ERROR+, we won't be able to test it properly.

Even if we'll decide not to remove it, it definitely will be promoted to `HIDDEN` deprecation and we won't be able to use it here, so the fallback is required

ndkoval: OK, it is better then to refactor the code another way, using the registration function directly.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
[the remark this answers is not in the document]
qwwdfsad: We won't be soon as we've discussed below

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
+        @Suppress("UNCHECKED_CAST")
+        val selectedClause = state.value as ClauseData<R>
+        // Process the internal result.
+        val blockArgument = selectedClause.processResult(internalResult)
+        // Perform the clean-up before the user-specified block
+        // invocation to guarantee the absence of memory leaks.
+        cleanup(selectedClause)
+        // TAIL-CALL OPTIMIZATION: the `suspend` block
+        // is invoked at the very end.
+        return selectedClause.invokeBlock(blockArgument)
```
qwwdfsad, line 641: There are two places where an exception can be thrown and thus stacktrace recovery _with outer continuation_ should be used: `processResult` and `invokeBlock`.

I remember we decided that the performance of the select _at this level_ does not really matter, so instead of trying to find a hacky workaround I suggest simply giving up TCO here (we're allocating a lot anyways) and use proper stacktrace recovery

kotlinx-coroutines-core/jvm/test/MutexStressTest.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlinx.coroutines.*
+import kotlinx.coroutines.internal.*
+import kotlinx.coroutines.selects.*
+import kotlinx.coroutines.sync.*
+import org.junit.*
+import java.util.concurrent.*
+
+class MutexStressTest : TestBase() {
+    @Test
+    fun testStressCancellationDoesNotBreakMutex() = runTest {
```
qwwdfsad, line 14: This test fails on a regular basis both locally and on TC: https://teamcity.jetbrains.com/viewLog.html?buildId=3801302&buildTypeId=KotlinTools_KotlinxCoroutines_BuildAggregated

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/channels/AbstractChannel.kt
```diff
@@ -188,20 +175,39 @@ internal abstract class AbstractSendChannel<E>(
...
             if (isFullImpl) {
-                val send = if (onUndeliveredElement == null)
-                    SendElement(element, cont) else
-                    SendElementWithUndeliveredHandler(element, cont, onUndeliveredElement)
+                val send: Send = if (waiter is SelectInstance<*>) {
+                    if (onUndeliveredElement == null) SendElementSelect(element, waiter, this)
+                    else SendElementSelectWithUndeliveredHandler(element, waiter, this, onUndeliveredElement)
+                } else {
+                    if (onUndeliveredElement == null) SendElement(element, waiter as CancellableContinuation<Unit>)
+                    else SendElementWithUndeliveredHandler(element, waiter as CancellableContinuation<Unit>, onUndeliveredElement)
```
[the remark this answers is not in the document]
ndkoval: Yes, I just do not care -- this code will be removed anyway

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: OK 👌

kotlinx-coroutines-core/common/src/selects/Select.kt
ndkoval: We use this code on the fast path, so am not sure that this is a good idea

ndkoval:
```
internal open suspend fun doSelect(): R =
        if (isSelected) complete()  // Fast path
        else doSelectSuspend()      // Slow path
```
Here, we invoke `complete()` on the fast path

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
+            // of registered clauses and store the provided via
+            // [invokeOnCompletion] completion action into the clause.
+            //
+            // Importantly, the [waitUntilSelected] function is implemented
+            // carefully to ensure that the cancellation handler has not been
+            // installed when clauses re-register, so the logic below cannot
+            // be invoked concurrently with the clean-up procedure.
+            if (!reregister) clauses!! += this
+            disposableHandle = this@SelectImplementation.disposableHandle
+            this@SelectImplementation.disposableHandle = null
```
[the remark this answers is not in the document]
ndkoval: That's a great idea, thanks! I will do the refactoring.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
+    /**
+     * Returns `true` if this `select` is cancelled.
+     */
+    private val isCancelled
+        get() = state.value is Cancelled
 
+    /**
+     * List of clauses waiting on this `select` instance.
+     */
+    private var clauses: MutableList<ClauseData<R>>? = ArrayList(2)
```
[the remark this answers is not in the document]
ndkoval: Unfortunately, `emptyList()` is not an instance of `MutableList`. Thus, I would keep it as is.

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: Actually, this change seems almost impossible, as I will add a special `disposeOnCompletion(segment, index)` method later, so no `DisposableHandle` instance will be returned in new channels and semaphore. Even now, semaphore modification seems weird with such API.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
-public interface SelectClause0 {
-    /**
-     * Registers this clause with the specified [select] instance and [block] of code.
-     * @suppress **This is unstable API and it is subject to change.**
-     */
-    @InternalCoroutinesApi
-    public fun <R> registerSelectClause0(select: SelectInstance<R>, block: suspend () -> R)
+public interface SelectClause0 : SelectClause
+@InternalCoroutinesApi
+public class SelectClause0Impl(
```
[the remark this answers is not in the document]
ndkoval: Marked them as internal, but kept them where they stay for code readability -- it is better not to split the logic.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
+            // be invoked concurrently with the clean-up procedure.
+            if (!reregister) clauses!! += this
+            disposableHandle = this@SelectImplementation.disposableHandle
+            this@SelectImplementation.disposableHandle = null
+        } else {
+            // This clause has been selected!
+            // Update the state correspondingly.
+            state.update {
+                if (it is Cancelled) {
+                    createOnCancellationAction(this@SelectImplementation, internalResult)?.invoke(it.cause)
```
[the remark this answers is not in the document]
ndkoval: You are right, this branch is never reachable -- I accidentally kept it when testing a prototype of the `selectUntil` feature. Removed, thanks for finding the mistake!

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: I do not know how to make this part better. Yet, one may think that `selectInRegistrationPhase` should be invoked when `resume` is called for plain continuations (we almost always invoke it during registration). Similarly, `trySelect` should be invoked in the cases when `tryResume` Is called for continuations. Similarly, `disposeOnCompletion` is invoked instead of `invokeOnCompletion`.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,696 @@ public interface SelectBuilder<in R> {
...
+     * updates the state to this clause reference.
+     */
+    protected fun ClauseData<R>.register(reregister: Boolean = false) {
+        // Is there already selected clause?
+        if (state.value.let { it is ClauseData<*> || it is Cancelled }) return
+        // For new clauses, check that there does not exist
+        // another clause with the same object.
+        if (!reregister) checkClauseObject(clauseObject)
+        // Try to register in the corresponding object.
+        if (tryRegister(this@SelectImplementation)) {
```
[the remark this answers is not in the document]
ndkoval: Clarified + renamed to `tryRegisterAsWaiter`

kotlinx-coroutines-core/jvm/test/MutexStressTest.kt
ndkoval: Works fine both locally and on Teamcity now

kotlinx-coroutines-core/common/src/selects/Select.kt
qwwdfsad: When we first discussed the review and you were asked about benchmarks, the answer was "performance of the select is not important", which is fine.
Now, when we are talking about debuggability and UX, the performance is important for some reason, which may be  also fine in isolation.

But these two reasonings together just don't add up, please stick with a consistent approach, and if you are choosing performance over user experience -- it's better have some numbers to support the decision about giving up user's convenience.

I've pushed `SelectBenchmark` that just measures the overhead of `select` and the experiment shows that losing tail-call in `doSelect` costs 5% of throughput _at best_, which is completely reasonable for stacktrace recovery, so please implement it

qwwdfsad: Also, I've measured the old select. New one is 50% faster, which is really good for the implementation that do not care about performance and indicates that we are on the right track

ndkoval: I found even better solution:
```
return if (!RECOVER_STACK_TRACES) {
            // TAIL-CALL OPTIMIZATION: the `suspend` block
            // is invoked at the very end.
            selectedClause.invokeBlock(blockArgument)
        } else {
            // In the debug mode, we need to recover
            // the stack-trace of possible exception.
            // Thus, the tail-call optimization cannot be applied.
            try {
                selectedClause.invokeBlock(blockArgument)
            } catch (e: Throwable) {
                throw recoverStackTrace(e)
            }
        }
```

Updated the code correspondingly.

qwwdfsad:
For the record, rebase on `develop` has rewritten the author of half of the commits, from a31621a to e38ca48

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
...
+     * It is worth mentioning that the algorithm above provides "obstruction-freedom" non-blocking guarantee
+     * instead of the standard "lock-freedom" to avoid using heavy descriptors. In practice, this relaxation
+     * does not make significant difference. However, it is vital for Kotlin Coroutines to provide some
+     * non-blocking guarantee, as users may add blocking code in [SelectBuilder], and this blocking code
+     * should not cause blocking behaviour in other places, such as an attempt to make a rendezvous with
+     * the `select` that is hang in REGISTRATION phase.
+     *
+     * Also, this implementation is NOT linearizable under some circumstances. The reason is that a rendezvous
+     * attempt with `select` (via [SelectInstance.trySelect]) may fail when this `select` operation is still
+     * in REGESTRATION phase. Consider the following situation on two empty rendezvous channels `c1` and `c2`
```
qwwdfsad, line 325:
```suggestion
     * in REGISTRATION phase. Consider the following situation on two empty rendezvous channels `c1` and `c2`
```

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
qwwdfsad: >Actually, this change seems almost impossible, as I will add a special disposeOnCompletion(segment, index) method later, so no DisposableHandle instance will be returned in new channels and semaphore. Even now, semaphore modification seems weird with such API.

Could you please elaborate on how exactly `disposableHandle` will be propagated to `ClauseData` if there is no disposable handle at all? I'm ready to take a look at any draft PR to get an idea of that, because for now it looks like this spilling protocol can be avoided

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
qwwdfsad: Then I suggest just calling `clear()` instead of nulling out the slot.

For now, `null` doesn't really prevent anything except making code much harder to read -- somewhere `!!` is used, and `?` in other places, making the reasoning really hard, I've stumbled across it once again

If users abuse API hard (e.g. by storing `select` instance in a field), they will get caught by our asserts, and if it's a race -- all bets are off anyway

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
...
+    val token = tryResume(Unit, null, onCancellation) ?: return false
+    completeResume(token)
+    return true
+}
 
-    private class DisposeNode(
-        @JvmField val handle: DisposableHandle
-    ) : LockFreeLinkedListNode()
+// Cancelled state with the specified cause.
+private class Cancelled(@JvmField val cause: Throwable)
```
qwwdfsad, line 770: `cause` is never used, so you can replace it with a singleton

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
...
+@SharedImmutable
+private val NO_RESULT = Symbol("NO_RESULT")
+// We use this marker parameter objects to distinguish
+// SelectClause[0,1,2] and invoke the user-specified block correctly.
+@SharedImmutable
+@JvmField
+internal val PARAM_CLAUSE_0 = Symbol("PARAM_CLAUSE_0")
+@SharedImmutable
+@JvmField
+internal val PARAM_CLAUSE_1 = Symbol("PARAM_CLAUSE_1")
```
qwwdfsad, line 805: This one seems redundant and is not checked for anywhere

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: Then there is another problem --  the `findClause` function iterates over the list of clauses and can be invoked concurrently with `clear()`, which leads to throwing a `ConcurrentModificationException`. Thus, the best way is to replace the existing mutable list with either `null` or `emptyList()`.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
ndkoval: The code overrides `onLockRegFunction` and `onLockProcessResult` functions, which are internal. After promoting the deprecation level for `onLock`, its usage can be replaced with an internal `onLockInternal`. 

We can keep the previous workaround, but it may introduce unpredictable behavior due to starting a new coroutine (even in the UNDISPATCHED mode), similar to the old `select` API implementation via the new one.

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: There will be an addition `disposeOnCompletion(segment: Segment<*>, index: Int)` function to avoid creating `DisposableHandle`-s. I suggested the same optimization for continuations in #3084.

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
qwwdfsad: Then probably setting it to empty mutable list with a corresponding commentary about why it is applicable should work

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
qwwdfsad: But `DisposableHandle` protocol will still be required for things like `onTimeout`, won't it?

qwwdfsad:
Also, [ChannelUndeliveredElementSelectOldStressTest](https://teamcity.jetbrains.com/viewLog.html?buildId=3863540&buildTypeId=KotlinTools_KotlinxCoroutines_BuildAggregated#testNameId1929048611229104549) fails, please fix that as well, we would like to avoid having any flaking tests in the mainline

ndkoval:
As for the flacking test, we have already had the discussion and decided to ignore them as everything works fine with the new channels. I have no idea how to fix the problem.

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: With the optimization that avoids creating `DisposableHandle`-s, there will be two different APIs: one that manipulates `DisposableHandle`-s and one that uses `Segment` and `index` directly. That is why I am suggesting keeping the API as-is. Also, I do like the fact that the current API is consistent with the `CancellableContinuation` API, which makes it easier to support `select` further.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
...
  * Clause for [select] expression without additional parameters that does not select any value.
  */
-public interface SelectClause0 {
-    /**
-     * Registers this clause with the specified [select] instance and [block] of code.
-     * @suppress **This is unstable API and it is subject to change.**
-     */
-    @InternalCoroutinesApi
-    public fun <R> registerSelectClause0(select: SelectInstance<R>, block: suspend () -> R)
+public interface SelectClause0 : SelectClause
```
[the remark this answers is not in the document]
ndkoval: `SelectClause` is already a `sealed interface`; should we mark `SelectClauseX` sealed then?

kotlinx-coroutines-core/jvm/test/channels/ChannelUndeliveredElementStressTest.kt
```diff
@@ -220,12 +221,22 @@ class ChannelUndeliveredElementStressTest(private val kind: TestChannelKind) : T
     }
 
     private inner class Data(val x: Long) {
-        private val failedToDeliver = atomic(false)
+        private val firstFailedToDeliverOrReceivedCallTrace = atomic<Exception?>(null)
```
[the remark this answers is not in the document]
ndkoval: While debugging this new `select` implementation and the new algorithm for the channels, I investigated many bugs, and all of them did reproduce with the tracing. As for the timings, without the tracing, the test takes 5 minutes on my laptop, while with the tracing enabled it takes 5 minutes and 25 seconds. Considering the usefulness of the tracing and the test timing effect, I made it optional.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
      * @param timeMillis timeout time in milliseconds.
      */
     @ExperimentalCoroutinesApi
-    public fun onTimeout(timeMillis: Long, block: suspend () -> R)
+    @Deprecated(
+        message = "Replaced with the same extension function",
+        level = DeprecationLevel.HIDDEN, replaceWith = ReplaceWith("onTimeout")
+    )
+    public fun onTimeout(timeMillis: Long, block: suspend () -> R): Unit = onTimeout(timeMillis, block)
```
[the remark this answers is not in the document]
ndkoval: I have updated the deprecation level, but the "replace-with" feature works buggy on my local machine. Could you please check it on your laptop?

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
...
+     * (3) Finally, the user-specified block is invoked
+     * with the processed result as an argument.
+     */
+    private suspend fun complete(): R {
+        assert { isSelected }
+        // Get the selected clause.
+        @Suppress("UNCHECKED_CAST")
+        val selectedClause = state.value as ClauseData<R>
+        // Process the internal result.
+        val blockArgument = selectedClause.processResult(internalResult)
```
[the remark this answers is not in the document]
ndkoval: We indeed need to perform the clean-up phase before processing the internal result; thank you for the catch! The issue, however, was in the `selectOld` implementation; it was buggy. Please review the updated version.

reactive/kotlinx-coroutines-reactive/src/Publish.kt
ndkoval: I did not know that this code is marginal. In this case, I agree with your point. I did a similar trick, please check the code.

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: I have improved the documentation.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +97,709 @@ public interface SelectBuilder<in R> {
...
 
 /**
- * Internal representation of select instance. This instance is called _selected_ when
- * the clause to execute is already picked.
+ * Internal representation of `select` instance.
  *
- * @suppress **This is unstable API and it is subject to change.**
+ * @suppress **This is unstable API, and it is subject to change.**
  */
 @InternalCoroutinesApi // todo: sealed interface https://youtrack.jetbrains.com/issue/KT-22286
```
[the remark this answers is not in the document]
qwwdfsad: Good point, will leave it as a documentation improvement

qwwdfsad:
TODO for me: cleanup documentation when it's the time

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
whyoleg: maybe make it sealed, and introduce some `internal interface InternalSelectInstance: SelectInstance` for internal coroutines usage, to make it truly sealed?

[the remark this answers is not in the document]
qwwdfsad: That's a good idea, thanks!
@ndkoval could you please explore it further?

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +98,725 @@ public interface SelectBuilder<in R> {
...
-        init {
-            desc.atomicOp = this
+    private suspend fun ClauseData<R>.invokeBlockRecoveringException(blockArgument: Any?): R =
+        try {
+            // In the debug mode, we need to recover
+            // the stack-trace of possible exception.
+            // Thus, the tail-call optimization cannot be applied.
+            invokeBlock(blockArgument)
+        } catch (e: Throwable) {
+            throw recoverStackTrace(e)
```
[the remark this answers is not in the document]
ndkoval: The `StackTraceRecoverySelectTest.testSelectJoin` is exactly the test we need. I do not see any difference when replacing 
```
            throw recoverStackTrace(e)
```
with 
```
            suspendCancellableCoroutine {
                it.resumeWithException(recoverStackTrace(e))
            }
```

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +98,728 @@ public interface SelectBuilder<in R> {
...
+         * Processes the internal result provided via either
+         * [SelectInstance.selectInRegistrationPhase] or
+         * [SelectInstance.trySelect] and returns an argument
+         * for the user-specified [block].
+         *
+         * Importantly, this function may throw an exception
+         * (e.g., when the channel is closed in [Channel.onSend], the
+         * corresponding [ProcessResultFunction] is bound to fail).
+         */
+        fun processResult(result: Any?) = try {
```
qwwdfsad, line 752: In order to make a proper recovery, this function should be `suspend` and it should use `recoverAndThrow`.

`recoverStackTrace(e)` is basically meaningless and was introduced for a very narrow use-case

qwwdfsad (changes requested):
Also, please revisit `SelectInstance` being sealed as Oleg suggested

kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
```diff
@@ -592,7 +593,6 @@ public abstract interface class kotlinx/coroutines/channels/ActorScope : kotlinx
 
 public final class kotlinx/coroutines/channels/ActorScope$DefaultImpls {
 	public static synthetic fun cancel (Lkotlinx/coroutines/channels/ActorScope;)V
-	public static fun getOnReceiveOrNull (Lkotlinx/coroutines/channels/ActorScope;)Lkotlinx/coroutines/selects/SelectClause1;
```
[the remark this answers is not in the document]
ndkoval: I will do an unchecked cast to `AbstractChannel`, we have discussed this in Slack

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
ndkoval: @whyoleg thanks for the suggestion!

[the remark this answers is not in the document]
whyoleg: looks like comment `todo: sealed interface https://youtrack.jetbrains.com/issue/KT-22286` could be also removed
BTW, why at all it should be public?
I see only one usage in public API - SelectImplementation$ClauseData.createOnCancellationAction which itself is internal + PublishedApi

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -52,606 +98,728 @@ public interface SelectBuilder<in R> {
...
-                    if (otherOp == null) {
-                        // regular trySelect -- just mark as select
-                        if (!_state.compareAndSet(NOT_SELECTED, null)) return@loop
-                    } else {
-                        // Rendezvous with another select instance -- install PairSelectOp
-                        val pairSelectOp = PairSelectOp(otherOp)
-                        if (!_state.compareAndSet(NOT_SELECTED, pairSelectOp)) return@loop
-                        val decision = pairSelectOp.perform(this)
-                        if (decision !== null) return decision
+    fun trySelectDetailed(clauseObject: Any, result: Any?) =
```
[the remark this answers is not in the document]
qwwdfsad: Thanks

kotlinx-coroutines-core/common/src/selects/Select.kt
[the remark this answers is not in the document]
qwwdfsad: It doesn't work, and I can't really make it work. We'll live with it unfortunately

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
```diff
@@ -140,74 +142,179 @@ private class SemaphoreImpl(private val permits: Int, acquiredPermits: Int) : Se
...
         val curTail = this.tail.value
         val enqIdx = enqIdx.getAndIncrement()
         val segment = this.tail.findSegmentAndMoveForward(id = enqIdx / SEGMENT_SIZE, startFrom = curTail,
             createNewSegment = ::createSegment).segment // cannot be closed
         val i = (enqIdx % SEGMENT_SIZE).toInt()
         // the regular (fast) path -- if the cell is empty, try to install continuation
-        if (segment.cas(i, null, cont)) { // installed continuation successfully
-            cont.invokeOnCancellation(CancelSemaphoreAcquisitionHandler(segment, i).asHandler)
+        if (segment.cas(i, null, waiter)) { // installed continuation successfully
+            when (waiter) {
```
dkhalanskyjb, line 293: I think that the code would be much cleaner if we introduced a private sealed interface to represent the possible types of waiters. This would cost us one additional application per slow-path acquire, but get rid of `Any`, impossible `error`, etc.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -117,307 +131,138 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-                else -> error("Illegal state $state")
-            }
+        while (true) {
+            // Is this mutex locked?
+            check(isLocked) { "This mutex is not locked" }
+            // Read the owner, waiting until it is set in a spin-loop if required.
+            val curOwner = this.owner.value
+            if (curOwner === NO_OWNER) continue // <-- ATTENTION, BLOCKING PART HERE
+            // Check the owner.
+            check(curOwner === owner) { "This mutex is locked by $curOwner, but $owner is expected" }
```
dkhalanskyjb, line 191: This breaks some existing and documented behavior: see #2796. Please also add a test for this.

dkhalanskyjb (changes requested):

qwwdfsad (approved):
Please address Dmitry's comments and it's good to go

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
ndkoval: I agree with you, but your suggestion requires either `CancellableContinuation` to implement this sealed interface or adding unsafe casts at every `CancellableContinuation` usage (in case only`CancellableContinuationImpl` implements the interface). I do not see a good solution, so I would keep the current code.

kotlinx-coroutines-core/common/src/sync/Mutex.kt
ndkoval: Fixed

ndkoval:
I went through Dmitry's comments and checked the ABI manually.

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
qwwdfsad: Dima, thanks, it's indeed worth doing. I'll try to do it separately in a branch with new channels, so the disturbance will be minimal. I have a prototype for selects only and it looks much more readable

hrach:
👏

kotlinx-coroutines-core/common/src/sync/Semaphore.kt
dkhalanskyjb: The idea was to do something like this (pseudocode):
```kotlin
private sealed interface ResumableComputation {
  fun resume()
  fun disposeOnCancellation(handle: DisposableHandle)
}

private class CancellableContinuationAsResumable(
  val continuation: CancellableContinuation<Unit>
): ResumableComputation {
  ...
}
```
So, `CancellableContinuation` would not itself implement anything (though maybe this interface makes sense in general and it should), it would be wrapped in an adapter. The cost is one more allocation on the slow path.

PR #3023, opened 2021-11-16
Documentation for the test module

### LINKED ISSUE #1390: Provided example test for withTimeout fails
I just copied the following into my IDE from the [README.md](https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-test#testing-launch-or-async-with-delay)

```kotlin
@Test(expected = TimeoutCancellationException::class)
fun testFooWithTimeout() = runBlockingTest {
    val uncompleted = CompletableDeferred<Unit>() // this Deferred<Foo> will never complete
    foo(uncompleted)
    advanceTimeBy(1_000) // advance time, which will cause the timeout to throw an exception
    // ...
}

fun CoroutineScope.foo(resultDeferred: Deferred<Unit>) {
    launch {
        withTimeout(1_000) {
            resultDeferred.await() // await() will suspend forever waiting for uncompleted
            // ...
        }
    }
}
```

I expected the test to pass, but it failed with the following exception:

```kotlin
java.lang.AssertionError: Expected exception: kotlinx.coroutines.TimeoutCancellationException

	at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:32)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
```

`advanceTimeBy(..)` does cause the timeout exception to be thrown, but `runBlockingTest` may be swallowing it.

dkhalanskyjb:
Fixes #1390


kotlinx-coroutines-test/README.md
```diff
@@ -52,372 +71,296 @@ class SomeTest {
...
 }
 ```
 
-*Note:* Prefer to provide `TestCoroutineScope` when it does not complicate code since it will also elevate exceptions 
-to test failures. However, exposing a `CoroutineScope` to callers of a function may lead to complicated code, in which 
-case this is the preferred pattern.
+## Detecting unfinished jobs
+
+The test framework ensures that coroutines are not leaked.
+To this end, the coroutines that aren't finished by the end of the test are reported.
```
qwwdfsad, line 324: Do we really need this part now when we have structured concurrency in `runTest`?

kotlinx-coroutines-test/MIGRATION.md
```diff
@@ -0,0 +1,324 @@
+# Migration to the new kotlinx-coroutines-test API
+
+In version 1.6.0, the API of the test module changed significantly.
+This is a guide for gradually adapting the existing test code to the new API.
+This guide is written step-by-step; the idea is to separate the migration into several sets of small changes.
+
+## Remove custom `UncaughtExceptionCaptor`, `DelayController`, and `TestCoroutineScope` implementations
+
+We couldn't find any code that defined new implementations of these interfaces, so they are deprecated. It's likely that
+you don't need to do anything for this section.
```
qwwdfsad, line 10: It means that either this section should be moved to the bottom or the link to the next relevant section should be added

kotlinx-coroutines-test/MIGRATION.md
```diff
@@ -0,0 +1,324 @@
...
+* Accessing `uncaughtExceptions` when the uncaught exceptions are actually expected.
+  In this case, `cleanupTestCoroutines` will fail with an exception that is being caught later.
+  It would be better in this case to use a custom `CoroutineExceptionHandler` so that actual problems that could be
+  found by the cleanup procedure are not superseded by the exceptions that are expected.
+  An example is shown below.
+
+```kotlin
+@Test
+fun testFoo() = runTest {
+    val exceptions = mutableListOf<Throwable>()
```
qwwdfsad, line 83: On the JVM, for the sake of simplicity, both `exceptions` and `customCaptor` can be extracted out of the `runTest`. Though here I may be a bit opinionated

kotlinx-coroutines-test/MIGRATION.md
```diff
@@ -0,0 +1,324 @@
...
+
+### `UncaughtExceptionCaptor`
+
+If the code base has an `UncaughtExceptionCaptor`, its special behavior as opposed to just `CoroutineExceptionHandler`
+was that, at the end of `runBlockingTest` or `cleanupTestCoroutines` (or both), its `cleanupTestCoroutines` procedure
+was called.
+
+We currently don't provide a replacement for this.
+However, `runTest` follows structured concurrency better than `runBlockingTest` did, so exceptions from child coroutines
+are propagated structurally in more cases than before, which makes uncaught exception handlers less useful.
```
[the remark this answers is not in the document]
dkhalanskyjb: I thought that launching a coroutine with a `SupervisorJob` still counts as launching a child coroutine.

kotlinx-coroutines-test/MIGRATION.md
dkhalanskyjb: Not quite, these are listed step-by-step, and this step is the first one to perform.

kotlinx-coroutines-test/MIGRATION.md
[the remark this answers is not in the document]
qwwdfsad: It is (well, at least if one gets it right via `SupervisorJob(coroutineContext.job)`), but I'd rather still phrase it as `so exceptions from child coroutines
are propagated structurally, which makes uncaught exception handlers less useful.`

kotlinx-coroutines-test/README.md
dkhalanskyjb: Probably not. I can still think of some cases where these errors would be hit (like using `GlobalScope.launch`), but then these checks are not reliable.

kotlinx-coroutines-test/MIGRATION.md
dkhalanskyjb: I don't feel one way or the other about this, so changed.

kotlinx-coroutines-test/MIGRATION.md
qwwdfsad: Maybe then it's worth adding a link to a next paragraph?

dkhalanskyjb: Does this idea generalize to adding a link to the next section to every section? If not, I don't get it.

qwwdfsad: Well, it's not; this section is just likely to be skipped by a large portion of readers, thus the suggestion. I do not insist though

dkhalanskyjb:
I needed to make `TestDispatcher` non-sealed again for the build to succeed: the JVM now has its own inheriting class.

qwwdfsad:
Sure

PR #3024, opened 2021-11-16
Promote deprecation levels for release 1.6.0

qwwdfsad:


kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -60,35 +60,6 @@ public fun <T> Flow<T>.catch(action: suspend FlowCollector<T>.(cause: Throwable)
...
- * Cancellation exceptions that were caused by the direct [cancel] call are not handled by this operator.
- *
- * @suppress **Deprecated**: Use `catch { e -> if (predicate(e)) emitAll(fallback) else throw e }`
- */
-@Deprecated(
-    level = DeprecationLevel.ERROR,
-    message = "Use catch { e -> if (predicate(e)) emitAll(fallback) else throw e }",
-    replaceWith = ReplaceWith("catch { e -> if (predicate(e)) emitAll(fallback) else throw e }")
-)
-public fun <T> Flow<T>.onErrorCollect(
```
qwwdfsad, line 84: This one was `@FlowPreview`

qwwdfsad:
`BroadcastChannel<T>.asFlow()` is left as warning deliberately for now

kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -60,35 +60,6 @@ public fun <T> Flow<T>.catch(action: suspend FlowCollector<T>.(cause: Throwable)
...
 
-/**
- * @suppress **Deprecated**: Use `(Throwable) -> Boolean` functional type
- */
-@Deprecated(
-    level = DeprecationLevel.ERROR,
-    message = "Use (Throwable) -> Boolean functional type",
-    replaceWith = ReplaceWith("(Throwable) -> Boolean")
-)
-public typealias ExceptionPredicate = (Throwable) -> Boolean
```
dkhalanskyjb, line 71: Do I understand correctly that type aliases don't leave a trace in target binaries, so it doesn't make sense to make them hidden, as they would simply become useless?

dkhalanskyjb (approved):

qwwdfsad:
>Do I understand correctly that type aliases don't leave a trace in target binaries, so it doesn't make sense to make them hidden, as they would simply become useless?

Right. Also, it used to be `@FlowPreview`

PR #3034, opened 2021-11-19
Revert "JobSupport: non-atomic adding to a Job's list of listeners (#…

qwwdfsad:
…2096)"

This reverts commit 2b1e2765f61fcf830f144651c50479f64d4a8558.


qwwdfsad:
So, the investigation has shown that this change leads to a consistent deadlock of `BlockingCoroutineDispatcherWorkSignallingStressTest`, but not related to the scheduler -- it's the job internals that got stuck in the intermediate state while all the work is scheduler and done correctly.

I'm reverting this commit to avoid further release delays, cc @dkhalanskyjb

PR #3046, opened 2021-11-24
Fix reading an uninitialized value of defaultMainDelayOptIn due to in…

### LINKED ISSUE #3044: defaultMainDelayOptIn is always false
I'm trying to check `delay` is used `Dispatchers.Main` on Android with Coroutines 1.6-RC. However `defaultMainDelayOptIn` is always false.

Below screenshots are debugging on Android Studio.

ss1

ss2

qwwdfsad:
…itialization cycle and add a test that ensures this behaviour in the future

Fixes #3044

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
+++ b/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
@@ -8,9 +8,9 @@ import kotlinx.coroutines.internal.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
 
-internal actual val DefaultDelay: Delay = initializeDefaultDelay()
-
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
```diff
@@ -8,10 +8,10 @@ import kotlinx.coroutines.internal.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
 
-internal actual val DefaultDelay: Delay = initializeDefaultDelay()
-
 private val defaultMainDelayOptIn = systemProp("kotlinx.coroutines.main.delay", true)
 
+internal actual val DefaultDelay: Delay = initializeDefaultDelay()
```
dkhalanskyjb, line 13: *Wow!* TIL. This changes everything. I will not be able to forget this.

Issue #3037, opened 2021-11-23
1.6.0-RC Changelog: FlowCollector is not fun interface


hfhbd:
From changelog: 
https://github.com/Kotlin/kotlinx.coroutines/blob/3574c2feca23c3e8a1ad00b5bf92e2bf04d95060/CHANGES.md?plain=1#L37

Linked PR does not affect `FlowCollector`: #2790

qwwdfsad:
Thanks! It seems like an automation issue, I'm looking into it, the PR didn't make it for some reason

hfhbd:
Fixed in 1.6.0-RC3

PR #3047, opened 2021-11-24
Make FlowCollector fun interface, remove redundant extensions

qwwdfsad:
* It also saves us from copy-without-imports problem

Addresses #3037

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -919,8 +919,6 @@ public final class kotlinx/coroutines/flow/FlowKt {
 	public static final fun catch (Lkotlinx/coroutines/flow/Flow;Lkotlin/jvm/functions/Function3;)Lkotlinx/coroutines/flow/Flow;
 	public static final fun channelFlow (Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/Flow;
 	public static final fun collect (Lkotlinx/coroutines/flow/Flow;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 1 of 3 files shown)


dkhalanskyjb (approved):
Found two implementations of `FlowCollector`.
* `Collect.kt`. That implementation can be replaced with
```kotlin
public suspend inline fun <T> Flow<T>.collectIndexed(crossinline action: suspend (index: Int, value: T) -> Unit) {
    var index = 0
    collect { value ->
        action(checkIndexOverflow(index++), value)
    }
}
```
* `Limit.kt`. Doesn't simplify easily.
* In benchmarks, `FlowPlaysScrabbleOpt`. Simplifies easily.

These simplifications are not obviously better, so it's up to you whether to make them.

qwwdfsad:
I've deliberately avoided touching benchmarks.
Regarding `collectIndexed` -- the proposed implementation is not identical, capturing mutable ints will be boxed on each emission, while spilling it into the state of flow collector explicitly leaves it as `int` type

olme04:
Looks like after this change, coroutines 1.6.0 isn't backward compatible, when `Flow<T>.collect(action)` used in library. Is it expected?

zjuhasz:
This change makes  the documentation confusing in some places.

In `SharedFlow.collect` it says "To emit values from a shared flow into a specific collector ... collect { ... } extension should be used." but collect extension no longer exists.

In `Flow.collect` it says "To ensure the context preservation property, it is not recommended implementing this method directly. Instead, AbstractFlow can be used as the base type to properly ensure flow's properties." but every time a call to collect is made on any Flow we're anonymously implementing collect directly, usually with the SAM conversion.

qwwdfsad:
@zjuhasz thanks for noticing, shared flow documentation is fixed in #3127

> but every time a call to collect is made on any Flow we're anonymously implementing collect directly, usually with the SAM conversion.

`Flow` interface is not anonymously implemented during `collect` SAM-conversion, only `FlowCollector` is, which is perfectly safe

PR #3048, opened 2021-11-25
Fix crashes with the test module on the JVM: take #1

dkhalanskyjb:
Now the test module correctly exposes that initially, the Main dispatcher is missing, so the default delay mechanism is not rewritten.

This is option №1.


qwwdfsad (approved):
This one is safe and sound. Now moving to the second PR

PR #3049, opened 2021-11-25
Make Dispatchers.setMain affect all delays in tests.

dkhalanskyjb:
Option 2.

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -249,6 +249,18 @@ public final class kotlinx/coroutines/DebugKt {
 	public static final field DEBUG_PROPERTY_VALUE_ON Ljava/lang/String;
 }
 
+public final class kotlinx/coroutines/DefaultExecutor : java/lang/Runnable {
+	public static final field INSTANCE Lkotlinx/coroutines/DefaultExecutor;
+	public static final field THREAD_NAME Ljava/lang/String;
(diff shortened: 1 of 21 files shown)


qwwdfsad (commented):
I'm a bit concerned this is way too intrusive change with potentially non-trivial interactions with existing/running code (e.g. now the source of default delay can be changed on the fly).

If you do not have objections, I would prefer sticking with option 1 for the initial release

dkhalanskyjb:
I would also prefer the first option initially, at least until we deal with the initial feedback on both the updated test framework and the change to the default delay. Otherwise, the changes could just be too overwhelming: it's not easy now to diagnose the situation where the default delay is not used for some reason, and the consequences of entering such a situation would become much more significant.

That said, I'd argued that, in code that's heavily using `Dispatchers.Main`, the problem of non-trivial interactions is already just as pronounced, as the only place where the default delay could change is in `setMain`.

dkhalanskyjb:
#3094 contains some examples of when delay-skipping is undesirable in tests.

1zaman:
This might cause similar issues as in #3173, with actual test infrastructure delays being controlled by the test dispatcher. Perhaps this can be addressed by using another delay-handling dispatcher to run delays in the test infrastructure, as [planned](https://github.com/cashapp/turbine/issues/71) by the Turbine library.

Note that as described in #3173, `Dispatchers.setMain` already affects all delays in Robolectric tests, which causes inconsistencies with non-Robolectric tests.

qwwdfsad:
This PR can be now closed, can't it?

dkhalanskyjb:
Sure.

PR #3056, opened 2021-12-01
Fix multiple exception handler invocation in Android

1zaman:
The `AndroidExceptionPreHandler` was changed in #1030 to not invoke the uncaught exception pre-handler on Android Pie and above, and to instead invoke the thread's uncaught exception handler directly on those versions, as Android Pie had added logic in the default uncaught exception handler to ensure that the pre-handler would be invoked.

However, the uncaught exception handler is already being invoked for all unhandled coroutine exceptions, so this change in the `AndroidExceptionPreHandler` caused the uncaught exception handler to be invoked twice in Android Pie and above.

Since the default uncaught exception handler implementation in Android kills the app immediately, this normally wouldn't make a difference. However, if an app sets a custom uncaught exception handler (either at the thread or global level) that doesn't always kill the app, then the duplicate invocations might cause undesired or wrong behaviour.

This is now fixed by removing the uncaught exception handler invocation from `AndroidExceptionPreHandler`.

Since the pre-handler was introduced in Android Oreo, I have also added a minimum SDK check for Android Oreo as well. Also edited the documentation in the comments to add more details and context.

### DIFF
--- a/ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt
+++ b/ui/kotlinx-coroutines-android/src/AndroidExceptionPreHandler.kt
@@ -34,20 +34,21 @@ internal class AndroidExceptionPreHandler :
 
     override fun handleException(context: CoroutineContext, exception: Throwable) {
         /*
-         * If we are on old SDK, then use Android's `Thread.getUncaughtExceptionPreHandler()` that ensures that
-         * an exception is logged before crashing the application.
+         * Android Oreo introduced private API for a global pre-handler for uncaught exceptions, to ensure that the
+         * exceptions are logged even if the default uncaught exception handler is replaced by the app. The pre-handler
+         * is invoked from the Thread's private dispatchUncaughtException() method, so our manual invocation of the
+         * Thread's uncaught exception handler bypasses the pre-handler in Android Oreo, and uncaught coroutine
+         * exceptions are not logged. This issue was addressed in Android Pie, which added a check in the default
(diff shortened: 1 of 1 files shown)


qwwdfsad (approved):
Thanks!

For the record, verified this fix on AS simulators with various API versions

1zaman:
Note that if a custom uncaught exception handler is set that doesn't call through to the default handler, then the pre-handler (and logging) would still be bypassed, since the additional checking in Android Pie was done only on the default handler implementation. As opposed to standard non-coroutine exceptions that are handled via the private `dispatchUncaughtException()` method, which _does_ invoke the pre-handler, even if a custom uncaught exception handler was set (which was the whole point behind introducing the pre-handler in the first place).

I think this is fine, since it works in the default use-case (without a custom exception handler), and it's probably preferable to avoid using unsupported API as much as possible, especially considering the warnings added in Android Pie, as shown in #822. Most Android apps will not bypass the default uncaught exception handler, and those that do can just add logging themselves if they're handling uncaught coroutine exceptions. Though it would potentially result in double logging if the exception can be thrown from outside of a coroutine context as well (and thus trigger the pre-handler), but that probably isn't a big deal.

Long-term, the special handling for Android can probably be removed completely, considering that it only affects Android Oreo (which is somewhat old at this point), and only developers for the most part (as it only impacts local logging). Those people who still use Android Oreo devices for development after a certain point can be asked to add the patch locally.

PR #3059, opened 2021-12-04
Update coroutine-context-and-dispatchers.md

parasdhawan17:
Fixed the order of output.


qwwdfsad:
The order is non-deterministic, though the current one is much more likely to happen.

We prefer not to explain concurrency in first chapters and touch this topic later

Issue #3031, opened 2021-11-18
NPE during StackTraceRecovery exception creation


gildor:
We got crashes on prod caused by StackTraceRecoveryKt. Not very common (a few cases per day).
Don't have any reproduction steps. 

```
Fatal Exception: kotlinx.coroutines.CoroutinesInternalError: Fatal exception in coroutines machinery for CancellableContinuation(DispatchedContinuation[Dispatchers.IO, Continuation at kotlinx.coroutines.flow.SharedFlowImpl.collect(SharedFlow.kt:348)@37836a93]){Cancelled}@2e1f5ad0. Please read KDoc to 'handleFatalException' method and report this incident to maintainers
       at kotlinx.coroutines.DispatchedTask.handleFatalException(DispatchedTask.kt:144)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:115)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.lang.StackTraceElement[].clone()' on a null object reference
       at kotlinx.coroutines.internal.StackTraceRecoveryKt.createFinalException(StackTraceRecovery.kt:107)
       at kotlinx.coroutines.internal.StackTraceRecoveryKt.recoverFromStackFrame(StackTraceRecovery.kt:78)
       at kotlinx.coroutines.internal.StackTraceRecoveryKt.access$recoverFromStackFrame(StackTraceRecovery.kt:1)
       at kotlinx.coroutines.CancellableContinuationImpl.getExceptionalResult$kotlinx_coroutines_core(CancellableContinuationImpl.kt:636)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:91)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
[...]

Issue #2367, opened 2020-11-04
Promote CopyableThrowable to stable


qwwdfsad:
The main benefit of making it stable is to enable other libraries that depend on `kotlinx.coroutines` leverage stacktrace.
E.g. retrofit is not ready to do so unless `CopyableThrowable` is stable: https://github.com/square/retrofit/issues/3474

We should evaluate potential enhancements and interactions with (also potential) global exception registry for exceptions beyond user's control and promote it to stable

dmarin:
will this be stable on 1.5.0?

qwwdfsad:
No such plans for 1.5.0 unfortunately

almozavr:
Retrofit is the de factor standart library for http in android world, and due to this issue it's really hard to read API error reports without hacky solutions, they are almost useless, it really hurts 😢

LouisCAD:
What's blocking this becoming stable? API design concerns? Multiplatform support? Something else?

dkhalanskyjb:
Hi! We decided to move this API to the standard library: https://github.com/Kotlin/KEEP/discussions/483 Please join the discussion, it's a good time to propose your changes to the `CopyableThrowable` API before it becomes stable (in the stdlib).

PR #3063, opened 2021-12-06
Validate that throwing tryCopy does not crash with an internal error

qwwdfsad:
Motivated by #3031


qwwdfsad:
I do not propose to make this behaviour final for #2367 or #2607, but at least it is worth sealing with a test and address later

dkhalanskyjb (approved):

PR #3071, opened 2021-12-08
Make UncompletedTestCoroutines more verbose

### LINKED ISSUE #3066: `runTest` doesn't report test failure if a non-cancellable coroutine happens to be hanging
I'm using `kotlinx-coroutines-test` 1.6.0-RC.

If some non-cancellable coroutine code is hanging, exceptions or assertion errors from the test are never reported despite being reached:

```kotlin
@Test
fun testFailureWithPendingCoroutine() = runTest(dispatchTimeoutMs = 5000) {
    launch {
        withContext(NonCancellable) {
            awaitCancellation() // hanging non-cancellable coroutine
        }
    }
    yield()
    fail("some test failure") // this line is reached and throws, but it's never reported
}
```

I could see with debugging that the `fail()` is indeed reached and called. It does fail, but the assertion error is just swallowed somewhere in `runTest` and never makes it to the output. I just get the error message about the hanging coroutine:

```
The test coroutine was not completed after waiting for 5000 ms
kotlinx.coroutines.test.UncompletedCoroutinesError: The test coroutine was not completed after waiting for 5000 ms
	at app//kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTestCoroutine$3$3.invokeSuspend(TestBuilders.kt:231)
	(Coroutine boundary)
	at kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt.runTestCoroutine(TestBuilders.kt:257)
	at kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTest$1$1.invokeSuspend(TestBuilders.kt:167)
	at kotlinx.coroutines.test.TestBuildersJvmKt$createTestResult$1.invokeSuspend(TestBuildersJvm.kt:13)
Caused by: kotlinx.coroutines.test.UncompletedCoroutinesError: The test coroutine was not completed after waiting for 5000 ms
	at app//kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTestCoroutine$3$3.invokeSuspend(TestBuilders.kt:231)
	at app//kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        ...
```

Moreover, the test spends the whole `dispatchTimeoutMs` before failing (this is why I reduced it to 5 seconds in the repro). It is somewhat understandable because of the non-cancellability of the coroutine, but it's quite annoying too. Not sure we could do anything about that though, at least the reporting of test failure in addition to the timeout would be great.

### LINKED ISSUE #3069: Feature request: different error messages for hanging test body vs launched coroutines
From my experience testing a coroutine-based library, with tons of fine-tuned suspension assertions, it has often been important to me to distinguish cases where the test body itself hangs from cases where some coroutines (spawned from deep in the code) are hanging.

When running tests locally it's easy to know by debugging tests and putting a breakpoint at the end of the test method, but when running on CI this is not an option. Sometimes I can't even reproduce test failures locally (because I'm developing on Windows, and the failing test happens in iOS test code on MacOS CI agent).

It would be nice if this error message could make the distinction too:

```
The test coroutine was not completed after waiting for 5000 ms
kotlinx.coroutines.test.UncompletedCoroutinesError: The test coroutine was not completed after waiting for 5000 ms
	at app//kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt$runTestCoroutine$3$3.invokeSuspend(TestBuilders.kt:231)
	(Coroutine boundary)
	at kotlinx.coroutines.test.TestBuildersKt__TestBuildersKt.runTestCoroutine(TestBuilders.kt:257)
```

Currently I get the same error for both of these test cases:

```kotlin
@Test
fun testTimeoutWithHangingBody() = runTest(dispatchTimeoutMs = 5000) {
    awaitCancellation() // hanging test body
}
    
@Test
fun testTimeoutWithHangingCoroutine() = runTest(dispatchTimeoutMs = 5000) {
    launch {
        awaitCancellation() // hanging launched coroutine
    }
}
```

dkhalanskyjb:
Fixes #3066
Fixes #3069

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
+++ b/kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
@@ -40,6 +40,7 @@ internal expect suspend inline fun recoverAndThrow(exception: Throwable): Nothin
  * The opposite of [recoverStackTrace].
  * It is guaranteed that `unwrap(recoverStackTrace(e)) === e`
  */
+@PublishedApi
 internal expect fun <E: Throwable> unwrap(exception: E): E
 
 internal expect class StackTraceElement
--- a/kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
+++ b/kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
@@ -40,7 +40,7 @@ internal expect suspend inline fun recoverAndThrow(exception: Throwable): Nothin
  * The opposite of [recoverStackTrace].
  * It is guaranteed that `unwrap(recoverStackTrace(e)) === e`
  */
-@PublishedApi
+@PublishedApi // only published for the multiplatform tests in our own code
 internal expect fun <E: Throwable> unwrap(exception: E): E
 
 internal expect class StackTraceElement
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -161,10 +161,11 @@ public fun runTest(
 public fun TestScope.runTest(
     dispatchTimeoutMs: Long = DEFAULT_DISPATCH_TIMEOUT_MS,
     testBody: suspend TestScope.() -> Unit
-): TestResult = asSpecificImplementation().let {
-    it.enter()
+): TestResult = with(asSpecificImplementation()) {
+    enter()
+    start(CoroutineStart.UNDISPATCHED, this, testBody)
     createTestResult {
-        runTestCoroutine(it, dispatchTimeoutMs, TestScopeImpl::tryGetCompletionCause, testBody) { it.leave() }
+        runTestCoroutine(this, dispatchTimeoutMs, { tryGetCompletionCause() }) { leave() }
     }
 }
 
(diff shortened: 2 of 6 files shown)


kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
```diff
@@ -40,6 +40,7 @@ internal expect suspend inline fun recoverAndThrow(exception: Throwable): Nothin
  * The opposite of [recoverStackTrace].
  * It is guaranteed that `unwrap(recoverStackTrace(e)) === e`
  */
+@PublishedApi
```
qwwdfsad, line 43: Please leave a comment that it is only used from other module tests, so when the time comes we know it can be safely deleted/reverted

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -228,7 +229,26 @@ internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
                     // we expect these and will instead throw a more informative exception just below.
                     emptyList()
                 }.throwAll()
-                throw UncompletedCoroutinesError("The test coroutine was not completed after waiting for $dispatchTimeoutMs ms")
+                var completing: Boolean
+                val completionCause = try {
+                    coroutine.tryGetCompletionCause().also { completing = true }
+                } catch (e: Throwable) {
+                    completing = false
```
qwwdfsad, line 236: instead of `try-catch`, it should be enough to check `coroutine.isCancelled`.
The same goes for `completing`

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -228,7 +229,26 @@ internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
                     // we expect these and will instead throw a more informative exception just below.
                     emptyList()
                 }.throwAll()
-                throw UncompletedCoroutinesError("The test coroutine was not completed after waiting for $dispatchTimeoutMs ms")
+                var completing: Boolean
+                val completionCause = try {
+                    coroutine.tryGetCompletionCause().also { completing = true }
```
qwwdfsad, line 234:
```suggestion
                    // Protected JobSupport.completionCause lifted in TestCoroutineScope and TestScope
                    coroutine.tryGetCompletionCause().also { completing = true }
```

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -228,7 +229,26 @@ internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
...
+                } catch (e: Throwable) {
+                    completing = false
+                    null
+                }
+                var message = "After waiting for $dispatchTimeoutMs ms"
```
qwwdfsad, line 239: It would be nice to follow our overall approach and extract slow-path to a separate method

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -196,6 +196,7 @@ internal const val DEFAULT_DISPATCH_TIMEOUT_MS = 60_000L
 internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
```
qwwdfsad, line 196: Do we still need generic parameter here instead of `AbstractCoroutine<Unit>`?

qwwdfsad (commented):
Overall looks good

kotlinx-coroutines-test/common/src/TestBuilders.kt
dkhalanskyjb: There's still the `TestScope`/`TestCoroutineScope` difference.

kotlinx-coroutines-test/common/src/TestBuilders.kt
qwwdfsad: Agree, then maybe it's worth extracting the whole onTimeout body?

qwwdfsad (approved):

PR #3073, opened 2021-12-08
Display artificial stack frames as calls to functions instead of text

dkhalanskyjb:
Originally implemented in #2461. I just cherry-picked it and made sure the project builds.

Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/2291


qwwdfsad (commented):
Could you please elaborate what is the latest IDEA version that do not support this change and when we've dropped the support of such version?

dkhalanskyjb:
This change is supported since 212. The support for 211 stopped with 2021.1.3 coming out, which was this summer: https://blog.jetbrains.com/idea/2021/06/intellij-idea-2021-1-3/

qwwdfsad:
Could you please figure out whether we are allowed to merge this change?

dkhalanskyjb:
The last Kotlin Plugin version to support IDEA 211 was 1.6.21. So, I think we're good to go.

qwwdfsad (approved):
LGTM.

A few obligatory things prior to merging:

* Please check that coroutines debugger in IDE works with this change
* Please link here the issue in KTIJ that supports this new behaviour, so we can track the change back to its roots if there is need to

qwwdfsad:
>I don't think such an issue was ever filed.

That's a bummer, could you please ask Nikita to do so and to link the commit with the change in IDEA to it?

dkhalanskyjb:
Here's the older comment: https://github.com/Kotlin/kotlinx.coroutines/issues/2291#issuecomment-1279142363 No harm in repeating this, I think.

PR #3079, opened 2021-12-13
Update Node version to 10.0.0 because Promise.finally since this version

ilgonmic:
To make work [KT-49738](https://youtrack.jetbrains.com/issue/KT-49738) with async `@AfterTest`, necessary to upgrade Node.JS version, because `Promise.finally` exists since 10.0.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -35,7 +35,7 @@ baksmali_version=2.2.7
 # JS
 kotlin.js.compiler=both
 gradle_node_version=1.2.0
-node_version=8.9.3
+node_version=10.0.0
 npm_version=5.7.1
 mocha_version=6.2.2
 mocha_headless_chrome_version=1.8.2


qwwdfsad (approved):

dkhalanskyjb:
Just curious: looking at the description of the method https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally#description, maybe `.then(onFinally, onFinally)` would suffice and allow to target more versions?

ilgonmic:
Yes, they are similar, but I wanted to get rid of duplicated callbacks.
Apart from that, now Kotlin/JS targeting 16.13.x, because of M1 support and ES modules, which are planned to be supported in Kotlin/JS.
I don't think that we need to target earlier than 10 version, because it is not supported more (https://nodejs.org/en/about/releases/)

PR #3082, opened 2021-12-14
Get rid of @InternalCoroutinesApi from Flow.collect member

### LINKED ISSUE #3078: FlowCollector fun interface errors due to collect being InternalCoroutinesApi
As per the title the change in 1.6 RC2 breaks code that are copy pasted from Compose for example:

```
@Composable
fun <T : R, R> Flow<T>.collectAsState(
    initial: R,
    context: CoroutineContext = EmptyCoroutineContext
): State<R> = produceState(initial, this, context) {
    if (context == EmptyCoroutineContext) {
        collect { value = it }
    } else withContext(context) {
        collect { value = it }
    }
}
```

Will produce `This is an internal kotlinx.coroutines API that should not be used from outside of kotlinx.coroutines. No compatibility guarantees are provided. It is recommended to report your use-case of internal API to kotlinx.coroutines issue tracker, so stable API could be provided instead`

qwwdfsad:
Fixes #3078

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Flow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Flow.kt
@@ -172,7 +172,7 @@ import kotlin.coroutines.*
  *
  * **The `Flow` interface is not stable for inheritance in 3rd party libraries**, as new methods
  * might be added to this interface in the future, but is stable for use.
- * Use the `flow { ... }` builder function to create an implementation or extend [AbstractFlow].
+ * Use the `flow { ... }` builder function to create an implementation, or extend [AbstractFlow].
  */
 public interface Flow<out T> {
     /**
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -172,7 +172,7 @@ import kotlin.coroutines.*
  *
  * **The `Flow` interface is not stable for inheritance in 3rd party libraries**, as new methods
  * might be added to this interface in the future, but is stable for use.
- * Use the `flow { ... }` builder function to create an implementation.
+ * Use the `flow { ... }` builder function to create an implementation or extend [AbstractFlow].
```
dkhalanskyjb, line 175: This parses as "Use the `flow { ... }` builder function to (create an implementation or extend [AbstractFlow])". I guess you probably meant
```suggestion
 * Use the `flow { ... }` builder function to create an implementation, or extend [AbstractFlow].
```
```suggestion
 * Use the `flow { ... }` builder function or extend [AbstractFlow] to create an implementation.
```

dkhalanskyjb (approved):
I checked with `grep -B2`, and it looks like there are no more `fun collect` that are `@InternalCoroutinesApi`.

PR #3087, opened 2021-12-15
Add TimeSource implementation to TestCoroutineScheduler

### LINKED ISSUE #3086: TestCoroutineScheduler: Make it a TimeSource
Already todo, and is very useful to avoid manually increasing `TestTimeSource`
Use case: 
testing timebased caching, timed retries, windowing, date time apis (using `Clock`)
```kt
class TimeCache(val timeSource: TimeSource = TimeSource.Monotonic, val valid: Duration = 1.seconds) {
    private var counter = 0
    private var cache: Data? = null

    suspend fun get(): Data {
        if (cache != null && cache!!.requestAt.elapsedNow() < valid) {
            return cache!!
        }
        cache = null
        return requestServer().also {
            cache = it
        }
    }

    private suspend fun requestServer(): Data {
        delay(3.seconds)
        return Data(counter.also { counter += 1 }, timeSource.markNow())
    }
}

data class Data(val data: Int, val requestAt: TimeMark)

@Test
fun blockingImprecise() = runBlocking {
    val cache = TimeCache()
    assertEquals(0, cache.get().data)
    delay(2.seconds)
    assertEquals(1, cache.get().data)
    delay(0.5.seconds)
    assertEquals(1, cache.get().data)
}

@Test
fun blockingManuallyPrecise() = runBlocking {
    val ts = TestTimeSource()
    val cache = TimeCache(ts)
    assertEquals(0, cache.get().data)
    delay(2.seconds)
    ts += 2.seconds
    assertEquals(1, cache.get().data)
    delay(0.5.seconds)
    ts += 0.5.seconds
    assertEquals(1, cache.get().data)
}

@Test
fun skippingManuallyPrecise() = runTest {
    val ts = TestTimeSource()
    val cache = TimeCache(ts)
    assertEquals(0, cache.get().data)
    delay(2.seconds)
    ts += 2.seconds
    assertEquals(1, cache.get().data)
    delay(0.5.seconds)
    ts += 0.5.seconds
    assertEquals(1, cache.get().data)
}

@Test
fun expectedPrecise() = runTest {
    val cache = TimeCache(timeSource)
    assertEquals(0, cache.get().data)
    delay(2.seconds)
    assertEquals(1, cache.get().data)
    delay(0.5.seconds)
    assertEquals(1, cache.get().data)
}

private val TestScope.timeSource: TimeSource get() = object : AbstractLongTimeSource(DurationUnit.MILLISECONDS) {
    override fun read(): Long = currentTime
}
```

hfhbd:
Fixes #3086 

I didn't inherit `TestCoroutineScheduler` from `TimeSource`, because you rarely use `TestCoroutineScheduler.markNow()` but rather use its `timeSource` to inject it in your time based components.
Also the method `elapsedNow` from `TimeSource` is quite confusing in a `Scheduler`. What happened when calling `elapsedNow` with its job? (Nothing...)

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
+++ b/kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
@@ -12,6 +12,7 @@ import kotlinx.coroutines.internal.*
 import kotlinx.coroutines.selects.*
 import kotlin.coroutines.*
 import kotlin.jvm.*
+import kotlin.time.*
 
 /**
  * This is a scheduler for coroutines used in tests, providing the delay-skipping behavior.
@@ -26,7 +27,6 @@ import kotlin.jvm.*
  * haven't yet been dispatched (via [runCurrent]).
  */
 @ExperimentalCoroutinesApi
-// TODO: maybe make this a `TimeSource`?
 public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCoroutineScheduler),
     CoroutineContext.Element {
 
@@ -43,7 +43,7 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
     /** This counter establishes some order on the events that happen at the same virtual time. */
     private val count = atomic(0L)
 
-    /** The current virtual time. */
+    /** The current virtual time in milliseconds. */
     @ExperimentalCoroutinesApi
     public var currentTime: Long = 0
         get() = synchronized(lock) { field }
@@ -193,6 +193,15 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
      * Consumes the knowledge that a dispatch event happened recently.
      */
     internal val onDispatchEvent: SelectClause1<Unit> get() = dispatchEvents.onReceive
+
+    /**
+     * Returns the [TimeSource] representation of the virtual time of this scheduler.
+     */
+    @ExperimentalCoroutinesApi
+    @ExperimentalTime
+    public val timeSource: TimeSource = object : AbstractLongTimeSource(DurationUnit.MILLISECONDS) {
+        override fun read(): Long = currentTime
+    }
 }
 
 // Some error-throwing functions for pretty stack traces
--- a/kotlinx-coroutines-test/common/src/TestScope.kt
+++ b/kotlinx-coroutines-test/common/src/TestScope.kt
@@ -7,6 +7,7 @@ package kotlinx.coroutines.test
 import kotlinx.coroutines.*
 import kotlinx.coroutines.internal.*
 import kotlin.coroutines.*
+import kotlin.time.*
 
 /**
  * A coroutine scope that for launching test coroutines.
@@ -83,6 +84,14 @@ public fun TestScope.runCurrent(): Unit = testScheduler.runCurrent()
  */
 @ExperimentalCoroutinesApi
 public fun TestScope.advanceTimeBy(delayTimeMillis: Long): Unit = testScheduler.advanceTimeBy(delayTimeMillis)
+
+/**
+ * The [test scheduler][TestScope.testScheduler] as a [TimeSource].
+ * @see TestCoroutineScheduler.timeSource
+ */
+@ExperimentalCoroutinesApi
+@ExperimentalTime
+public val TestScope.testTimeSource: TimeSource get() = testScheduler.timeSource
 
 /**
  * Creates a [TestScope].
--- a/kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
+++ b/kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
@@ -65,6 +65,7 @@ public final class kotlinx/coroutines/test/TestCoroutineScheduler : kotlin/corou
 	public final fun advanceTimeBy (J)V
 	public final fun advanceUntilIdle ()V
 	public final fun getCurrentTime ()J
+	public final fun getTimeSource ()Lkotlin/time/TimeSource;
 	public final fun runCurrent ()V
 }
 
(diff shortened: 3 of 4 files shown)


dkhalanskyjb (changes requested):
Nicely implemented! It was a `TODO` not because it was bothersome, but because we didn't want to decide immediately whether to deprecate `currentTime` in favor of manipulating time marks. On one hand, `measureTime`, `markTime`, etc., are more structured than manipulating `currentTime`'s `Long` values, but on the other hand, `TimeSource` is still an experimental API, and also, a lot of tests explicitly compare `currentTime` with specific `Long` values.

However, your approach provides the functionality without railroading us into some particular API. If we decide after all to make `TestCoroutineScheduler` itself a `TimeSource`, we can just easily deprecate `timeSource` in favor of `this` and be done.

kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
```diff
@@ -193,6 +193,18 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
...
     internal val onDispatchEvent: SelectClause1<Unit> get() = dispatchEvents.onReceive
+
+    /**
+     * Returns a [TimeSource] synced with this [currentTime] to allow time based consumers handling this delay-skipping
+     * behavior.
+     */
+    @ExperimentalTime
+    public val timeSource: TimeSource = VirtualTimeSource()
+
+    @ExperimentalTime
```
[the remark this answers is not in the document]
hfhbd: It is needed

kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
```diff
@@ -193,6 +193,18 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
...
+    private inner class VirtualTimeSource: AbstractLongTimeSource(DurationUnit.MILLISECONDS) {
```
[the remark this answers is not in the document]
hfhbd: simply changing it to an object does not work, because the class and the object are unrelated, so it is not possible to access `currentTime` (and inner objects are not available). An anonymous class/object in the getter does work, but would allocate a new instance every call, would it?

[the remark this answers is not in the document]
hfhbd: Oh yeah, that's true :D Changed

dkhalanskyjb (approved):
LGTM. @qwwdfsad, what do you think about these changes?

hfhbd:
Loosely related: https://github.com/Kotlin/kotlinx-datetime/pull/164 to handle this new delay skipping behavior when using `datetime.Clock` as well

kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
```diff
@@ -193,6 +193,14 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
      * Consumes the knowledge that a dispatch event happened recently.
      */
     internal val onDispatchEvent: SelectClause1<Unit> get() = dispatchEvents.onReceive
+
+    /**
+     * Returns the [TimeSource] representation of the virtual time of this scheduler.
+     */
+    @ExperimentalTime
```
qwwdfsad, line 200: Maybe it is worth marking it as `ExperimentalCoroutinesApi` as well in case we'll decide to implement `TimeSource` interface when it's stable?

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -84,6 +85,13 @@ public fun TestScope.runCurrent(): Unit = testScheduler.runCurrent()
 @ExperimentalCoroutinesApi
 public fun TestScope.advanceTimeBy(delayTimeMillis: Long): Unit = testScheduler.advanceTimeBy(delayTimeMillis)
 
+/**
+ * The [test scheduler][TestScope.testScheduler] as a [TimeSource].
+ * @see TestCoroutineScheduler.timeSource
+ */
+@ExperimentalTime
+public val TestScope.testTimeSource: TimeSource get() = testScheduler.timeSource
```
qwwdfsad, line 93: Shouldn't it be `timeSource` or `schedulerTimeSource`?
Also lacks `ExperimentalCoroutinesApi`

qwwdfsad (commented):
Conceptually, I have no objections. Thanks for the listed use-case in the issue!

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: It was initially `timeSource`, but I advised against it because seeing just a bare `timeSource` in the middle of a test will be disheartening, I assume. `schedulerTimeSource` is not much better than `testScheduler.timeSource`, so I think it's either `testTimeSource` or no extension on `TestScope` at all.

kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
dkhalanskyjb: Good point!

hfhbd: Sure

kotlinx-coroutines-test/common/src/TestScope.kt
hfhbd: Added annotation as well

dkhalanskyjb (approved):
Thanks!

dkhalanskyjb:
Merged manually: https://github.com/Kotlin/kotlinx.coroutines/commit/d5a66e0adc747f1748dc3692fc9e7f8580d22f4f

PR #3088, opened 2021-12-15
Updated SLF4J Dependencies In SLF4J Integration Module To 1.7.32

severn-everett:
The SLF4J integration module now also no longer uses the deprecated Gradle dependency declaration keywords


qwwdfsad (approved):
Thanks! I'll merge it after 1.6.0

PR #3090, opened 2021-12-15
Fix R8 warnings

### LINKED ISSUE #3058: Upgrading from 1.5.2 to 1.6.0-RC in Android project breaks R8 obfuscation
Build fails because of missing classes.

It is fixable by adding:
```
-dontwarn java.lang.ClassValue # Needed by kotlinx.coroutines 1.6.0-RC
-dontwarn org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement # Needed by kotlinx.coroutines 1.6.0-RC
```
to `proguard-rules.pro`.

I though it was worth reporting this because no specific R8 pro guard config was needed with 1.5.2.

dkhalanskyjb:
Fixes #3058

Tried to get creative here but to no avail. Looks like just adding the new rules is the way to go.
I had my doubts about `dontwarn` for `IgnoreJRERequirements`, but, searching the net, it looks like this rule is common practice, and given that only methods and internal and private classes are marked with it, everything should be fine.


qwwdfsad (approved):
👍

PR #3092, opened 2021-12-16
Created integration module for Kotlin Logging

severn-everett:
I'm not sure how I am supposed to populate the `package.list` file, so feedback on that would be welcome.


qwwdfsad (commented):
Unfortunately, we are not ready to accept a new module and further maintain it.

I would suggest discussing such module as part of the kotlin-logging with its maintainer

PR #3093, opened 2021-12-16
Introduce new `CancellableContinuation.resume` with `onCancellation` lambda that takes the resumption value

ndkoval:
Introduce new `CancellableContinuation.resume` with `onCancellation` lambda that takes the resumption value, cause, and the coroutine context as arguments


ndkoval:
Some Lincheck tests fail as they require a small integration fix in Lincheck. 
I have already tested the fix locally, but cannot release it right now, as we need to agree on the `OnCancellationHandler` name first.

qwwdfsad:
Superseded by Dmitry's work

PR #3101, opened 2021-12-20
Fix Dispatchers.Main not being fully initialized on Android and Swing

dkhalanskyjb:
* If `unitTests.returnDefaultValues=true` is set on Android, then
`Looper.getMainLooper()` may return `null`. The type system of
Kotlin is tricked to believe that the method can't ever return
`null`, so doesn't check for it anywhere. As a result, despite not
being fully initialized, `Dispatchers.Main` is considered
non-missing.
* When running in a headless environment, Swing crashes, but
the dispatcher successfully initializes nonetheless.

This was not an issue before, as it only surfaced when
`Dispatchers.Main` was used. However, now `Main` is the source of
time for delays, so any delay will throw something
incomprehensible if this happens.

### DIFF
--- a/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
+++ b/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
@@ -51,9 +51,9 @@ public sealed class HandlerDispatcher : MainCoroutineDispatcher(), Delay {
 
 internal class AndroidDispatcherFactory : MainDispatcherFactory {
 
-    override fun createDispatcher(allFactories: List<MainDispatcherFactory>) {
+    override fun createDispatcher(allFactories: List<MainDispatcherFactory>): MainCoroutineDispatcher {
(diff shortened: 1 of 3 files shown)


ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -52,7 +52,7 @@ public sealed class HandlerDispatcher : MainCoroutineDispatcher(), Delay {
 internal class AndroidDispatcherFactory : MainDispatcherFactory {
 
     override fun createDispatcher(allFactories: List<MainDispatcherFactory>) =
-        HandlerContext(Looper.getMainLooper().asHandler(async = true))
+        HandlerContext(Looper.getMainLooper()!!.asHandler(async = true))
```
qwwdfsad, line 55: Please leave a comment in the code why it is necessary

qwwdfsad (approved):

qwwdfsad (approved):

PR #3104, opened 2021-12-21
Restore MPP backwards compatibility on K/JS and K/N

qwwdfsad:
See https://github.com/rsocket/rsocket-kotlin/issues/201

Seems like it can be quite a disrupting change otherwise


dkhalanskyjb (commented):
https://github.com/Kotlin/kotlinx.coroutines/pull/3047 also removes this on `SharedFlow`, which is more obscure than this one, but still, something to consider.

qwwdfsad:
`SharedFlow` overload was added in `1.6.0-RC`, previous versions are not affected

PR #3130, opened 2022-01-11
Make multithreadingSupported a property with a getter

ilmat192:
This change fixes an issue with initialization order observed when
lazy initialization is disabled (-Xir-property-lazy-initialization=
disabled compiler flag) and the new MM is used.

The problem is the following: the initialization of DefaultDelay
happens before initialization of multithreadingSupported. Thus
the initialization of DefaultDelay gets an uninitialized value
of multithreadingSupported and behaves as if the old MM is used.

Related issue: #KT-50491.

### DIFF
--- a/kotlinx-coroutines-core/native/src/internal/Concurrent.kt
+++ b/kotlinx-coroutines-core/native/src/internal/Concurrent.kt
@@ -32,6 +32,6 @@ internal open class SuppressSupportingThrowableImpl : Throwable() {
     }
 }
 
-@SharedImmutable
 @OptIn(ExperimentalStdlibApi::class)
-internal val multithreadingSupported: Boolean = kotlin.native.isExperimentalMM()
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/native/src/internal/Concurrent.kt
```diff
@@ -32,6 +32,6 @@ internal open class SuppressSupportingThrowableImpl : Throwable() {
     }
 }
 
-@SharedImmutable
 @OptIn(ExperimentalStdlibApi::class)
```
qwwdfsad, line 35:
```suggestion
// getter instead of a property due to the bug in the initialization dependencies tracking with '-Xir-property-lazy-initialization=disabled' that Ktor uses 
@OptIn(ExperimentalStdlibApi::class)
```

PR #3131, opened 2022-01-11
Rollback #2972, but leave a compatibility option with 1.6.0

### LINKED ISSUE #3106: Problem with delay with kotlinx-coroutines-javafx
**versions**:
- java - 17.0.1
- javafx - 17.0.1
- kotlin-stdlib-jdk8 - 1.6.10
- kotlinx-coroutines-core - 1.6.0
- kotlinx-coroutines-javafx - 1.6.0

**smallest example of bug**
```kotlin
import kotlinx.coroutines.withContext

fun main() {
    val jobCoroutineScope = CoroutineScope(Dispatchers.IO)
    println("started")
    for (job in 1..3) {
        jobCoroutineScope.launch {
            println("job[$job] started")
            for (step in 1..10) {
                  delay(100)
            }
            println("job[$job] ended")
        }
    }
    Thread.sleep(2000)
    println("ended")
}
```
**BUG:** Some jobs stay blocked forever in `delay` function. 

I've been trying to figure out what the problem is. And it is because the delay is redirected to `kotlinx.coroutines.javafx.JavaFxDispatcher#schedule`, where the delay is implemented as a `TimeLine` for JavaFx. However, this Timeline is started outside of the `JavaFX Thread` and this will lead to unexpected results.

**Workarounds:**
- wrap delay as `withContext(Dispatchers.Main) { delay(100) }` - Dispatchers.Main is JavaFx thread in this case.
- add SysProp `kotlinx.coroutines.main.delay=false` - switch off redirecting to JavaFxDispatcher (Dispatchers.Main)
- remove `kotlinx-coroutines-javafx` dependency (not possible for me)

### LINKED ISSUE #3113: Coroutine 1.6.0 ANR on Android 11
Took me a while to figure out but it seems the changes to `delay` to use MainDispatcher have a side effect on Android 11 (And only Android 11).

I have yet to build a repro but posting before in case there's something obvious.

In an activity in `onCreate` I have a simple runBlocking (so on main thread ui) for a call that is nearly instant.

That call use a workerPool on dispatchers.IO and at some point have a simple :

```
private suspend fun isLoaded(): Boolean {
        return loaded ?: run {
            do {
                delay(1)
            } while (loaded == null)
            loaded
        } ?: false
    }
```

This works well on first start, but if the app is killed by OS or you swipe the app from the drawer (but don't use the kill) then further start will ANR with below trace.

ANR:
```
 at sun.misc.Unsafe.park (Native method)
  at java.util.concurrent.locks.LockSupport.parkNanos (LockSupport.java:230)
  at kotlinx.coroutines.BlockingCoroutine.joinBlocking (BlockingCoroutine.java:88)
  at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking (BuildersKt__Builders.kt:59)
  at kotlinx.coroutines.BuildersKt.runBlocking (Builders.kt:1)
  at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default (BuildersKt__Builders.kt:38)
  at kotlinx.coroutines.BuildersKt.runBlocking$default (Builders.kt:1)
  at xxxxx.onCreate (xxxx.kt:127)
  at android.app.Activity.performCreate (Activity.java:8215)
  at android.app.Activity.performCreate (Activity.java:8199)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1309)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3824)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:4027)
  at android.app.servertransaction.LaunchActivityItem.execute (LaunchActivityItem.java:85)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2336)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:247)
  at android.app.ActivityThread.main (ActivityThread.java:8676)
  at java.lang.reflect.Method.invoke (Native method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:602)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1130)
```

Since it's very specific to a version of Android I'm not sure there's something to be done at coroutines levels, and I'll try to increase the delay to see, but reporting in case the delay(1) with runBlocking on mainthread can trigger issues in other unseen cases.

qwwdfsad:
The approach from 1.6.0 has proven itself as unstable and multiple hard-to-understand bugs have been reported:

* JavaFx timer doesn't really work outside the main thread
* The frequent initialization pattern "runBlocking { doSomethingThatMayCallDelay() }" used on the main thread during startup now silently deadlocks
* The latter issue was reported both by Android and internal JB Compose users
* The provided workaround with system property completely switches off the desired behaviour that e.g. Compose may rely on, potentially introducing new sources of invalid behaviour

The original benefits does not outweigh these pitfalls, so the decision is to revert this changes in the minor release

Fixes #3113
Fixes #3106

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
+++ b/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
@@ -8,7 +8,7 @@ import kotlinx.coroutines.internal.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
 
-private val defaultMainDelayOptIn = systemProp("kotlinx.coroutines.main.delay", true)
+private val defaultMainDelayOptIn = systemProp("kotlinx.coroutines.main.delay", false)
 
 internal actual val DefaultDelay: Delay = initializeDefaultDelay()
 
--- a/ui/kotlinx-coroutines-android/test/HandlerDispatcherTest.kt
+++ b/ui/kotlinx-coroutines-android/test/HandlerDispatcherTest.kt
@@ -163,20 +163,19 @@ class HandlerDispatcherTest : TestBase() {
     }
 
     @Test
-    fun testDelayIsDelegatedToMain() = runTest {
+    fun testDelayIsNotDelegatedToMain() = runTest {
         val mainLooper = shadowOf(Looper.getMainLooper())
         mainLooper.pause()
         val mainMessageQueue = shadowOf(Looper.getMainLooper().queue)
         assertNull(mainMessageQueue.head)
         val job = launch(Dispatchers.Default, start = CoroutineStart.UNDISPATCHED) {
             expect(1)
-            delay(10_000_000)
-            expect(3)
+            delay(Long.MAX_VALUE)
+            expectUnreached()
         }
         expect(2)
-        assertNotNull(mainMessageQueue.head)
-        mainLooper.runOneTask()
-        job.join()
-        finish(4)
+        assertNull(mainMessageQueue.head)
+        job.cancelAndJoin()
+        finish(3)
     }
 }


slavonnet:
1. This is critical issue. 
2. . Why don't you want to set the maximum parking time to 1-5 seconds and recheck the event loop? Yes, ideally it should be an unpark, but why not recheck the status and show the warning with additional information?
3. Why not limit the blocking time of the main thread to the size of the frame time or maximum delay and not unblock the thread with the completion of the task in an exception?
4. Why not add support for nested rubBlocking by adding it to an already existing and running coroutine as a child task? Is it enough to keep the link locally in the treadlocal?
5. CoroutineScheduler has MAX_POOL_SIZE over 2M. At the same time, CoroutineScheduler.toString has an enumeration of index in "1 until workers.length". When debugging, they often work. Why not increase the pool by 2? It makes sense ? The costs are not significant compared to the cost of creating one thread.
6. afterCompletion in JobSupport and BlockingCoroutine works only on successful completion, judging by the code. And if an exception or simply destroy another thread, then the blocked thread will be blocked for an infinite time?

dkhalanskyjb (approved):

PR #3137, opened 2022-01-12
Optimization: resizable workers array

elizarov:
Instead of allocating an array of maxPoolSize (~2M) elements for the worst-case supported scenario that may never be reached in practice and takes considerable memory, allocate just an array of corePoolSize elements and grow it dynamically if needed to accommodate more workers.

The data-structure to make it happen must support lock-free reads for performance reasons, but it is simple, since workers array is modified exclusively under synchronization.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
@@ -8,7 +8,7 @@ import kotlinx.coroutines.*
 import java.util.concurrent.atomic.*
 
 /**
- * Atomic array with lock-free reads and synchronized modifications. It is logically has unbounded size,
+ * Atomic array with lock-free reads and synchronized modifications. It logically has an unbounded size,
  * is implicitly filled with nulls, and is resized on updates as needed to grow.
  */
 internal class ResizableAtomicArray<T>(initialLength: Int) {
--- a/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
@@ -4,7 +4,6 @@
 
 package kotlinx.coroutines.internal
 
-import kotlinx.coroutines.*
 import java.util.concurrent.atomic.*
 
 /**
@@ -23,8 +22,8 @@ internal class ResizableAtomicArray<T>(initialLength: Int) {
         return if (index < array.length()) array[index] else null
     }
 
-    @Synchronized
-    operator fun set(index: Int, value: T?) {
+    // Must not be called concurrently, e.g. always use synchronized(this) to call this function
+    fun setSynchronized(index: Int, value: T?) {
         val curArray = this.array
         val curLen = curArray.length()
         if (index < curLen) {
--- a/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
@@ -19,14 +19,12 @@ internal class ResizableAtomicArray<T>(initialLength: Int) {
     public fun currentLength(): Int = array.length()
 
     public operator fun get(index: Int): T? {
-        assert { index >= 0 }
         val array = this.array // volatile read
         return if (index < array.length()) array[index] else null
     }
 
     @Synchronized
     operator fun set(index: Int, value: T?) {
-        assert { index >= 0 }
         val curArray = this.array
         val curLen = curArray.length()
         if (index < curLen) {
--- a/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
@@ -15,7 +15,8 @@ internal class ResizableAtomicArray<T>(initialLength: Int) {
     @Volatile
     private var array = AtomicReferenceArray<T>(initialLength)
 
-    public fun length(): Int = array.length()
+    // for debug output
+    public fun currentLength(): Int = array.length()
 
     public operator fun get(index: Int): T? {
         assert { index >= 0 }
--- a/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2016-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.*
+import java.util.concurrent.atomic.*
+
+/**
+ * Atomic array with lock-free reads and synchronized modifications. It is logically has unbounded size,
+ * is implicitly filled with nulls, and is resized on updates as needed to grow.
+ */
+internal class ResizableAtomicArray<T>(initialLength: Int) {
+    @Volatile
+    private var array = AtomicReferenceArray<T>(initialLength)
+
+    public fun length(): Int = array.length()
+
+    public operator fun get(index: Int): T? {
+        assert { index >= 0 }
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -261,7 +260,7 @@ internal class CoroutineScheduler(
      * works properly
      */
     @JvmField
-    val workers = AtomicReferenceArray<Worker?>(maxPoolSize + 1)
+    val workers = ResizableAtomicArray<Worker?>(corePoolSize + 1)
```
dkhalanskyjb, line 263:
```suggestion
    val workers = ResizableAtomicArray<Worker>(corePoolSize + 1)
```

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Atomic array with lock-free reads and synchronized modifications. It is logically has unbounded size,
+ * is implicitly filled with nulls, and is resized on updates as needed to grow.
+ */
+internal class ResizableAtomicArray<T>(initialLength: Int) {
+    @Volatile
+    private var array = AtomicReferenceArray<T>(initialLength)
+
+    public fun length(): Int = array.length()
```
dkhalanskyjb, line 18: (Minor) This operation is only used for debug output and looks out-of-place in a logically unbounded structure. Maybe call this something like `currentLength` instead?

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright 2016-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.*
+import java.util.concurrent.atomic.*
+
+/**
+ * Atomic array with lock-free reads and synchronized modifications. It is logically has unbounded size,
```
dkhalanskyjb, line 11:
```suggestion
 * Atomic array with lock-free reads and synchronized modifications. It logically has an unbounded size,
```

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
elizarov: Makes sense!

dkhalanskyjb (approved):

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,41 @@
...
+    // for debug output
+    public fun currentLength(): Int = array.length()
+
+    public operator fun get(index: Int): T? {
+        assert { index >= 0 }
+        val array = this.array // volatile read
+        return if (index < array.length()) array[index] else null
+    }
+
+    @Synchronized
```
qwwdfsad, line 27: `set` is always guarded by `synchronized(workers)`.

It's better to avoid excessive synchronization; alternatively, you can do `@GuardedBy("this")` or `assert { Thread.holdsLock(this) }`

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+internal class ResizableAtomicArray<T>(initialLength: Int) {
+    @Volatile
+    private var array = AtomicReferenceArray<T>(initialLength)
+
+    // for debug output
+    public fun currentLength(): Int = array.length()
+
+    public operator fun get(index: Int): T? {
+        assert { index >= 0 }
```
qwwdfsad, line 22: This assert is redundant, `ARA` will perform it for you with better error message

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,41 @@
...
+        val array = this.array // volatile read
+        return if (index < array.length()) array[index] else null
+    }
+
+    @Synchronized
+    operator fun set(index: Int, value: T?) {
+        assert { index >= 0 }
```
qwwdfsad, line 29: Same here

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
elizarov: The `Thread.holdsLock(this)` call is extremely expensive. I don't feel it is Ok to have it running even inside `assert { ... }` (that can be still turned on in production when assertions are enabled)>

Just removing `@Synchronized` and documenting that it can be called only under lock is error-prone, since that code lives in a different package from the scheduler and could be, one day, used by someone who did not read the docs (inside or outside scheduler). 

On the other hand, an extra nested `@Synchronized` is usually removed by HotSpot after inlining and it seems, that this accessor is a small-enough method to be always inlined. I have not checked if that is the case for this particular code, though.

As a compromise, `operator fun set(...)` can be renamed to `fun setSynchonized(...)`. This will make all its call-sites more verbose, so we can remove `@Synchronize` without any fear of it being error-prone.

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package kotlinx.coroutines.internal
+
+import java.util.concurrent.atomic.*
+
+/**
+ * Atomic array with lock-free reads and synchronized modifications. It logically has an unbounded size,
+ * is implicitly filled with nulls, and is resized on updates as needed to grow.
+ */
+internal class ResizableAtomicArray<T>(initialLength: Int) {
```
ndkoval, line 13: Let's add a Lincheck test

qwwdfsad (approved):

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -0,0 +1,38 @@
...
+    fun setSynchronized(index: Int, value: T?) {
+        val curArray = this.array
+        val curLen = curArray.length()
+        if (index < curLen) {
+            curArray[index] = value
+        } else {
+            val newArray = AtomicReferenceArray<T>((index + 1).coerceAtLeast(2 * curLen))
+            for (i in 0 until curLen) newArray[i] = curArray[i]
+            newArray[index] = value
+            array = newArray // copy done
```
slavonnet, line 35:
```
class test<E> (initialLength: Int = 2) {
    
    private var array: AtomicReferenceArray<E?> = AtomicReferenceArray(initialLength)
    private val size: AtomicInteger = AtomicInteger(initialLength)

    fun get(idx: Int): E? = idx.coerceAtMost(size.get()).let { array.get(it) }

    fun set(idx: Int, newValue: E?) {
        size.get().takeIf { it < idx }?.let { s ->
            arrayListOf(
                0.rangeTo(s).asSequence().map { array.get(it) }.toList()
            ).toArray().copyOf(
                idx.inc().coerceAtLeast(s.times(2))
            ).let { newArray ->
                AtomicReferenceArray(newArray as Array<out E?>).let {
                    array = it
                    size.set(newArray.size)
                }
            }
        }
        array.set(idx, newValue)
    }
}

```
synchronized and Volatile do not need

elizarov: I'm not sure what your comments mean, but the implementation you've provided is clearly non-linearizable (however, our current testing infrastructure is unlikely to catch it).

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
elizarov: Ok. Added. Though it still cannot catch, for example, a missing `@Volatile` annotation in this particular code.

Issue #2624, opened 2021-04-04
Add timeout operator to Flow


pablobaxter:
Enhancement: Add `timeout` operator to Flow

This is in response to the updated contribution guidelines, and to push for a `timeout` mechanism for Flow (https://github.com/Kotlin/kotlinx.coroutines/pull/2597). Currently, there is no easy way to timeout a flow if the upstream takes too long to produce an event.

Use case:
A flow when collected may not always emit items at a regular interval. If this flow is using some resource intensive process or enabling some hardware, leaving it active with no events for an extended period of time is wasteful. By having a timeout operator on the flow if no events have emitted, it allows for creating flows based with a defined timeout strategy, to ensure these resources aren't being accidentally misused.

joffrey-bion:
I'm not against such operator but I am not sure this specific use case calls for it.

This use case seems to need the timeout on the *consumer* side, and for this I wonder why a `withTimeout` around the `collect` call is not sufficient. Could you please elaborate?

I would imagine that `.timeout()` operator for a flow would be useful for the *producer* side instead, so that a returned flow can be forced to have a timeout. For instance, if the producer knows that the events are too irregular and prefers to send "stub" events or something to indicate it's taking too long, it would return a `Flow` with the `timeout` operator already set up.

pablobaxter:
To give a more specific use-case, I'll go with one for Android:

With GPS location on Android, it's not always guaranteed that location events come in at the desired interval (it could vary based on the environment). It's even possible that no GPS location will come in if there is no clear signal. However, even if no signal is available, the GPS radio will remain on, draining battery power. To prevent this, you would use a timeout on the location events coming in, and if there are none after a period of time, turn off the radio.

An example of what a location emitting flow would look like (without a timeout):
 
```
fun getLocationFlow(context: Context): Flow<Location> {
    val locationManager = context.getSystemService(Context.LOCATION_SERVICE) as LocationManager
    return callbackFlow {
        val listener = LocationListener { sendBlocking(it) }
        locationManager.requestLocationUpdates(LocationManager.GPS_PROVIDER, 0L, 0F, listener)
        invokeOnClose { locationManager.removeUpdates(listener) }
    }.buffer(Channel.UNLIMITED)
}
```

Every consumer of this flow could have a different requirement for what they consider to be a timeout event. One example is that these locations may not be accurate enough. In this case, a filter would be added to the flow, which filters out inaccurate locations.

```
getLocationFlow(context)
[...]

PR #3148, opened 2022-01-19
Add timeout operator to Flow (redux)

pablobaxter:
This adds functionality similar to the [`timeout()`](http://reactivex.io/documentation/operators/timeout.html) in Rx. This invokes a timeout if the upstream flow doesn't emit after the set period of time, with the option to perform a custom action on timeout.

This PR is in support of #2624. One item that was discussed was the name of this new function. I couldn't think of a better one, but I'm not opposed to renaming from `timeout()` to something else.


pablobaxter:
Any update on this PR?

qwwdfsad:
Hi, sorry, this PR eluded me.

We are very careful with time-related API and still not sure how to properly expose it ( https://github.com/Kotlin/kotlinx.coroutines/pull/2745#pullrequestreview-681969186 and https://github.com/Kotlin/kotlinx.coroutines/pull/2738)

We'll use this work as a reference, but we still haven't found a proper chunk of time to get back to it

pablobaxter:
Thanks for the update!

Regarding the reason from #2745, I understand the concern, but the problem currently is that there is no easy way to shutdown an upstream flow if nothing has been emitted after some period of time, without also shutting down the downstream. What this operator provides does not exist with Flow currently. As mentioned in the issue (https://github.com/Kotlin/kotlinx.coroutines/issues/2624), this operator will be a huge benefit for better handling of resource intensive tasks (ex: GPS emitting on Android, which is battery intensive) when used with a flow.

Let me know if there's anything more I can do to help push this along.

He-Pin:
I think this kind of operator is essential，bug do need more careful to took in as a stable.api

qwwdfsad (commented):
Thanks for pushing it!

I did an initial round of review of public API (i.e. I haven't checked the implementation) and it seems good as a potential candidate for 1.7.0.
If you are still ready to work on it -- let's proceed!

qwwdfsad:
Also, please change the base branch to `develop` and rebase on the latest `develop` for the sake of simplicity

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +347,150 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+ * <!--- TEST -->
+ *
+ * Note that delaying on the downstream doesn't trigger the timeout.
+ *
+ * @param timeoutMillis Timeout period in millis
+ * @param action Action to invoke on timeout. Default is to throw [FlowTimeoutException]
+ */
+@FlowPreview
+public fun <T> Flow<T>.timeout(
+    timeoutMillis: Long,
```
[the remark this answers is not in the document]
pablobaxter: Got it. Will 1.7 deprecate the overloads for the other `Long` overloads?

He-Pin:
@pablobaxter maybe `idleTimeout` ?

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +347,150 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+ * ```kotlin
+ * flow {
+ *     emit(1)
+ *     delay(100)
+ *     emit(2)
+ *     delay(100)
+ *     emit(3)
+ *     delay(1000)
+ *     emit(4)
+ * }.timeout(100.milliseconds) {
```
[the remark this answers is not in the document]
pablobaxter: I see the modifications to the scaffolding to support this. I'll re-run the Knit tool.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +347,150 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+public fun <T> Flow<T>.timeout(
+    timeoutMillis: Long,
+    @BuilderInference action: suspend FlowCollector<T>.() -> Unit = { throw FlowTimeoutException(timeoutMillis) }
+): Flow<T> = timeoutInternal(timeoutMillis, action)
+
+private fun <T> Flow<T>.timeoutInternal(
+    timeoutMillis: Long,
+    action: suspend FlowCollector<T>.() -> Unit
+): Flow<T> = scopedFlow { downStream ->
+    require(timeoutMillis >= 0L) { "Timeout should not be negative" }
```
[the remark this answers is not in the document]
pablobaxter: No. I think this should actually follow how `withTimeout` handles a zero timeout.

pablobaxter:
> @pablobaxter maybe rename to `idleTimeout` ?

@qwwdfsad Thoughts? I know you had a concern about the name of the API as well.

He-Pin:
incase there will be something like `backpressureTimeout`  `completionTimeout` etc someday.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +347,150 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+        }
+    }
+}
+
+/**
+ * This exception is thrown by [timeout] to indicate an upstream flow timeout.
+ *
+ * @constructor Creates a timeout exception with the given message. This constructor is needed for exception stack-traces recovery.
+ */
+public class FlowTimeoutException internal constructor(message: String) : CancellationException(message), CopyableThrowable<FlowTimeoutException> {
```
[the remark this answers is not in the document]
qwwdfsad: I'm afraid that this ship has sailed, and we cannot rely on exception type either. So, if solving this, it probably should be much more complex than a separate exception type (e.g. internally, we do `checkOwnership`, but this is definitely not the best API for public use).

The problem is that you can have `FlowTimeoutException` thrown by the flow body anyways -- due to `flat*` operators while awaiting something etc. The separate exception type just cures the symptoms, not the decease and masks this problem.

I'd suggest leaving `TimeoutException` and see at particular use-cases where handling of a this specific exception is required

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +346,105 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+ * produces the following emissions
+ *
+ * ```text
+ * 1, 2, 3, -1
+ * ```
+ * <!--- TEST -->
+ *
+ * Note that delaying on the downstream doesn't trigger the timeout.
+ *
+ * @param timeout Timeout period
```
qwwdfsad, line 381:
```suggestion
 * @param timeout period. If non-positive, the flow is timed out immediately
```

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +346,105 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+            }
+        } finally {
+            timeoutJob.cancel()
+            send(DONE) // Special signal to let flow end
+        }
+    }
+
+    // Await for values from our producer now
+    whileSelect {
+        values.onReceive { value ->
```
qwwdfsad, line 421: You don't need a select here -- `for (value in values)` should be sufficient.

`select` is designed for use-cases when you have more than one source you'd like to cosume

kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+package kotlinx.coroutines.flow.operators
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.*
+import kotlin.test.*
+import kotlin.time.Duration.Companion.milliseconds
+
+class TimeoutTest : TestBase() {
```
qwwdfsad, line 12: Please add a test that uses `null` value and fails if you replace `send(it ?: NULL)` with `send(it)` in your implementation

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +346,105 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+): Flow<T> = timeoutInternal(timeout.toDelayMillis())
+
+private fun <T> Flow<T>.timeoutInternal(
+    timeoutMillis: Long
+): Flow<T> = scopedFlow { downStream ->
+    if (timeoutMillis <= 0L) throw FlowTimeoutException("Timed out immediately")
+
+    // Produce the values using the default (rendezvous) channel
+    // Similar to [debounceInternal]
+    val values = produce<Any?> {
```
qwwdfsad, line 395: Also, it seems like there exists both more straightforward, idiomatic and efficient solution (tho I haven't really tested it):

```
private fun <T> Flow<T>.timeoutInternal(
    timeoutMillis: Long
): Flow<T> = scopedFlow { downStream ->
    if (timeoutMillis <= 0L) throw FlowTimeoutException("Timed out immediately")
    val values = buffer(Channel.RENDEZVOUS).produceIn(this)
    whileSelect {
        values.onReceiveCatching { value ->
            value.onSuccess {
                downStream.emit(it)
            }.onClosed {
                return@onReceiveCatching false
            }
            return@onReceiveCatching true
        }
        onTimeout(timeoutMillis) {
            throw FlowTimeoutException(timeoutMillis)
        }
    }
}
```

Could you please take a look at it?

qwwdfsad:
Sorry for the delay with review -- caught COVID and became even less responsive :(

pablobaxter:
> Sorry for the delay with review -- caught COVID and became even less responsive :(

No worries. I hope you are resting up and taking care of yourself. Get better soon. I'll work on the changes and update accordingly.

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
pablobaxter: I'm seeing `testScalar` and `testEmpty` tests failing intermittently with this implementation. It seems that it takes some measure of time to hit `onReceiveCatching`, which is just long enough in the case of these two tests to trigger the `onTimeout()` select block.

Not entirely sure why this fails, as I believe it is very close to what I also had, but my guess is that the `invokeOnTimeout()` in the `Select.kt` is posted on a different context or dispatcher, rather than the same one used for the test (given these tests are running with `runTest` instead of `withVirtualTime`).

He-Pin:
> > Sorry for the delay with review -- caught COVID and became even less responsive :(
> 
> No worries. I hope you are resting up and taking care of yourself. Get better soon. I'll work on the changes and update accordingly.

Take good care

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
qwwdfsad: I'll check. `withVirtualTime` is really hacky and written in an an ad-hoc manner, so maybe it causes unforeseen issues

pablobaxter: I don't think the problem is with `withVIrtualTime` in this case, as both tests are running in `runTest`. I dug a bit deeper, and saw that the `onTimeout` was eventually leading to `EventLoopImplBase.scheduleInvokeOnTimeout`, which led to `DelayedRunnableTask`, where in my implementation, it would go to `EventLoopImplBase.scheduleResumeAfterDelay` which leads to the `DelayedResumeTask`. I believe it's this difference that's causing the tests to be flaky.

qwwdfsad: Please update your branch on the most recent `develop`, and then this implementation should pass all the tests.

I suggest using it as the most straightforward and performant

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -348,3 +348,92 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+            timeoutJob.cancel()
+            send(DONE) // Special signal to let flow end
+        }
+    }
+
+    // Await for values from our producer now
+    for (value in values) {
+        if (value !== DONE) {
+            if (value === TIMEOUT) {
+                throw TimeoutCancellationException("Timed out waiting for $timeoutMillis ms")
```
qwwdfsad, line 425: Probably worth using `duration` everywhere and avoid using `Long`-based MS everywhere: `onTimeout` already supports it and exception message will be more friendly

qwwdfsad (commented):
Almost there!

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
qwwdfsad: (just to be clear, "this implementation" -- the one proposed in the comment https://github.com/Kotlin/kotlinx.coroutines/pull/3148#discussion_r981143278)

pablobaxter: Although the tests are much more stable now, there is still a bit of flakiness with this implementation for the same two tests (`testScalar` and `testEmpty`). However, it is still much more performant, and I have to repeat the tests a few million times to even see a failure. I'll make the change, but add a comment to both of these tests about them being flaky.

kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -0,0 +1,210 @@
...
+        assertEquals(1, flow.singleOrNull())
+        finish(5)
+    }
+
+    // Test may rarely fail due to timeout being reached before test has completed
+    @Test
+    fun testEmpty() = runTest {
+        val flow = emptyFlow<Any?>().timeout(1.milliseconds)
+        assertNull(flow.singleOrNull())
+    }
```
qwwdfsad, line 87:
```suggestion
    @Test
    fun testEmpty() = withVirtualTime {
        val flow = emptyFlow<Any?>().timeout(1.milliseconds)
        assertNull(flow.singleOrNull())
        finish(1)
    }
```

kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -0,0 +1,210 @@
...
+
+    // Test may rarely fail due to timeout being reached before test has completed
+    @Test
+    fun testScalar() = runTest {
+        val flow = flowOf(1, 2, 3).timeout(1.milliseconds)
+        assertEquals(listOf(1, 2, 3), flow.toList())
+    }
```
qwwdfsad, line 94:
```suggestion
    @Test
    fun testScalar() = withVirtualTime {
        val flow = flowOf(1, 2, 3).timeout(1.milliseconds)
        assertEquals(listOf(1, 2, 3), flow.toList())
        finish(1)
    }
```

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -345,3 +344,63 @@ internal fun CoroutineScope.fixedPeriodTicker(delayMillis: Long, initialDelayMil
...
+): Flow<T> = timeoutInternal(timeout)
+
+private fun <T> Flow<T>.timeoutInternal(
+    timeout: Duration
+): Flow<T> = scopedFlow { downStream ->
+    if (timeout <= Duration.ZERO) throw TimeoutCancellationException("Timed out immediately")
+
+    val values = buffer(Channel.RENDEZVOUS).produceIn(this)
+
+    whileSelect {
```
qwwdfsad, line 393:
```suggestion
    if (timeout <= Duration.ZERO) throw TimeoutCancellationException("Timed out immediately")
    val values = buffer(Channel.RENDEZVOUS).produceIn(this)
    whileSelect {
```

qwwdfsad (approved):
Great job, thanks for the PR and the overall perseverance!

I'll push a few minor codestyle fixes, wait for the CI and merge it as soon as quality gates are passed

kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -0,0 +1,210 @@
...
+        }.flowOn(NamedDispatchers("upstream")).timeout(100.milliseconds)
+
+        assertEquals(listOf(1), flow.toList())
+        finish(3)
+    }
+
+    @Test
+    fun testSharedFlowTimeout() = runTest {
+        assertFailsWith<TimeoutCancellationException>(MutableSharedFlow<Int>().asSharedFlow().timeout(100.milliseconds))
+    }
```
qwwdfsad, line 178:
```suggestion
    @Test
    fun testSharedFlowTimeout() = runTest {
        // Workaround for JS legacy bug
        try {
            MutableSharedFlow<Int>().asSharedFlow().timeout(100.milliseconds).collect()
            expectUnreached()
        } catch (e: TimeoutCancellationException) {
            finish(1)
        }
    }
```

kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -0,0 +1,210 @@
...
+    fun testUpstreamErrorTimeoutException() = testUpstreamError(TimeoutCancellationException(0, Job()))
+
+    private inline fun <reified T: Throwable> testUpstreamError(cause: T) = runTest {
+        val flow = flow {
+            emit(1)
+            throw cause
+        }.timeout(1.milliseconds)
+
+        assertFailsWith<T>(flow)
+    }
```
qwwdfsad, line 109:
```suggestion
    private inline fun <reified T: Throwable> testUpstreamError(cause: T) = runTest {
        try {
            // Workaround for JS legacy bug
            flow {
                emit(1)
                throw cause
            }.timeout(1.milliseconds).collect()
            expectUnreached()
        } catch (e: Throwable) {
            assertTrue { e is T }
            finish(1)
        }
    }
```

qwwdfsad:
Thank you! 👍

pablobaxter:
> Thank you! 👍

Thank you for helping push this along!

PR #3150, opened 2022-01-21
CoroutineDispatcher.asScheduler for the Rx modules

### LINKED ISSUE #548: Way to use a CoroutineDispatcher as a RxJava Scheduler

### LINKED ISSUE #968: Allow converting CoroutineDispatcher to RxJava scheduler

dkhalanskyjb:
Supersedes https://github.com/Kotlin/kotlinx.coroutines/pull/1923
Fixes #968
Fixes #548

Authored by @recheej

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
@@ -8,7 +8,6 @@ import io.reactivex.*
 import io.reactivex.disposables.*
 import io.reactivex.plugins.*
 import kotlinx.coroutines.*
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.channels.*
 import java.util.concurrent.*
 import kotlin.coroutines.*
@@ -39,7 +38,7 @@ public fun CoroutineDispatcher.asScheduler(): Scheduler =
         DispatcherScheduler(this)
     }
 
-private class DispatcherScheduler(val dispatcher: CoroutineDispatcher) : Scheduler() {
+private class DispatcherScheduler(@JvmField val dispatcher: CoroutineDispatcher) : Scheduler() {
 
     private val schedulerJob = SupervisorJob()
 
@@ -50,16 +49,10 @@ private class DispatcherScheduler(val dispatcher: CoroutineDispatcher) : Schedul
      */
     private val scope = CoroutineScope(schedulerJob + dispatcher)
 
-    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
-        if (!scope.isActive) return Disposables.disposed()
-        var handle: DisposableHandle? = null
-        val task = buildTask(block) { handle!! }
-        val newBlock = Runnable { scope.launch { task.block() } }
-        val ctx = scope.coroutineContext
-        @Suppress("INVISIBLE_MEMBER")
-        ctx.delay.invokeOnTimeout(unit.toMillis(delay), newBlock, ctx).let { handle = it }
-        return task
-    }
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable =
+        scope.scheduleTask(block, unit.toMillis(delay)) { task ->
+            Runnable { scope.launch { task() } }
+        }
 
     override fun createWorker(): Worker = DispatcherWorker(dispatcher, schedulerJob)
 
@@ -82,17 +75,10 @@ private class DispatcherScheduler(val dispatcher: CoroutineDispatcher) : Schedul
             }
         }
 
-        override fun schedule(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
-            if (!workerScope.isActive) return Disposables.disposed()
-            val timeMillis = unit.toMillis(delay)
-            var handle: DisposableHandle? = null
-            val task = buildTask(block) { handle!! }
-            val newBlock = Runnable { blockChannel.trySend(task.block) }
-            val ctx = workerScope.coroutineContext
-            @Suppress("INVISIBLE_MEMBER")
-            ctx.delay.invokeOnTimeout(timeMillis, newBlock, ctx).let { handle = it }
-            return task
-        }
+        override fun schedule(block: Runnable, delay: Long, unit: TimeUnit): Disposable =
+            workerScope.scheduleTask(block, unit.toMillis(delay)) { task ->
+                Runnable { blockChannel.trySend(task) }
+            }
 
         override fun isDisposed(): Boolean = !workerScope.isActive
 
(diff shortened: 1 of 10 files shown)


qwwdfsad:
Could you please squash all of `@recheej` commits into one?

I find it really hard to navigate, especially now with incorrect GH rendering where all changes seems to be contributed by you: 
Screenshot 2022-01-24 at 17 20 01

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,128 @@
...
+
+    /**
+     * The scope for everything happening in this [DispatcherScheduler].
+     *
+     * Running tasks, too, get launched under this scope, because [shutdown] should cancel the running tasks as well.
+     */
+    private val scope = CoroutineScope(schedulerJob + dispatcher)
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+        if (!scope.isActive) return Disposables.disposed()
```
qwwdfsad, line 54: Is it expected behaviour? E.g. do default Rx schedulers follow it?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,128 @@
...
+
+        override fun dispose() {
+            blockChannel.close()
+            workerJob.cancel()
+        }
+    }
+
+    data class Task(val disposable: Disposable, val block: suspend () -> Unit): Disposable by disposable
+
+    companion object {
```
qwwdfsad, line 107: Is this companion really needed?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,128 @@
...
+     * The scope for everything happening in this [DispatcherScheduler].
+     *
+     * Running tasks, too, get launched under this scope, because [shutdown] should cancel the running tasks as well.
+     */
+    private val scope = CoroutineScope(schedulerJob + dispatcher)
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+        if (!scope.isActive) return Disposables.disposed()
+        var handle: DisposableHandle? = null
+        val task = buildTask(block) { handle!! }
```
qwwdfsad, line 56: Given it more thought, could you please elaborate on why such complex interaction exists?

E.g. consider this:

```
return scope.launch {
    delay(...)
    runInterruptible(block) 
} // Here a small manual DisposableHandle-Disposable conversion will be required
```

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,128 @@
...
+                        }
+                    } catch (e: Throwable) {
+                        // TODO: what about TimeoutCancellationException?
+                        if (e !is CancellationException)
+                            RxJavaPlugins.onError(e)
+                    }
+                }
+            }
+        }
+    }
```
qwwdfsad, line 127: `toString` implementation won't hurt here

qwwdfsad (changes requested):

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
dkhalanskyjb: They do return this if they fail to submit a task, for example
https://github.com/ReactiveX/RxJava/blob/f3d8583243e81c0b2b3da1870b9430fc85e95f99/src/main/java/io/reactivex/rxjava3/internal/schedulers/ExecutorScheduler.java#L80
https://github.com/ReactiveX/RxJava/blob/5bb119b1489afe4fbc5bb51b828be95971ee8f1d/src/main/java/io/reactivex/rxjava3/disposables/Disposable.java#L159

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,128 @@
...
+    private val scope = CoroutineScope(schedulerJob + dispatcher)
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable {
+        if (!scope.isActive) return Disposables.disposed()
+        var handle: DisposableHandle? = null
+        val task = buildTask(block) { handle!! }
+        val newBlock = Runnable { scope.launch { task.block() } }
+        val ctx = scope.coroutineContext
+        @Suppress("INVISIBLE_MEMBER")
+        ctx.delay.invokeOnTimeout(unit.toMillis(delay), newBlock, ctx).let { handle = it }
```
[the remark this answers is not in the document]
dkhalanskyjb: `disposable.isDisposed` will not call `handle!!`. `handle()` in `buildTask` is only used in the `Disposable.dispose()` implementation, and there's no one to call `dispose()` before the method ends.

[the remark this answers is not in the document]
qwwdfsad: Oh, okay, thanks you

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
dkhalanskyjb: Though they do this only after `onSchedule`. Fixed in one of the latest commits.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,131 @@
...
+    private class DispatcherWorker(dispatcher: CoroutineDispatcher, parentJob: Job) : Worker() {
+
+        private val workerJob = SupervisorJob(parentJob)
+        private val workerScope = CoroutineScope(workerJob + dispatcher)
+        private val blockChannel = Channel<suspend () -> Unit>(Channel.UNLIMITED)
+
+        init {
+            workerScope.launch {
+                while (isActive) {
+                    val task = blockChannel.receive()
```
qwwdfsad, line 72: This `receive` is racing with the `dispose`.

The following test will result in exception reported to global exception handler:

```
val scheduler = Dispatchers.Default.asScheduler()
val worker = scheduler.createWorker()
delay(500)
worker.dispose()
delay(500)
```

qwwdfsad: You probably need `consumeEach` here

qwwdfsad (changes requested):
Overall looks good

dkhalanskyjb:
Kover finds non-covered code in two places:
* `runInterruptible` has some uncovered branch;
* A method called `DispatcherScheduler$DispatcherWorker$schedule$1$invoke$$inlined$Runnable$1` was not tested at all, but whatever it is, all lines in the `DispatcherWorker` class are marked green.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -4,16 +4,134 @@
...
+     * Running tasks, too, get launched under this scope, because [shutdown] should cancel the running tasks as well.
+     */
+    private val scope = CoroutineScope(schedulerJob + dispatcher)
+
+    override fun scheduleDirect(block: Runnable, delay: Long, unit: TimeUnit): Disposable =
+        scope.scheduleTask(block, unit.toMillis(delay)) { task ->
+            Runnable { scope.launch { task() } }
+        }
+
+    private val workerCounter = atomic(1L)
```
qwwdfsad, line 58: [code style] it would be nice to both have a line indent here and to keep both properties and functions grouped separately, not mixing them all over the place

qwwdfsad (approved):
👍

dkhalanskyjb:
Merged manually to preserve authorship: https://github.com/Kotlin/kotlinx.coroutines/commit/d5f852c3345438d9b8196a59af3482afc50d857b

PR #3155, opened 2022-01-25
Confine context-specific state to the thread in UndispatchedCoroutine…

### LINKED ISSUE #2930: Thread local is not cleaned up sometimes
I'm using a `ThreadContextElement` that sets value of a `ThreadLocal`. After resolving of #985 it worked perfectly.
But after upgrade to 1.5.0 I've got a similar problem: sometimes the last value of the thread local stucks in a worker thread.
Equivalent code:
```
while(true) {
    someCode {
        // here the thread local may already have a value from previous iteration
        withContext(threadLocal.asContextElement("foo")) {
            someOtherCode()
        }
    }
}
```
Actual code of the `ThreadContextElement` implementation is [here](https://github.com/kinfra/kinfra-logging/blob/0.3/src/main/kotlin/ru/kontur/kinfra/logging/LoggingContext.kt).

It is hard to reproduce the issue, but I'm facing it periodically in production (it may take hours or days to arise).
Tested 1.5.0 and 1.5.2, both behaves the same. Running it with `-ea`.

qwwdfsad:
… in order to avoid state interference when the coroutine is updated concurrently.

Concurrency is inevitable in this scenario: when the coroutine that has UndispatchedCoroutine as its completion suspends, we have to clear the thread context, but while we are doing so, concurrent resume of the coroutine could've happened that also ends up in save/clear/update context

Fixes #2930

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
@@ -107,7 +107,7 @@ internal tailrec fun CoroutineStackFrame.undispatchedCompletion(): UndispatchedC
 
 /**
  * Marker indicating that [UndispatchedCoroutine] exists somewhere up in the stack.
- * Used as a performance optimization to avoid stack walking where it is not nesessary.
+ * Used as a performance optimization to avoid stack walking where it is not necessary.
  */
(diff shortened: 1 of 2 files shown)


dkhalanskyjb (commented):
Can't several calls to `thisCoroutine().clearThreadContext()` happen concurrently?

dkhalanskyjb (approved):
Now that I'm more awake I see that the question was wildly irrelevant.

PR #3157, opened 2022-01-25
Fix example code in comment of CopyableThreadContextElement

OliverO2:
The example code in the comment of `CopyableThreadContextElement` contained references to non-existing (property, parameter, class) names. This PR fixes those, simplifies the `Key` initialization and adds a `copy()` invocation in `copyForChildCoroutine()` for a more realistic use case.


qwwdfsad (approved):
Thanks!

PR #3183, opened 2022-02-10
Update Kover to 0.5.0

qwwdfsad:


qwwdfsad:
This PR also fixes our build as I initially thought `kover` is the root cause here

dkhalanskyjb (approved):
I don't understand the difference between doing and not doing `ShadowChoreographer.reset`; other than that, LGTM.

Issue #3177, opened 2022-02-08
Broken documentation about `runTest` in `kotlinx-coroutines-test` main page


joffrey-bion:
Here is what the docs look like at the bottom of [this page](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/):
![image](https://user-images.githubusercontent.com/1913296/153003076-bf086106-56d3-4053-a772-6599c4b23b50.png)

dkhalanskyjb:
Hilarious! Also, the problem is more pronounced than that: in fact, most of the https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-test page is absent.

joffrey-bion:
Yeah I noticed it after checking the README, the rest of the page was also dropped 😄

hfhbd:
I think this is the cause https://github.com/Kotlin/dokka/issues/1838

IgnatBeresnev:
JFYI this has been fixed in Dokka 1.6.20, I can send a PR your way if you want :)

PR #3184, opened 2022-02-10
Update Dokka to 1.6.10

qwwdfsad:


qwwdfsad:
I've also tried to fix #3177 but it is blocked by https://github.com/Kotlin/dokka/issues/1910 without a known workaround

dkhalanskyjb (approved):
Makes sense, though I didn't check the generated docs. Should I?

qwwdfsad:
No, I've skimmed through the docs. They changed a visual style a bit (margins mostly), but in general it's the same

PR #3199, opened 2022-02-21
Properly cleanup completion in SafeCollector to avoid unintended memo…

### LINKED ISSUE #3197: Unused objects not collected in Flow.transform
I produce chunks of data with `flow {}` and then transform them into smaller parts as in the following example:

```
import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.transform

fun main() {
    runBlocking {
        generate().flattenToBooleans().collect()
    }
}

fun generate() = flow {
    emit((1..99999).toMutableList())
    delay(99999)
}

fun Flow<MutableList<Int>>.flattenToBooleans() = transform { list ->
    list.forEach { emit(it == 99999) }
}

suspend fun Flow<Boolean>.collect() = collect { isLast ->
    if (isLast) {
        println("the last has been collected")
    }
}
```

I expect the initial list of 99999 integers to be removed by the GC after it is transformed to booleans, but for some reason, it is still kept in the memory after "the last has been collected" is printed:
image

If I manually clear the list, the objects are removed and everything works normally:

```
// ...
fun Flow<MutableList<Int>>.flattenToBooleans() = transform { list ->
    list.forEach { emit(it == 99999) }
    list.clear()
}
// ...
```
image

If I send another list of integers in `flow {}`, the first one is expectedly removed from the heap, but the new one is still present somewhere in another place:
image

I expect the processed elements to be removed by the GC as soon as possible when using `transform` to free up the space for the other coroutines. This is the code example where commenting out `list.clear()` makes the program run out of the heap space with `-Xmx500m`:

```
fun main() {
    runBlocking {
        repeat(3) {
            launch {
                generate().flattenToBooleans().collect()
            }
            delay(500)
        }
    }
}

fun generate() = flow {
    emit((1..9999999).toMutableList())
    delay(99999)
}

fun Flow<MutableList<Int>>.flattenToBooleans() = transform { list ->
    list.forEach { emit(it == 9999999) }
    list.clear() // comment this line to make the code fail with -Xmx500m
}

suspend fun Flow<Boolean>.collect() = collect { isLast ->
    if (isLast) {
        println("last collected")
    }
}
```

Kotlin 1.6.10, coroutines 1.6.0, azul-15.0.6 JVM on an ARM macOS.

qwwdfsad:
…ry leak that regular coroutines (e.g. unsafe flow) are not prone to

Fixes #3197

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/flow/internal/SafeCollector.kt
+++ b/kotlinx-coroutines-core/jvm/src/flow/internal/SafeCollector.kt
@@ -30,7 +30,7 @@ internal actual class SafeCollector<T> actual constructor(
     @JvmField // Note, it is non-capturing lambda, so no extra allocation during init of SafeCollector
     internal actual val collectContextSize = collectContext.fold(0) { count, _ -> count + 1 }
 
-    // Either context of the last emission or wrapper 'DownstreamExceptionElement'
+    // Either context of the last emission or wrapper 'DownstreamExceptionContext'
     private var lastEmissionContext: CoroutineContext? = null
     // Completion if we are currently suspended or within completion body or null otherwise
     private var completion: Continuation<Unit>? = null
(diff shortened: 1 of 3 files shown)


elizarov (approved):
I've made some tweaks to make the code easier to understand. Tests were tweaked to avoid the usage of unwalkable Interator. Also, FieldWalker is improved to avoid "illegal reflective access".

PR #3200, opened 2022-02-21
Add references to available implementations in TestDispatcher docs

zsmb13:
Currently, linking to [TestDispatcher](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-dispatcher/index.html) directly leads the user to a page where they can't find the available implementations of this interface. This is mostly caused by the implementations being private classes with factory functions instead of inheritors.

This PR adds references to the two implementations. The description of the implementations here is not very important, feel free to modify it if needed.

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestDispatcher.kt
+++ b/kotlinx-coroutines-test/common/src/TestDispatcher.kt
@@ -11,8 +11,10 @@ import kotlin.jvm.*
 /**
  * A test dispatcher that can interface with a [TestCoroutineScheduler].
  * 
- * @see StandardTestDispatcher A predictable, controllable dispatcher.
- * @see UnconfinedTestDispatcher A dispatcher that is not confined to any particular thread.
+ * The available implementations are:
+ * * [StandardTestDispatcher] is a dispatcher that places new tasks into a queue.
+ * * [UnconfinedTestDispatcher] is a dispatcher that behaves like [Dispatchers.Unconfined] while allowing to control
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-test/common/src/TestDispatcher.kt
```diff
@@ -10,6 +10,9 @@ import kotlin.jvm.*
 
 /**
  * A test dispatcher that can interface with a [TestCoroutineScheduler].
+ * 
+ * @see StandardTestDispatcher A predictable, controllable dispatcher.
+ * @see UnconfinedTestDispatcher A dispatcher that is not confined to any particular thread.
```
dkhalanskyjb, line 15:
```suggestion
 * The available implementations are:
 * * [StandardTestDispatcher] is a dispatcher that places new tasks into a queue.
 * * [UnconfinedTestDispatcher] is a dispatcher that behaves like [Dispatchers.Unconfined] while allowing to control
 *   the virtual time.
```
Two things here:
* The description after `@see` doesn't seem to have an effect. 
2022-03-17-112319_794x207_scrot
* I think not being confined to a particular thread is not the most important property of `UnconfinedTestDispatcher`, as tests are typically single-threaded. Forming an event loop is the useful property here. However, describing the event loop here would, in my opinion, be excessive, so, instead, `Dispatchers.Unconfined` is referenced, which explains this.

What do you think?

PR #3202, opened 2022-02-24
docs: Update Coroutine context and dispatchers doc

Pearcekieser:
I think flipping the order of the print and delay makes it more clear that one coroutine is canceled while the other lives on.  Also I added a note that explains why the delay is necessary.  The need for the second delay wasn't immediately clear to me on first read of the coroutine guide.


qwwdfsad:
Thanks for the idea! Opened https://github.com/Kotlin/kotlinx.coroutines/pull/3292 with a full update

PR #3206, opened 2022-02-26
Convert SAM

Goooler:


kotlinx-coroutines-core/common/src/Runnable.common.kt
```diff
@@ -7,15 +7,9 @@ package kotlinx.coroutines
...
      * @suppress
      */
     public fun run()
 }
-
-/**
- * Creates [Runnable] task instance.
- */
-@Suppress("FunctionName")
-public expect inline fun Runnable(crossinline block: () -> Unit): Runnable
```
Goooler, line 21: This `fun Runnable` could be handled by `fun interface Runnable` now?

kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -177,7 +177,7 @@ import kotlin.coroutines.*
  * These implementations ensure that the context preservation property is not violated, and prevent most
  * of the developer mistakes related to concurrency, inconsistent flow dispatchers, and cancellation.
  */
-public interface Flow<out T> {
+public fun interface Flow<out T> {
```
olme04, line 180: IMO, it's better not to make it `fun` as there will be confusion, on what to use when building flow:
1. `Flow { collector -> ... }` 
2. `flow { //this: FlowCollector ... }` 
And second is preferred, because it uses `SafeFlow` which checks `emit` context
https://github.com/Kotlin/kotlinx.coroutines/blob/6e0099d3dd3f54aa740f798359485b753a76a65c/kotlinx-coroutines-core/common/src/flow/Builders.kt#L56

Goooler: The first one is uppercase, the second one is lowercase, it should be fine?

olme04: I mean, that they are too similar to misuse them, not to get conflict of declarations

Goooler: Alright, reverted.

qwwdfsad (commented):
Sorry for being super-late here, but we would prefer keeping the code the way it is:

* `CopyableThrowable` is not expected to be `fun` at all, even though it's technically SAM
* With `Runnables` these changes are backwards-incompatible in K/N and we would prefer to maintain binary compatibility (even though it's quite unstable for K/N) when possible

PR #3207, opened 2022-02-26
Code cleanups

Goooler:


qwwdfsad (commented):
Thanks for the effort, but unfortunately we cannot accept a PR that is spread across hundred of files with changes of style, some of that are indeed acceptable, while others (i.e. explicit visibility modifiers) are deliberate and I would really like to avoid splitting former from latter

It will also clash with most of the WIP PRs and will create an additional maintenance burden

Goooler:
Alright

PR #3225, opened 2022-03-29
docs: clarify section on coroutine memory consumption

scolsen:
Updates the section "Coroutines are light-weight" to clarify its
comparison of the resource-intensiveness of coroutines and threads. The
content is much the same, I've attempted to make it follow the pattern
of the other sections by introducing the conceptual point first, then
presenting the code sample.

This commit also makes the code sample in this section runnable and
omits the surrounding runBlocking block.

### DIFF
--- a/docs/topics/coroutines-basics.md
+++ b/docs/topics/coroutines-basics.md
@@ -245,24 +245,29 @@ Done
 
 <!--- TEST -->
 
-## Coroutines ARE light-weight
+## Coroutines are light-weight
 
-Run the following code:
+Coroutines are less resource-intensive than JVM threads. Code that exhausts the
+JVM's available memory when using threads can be expressed using coroutines
+without hitting resource limits. For example, the following code launches
+100000 distinct coroutines that each wait 5 seconds and then print a period
+('.'):
 
 ```kotlin
 import kotlinx.coroutines.*
 
-//sampleStart
 fun main() = runBlocking {
+//sampleStart
     repeat(100_000) { // launch a lot of coroutines
         launch {
             delay(5000L)
             print(".")
         }
     }
-}
 //sampleEnd
+}
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 > You can get the full code [here](../../kotlinx-coroutines-core/jvm/test/guide/example-basic-06.kt).
 >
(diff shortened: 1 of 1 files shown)


docs/topics/coroutines-basics.md
```diff
@@ -245,35 +245,39 @@ Done
...
 
-## Coroutines ARE light-weight
+## Coroutines are light-weight
 
-Run the following code:
+Coroutines are less resource-intensive than JVM threads. Code that exhausts the
+JVM's available memory when using threads can be expressed using coroutines
+without hitting resource limits. For example, the following code launches
+100000 distinct coroutines that each wait 5 seconds and then print a period
+('.'):
```
dkhalanskyjb, line 254:
```suggestion
('.'), all the while consuming very little memory:
```
Or something in this vein. Otherwise, the whole point is obscured.

docs/topics/coroutines-basics.md
```diff
@@ -245,35 +245,39 @@ Done
...
         launch {
             delay(5000L)
             print(".")
         }
     }
-}
 //sampleEnd
+}
 ```
+{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
```
dkhalanskyjb, line 270: This would go against our narrative :)
Try running https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjEwIiwiY29kZSI6ImltcG9ydCBrb3RsaW54LmNvcm91dGluZXMuKlxuLy9zYW1wbGVTdGFydFxuZnVuIG1haW4oKSA9IHJ1bkJsb2NraW5nIHtcbi8vc2FtcGxlU3RhcnRcbiAgICByZXBlYXQoMTAwXzAwMCkgeyAvLyBsYXVuY2ggYSBsb3Qgb2YgY29yb3V0aW5lc1xuICAgICAgICBsYXVuY2gge1xuICAgICAgICAgICAgZGVsYXkoNTAwMEwpXG4gICAgICAgICAgICBwcmludChcIi5cIilcbiAgICAgICAgfVxuICAgIH1cbn1cbi8vc2FtcGxlRW5kIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiJ9
However lightweight coroutines are, they still take *some* memory, more that our playground allocates.

docs/topics/coroutines-basics.md
```diff
@@ -245,35 +245,39 @@ Done
...
+//sampleStart
     repeat(100_000) { // launch a lot of coroutines
         launch {
             delay(5000L)
             print(".")
         }
     }
-}
 //sampleEnd
+}
```
dkhalanskyjb, line 268: Given that the code is not runnable, I think a better choice would be to remove `//sampleStart` and `//sampleEnd`, given how they aren't recognized for code that is not runnable: https://kotlinlang.org/docs/coroutines-basics.html#coroutines-are-light-weight

dkhalanskyjb (changes requested):

docs/topics/coroutines-basics.md
scolsen: Good point! Added.

docs/topics/coroutines-basics.md
scolsen: ah, I figured something like this might be the case--I added a comment noting as much so that future contributors don't make the same mistake, I felt there was probably a good reason for this one not being runnable but didn't see any immediate indication as to why, thus the PR--thanks for letting me know!

PR #3227, opened 2022-03-30
CopyableThreadContextElement implementation

qwwdfsad:


qwwdfsad:
PTAL, especially at how clear the documentation (both public and internal) is.

I've measured the performance impact and for a classic ping-pong application an additional context scan is negligible and practically undetectable.

It neither means that it's zero nor that we won't encounter any issues with it in the future. The biggest potential impact is our future selves --  I can easily imagine 30-50+ elements context in enterprise-grade applications, but it requires a completely different solution, namely rework of the context elements themselves (now being a straightforward immutable map is not really the fastest solution, especially with weird `plus` and `minusKey` contracts)

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +111,52 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
  *         // child coroutine visible to the child.
  *         return TraceContextElement(traceThreadLocal.get()?.copy())
  *     }
+ *
+ *     override fun mergeForChild(overwritingElement: CoroutineContext.Element): CoroutineContext
+ *         // Merge operation defines how to handle situation when both
+ *         // parent coroutine has the element in the context and it was also
+ *         // explicitly passed to a child coroutine.
+ *         // If merge is not defined, the copy of the element can be returned.
+ *         return TraceContextElement(traceThreadLocal.get()?.copy())
```
dkhalanskyjb, line 126:
```suggestion
 *         // Merge operation defines how to handle situation when both
 *         // the parent coroutine has an element in the context and
 *         // an element with the same key was also
 *         // explicitly passed to the child coroutine.
 *         // If merging does not require special behavior,
 *         // the copy of the element can be returned.
 *         return TraceContextElement(traceThreadLocal.get()?.copy())
```

dkhalanskyjb (changes requested):
Didn't look at the tests yet, but I did try breaking the code in a few non-obvious places to ensure that these are tested.

kotlinx-coroutines-core/jvm/test/ThreadContextMutableCopiesTest.kt
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright 2016-2018 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
dkhalanskyjb, line 2:
```suggestion
 * Copyright 2016-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```

dkhalanskyjb (commented):
Looked through the tests as well.

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -10,35 +10,84 @@ import kotlin.coroutines.jvm.internal.CoroutineStackFrame
...
+        return@fold result + (element as CopyableThreadContextElement<Any?>).mergeForChild(newElement)
+    }
+
+    if (hasElementsRight) {
+        leftoverContext = leftoverContext.fold<CoroutineContext>(EmptyCoroutineContext) { result, element ->
+            // We're appending new context element -- we have to copy it, otherwise it may be shared with others
+            if (element is CopyableThreadContextElement<*>) {
+                return@fold result + element.copyForChild()
+            }
+            return@fold result + element
```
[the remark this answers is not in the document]
qwwdfsad: I believe the original version is more readable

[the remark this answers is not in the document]
dkhalanskyjb: You're right, for some reason I can't place a breakpoint on block subexpressions here. How about this then?
```suggestion
            result + if (element is CopyableThreadContextElement<*>) {
                element.copyForChild()
            } else {
                element
            }
```
If not, feel free to just disregard this.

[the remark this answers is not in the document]
qwwdfsad: In general, I prefer
```
if (condition) {
    return foo
}
doOrReturnOtherFoo
```

over
```
return if (condition) {
    foo
} else {
   otherFoo
}
```

to keep the number of nested blocks minimal. This is, of course, opinionated, so we can stick to the most "concise" one or to the proposed middle-ground if you do object

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -10,35 +10,84 @@ import kotlin.coroutines.jvm.internal.CoroutineStackFrame
...
+    /*
+     * Fast-path: we only have to copy/merge if 'addedContext' (which typically has one or two elements)
+     * contains copyable element.
+     */
+    if (!addedContext.fold(false, hasCopyableElements)) return this + addedContext
+    return foldCopies(this, addedContext, false)
+}
+
+private val hasCopyableElements: (Boolean, CoroutineContext.Element) -> Boolean = { result, it ->
+    result || it is CopyableThreadContextElement<*>
```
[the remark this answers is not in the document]
qwwdfsad: Good suggestion, thanks! I've originally extracted the function to save the number of classfiles, but this is even better

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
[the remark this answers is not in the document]
dkhalanskyjb: In general, I'm the opposite and will typically prefer the other way because it highlights the symmetry between the cases and simplifies mental parsing of the control flow. In particular, in this case, I certainly don't find nesting to be an issue.

In any case, we're splitting hairs here. Both versions are perfectly fine, I think.

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,53 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
- *     override fun copyForChildCoroutine(): CopyableThreadContextElement<TraceData?> {
+ *     override fun copyForChild(): TraceContextElement {
  *         // Copy from the ThreadLocal source of truth at child coroutine launch time. This makes
  *         // ThreadLocal writes between resumption of the parent coroutine and the launch of the
  *         // child coroutine visible to the child.
  *         return TraceContextElement(traceThreadLocal.get()?.copy())
  *     }
+ *
+ *     override fun mergeForChild(overwritingElement: CoroutineContext.Element): CoroutineContext {
+ *         // Merge operation defines how to handle situation when both
```
dkhalanskyjb, line 123:
```suggestion
 *         // Merge operation defines how to handle situations when both
```
```suggestion
 *         // Merge operation defines how to handle a situation when both
```
Either of the two is fine.

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,53 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
+ *         // explicitly passed to the child coroutine.
+ *         // If merging does not require special behavior,
+ *         // the copy of the element can be returned.
+ *         return TraceContextElement(traceThreadLocal.get()?.copy())
+ *     }
  * }
  * ```
  *
  * A coroutine using this mechanism can safely call Java code that assumes it's called using a
  * `Thread`.
```
dkhalanskyjb, line 135: I don't understand the meaning here. Is it possible to call any Java code without using a `Thread`?

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,53 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
  */
+@DelicateCoroutinesApi
```
dkhalanskyjb, line 137: Is this due to the implementations being performance-sensitive? Maybe mention this at the top level, too, then?

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,53 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
 @ExperimentalCoroutinesApi
 public interface CopyableThreadContextElement<S> : ThreadContextElement<S> {
 
     /**
      * Returns a [CopyableThreadContextElement] to replace `this` `CopyableThreadContextElement` in the child
-     * coroutine's context that is under construction.
+     * coroutine's context that is under construction if the child coroutine's context does not contain
```
dkhalanskyjb, line 143: "Child coroutine's context" still irks me: "Returns a Z to replace Y in the X that is under construction if *X* does not contain…" Taken literally, this definition doesn't converge! This probably *is* fairly unambiguous, but still.

Could we maybe evaluate some other options for ?
* "the appended part"
* "the new part of the context" 
* "the added context" (the parameter name for `CoroutineContext.newCoroutineContext`)
* Something else?

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,53 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
      *
      * Since this method is called whenever a new coroutine is launched in a context containing this
      * [CopyableThreadContextElement], implementations are performance-sensitive.
      */
-    public fun copyForChildCoroutine(): CopyableThreadContextElement<S>
+    public fun copyForChild(): CopyableThreadContextElement<S>
+
+    /**
+     * Returns a [CopyableThreadContextElement] to replace `this` `CopyableThreadContextElement` in the child
+     * coroutine's context when the child coroutine's context contains an element with the same [key].
```
dkhalanskyjb, line 156: Ditto.

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
qwwdfsad: >Returns a [CopyableThreadContextElement] to replace `this` `CopyableThreadContextElement` in the child
coroutine's context that is under construction if the inherited parent context does not contain
an element with the same [key].

WDYT?

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
qwwdfsad: Mostly due to how easy it is to get it wrong and to nudge people into reading the doc.

In general, I prefer not to mention any performance characteristics in the doc as it can easily lead folks into believing something is "fast" or "slow".

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
dkhalanskyjb: I think that, since `merge` is called on the context element inherited from the parent, we need to phrase that in terms of an operation on the parent context, and therefore, for consistency, here as well.

qwwdfsad: Could you please elaborate on what do you propose?

>if the inherited parent context does not contain an element with the same [key].

For me it seems like the doc is phrased in terms of parent context

dkhalanskyjb: It's not "frome the point of view" of the parent context, if that makes sense.

Some specific proposals:
```suggestion
     * coroutine's context that is under construction if the added context does not contain an element with the same key.
```
```suggestion
     * coroutine's context that is under construction if the context explicitly passed for construction does not contain an element with the same key.
```

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,54 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
  */
+@DelicateCoroutinesApi
 @ExperimentalCoroutinesApi
 public interface CopyableThreadContextElement<S> : ThreadContextElement<S> {
 
     /**
      * Returns a [CopyableThreadContextElement] to replace `this` `CopyableThreadContextElement` in the child
-     * coroutine's context that is under construction.
+     * coroutine's context that is under construction if the added context does not contain an element with the same key.
+     * an element with the same [key].
```
dkhalanskyjb, line 144: Extraneous line.

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -111,32 +112,54 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
      * Since this method is called whenever a new coroutine is launched in a context containing this
      * [CopyableThreadContextElement], implementations are performance-sensitive.
      */
-    public fun copyForChildCoroutine(): CopyableThreadContextElement<S>
+    public fun copyForChild(): CopyableThreadContextElement<S>
+
+    /**
+     * Returns a [CopyableThreadContextElement] to replace `this` `CopyableThreadContextElement` in the child
+     * coroutine's context that is under construction if the added context does contain an element with the same key.
+     * an element with the same [key].
```
dkhalanskyjb, line 157: Here as well.

dkhalanskyjb (changes requested):

dkhalanskyjb (approved):

PR #3235, opened 2022-04-05
Update Kotlin to 1.6.20

qwwdfsad:
It should finally fix native test hang on linux


dkhalanskyjb:
The `NoJs` annotation can be removed from `testMainMocking`.

dkhalanskyjb (approved):

PR #3252, opened 2022-04-13
    Properly preserve thread local values for coroutines that are not intercepted with DispatchedContinuation

qwwdfsad:


kotlinx-coroutines-core/jvm/test/ThreadLocalStressTest.kt
```diff
@@ -63,10 +67,99 @@ class ThreadLocalStressTest : TestBase() {
...
+        repeat(100) {
+            doTestWithPreparation(
+                ::doTestWithContextSwitch,
+                { threadLocal.set("initial") }) { false /* can randomly wake up on the non-main thread */ }
+            // Here we are always on the main thread
+            assertEquals("initial", threadLocal.get())
+        }
+    }
+
+    private fun doTestWithPreparation(testBody: suspend () -> Unit, setup: () -> Unit, isInvalid: () -> Boolean) {
```
dkhalanskyjb, line 122: Minor, opinionated: `isInvalid` is a reverse of `isValid`, and most implementations here are of the form `!=` or `false`, which is a double negation, adding mental load to this and making code asymmetrical and, arguably, even non-idiomatic, as asserts are most of the time state success conditions positive.

dkhalanskyjb (approved):

PR #3260, opened 2022-04-20
Remove duplicate allowed method in BlockHound integration

koo-taejin:
There are two "pollInternal". :)


dkhalanskyjb (approved):
Thanks! FYI: typically the changes to the code go to the `develop` branch, though in this case, given that the behavior doesn't change, `master` is also fine.

koo-taejin:
It's a simple pr, so I didn't worry too much.
If I were improve or develop something, I will follow what you said.
Thank you for making such a great library. :)

PR #3262, opened 2022-04-21
Remove unnecessary -keep rule for AndroidDispatcherFactory, AndroidExceptionPreHandler

agrieve:
R8 supports keeping and/or optimizing classes found in META-INF/services:
https://b.corp.google.com/issues/120436373#comment7

The last R8 commit I can find related to ServerLoader is
https://r8-review.googlesource.com/53824 (Sept 2020),
so I think R8 1.6 still needs the -keep rule, but 3.0+ should not.

I tested with `./gradlew test`, but if failed in the same way with & without my change on javafx tests.

Fixes: 3111


agrieve:
wrong branch...

PR #3281, opened 2022-05-12
Gradle 7.4.2, Kotlin 1.6.21 and other plugin updates

wojtek-kalicinski:
Upgrading to Gradle 7.4.2 also includes updating or removing plugins that were outdated or incompatible with Gradle 7.x

Kotlin 1.6.21 update is not required but bumping anyway since tests pass.

One difference that I noticed is that dokka now generates absolute paths for links, and not `../../`, if this is a problem it's something that should be looked into. 

`./gradlew build` passes, and I also verified (more or less) that all files are generated in the right paths in the maven publication.

### DIFF
--- a/buildSrc/src/main/kotlin/Publishing.kt
+++ b/buildSrc/src/main/kotlin/Publishing.kt
@@ -7,6 +7,7 @@
 import org.gradle.api.Project
 import org.gradle.api.artifacts.dsl.*
 import org.gradle.api.publish.maven.*
+import org.gradle.kotlin.dsl.*
 import org.gradle.plugins.signing.*
 import java.net.*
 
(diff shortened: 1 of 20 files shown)


qwwdfsad (commented):
I see that the build consistently fails on our CI (e.g. https://teamcity.jetbrains.com/viewLog.html?buildTypeId=KotlinTools_KotlinxCoroutines_Build&buildId=3854604&tab=buildLog&logTab=tree&filter=debug&expand=all&_focus=961):

>Script '/mnt/agent/work/44ec6e850d5c63f0/gradle/test-mocha-js.gradle' line: 12

Could you please take a look and ensure that `./gradlew build` passes?
I'll handle Dokka

wojtek-kalicinski:
There was a code path that was only accessed with -Pteamcity set that I didn't notice. Fixed, can you try it now?

buildSrc/src/main/kotlin/Publishing.kt
```diff
@@ -56,6 +57,11 @@ fun configureMavenPublication(rh: RepositoryHandler, project: Project) {
             password = project.getSensitiveProperty("libs.sonatype.password")
         }
     }
+
+    // Something that's easy to "clean" for development, not mavenLocal
```
qwwdfsad, line 61: 👍

qwwdfsad (approved):
Great job, thanks for looking into that!

PR #3289, opened 2022-05-17
Fix potential data-race in EventLoop

### LINKED ISSUE #3251: Data race in EventLoop
Hey there,

Similar to https://github.com/Kotlin/kotlinx.coroutines/issues/2660, our internal Java TSAN tests in Google picked up a race in EventLoop.common.kt

```
Write of size 4 at
kotlinx.coroutines.EventLoopImplBase$DelayedTask.setHeap(Lkotlinx/coroutines/internal/ThreadSafeHeap;)V EventLoop.common.kt:419
versus read of size 4 at
kotlinx.coroutines.EventLoopImplBase$DelayedTask.dispose()V EventLoop.common.kt:480
```

The race in EventLoop seems to be here
https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/EventLoop.common.kt#L417-L420
that write is not thread safe, but the `dispose` function is marked as synchronized
https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/EventLoop.common.kt#L479-L485
leading to the race in `_heap` reading & writing.

Is this race safe?

Thanks!

qwwdfsad:
Fixes #3251


kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -410,6 +415,7 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
          */
         @JvmField var nanoTime: Long
     ) : Runnable, Comparable<DelayedTask>, DisposableHandle, ThreadSafeHeapNode {
+        @Volatile
```
qwwdfsad, line 418: Without this `volatile`, `removeAtImpl` may see an arbitrary value at `assert { result.heap === this }`, leading to a potential crash.

dkhalanskyjb (commented):
Couldn't find any negative consequences of the change of order, looks good.

dkhalanskyjb (approved):

PR #3291, opened 2022-05-18
Remove TestCoroutineContext.kt completely

### LINKED ISSUE #3218: JPMS: coroutines-core-jvm uses package name kotlinx.coroutines.test
I want to use the new multiplatform test feature from 1.6.0. 
Unfortunately, we use JPMS, which fails with this error message:
```
module io.ktor.io reads package kotlinx.coroutines.test from both kotlinx.coroutines.test.jvm and kotlinx.coroutines.core.jvm
```
Reason: `coroutines-core` uses the same package `kotlinx.coroutines.test`: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/test_/TestCoroutineContext.kt

Maybe related: https://github.com/Kotlin/kotlinx.coroutines/issues/3128

qwwdfsad:
While we typically avoid breaking experimental API in minor releases, even if we're allowed to, this one seems to be acceptable. It was obsolete since the very 1.0.0 and was never integrated into structured concurrency, and on the other side it prevents using `kotlinx-coroutines-core` along with `kotlinx-coroutines-test` in a modularized environment due to split package `kotlinx.coroutines.test`.

Fixes #3218


kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
```diff
@@ -77,9 +77,6 @@ public class CoroutinesBlockHoundIntegration : BlockHoundIntegration {
         for (method in listOf("clear", "peek", "removeFirstOrNull", "addLast")) {
             allowBlockingCallsInside("kotlinx.coroutines.internal.ThreadSafeHeap", method)
         }
-        // [addLastIf] is only used in [EventLoop.common]. Users of [removeFirstIf]:
-        allowBlockingCallsInside("kotlinx.coroutines.test.TestCoroutineDispatcher", "doActionsUntil")
-        allowBlockingCallsInside("kotlinx.coroutines.test.TestCoroutineContext", "triggerActions")
```
dkhalanskyjb, line 82: Not directly related, but looking at this, I don't know how I feel about it being forbidden to do the test module's `runCurrent` and `advance*` from `Dispatchers.Default` when BlockHound is used, even if the operations themselves are completely fine. I don't think this is a realistic concern, but it still feels a bit semantically unclean.

dkhalanskyjb (changes requested):
Grepping for the mentions of the package, I also see things to remove in
* `kotlinx-coroutines-core/README.md`
* `kotlinx-coroutines-core/build.gradle`

kotlinx-coroutines-core/README.md
```diff
@@ -83,11 +83,7 @@ Select expression to perform multiple suspending operations simultaneously until
 
 Low-level primitives for finer-grained control of coroutines.
 
-# Package kotlinx.coroutines.test
-
-Obsolete and deprecated module to test coroutines. Replaced with `kotlinx-coroutines-test` module.
-
-<!--- MODULE kotlinx-coroutines-core -->
+!--- MODULE kotlinx-coroutines-core -->
```
dkhalanskyjb, line 86:
```suggestion
<!--- MODULE kotlinx-coroutines-core -->
```

dkhalanskyjb (approved):

PR #3293, opened 2022-05-18
Add Android-JAR plugin to kx-coroutines-android

wojtek-kalicinski:
The Android-JAR plugin works based on the following principles:
* can be added to a project with both plugins: `java-library` and `maven-publish`
* creates 2 new configurations: `aarRuntime` and `aarApi` with attributes that are requested by Android Gradle plugins
* creates .aar artifacts for these configurations that pull in the regular .jar and additional files specified through extension
* provides extension for configuring required values for the AAR manifest: `minSdk` and `namespace` 
* adds the AAR configurations/artifacts to the maven publication throught the `components.java`

As a result, we'll be able to continue shipping JARs to JVM clients (e.g. host side tests) and AAR to Android clients through the resulting Gradle module metadata.

In the future, I'm hoping to ship this plugin as part of Android Gradle plugins (com.android.<something>) and this code could then be removed.

### DIFF
--- a/buildSrc/src/main/kotlin/AndroidJarPlugin.kt
+++ b/buildSrc/src/main/kotlin/AndroidJarPlugin.kt
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2016-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.Plugin
+import org.gradle.api.Project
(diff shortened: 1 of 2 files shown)


qwwdfsad (commented):
In general, LGTM.

I would prefer not to merge it into `develop` until the final solution arrives (we do not need this CR otherwise).

Would it be okay to merge it into a separate `aar-plugin` branch and target 3rd PR there?

qwwdfsad:
Closing as shelved, feel free to reopen when you are back on it

PR #3297, opened 2022-05-19
Add explicit module-info.java for JPMS compatibility

lion7:


build.gradle
```diff
@@ -300,6 +300,20 @@ allprojects {
     CacheRedirector.configure(project)
 }
 
+// --------------- Java 9 modularity ---------------
+
+// Currently the compilation of the module-info fails for both
+// kotlinx.coroutines.debug and kotlinx.coroutines.test
+// because they export packages that also exist in kotlinx.coroutines.core.
+// Note that only 1 module can export a package.
+def invalidModules = ["kotlinx-coroutines-debug", "kotlinx-coroutines-test"]
```
qwwdfsad, line 309: `kotlinx-coroutines-test` has been fixed in 1.6.2 and no longer contains split package

qwwdfsad (commented):
Thanks for taking care of it! 

This is a change we'll be able to ship in the next major release as it carries compatibility risks and may require and RC first, so I will properly investigate the PR around `kotlinx.coroutines` 1.7.0

Unfortunately, IDEA issues won't be fixed in a foreseeable future, so could you please conditionally enable modules if and only if `Idea.isActive` is false? In `kotlinx.serialization` we had plenty of IDEA issues due to that in JVM-nly modules

lion7:
Unfortunately I cannot use `Idea.isActive` because whatever the Gradle script configures is ignored by IDEA.
It just immediately starts analyzing the module graph if it detects a `module-info.java` in the source root 
(if I understood the logic in `https://github.com/JetBrains/intellij-community/blob/master/java/java-analysis-impl/src/com/intellij/codeInsight/daemon/impl/analysis/JavaModuleGraphUtil.java` correctly).

For now I just moved the `module-info.java` files to a separate `java9` source folder which is included explicitly when compile the module descriptor, similar to how it's done in the `kotlinx.serialization` project.

elizarov:
AFAIU, module declarations must also declare all the services with `uses` and `provides` declarations or the corresponding functionality (like the main dispatcher) will not work. For this purse, it also makes sense to start running all the tests with a module path. This way, these kinds of omissions will be caught during testing.

qwwdfsad:
A short update: the corresponding support of JPMS should land in IDEA 2022.3, so we are ready to proceed with the change.
Could you please rebase it on the most recent `develop` when you have time to spare?

lion7:
I just rebased and can confirm that the highlighting issues described in https://youtrack.jetbrains.com/issue/KTIJ-19614 are resolved in IDEA 2022.3 (EAP)

integration/kotlinx-coroutines-guava/src/module-info.java
```diff
@@ -0,0 +1,8 @@
+module kotlinx.coroutines.guava {
+    requires kotlin.stdlib;
+    requires kotlinx.coroutines.core;
+    requires com.google.common;
+    requires failureaccess;
```
qwwdfsad, line 5: What is `failureaccess` and why it is needed?

kotlinx-coroutines-core/jvm/src/module-info.java
```diff
@@ -0,0 +1,23 @@
...
+    // these are used by kotlinx.coroutines.debug.AgentPremain
+    requires static java.instrument; // contains java.lang.instrument.*
+    requires static jdk.unsupported; // contains sun.misc.Signal
+
+    exports kotlinx.coroutines;
+    exports kotlinx.coroutines.channels;
+    exports kotlinx.coroutines.debug;
+    exports kotlinx.coroutines.debug.internal;
+    exports kotlinx.coroutines.flow;
+    exports kotlinx.coroutines.flow.internal;
```
qwwdfsad, line 14: Could you please elaborate on how exactly you got these packages?

It seems like some of them should not be exported in the first place -- all `internal` (though I have to double check some of them), `scheduling` and `intrinsics`

kotlinx-coroutines-debug/src/module-info.java
```diff
@@ -0,0 +1,16 @@
+module kotlinx.coroutines.debug {
+    requires java.management;
+    requires java.instrument;
+    requires kotlin.stdlib;
+    requires kotlinx.coroutines.core;
+    requires net.bytebuddy;
+    requires net.bytebuddy.agent;
```
qwwdfsad, line 7: Is it going to work with our ByteBuddy shadowing?
https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-debug/build.gradle#L40

reactive/kotlinx-coroutines-reactive/src/module-info.java
```diff
@@ -0,0 +1,8 @@
+module kotlinx.coroutines.reactive {
+    requires kotlin.stdlib;
+    requires kotlinx.coroutines.core;
+    requires kotlinx.atomicfu;
```
qwwdfsad, line 4: All `atomicfu` mentions should be erased at the runtime, so this one should not be necessary

qwwdfsad (changes requested):

qwwdfsad:
Also, it seems like some modules cannot be compiled: https://teamcity.jetbrains.com/viewLog.html?buildId=3984907&tab=buildResultsDiv&buildTypeId=KotlinTools_KotlinxCoroutines_Build

build.gradle
```diff
@@ -290,6 +290,22 @@ allprojects {
...
 
+// --------------- Java 9 modularity ---------------
+
+// Currently the compilation of the module-info fails for
+// kotlinx-coroutines-play-services because it depends on Android JAR's
+// which do not have an explicit module-info descriptor.
+// Because the JAR's are all named `classes.jar`,
+// the automatic module name also becomes `classes`.
+// This conflicts since there are multiple JAR's with identical names.
+def invalidModules = ["kotlinx-coroutines-play-services"]
```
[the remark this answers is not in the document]
lion7: Moved the mentioned logic to a separate file

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,106 @@
...
+            // this is needed for Gradle's module detection to work in
+            // org.gradle.api.tasks.compile.JavaCompile.createSpec
+            source(sourceFile.parentFile)
+            include { it.file == sourceFile }
+
+            // The Kotlin compiler will parse and check module dependencies,
+            // but it currently won't compile to a module-info.class file.
+            // Note that module checking only works on JDK 9+,
+            // because the JDK built-in base modules are not available in earlier versions.
+            val javaVersion = compileKotlinTask.kotlinJavaToolchain.javaVersion.getOrNull()
```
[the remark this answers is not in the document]
lion7: Done

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.gradle.api.tasks.bundling.*
+import org.gradle.api.tasks.compile.*
+import org.gradle.jvm.toolchain.*
+import org.gradle.kotlin.dsl.*
+import org.jetbrains.kotlin.gradle.dsl.*
+
+object Java9Modularity {
+
+    @JvmStatic
+    fun Project.configure() {
```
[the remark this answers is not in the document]
lion7: Done

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,106 @@
...
+            // Ignore warnings about using 'requires transitive' on automatic modules.
+            // not needed when compiling with recent JDKs, e.g. 17
+            options.compilerArgs.add("-Xlint:-requires-transitive-automatic")
+
+            // Patch the compileKotlinJvm output classes into the compilation so exporting packages works correctly.
+            val kotlinCompileDestinationDir = compileKotlinTask.destinationDirectory.asFile.get()
+            options.compilerArgs.addAll(listOf("--patch-module", "$moduleName=$kotlinCompileDestinationDir"))
+
+            // Use the classpath of the compileKotlinJvm task.
+            // Also ensure that the module path is used instead of classpath.
```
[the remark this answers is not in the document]
lion7: Yes, that is correct. You can add new packages, but if you don't export them explicitly in the `module-info.java` they will not be reachable when called from a modular Java application.

integration-testing/src/debugAgentTest/kotlin/DebugProbes.kt
```diff
@@ -1,14 +0,0 @@
-/*
- * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
- */
-@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
-package kotlin.coroutines.jvm.internal
```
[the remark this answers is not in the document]
lion7: It actually wasn't necessary. It seemed to be working fine without this file, but that was only because I still had the compiled class file locally available. I have reverted the file back to it's original form.

integration/kotlinx-coroutines-guava/src/module-info.java
lion7: This is a transitive dependency of Guava. One of the classes inside it is used in the `kotlinx-coroutines-guava` module. 
More specifically, there is a typecheck for `com.google.common.util.concurrent.internal.InternalFutureFailureAccess`.

lion7: Compilation works fine without this `requires failureaccess`, so I removed it.

lion7:
> AFAIU, module declarations must also declare all the services with `uses` and `provides` declarations or the corresponding functionality (like the main dispatcher) will not work. For this purse, it also makes sense to start running all the tests with a module path. This way, these kinds of omissions will be caught during testing.

I also added the necessary `uses` and `provides` declarations to the module descriptors.

reactive/kotlinx-coroutines-reactive/src/module-info.java
lion7: There are multiple codepaths where `kotlinx.atomicfu.AtomicRef` is used, which causes the requires on `atomicfu` to be needed. Without it, the module system can't check / guarantee that `atomicfu` is on the classpath (or more correctly, the module path) which could cause a `ClassNotFoundException` during runtime.

kotlinx-coroutines-core/jvm/src/module-info.java
lion7: These packages are mostly used in `reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt`. So I need to export them from core so I don't get a compile error in `kotlinx-coroutines-reactive`.

qwwdfsad:
Is it ready for review or are you still working on it? (I'm not sure if "re-request review" button is available for external contributors, thus the question)

CodeMason:
Hey @lion7 -- is there any holdup on this? Having proper JPMS support would really be fantastic.

lion7:
@qwwdfsad I think this is read enough for review, I'm not working on it anymore and just rebased it on the latest develop. The only thing that still bothers me is that there are currently no tests executed with a module path, especially after re-reading the comment from @elizarov. 
I did do some manual testing and it should all work, but there might be regressions or that I missed something. Should that kind of testing also be part of this PR or is it ok to do that in a separate PR?

qwwdfsad:
Thanks! Will take care of it during this week

qwwdfsad:
I've started the internal testing (mostly our internal projects, Ktor and some Android apps) process and, as soon as all perturbations are resolved, will publish the dev build for everybody interested in trying it out. 

For now, it seems like JavaFx ahs some non-trivial setup issues: none of the tests is able to start and fails with `java.lang.ClassNotFoundException: kotlinx.coroutines.javafx.JavaFxStressTest` (or any other tests).
@lion7 could you please provide any additional pointers here while I'm looking at other modules?

buildSrc/src/main/kotlin/Java9Modularity.kt
```diff
@@ -0,0 +1,116 @@
...
+            classpath = compileKotlinTask.libraries
+            modularity.inferModulePath.set(true)
+        }
+
+        tasks.getByName<Jar>(target.artifactsTaskName) {
```
ALikhachev, line 107: Please change it to
```suggestion
        tasks.named<Jar>(target.artifactsTaskName) {
```
You're triggering task unconditional task configuration even on running `./gradlew help`:
Screenshot 2023-02-16 at 14 25 16

ALikhachev (changes requested):

qwwdfsad:
I've applied all the proposed suggestions in #3629, will keep you posted regarding further actions

qwwdfsad:
I've run a few out integration tests, so far so good. Our JPMS-related testing capabilities are limited (because we are not really using JPMS ourselves anywhere),  so I've published `1.7.0-jpms-preview` to `https://maven.pkg.jetbrains.space/public/p/kotlinx-coroutines/maven` repository; if you are relying on JPMS in your project, consider checking it out.

qwwdfsad:
Merged manually. Thanks!

PR #3302, opened 2022-05-24
Mention swallowing CancellationExceptions in the docs

dkhalanskyjb:
No idea how to make the diff more structured/less ugly.


qwwdfsad (approved):
Unfortunately, there is no way to make the diff any better

PR #3306, opened 2022-05-30
Update docs about creating custom CoroutineScopes

zsmb13:
Clarify the difference between `MainScope()` and `CoroutineScope()`, remove mention of `CustomScope`

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineScope.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineScope.kt
@@ -42,14 +42,14 @@ import kotlin.coroutines.intrinsics.*
  * ### Custom usage
  *
  * `CoroutineScope` should be declared as a property on entities with a well-defined lifecycle that are
- * responsible for launching children coroutines. The corresponding instance of `CoroutineScope` shall be created
- * with either `CoroutineScope()` or `MainScope()` functions. The difference between them is only in the
- * [CoroutineDispatcher]:
+ * responsible for launching child coroutines. The corresponding instance of `CoroutineScope` shall be created
+ * with either `CoroutineScope()` or `MainScope()`:
  *
- * * `CoroutineScope()` uses [Dispatchers.Default] for its coroutines.
- * * `MainScope()` uses [Dispatchers.Main] for its coroutines.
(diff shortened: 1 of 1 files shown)


qwwdfsad (approved):
Thanks!

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -42,14 +42,14 @@ import kotlin.coroutines.intrinsics.*
...
- * responsible for launching children coroutines. The corresponding instance of `CoroutineScope` shall be created
- * with either `CoroutineScope()` or `MainScope()` functions. The difference between them is only in the
- * [CoroutineDispatcher]:
+ * responsible for launching child coroutines. The corresponding instance of `CoroutineScope` shall be created
+ * with either `CoroutineScope()` or `MainScope()`:
  *
- * * `CoroutineScope()` uses [Dispatchers.Default] for its coroutines.
- * * `MainScope()` uses [Dispatchers.Main] for its coroutines.
+ * * `CoroutineScope()` uses the [context][CoroutineContext] provided to it as a parameter for its coroutines 
+ *   and adds a [Job] if one is not provided as part of the context.
```
dkhalanskyjb, line 49: The part about `Dispatchers.Default` is important though. If one knows that `Dispatchers.Default` is used by default, yes, it's obvious what happens when a default dispatcher is not specified, but if one does not, it's unclear where to look.

qwwdfsad: Every coroutine builder mentions `Dispatchers.Default` though:

>If the context does not have any dispatcher or other [ContinuationInterceptor], then [Dispatchers.Default] is used.
The parent job is inherited from the [CoroutineScope] as well, but it can also be overridden
with a corresponding [context] element.

And mentioning `Dispatchers.Default` here is just being incorrect -- `CoroutineScope` does not provide any dispatcher and each coroutine decides what to use on its own. E.g. some may use `Dispatches.Unconfined` when the dispatcher is not supplied

dkhalanskyjb: Ah, ok, didn't notice `launch` mentioning `Dispatchers.Default`.

Issue #3305, opened 2022-05-28
1.6.2: JS(IR) - Atomicfu Runtime not found


hfhbd:
After updating to `1.6.2`, it is not possible to build the project, if it contains a JS(IR) sourceset.

reproducer:
```kt
plugins {
    kotlin("multiplatform") version "1.6.21"
}

repositories {
    mavenCentral()
}

kotlin {
    jvm()
    js(IR) {
        browser()
    }

    sourceSets {
        commonMain {
            dependencies {
                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.2")
            }
        }
    }
}
```

Project: https://github.com/hfhbd/repro-coroutines162-JS-IR

```stacktrace
e: Could not find "org.jetbrains.kotlin:kotlinx-atomicfu-runtime" in [/Users/philipwedemann/Library/Application Support/kotlin/daemon]
e: java.lang.IllegalStateException: FATAL ERROR: Could not find "org.jetbrains.kotlin:kotlinx-atomicfu-runtime" in [/Users/philipwedemann/Library/Application Support/kotlin/daemon]
	at org.jetbrains.kotlin.ir.backend.js.KlibKt$toResolverLogger$1.fatal(klib.kt:117)
	at org.jetbrains.kotlin.library.KotlinLibrarySearchPathResolver.resolve(SearchPathResolver.kt:175)
	at org.jetbrains.kotlin.library.SearchPathResolver$DefaultImpls.resolve$default(SearchPathResolver.kt:20)
	at org.jetbrains.kotlin.library.SearchPathResolverKt.resolve(SearchPathResolver.kt:29)
	at org.jetbrains.kotlin.library.resolver.impl.KotlinLibraryResolverImpl$resolveDependencies$2$2.invoke(KotlinLibraryResolverImpl.kt:122)
	at org.jetbrains.kotlin.library.resolver.impl.KotlinLibraryResolverImpl$resolveDependencies$2$2.invoke(KotlinLibraryResolverImpl.kt:122)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:210)
	at kotlin.sequences.FilteringSequence$iterator$1.calcNext(Sequences.kt:170)
	at kotlin.sequences.FilteringSequence$iterator$1.hasNext(Sequences.kt:194)
	at kotlin.sequences.TransformingSequence$iterator$1.hasNext(Sequences.kt:214)
[...]

PR #3307, opened 2022-05-31
Integration test project

mvicsokolova:


integration-test/src/commonMain/kotlin/main.kt
```diff
@@ -0,0 +1,9 @@
+import kotlinx.coroutines.*
+
+suspend fun doWorld() = coroutineScope {
```
qwwdfsad, line 3: Please add it as `@Test`, so this code will be ran as well

integration-test/settings.gradle
```diff
@@ -0,0 +1,18 @@
...
+            if (requested.id.id == "kotlin-multiplatform") {
+                useModule("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
+            }
+        }
+    }
+
+    repositories {
+        mavenCentral()
+        maven { url "https://plugins.gradle.org/m2/" }
+        maven { url "https://cache-redirector.jetbrains.com/maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
```
qwwdfsad, line 13: Why do we need `dev` repo?

integration-test/gradlew
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
qwwdfsad, line 1: Please also add Gradle wrapper, so it can be run locally without Gradle 7.3 installed in the system

qwwdfsad (changes requested):

integration-test/gradle.properties
```diff
@@ -0,0 +1,4 @@
+kotlin_version=1.6.21
+coroutines_version=1.6.2-SNAPSHOT
```
qwwdfsad, line 2: TODO before merge: tweak `bump-version.sh`

qwwdfsad:
@dkhalanskyjb it is possible to do so, but such an approach may be not robust enough -- it's easy to have some dependency slipped into subproject via all our `configure.all` blocks and defeat the purpose of the test: whether a standalone, fresh MPP project can depend on `kotlinx.coroutines` (see #3305)

Taking it into account, could you please also take a look at this PR?

integration-test/settings.gradle
mvicsokolova: Removed it, we only need `mavenLocal() `to get coroutines published on the previous step

dkhalanskyjb:
Pros of having a separate subproject:
* If we make a mistake in the build scripts, globally applying some plugin for all subprojects and not excluding the integration testing one, we may miss a production issue that will be caught on the RC stage (if it's not caught then, the smoke test will not do much good). The dependencies can be handled fairly robustly: https://github.com/Kotlin/kotlinx.coroutines/pull/3315

Cons:
* The subproject is not recognized for automatic refactorings (this was the problem that initially led us to move `mavenTest` to be in a subproject and not just lying nearby).
* A complication of the build process.
* *As implemented*, a complication of the project structure, as we already have `integration-testing`, and the split between their responsibilities is very vague. Can be mitigated, see below.

Another consideration is that this problem is not limited to the coroutines library. For example, we encountered a similar issue with the datetime, where the project itself behaved just fine, but projects using it, but not the serialization plugin, did not.

*Maybe this is a sign that we need a more general solution for this issue, like the ability to have subprojects that run the Kotlin compiler only with the stock arguments*, which would allow us to have all the pros without any of the cons. Do we maybe have such an ability already? If so, we should employ that.

If not, let's at least reduce the complexity of the file structure. I see these solutions:
1. Clarify the separation between various types of integration tests.
* Instead of just making a project called `integration-test`, introduce a more general `publication-testing`, for testing the published versions. This includes `smokeTest` (I think this is a more fitting name for what is now `integration-test`), `npmTest`, and `mavenTest`.
* Keep `integration-testing` for tests that depend on the other subprojects—for now, those are `debugAgentTest` and `coreAgentTest`.
2. Change `integration-testing` not to be a subproject, so that all integration tests run on published versions, and let `integration-test` be one of its tests.

qwwdfsad (changes requested):
.

dkhalanskyjb:
Also, my continuing the discussion is not a sign that I'm against us going forward with the current solution. We do need to publish something quickly, and the maintenance burden is not high, so a more clean solution can wait.

integration-testing/build.gradle
```diff
@@ -43,70 +62,34 @@ compileDebugAgentTestKotlin {
...
 
 task debugAgentTest(type: Test) {
     def sourceSet = sourceSets.debugAgentTest
-    dependsOn(project(':kotlinx-coroutines-debug').shadowJar)
-    jvmArgs ('-javaagent:' + project(':kotlinx-coroutines-debug').shadowJar.outputs.files.getFiles()[0])
+    def coroutinesDebugJar = sourceSet.runtimeClasspath.filter {it.name == "kotlinx-coroutines-debug-${coroutines_version}.jar" }.singleFile
+    jvmArgs ('-javaagent:' + coroutinesDebugJar)
     testClassesDirs = sourceSet.output.classesDirs
     classpath = sourceSet.runtimeClasspath
     systemProperties project.properties.subMap(["overwrite.probes"])
```
dkhalanskyjb, line 78: Now, that's an issue.

This flag control the functionality in `integration-testing/src/debugAgentTest/kotlin/PrecompiledDebugProbesTest.kt` that allows us to replace `DebugProbesKt.bin`, a file in the repository, with a new version. I'm sure this functionality is now broken. We should either fix this or adapt in some other way. @qwwdfsad, do we still need `-Poverwrite.probes`? Does anyone rely but us on this, or can we provide a more fitting interface? Also, is the comment below still relevant (from the file with the test)?
> ensure that classfile has major version equal to 50 (Java 6 compliance)

dkhalanskyjb (commented):
Please rebase on top of `develop`, so that git is not confused.

integration-testing/build.gradle
```diff
@@ -5,30 +5,54 @@
...
 
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    testImplementation "org.jetbrains.kotlin:kotlin-test:$kotlin_version"
+    // Workaround to make addSuppressed work in tests
+    testImplementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
```
qwwdfsad, line 24: It doesn't seem to be required here

integration-testing/smokeTest/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+        jvmTest {
+            dependencies {
+                implementation kotlin('test')
+                implementation kotlin('test-junit')
+            }
+        }
+    }
+}
```
qwwdfsad, line 43: Please configure your IDEA with "Ensure an empty line at the end of a file" (https://stackoverflow.com/a/63630815/4465354)

integration-testing/build.gradle
qwwdfsad: The comment is indeed redundant.

>We should either fix this or adapt in some other way

Right. @mvicsokolova please ensure that ` ./gradlew debugAgentTest -Poverwrite.probes=true` properly works. For now it fails with `FileNotFoundException`

qwwdfsad: Apart from that, `./gradlew publishToMavenLocal` + `cd integration-testing` + ` ./gradlew debugAgentTest` fails locally at `PrecompiledDebugProbesTest`. It shouldn't be be the case, please investigate

qwwdfsad (changes requested):
@dkhalanskyjb could you please take a look if general Gradle project structure looks good to you?

dkhalanskyjb (commented):
Yeah, Maria and I discussed the structure already, I don't think we can do much better than that here.

integration-testing/src/debugAgentTest/kotlin/PrecompiledDebugProbesTest.kt
```diff
@@ -23,17 +23,16 @@ class PrecompiledDebugProbesTest {
         val binFile = clz.classLoader.getResourceAsStream("DebugProbesKt.bin")!!
         val binContent = binFile.readBytes()
         if (overwrite) {
-            val url = clz.classLoader.getResource("DebugProbesKt.bin")!!
-            val base = url.toExternalForm().toString().removePrefix("jar:file:").substringBefore("/build")
-            val probes = File(base, "jvm/resources/DebugProbesKt.bin")
+            val base = File("kotlin.coroutines.jvm.internal.DebugProbesKt").absolutePath.substringBefore("integration-testing")
+            val probes = File(base, "kotlinx-coroutines-core/jvm/resources/DebugProbesKt.bin")
```
dkhalanskyjb, line 27: I think this is a fine approach, if not the cleanest. It may be surprising that we use the knowledge that this test project is always located in a specific directory of the main project, but it likely won't bite anyone.

A possible alternative would be to replace `overwrite.probes` with another Gradle property that contained the path to `DebugProbesKt.bin`. On the other hand, I think the cost of the hardcode here is fairly low. @qwwdfsad, do we want to preserve the name/function of the Gradle property?

dkhalanskyjb (changes requested):

dkhalanskyjb (changes requested):
Also, `integration-testing/README.md` is now inaccurate.

mvicsokolova:
Updated the README 🙂

dkhalanskyjb (approved):

PR #3309, opened 2022-06-02
Update withContext KDoc

### LINKED ISSUE #3167: Describe non-dispatching fast-path in 'withContext'
It seems to slightly confuse people that having defensive `withContext(Dispatchers.Main)` or `withContext(Dispatchers.IO)` cause an unnecessary context-switch while it doesn't.

qwwdfsad:
Fixes #3167

### DIFF
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -128,7 +128,7 @@ private class LazyDeferredCoroutine<T>(
  *
  * This function redispatches the given [block] if and only if a different [CoroutineDispatcher]
  * has been explicitly supplied with `coroutineContext` argument, shifting the execution
- * back to an original dispatcher when the [block] completes.
+ * back to the original dispatcher when the [block] completes.
  *
  * Note that the result of `withContext` invocation is dispatched into the original context in a cancellable way
  * with a **prompt cancellation guarantee**, which means that if the original [coroutineContext],
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -127,7 +127,7 @@ private class LazyDeferredCoroutine<T>(
  * the resulting context and throws [CancellationException] if it is not [active][CoroutineContext.isActive].
  *
  * This function redispatches the given [block] if and only if a different [CoroutineDispatcher]
- * has been explicitly supplied with `coroutineContext` argument, shifting the execution
+ * was explicitly supplied with the `coroutineContext` argument, shifting the execution
  * back to the original dispatcher when the [block] completes.
  *
  * Note that the result of `withContext` invocation is dispatched into the original context in a cancellable way
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -126,12 +126,13 @@ private class LazyDeferredCoroutine<T>(
  * This suspending function is cancellable. It immediately checks for cancellation of
  * the resulting context and throws [CancellationException] if it is not [active][CoroutineContext.isActive].
  *
- * This function uses dispatcher from the new context, shifting execution of the [block] into the
- * different thread if a new dispatcher is specified, and back to the original dispatcher
- * when it completes. Note that the result of `withContext` invocation is
- * dispatched into the original context in a cancellable way with a **prompt cancellation guarantee**,
- * which means that if the original [coroutineContext], in which `withContext` was invoked,
- * is cancelled by the time its dispatcher starts to execute the code,
+ * This function redispatches the given [block] if and only if a different [CoroutineDispatcher]
+ * has been explicitly supplied with `coroutineContext` argument, shifting the execution
+ * back to an original dispatcher when the [block] completes.
+ *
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -126,12 +126,13 @@ private class LazyDeferredCoroutine<T>(
...
  *
- * This function uses dispatcher from the new context, shifting execution of the [block] into the
- * different thread if a new dispatcher is specified, and back to the original dispatcher
- * when it completes. Note that the result of `withContext` invocation is
- * dispatched into the original context in a cancellable way with a **prompt cancellation guarantee**,
- * which means that if the original [coroutineContext], in which `withContext` was invoked,
- * is cancelled by the time its dispatcher starts to execute the code,
+ * This function redispatches the given [block] if and only if a different [CoroutineDispatcher]
+ * has been explicitly supplied with `coroutineContext` argument, shifting the execution
+ * back to an original dispatcher when the [block] completes.
```
dkhalanskyjb, line 131:
```suggestion
 * back to the original dispatcher when the [block] completes.
```

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -126,12 +126,13 @@ private class LazyDeferredCoroutine<T>(
...
  * the resulting context and throws [CancellationException] if it is not [active][CoroutineContext.isActive].
  *
- * This function uses dispatcher from the new context, shifting execution of the [block] into the
- * different thread if a new dispatcher is specified, and back to the original dispatcher
- * when it completes. Note that the result of `withContext` invocation is
- * dispatched into the original context in a cancellable way with a **prompt cancellation guarantee**,
- * which means that if the original [coroutineContext], in which `withContext` was invoked,
- * is cancelled by the time its dispatcher starts to execute the code,
+ * This function redispatches the given [block] if and only if a different [CoroutineDispatcher]
+ * has been explicitly supplied with `coroutineContext` argument, shifting the execution
```
dkhalanskyjb, line 130:
```suggestion
 * was explicitly supplied with the `coroutineContext` argument, shifting the execution
```

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -126,12 +126,13 @@ private class LazyDeferredCoroutine<T>(
...
+ * back to an original dispatcher when the [block] completes.
+ *
+ * Note that the result of `withContext` invocation is dispatched into the original context in a cancellable way
+ * with a **prompt cancellation guarantee**, which means that if the original [coroutineContext],
+ * in which `withContext` was invoked, is cancelled by the time its dispatcher starts to execute the code,
```
dkhalanskyjb, line 135:
```suggestion
 * with a **prompt cancellation guarantee**, which means that if the original [coroutineContext]
 * in which `withContext` was invoked is cancelled by the time its dispatcher starts to execute the code,
```

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -126,12 +126,13 @@ private class LazyDeferredCoroutine<T>(
...
  *
- * This function uses dispatcher from the new context, shifting execution of the [block] into the
- * different thread if a new dispatcher is specified, and back to the original dispatcher
- * when it completes. Note that the result of `withContext` invocation is
- * dispatched into the original context in a cancellable way with a **prompt cancellation guarantee**,
- * which means that if the original [coroutineContext], in which `withContext` was invoked,
- * is cancelled by the time its dispatcher starts to execute the code,
+ * This function redispatches the given [block] if and only if a different [CoroutineDispatcher]
+ * was explicitly supplied with the `coroutineContext` argument, shifting the execution
+ * back to the original dispatcher when the [block] completes.
```
dkhalanskyjb, line 131: I think a newcomer may not have enough perspective to judge the importance of this, so I would reword it a bit to denote that this only affects performance. For example:
```suggestion
 * Performance-wise, using `withContext` typically has an impact, as [block] can not be executed
 * immediately and instead needs to be scheduled on the correct dispatcher.
 * However, such a redispatch only happens if a different [CoroutineDispatcher]
 * was explicitly supplied with the [context] argument. In that case, the execution
 * is shifted back to the original dispatcher when the [block] completes.
 * If the [CoroutineDispatcher] for [block] is the same as the original one, [withContext]
 * is cheap performance-wise.
```
Also, noticed that the argument is `context`, not `coroutineContext`.

qwwdfsad: I would strongly suggest avoiding mentioning "performance impact" and "performance-wise" in any official KDocs for a high-level API, otherwise, it becomes a minefield for public speculation about what usages are "fast" and what are not etc.

dkhalanskyjb: Or, to avoid mentioning performance at all, while still leaving the impression that this is just an implementation detail and not something that affects the results meaningfully,
```suggestion
 * Calls to [withContext] whose [context] argument provides a [CoroutineDispatcher] that is
 * different from the current one, by necessity, perform additional dispatches: the [block]
 * can not be executed immediately and needs to be dispatched for execution on
 * the passed [CoroutineDispatcher], and then when the [block] completes, the execution
 * has to shift back to the original dispatcher.
```

PR #3311, opened 2022-06-04
Docs: change volatile example to use IO, not Default

aSemy:
The Default dispatcher regularly returns the exact amount, 100000. https://old.reddit.com/r/Kotlin/comments/v4j01h/volatile_works_in_coroutine/

This seems to happen more often when I run the code on my mobile... but that's a guess.

When I change `Dispatchers.Default` to `Dispatchers.IO` it appears to demonstrate the problem more reliably.

- [ ] Should the Dispatcher in the other examples also be changed?

### DIFF
--- a/docs/topics/shared-mutable-state-and-concurrency.md
+++ b/docs/topics/shared-mutable-state-and-concurrency.md
@@ -110,7 +110,7 @@ suspend fun massiveRun(action: suspend () -> Unit) {
 var counter = 0
 
 fun main() = runBlocking {
-    withContext(Dispatchers.Default) {
+    withContext(Dispatchers.IO) {
         massiveRun {
             counter++
         }


qwwdfsad:
This is unlucky timing, sometimes it flickers. Any other dispatcher will behave the same way and show "10000" occasionally.

FTR, I've verified that current kotlinlang consistently shows numbers different from 10000

aSemy:
I've just tried it again 10 times and it's always returned exactly 100000.

qwwdfsad:
Yeah, that's a good point, I'll update it, thanks!

qwwdfsad:
https://github.com/Kotlin/kotlinx.coroutines/commit/9e1eb9ecd62657f3030af2378a0bd661fd9fad06 done

PR #3312, opened 2022-06-06
Remove NPM validation

qwwdfsad:
We will drop complete support of NPM as soon as JS IR is declared stable and this test is already a deadweight as it was never run due to misconfiguration


dkhalanskyjb (approved):
Ok, sure, but where was the misconfiguration? On the TeamCity side?

qwwdfsad:
Yes, `dryRun` was never applied

PR #3314, opened 2022-06-07
Clarify implementation contract of ThreadContextElement and CopyableThreadContextElement

qwwdfsad:


kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -48,6 +48,16 @@ import kotlin.coroutines.*
  * this coroutine suspends.
  *
  * To use [ThreadLocal] variable within the coroutine use [ThreadLocal.asContextElement][asContextElement] function.
+ *
+ * ### Reentrancy and thread-safety
+ *
+ * Correct implementation of this interface should be reentrant and expect
+ * that [updateThreadContext] can be invoked multiple times prior to matching [restoreThreadContext],
```
dkhalanskyjb, line 55:
```suggestion
 * that [updateThreadContext] can be invoked multiple times prior to the matching [restoreThreadContext],
```

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -133,6 +143,28 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
  *
  * A coroutine using this mechanism can safely call Java code that assumes the corresponding thread local element's
  * value is installed into the target thread local.
+ *
+ * ### Reentrancy and thread-safety
+ *
+ * Correct implementation of this interface should be reentrant and expect
```
dkhalanskyjb, line 149:
```suggestion
 * A correct implementation of this interface should be reentrant and expect
```

dkhalanskyjb, line 149:
```suggestion
 * A correct implementation of this interface should be reentrant and expect
```

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -133,6 +143,28 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
+ * that [updateThreadContext] can be invoked multiple times prior to matching [restoreThreadContext],
+ * but it is guaranteed that every update will eventually receive its matching [restoreThreadContext].
+ *
+ * Even though an element is copied for each child coroutine, an implementation should be ready for the following
+ * interleaving, when a coroutine with the corresponding element is launched on a multithreaded dispatcher:
```
dkhalanskyjb, line 154:
```suggestion
 * interleaving when a coroutine with the corresponding element is launched on a multithreaded dispatcher:
```

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -133,6 +143,28 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
  * A coroutine using this mechanism can safely call Java code that assumes the corresponding thread local element's
  * value is installed into the target thread local.
+ *
+ * ### Reentrancy and thread-safety
+ *
+ * Correct implementation of this interface should be reentrant and expect
+ * that [updateThreadContext] can be invoked multiple times prior to matching [restoreThreadContext],
+ * but it is guaranteed that every update will eventually receive its matching [restoreThreadContext].
+ *
+ * Even though an element is copied for each child coroutine, an implementation should be ready for the following
```
dkhalanskyjb, line 153: Minor, style
```suggestion
 * Even though an element is copied for each child coroutine, an implementation should be able to handle the following
```

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -133,6 +143,28 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
  *
  * A coroutine using this mechanism can safely call Java code that assumes the corresponding thread local element's
  * value is installed into the target thread local.
+ *
+ * ### Reentrancy and thread-safety
+ *
+ * Correct implementation of this interface should be reentrant and expect
+ * that [updateThreadContext] can be invoked multiple times prior to matching [restoreThreadContext],
+ * but it is guaranteed that every update will eventually receive its matching [restoreThreadContext].
```
dkhalanskyjb, line 151: Same as above.

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
```diff
@@ -133,6 +142,27 @@ public interface ThreadContextElement<S> : CoroutineContext.Element {
...
+ * interleaving when a coroutine with the corresponding element is launched on a multithreaded dispatcher:
+ *
+ * ```
+ * coroutine.updateThreadContext() // Thread #1
+ * ... coroutine body ...
+ * // suspension + immediate dispatch happen here
+ * coroutine.updateThreadContext() // Thread #2, coroutine is already resumed
+ * // ... coroutine body after suspension point on Thread #2 ...
+ * coroutine.restoreThreadContext() // Thread #1, is invoked late because Thread #1 is slow
+ * coroutine.restoreThreadContext() // Thread #2, may happen concurrently with the previous restore
```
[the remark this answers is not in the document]
qwwdfsad: In parallel.

For me, non-nested order (aka "non-balanced") is more unnatural

dkhalanskyjb (approved):

kotlinx-coroutines-core/jvm/src/ThreadContextElement.kt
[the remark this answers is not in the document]
dkhalanskyjb: I think it depends on the context. I'd define the order as unnatural if it defies some reasonable expectations. Expecting one `update`/`restore` pair to wrap another one is less reasonable than expecting `restore` corresponding to an earlier `update` to also happen earlier than the `restore` corresponding to the later one.

PR #3315, opened 2022-06-07
Make the mavenTest integration test more robust

dkhalanskyjb:
Now, it checks explicitly that the classes it analyzes were taken
from the local Maven repository. Additionally, the build script was
tweaked so that accidentally setting the dependencies for this
subproject is ignored.


integration-testing/build.gradle
```diff
@@ -68,12 +69,12 @@ task mavenTest(type: Test) {
...
             project.dependencies.create("org.jetbrains.kotlinx:kotlinx-coroutines-android:$version"))
-
     mavenTestClasspathConfiguration.attributes {
         attribute(KotlinPlatformType.attribute, KotlinPlatformType.jvm)
     }
-
-    classpath += mavenTestClasspathConfiguration
+    /* the order is this way because the first match is used, and if we ever accidentally add the wrong dependencies to
+    this source-set, they shouldn't override the dependencies to the local Maven repository */
+    classpath = mavenTestClasspathConfiguration + classpath
```
dkhalanskyjb, line 77: With this, I was unable to make the new check fail, even trying to specify the dependencies for this source set in various places.

dkhalanskyjb:
Yes. If we decide to make `integration-testing` a separate project, this becomes irrelevant.

PR #3325, opened 2022-06-17
A prototype of ThreadContextElement on k/js and k/native

eymar:
It's a mere copy/paste from jvm sources + 2 common tests to showcase how `ThreadContextElement` can be used (similar to Compose Runtime needs). 

It's desirable to have some common CoroutineContext.Element similar to ThreadContextElement: 

> Specifically, such a CoroutineContext.Element would have 2 common things with ThreadContextElement:
> a function e.g. onResume(context: CoroutineContext): S, which is invoked before the coroutine in the specified context is resumed
> a function e.g. onSuspend(context: CoroutineContext, oldState: S), which is invoked after the coroutine in the specified context is suspended


dkhalanskyjb:
Closing in favor of https://github.com/Kotlin/kotlinx.coroutines/pull/4208

PR #3327, opened 2022-06-17
Draft: Add onUndeliveredElement callback to ArrayChannels

sakex:
Hi,

This PR is more of a feature request. It would be great if we had a way to get notified when an Element is dropped with `BufferOverflow.DROP_OLDEST`, are there reasons why it shouldn't be done?


qwwdfsad (commented):
We'll fix that separately, thanks!

See https://github.com/Kotlin/kotlinx.coroutines/pull/3331#pullrequestreview-1014987829 and #3330

qwwdfsad:
Also, could you please share your use case in #3330?

sakex:
@qwwdfsad Done

Issue #3330, opened 2022-06-20
Channel onUndeliveredElement doesn't work


artemkaxboy:
I wanted to use channel with buffer but was disappointed to find out that it doesn't work as expected.

When I use a channel with **capacity = 1**, like this:

```kotlin
val capacity = 1

val channel = Channel<Int>(
    capacity = capacity,
    onBufferOverflow = BufferOverflow.DROP_OLDEST,
    onUndeliveredElement = { value -> println("Dropped value: $value") }
)

runBlocking {
    (1..3).forEach { value ->
        channel.send(value)
        println("Sent value: $value")
    }
}
```

It **works perfectly** as I want, output:

```
Sent value: 1
Dropped value: 1
Sent value: 2
Dropped value: 2
Sent value: 3
```

So I expected the same behavior with **greater capactity = 2**:

```kotlin
val capacity = 2

val channel = Channel<Int>(
    capacity = capacity,
    onBufferOverflow = BufferOverflow.DROP_OLDEST,
    onUndeliveredElement = { value -> println("Dropped value: $value") }
)

runBlocking {
    (1..3).forEach { value ->
        channel.send(value)
        println("Sent value: $value")
    }
}
```

But **onUndeliveredElement never called** in that case

```
Sent value: 1
Sent value: 2
Sent value: 3
```

BufferOverflow.DROP_LATEST doesn't work either.

qwwdfsad:
This is indeed a bug, but it seems to be only fixable in 1.7.0.

The story is the following:

* `BufferOverflow` doesn't acknowledge `onUndeliveredElement` in any channels at all, except `CONFLATED`
* `Channel()` function has an internal optimization -- if capacity is `1` and strategy is `DROP_LATEST`, it picks `CONFLATED` as an underlying implementation because it's semantically the same.

So we have to fix all channels except `CONFLATED` which potentially can be considered as breaking change for most of the channels

qwwdfsad:
While we're evaluating it, could you folks please describe your use case for this?

artemkaxboy:
We have almost the same use case. We receive some data from the network and send items to a channel while the background worker processes them. When we have too many messages we'd like to drop the oldest. Planned to use callback for metrics and logging.

qwwdfsad:
Fixed by #3621

PR #3331, opened 2022-06-20
Fix channel BufferOverflow behavior

artemkaxboy:
Fixes unexpected behavior #3330


qwwdfsad (commented):
Sorry, but I can neither accept this PR nor guide you through the whole contribution process here:

* It will be postponed to 1.7.0, so we cannot merge it
* It requires a lot of analysis to do (e.g. it doesn't work for `capacity = 0` for now and probably there are other places that have to be thoughtfully checked)
* More importantly, we are in the process of rewriting the core of our channels #3103. It's all in progress and it will be really tough to keep this work in sync with develop

Sorry for the inconvenience and thanks for the effort

PR #3335, opened 2022-06-21
Fix a flaky test

dkhalanskyjb:
Not sure what the test checks for. Does it really need to make all these iterations, or just waiting for every type of exception to fire is enough?

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/exceptions/WithContextCancellationStressTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/exceptions/WithContextCancellationStressTest.kt
@@ -27,8 +27,9 @@ class WithContextCancellationStressTest : TestBase() {
         var eCnt = 0
         var e1Cnt = 0
         var e2Cnt = 0
+        var loopsToMake = iterations
 
-        repeat(iterations) {
+        while (eCnt == 0 || e1Cnt == 0 || e2Cnt == 0 || loopsToMake-- > 0) {
             val barrier = CyclicBarrier(4)
             val ctx = pool + NonCancellable
             var e1 = false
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/jvm/test/exceptions/WithContextCancellationStressTest.kt
```diff
@@ -27,8 +27,9 @@ class WithContextCancellationStressTest : TestBase() {
         var eCnt = 0
         var e1Cnt = 0
         var e2Cnt = 0
+        var loopsToMake = iterations
 
-        repeat(iterations) {
+        while (eCnt == 0 || e1Cnt == 0 || e2Cnt == 0 || loopsToMake-- > 0) {
```
qwwdfsad, line 32: The rationale behind this test was the following: given enough parallelism (3 threads) and iterations, we expect to observe every exception type at least once.

Taking into account recent changes in our CI and the nature of this test, I would suggest changing it to be more robust, but keep the original intent: iterate until all exception types are observed OR really many iterations passed (to address flakyness), but add a large enough timeout so CI won't hang for four hours

dkhalanskyjb: Given the rationale, I understand the need to count iterations, as opposed to just doing this, with something like `withTimeout` wrapping everything:
```suggestion
        while (eCnt == 0 || e1Cnt == 0 || e2Cnt == 0) {
```

qwwdfsad: I was thinking about iterations for fast debugging locally and a large enough timeout for slow CI, but this should work as well, up tp you

dkhalanskyjb: I think not having iterations, in this case, is more semantically clear. Most other stress tests perform many iterations so that a problematic condition has more chances to arise. Here, it's not the case: we want to wait for a *success* condition. Waiting for a given number of iterations obscures this, whereas the concept of a timeout is clear and self-evident.

Issue #3287, opened 2022-05-16
Support testing coroutines that outlive the test


dkhalanskyjb:
In general, coroutines should finish their execution at some point, as otherwise, they may stick around, wasting memory, CPU time, and possibly hogging some other resources. For this reason, when a singular piece of work is being tested, it's typically a mistake to have coroutines that don't finish by the end of the test. For this reason, we report unfinished coroutines as an error.

However, sometimes, there are coroutines that are expected to live throughout the lifetime of the program; for example, if the program is processing data from some source, like a sensor, there's no reason to ever stop the data flow from the sensor.

In order to test code that uses such long-running coroutines, it's required to somehow exempt them from being reported at the end.

For now, the workaround is to create a scope for the background tasks and cancel it at the end of the test:
```kotlin
@Test
fun testFoo() = runTest {
    // create a new scope that contains the test framework facilities,
    // but with an unrelated job
    val backgroundScope = CoroutineScope(coroutineContext + Job())
    try {
        // start a never-ending coroutine
        backgroundScope.launch {
            while (true) {
                yield()
            }
        }
        // the actual test here
    } finally {
        // cancel the background work
        backgroundScope.cancel()
    }
}
```

This is a significant pain point, judging by the activity in the issues regarding it: https://github.com/Kotlin/kotlinx.coroutines/issues/1531, https://github.com/Kotlin/kotlinx.coroutines/issues/3283, https://youtrack.jetbrains.com/issue/KT-52274, https://github.com/Kotlin/kotlinx.coroutines/issues/3323.

Some options:
* If #1065 gets implemented, then it will be possible to avoid doing anything else: it will suffice to mark the long-running coroutines for cancellation on scope termination. In production, their scope will just not be terminated, and in testing, on termination of the mocked scope, the long-running coroutines will be canceled. Pros: general, orthogonal. Cons: is not completely clean semantically, as cancellation happening in the test is not what happens in production.
* Communicate the expected usage patterns better. This will still be clunky, but at least not confusing. The path of least resistance.
* Some test-specific API. For example, `TestScope` could provide access to some child scope in which all tasks are ignored when checking for leaks. Such API can directly reflect the intended purpose, which is good, but there are issues when trying to introduce the concept of background work. In essence, we don't want to run the background work to completion when doing `advanceUntilIdle`, but in that case, `finally` blocks in background tasks are not guaranteed to be entered during the test. This needs careful thought.

Some proposed approaches that don't seem viable:
[...]

PR #3348, opened 2022-06-27
Add a scope for launching background work in tests

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/3287

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -253,6 +253,7 @@ internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutin
             select<Unit> {
                 coroutine.onJoin {
                     // observe that someone completed the test coroutine and leave without waiting for the timeout
+                    // TODO: looks like this can never happen
                     completed = true
                 }
                 scheduler.onDispatchEvent {
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -211,6 +211,27 @@ internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutin
     coroutine.start(CoroutineStart.UNDISPATCHED, coroutine) {
         testBody()
     }
+    /**
+     * The general procedure here is as follows:
+     * 1. Try running the work that the scheduler knows about, both background and foreground.
+     *
+     * 2. Wait until we run out of foreground work to do. This could mean one of the following:
+     *    * The main coroutine is already completed. This is checked separately; then we leave the procedure.
+     *    * It's switched to another dispatcher that doesn't know about the [TestCoroutineScheduler].
+     *    * Generally, i's waiting for something external (like a network request, or just an arbitrary callback).
+     *    * The test simply hanged.
+     *    * The main coroutine is waiting for some background work.
+     *
+     * 3. We await progress from things that are not the code under test:
+     *    the background work that the scheduler knows about, the external callbacks,
+     *    the work on dispatchers not linked to the scheduler, etc.
+     *
+     *    When we observe that the code under test can proceed, we go to step 1 again.
+     *    If there is no activity for [dispatchTimeoutMs] milliseconds, we consider the test to have hanged.
+     *
+     *    The background work is not running on a dedicated thread.
+     *    Instead, the test thread itself is used, by spawning a separate coroutine.
+     */
     var completed = false
     while (!completed) {
         scheduler.advanceUntilIdle()
@@ -224,12 +245,14 @@ internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutin
         val backgroundWorkRunner = launch(CoroutineName("background work runner")) {
             while (true) {
                 scheduler.tryRunNextTaskUnless { !isActive }
+                // yield so that the `select` below has a chance to check if its conditions are fulfilled
                 yield()
             }
         }
         try {
             select<Unit> {
                 coroutine.onJoin {
+                    // observe that someone completed the test coroutine and leave without waiting for the timeout
                     completed = true
                 }
                 scheduler.onDispatchEvent {
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -165,7 +165,7 @@ public fun TestScope.runTest(
     it.enter()
     createTestResult {
         runTestCoroutine(it, dispatchTimeoutMs, TestScopeImpl::tryGetCompletionCause, testBody) {
-            backgroundWorkScope.cancel()
+            backgroundScope.cancel()
             testScheduler.advanceUntilIdleOr { false }
             it.leave()
         }
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -164,7 +164,11 @@ public fun TestScope.runTest(
 ): TestResult = asSpecificImplementation().let {
     it.enter()
     createTestResult {
-        runTestCoroutine(it, dispatchTimeoutMs, TestScopeImpl::tryGetCompletionCause, testBody) { it.leave() }
+        runTestCoroutine(it, dispatchTimeoutMs, TestScopeImpl::tryGetCompletionCause, testBody) {
+            backgroundWorkScope.cancel()
+            testScheduler.advanceUntilIdle(backgroundIsIdle = false)
+            it.leave()
+        }
     }
 }
 
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -166,7 +166,7 @@ public fun TestScope.runTest(
     createTestResult {
         runTestCoroutine(it, dispatchTimeoutMs, TestScopeImpl::tryGetCompletionCause, testBody) {
             backgroundWorkScope.cancel()
-            testScheduler.advanceUntilIdle(backgroundIsIdle = false)
+            testScheduler.advanceUntilIdleOr { false }
             it.leave()
         }
     }
@@ -176,7 +176,7 @@ public fun TestScope.runTest(
  * Runs [testProcedure], creating a [TestResult].
  */
 @Suppress("NO_ACTUAL_FOR_EXPECT") // actually suppresses `TestResult`
-internal expect fun createTestResult(testProcedure: suspend () -> Unit): TestResult
+internal expect fun createTestResult(testProcedure: suspend CoroutineScope.() -> Unit): TestResult
 
 /** A coroutine context element indicating that the coroutine is running inside `runTest`. */
 internal object RunningInRunTest : CoroutineContext.Key<RunningInRunTest>, CoroutineContext.Element {
@@ -199,7 +199,7 @@ internal const val DEFAULT_DISPATCH_TIMEOUT_MS = 60_000L
  * The [cleanup] procedure may either throw [UncompletedCoroutinesError] to denote that child coroutines were leaked, or
  * return a list of uncaught exceptions that should be reported at the end of the test.
  */
-internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
+internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutine(
     coroutine: T,
     dispatchTimeoutMs: Long,
     tryGetCompletionCause: T.() -> Throwable?,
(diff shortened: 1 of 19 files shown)


kotlinx-coroutines-test/common/test/TestScopeTest.kt
```diff
@@ -169,6 +169,167 @@ class TestScopeTest {
...
+        delay(10)
+        assertEquals(1, i)
+        assertEquals(1, j)
+    }
+
+    /**
+     * Tests that the background work gets cancelled after the test body finishes.
+     */
+    @Test
+    @NoNative
```
dkhalanskyjb, line 198: This is a draft until we can make the new tests run on Native.

dkhalanskyjb:
Reported the compiler bug here: https://youtrack.jetbrains.com/issue/KT-52985/Native-a-function-with-type-T-returned-a-kotlinUnit-instead-of-null

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -46,6 +46,23 @@ public sealed interface TestScope : CoroutineScope {
...
+    /**
+     * A scope for background work.
+     *
+     * This scope is automatically cancelled when the test finishes.
+     * Additionally, while the coroutines in this scope are run as usual when
+     * using [advanceTimeBy] and [runCurrent], [advanceUntilIdle] will stop advancing the virtual time
+     * once only the coroutines in this scope are left unprocessed.
+     *
+     * Failures in coroutines in this scope do not terminate the test.
+     * Instead, they are reported at the end of the test.
```
qwwdfsad, line 59: It worth mentioning that this scope is not bound by parent-child relationship with test scope itself, so any failure in the test itself won't affect this scope

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -170,6 +187,9 @@ internal class TestScopeImpl(context: CoroutineContext) :
     private val uncaughtExceptions = mutableListOf<Throwable>()
     private val lock = SynchronizedObject()
 
+    override val backgroundWorkScope: CoroutineScope =
+        CoroutineScope(coroutineContext + SupervisorJob() + BackgroundWork)
```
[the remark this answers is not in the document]
dkhalanskyjb: This reveals that we need to improve the name/framing.

The purpose of this scope is not the fuzzy notion of background work, but, specifically, for coroutines that, from the perspective of the test, are always there. They are only cancelled so that resources are not leaked; semantically, they keep running for the whole duration of the test. I would even expect such background work to be launched outside of the test definition, in fields of the test class etc. So, if something significant happens on cancellation of a background task, using this scope is not the proper way to check that.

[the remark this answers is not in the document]
qwwdfsad: Even with a better name, the problem will still exist. It would be nice to report _any_ exceptions that happen within the scope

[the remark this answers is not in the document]
dkhalanskyjb: Not sure about that.

Given the more clear definition of background work, we see that, typically, the background work executes in places that suffer from the same problem, only in production. For example, `viewModelScope` uses `SupervisorJob`, and something crashing in `GlobalScope` is notoriously elusive.

So, let's consider the possibilities.
* The background code never crashes silently in production. Ok, fair enough, nothing to discuss.
* It may crash silently in production, but in a controllable manner that the foreground code expects. Then we would just complicate the tests with an unneeded error-handling ceremony if we forced the error to surface.
* It may crash silently in production, and this is unexpected and should be tested for. Well, why is that not a separate test then?

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -46,6 +47,43 @@ public sealed interface TestScope : CoroutineScope {
      */
     @ExperimentalCoroutinesApi
     public val testScheduler: TestCoroutineScheduler
+
+    /**
+     * A scope for background work.
+     *
+     * This scope is automatically cancelled when the test finishes.
```
whyoleg, line 54: will `runTest` await until `backgroundScope` finished after cancelling? Or it will just cancel it and complete test leaving `backgroundScope` cancelling in background?

dkhalanskyjb: It will await the background scope. There's no "background", that is, no separate thread for this.

kotlinx-coroutines-test/common/test/TestScopeTest.kt
```diff
@@ -169,6 +169,164 @@ class TestScopeTest {
         }
     }
 
+    /** Tests that the background work is being run at all. */
+    @Test
+    fun testBackgroundWorkBeingRun(): TestResult = runTest {
```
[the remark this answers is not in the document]
dkhalanskyjb: Provided a test in that style.

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: Done.

kotlinx-coroutines-test/common/src/TestScope.kt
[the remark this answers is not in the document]
dkhalanskyjb: After an internal discussion, we arrived at this solution: do keep `SupervisorJob`, because it allows the background tasks to fail independently of the test code and one another, which would, in our opinion, lead to the least surprising behavior, but *do* report all the exceptions that it is possible for us to observe. For now, we will assume that the actual production code does not rely on `async` and `produce` not surfacing the errors until queried. If this helps surface some errors, great. If it turns out that people *are* using the fact that exceptions are not reported, we will learn about this with actual use cases. In general, we follow the rule of thumb that exceptions are probably something that's not to be unhandled.

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -46,6 +46,23 @@ public sealed interface TestScope : CoroutineScope {
...
+     * once only the coroutines in this scope are left unprocessed.
+     *
+     * Failures in coroutines in this scope do not terminate the test.
+     * Instead, they are reported at the end of the test.
+     *
+     * A typical use case for this scope is to launch tasks that would outlive the tested code in
+     * the production environment.
+     */
+    @ExperimentalCoroutinesApi
+    public val backgroundWorkScope: CoroutineScope
```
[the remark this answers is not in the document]
dkhalanskyjb: Sure. I assumed that even `backgroundScope` would be too long, so people would do something like
```kotlin
backgroundScope.apply {
    launch { ... }
    launch { ... }
}
```
or pass the scope somewhere. Then I couldn't think of an even shorter name, so I decided to go all-out. However, if people are already using the name `backgroundScope`, it's a great way to tap into an existing convention.

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -216,16 +220,27 @@ internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
...
-        select<Unit> {
-            coroutine.onJoin {
-                completed = true
-            }
-            scheduler.onDispatchEvent {
-                // we received knowledge that `scheduler` observed a dispatch event, so we reset the timeout
+        // in case progress depends on some background work, we need to keep spinning it.
+        val backgroundWorkRunner = launch(CoroutineName("background work runner")) {
+            while (true) {
+                scheduler.tryRunNextTaskUnless { !isActive }
```
qwwdfsad, line 226: Is my understanding of this machinery correct?

* We cannot run background jobs in the previous `advanceUntilIdle` because otherwise, we may be stuck in an endless loop of background work
* `yield` is required to get the chance to `select` to complete and switch `isActive` trigger
* We cannot avoid this `launch` even though we get notified in `onDispatchEvent` because of p.1

qwwdfsad: I've tried to figure out the way to avoid ceremonies with extra `launch`, but it doesn't seem to be that easy.

The machinery itself is already quite complex and it would be nice if you could drop a comment with a rationale behind it

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -216,16 +220,27 @@ internal suspend fun <T: AbstractCoroutine<Unit>> runTestCoroutine(
...
-            onTimeout(dispatchTimeoutMs) {
-                handleTimeout(coroutine, dispatchTimeoutMs, tryGetCompletionCause, cleanup)
+        }
+        try {
+            select<Unit> {
+                coroutine.onJoin {
+                    completed = true
+                }
+                scheduler.onDispatchEvent {
+                    // we received knowledge that `scheduler` observed a dispatch event, so we reset the timeout
```
qwwdfsad, line 236: also worth noting that it doesn't receive events from background jobs

kotlinx-coroutines-test/common/src/internal/ReportingSupervisorJob.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import kotlinx.coroutines.*
+
+/**
+ * A variant of [SupervisorJob] that additionally notifies about child failures via a callback.
+ */
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE", "CANNOT_OVERRIDE_INVISIBLE_MEMBER")
+internal class ReportingSupervisorJob(parent: Job? = null, val onChildCancellation: (Throwable) -> Unit) :
+    JobImpl(parent) {
+    override fun childCancelled(cause: Throwable): Boolean = onChildCancellation(cause).let { false }
```
qwwdfsad, line 15: It's potato-potato of course, but personally I find multi-liners here much more straightforward:
```
onChildCancellation(cause)
return false
```

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -170,14 +208,56 @@ internal class TestScopeImpl(context: CoroutineContext) :
     private val uncaughtExceptions = mutableListOf<Throwable>()
     private val lock = SynchronizedObject()
 
+    override val backgroundScope: CoroutineScope =
+        CoroutineScope(coroutineContext + BackgroundWork + ReportingSupervisorJob {
+            if (it !is CancellationException) reportException(it)
```
qwwdfsad, line 213: `reportException` may throw an exception in edge cases. Job's machinery is unlikely to be ready for that and it can lead to very weird deadlocks or inconsistent states.

I suggest doing something else, for example reporting it to global EH

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -170,14 +208,56 @@ internal class TestScopeImpl(context: CoroutineContext) :
...
+     * ```
+     */
+    // Only call this in a `synchronized(lock)` block.
+    @Suppress("INVISIBLE_MEMBER")
+    private fun collectUncaughtExceptions(): List<Throwable> {
+        val exceptions = LinkedHashSet(uncaughtExceptions)
+        val unwrappedExceptions = LinkedHashSet(uncaughtExceptions.map { unwrap(it) })
+        for (child in backgroundScope.coroutineContext.job.children) {
+            val exception = try {
+                child.getCancellationException()
```
qwwdfsad, line 243: You can probably guard it by `if (!child.isCancelled) continue` to save some time

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -170,14 +208,56 @@ internal class TestScopeImpl(context: CoroutineContext) :
...
+     *     throw TestException()
+     *   }
+     *   delay(1)
+     * }
+     * ```
+     */
+    // Only call this in a `synchronized(lock)` block.
+    @Suppress("INVISIBLE_MEMBER")
+    private fun collectUncaughtExceptions(): List<Throwable> {
+        val exceptions = LinkedHashSet(uncaughtExceptions)
```
qwwdfsad, line 239: In `JobSupport` we use an identity hash set to workaround throwables that throw exceptions from their `hashCode`/`equals`. 
I don't think it's worth fixing here, but it's definitely worth to file an issue here and I'll escalate it to MPP stdlib

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: Do you have good ideas for the `context` for the `handleCoroutineException` invocation?

kotlinx-coroutines-test/common/src/internal/ReportingSupervisorJob.kt
dkhalanskyjb: I like the one-liner approach because the exit points are self-evident. In the example you commented on, yes, this can go both ways. The code was modified and has more complex control flow, but the intent to always return `false` didn't change, and here the benefit, if there is any, should be more clear: the code clearly signifies that, no matter what happens, `false` should always be returned here. One doesn't have to scan the function for other possible exit points.

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: Or I can just remove the code altogether.

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -218,6 +260,12 @@ internal class TestScopeImpl(context: CoroutineContext) :
             if (finished) {
                 throw throwable
             } else {
+                @Suppress("INVISIBLE_MEMBER")
+                for (existingThrowable in uncaughtExceptions) {
+                    // avoid reporting exceptions that already were reported.
+                    if (unwrap(throwable) == unwrap(existingThrowable))
```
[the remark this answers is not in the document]
dkhalanskyjb: I looked at this for inspiration:
```kotlin
private fun Throwable.isSameExceptionAs(other: Throwable?): Boolean =
    other != null && unwrap(other) == unwrap(this)
```

qwwdfsad (approved):

PR #3353, opened 2022-06-29
Reduce reachable references of disposed invokeOnTimeout handle

### LINKED ISSUE #3351: Memory leak with new memory model and withTimeout
See https://youtrack.jetbrains.com/issue/KTOR-4298 for details.

The root cause is our dispatcher implementation:

```

override fun invokeOnTimeout(timeMillis: Long, block: Runnable, context: CoroutineContext): DisposableHandle {
    // No API to cancel on timeout
    worker.executeAfter(timeMillis.toMicrosSafe()) { block.run() }
    return NonDisposableHandle
}
```

Yet it can be workarounded by carefully replacing lambda with something holding an atomic reference to a `block` so it can be nulled out

qwwdfsad:
Fixes #3351

### DIFF
--- a/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
+++ b/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
@@ -38,6 +38,7 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(), 
         // Workers don't have an API to cancel sent "executeAfter" block, but we are trying
         // to control the damage and reduce reachable objects by nulling out `block`
         // that may retain a lot of references, and leaving only an empty shell after a timely disposal
+        // This is a class and not an object with `block` in a closure because that would defeat the purpose.
         class DisposableBlock(block: Runnable) : DisposableHandle, Function0<Unit> {
             private val disposableHolder = AtomicReference<Runnable?>(block)
 
--- a/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
+++ b/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
@@ -35,9 +35,24 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(), 
     }
 
     override fun invokeOnTimeout(timeMillis: Long, block: Runnable, context: CoroutineContext): DisposableHandle {
-        // No API to cancel on timeout
-        worker.executeAfter(timeMillis.toMicrosSafe()) { block.run() }
-        return NonDisposableHandle
+        // Workers don't have an API to cancel sent "executeAfter" block, but we are trying
+        // to control the damage and reduce reachable objects by nulling out `block`
+        // that may retain a lot of references, and leaving only an empty shell after a timely disposal
+        class DisposableBlock(block: Runnable) : DisposableHandle, Function0<Unit> {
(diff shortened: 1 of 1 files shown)


qwwdfsad:
I cannot write a proper unit test though, but we've discussed it with K/N folks and it seems like there is a potential API for that  `fun Any.getObjectReferenceFields() = List<Any>`. I'll keep you updated about that

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -35,9 +35,24 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
-        worker.executeAfter(timeMillis.toMicrosSafe()) { block.run() }
-        return NonDisposableHandle
+        // Workers don't have an API to cancel sent "executeAfter" block, but we are trying
+        // to control the damage and reduce reachable objects by nulling out `block`
+        // that may retain a lot of references, and leaving only an empty shell after a timely disposal
+        class DisposableBlock(block: Runnable) : DisposableHandle, Function0<Unit> {
+            private val disposableHolder = AtomicReference<Runnable?>(block)
+
+            override fun invoke() {
+                disposableHolder.value?.run()
```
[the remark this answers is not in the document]
qwwdfsad: No need, if it is being invoked then it's already removed from worker's queue

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -35,9 +35,24 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
     }
 
     override fun invokeOnTimeout(timeMillis: Long, block: Runnable, context: CoroutineContext): DisposableHandle {
-        // No API to cancel on timeout
-        worker.executeAfter(timeMillis.toMicrosSafe()) { block.run() }
-        return NonDisposableHandle
+        // Workers don't have an API to cancel sent "executeAfter" block, but we are trying
+        // to control the damage and reduce reachable objects by nulling out `block`
+        // that may retain a lot of references, and leaving only an empty shell after a timely disposal
+        class DisposableBlock(block: Runnable) : DisposableHandle, Function0<Unit> {
```
dkhalanskyjb, line 41: It could be nice to highlight that this code being non-idiomatic is intentional.
```suggestion
        // This is a class and not an object with `block` in a closure because that would defeat the purpose.
        class DisposableBlock(block: Runnable) : DisposableHandle, Function0<Unit> {
```

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -35,9 +35,24 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+            }
+
+            override fun dispose() {
+                disposableHolder.value = null
+            }
+        }
+
+        val disposableBlock = DisposableBlock(block)
+        worker.executeAfter(timeMillis.toMicrosSafe(), disposableBlock)
+        return disposableBlock
```
dkhalanskyjb, line 55:
```suggestion
        DisposableBlock(block).also { worker.executeAfter(timeMillis.toMicrosSafe(), it) }
```
The variable name does not seem to be informative.

Issue #3345, opened 2022-06-24
kotlinx-coroutines-debug version 1.6.3 breaks build with Spring Boot Upgrade to 2.7.1


jesperancinha:
Hi there!

Yesterday I noticed that on several of my repos the build was breaking and dependabot could not automate any of the Spring Boot updates. After an exhaustive investigation I came accross for `kotlinx-coroutines-debug` which is included in spring boot 2.7.1, there is jar dependency configured for the build of materials and that of course generates:

```shell
[ERROR] Failed to execute goal on project stamps-and-coins-service: Could not resolve dependencies for project <> org.jetbrains.kotlinx:kotlinx-coroutines-bom:jar:1.6.3 was not found in https://repo.maven.apache.org/maven2 during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]
```

This is what I see when looking into the `kotlinx-coroutines-debug` dependency:

```xml
 <dependencies>
    <dependency>
      <groupId>org.jetbrains.kotlinx</groupId>
      <artifactId>kotlinx-coroutines-core</artifactId>
      <version>1.6.3</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlinx</groupId>
      <artifactId>kotlinx-coroutines-bom</artifactId>
      <version>1.6.3</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>net.java.dev.jna</groupId>
      <artifactId>jna</artifactId>
      <version>5.9.0</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>net.java.dev.jna</groupId>
      <artifactId>jna-platform</artifactId>
      <version>5.9.0</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib-jdk8</artifactId>
      <version>1.6.21</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
```

So I'm guessing the problem is here:

```xml
    <dependency>
      <groupId>org.jetbrains.kotlinx</groupId>
      <artifactId>kotlinx-coroutines-bom</artifactId>
      <version>1.6.3</version>
      <scope>runtime</scope>
    </dependency>
```

Are you planning on making a release soon about this?

Thanks!

[...]

PR #3357, opened 2022-07-05
Fix debug module publication with shadow plugin

qwwdfsad:
It seems like Maven's BOM has never worked with shadow plugin in our previous releases: added `api` dependency was just too late.
It has been changed in the latest shadow plugin release and it automatically promoted all `api` dependencies to `runtime` (https://github.com/johnrengelman/shadow/issues/321), triggering #3345. Gradle builds were never affected though because they were primarily parsing Gradle metadata.

<details>
<summary>Resulting debug module POM</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.jetbrains.kotlinx</groupId>
  <artifactId>kotlinx-coroutines-debug</artifactId>
  <version>1.6.3-SNAPSHOT</version>
  <packaging>pom</packaging>
  <name>kotlinx-coroutines-debug</name>
  <description>Coroutines support libraries for Kotlin</description>
  <url>https://github.com/Kotlin/kotlinx.coroutines</url>
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
  <developers>
    <developer>
      <id>JetBrains</id>
      <name>JetBrains Team</name>
      <organization>JetBrains</organization>
      <organizationUrl>https://www.jetbrains.com</organizationUrl>
    </developer>
  </developers>
  <scm>
    <url>https://github.com/Kotlin/kotlinx.coroutines</url>
  </scm>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.jetbrains.kotlinx</groupId>
        <artifactId>kotlinx-coroutines-bom</artifactId>
        <version>1.6.3-SNAPSHOT</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>org.jetbrains.kotlinx</groupId>
      <artifactId>kotlinx-coroutines-core-jvm</artifactId>
      <version>1.6.3-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>net.java.dev.jna</groupId>
      <artifactId>jna</artifactId>
      <version>5.9.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>net.java.dev.jna</groupId>
      <artifactId>jna-platform</artifactId>
      <version>5.9.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib-jdk8</artifactId>
      <version>1.6.21</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>net.bytebuddy</groupId>
      <artifactId>byte-buddy</artifactId>
      <version>1.10.9</version>
      <scope>runtime</scope>
    </dependency>
    <dependency>
      <groupId>net.bytebuddy</groupId>
      <artifactId>byte-buddy-agent</artifactId>
      <version>1.10.9</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>
</project>

```

### DIFF
--- a/integration-testing/build.gradle
+++ b/integration-testing/build.gradle
@@ -53,6 +53,18 @@ sourceSets {
             implementation "org.jetbrains.kotlinx:kotlinx-coroutines-debug:$coroutines_version"
         }
     }
+
+    debugDynamicAgentTest {
+        kotlin
+        compileClasspath += sourceSets.test.runtimeClasspath
+        runtimeClasspath += sourceSets.test.runtimeClasspath
+
+        dependencies {
+            implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version"
+            implementation "org.jetbrains.kotlinx:kotlinx-coroutines-debug:$coroutines_version"
+        }
+    }
+
     coreAgentTest {
         kotlin
         compileClasspath += sourceSets.test.runtimeClasspath
@@ -93,6 +105,14 @@ task debugAgentTest(type: Test) {
     systemProperties project.properties.subMap(["overwrite.probes"])
 }
 
+task debugDynamicAgentTest(type: Test) {
+    environment "version", coroutines_version
+    def sourceSet = sourceSets.debugDynamicAgentTest
+    testClassesDirs = sourceSet.output.classesDirs
+    classpath = sourceSet.runtimeClasspath
+}
+
+
 task coreAgentTest(type: Test) {
     def sourceSet = sourceSets.coreAgentTest
     def coroutinesCoreJar = sourceSet.runtimeClasspath.filter {it.name == "kotlinx-coroutines-core-jvm-${coroutines_version}.jar" }.singleFile
@@ -106,5 +126,5 @@ compileTestKotlin {
 }
 
 check {
-    dependsOn([withGuavaTest, mavenTest, debugAgentTest, coreAgentTest, 'smokeTest:build'])
+    dependsOn([withGuavaTest, debugDynamicAgentTest, mavenTest, debugAgentTest, coreAgentTest, 'smokeTest:build'])
 }
--- a/integration-testing/build.gradle
+++ b/integration-testing/build.gradle
@@ -23,6 +23,7 @@ dependencies {
 }
 
 sourceSets {
+    // Test that relies on Guava to reflectively check all Throwable subclasses in coroutines
     withGuavaTest {
         kotlin
         compileClasspath += sourceSets.test.runtimeClasspath
@@ -33,6 +34,7 @@ sourceSets {
             implementation 'com.google.guava:guava:31.1-jre'
         }
     }
+    // Checks correctness of Maven publication (JAR resources) and absence of atomicfu symbols
     mavenTest {
         kotlin
         compileClasspath += sourceSets.test.runtimeClasspath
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-debug/build.gradle
```diff
@@ -39,19 +31,27 @@ java {
 }
 
 jar {
+    setEnabled(false)
     manifest {
         attributes "Premain-Class": "kotlinx.coroutines.debug.AgentPremain"
         attributes "Can-Redefine-Classes": "true"
     }
 }
```
dkhalanskyjb, line 39: I find this very counterintuitive: if a task is not enabled, why does its configuration contribute to anything? I propose to move the `manifest` block to `shadowJar`. If I understood correctly (and from a quick check of the resulting `META-INF/MANIFEST.MF` in the published JAR), this shouldn't change the result, but it would look much more clear.

dkhalanskyjb: That's what I was proposing to do, but then I noticed the `Class-Path` line disappearing from the manifest.

qwwdfsad: The previous version also does not generate `Class-Path` attribute. 
This is a nice catch, though I cannot convince myself that it actually ever affected anything.

dkhalanskyjb: This is what I observed as well, yes, just wasn't sure.

Could you please try using the debug module in a minimal project as a dependency and check that everything still works, just to be safe? We do have integration tests for using it as a Java agent, but not for just using `DebugProbes`, right?

qwwdfsad: Restored it back, created https://github.com/Kotlin/kotlinx.coroutines/issues/3361

Will push integration test soon as well

integration-testing/build.gradle
```diff
@@ -93,6 +105,14 @@ task debugAgentTest(type: Test) {
     systemProperties project.properties.subMap(["overwrite.probes"])
 }
 
+task debugDynamicAgentTest(type: Test) {
+    environment "version", coroutines_version
```
dkhalanskyjb, line 109: Could you drop a comment on why this is needed?

qwwdfsad: oh, this one is indeed not needed, I'll remove it

integration-testing/src/debugDynamicAgentTest/kotlin/DynamicAttachDebugTest.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
dkhalanskyjb, line 2:
```suggestion
 * Copyright 2016-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```

dkhalanskyjb (approved):

PR #3363, opened 2022-07-12
Fix Mono.awaitSingleOrNull non-compliance with 2.7

dkhalanskyjb:


reactive/kotlinx-coroutines-reactor/test/MonoAwaitCancellationStressTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.reactivestreams.*
+import reactor.core.*
+import reactor.core.publisher.*
+import java.util.concurrent.locks.*
+
+class MonoAwaitCancellationStressTest {
+    private val iterations = 10_000 * stressTestMultiplier
+
+    @Test
+    fun testAwaitCancellationOrder() = runBlocking {
```
qwwdfsad, line 18: This test consistently deadlocks on the very first iteration on my machine

dkhalanskyjb: For me, the behavior here is *extremely* odd. When I run the test, I get dozens of times in a row when *all* iterations pass, and then dozens of times in a row when, like for you, the first iteration hangs.

OlegDokuka:
Just FYI. Having cancel racing with request is expected behavior but rule 2.7. Please see related PR I have done in the past for the spec see https://github.com/reactive-streams/reactive-streams-jvm/pull/489

OlegDokuka:
@dkhalanskyjb rule 3.5 is not about race conditions but rather about idempotency and non-blocking behaviour

OlegDokuka:
@qwwdfsad this PR is not really relevant, so my suggestion is to reject it.

dkhalanskyjb:
Hey @OlegDokuka, any progress on updating the spec? I don't see any new proposals in https://github.com/reactive-streams/reactive-streams-jvm/pulls — is that the place to look?

dkhalanskyjb:
Ok, I give up, this PR is certainly not useful anymore.

PR #3376, opened 2022-07-25
Get rid of old Kotlin/Native memory model

### LINKED ISSUE #3375: Stop supporting "old" K/N memory model in the favour of new one
To significantly simplify maintenance and implementation of new features for `kotlinx.coroutines` 1.7.0.

The new memory model will be enabled by default in Kotlin 1.7.20, and the old one will likely be deprecated in 1.8.0.
We expect to release `kotlinx.coroutines` 1.7.0 with Kotlin 1.8.0, so at this moment it will be no longer relevant

qwwdfsad:
Fixes #3375


qwwdfsad:
I see, will fix `kotlinx-coroutines-test` as well

kotlinx-coroutines-test/common/test/RunTestTest.kt
```diff
@@ -112,7 +110,7 @@ class RunTestTest {
                 it()
                 fail("unreached")
             } catch (e: UncompletedCoroutinesError) {
-                assertTrue((e.message ?: "").contains(name1))
+                assertContains(e.message ?: "", name1)
```
qwwdfsad, line 113: Filed https://youtrack.jetbrains.com/issue/KT-53336/Improve-assertContains-and-friends-provide-negation-and-nullability-support

dkhalanskyjb (changes requested):
I think we should be a bit more thorough here. In particular, I think it's worth it to review all files that use `kotlin.native.concurrent` and get rid of the hacks used to support the old memory model.

I went through the codebase, and here are the results: https://gist.github.com/dkhalanskyjb/ed533dfbaff5b5d39cba8bcc43f13921 Lots of unused imports were eliminated at least.

Of the remaining uses of `kotlin.native.concurrent`, all seem useful to me, except for `kotlinx-coroutines-core/common/src/EventLoop.common.kt`. I don't know that code well, so I'm not sure. Is there a benefit to have `ThreadLocalEventLoop` and it being `@ThreadLocal`? Sure does look like a measure put in place for the old memory model.

kotlinx-coroutines-test/build.gradle.kts
```diff
@@ -10,4 +12,10 @@ val experimentalAnnotations = listOf(
 
 kotlin {
     sourceSets.all { configureMultiplatform() }
+
+    targets.withType(KotlinNativeTargetWithTests::class.java).configureEach {
+        val tests = binaries.getTest("DEBUG")
+        tests.optimized = true
```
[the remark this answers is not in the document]
qwwdfsad: It's not needed, but it's convenient:

* Tests are faster
* It helps with finding bugs in K/N optimizer (i.e. these tests are run on kotlinx.train) which is far more important

qwwdfsad:
>. Is there a benefit to have ThreadLocalEventLoop and it being @ThreadLocal? 

I've dropped a comment

kotlinx-coroutines-test/build.gradle.kts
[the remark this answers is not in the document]
dkhalanskyjb: I get the second point, but with how slow the Native test compilation is now, I doubt the tests are actually faster to run in practice.

kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -123,6 +123,10 @@ internal abstract class EventLoop : CoroutineDispatcher() {
     open fun shutdown() {}
 }
 
+/*
+ * This is marked with `ThreadLocal` on purpose: the loop itself is supposed to be thread-local,
+ * but `CommonThreadLocal` is not supported on K/N properly, so we emulate it with this annotation.
+ */
```
dkhalanskyjb, line 129: `CommonThreadLocal` is quite a leaky abstraction then. Maybe we should fix it instead? Here's my attempt to do it: https://gist.github.com/dkhalanskyjb/7cc2d6c752d97217c6c288304253ba11

kotlinx-coroutines-test/build.gradle.kts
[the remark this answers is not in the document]
qwwdfsad: It is important for stress tests in core, but here it's unlikely to make any difference.
If you insist or have a strong opinion why it shouldn't be enabled, then I do not mind to revert it.

kotlinx-coroutines-core/common/src/EventLoop.common.kt
qwwdfsad: In general, I prefer to extract abstraction (if we are talking about internal API of course) only when it has at least two uses because otherwise, it's much easier to either over-abstract or write too much code for no good.

But it is already here, so I've pushed the change

dkhalanskyjb: Agreed: if there was no `CommonThreadLocal`, I wouldn't propose one.

dkhalanskyjb (approved):

kotlinx-coroutines-core/common/src/EventLoop.common.kt
dkhalanskyjb: BTW, https://stackoverflow.com/a/50329788

PR #3380, opened 2022-07-27
Introduce a rendezvous on CancellableContinuationImpl.parentHandle to…

qwwdfsad:
… avoid race with original thread setting parent handle, while another thread nulls it out during cancellation via detachChild

Fixes on of the races found in #3378


qwwdfsad:
Both JVM and K/N are affected, though I've failed to find any potential harm from this race.
The only scenario that comes to my mind is that the published CC instance may retain its [cancelled parent in memory, while CC itself is not cleaned

elizarov (approved):

PR #3384, opened 2022-07-29
Introduce Job.parent API

### LINKED ISSUE #3201: [DEBUG-IMPROVEMENT] expose `job.parentJob`
currently `job.parentJob` is not exposed, making it harder to build a parent-child relational job graph when debug
according to this https://github.com/Kotlin/kotlinx.coroutines/issues/3198 it should be fairly easy(and safe?) to expose that

qwwdfsad:
* The API is crucial for debugger extension and deadlock detection
* It enables allows more fluent coroutines hierarchy inspection, e.g. capability to build a list of roots and procrss them top-bottom separately

Fixes #3201

### DIFF
--- a/kotlinx-coroutines-core/common/src/Job.kt
+++ b/kotlinx-coroutines-core/common/src/Job.kt
@@ -125,7 +125,7 @@ public interface Job : CoroutineContext.Element {
      * as the job is transitioned to its final state, whether it is cancelled or completed,
      * and all job children are completed.
      *
-     * For coroutines, its corresponding job completes as soon as the coroutine itself
+     * For a coroutine, its corresponding job completes as soon as the coroutine itself
      * and all its children are complete.
      *
      * @see [Job] state transitions for additional details.
--- a/kotlinx-coroutines-core/common/src/Job.kt
+++ b/kotlinx-coroutines-core/common/src/Job.kt
@@ -31,7 +31,7 @@ import kotlin.jvm.*
  *   It is completed by calling [CompletableJob.complete].
  *
  * Conceptually, an execution of a job does not produce a result value. Jobs are launched solely for their
- * side-effects. See [Deferred] interface for a job that produces a result.
+ * side effects. See [Deferred] interface for a job that produces a result.
  *
  * ### Job states
  *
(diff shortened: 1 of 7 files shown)


kotlinx-coroutines-core/common/src/NonCancellable.kt
```diff
@@ -29,6 +29,14 @@ public object NonCancellable : AbstractCoroutineContextElement(Job), Job {
 
     private const val message = "NonCancellable can be used only as an argument for 'withContext', direct usages of its API are prohibited"
 
+    /**
+     * Always returns `null`.
+     * @suppress **This an internal API and should not be used from general code.**
+     */
+    @Deprecated(level = DeprecationLevel.WARNING, message = message)
```
dkhalanskyjb, line 36: Does it make sense to add `ReplaceWith(null)` to allow for automatic replacements in bulk? I don't see the downsides.

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -117,6 +117,22 @@ public interface Job : CoroutineContext.Element {
 
     // ------------ state query ------------
 
+    /**
+     * Returns a parent of the current job if the parent-child relationship
```
dkhalanskyjb, line 121:
```suggestion
     * Returns the parent of the current job if the parent-child relationship
```
IIRC, the parent, if it exists, is unique?

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -117,6 +117,22 @@ public interface Job : CoroutineContext.Element {
...
+     * is established or `null` if the job has no parent or was successfully completed.
+     *
+     * Accesses to this property are not idempotent, the property becomes `null` as soon
+     * as the job is transitioned to its final state, whether it is cancelled or completed,
+     * and all job children are completed.
```
dkhalanskyjb, line 126:
```suggestion
     * Accesses to this property are not idempotent: the property becomes `null` as soon
     * as the job is transitioned to its final state—whether it is cancelled or completed—
     * and all job children are completed.
```
Do we allow Unicode in comments?

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -117,6 +117,22 @@ public interface Job : CoroutineContext.Element {
...
+     *
+     * For coroutines, its corresponding job completes as soon as the coroutine itself
```
dkhalanskyjb, line 128:
```suggestion
     * For a coroutine, its corresponding job completes as soon as the coroutine itself
```

dkhalanskyjb (approved):

kotlinx-coroutines-core/common/src/NonCancellable.kt
qwwdfsad: Not really, it rather means that the whole usage of this API is incorrect and it's time to rethink what is written :)

kotlinx-coroutines-core/common/src/Job.kt
qwwdfsad: I would prefer to avoid it, we never know whether Dokka and kotlinlang support it

dkhalanskyjb: This argument cuts both ways: we may treat this as a good chance to find out. I don't care that much though.

PR #3388, opened 2022-08-03
Provide a way for customization Dokka's HTML output

nikpachoo:
We want to customize the header and the footer in coroutines API reference on kotlinlang.org to provide a better user experience and consistency between all parts of kotlinlang.org. 

Starting from the 1.6.20 version, Dokka supports customization of HTML output. You can read about it [here](https://kotlin.github.io/dokka/1.6.20/user_guide/base-specific/frontend/#custom-html-pages).
To use this opportunity, we need a small preparation on your side.
These changes allow HTML customization. 
The rest we will do on our side.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -355,3 +355,7 @@ allprojects { subProject ->
                 }
             }
 }
+
+tasks.named("dokkaHtmlMultiModule") {
+    pluginsMapConfiguration.set(["org.jetbrains.dokka.base.DokkaBase": """{ "templatesDir": "${projectDir.toString().replace('\\', '/')}/dokka-templates" }"""])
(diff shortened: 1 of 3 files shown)


qwwdfsad (changes requested):
Build failed twice in a row on Windows with Dokka-specific issue: https://teamcity.jetbrains.com/viewType.html?buildTypeId=KotlinTools_KotlinxCoroutines_BuildAggregated&branch_KotlinTools_KotlinxCoroutines=ktl-695-dokka-custom-templates&tab=buildTypeStatusDiv

Please check it out

qwwdfsad (approved):

PR #3389, opened 2022-08-03
Fix documentation for `DEBUG_PROPERTY_VALUE_OFF`

Lukellmann:
Based against `master` as this PR only fixes documentation and doesn't change line numbers.


qwwdfsad (approved):
Good catch, thanks!

PR #3391, opened 2022-08-04
Introduce Java Object Layout to the codebase and write tests for core…

qwwdfsad:
… primitives

It should further simplify reasoning about new fields in core classes


qwwdfsad:
@dkhalanskyjb pinging you so you are aware we are doing this

PR #3393, opened 2022-08-05
Introduce non-nullable types in reactive integrations where appropriate

qwwdfsad:
* For RxJava2, use them in internal implementations where appropriate
* For RxJava3, introduce & Any bound to generic argument in our extensions to avoid errors in Kotlin 1.8.0 due to non-nullability rx3 annotations being part of generics upper bound. This change went through commitee, and all the "unsound" declarations such as "RxSignature<Foo?>" were properly highlighted as warning that will become an error.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -176,8 +176,7 @@ configure(subprojects.findAll { !sourceless.contains(it.name) }) {
     tasks.withType(AbstractKotlinCompile).all {
         kotlinOptions.freeCompilerArgs += OptInPreset.optInAnnotations.collect { "-Xopt-in=" + it }
         kotlinOptions.freeCompilerArgs += "-progressive"
-        // Disable KT-36770 for RxJava2 integration
-        kotlinOptions.freeCompilerArgs += "-XXLanguage:-ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated"
+        kotlinOptions.freeCompilerArgs += "-XXLanguage:+ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated"
         // Remove null assertions to get smaller bytecode on Android
         kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
     }
--- a/build.gradle
+++ b/build.gradle
@@ -176,7 +176,6 @@ configure(subprojects.findAll { !sourceless.contains(it.name) }) {
     tasks.withType(AbstractKotlinCompile).all {
         kotlinOptions.freeCompilerArgs += OptInPreset.optInAnnotations.collect { "-Xopt-in=" + it }
         kotlinOptions.freeCompilerArgs += "-progressive"
-        kotlinOptions.freeCompilerArgs += "-XXLanguage:+ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated"
         // Remove null assertions to get smaller bytecode on Android
         kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
     }
--- a/reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
@@ -24,9 +24,17 @@ import kotlin.coroutines.*
  */
 public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine { cont ->
     subscribe(object : CompletableObserver {
-        override fun onSubscribe(d: Disposable) { cont.disposeOnCancellation(d) }
-        override fun onComplete() { cont.resume(Unit) }
-        override fun onError(e: Throwable) { cont.resumeWithException(e) }
+        override fun onSubscribe(d: Disposable) {
+            cont.disposeOnCancellation(d)
+        }
+
+        override fun onComplete() {
+            cont.resume(Unit)
+        }
+
+        override fun onError(e: Throwable) {
+            cont.resumeWithException(e)
+        }
     })
 }
 
@@ -43,12 +51,23 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
  */
 @Suppress("UNCHECKED_CAST")
 public suspend fun <T> MaybeSource<T>.awaitSingleOrNull(): T? = suspendCancellableCoroutine { cont ->
-    subscribe(object : MaybeObserver<T> {
-        override fun onSubscribe(d: Disposable) { cont.disposeOnCancellation(d) }
-        override fun onComplete() { cont.resume(null) }
-        override fun onSuccess(t: T) { cont.resume(t) }
-        override fun onError(error: Throwable) { cont.resumeWithException(error) }
-    })
+    subscribe(object : MaybeObserver<T & Any> {
+        override fun onSubscribe(d: Disposable) {
+            cont.disposeOnCancellation(d)
+        }
+
+        override fun onComplete() {
+            cont.resume(null)
+        }
+
+        override fun onSuccess(t: T & Any) {
+            cont.resume(t)
+        }
+
+        override fun onError(error: Throwable) {
+            cont.resumeWithException(error)
+        }
+    } as MaybeObserver<T>)
 }
 
 /**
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -176,8 +176,7 @@ configure(subprojects.findAll { !sourceless.contains(it.name) }) {
     tasks.withType(AbstractKotlinCompile).all {
         kotlinOptions.freeCompilerArgs += OptInPreset.optInAnnotations.collect { "-Xopt-in=" + it }
         kotlinOptions.freeCompilerArgs += "-progressive"
-        // Disable KT-36770 for RxJava2 integration
-        kotlinOptions.freeCompilerArgs += "-XXLanguage:-ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated"
+        kotlinOptions.freeCompilerArgs += "-XXLanguage:+ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated"
```
dkhalanskyjb, line 179: I don't see any difference between this line being present and not present. Isn't it included in `-progressive`?

reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
```diff
@@ -43,12 +51,23 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
  */
 @Suppress("UNCHECKED_CAST")
 public suspend fun <T> MaybeSource<T>.awaitSingleOrNull(): T? = suspendCancellableCoroutine { cont ->
-    subscribe(object : MaybeObserver<T> {
-        override fun onSubscribe(d: Disposable) { cont.disposeOnCancellation(d) }
-        override fun onComplete() { cont.resume(null) }
-        override fun onSuccess(t: T) { cont.resume(t) }
-        override fun onError(error: Throwable) { cont.resumeWithException(error) }
-    })
+    subscribe(object : MaybeObserver<T & Any> {
```
dkhalanskyjb, line 54:
```suggestion
    subscribe(object : MaybeObserver<T> {
```
Then, the cast can be removed, along with the suppression.

reactive/kotlinx-coroutines-rx2/src/RxAwait.kt
```diff
@@ -117,12 +136,21 @@ public suspend fun <T> MaybeSource<T>.awaitOrDefault(default: T): T = awaitSingl
...
  * function immediately disposes of its subscription and resumes with [CancellationException].
  */
+@Suppress("UNCHECKED_CAST")
 public suspend fun <T> SingleSource<T>.await(): T = suspendCancellableCoroutine { cont ->
-    subscribe(object : SingleObserver<T> {
-        override fun onSubscribe(d: Disposable) { cont.disposeOnCancellation(d) }
-        override fun onSuccess(t: T) { cont.resume(t) }
-        override fun onError(error: Throwable) { cont.resumeWithException(error) }
-    })
+    subscribe(object : SingleObserver<T & Any> {
```
dkhalanskyjb, line 141: Ditto.

reactive/kotlinx-coroutines-rx2/src/RxChannel.kt
```diff
@@ -31,22 +31,24 @@ public suspend inline fun <T> ObservableSource<T>.collect(action: (T) -> Unit):
...
     val channel = SubscriptionChannel<T>()
-    subscribe(channel)
+    (this as ObservableSource<T & Any>).subscribe(channel)
     return channel
 }
 
 @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 private class SubscriptionChannel<T> :
-    LinkedListChannel<T>(null), Observer<T>, MaybeObserver<T>
+    LinkedListChannel<T>(null), Observer<T & Any>, MaybeObserver<T & Any>
```
dkhalanskyjb, line 51: Ditto: the casts above will not be needed if these bounds are removed.

dkhalanskyjb: But this change *is* required for Rx3. Shouldn't `RxChannel` be changed for rx3 as well?

dkhalanskyjb: Also, rx3's `RxMaybe`, too, gives me warnings.

dkhalanskyjb (changes requested):

reactive/kotlinx-coroutines-rx2/src/RxChannel.kt
```diff
@@ -31,16 +31,18 @@ public suspend inline fun <T> ObservableSource<T>.collect(action: (T) -> Unit):
     toChannel().consumeEach(action)
 
 @PublishedApi
+@Suppress("UNCHECKED_CAST")
 internal fun <T> MaybeSource<T>.toChannel(): ReceiveChannel<T> {
     val channel = SubscriptionChannel<T>()
-    subscribe(channel)
+    (this as MaybeSource<T & Any>).subscribe(channel)
```
dkhalanskyjb, line 37: The code works even without this cast now.

reactive/kotlinx-coroutines-rx2/src/RxChannel.kt
```diff
@@ -31,16 +31,18 @@ public suspend inline fun <T> ObservableSource<T>.collect(action: (T) -> Unit):
...
     return channel
 }
 
 @PublishedApi
+@Suppress("UNCHECKED_CAST")
 internal fun <T> ObservableSource<T>.toChannel(): ReceiveChannel<T> {
     val channel = SubscriptionChannel<T>()
-    subscribe(channel)
+    (this as ObservableSource<T & Any>).subscribe(channel)
```
dkhalanskyjb, line 45: Ditto.

dkhalanskyjb (approved):

reactive/kotlinx-coroutines-rx3/src/RxAwait.kt
```diff
@@ -41,9 +41,8 @@ public suspend fun CompletableSource.await(): Unit = suspendCancellableCoroutine
...
  * function immediately resumes with [CancellationException] and disposes of its subscription.
  */
-@Suppress("UNCHECKED_CAST")
-public suspend fun <T> MaybeSource<T>.awaitSingleOrNull(): T? = suspendCancellableCoroutine { cont ->
-    subscribe(object : MaybeObserver<T> {
+public suspend fun <T> MaybeSource<T & Any>.awaitSingleOrNull(): T? = suspendCancellableCoroutine { cont ->
+    subscribe(object : MaybeObserver<T & Any> {
         override fun onSubscribe(d: Disposable) { cont.disposeOnCancellation(d) }
         override fun onComplete() { cont.resume(null) }
         override fun onSuccess(t: T) { cont.resume(t) }
```
cpovirk, line 48: `T & Any`

reactive/kotlinx-coroutines-rx3/src/RxChannel.kt
```diff
@@ -54,12 +54,11 @@ public suspend inline fun <T> MaybeSource<T>.collect(action: (T) -> Unit): Unit
...
  * [collect]. Also, if the [ObservableSource] signals an error, that error is rethrown from [collect].
  */
-public suspend inline fun <T> ObservableSource<T>.collect(action: (T) -> Unit): Unit =
-    openSubscription().consumeEach(action)
+public suspend inline fun <T> ObservableSource<T & Any>.collect(action: (T) -> Unit): Unit = openSubscription().consumeEach(action)
 
 @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 private class SubscriptionChannel<T> :
-    LinkedListChannel<T>(null), Observer<T>, MaybeObserver<T>
+    LinkedListChannel<T>(null), Observer<T & Any>, MaybeObserver<T & Any>
```
cpovirk, line 61: `T & Any` for `onSuccess` and `onNext`, too.

build.gradle
```diff
@@ -176,8 +176,6 @@ configure(subprojects.findAll { !sourceless.contains(it.name) }) {
     tasks.withType(AbstractKotlinCompile).all {
         kotlinOptions.freeCompilerArgs += OptInPreset.optInAnnotations.collect { "-Xopt-in=" + it }
         kotlinOptions.freeCompilerArgs += "-progressive"
-        // Disable KT-36770 for RxJava2 integration
-        kotlinOptions.freeCompilerArgs += "-XXLanguage:-ProhibitUsingNullableTypeParameterAgainstNotNullAnnotated"
         // Remove null assertions to get smaller bytecode on Android
         kotlinOptions.freeCompilerArgs += ["-Xno-param-assertions", "-Xno-receiver-assertions", "-Xno-call-assertions"]
```
cpovirk, line 180: I'm not sure exactly what [`-progressive`](https://kotlinlang.org/docs/compiler-reference.html#progressive) enables, but I notice that I can get a few more errors (noted below) if I enable the flags from #3007. (Thanks for pointing me here from there!)

```
        // Recognize rxjava3 nullness annotations even before that becomes the default (which may happen in 1.8): https://kotlinlang.org/docs/java-interop.html#nullability-annotations
        kotlinOptions.freeCompilerArgs += "-Xnullability-annotations=@io.reactivex.rxjava3.annotations:strict"
        // Recognize nullness annotations on type arguments, etc.: https://kotlinlang.org/docs/java-interop.html#annotating-type-arguments-and-type-parameters
        kotlinOptions.freeCompilerArgs += "-Xtype-enhancement-improvements-strict-mode"
```

cpovirk (commented):
Thanks for making this happen!

reactive/kotlinx-coroutines-rx3/src/RxAwait.kt
qwwdfsad: This bound seems to be properly propagated from the declaration site

cpovirk: Since the object is implementing `MaybeObserver<T & Any>`, I think the method parameter needs to be `T & Any` to match.

Or you know what, actually, even if we were implementing plain `MaybeObserver<T>`, we'd probably still need the method parameter to be `T & Any`, since the interface declares its method as requiring a `@NonNull T`:

https://github.com/ReactiveX/RxJava/blob/8e53e0ee1637791923a81fdb9d17dfe8c1569250/src/main/java/io/reactivex/rxjava3/core/MaybeObserver.java#L76

I think that's why, with the current parameter type of `T`, the build fails with the flags I posted in the build script.

qwwdfsad: Thanks for pointing it out. T & Any is required since Kotlin 1.8.0, it was an omission on the language site

PR #3421, opened 2022-08-23
New implementation of K/N multi-threaded dispatchers 

qwwdfsad:


kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -63,28 +64,54 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+             */
             task.run()
         }
     }
 
     override fun dispatch(context: CoroutineContext, block: Runnable) {
-        // TODO handle rejections
-        tasksQueue.trySend(block)
+        if (activeWorkers.value != workersCount) {
+            tryAddWorker()
```
qwwdfsad, line 86: I would really like to provide a proper keep-alive here, but it doesn't seem adequately possible: we **must** terminate workers and we **must** await termination in a blocking way, otherwise native resources leak.

The preliminary decision is to get rid of workers and design something better (i.e. `Thread`) in K/N

qwwdfsad:
Long story short, I was trying to rewrite the dispatcher with proper locks (i.e. https://github.com/Kotlin/kotlinx.coroutines/commit/c414dce40d749436d1681f8fac8ad47fe2f2b17b) in order to make it faster.

Apparently, zero-contention cases indeed become significantly faster, but any contention will kill all the dispatcher performance (I mean it, orders of magnitude). In order to avoid it, we need something that barely resembles a scalable lock (i.e. `AbstractQueuedSynchronizer`). The closest thing to it is, surprise, a channel :)

So we invested around ~10 human hours (I've also included K/N folk to help me and avoid some stupid gotcha) to keep the code the way it is :) 

I do have ideas how to get rid of channel overhead for uncontended case, but it doesn't seem to be worth it

qwwdfsad:
On a side-note, the documentation to the native declaration of `newFixedThreadPoolContext` will arrive later, first, we'll merge it, and I'll create `Dispatchers.IO` on top of that and only then KDoc will follow

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -63,28 +64,54 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+                return
+            }
+        }
     }
 
     override fun close() {
         tasksQueue.close()
-        workers.forEach { it.requestTermination().result }
+        /*
+         * Here we cannot avoid not to wait on `.result`, otherwise it will lead
```
dkhalanskyjb, line 109:
```suggestion
         * Here we cannot avoid waiting on `.result`, otherwise it will lead
```

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -63,28 +64,54 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+         * to a native memory leak, including pthread handle.
```
dkhalanskyjb, line 110:
```suggestion
         * to a native memory leak, including a pthread handle.
```

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -63,28 +64,54 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+         */
+        val requests = Array(workersCount) {
+            workers[it].value?.requestTermination()
```
dkhalanskyjb, line 113: Maybe I'm missing something, but isn't there still a race between this line and the assignment of a worker to the array, leading to a possible leak?

Suggestion:
* designate a bit in `activeWorkers` for the dispatcher being closed or not,
* only allow updating `activeWorkers` when the bit is not set,
* when terminating the workers, traverse the `activeWorkers` array elements and spin on each one because someone is guaranteed to add a worker there.

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
dkhalanskyjb, line 86: I don't understand this whole approach. So, each `MultiWorkerDispatcher` has the target number of workers. On each dispatch, we check if this number was reached, and if not, we create a new worker—even if all the existing ones are idle. So, in effect, several dispatches later, the `workersCount` number of workers will be reached. So, why bother at all and not just spawn all the workers immediately, as was done before? I guess the initialization phase will be faster, but is it by much?

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
qwwdfsad: >I guess the initialization phase will be faster, but is it by much?

To avoid paying the cost of the initialization by the code that does not use it.
E.g. I'm going to introduce `Dispatchers.IO` with 64 workers capacity. We definitely do not want to start 64 workers for the very first use of `Dispatchers.Default` or `Dispatchers.Unconfined`.

If I find a spare time, I'll try to introduce both idle-check and withTimeout-based idle worker termination prior to 1.8.0. 
The former is likely to happen, the latter is unlikely to happen tho.

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -67,28 +68,81 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+    private val workers = atomicArrayOfNulls<Worker>(workersCount)
+
+    // Number of active workers + isClosed flag in the highest bit
+    private val controlState = atomic(0)
+    private val activeWorkers: Int get() = controlState.value and (IS_CLOSED_MASK.inv())
+    private var isClosed: Boolean
+        get() = controlState.value.isClosed()
+        set(value) {
+            assert { value }
+            controlState.value = controlState.value or IS_CLOSED_MASK
```
dkhalanskyjb, line 87: We may still leak some workers this way.
1. `controlState.value` is read in the `isClosed = true` statement.
2. Some new workers arrive. They increment `controlState.value` and do nothing else.
3. The old `controlState.value` is `or`-ed with `IS_CLOSED_MASK` and written to `controlState.value`.
4. The cleanup procedure is done.
5. The workers add themselves to the array and begin to do their thing.

qwwdfsad: Good catch, thanks

dkhalanskyjb:
I am wary now that we found two races in the implementation, so I extracted the data structure and added a test for it. Sure, it's only used in one place, but the benefit of having that code tested (and also a bit more transparent, IMHO) outweighs this, I think.

kotlinx-coroutines-core/common/src/internal/OnDemandAllocatingPool.kt
```diff
@@ -0,0 +1,101 @@
...
+
+package kotlinx.coroutines.internal
+
+import kotlinx.atomicfu.*
+
+/**
+ * A thread-safe resource pool.
+ *
+ * [maxCapacity] is the maximum amount of elements.
+ * [create] is the function that creates a new element.
```
qwwdfsad, line 13:
```suggestion
 * [create] is the function that creates a new element.
 * It is part of 'common' source-set in order to be testable on JVM with Lincheck
```

qwwdfsad: Also, probably can be moved to 'concurrent'?

kotlinx-coroutines-core/common/src/internal/OnDemandAllocatingPool.kt
```diff
@@ -0,0 +1,101 @@
...
+ */
+internal class OnDemandAllocatingPool<T>(
+    private val maxCapacity: Int,
+    private val create: (Int) -> T
+) {
+    /** Number of existing elements + isClosed flag in the highest bit.
+     * Once the flag is set, the value is guaranteed not to change anymore. */
```
qwwdfsad, line 20: I would stick with the default Java convention, that autoformats this comment into the following form:
```
/*
 * Comment
 */
```
I've stressed our (Kotlin) auto-formatter for a while, and it seems like it just doesn't work at all, I'll file all the bugs separately

qwwdfsad (commented):
Otherwise good to go, cannot approve though due to GH limitations

dkhalanskyjb (approved):

PR #3431, opened 2022-09-02
Ensure `awaitFrame()` only awaits a single frame when used from the main thread

pablobaxter:
Fix to resolve https://github.com/Kotlin/kotlinx.coroutines/issues/3432

### DIFF
--- a/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
+++ b/ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
@@ -190,11 +190,14 @@ public suspend fun awaitFrame(): Long {
             postFrameCallback(choreographer, cont)
         }
     }
-    // post into looper thread to figure it out
     return suspendCancellableCoroutine { cont ->
-        Dispatchers.Main.dispatch(EmptyCoroutineContext, Runnable {
+        if (Looper.myLooper() === Looper.getMainLooper()) { // Check if we are already in the main looper thread
             updateChoreographerAndPostFrameCallback(cont)
-        })
+        } else { // post into looper thread to figure it out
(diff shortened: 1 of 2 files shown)


ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -190,11 +190,14 @@ public suspend fun awaitFrame(): Long {
             postFrameCallback(choreographer, cont)
         }
     }
-    // post into looper thread to figure it out
     return suspendCancellableCoroutine { cont ->
-        Dispatchers.Main.dispatch(EmptyCoroutineContext, Runnable {
+        if (Looper.myLooper() === Looper.getMainLooper()) { // Check if we are already in the main looper thread
```
qwwdfsad, line 194: Thanks! 

I'm not yet convinced this should be **the** solution though.
I would expect that `Dispatchers.Main.immediate` would be enough, but it is not due to our implementation details.

We'll fix it in the next release, but it is an open question whether we are going to use this particular approach though

pablobaxter: No probs!

The only reason I took the `Looper` approach here is because this code is specific to Android, and I didn't see anything similar for JavaFX or Swing implementations (`awaitPulse()` was the closest parallel). I see no problems and I am not opposed to using `Dispatchers.Main.immediate`.

qwwdfsad: It seems to be the way though!

I've opened https://github.com/Kotlin/kotlinx.coroutines/pull/3437 with a few changes on top of yours that I'll squash into your commit. Thanks for the contribution, we'll merge it in the next release!

PR #3437, opened 2022-09-05
Ensure awaitFrame() only awaits a single frame when used from the main looper

qwwdfsad:
See https://github.com/Kotlin/kotlinx.coroutines/pull/3431 for the additional context


ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -185,22 +185,27 @@ private var choreographer: Choreographer? = null
...
+        awaitFrameSlowPath()
     }
-    // post into looper thread to figure it out
-    return suspendCancellableCoroutine { cont ->
-        Dispatchers.Main.dispatch(EmptyCoroutineContext, Runnable {
+}
+
+private suspend fun awaitFrameSlowPath(): Long = suspendCancellableCoroutine { cont ->
+    if (Looper.myLooper() === Looper.getMainLooper()) { // Check if we are already in the main looper thread
+        updateChoreographerAndPostFrameCallback(cont)
```
dkhalanskyjb, line 199: I saw a discussion about `Dispatchers.Main.immediate` not being good enough for this and something about intending to fix it. Could you elaborate?

ui/kotlinx-coroutines-android/test/HandlerDispatcherTest.kt
```diff
@@ -109,14 +109,14 @@ class HandlerDispatcherTest : TestBase() {
...
         expect(2)
         // Run choreographer detection
         mainLooper.runOneTask()
-        expect(3)
-        mainLooper.scheduler.advanceBy(50, TimeUnit.MILLISECONDS)
         expect(4)
+        mainLooper.scheduler.advanceBy(50, TimeUnit.MILLISECONDS)
+        expect(5)
         mainLooper.scheduler.advanceBy(51, TimeUnit.MILLISECONDS)
         finish(6)
```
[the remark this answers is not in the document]
qwwdfsad: Does not seem so

ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
qwwdfsad: Well, I was hoping that adding `Dispatchers.main.immediate` would solve the problem, but it won't due to how we do structure our code: `CancellableContinuation` already incapsulates the dispatcher (for this case -- `Dispatchers.Main`) and we cannot "just" resume it in place.

Neither can we do `Dispatchers.Main.immediate.resumeUndispatched(cont)` due to how it is implemented -- that's a grey area in "how should 'resumeUndispatched' behave if extension receiver and actual dispatcher do not match". And they do not match because `Dispatchers.Main.immediate` and `Dispatchers.Main` are different dispatchers, non-transparent to each other

ui/kotlinx-coroutines-android/test/HandlerDispatcherTest.kt
```diff
@@ -116,9 +116,7 @@ class HandlerDispatcherTest : TestBase() {
         mainLooper.runOneTask()
         expect(4)
         mainLooper.scheduler.advanceBy(50, TimeUnit.MILLISECONDS)
-        expect(5)
-        mainLooper.scheduler.advanceBy(51, TimeUnit.MILLISECONDS)
-        finish(6)
+        finish(5)
```
dkhalanskyjb, line 119: Then it's `finish(4)`, as the action after `expect(4)` also does nothing.

qwwdfsad: Done

PR #3438, opened 2022-09-05
Replace hand-rolled ArrayQueue with ArrayDeque in standard library in …

qwwdfsad:
…order to (potentially) reduce dex size

Move ArrayQueue to JS-specific source-set to avoid touching JS


dkhalanskyjb:
I get the point of not touching the JS part, but in this case, the changes required seem completely trivial, and removing any code is a win in my book. https://gist.github.com/dkhalanskyjb/419d1bb27866dc3c6f8e9bf06b751100

qwwdfsad:
Good point, done

dkhalanskyjb (approved):

PR #3446, opened 2022-09-08
Get rid of workaround for KT-16222 that is long time fixed

qwwdfsad:
The correctness of the change is verified by ConsumeAsFlowLeakTest


qwwdfsad:
And a bit of warnings removal here and there

dkhalanskyjb (approved):
(Irrelevant musings) Before I read the actual code we're removing, I thought that this PR was just a trojan horse containing a clean-up of warnings and KT-16222 is just a pretense, but then, oh, wow, this is nasty. Great call purging it!

Issue #1205, opened 2019-05-17
TestCoroutineDispatcher swallows exceptions


manuelvicnt:
@julioyg found an interesting bug(?) regarding the TestCoroutineDispatcher.

The fact that `runBlockingTest` uses `async` to run the test body ([TestBuilders.runBlockingTest#L49](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-test/src/TestBuilders.kt#L49)), by the time it tries to throw an exception, the test has already completed.

Problems:
1) Silent exceptions
If the coroutine inside a function throws an exception, the test passes and the exception is thrown silently to the console.

2) Tests that expect an exception to be thrown
Consider the following scenario. Since we're running this in a blocking way, the expectation is that the exception is thrown, but it is not.

```
class MyViewModel(): ViewModel() {
   fun throwException() {
        viewModelScope.launch {
            throw Exception("Not found")
        }
    }
}
```

```
class MyViewModelTest {

   @get:Rule
   var coroutinesTestRule = CoroutinesTestRule()

   @Test
   fun testExceptionThrown() = coroutinesTestRule.testDispatcher.runBlockingTest {
        val assertionThrown = false
        try {
            val viewModel = MyViewModel()
            viewModel.throwException()
        } catch(e: Exception) {
            assertionThrown = true
        }    
        assertTrue(assertionThrown)
   }
}
```

manuelvicnt:
cc @objcode

objcode:
Ah so this one is a bit weird - since you're not passing the job from the TestCoroutineScope the exception is handled by viewModelScope.

This might be more properly a bug against the viewModelScope implementation (allow the injection of parent scope) WDYT?

objcode:
Two options here:

1. Add test-injection of uncaught exception handler to viewModelScope for a specific viewModel
2. Add a global uncaught exception handler on CoroutineScope (CoroutineScope.addGlobalUncaughtExceptionHandler) - this mirrors the thread APIs. However, it's really quite global (and note that it's not normal to set the thread version in test code because it breaks test parallelism)

yigit:
I've not thought about the whole problem but there is a flaw in that test code.

```
class MyViewModel(): ViewModel() {
   fun throwException() {
        viewModelScope.launch {
            throw Exception("Not found")
        }
    }
}
```

You are calling a regular function and launching inside it. It is async, there is no reason for your test to wait for it to finish. 

[...]

PR #3449, opened 2022-09-12
In the test module, collect and report the actual uncaught exceptions

dkhalanskyjb:
An attempt to mitigate #1205.


dkhalanskyjb:
A draft because it's untested and has (a fairly unpleasant, it seems?) TODO. @qwwdfsad, what do you think of the general approach?

kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+internal expect fun propagateExceptionToPlatform(context: CoroutineContext, exception: Throwable)
+
+/**
+ * If [addOnExceptionCallback] is called, the provided callback will be evaluated each time
+ * [handleCoroutineException] is executed and can't find a [CoroutineExceptionHandler] to
+ * process the exception.
+ *
+ * When a callback is registered once, even if it's later removed, the system starts to assume that
+ * other callbacks will eventually be registered, and so collects the exceptions.
+ * Once a new callback is registered, it tries.
```
qwwdfsad, line 21: it tries...?

kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+             * it's not enough to have the exception processed by one of them.
+             * The reason is, it's less big of a deal to observe multiple concurrent reports of bad behavior than not
+             * to observe the report in the exact callback that is connected to that bad behavior. */
+        }
+        return executedACallback
+    }
+}
+
+internal fun handleUncaughtCoroutineException(context: CoroutineContext, exception: Throwable) {
+    if (ANDROID_DETECTED) {
```
qwwdfsad, line 88: I got the idea behind this guard-check, but it still confuses me.

Can we instead guard it by, forgive me, Father, global mutable atomic boolean that the test module switches on? 
In addition, we can add `assumenosideeffects EXCEPTIONS_TEST_MODULE_FLAG false` to save Android a few bytes in dex

qwwdfsad (commented):
I'm not really happy with the approach, but it seems like we have no other (reasonable) options.

I'll revisit it this week along with an evaluation of user requests

kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.kt
dkhalanskyjb: Fixed.

kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.kt
dkhalanskyjb: Do the proguard rules work correctly with atomic values?

Also, the guard check is needed solely for the Android to save a few bytes, it has no significance other than that. I dropped a comment to clarify this.

qwwdfsad: They don't, but you do not have to expose the atomic, it can be a property with custom get/set and atomic as backing field

qwwdfsad: Could you please add a corresponding check to our r8 test?
It probably should be `-checkdiscard kotlinx.coroutines.internal.ExceptionCollector`

dkhalanskyjb:
@qwwdfsad, WDYT about this big rework? This is much less intrusive than the previous version, but it is still a bit unpleasant due to `CoroutineExceptionHandler` not being flexible enough. I described my thoughts about the underlying issue with `CoroutineExceptionHandler` here: https://github.com/Kotlin/kotlinx.coroutines/pull/3554

kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2016-2022 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.*
+import kotlin.coroutines.*
+
```
qwwdfsad, line 9: We have a different naming scheme for expect/actuals: common files are named as `Foo.common.kt`, while platform-specific files are named `Foo.kt`.

Apart from consistency, it will help to reduce the diff and make historic git-digging much easier

qwwdfsad (commented):
I like the idea in general.

I'll take a final look after the rename and I believe it is good to go

qwwdfsad:
Let's finalize and merge it?

qwwdfsad:
(preliminary looks good, will take a one final look when #3603 is merged)

qwwdfsad:
Could you please rebase this one on the recent `develop`?

dkhalanskyjb:
Sure, done.

kotlinx-coroutines-test/jvm/resources/META-INF/services/kotlinx.coroutines.CoroutineExceptionHandler
```diff
@@ -0,0 +1 @@
+kotlinx.coroutines.test.internal.ExceptionCollectorAsService
```
qwwdfsad, line 1: This one should have the corresponding `provides` in `module-info.java`

kotlinx-coroutines-test/common/src/internal/ExceptionCollector.kt
```diff
@@ -0,0 +1,98 @@
...
+ * process the exception.
+ *
+ * When a callback is registered once, even if it's later removed, the system starts to assume that
+ * other callbacks will eventually be registered, and so collects the exceptions.
+ * Once a new callback is registered, the collected exceptions are used with it.
+ *
+ * The callbacks in this object are the last resort before relying on platform-dependent
+ * ways to report uncaught exceptions from coroutines.
+ */
+internal object ExceptionCollector: AbstractCoroutineContextElement(CoroutineExceptionHandler), CoroutineExceptionHandler {
```
qwwdfsad, line 23:
```suggestion
internal object ExceptionCollector : AbstractCoroutineContextElement(CoroutineExceptionHandler), CoroutineExceptionHandler {
```

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -251,6 +259,8 @@ internal class TestScopeImpl(context: CoroutineContext) :
     fun legacyLeave(): List<Throwable> {
         val exceptions = synchronized(lock) {
             check(entered && !finished)
+            /** After [finished] becomes `true`, it is no longer valid to have [reportException] as the callback. */
+            ExceptionCollector.removeOnExceptionCallback(lock)
```
qwwdfsad, line 263: Callbacks are never removed for non-legacy version, consuming the memory lineary to number of tests run, please address that

qwwdfsad (approved):
LGTM module issues found.

While working on that, a few iterations were done, and it's easy to forget the reasning for the final solution over time. Issue fixed is also quite heavy on the discussion and potential solutions. While squashing, could you please provide an extensive commit message with the explanation of refactoring's rationale and problem solved?

dkhalanskyjb:
I squashed the commits into one and wrote a detailed commit message. Did you have something like that in mind?

qwwdfsad:
Exactly, nicely done!

PR #3460, opened 2022-09-23
Use outputFileProperty instead of outputFile

ilgonmic:
Because of changes in API of compile task

### DIFF
--- a/gradle/test-mocha-js.gradle
+++ b/gradle/test-mocha-js.gradle
@@ -23,7 +23,7 @@ def compileTestJsLegacy = tasks.hasProperty("compileTestKotlinJsLegacy")
 // todo: use atomicfu-transformed test files here (not critical)
 task testMochaNode(type: NodeTask, dependsOn: [compileTestJsLegacy, installDependenciesMochaNode]) {
     script = file("${node.nodeProjectDir.getAsFile().get()}/node_modules/mocha/bin/mocha")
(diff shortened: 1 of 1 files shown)


qwwdfsad (changes requested):
It does not build that implies breaking changes in the plugin.

Please elaborate on why these changes are necessary in the first place and whether the breaking changes are intended

ilgonmic:
Changes with migration from `outputFile` to `outputDirectory` in Gradle plugin are necessary, because can be situation, when users want to use complex module name e.g. (`@foo/bar`). And for these one `outputFile` migrated to 2 properties `outputDir` and `moduleName`.
Apart from that, `outputFile` was deprecated earlier with migration plan to `outputFileProperty` (because of Gradle property API), so there is a plan to leve only property-based inputs and outputs

qwwdfsad:
Thanks, but why is it non-buildable then?

I find it concerning that we have a deprecated API that has no replacement until Kotlin 1.8.0, where the original API is dropped

qwwdfsad:
@ilgonmic can you please provide an update regarding this one?

ilgonmic:
Yes, sorry

It has a replacement. `outputFileProperty` is the replacement, using Gradle Property API.
And it existed during deprecation cycle of `outputFile`

qwwdfsad:
I see, thanks. Please target the PR into `develop` branch and let release team rebase on it in such case

ilgonmic:
Done

PR #3475, opened 2022-10-08
Make use of `CompletionStage#handle` instead of `whenComplete`.

He-Pin:
Which leads a better performance.
https://github.com/line/armeria/pull/1440
https://github.com/akka/akka/pull/31283

### DIFF
--- a/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
+++ b/integration/kotlinx-coroutines-jdk8/src/future/Future.kt
@@ -40,7 +40,7 @@ public fun <T> CoroutineScope.future(
     val newContext = this.newCoroutineContext(context)
     val future = CompletableFuture<T>()
     val coroutine = CompletableFutureCoroutine(newContext, future)
-    future.whenComplete(coroutine) // Cancel coroutine if future was completed externally
+    future.handle(coroutine) // Cancel coroutine if future was completed externally
     coroutine.start(start, coroutine, block)
     return future
 }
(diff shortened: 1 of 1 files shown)


qwwdfsad (changes requested):
I've checked the implementation of `uniWhenCompleteStage` and it seems like the issue described in Akka/armeria is only applicable for `*Async` methods, not to regular ones.

I've also benchmarked the simplest `asCompletableFuture` and `future{}` builders and found no difference.

Taking into account that `handle` is semantically heavier than `when*` as it returns new value and potentially can change the result of the computation when used carelessly, it would be nice to see a benchmark that shows that this change yields a performance improvement prior to merge.

Code-wise changes looks good

He-Pin:
@qwwdfsad this link https://github.com/line/armeria/pull/1444, `handle` helps especially when the CompletionStage is failed.

He-Pin:
I will submit an jmh bench later。

He-Pin:
image

jmh:run -i 5 -wi 5 -f1 -t1 .CompletionStageBenchmark.

qwwdfsad (approved):
Thanks!

qwwdfsad:
Please rebase your PR on `develop` branch and change the target branch to `develop` as well

He-Pin:
Have rebased on to the `develop` and retarget `develop` branch.

qwwdfsad (approved):
Thanks!

PR #3477, opened 2022-10-11
Fix flakiness of TimeoutTest

qwwdfsad:
Increase timeout, so it's never triggered as these tests do not require it to race or trigger with the upstream

### DIFF
--- a/kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
@@ -99,16 +99,19 @@ class TimeoutTest : TestBase() {
     @Test
     fun testUpstreamErrorTimeoutException() = testUpstreamError(TimeoutCancellationException(0, Job()))
 
+    @Test
+    fun testUpstreamErrorCancellationException() = testUpstreamError(CancellationException(""))
+
     private inline fun <reified T: Throwable> testUpstreamError(cause: T) = runTest {
         try {
             // Workaround for JS legacy bug
             flow {
                 emit(1)
                 throw cause
-            }.timeout(1.milliseconds).collect()
+            }.timeout(1000.milliseconds).collect()
             expectUnreached()
         } catch (e: Throwable) {
-            assertTrue { e is T }
+            assertIs<T>(e)
             finish(1)
         }
     }


kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -99,16 +99,19 @@ class TimeoutTest : TestBase() {
...
                 throw cause
-            }.timeout(1.milliseconds).collect()
+            }.timeout(1000.milliseconds).collect()
             expectUnreached()
         } catch (e: Throwable) {
-            assertTrue { e is T }
+            assertIs<T>(e)
             finish(1)
         }
     }
```
dkhalanskyjb, line 117: You raise a good point about a race not being needed. I think though that we should also check what happens in case of a race between the `flow` block throwing an exception and the `timeout` directive. WDYT?
```suggestion
    }

    @Test
    fun testUpstreamExceptionsTakingPriority() = withVirtualTime {
        val flow = flow<Unit> {
            expect(2)
            withContext(NonCancellable) {
                delay(2.milliseconds)
            }
            assertFalse(currentCoroutineContext().isActive) // cancelled already
            expect(3)
            throw TestException()
        }.timeout(1.milliseconds)
        expect(1)
        assertFailsWith<TestException> {
            flow.collect {
                expectUnreached()
            }
        }
        finish(4)
    }
```

qwwdfsad: Makes sense, let's check it as well

kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
```diff
@@ -99,16 +99,19 @@ class TimeoutTest : TestBase() {
     @Test
     fun testUpstreamErrorTimeoutException() = testUpstreamError(TimeoutCancellationException(0, Job()))
 
+    @Test
+    fun testUpstreamErrorCancellationException() = testUpstreamError(CancellationException(""))
+
     private inline fun <reified T: Throwable> testUpstreamError(cause: T) = runTest {
```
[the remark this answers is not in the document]
qwwdfsad: In general, the less test infrastructure we use for tests that do not need it, the better.

On a side note, `withVirtualTime` is really ad-hoc, e.g.: "Regular tasks always dominate delayed ones. It means that *    `launch { while(true) yield() }` will block the progress of the delayed tasks"

dkhalanskyjb (approved):

PR #3486, opened 2022-10-17
Integration test step in Train 

mvicsokolova:
Set up for Integration Test build step.

Here is the [Train build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_BuildPlayground_MilkyWayMargaritaBobova_kotlinxtraincoroutines/224222229?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true) with Integration Test step and applied atomicfu JVM compiler plugin.


integration-testing/README.md
```diff
@@ -3,9 +3,9 @@
 This is a supplementary project that provides integration tests.
 
 The tests are the following:
-* `MavenPublicationValidator` depends on the published artifacts and tests artifacts binary content and absence of atomicfu in the classpath.
-* `CoreAgentTest` checks that `kotlinx-coroutines-core` can be run as a Java agent.
-* `DebugAgentTest` checks that the coroutine debugger can be run as a Java agent.
+* `mavenTest` depends on the published artifacts and tests artifacts binary content for absence of atomicfu in the classpath.
+* `coreAgentTest` checks that `kotlinx-coroutines-core` can be run as a Java agent.
+* `debugAgentTest` checks that the coroutine debugger can be run as a Java agent.
 * `smokeTest` builds the test project that depends on coroutines.
```
mvicsokolova, line 9: @qwwdfsad could you please drop a general comment here about why we need `jvmCoreTest` and `debugDynamicAgentTest`?

integration-testing/smokeTest/build.gradle
```diff
@@ -40,4 +41,12 @@ kotlin {
             }
         }
     }
+    targets {
+        configure([]) {
+            tasks.getByName(compilations.main.compileKotlinTaskName).kotlinOptions {
+                jvmTarget = "1.8"
```
qwwdfsad, line 47: Isn't it enabled by default?

integration-testing/build.gradle
```diff
@@ -130,3 +172,8 @@ compileTestKotlin {
 check {
     dependsOn([jvmCoreTest, debugDynamicAgentTest, mavenTest, debugAgentTest, coreAgentTest, 'smokeTest:build'])
 }
+compileKotlin {
+    kotlinOptions {
+        jvmTarget = "1.8"
```
qwwdfsad, line 177: Isn't it enabled by default?

integration-testing/build.gradle
```diff
@@ -2,11 +2,52 @@
  * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
+
+buildscript {
```
qwwdfsad, line 6: The script is already close to being 200-lines long.

Could you please migrate it to `.kts` while we are here, so it's easier to maintain?

integration-testing/build.gradle
```diff
@@ -2,11 +2,52 @@
...
+    }
+
+}
+
 plugins {
     id "org.jetbrains.kotlin.jvm" version "$kotlin_version"
 }
 
 repositories {
+    maven { url "https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev" }
```
qwwdfsad, line 50: It should be added only if `build_snapshot_train` is true, shouldn't it?

integration-testing/README.md
qwwdfsad: Done

integration-testing/smokeTest/build.gradle
mvicsokolova: @woainikk could you please eleborate why we need to set jvmTarget here?

integration-testing/build.gradle
woainikk: Yes, 1.8 is enabled by default, but jvmTarget = "1.8" is used everywhere in the project build scripts.  Also, maybe it's better to define the version explicitly?

integration-testing/build.gradle
mvicsokolova: This script contains some non trivial tasks configuration (e.g. [here](https://github.com/Kotlin/kotlinx.coroutines/blob/fbf966e22f9000bc3da262776e82e1221bd08136/integration-testing/build.gradle.kts#L68)) and it turned out to be very time consuming to migrate. It requires some special knowledge, and I failed do it quickly now ☹️. Maybe it's better to do it as a separate PR later?

qwwdfsad: Works for me

qwwdfsad (approved):

PR #3488, opened 2022-10-18
Fix atomicfu plugin application

mvicsokolova:
The new way of atomicfu compiler plugin applicaition takes 3 steps: 
- first the compiler plugin `AtomicfuKotlinGradleSubplugin` is [applied](https://github.com/Kotlin/kotlinx-atomicfu/blob/70dd7aaa2c51d5ca2556e981581830c5696fcbd3/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt#L119). It adds an extension `AtomicfuKotlinGradleExtension` with properties for turning on jvm and js transformation.
- Then these extension properties are [set](https://github.com/Kotlin/kotlinx-atomicfu/blob/70dd7aaa2c51d5ca2556e981581830c5696fcbd3/atomicfu-gradle-plugin/src/main/kotlin/kotlinx/atomicfu/plugin/gradle/AtomicFUGradlePlugin.kt#L121) in the AtomicfuGradlePlugin.
- When `AtomicfuKotlinGradleSubplugin$isApplicable` is invoked, the values of the extension properties are already set and the plugin is applied accordingly.

When we apply the atomicfu plugin too late in the configuration -- the `AtomicfuKotlinGradleSubplugin$isApplicable` may be called earlier, when the properties are not set yet and the compiler plugin is not applied.

[The Train build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_BuildPlayground_MilkyWayMargaritaBobova_kotlinxtraincoroutines/224222229?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true) with these changes + atomicfu JVM compiler plugin support

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,6 +76,9 @@ def configureKotlinJvmPlatform(configuration) {
     configuration.attributes.attribute(KotlinPlatformType.attribute, KotlinPlatformType.jvm)
 }
 
+// Configure subprojects with Kotlin sources
+apply plugin: "configure-compilation-conventions"
+
 allprojects {
     // the only place where HostManager could be instantiated
     project.ext.hostManager = new HostManager()
@@ -166,9 +169,6 @@ configure(subprojects.findAll { !sourceless.contains(it.name) && it.name != core
 }
 
 apply plugin: "bom-conventions"
-
-// Configure subprojects with Kotlin sources
-apply plugin: "configure-compilation-conventions"
 
 if (build_snapshot_train) {
     println "Hacking test tasks, removing stress and flaky tests"


Tapchicoma:
The problem is in [this line](https://github.com/Kotlin/kotlinx.coroutines/blob/02bf3569a6c0d59a2257eef47d66a038b0875f32/build.gradle#L139). 

It triggers `:kotlinx-core-coroutines` project evaluation including all `afterEvaluate { .. }` closures, where subplugin options are configured. The problem is that `apply plugin: "configure-compilation-conventions"` goes after this line meaning it configures almost nothing in `:kotlinx-core-coroutines` subproject as it is already evaluated. By moving "configure-compilaiton-conventions" above line #139, conventions are applied before core project evaluation fixing the issue.

Tapchicoma (approved):

build.gradle
```diff
@@ -76,6 +76,9 @@ def configureKotlinJvmPlatform(configuration) {
     configuration.attributes.attribute(KotlinPlatformType.attribute, KotlinPlatformType.jvm)
 }
 
+// Configure subprojects with Kotlin sources
+apply plugin: "configure-compilation-conventions"
```
qwwdfsad, line 80: I'm okay with this particular kludge.

What puzzles me is the general debuggability, i.e. consider the same situation happens to an external user:

>When we apply the atomicfu plugin too late in the configuration -- the AtomicfuKotlinGradleSubplugin$isApplicable may be called earlier, when the properties are not set yet and the compiler plugin is not applied.

What are the option for them to debug it and understand what's going on? Are there better options?

qwwdfsad (approved):

build.gradle
Tapchicoma: imho this case is nothing to do with Atomicfu plugin as it is not even present when core project evaluates. This is more Gradle edge-case. BTW why do you need `evaluationDependsOn` for core project? :thinking:

Issue #2387, opened 2020-11-13
Unexpected tryEmit behaviour


jmp0x0000:
`tryEmit` doesn't attempt to emit a value after first subscriber joined and returns `false`. 
Setting `replay` or `extraBufferCapacity` > 0 or replacing `tryEmit` by `emit` resolves the issue

```
    @Test
    fun tryEmitExample() = runBlocking {
        val sharedFlow = MutableSharedFlow<Long>()

        val asyncReceiver = async(){
            delay(300)
                sharedFlow.collect{
                    println("Received on 1 $it")
                }
            println("Done")
        }

        repeat(4) {
            delay(100)
            val res = sharedFlow.tryEmit(System.currentTimeMillis())
            println("Emitted ${System.currentTimeMillis()} Subscribers: ${sharedFlow.subscriptionCount.value} try: $res")
        }
        asyncReceiver.cancel()
    }
```
```
Emitted 1605303026489 Subscribers: 0 try: true
Emitted 1605303026592 Subscribers: 0 try: true
Emitted 1605303026693 Subscribers: 1 try: false
Emitted 1605303026799 Subscribers: 1 try: false
```

```
    @Test
    fun emitExample() = runBlocking {
        val sharedFlow = MutableSharedFlow<Long>()

        val asyncReceiver = async(){
            delay(300)
            sharedFlow.collect{
                println("Received on 1 $it")
            }
            println("Done")
        }

        repeat(4) {
            delay(100)
            sharedFlow.emit(System.currentTimeMillis())
            println("Emitted ${System.currentTimeMillis()} Subscribers: ${sharedFlow.subscriptionCount.value}")
        }
        asyncReceiver.cancel()
    }
```
```
Emitted 1605303080955 Subscribers: 0
Emitted 1605303081061 Subscribers: 0
Received on 1 1605303081162
Emitted 1605303081166 Subscribers: 1
Received on 1 1605303081267
Emitted 1605303081267 Subscribers: 1
```

Dominaezzz:
Have a look at #2346

jmp0x0000:
It's not a bug, it's a feature...  :)
This kind of implementation feels to me counterintuitive.

elizarov:
[...]

PR #3497, opened 2022-10-20
suggested update to SharedFlow.kt

Nodrex:
Coroutines and Flows are amazing I love them and many many thanks for them 🤩🤩🤩🤩 You guys in JetBrains rock 🤘🤘🤘 🚀🚀🚀
I have a small suggestion:
According to this big discussion: 
https://github.com/Kotlin/kotlinx.coroutines/issues/2387  and a lot of confusion, including myself, default value of `extraBufferCapacity` should be 1 instead of 0 and I think this parameter also needs better explanation


qwwdfsad:
Thanks!

Unfortunately, we cannot accept this contribution -- it is a breaking change for stable API that will break almost every client of `MutableSharedFlow` on update.
The best way to affect the current behaviour is to initiate a discussion and maybe propose a draft solution/idea, as well as outline the downsides of the current behaviour and why the proposed solution outweighs **benefits** of the current solution

Nodrex:
At least we can update docs, cause `onBufferOverflow` can be used to control backpressure on emitters (slowing them down). Is it not?

qwwdfsad:
Yes, it probably makes sense

Nodrex:
So should I send another pull request with just doc changes?

Issue #3143, opened 2022-01-17
Testing 1.6.0 - Correct way to collect and verify StateFlow<T> emissions


TreeFrogApps:
Hello,

After updating to version 1.6.0 I have test suites failing. I have read through the migration guide as best as I could understand, but still tests fail.  I have spent some time on this trying to pinpoint the issue to no avail.

The failing tests center around accumulating values emitted from a `StateFlow<T>`, which itself is prompted by events from a `MutableSharedFlow<T>` and then verifying the Item emission count and emission order/data.  I have not changed any of the mechanics of the tests, just updated using the migration guide to remove deprecation warnings (which will turn into errors on 1.7.0).

I have provided a simple contrived example of what fails and ultimately what passes. My actual tests are more complicated but the general mechanics are the same and anymore just creates unnecessary noise  :

Note : The example `TestViewModel` extends `androidx.lifecycle.ViewModel`

```
@OptIn(
    FlowPreview::class,
    ExperimentalCoroutinesApi::class
)
class MVPTest {

    class TestViewModel : ViewModel() {

        private val events = MutableSharedFlow<Int>(0)

        val eventsFlow: StateFlow<Int> by lazy {
            events.flatMapConcat { int ->
                flow { emit(int) }
                    .onStart { emit(int - 1) }
                    .onCompletion { emit(int + 1) }
            }.stateIn(viewModelScope, SharingStarted.Lazily, 0)
        }

        fun pushEvent(id: Int) = viewModelScope.launch {
            events.emit(id)
        }
    }

    @Before
    fun setup() {
        Dispatchers.setMain(UnconfinedTestDispatcher())
    }

    @After
    fun teardown() {
        Dispatchers.resetMain()
    }

    @Test fun failing() = runTest(UnconfinedTestDispatcher()) {
        // Setup
        val results = mutableListOf<Int>()
        val viewModel = TestViewModel()
        val job = viewModel.eventsFlow
            .onEach(results::add)
            .launchIn(this)

        // Test
        viewModel.pushEvent(1)
        viewModel.pushEvent(4)

        //Verify
        assertEquals(listOf(0, 1, 2, 3, 4, 5), results) // fails results = [0, 2, 5]
        job.cancel()
    }

    @Test fun failing2() = runTest {
        // Setup
        val results = mutableListOf<Int>()
        val viewModel = TestViewModel()
        val job = viewModel.eventsFlow
            .onEach(results::add)
            .launchIn(this)

[...]

PR #3500, opened 2022-10-24
Deprecate SharedFlow.toList(dst) and toSet(dst)

dkhalanskyjb:
See https://github.com/Kotlin/kotlinx.coroutines/issues/3143#issuecomment-1286004389


qwwdfsad (approved):
No questions about this particular change standalone.

Though the referred comment is using `take(1).toList(result)`, so the problem is probably in something else

dkhalanskyjb:
Yeah, this change would not help with that comment, it's just an example of code that would correctly get red.

PR #3501, opened 2022-10-25
Update atomicfu to 0.18.5

mvicsokolova:
Apply JVM IR compiler plugin

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ kotlin_version=1.7.20
 # Dependencies
 junit_version=4.12
 junit5_version=5.7.0
-atomicfu_version=0.18.4
+atomicfu_version=0.18.5
 knit_version=0.4.0
 html_version=0.7.2
 lincheck_version=2.14.1
@@ -58,4 +58,5 @@ org.gradle.jvmargs=-Xmx3g
 
 kotlin.mpp.enableCompatibilityMetadataVariant=true
 kotlin.mpp.stability.nowarn=true
-kotlinx.atomicfu.enableIrTransformation=true
+kotlinx.atomicfu.enableJvmIrTransformation=true
+kotlinx.atomicfu.enableJsIrTransformation=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@
 # Kotlin
 version=1.6.4-SNAPSHOT
 group=org.jetbrains.kotlinx
-kotlin_version=1.7.20
+kotlin_version=1.7.21
 
 # Dependencies
 junit_version=4.12


mvicsokolova:
A very dumb problem ocurred:

`:kotlinx-coroutines-core:jvmApiCheck` in this build [failed](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildAggregated/3982790?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true) because atomicfu JVM IR compiler plugin generated some declarations with `PUBLIC` visibility modifier, and those appeared in the public api, that is obviously not right.

For some reason this simple bug was not revealed by the test [build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_BuildPlayground_MilkyWayMargaritaBobova_kotlinxtraincoroutines/224222229?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true)  of coroutines with atomicfu JVM IR plugin applied. The config of the build is equal to the train config, actually, but seems that `:kotlinx-coroutines-core:jvmApiCheck` was not called there at all, I could not find it in the build log.

I filed this issue in `kotlinx-atomicfu`: https://github.com/Kotlin/kotlinx-atomicfu/issues/258
And prepared a fix for this problem for Kotlin `1.7.21`, `1.8.0` and `master`: https://jetbrains.team/p/kt/reviews/7497/timeline

mvicsokolova:
That's too bad, and really reveals the problem of the lack of atomicfu tests and our test train builds ☹️. That code in the atomicfu  JVM IR compiler plugin is very old, but we never ran into this issue with visibility modifiers.

I probably need to add some tests to atomicfu-gradle-plugin that would also dump the public api and check that no generated declarations leak

qwwdfsad:
I see that `apiCheck` is explicitly disabled in these `train` branches which probably should not be the case

mvicsokolova:
Yes, `apiCheck` and `knitCheck` tasks were always excluded in the train builds 😞. I've asked to turn them on.

And I turned on the `apiCheck` in that test build, now it [fails](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_BuildPlayground_MilkyWayMargaritaBobova_kotlinxtraincoroutines/227455465?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=227455465_40930_5606&logView=flowAware) at `jvmApiCheck` as it should have :(

qwwdfsad (approved):

qwwdfsad:
@mvicsokolova is it safe to merge or we do wait for something else?

mvicsokolova:
Yes I'll merge it. I just forgot to press the button.

PR #3512, opened 2022-10-31
Fix suspend-resume race in coroutine debugger

qwwdfsad:


qwwdfsad:
I will overwrite the change with force-push, here it is for the history:

https://github.com/Kotlin/kotlinx.coroutines/blob/0c653aee75fc3a72d29424fcb34a3324abc144bd/kotlinx-coroutines-core/common/src/Yield.kt#L51

This patch is the best way to **reliably** reproduce resume-after-suspend races

PR #3513, opened 2022-10-31
Remove some broken tests

dkhalanskyjb:
These tests haven't been reliable for a very long time for some reason. We could ensure that they pass with a manual check, but it feels like nobody actually relies on this behavior, and it's not clear who even would want to.

https://github.com/search?p=3&q=%22dispatchTimeoutMs+%3D+0%22&type=Code https://grep.app/search?q=dispatchTimeoutMs%20%3D%200

### DIFF
--- a/kotlinx-coroutines-test/common/test/RunTestTest.kt
+++ b/kotlinx-coroutines-test/common/test/RunTestTest.kt
@@ -69,20 +69,6 @@ class RunTestTest {
         deferred.await()
     }
 
-    /** Tests that a dispatch timeout of `0` will fail the test if there are some dispatches outside the scheduler. */
-    @Test
-    fun testRunTestWithZeroTimeoutWithUncontrolledDispatches() = testResultMap({ fn ->
-        assertFailsWith<UncompletedCoroutinesError> { fn() }
-    }) {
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-test/common/test/RunTestTest.kt
```diff
@@ -69,20 +69,6 @@ class RunTestTest {
...
-    @Test
-    fun testRunTestWithZeroTimeoutWithUncontrolledDispatches() = testResultMap({ fn ->
-        assertFailsWith<UncompletedCoroutinesError> { fn() }
-    }) {
-        runTest(dispatchTimeoutMs = 0) {
-            withContext(Dispatchers.Default) {
-                delay(10)
-                3
-            }
-            fail("shouldn't be reached")
```
qwwdfsad, line 82: For the future reference -- we have `expectUnreached`

qwwdfsad (approved):

kotlinx-coroutines-test/common/test/RunTestTest.kt
dkhalanskyjb: This class is not inherited from `TestBase`.

PR #3514, opened 2022-10-31
Fix suspend-resume race in DebugProbes

### LINKED ISSUE #3193: Coroutines debugger may overwrite the actual state of coroutine during asynchronous resume
Consider the following situation when the dispatcher in the context is multithreaded

```
suspend fun foo() = yield()
```

It basically can be boiled down to the following pseudocode:
```
fun foo(...) {
    uCont.intercepted().dispatchUsingDispatcher()
    // Notify the debugger the coroutine is suspended 
    probeCoroutineSuspended()
    return COROUTINE_SUSPENDED // Unroll the stack
}
```

where `dispatch` **concurrently** invokes `probeCoroutineResumed` to notify the debugger the coroutine is actually executing right now.

The problem is straightforward -- if timings are unlucky enough, `probeCoroutineResumed` can be invoked *before* `probeCoroutineSuspended`, with `probeCoroutineSuspended` overwriting the state and leaving it as `SUSPENDED`

qwwdfsad:
### Reproducing scenario

For the following code:
```
suspend fun foo() = yield()
```

the following bytecode is produced:

```
fun foo(...) {
    uCont.intercepted().dispatchUsingDispatcher() // 1
    probeCoroutineSuspended() // 2
    return COROUTINE_SUSPENDED // Unroll the stack
}
```

And it is possible to observe the next 'probeCoroutineSuspended' **prior** to receiving 'probeCoroutineSuspended'.

### Steps taken

To address this, a dedicated 'unmatchedResumes' field is introduced to the coroutine state, which keeps track of consecutive 'probeCoroutineResumed' without matching 'probeCoroutineSuspended' and attributes lately arrived probes to it.

Unfortunately, it introduces a much more unlikely race when **two** 'probeCoroutineSuspended' are reordered, then we misattribute 'lastObservedFrame', but it is still better than misattributing the actual coroutine state.

@Volatile and @Synchronized are also introduced to DebugCoroutineInfoImpl as previously they have been subject to data races as well

Fixes #3193

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/DebugCoroutineInfoImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/DebugCoroutineInfoImpl.kt
@@ -41,8 +41,85 @@ internal class DebugCoroutineInfoImpl(
      * Can be CREATED, RUNNING, SUSPENDED.
      */
     public val state: String get() = _state
+
     @Volatile
     private var _state: String = CREATED
+
+    /*
+     * How many consecutive unmatched 'updateState(RESUMED)' this object has received.
+     * It can be `> 1` in two cases:
+     *
+     * * The coroutine is finishing and its state is being unrolled in BaseContinuationImpl, see comment to DebugProbesImpl#callerInfoCache
+     *   Such resumes are not expected to be matched and are ignored.
+     * * We encountered suspend-resume race explained above, and we do wait for a match.
+     */
+    private var unmatchedResume = 0
+
+    /**
+     * Here we orchestrate overlapping state updates that are coming asynchronously.
+     * In a nutshell, `probeCoroutineSuspended` can arrive **later** than its matching `probeCoroutineResumed`,
+     * e.g. for the following code:
+     * ```
+     * suspend fun foo() = yield()
+     * ```
+     *
+     * we have this sequence:
+     * ```
+     * fun foo(...) {
+     *     uCont.intercepted().dispatchUsingDispatcher() // 1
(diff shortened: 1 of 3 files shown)


qwwdfsad:
I was literally at my wit's end while debugging it and figuring out whether my potential fixes are reasonable, so it's possible that I've missed something obvious

kotlinx-coroutines-core/jvm/src/debug/internal/DebugCoroutineInfoImpl.kt
```diff
@@ -40,19 +40,101 @@ internal class DebugCoroutineInfoImpl(
...
+     * suspend fun foo() = yield()
+     * ```
+     *
+     * we have this sequence:
+     * ```
+     * fun foo(...) {
+     *     uCont.intercepted().dispatchUsingDispatcher() // 1
+     *     // Notify the debugger the coroutine is suspended
+     *     probeCoroutineSuspended() // 2
+     *     return COROUTINE_SUSPENDED // Unroll the stack
```
dkhalanskyjb, line 71: Is it correct that (though astronomically unlikely?) `probeCoroutineSuspended` can even come out of order, and the fix doesn't prevent that? Let's say that, while the first `probeCoroutineSuspended` was awaiting execution, the coroutine was suspended again, and the new `probeCoroutineSuspended` managed to happen before the first one did. So, could we discard the wrong `probeCoroutineSuspend` here?

dkhalanskyjb (approved):
If you missed something obvious, then both of us did.

PR #3515, opened 2022-10-31
Make withTimeout throw not a CancellationException

### LINKED ISSUE #1374: Replace TimeoutCancellationException with TimeoutException
In the current version of the library, `withTimeout` throws `TimeoutCancellationException` when a timeout was exceeded.

It can lead to very subtle errors, for example:
```
launch {
    val result = withTimeout(...) {
        // ... some computation ...
    }
    // process result
}
```

`TimeoutCancellationException` is `CancellationException`, thus is never reported.
But in the snippet above, it's likely to be a programmatic error. If it is expected to miss the deadline, then `withTimeoutOrNull` should be used **explicitly**.

dkhalanskyjb:
Fixes #1374

### DIFF
--- a/integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
+++ b/integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
@@ -749,7 +749,7 @@ class ListenableFutureTest : TestBase() {
             }
         }
         future.set(1)
-        withTimeout(60_000) {
+        kotlinx.coroutines.time.withTimeout(60_000) {
             children.forEach { it.join() }
             assertEquals(count, completed.get())
         }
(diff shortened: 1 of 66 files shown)


dkhalanskyjb:
Report of how this went:
* The IDE managed to correctly replace the deprecated `withTimeout` with the new overload.
* When replacing `TimeoutCancellationException` with `TimeoutException`, the IDE did not suggest replacing all the instances. Tought to say whether this is good. In our case, it wouldn't be. Also, there's the possible pattern `catch (e: CancellationException) { if (e is TimeoutCancellationException) X() else Y() }`, which would break if `TimeoutCancellationException` is blindly replaced everywhere.

The manual intervention by me was minimal (see the individual commits), but it seems that only a couple of tests have failed, and those are the stacktrace recovery tests. Not immediately obvious to me what's going on there.

elizarov:
How about trying to find a different name for a new `withTimeout` implementation?

qwwdfsad:
If we decide to follow a road with a new name, I suggest making it consistent with a potential replacement of `withContext`*.

* -- it allows to break of structured concurrency in a bit unexpected ways (e.g. we probably should've prohibited `withContext(Job())` and its behaviour with `CopyableThreadContextElement` may (not *should*) be reconsidered

dkhalanskyjb:
The code changed somewhat, as has the auto-replacement behavior. I feel it's going to be easier to redo this from scratch relying on the knowledge gained from this PR than it would be to change the PR itself.

PR #3516, opened 2022-10-31
Disable tests with real source of the time on Java's Windows to preve…

qwwdfsad:
…nt flakiness

We still want to test real code, not virtualized mocked clocks, but Windows itself is virtualized on our TeamCity, making Java's parker resolution really low and unstable for such tests


qwwdfsad:
Let's see if it's `os.name` or `teamcity.agent.jvm.os.name` though

kotlinx-coroutines-core/jvm/test/TestBase.kt
```diff
@@ -254,3 +254,10 @@ public actual open class TestBase(private var disableOutCheck: Boolean)  {
 
     protected suspend fun currentDispatcher() = coroutineContext[ContinuationInterceptor]!!
 }
+
+/*
+ * We ignore tests that test **real** non-virtualized tests with time on Windows, because
+ * our TC Windows is virtualized itself (oh, the irony) and its clock resolution is dozens of ms,
```
dkhalanskyjb, line 260:
```suggestion
 * our CI Windows is virtualized itself (oh, the irony) and its clock resolution is dozens of ms,
```
Should be more clear to external contributors.

dkhalanskyjb (commented):
Do you happen to know what difference Java makes here? Why don't we observe the same on Native?

qwwdfsad:
Java uses `WaitForSingleObject` that on its own happens to have resolution problems (see https://github.com/Kotlin/kotlinx.coroutines/pull/3301#issue-1245498016, dated back to 1998!), it seems like in virtualized environments it just gets exponentially worse.

No idea why it is not the issue on Windows though, probably different OS-level API is used

dkhalanskyjb (approved):

PR #3517, opened 2022-11-01
Stabilize isClosedForSend and isClosedForReceive

qwwdfsad:
Both seem to have their proper niche (whether it is a fast-path optimization for single-threaded code or some sanity checks), both have a well-defined contract and are consistent.

The only open question is their potential linearizability issues (that are partially addressed in #3103), but it won't make us change this API in any way

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -26,7 +26,7 @@ public interface SendChannel<in E> {
      * Returns `true` if this channel was closed by an invocation of [close] or its receiving side was [cancelled][ReceiveChannel.cancel].
      * This means that calling [send] will result in an exception.
      *
-     * Note that if this property returns `false`, it does not guarantee that consecutive call [send] will succeed, as the
+     * Note that if this property returns `false`, it does not guarantee that consecutive call to [send] will succeed, as the
      * channel can be concurrently closed right after the check. For such scenarios, it is recommended to use [trySend] instead.
      *
      * @see SendChannel.close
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -23,12 +23,15 @@ import kotlin.jvm.*
...
     /**
-     * Returns `true` if this channel was closed by an invocation of [close]. This means that
-     * calling [send] will result in an exception.
+     * Returns `true` if this channel was closed by an invocation of [close] or its receiving side was [cancelled][ReceiveChannel.cancel].
+     * This means that calling [send] will result in an exception.
      *
-     * **Note: This is an experimental api.** This property may change its semantics and/or name in the future.
+     * Note that if this property returns `false`, it does not guarantee that consecutive call to [send] will succeed, as the
+     * channel can be concurrently closed right after the check. For such scenarios, it is recommended to use [trySend] instead.
+     *
```
dkhalanskyjb, line 31:
```suggestion
     *
     * @see SendChannel.trySend
```
to highlight what is actually meant in the majority of cases?

Also, WDYT about marking this API as delicate?

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -174,14 +177,18 @@ public interface SendChannel<in E> {
...
+     * side and all previously sent items were already received, or if the receiving side was [cancelled][ReceiveChannel.cancel].
+     *
+     * This means that calling [receive] will result in a [ClosedReceiveChannelException] or a corresponding cancellation cause.
+     * If the channel was closed because of an exception, it is considered closed, too, but is called a _failed_ channel.
+     * All suspending attempts to receive an element from a failed channel throw the original [close][SendChannel.close] cause exception.
      *
-     * **Note: This is an experimental api.** This property may change its semantics and/or name in the future.
+     * Note that if this property returns `false`, it does not guarantee that consecutive call to [receive] will succeed, as the
+     * channel can be concurrently closed right after the check. For such scenarios, it is recommended to use [receiveCatching] instead.
+     *
```
dkhalanskyjb, line 188: Ditto.

kotlinx-coroutines-core/common/src/channels/Channel.kt
dkhalanskyjb: > Also, WDYT about marking this API as delicate?

The question probably got lost.

qwwdfsad: Not very keen on the idea, to be honest. I would prefer having `delicate` marker on a widespread or potentially really dangerous API, so it keeps having its weight.
I would like to avoid using it as "please just read a documentation" marker

qwwdfsad: Sold!

dkhalanskyjb (approved):

PR #3534, opened 2022-11-21
Improve DebugProbes performance

### LINKED ISSUE #3527: Performance overhead of DebugProbesImpl for non-concurrent coroutines
Snippet to reproduce:

```
class DebugProbesTest : DebugTestBase() {
    fun <Node> generateRecursiveSequence(initialSequence: Sequence<Node>, children: (Node) -> Sequence<Node>): Sequence<Node> {
        return sequence {
            val initialIterator = initialSequence.iterator()
            if (!initialIterator.hasNext()) {
                return@sequence
            }
            val visited = HashSet<Node>()
            val sequences = ArrayDeque<Sequence<Node>>()
            sequences.addLast(initialIterator.asSequence())
            while (sequences.isNotEmpty()) {
                val currentSequence = sequences.removeFirst()
                for (node in currentSequence) {
                    if (visited.add(node)) {
                        yield(node)
                        sequences.addLast(children(node))
                    }
                }
            }
        }
    }

    @Volatile
    var a = 2
    @Test
    fun stressGenerateRecursive() {
        while (true) {
            runBlocking {
                repeat(8) {
                    launch(Dispatchers.Default) {
                        val seq = generateRecursiveSequence((1..100).asSequence()) {
                            (1..it).asSequence()
                        }

                        for (i in seq) {
                            a = i
                        }
                    }
                }
            }
        }
    }
}
```

Profile:
deb

Up to 30% is occupied by `updateState`

qwwdfsad:
* Get rid of RW lock that was contended enough on reads to be observable by non-concurrent coroutines-heavy code (up to 30% of throughput on IDEA-specific benchmark)
* Tweak the code to be DRF in the absence of RW lock
* Document snapshots' weak consistency guarantee 

Fixes #3527

### DIFF
--- a/kotlinx-coroutines-debug/src/DebugProbes.kt
+++ b/kotlinx-coroutines-debug/src/DebugProbes.kt
@@ -20,13 +20,20 @@ import kotlin.coroutines.*
  * asynchronous stack-traces and coroutine dumps (similar to [ThreadMXBean.dumpAllThreads] and `jstack` via [DebugProbes.dumpCoroutines].
  * All introspecting methods throw [IllegalStateException] if debug probes were not installed.
  *
- * Installed hooks:
+ * ### Consistency guarantees
  *
+ * All snapshotting operations (e.g. [dumpCoroutines]) are *weakly-consistent*, meaning that they happen
+ * concurrently with coroutines progressing their own state. These operations are guaranteed to observe
+ * each coroutine's state exactly once, but the state is not guaranteed to be the most recent before the operation.
+ * In practice, it means that for snapshotting operation being in progress, for each concurrent coroutine either
+ * the state prior to the operation or the state that was reached during the current operation is observed.
+ *
+ * ### Installed hooks
  * * `probeCoroutineResumed` is invoked on every [Continuation.resume].
  * * `probeCoroutineSuspended` is invoked on every continuation suspension.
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-debug/src/DebugProbes.kt
```diff
@@ -20,13 +20,20 @@ import kotlin.coroutines.*
  * asynchronous stack-traces and coroutine dumps (similar to [ThreadMXBean.dumpAllThreads] and `jstack` via [DebugProbes.dumpCoroutines].
  * All introspecting methods throw [IllegalStateException] if debug probes were not installed.
  *
- * Installed hooks:
+ * ### Consistency guarantees
  *
+ * All snapshotting operations (e.g. [dumpCoroutines]) are *weakly-consistent*, meaning that they happen
+ * concurrently with coroutines progressing their own state. These operations are guaranteed to observe
+ * each coroutine's state exactly once, but the state is not guaranteed to be the most recent before the operation.
+ * In practice, it means that for snapshotting operation being in progress, for each concurrent coroutine either
```
dkhalanskyjb, line 28:
```suggestion
 * In practice, it means that for snapshotting operations in progress, for each concurrent coroutine either
```

dkhalanskyjb (approved):

PR #3537, opened 2022-11-22
Introduce a separate slot for stealing tasks into in CoroutineScheduler

### LINKED ISSUE #3416: CoroutineScheduler: do not re-add stolen task to the head of the WorkerQueue where the task can be stolen from again
We should hand it directly to the stealing worker without intermediate stores.

Apart from code being more straightforward, it also saves us at least 2 CASes on a fairly contended memory location

qwwdfsad:
It solves two problems:

* Stealing into exclusively owned local queue does no longer require and CAS'es or atomic operations where they were previously not needed. It should save a few cycles on the stealing code path
* The overall timing perturbations should be slightly better now: previously it was possible for the stolen task to be immediately got stolen again from the stealer thread because it was actually published to owner's queue, but its submission time was never updated

Fixes #3416

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -726,7 +726,6 @@ internal class CoroutineScheduler(
                 parkedWorkersStackPush(this)
                 return
             }
-            assert { localQueue.size == 0 }
             workerCtl.value = PARKED // Update value once
             /*
              * inStack() prevents spurious wakeups, while workerCtl.value == PARKED
@@ -873,15 +872,10 @@ internal class CoroutineScheduler(
             }
         }
 
-        fun findTask(scanLocalQueue: Boolean): Task? {
-            if (tryAcquireCpuPermit()) return findAnyTask(scanLocalQueue)
+        fun findTask(mayHaveLocalTasks: Boolean): Task? {
+            if (tryAcquireCpuPermit()) return findAnyTask(mayHaveLocalTasks)
             // If we can't acquire a CPU permit -- attempt to find blocking task
-            val task = if (scanLocalQueue) {
-                localQueue.poll() ?: globalBlockingQueue.removeFirstOrNull()
-            } else {
-                globalBlockingQueue.removeFirstOrNull()
-            }
-            return task ?: trySteal(blockingOnly = true)
+            return globalBlockingQueue.removeFirstOrNull() ?: trySteal(blockingOnly = true)
         }
 
         private fun findAnyTask(scanLocalQueue: Boolean): Task? {
@@ -911,7 +905,6 @@ internal class CoroutineScheduler(
         }
 
         private fun trySteal(blockingOnly: Boolean): Task? {
-            assert { localQueue.size == 0 }
             val created = createdWorkers
             // 0 to await an initialization and 1 to avoid excess stealing on single-core machines
             if (created < 2) {
@@ -925,7 +918,6 @@ internal class CoroutineScheduler(
                 if (currentIndex > created) currentIndex = 1
                 val worker = workers[currentIndex]
                 if (worker !== null && worker !== this) {
-                    assert { localQueue.size == 0 }
                     val stealResult = if (blockingOnly) {
                         worker.localQueue.tryStealBlocking(stolenTask)
                     } else {
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -10,6 +10,7 @@ import kotlinx.coroutines.internal.*
 import java.io.*
 import java.util.concurrent.*
 import java.util.concurrent.locks.*
+import kotlin.jvm.internal.Ref.ObjectRef
 import kotlin.math.*
 import kotlin.random.*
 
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -873,15 +872,10 @@ internal class CoroutineScheduler(
...
+        fun findTask(mayHaveLocalTasks: Boolean): Task? {
+            if (tryAcquireCpuPermit()) return findAnyTask(mayHaveLocalTasks)
             // If we can't acquire a CPU permit -- attempt to find blocking task
-            val task = if (scanLocalQueue) {
-                localQueue.poll() ?: globalBlockingQueue.removeFirstOrNull()
-            } else {
-                globalBlockingQueue.removeFirstOrNull()
-            }
-            return task ?: trySteal(blockingOnly = true)
+            return globalBlockingQueue.removeFirstOrNull() ?: trySteal(blockingOnly = true)
```
qwwdfsad, line 878: This fix seems to be incorrect, I will ament the last commit

qwwdfsad: For each blocking task scheduled, the scheduler unparks/creates a separate worker to steal that the current thread blocks indefinitely (i.e. "no progress from here now on in this thread"), so no liveness issues or bugs, but it still worth trying to poll thread's local queue for blocking tasks only

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -100,41 +101,37 @@ internal class WorkQueue {
...
-     * Tries stealing from [victim] queue into this queue.
+     * Tries stealing from this queue into the [stolenTaskRef] argument.
      *
      * Returns [NOTHING_TO_STEAL] if queue has nothing to steal, [TASK_STOLEN] if at least task was stolen
      * or positive value of how many nanoseconds should pass until the head of this queue will be available to steal.
      */
-    fun tryStealFrom(victim: WorkQueue): Long {
-        assert { bufferSize == 0 }
-        val task  = victim.pollBuffer()
+    fun trySteal(stolenTaskRef: ObjectRef<Task?>): Long {
```
dkhalanskyjb, line 111: The names no longer reflect who's the receiver and who's doing the action. The semantics of "I steal (from myself) (into this bag that was given to me)" is not really intuitive!
* `tryGift`? Semantically clear, but "stealing" is a well-established term.
* `tryStealFromThisInto`? Wordy but much less confusing.
* Make it a separate static function, like `trySteal(victim, stolenTaskRef)`.
* ...

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -100,41 +103,63 @@ internal class WorkQueue {
...
+
+        while (start != end) {
+            --end
+            val index = end and MASK
+            if (blockingTasksInBuffer.value == 0) break
+            val value = buffer[index]
+            if (value != null && value.isBlocking && buffer.compareAndSet(index, value, null)) {
+                blockingTasksInBuffer.decrementAndGet()
+                return value
+            }
```
dkhalanskyjb, line 160: How about moving this into a separate method, like `tryExtractBlockingTask(index: Int)`?

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
qwwdfsad: Maybe `tryStealInto()`? `victim.tryStealInto(stolenTaskSlot)` seems quite straightforward

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
qwwdfsad: Done

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
dkhalanskyjb: "I steal (into this bag)" begs the question, from whom do you steal? The answer is, from yourself, which is just not obvious. When I see `X.y()`, I always read this as "`X`, I order you to `y`".

qwwdfsad: I agree with the general reasoning. That's why I added context (receiver name), as this is the method used in a very specific manner on call sites we control, including the receiver's name.

I'm reluctant to both adding a companion (so the static method has FQN name like `WorkQueue.trySteal(victim, into)`) and to giving up on `steal` terminology

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -100,41 +103,70 @@ internal class WorkQueue {
...
             val value = buffer[index]
             if (value != null && value.isBlocking && buffer.compareAndSet(index, value, null)) {
-                victim.blockingTasksInBuffer.decrementAndGet()
-                add(value)
+                blockingTasksInBuffer.decrementAndGet()
+                stolenTaskRef.element = value
                 return TASK_STOLEN
             } else {
                 ++start
             }
```
dkhalanskyjb, line 134:
```kotlin
val task = tryExtractBlockingTask(start)
if (task != null) {
  stolenTaskRef.element = task
  return TASK_STOLEN
} else {
  ++start
}
```
or even
```kotlin
stolenTaskRef.element = tryExtractBlockingTask(start++) ?: continue
return TASK_STOLEN
```
?

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
dkhalanskyjb: Eh , ok. I think this warrants a comment though, in the vein of
"Counterintuitively, the queue from which a task is stolen is [this], so we are stealing from ourselves here. The reason for the name is, *stealing* is an established concept in concurrency".

PR #3538, opened 2022-11-23
typo fix

surajsahani:
returned typo fix


qwwdfsad:
It's not a typo, it's intended format of the keyword `return`.

PR #3539, opened 2022-11-24
Mention value-based conflation in MutableStateFlow

### LINKED ISSUE #3379: Docs for StateFlow.emit() do not mention that an equal value will not result in an emit
I noticed a discrepancy between the docs of `StateFlow`'s `emit()` function and its `value` property:

The docs of StateFlow's `value` property say that
> Setting a value that is equal to the previous one does nothing.

The docs of `emit()` are inherited from `Flow.emit()`, and therefore do not mention anything about equal values. However, the implementation of `StateFlow.emit()` looks like this:
```kotlin
    override suspend fun emit(value: T) {
        this.value = value
    }
```

So, clearly, the restriction that equal values will not lead to collectors receiving the update applies to `emit()` just as it does to `value`.

I think the docs here are quite misleading however, as they suggest that `emit()` will send an update to collectors even if the value is the same. It lead me to try for quite a while to use `emit()` to fix a problem with a flow that didn't emit as I expected. The reason was something completely different, but this discrepancy confused me for quite some time and I would like to spare others from making the same mistake.

Therefore I suggest to add the hint from the docs of `value` to the docs of `emit()` as well.

qwwdfsad:
Fixes #3379

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/StateFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/StateFlow.kt
@@ -145,8 +145,9 @@ public interface StateFlow<out T> : SharedFlow<T> {
  * A mutable [StateFlow] that provides a setter for [value].
  * An instance of `MutableStateFlow` with the given initial `value` can be created using
  * `MutableStateFlow(value)` constructor function.
- *
+
  * See the [StateFlow] documentation for details on state flows.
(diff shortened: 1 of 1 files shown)


qwwdfsad:
The mention itself is harmless, and is much easier than overriding and re-documenting `emit` and `tryEmit` members

dkhalanskyjb (approved):
This seems to be enough, given that conflation is mentioned in the `StateFlow` documentation as well. Maybe I'd also reference that fact. `, like for any [StateFlow]`.

PR #3562, opened 2022-12-28
Eliminate unneeded `LimitedDispatcher` instances on `Dispatchers.Default` and `Dispatchers.IO`

dovchinnikov:
Before the change `LimitedDispatcher(parallelism)` was returned. While it does work as expected, any submitted task 
goes through its queue and `parallelism` number of workers. The change allows to eliminate the `LimitedDispatcher` 
instance and its queue in case the requested parallelism is greater or equal to the effective parallelism of `Dispatchers.Default` and/or `Dispatchers.IO`

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
@@ -50,7 +50,7 @@ private object UnlimitedIoScheduler : CoroutineDispatcher() {
     @ExperimentalCoroutinesApi
     override fun limitedParallelism(parallelism: Int): CoroutineDispatcher {
         parallelism.checkParallelism()
-        if (parallelism == Int.MAX_VALUE) return this
+        if (parallelism >= MAX_POOL_SIZE) return this
         return super.limitedParallelism(parallelism)
     }
 }
--- a/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
@@ -38,6 +38,13 @@ private object UnlimitedIoScheduler : CoroutineDispatcher() {
     override fun dispatch(context: CoroutineContext, block: Runnable) {
         DefaultScheduler.dispatchWithContext(block, BlockingContext, false)
     }
+
+    @ExperimentalCoroutinesApi
+    override fun limitedParallelism(parallelism: Int): CoroutineDispatcher {
+        parallelism.checkParallelism()
+        if (parallelism == Int.MAX_VALUE) return this
+        return super.limitedParallelism(parallelism)
+    }
 }
 
 // Dispatchers.IO
--- a/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
@@ -14,6 +14,14 @@ internal object DefaultScheduler : SchedulerCoroutineDispatcher(
     CORE_POOL_SIZE, MAX_POOL_SIZE,
     IDLE_WORKER_KEEP_ALIVE_NS, DEFAULT_SCHEDULER_NAME
 ) {
+
+    @ExperimentalCoroutinesApi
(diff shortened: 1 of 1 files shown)


qwwdfsad:
No conceptual objections from me. @dkhalanskyjb mind taking a look at the general approach?

kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
```diff
@@ -38,6 +46,13 @@ private object UnlimitedIoScheduler : CoroutineDispatcher() {
     override fun dispatch(context: CoroutineContext, block: Runnable) {
         DefaultScheduler.dispatchWithContext(block, BlockingContext, false)
     }
+
+    @ExperimentalCoroutinesApi
+    override fun limitedParallelism(parallelism: Int): CoroutineDispatcher {
+        parallelism.checkParallelism()
+        if (parallelism == Int.MAX_VALUE) return this
```
dkhalanskyjb, line 53:
```suggestion
        if (parallelism >= MAX_POOL_SIZE) return this
```
Symmetrical and also a bit more general, in case someone overrides `MAX_POOL_SIZE` to a lower value.

dovchinnikov:
Another approach was to expose `open val effectiveParallelism: Int get() = Int.MAX_VALUE` to `CoroutineDispatcher`, move the check `if (parallelism >= effectiveParallelism) return this`, then replace `limitedParallelism` implementations with `override val effectiveParallelism`. I've decided not to go that way since it requires exposing additional member (`@InternalCoroutinesApi` but still), and it can still be done later.

kotlinx-coroutines-core/jvm/src/scheduling/Dispatcher.kt
dovchinnikov: I can rebase the commit to include MAX_POOL_SIZE into original commit description instead of Int.MAX_VALUE if you wish

dkhalanskyjb: We squash PRs into one commit and manually go through commit messages before merging, so it's not a big deal.

dkhalanskyjb (approved):
`effectiveParallelism` seems to me like a leaky abstraction because `Dispatchers.IO` has an effective parallelism of 64, but `Dispatchers.IO.limitedParallelism(n) == UnlimitedIoDispatcher.limitedParallelism(n)` has an effective parallelism of two million threads. So, I like what you propose now more.

dovchinnikov:
I'd argue that `Dispatchers.IO.limitedParallelism()` providing views of unbounded scheduler is leaky :) 
I still think the unbounded scheduler should be exposed somehow differently, e.g. via a top-level function `fun giveMeNewIoDispatcherView(parallelism: Int)`; then `Dispatchers.IO.limitedParallelism()` would cut from 64 threads as expected, and `Dispatchers.IO.effectiveParallelism` would return 64 as expected. Consistency FTW

dkhalanskyjb:
Practicality outweighed consistency in this case. I'm sure *nobody* will be sad if they try to limit `Dispatchers.IO` to 100 threads and, in the worst case, observe those 100 threads and not 64. However, people *will* be sad if they observe starvation, which may well happen. The following is an excerpt from our design document:

A naive and problematic attempt at migrating some backend code to `limitedParallelism` from dedicated thread pools:

```kotlin
val dbConnections = Dispatchers.IO.limitedParallelism(100)
val loggers = Dispatchers.IO.limitedParallelism(2)
val cachers = Dispatchers.IO.limitedParallelism(5)
```

The number of max connections for PostgreSQL and MySQL is about 100, and the need to further increase this limit is common.
The number of threads in Dispatchers.IO is **64**.

* The code will run just fine until Christmas Eve when it will become evident that the number of connections is half of what was possible before the migration. Given how rare the condition is, no tutorials will cover this.
* With fairness = 16, even assuming an equal length of all tasks, about 4 threads on average will be dedicated to fair execution; the rest may be busy serving the DB connections. This is fairly small, which may affect the system's responsiveness.

qwwdfsad:
(Changed the base twice for GH to catch the most recent `develop` state).

Thanks! 
Additional kudos for detailed commit message

Issue #3439, opened 2022-09-05
Internal API to retreive a single CPU-bound task from Dispatchers.Default


qwwdfsad:
IntelliJ IDEA platform has the following **core** API:

```
// In a blocking, non-suspendable world that is executed in ForkJoinPool

decompose() {
    // here happens decomposition of the task and .join()
}
```

Such API is composable in a way that it is possible to call `decompose` recursively from within subtasks.

It may be tempting to rewrite it with coroutines in a following manner:

```
fun decompose() = runBlocking {
    for (task in decomposition) {
        launch(Dispatchers.Default) { task() }
    }
}
```

The problem is -- this code is prone to both starvation and deadlocks -- each call to decompose from `Dispatchers.Default` occupies at least one thread in the backing executor for `joinBlocking()`, meaning that eventually, such code may deadlock.

It is not the case for the FJP-based solution, because `.join` on any FJP task not only awaits for the task to complete but also helps to execute subtasks if the joining thread is part of the target's FJP.

It doesn't seem that we can provide a full-blown FJP-like API out-of-the-box, but as the first steps we can provide an internal TBD API `getSingleDispatchersDefaultTaskOrNull`. With such API, it is possible to emulate `decompose` API and, after initial evaluation by the IDEA team, we may want to promote it to `@Delicate` API as one more way to bridge blocking calls with coroutine dispatchers.

qwwdfsad:
I would love to know how we can implement `restoreThreadContext`.

AFAIR there is no API in Java to clean up and later recover all thread locals. 
But, given that we know the current `ThreadContextElement` (e.g. if it works for you to pass it as an argument), we can probably provide an API for isolating **these specific elements**.

dovchinnikov:
We've recently had a starved thread situation after incorrect migration to coroutines in IJ (https://youtrack.jetbrains.com/issue/IDEA-302358), but it was using `Dispatchers.IO.limitedParallelism(xx)`. Thus I'd like to update the request to: API to retrieve a task from the current dispatcher. We are interested in `Dispatchers.Default`, `Dispatchers.IO` and `limitedParallelism` on any of them.

dovchinnikov:
Another possible approach is to provide a single-function API with the semantics like this: 

// NB return type is intentionally Nothing
public suspend fun processEventsInCurrentThread(): Nothing {
    assert {
        ThreadLocalEventLoop.currentOrNull() != null // only allowed inside runBlocking
    }
[...]

PR #3572, opened 2023-01-03
Preliminary prototype of #3439

qwwdfsad:


kotlinx-coroutines-core/jvm/src/EventLoop.kt
```diff
@@ -45,6 +49,78 @@ internal actual fun createEventLoop(): EventLoop = BlockingEventLoop(Thread.curr
...
+ *
+ * This function is inteded to be used on the result of `Thread.currentThread()` for diagnostic
+ * purposes, and is declared as an extension only to avoid top-level scope pollution.
+ */
+@InternalCoroutinesApi
+@DelicateCoroutinesApi
+public fun Thread.isIoDispatcherThread(): Boolean {
+    if (this !is CoroutineScheduler.Worker) return false
+    return isIo()
+}
```
dkhalanskyjb, line 125: Looks very similar to the functions at the bottom of `CoroutineScheduler.kt` and would probably enjoy their company.

kotlinx-coroutines-core/jvm/src/EventLoop.kt
```diff
@@ -45,6 +49,78 @@ internal actual fun createEventLoop(): EventLoop = BlockingEventLoop(Thread.curr
...
+ *   work to current dispatcher may arrive **later** from external sources [1]
+ *
+ * [1] -- this is also a technicality that can be solved in kotlinx.coroutines itself, but unfortunately requires
+ *        a tremendous effort.
+ *
+ * @throws IllegalStateException if the current thread is not system dispatcher thread
+ */
+@InternalCoroutinesApi
+@DelicateCoroutinesApi
+public fun runSingleTaskFromCurrentSystemDispatcher(): Long {
```
dkhalanskyjb, line 106: Instead of `public`, we could make them `@PublishedApi internal`. This way, we would have the guarantees of binary compatibility but also not expose an API that we don't fully understand yet. We could offer help with `@Suppress`ing this to any parties that are large enough to need access to this.

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -138,23 +153,41 @@ internal class WorkQueue {
...
+
+    private fun pollWithMode(/* Only blocking OR only CPU */onlyBlocking: Boolean): Task? {
         val start = consumerIndex.value
         var end = producerIndex.value
-
-        while (start != end && blockingTasksInBuffer.value > 0) {
-            val task = tryExtractBlockingTask(--end)
+        // CPU or (BLOCKING & hasBlocking)
+        val shouldProceed = !onlyBlocking || blockingTasksInBuffer.value > 0
+        while (start != end && shouldProceed) {
```
dkhalanskyjb, line 176: `shouldProceed` doesn't update. It doesn't feel like it's intentional, but if it is, it can be moved to an `if` wrapping the `while`.

dkhalanskyjb (commented):
Only glanced through it quickly for now since it's still a draft.

kotlinx-coroutines-core/jvm/src/EventLoop.kt
qwwdfsad: I've put it here solely because `processNextEventInCurrentThread` and I wanted to keep these "internal event loop spinners" in a single place to keep track of

kotlinx-coroutines-core/jvm/src/EventLoop.kt
```diff
@@ -45,6 +49,78 @@ internal actual fun createEventLoop(): EventLoop = BlockingEventLoop(Thread.curr
...
+ * thus blocking at least one thread. To avoid deadlocks and starvation during large hierarchical decompositions,
+ * 'runDecomposedTaskAndJoinIt' should not only block, but also **help** executing the task or other tasks
+ * until an arbitrary condition is satisfied.
+ *
+ * See #3439 for additional details.
+ *
+ * ### Limitations and caveats
+ *
+ * - Executes tasks in-place, thus potentially leaking irrelevant thread-locals from the current thread
+ * - Is not 100% effective, it waits for [Long] returned nanoseconds even when work arrives immediately after returning
```
dovchinnikov, line 94: re:
> it waits for [Long] returned nanoseconds

This might be an artefact of a previous implementation. AFAIU at this moment the function returns the amount of time until the next task would be available for stealing, and it's up to the caller what to do with that time.

qwwdfsad:
Closing in the favor of #3641

PR #3580, opened 2023-01-06
Do not propagate exceptions to CoroutineExceptionHandler in 'future' …

### LINKED ISSUE #3452: CoroutineScope.future() may trigger global exception handler on cancellation
### Expected behavior

`future` coroutine builder delivers all errors via the returned object, just like `async`.

### Actual behavior

`future` coroutine builder can suddenly trigger global exception handler that may crash the application.

Sample code:
```kotlin
fun main() {
    val future = GlobalScope.future {
        error("foo")
    }
    future.cancel(true)
    Thread.sleep(100)
}
```

Result:
`Exception in thread "DefaultDispatcher-worker-1" java.lang.IllegalStateException: foo`

qwwdfsad:
…builder if it has been cancelled in order to be consistent with other future implementations

Fixes #3452


qwwdfsad:
I've also checked the implementation of jdk HTTP client -- they indeed do the same and ignore `false` returned from `completeExceptionally` during any exceptional code-path

dkhalanskyjb (approved):

PR #3581, opened 2023-01-09
Remove confusing comma in Coroutines Hands-on

SebastianAigner:
This change makes sure the sentence reflects the actual meaning: We're switching to the type of API that uses callbacks, rather than switching to the API, which incidentally happens to use callbacks.


SebastianAigner:
It's precisely _not_ about switching to an API called Retrofit. The previous example also uses the Retrofit API. This sentence specifically talks about _the part of the Retrofit API that happens to use callbacks_.

In other words, we're switching "[from the blocking Retrofit API] to the callback-based Retrofit API".

dkhalanskyjb:
Ah, yes, you're right! Sorry for the noise.

PR #3582, opened 2023-01-10
Fix example code

isaialvarado:
GlobalScope should not be in loadContributorsConcurrent if you want the function to be cancellable by using the cancel button.

### DIFF
--- a/docs/topics/coroutines-and-channels.md
+++ b/docs/topics/coroutines-and-channels.md
@@ -905,7 +905,7 @@ the parent coroutine.
         req: RequestData
     ): List<User> = coroutineScope {
         // ...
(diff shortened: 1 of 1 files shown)


qwwdfsad:
@danil-pavlov PTAL, it seems like the guide have this mixed up

danil-pavlov:
@svtk, could you take a look, please? Maybe we need to update the solution in the example project as well

svtk:
@danil-pavlov The solution and the previous version are correct. It seems to be introduced when we moved the text to the docs.

@isaialvarado Thanks for pointing it out!

danil-pavlov (approved):
Thank you!

Issue #3588, opened 2023-01-18
Code that uses `withTimeout` often fails and cannot be safely test covered using the `kotlinx.coroutines.test`


amal:
Right now any code that uses `withTimeout` in the `TestScope` or background `TestScope` [can be failed](https://github.com/Kotlin/kotlinx.coroutines/blob/dea2ca5/kotlinx-coroutines-test/README.md#using-withtimeout-inside-runtest) with `TimeoutCancellationException` on any suspension. And in such case it will fail with misleading, I would even say completely false message, e.g.:
```
Timed out waiting for 5000 ms
```
![image](https://user-images.githubusercontent.com/423417/213269000-c4c9a8f0-3292-47be-a59b-94bb6dc5f4d8.png)

This is an exaggerated example where it is obvious (long timeout with much shorter test execution time). In reality, it is often difficult to understand that the matter is actually in `TestScope` usage.

I believe that it's a very unexpected behavior for developers and a source of lost hours and maybe even days to understand the source of the problem. I even knew about such behavior earlier, but in a few months forgot about it and again fell into this trap, wasting hours when tried to write more tests for the [MVI library](https://github.com/fluxo-kt/fluxo-mvi).

As already mentioned in #1374, the problem is further complicated by the fact that `TimeoutCancellationException` is `CancellationException`, thus silently canceling parent coroutines.

Just to mention, here [are](https://github.com/Kotlin/kotlinx.coroutines/issues/3270#issuecomment-1127549299) [examples](https://github.com/Kotlin/kotlinx.coroutines/issues/3270#issuecomment-1145786951) where other people think about `withTimeout` as reliable function in the context of coroutines tests, which it is not right now!

## Proposal

1. I am sure that **code wrapped in the `withTimeout` should work as usual in any tests without sudden fails**.

2. If somehow it's not possible or as a migration measure what can be done better:
**If `withTimeout` fails because of the `TestScope`, change the error message so that the cause would be obvious.**

## Workaround

For those suffering in the same situation 🤗
Right now I use this replacement instead of the unsafe `withContext`. At least it works as expected right now.
I hope it will not be broken with something like #982

```kotlin
public suspend inline fun <T> withTimeoutSafe(timeMillis: Long, crossinline block: suspend CoroutineScope.() -> T): T {
    // just for convenience, can be safely removed
    contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }
    val context = currentCoroutineContext()
[...]

Issue #3179, opened 2022-02-08
contlix.coroutines.test: Option to disable cirtual time and delay skipping


sergeych:
It is often a need to keep regular `delay()` behavior with actual delay, for example:

- debugging code that uses time-depending libraries that know nothing of virtual time (e.g. `kotlinx.datetime`)
- debugging network protocols where real delays are a must.

It could be something like `runTest(noVirtualTime=true) { /}`

joffrey-bion:
Yep, I also think it's really important. Some integration tests need to interact with real-time systems, and delay-skipping just breaks here.

The purpose of `kotlinx-coroutines-test` changed when it became multiplatform. It is no longer *only* meant for delay skipping, it is meant to help run multiplatform tests using coroutines (especially regarding the JS stuff).

Note that there already was a discussion on this topic, but I think it's worth bringing back:
https://github.com/Kotlin/kotlinx.coroutines/issues/3094

dkhalanskyjb:
Sure, we recognize the problem, it's just that the API shape is currently unclear. `runTest(noVirtualTime = true) { }` is simply not flexible enough: it's possible that some parts of the test do need to avoid delay-skipping, but some do not. Adding it as a parameter to the test coroutine dispatcher would be a better option, like
```kotlin
withContext(StandardTestDispatcher(scheduler = testScheduler, skipDelays = false)) {
  // do whatever
}
```
But then, there's not much reason to use the test dispatchers at all.

We would prefer to take our time with this because there are other, more intrusive aspects of the test API that would influence the API of this functionality—like mocking the IO and Default dispatchers—and it doesn't look like the issue is urgent, as there is a workaround (from the linked thread):
```kotlin
withContext(Dispatchers.Default.limitedParallelism(1)) {
  // delays are not skipped here
}
```
If there are usage patterns not covered by this, please share them.

EanLombardo:
I think the issue perhaps with the current API is the naming, not the functionality.

"runTest" is such a generic name it implies that it is just "the way to do tests", StandardTestDispatcher implies this even stronger.

With such a generic name I would think think that the functionality should be generic. In some contexts skipping delays breaks correctness contracts (networking), and in others it can cause heavy workloads to overrun (polling services in a cross system test). Performing a delay when it is not needed however at worst makes a test time out or hang, which can be hard to debug sure, but those cases should be fairly rare. The name "runTest" doesn't really imply that skipping should happen by default.

[...]

PR #3591, opened 2023-01-19
Indicate the use of virtual time in withTimeout in TestScope

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/3588

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -561,6 +561,7 @@ public final class kotlinx/coroutines/ThreadPoolDispatcherKt {
 }
 
 public final class kotlinx/coroutines/TimeoutCancellationException : java/util/concurrent/CancellationException, kotlinx/coroutines/CopyableThrowable {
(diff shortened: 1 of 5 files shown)


dkhalanskyjb:
I wanted to do both because I didn't think it was reasonable to do something like
```kotlin
public interface Delay {
  val whatCanBeDoneToAvoidATimeout: String?
```
and just a message from `withTimeout` is not very actionable. The intention was for people to see that something test-module-related is wrong, try `withTimeout` in `runTest` itself in the debugging process, and then see the solution.

Though I guess if I manage to implement https://github.com/Kotlin/kotlinx.coroutines/issues/3179 after all (it turned out to be very demanding) + an informative message in `withTimeout`, we will be able to rely on word of mouth to spread the correct solution to timeouts that have a specific test-module related message.

qwwdfsad:
The reasoning works for me.

It seems like an alternative could be a slightly better `Delay` interface -- _maybe_ it should be in control of both scheduling and timeout exceptions (e.g. to enhance stracktraces, add diagnostics to exceptions, provide domain-specific message etc.).
It should be achievable the moment we'll figure `-Xjvm-defaults`

PR #3593, opened 2023-01-19
Remove ThreadLocal from ThreadLocalMap when finishing UndispatchedCor…

### LINKED ISSUE #3592: UndispatchedCoroutine in withContext creates unreasonable CPU pressure due to GC nepotism
This is a condensed description of the issue Space is suffering, so some details (e.g. the actual CPU impact, CPU snapshots, etc.) are omitted.

For coroutines, in order to keep track of `ThreadContextElement`, each undispatched (the one that doesn't change current coroutine dispatcher) `withContext`, a special thread-local variable is spawned in order to preserve and restore execution context in the corresponding thread.

It leads to multiple problems:

* If the coroutine *does not* use thread-local elements, the corresponding system `ThreadLocal` is instantiated and queried, while being later reset to `null`
* If system-wide `ThreadContextElement` is used in the application, then application spawns as much thread locals as there are calls to `withContext.

The rest is `ThreadLocal` implementation peculiarities: 

* `ThreadLocal` is actually stored in Thread's local map with open addressing by weak reference
* `ThreadLocal` map entry is allocated on the first access (e.g. `get()`) even if it's `null` or wasn't `set` and has no initializer
* If `ThreadLocal` is unreachable, it's collected, but its weak-reference slot is not 
* When thread-local map has reached its fill factor or encountered "stale" (GC-collected) entry, it attempts to clean it up with potential rehash
* If coroutine lives long enough, it is promoted to old generation along with its `ThreadLocal` object
   - Because of that, entry can only be collected after an old generation GC which is rare enough event
   - Until then, any thread-local op that encounters rehash/stale entry will burn quite a lot of CPU processing 4k+ open-addressing hashmap

qwwdfsad:
…outine

* It addresses the problem with ThreadLocalMap.entries that may outlive the coroutine lifecycle and interfere with CPU consumption of other thread-locals on the same thread
* No test provided as this is a non-functioanl change. The only reasonable way to check it is to reflectively walk over Thread class which is prohibited by Java since 11+. The only way is to eyeball Thread.currentThread().threadLocals size in debugger in the properly crafted unit test

Fixes #3592

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
@@ -217,16 +217,19 @@ internal actual class UndispatchedCoroutine<in T>actual constructor (
     }
 
     fun clearThreadContext(): Boolean {
-        if (threadStateToRecover.get() == null) return false
-        threadStateToRecover.set(null)
+        if (threadStateToRecover.get() == null) {
+            threadStateToRecover.remove()
+            return false
+        }
+        threadStateToRecover.remove()
         return true
     }
 
     override fun afterResume(state: Any?) {
         threadStateToRecover.get()?.let { (ctx, value) ->
             restoreThreadContext(ctx, value)
-            threadStateToRecover.set(null)
         }
+        threadStateToRecover.remove()
         // resume undispatched -- update context but stay on the same dispatcher
         val result = recoverResult(state, uCont)
         withContinuationContext(uCont, null) {
--- a/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
@@ -179,7 +179,21 @@ internal actual class UndispatchedCoroutine<in T>actual constructor (
      *     // and it also calls saveThreadContext and clearThreadContext
      * }
      */
-    private var threadStateToRecover = ThreadLocal<Pair<CoroutineContext, Any?>>()
+    private val threadStateToRecover = ThreadLocal<Pair<CoroutineContext, Any?>>()
+
+    /*
+     * Indicates that a coroutine has a thread local elements associated with it
+     * and that 'threadStateToRecover'.
+     * Better than nullable thread-local for easier debugging.
+     *
+     * It is used as a performance optimization to avoid 'threadStateToRecover' and
+     * is prone to false-positives as it is never reset: otherwise
+     * it may lead to logical data races between suspensions point where
+     * coroutine is yet being suspended in one thread while already being resumed
+     * in another.
+     */
+    @Volatile
+    private var threadLocalIsSet = false
 
     init {
         /*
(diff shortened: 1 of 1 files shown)


qwwdfsad:
The next step is not to create thread-local if there doesn't exist the corresponding thread context element

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -217,15 +217,18 @@ internal actual class UndispatchedCoroutine<in T>actual constructor (
...
+        if (threadStateToRecover.get() == null) {
+            threadStateToRecover.remove()
+            return false
+        }
+        threadStateToRecover.remove()
         return true
     }
 
     override fun afterResume(state: Any?) {
         threadStateToRecover.get()?.let { (ctx, value) ->
```
whyoleg, line 229: isn't it needed to call `remove()` even if `get()` returns `null` (by analogy with `clearThreadContext`?

qwwdfsad: Nice catch!

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -213,19 +227,25 @@ internal actual class UndispatchedCoroutine<in T>actual constructor (
...
     fun saveThreadContext(context: CoroutineContext, oldValue: Any?) {
+        threadLocalIsSet = true // Specify that thread-local is touched at all
         threadStateToRecover.set(context to oldValue)
     }
 
     fun clearThreadContext(): Boolean {
-        if (threadStateToRecover.get() == null) return false
-        threadStateToRecover.set(null)
+        if (threadLocalIsSet && threadStateToRecover.get() == null) {
+            threadStateToRecover.remove()
```
dkhalanskyjb, line 236: I think such brittle changes should be heavily documented so that they don't get reverted by seemingly innocent edits. I *guess* the reason to call `threadStateToRecover.remove()` here is to completely eliminate the thread-local variable so that it doesn't get traversed (is this correct at all?), but there's no chance I'd understand the danger of removing this line if I didn't know or forgot about this issue.

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -213,19 +227,25 @@ internal actual class UndispatchedCoroutine<in T>actual constructor (
...
+            return false
+        }
+        threadStateToRecover.remove()
         return true
```
dkhalanskyjb, line 240: Minor:
```kotlin
!(threadLocalIsSet && threadStateToRecover.get() == null).also { threadStateToRecover.remove() }
```
I think it conveys the idea much better.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
qwwdfsad: I've added some, could you please check if that is enough along with commit messages?

dkhalanskyjb (approved):
This is a really good explanation! And appropriately scarily long, too, everyone will be sufficiently intimidated by this.

PR #3595, opened 2023-01-20
Don't allocate threads on every dispatch in Native's thread pools

dkhalanskyjb:
See also https://github.com/Kotlin/kotlinx.coroutines/pull/3576


kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -73,38 +74,75 @@ private class MultiWorkerDispatcher(
...
+            val state = tasksAndWorkersCounter.getAndUpdate {
+                if (it.isClosed() && !it.hasTasks()) return@runBlocking
+                it - 2
+            }
+            if (state.hasTasks()) {
+                // we promised to process a task, and there are some
+                tasksQueue.receive().run()
+            } else {
+                availableWorkers.send(privateChannel)
+                val task = privateChannel.receiveCatching().getOrNull()?.run()
```
dkhalanskyjb, line 105: I think we could quite easily add a `withTimeout` here, then `getAndUpdate { if still considered to be sleeping, remove oneself and terminate }`, and retry the `obtainWorker()`, and this will allow graceful worker termination. I think it would complicate the code somewhat, so didn't initially implement this. @qwwdfsad, do you think this could be worth it?

dkhalanskyjb:
@ndkoval, the spirit of the SQS is strong here! :)

kotlinx-coroutines-core/jvm/test/knit/ClosedAfterGuideTestExecutor.kt
```diff
@@ -44,7 +44,10 @@ private class ClosedAfterGuideTestDispatcher(
     }
 
     override fun close() {
-        (executor as ExecutorService).shutdown()
+        (executor as ExecutorService).apply {
+            shutdown()
+            awaitTermination(1, TimeUnit.MINUTES)
+        }
```
dkhalanskyjb, line 50: No strong opinion about this change, I just found it severely surprising that there's no way to wait for the tasks to finish executing. Can undo this, move the corresponding test to native-only, and change the doc to state that the previously-submitted tasks may execute *in parallel* to the caller of `close()`.

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -73,38 +74,80 @@ private class MultiWorkerDispatcher(
     workersCount: Int
 ) : CloseableCoroutineDispatcher() {
     private val tasksQueue = Channel<Runnable>(Channel.UNLIMITED)
+    private val availableWorkers = Channel<CancellableContinuation<Runnable>>(Channel.UNLIMITED)
```
qwwdfsad, line 77: Shouldn't its capacity be limited with `workersCount`?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -73,38 +74,80 @@ private class MultiWorkerDispatcher(
...
     private val workerPool = OnDemandAllocatingPool(workersCount) {
         Worker.start(name = "$name-$it").apply {
             executeAfter { workerRunLoop() }
         }
     }
 
+    /**
+     * (number of tasks - number of workers) * 2 + (1 if closed)
```
qwwdfsad, line 85: Neat!

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -73,38 +74,80 @@ private class MultiWorkerDispatcher(
...
-            /**
-             * Any unhandled exception here will pass through worker's boundary and will be properly reported.
-             */
-            task.run()
+        while (true) {
+            val state = tasksAndWorkersCounter.getAndUpdate {
+                if (it.isClosed() && !it.hasTasks()) return@runBlocking
+                it - 2
+            }
+            if (state.hasTasks()) {
```
qwwdfsad, line 99: Nice idea with state snapshotting!

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -73,38 +74,80 @@ private class MultiWorkerDispatcher(
...
+                // we promised to process a task, and there are some
+                tasksQueue.receive().run()
+            } else {
+                try {
+                    suspendCancellableCoroutine {
+                        availableWorkers.trySend(it)
+                    }.run()
+                } catch (e: CancellationException) {
+                    /** we are cancelled from [close] and thus will never get back to this branch of code,
+                    but there may still be pending work, so we can't just exit here. */
```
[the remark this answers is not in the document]
dkhalanskyjb: The code is written with this in mind: we can only end up on line 101 if we decremented the task count from a positive number, and it was positive because someone promised to provide a task (by incrementing the number). Notice that, if the counter was successfully incremented, there's no choice but to submit a task, even if the dispatcher is closed after that increment, so the task will always end up in the queue! As a result, the number of threads waiting on line 101 (now or in the near future) is always the same as the number of tasks that entered the channel (or will enter it shortly).

If this is true, all threads are guaranteed to leave this loop once no more tasks are left for them to execute. The stress test suggests that this does hold.

Does this make sense? Maybe there are some places where I could add a helpful comment?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
dkhalanskyjb: Not do much an idea as a habit at this point!

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
dkhalanskyjb: It can. Would it improve something? I wanted the code to say "we don't want to ever drop anything here", and "buffer size N" does imply "drop if larger than N."

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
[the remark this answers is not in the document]
qwwdfsad: It seems like I'm missing something.

Consider the following scenario: 

* `close` is being in progress
* `close` cancels one of the waiting workers (W1), it catches `CancellationException` at 107, gets preempted
* W2 invokes `dispatch`: it increments the counter, gets preempted
* W1 gets scheduled, goes to the next loop iteration, observes the positive tasks counter, decrements it, goes to `receive`. There is no tasks in the channel (W2 got preempted), so it attempts to suspend and immediately throws `CancellationException` right from 101.
* W2 finished `dispatch` and sends task to the channel. Counter is already 0 and there are no worker who can execute it.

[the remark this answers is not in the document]
dkhalanskyjb: > throws `CancellationException` right from 101 because it's already cancelled.

Why? Only the channel for personal communication is closed, the global one always lives, exactly because of the scenario you're describing.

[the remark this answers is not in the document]
qwwdfsad: Oh, I see what I got mixed up. Thanks!

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
qwwdfsad: Nothing really, was rather surprised that it has unlimited, while effectively never having more than `workersCount` elements/waiters.

>we don't want to ever drop anything here

Maybe its worth adding asserts/postconditions to each `trySend`?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
qwwdfsad: My vote is to have a follow-up PR with the change

qwwdfsad:
Seems like we only have two minor questions and then it's good to go

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
dkhalanskyjb: I added some checks. Are they what you meant?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -73,38 +74,80 @@ private class MultiWorkerDispatcher(
...
     }
 
-    override fun dispatch(context: CoroutineContext, block: Runnable) {
-        fun throwClosed(block: Runnable) {
-            throw IllegalStateException("Dispatcher $name was closed, attempted to schedule: $block")
+    private fun obtainWorker(): CancellableContinuation<Runnable> {
+        // spin loop until a worker that promised to be here actually arrives.
+        while (true) {
+            val result = availableWorkers.tryReceive()
+            return result.getOrNull() ?: continue
```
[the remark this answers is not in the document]
dkhalanskyjb: Fixed.

kotlinx-coroutines-core/native/test/MultithreadedDispatchersTest.kt
```diff
@@ -0,0 +1,50 @@
...
+                        // intentionally empty
+                    }
+                    spin(encounteredWorkers)
+                }
+                listOf(coroutine1, coroutine2).joinAll()
+                assertEquals(2, encounteredWorkers.size)
+            }
+        } finally {
+            dispatcher.close()
+        }
```
dkhalanskyjb, line 48: This test failed in the CI with `encounteredWorkers.size == 1`, but I fail to see how it's possible. Can different workers be equal to each other?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
qwwdfsad: That's it, thanks

kotlinx-coroutines-core/native/test/MultithreadedDispatchersTest.kt
qwwdfsad: It seems like there are at least two data-races:

* One is over `canStart1`/`canStart2`
* Another one is concurrent mutation of non-thread safe `MutableSet` which probably causes the failure

dkhalanskyjb: > non-thread-safe `MultiSet`

Wow, what a ridiculous mistake. Looks like I've exhausted my resource for the understanding of concurrency for this PR on the main code.

> `canStart1`/`canStart2`

I still fail to see the problem, sorry, could you spell it out?

qwwdfsad: We all have been there :)

>I still fail to see the problem, sorry, could you spell it out?

Non-atomic variable is read and write concurrently, I honestly not sure what guarantees K/N (or LLVM) provides in such scenarios. In theory, the compiler is allowed to host memory read in `while (!canStart*)` from the loop

qwwdfsad (approved):
LGTM. It seems like Kover verification started to fail, probably code coverage dropped somewhere, please take a look before merging

dkhalanskyjb:
Kover failed because it tried to run all the tests, and `testStackMergeWithContext` failed. Locally, `./gradlew koverVerify koverReport -Pkover.enabled=true` passes without issues.

PR #3602, opened 2023-01-26
Add optins required in Kotlin 1.9

demiurg906:


kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
```diff
@@ -43,6 +43,7 @@ internal actual object DefaultExecutor : EventLoopImplBase(), Runnable {
 
     @Suppress("ObjectPropertyName")
     @Volatile
+    @OptIn(ExperimentalStdlibApi::class)
```
qwwdfsad, line 46: This is clearly a breaking change in the language, something suspicious is here and we cannot accept it as is.

demiurg906: I reported it to `kotlin-design`. Fix of this issue is definitely not the part of migration master of kotlin to 1.9

qwwdfsad: Well, it means that the master is broken right now, and that it is definitely the part of preparing master to 1.9

woainikk (approved):

PR #3603, opened 2023-01-27
Allow specifying the timeout for `runTest`

### LINKED ISSUE #3270: Consider lowering `runTest` default timeout
60 seconds is a _really_ long time for nothing to happen. It's like an eternity for a CPU, especially one which is skipping delays. I would personally prefer something low like 2s (as is the default for JS's Mocha runner) but something like 10s seems a reasonable default that it covers 99.9% of real-world usage without also being a whole minute.

dkhalanskyjb:
Deprecate `dispatchTimeoutMs`, as this is a confusing implementation detail that made it to the final API.

We use the fact that the `runTest(Duration)` overload was never published, so we can reuse it to have the `Duration` mean the whole-test timeout in a backward-compatible manner.

Fixes #3270

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -160,8 +160,9 @@ public fun runTest(
     timeout: Duration = DEFAULT_TIMEOUT,
     testBody: suspend TestScope.() -> Unit
 ): TestResult {
-    if (context[RunningInRunTest] != null)
-        throw IllegalStateException("Calls to `runTest` can't be nested. Please read the docs on `TestResult` for details.")
+    check(context[RunningInRunTest] == null) {
+        "Calls to `runTest` can't be nested. Please read the docs on `TestResult` for details."
+    }
     return TestScope(context + RunningInRunTest).runTest(timeout, testBody)
 }
 
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -303,19 +303,19 @@ public fun runTest(
 public fun TestScope.runTest(
     timeout: Duration = DEFAULT_TIMEOUT,
     testBody: suspend TestScope.() -> Unit
-): TestResult = asSpecificImplementation().let {
-    it.enter()
+): TestResult = asSpecificImplementation().let { scope ->
+    scope.enter()
     createTestResult {
         /** TODO: moving this [AbstractCoroutine.start] call outside [createTestResult] fails on JS. */
-        it.start(CoroutineStart.UNDISPATCHED, it) {
+        scope.start(CoroutineStart.UNDISPATCHED, scope) {
+            /* we're using `UNDISPATCHED` to avoid the event loop, but we do want to set up the timeout machinery
+            before any code executes, so we have to park here. */
+            yield()
             testBody()
         }
-        /**
-         * We run the tasks in the test coroutine using [Dispatchers.Default]. On JS, this does nothing particularly,
-         * but on the JVM and Native, this means that the timeout can be processed even while the test runner is busy
-         * doing some synchronous work.
-         */
-        val workRunner = launch(Dispatchers.Default + CoroutineName("kotlinx.coroutines.test runner")) {
+        var timeoutError: Throwable? = null
+        var cancellationException: CancellationException? = null
+        val workRunner = launch(CoroutineName("kotlinx.coroutines.test runner")) {
             while (true) {
                 val executedSomething = testScheduler.tryRunNextTaskUnless { !isActive }
                 if (executedSomething) {
@@ -323,47 +323,49 @@ public fun TestScope.runTest(
                      * procedure needs a chance to run concurrently. */
                     yield()
                 } else {
-                    // no more tasks, we should suspend until there are some more
+                    // waiting for the next task to be scheduled, or for the test runner to be cancelled
                     testScheduler.receiveDispatchEvent()
                 }
             }
         }
-        var timeoutError: Throwable? = null
         try {
             withTimeout(timeout) {
-                it.join()
+                coroutineContext.job.invokeOnCompletion(onCancelling = true) { exception ->
+                    if (exception is TimeoutCancellationException) {
+                        dumpCoroutines()
+                        val activeChildren = scope.children.filter(Job::isActive).toList()
+                        val completionCause = if (scope.isCancelled) scope.tryGetCompletionCause() else null
+                        var message = "After waiting for $timeout"
+                        if (completionCause == null)
+                            message += ", the test coroutine is not completing"
+                        if (activeChildren.isNotEmpty())
+                            message += ", there were active child jobs: $activeChildren"
+                        if (completionCause != null && activeChildren.isEmpty()) {
+                            message += if (scope.isCompleted)
+                                ", the test coroutine completed"
+                            else
+                                ", the test coroutine was not completed"
+                        }
+                        timeoutError = UncompletedCoroutinesError(message)
+                        cancellationException = CancellationException("The test timed out")
+                        (scope as Job).cancel(cancellationException!!)
+                    }
+                }
+                scope.join()
                 workRunner.cancelAndJoin()
             }
         } catch (_: TimeoutCancellationException) {
-            val activeChildren = it.children.filter(Job::isActive).toList()
-            val completionCause = if (it.isCancelled) it.tryGetCompletionCause() else null
-            var message = "After waiting for $timeout"
-            if (completionCause == null)
-                message += ", the test coroutine is not completing"
-            if (activeChildren.isNotEmpty())
-                message += ", there were active child jobs: $activeChildren"
-            if (completionCause != null && activeChildren.isEmpty()) {
-                message += if (it.isCompleted)
-                    ", the test coroutine completed"
-                else
-                    ", the test coroutine was not completed"
-            }
-            timeoutError = UncompletedCoroutinesError(message)
-            dumpCoroutines()
-            val cancellationException = CancellationException("The test timed out")
-            (it as Job).cancel(cancellationException)
-            // we can't abandon the work we're doing, so if it hanged, we'll still hang, despite the timeout.
-            it.join()
-            val completion = it.getCompletionExceptionOrNull()
+            scope.join()
+            val completion = scope.getCompletionExceptionOrNull()
             if (completion != null && completion !== cancellationException) {
-                timeoutError.addSuppressed(completion)
+                timeoutError!!.addSuppressed(completion)
             }
             workRunner.cancelAndJoin()
         } finally {
             backgroundScope.cancel()
             testScheduler.advanceUntilIdleOr { false }
-            val uncaughtExceptions = it.leave()
-            throwAll(timeoutError ?: it.getCompletionExceptionOrNull(), uncaughtExceptions)
+            val uncaughtExceptions = scope.leave()
+            throwAll(timeoutError ?: scope.getCompletionExceptionOrNull(), uncaughtExceptions)
         }
     }
 }
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -122,8 +122,7 @@ public expect class TestResult
  * #### Timing out
  *
  * There's a built-in timeout of 10 seconds for the test body. If the test body doesn't complete within this time,
- * then the test fails with an [UncompletedCoroutinesError]. The timeout can be changed by setting the [timeout]
- * parameter.
+ * then the test fails with an [AssertionError]. The timeout can be changed by setting the [timeout] parameter.
  *
  * The test finishes by the timeout procedure cancelling the test body. If the code inside the test body does not
  * respond to cancellation, we will not be able to make the test execution stop, in which case, the test will hang
@@ -294,7 +293,7 @@ public fun runTest(
 }
 
 /**
- * Performs [runTest] on an existing [TestScope].
+ * Performs [runTest] on an existing [TestScope]. See the documentation for [runTest] for details.
  */
 public fun TestScope.runTest(
     timeout: Duration = DEFAULT_TIMEOUT,
@@ -466,7 +465,8 @@ internal suspend fun <T : AbstractCoroutine<Unit>> CoroutineScope.runTestCorouti
                 lastKnownPosition.value = getLastKnownPosition()
                 val executedSomething = scheduler.tryRunNextTaskUnless { !isActive }
                 if (executedSomething) {
-                    // yield to check for cancellation
+                    /** Yield to check for cancellation. On JS, we can't use [ensureActive] here, as the cancellation
+                     * procedure needs a chance to run concurrently. */
                     yield()
                 } else {
                     // no more tasks, we should suspend until there are some more
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -301,10 +301,76 @@ public fun TestScope.runTest(
 ): TestResult = asSpecificImplementation().let {
     it.enter()
     createTestResult {
-        runTestCoroutine(it, null, timeout, TestScopeImpl::tryGetCompletionCause, testBody) {
+        /** TODO: moving this [AbstractCoroutine.start] call outside [createTestResult] fails on JS. */
+        it.start(CoroutineStart.UNDISPATCHED, it) {
+            testBody()
+        }
+        /**
+         * The thread in which the task was last seen executing.
+         */
+        val lastKnownPosition = MutableStateFlow<Any?>(null)
+
+        /**
+         * We run the tasks in the test coroutine using [Dispatchers.Default]. On JS, this does nothing particularly,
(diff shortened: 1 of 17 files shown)


kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
```diff
@@ -177,6 +177,14 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
         }
     }
 
+    /**
+     * Moves the virtual clock of this dispatcher forward by [the specified amount][delayTime], running the
+     * scheduled tasks in the meantime.
+     *
+     * @throws IllegalStateException if passed a negative [delay][delayTime].
+     */
+    public fun advanceTimeBy(delayTime: Duration): Unit = advanceTimeBy(delayTime.inWholeMicroseconds)
```
Lukellmann, line 186: I think this should use `delayTime.inWholeMilliseconds`, the parameter in the overload taking a `Long` is named `delayTimeMillis`.

dkhalanskyjb: Yep, you're right. This PR is marked as a draft because it's undertested, so such silly mistakes are very likely.

dkhalanskyjb:
Regarding the two changes (timeout + experimentality markers) for the price of one: I'll split the PR in two if it makes it easier to review this.

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,28 +276,33 @@ public fun runTest(
  *
  * @throws IllegalArgumentException if the [context] is invalid. See the [TestScope] constructor docs for details.
  */
-@ExperimentalCoroutinesApi
+@Deprecated(
+    "Define a total timeout for the whole test instead of using dispatchTimeoutMs. " +
+        "Warning: the proposed replacement is not identical as it uses 'dispatchTimeoutMs' as the timeout for the whole test!",
+    ReplaceWith("runTest(context, timeout = with(kotlin.time.Duration.Companion) { dispatchTimeoutMs.milliseconds }, testBody)"),
```
qwwdfsad, line 282: This replacement is way too verbose (also, I doubt it's going to work bug-less in complex contexts).
You probably can replace it (pun intended) with `ReplaceWith("runTest(context, timeout = dispatchTimeoutMs.milliseconds, testBody), imports = [kotlin.time.Duration.Companion]"` (maybe `kotlin.time.Duration.Companion.milliseconds` is required though)

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,28 +276,33 @@ public fun runTest(
...
+): TestResult {
+    if (context[RunningInRunTest] != null)
+        throw IllegalStateException("Calls to `runTest` can't be nested. Please read the docs on `TestResult` for details.")
+    @Suppress("DEPRECATION")
+    return TestScope(context + RunningInRunTest).runTest(dispatchTimeoutMs = dispatchTimeoutMs, testBody)
+}
 
 /**
  * Performs [runTest] on an existing [TestScope].
  */
```
qwwdfsad, line 298: Probably worth mention that [runTest] contains the full-blown documentation

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -350,59 +385,115 @@ internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutin
...
-     *    Instead, the test thread itself is used, by spawning a separate coroutine.
-     */
-    var completed = false
-    while (!completed) {
-        scheduler.advanceUntilIdle()
-        if (coroutine.isCompleted) {
-            /* don't even enter `withTimeout`; this allows to use a timeout of zero to check that there are no
+    if (dispatchTimeout != null) {
+        check(testTimeout == null) { "Only one of dispatchTimeout and testTimeout can be specified" }
+        /**
```
qwwdfsad, line 390: This is a bit opionated, so feel free to ignore.

I found that extracting this branch's content into  separate `runTestCoroutineLegacy` is much more straightforward -- it immediately becomes clear which code is meaningful and should be future-proof and which is not so. Also, simplifies the body for readers that do not care about legacy behaviour

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -350,59 +385,115 @@ internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutin
...
+        val workRunner = launch(Dispatchers.Default + CoroutineName("kotlinx.coroutines.test runner")) {
             while (true) {
-                scheduler.tryRunNextTaskUnless { !isActive }
-                // yield so that the `select` below has a chance to check if its conditions are fulfilled
-                yield()
+                lastKnownPosition.value = getLastKnownPosition()
+                val executedSomething = scheduler.tryRunNextTaskUnless { !isActive }
+                if (executedSomething) {
+                    // yield to check for cancellation
+                    yield()
```
qwwdfsad, line 470: We have `ensureActive` for that

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -350,59 +385,115 @@ internal suspend fun <T: AbstractCoroutine<Unit>> CoroutineScope.runTestCoroutin
...
+             * There's a race that may lead to the misleading results here, but it's better than nothing.
+             * The race: `lastKnownPosition` is read, then the task executed in `workRunner` completes,
+             * then `updateStacktrace` does its thing, but the thread is already busy doing something else.
+             */
+            updateStacktrace(exception, lastKnownPosition.value)
+            // we can't abandon the work we're doing, so if it hanged, we'll still hang, despite the timeout.
+            (coroutine as CoroutineScope).cancel("The test timed out", exception)
+            coroutine.join()
+            workRunner.cancelAndJoin()
+            scheduler.advanceUntilIdleOr { false }
```
qwwdfsad, line 495: Consider the following scenario: the coroutine is timed out and properly joined, but some auxiliary watchdog wasn't cancelled. Does it mean that here we'll hang? 

I do not object against this behaviour, rather pointing out potential scenarios

qwwdfsad (changes requested):

kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
```diff
@@ -153,10 +154,18 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
...
     @ExperimentalCoroutinesApi
-    public fun advanceTimeBy(delayTimeMillis: Long) {
-        require(delayTimeMillis >= 0) { "Can not advance time by a negative delay: $delayTimeMillis" }
+    public fun advanceTimeBy(delayTimeMillis: Long): Unit = advanceTimeBy(delayTimeMillis.milliseconds)
+
+    /**
+     * Moves the virtual clock of this dispatcher forward by [the specified amount][delayTime], running the
+     * scheduled tasks in the meantime.
+     *
+     * @throws IllegalStateException if passed a negative [delay][delayTime].
```
hfhbd, line 163: Nit: `require` throws an `IllegalArgumentException`, `check` throws `IllegalStateException`

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -123,6 +120,16 @@ public fun TestScope.runCurrent(): Unit = testScheduler.runCurrent()
 @ExperimentalCoroutinesApi
 public fun TestScope.advanceTimeBy(delayTimeMillis: Long): Unit = testScheduler.advanceTimeBy(delayTimeMillis)
 
+/**
+ * Moves the virtual clock of this dispatcher forward by [the specified amount][delayTime], running the
+ * scheduled tasks in the meantime.
+ *
+ * @throws IllegalStateException if passed a negative [delay][delayTime].
```
hfhbd, line 127: Here too

kotlinx-coroutines-test/common/src/TestBuilders.kt
dkhalanskyjb: If something's executing in the test scheduler and doesn't respond to cancellation, yes, we'll keep executing it even if it means that we'll hang.

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,46 +277,138 @@ public fun runTest(
...
+        val lastKnownPosition = MutableStateFlow<Any?>(null)
+
+        /**
+         * We run the tasks in the test coroutine using [Dispatchers.Default]. On JS, this does nothing particularly,
+         * but on the JVM and Native, this means that the timeout can be processed even while the test runner is busy
+         * doing some synchronous work.
+         */
+        val workRunner = launch(Dispatchers.Default + CoroutineName("kotlinx.coroutines.test runner")) {
+            while (true) {
+                lastKnownPosition.value = getLastKnownPosition()
```
qwwdfsad, line 324: It seems like I'm missing something.

Could you please explain the scenario where `lastKnownPosition` is something different from an arbitrary `Dispatchers.Default` thread?

dkhalanskyjb: This question prompted me to remove the code that attempts to do tricks with stacktraces. I agree that it was only situationally useful but could also cause much confusion. Unfortunately, I don't see a way to make the timeout stacktrace consistently useful.

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,46 +278,126 @@ public fun runTest(
...
+        /** TODO: moving this [AbstractCoroutine.start] call outside [createTestResult] fails on JS. */
+        it.start(CoroutineStart.UNDISPATCHED, it) {
+            testBody()
+        }
+        /**
+         * We run the tasks in the test coroutine using [Dispatchers.Default]. On JS, this does nothing particularly,
+         * but on the JVM and Native, this means that the timeout can be processed even while the test runner is busy
+         * doing some synchronous work.
+         */
+        val workRunner = launch(Dispatchers.Default + CoroutineName("kotlinx.coroutines.test runner")) {
```
qwwdfsad, line 318: Could you please elaborate a bit on the reasoning behind this change?

From now on, `runTest` behaves in a bit surprising manner -- it changes the execution thread from "main" (where all tests are launched, a single-threaded default) to "worker" and it may be an observable behaviour:
```
    val tl = ThreadLocal.withInitial { 42 }

    @Test
    fun foo() = runTest {
        println(Thread.currentThread())
        tl.set(41)
        println(tl.get())
        yield()
        println(Thread.currentThread())
        println(tl.get())
    }
```

it also may pollute thread-locals arbitrarily (I believe this is the reason `TimeoutRule` creates standalone thread each time).

It may potentially confuse tooling around tests and interferes with production code (less threads in `Dispatchers.Default` that can actually matter on low-number-of-cores-machines, and interferes with thread-locals when used carelessly) for the sake of a more granular timeout.

Maybe it's worth to run this loop in-place (instead of standalone coroutine) and check for timeout cancellation at the end. We are going to wait for syncrhnous work anyway in order to terminate the test, but the interference issues will be addressed

dkhalanskyjb: The reason is simple: I wanted to get a coroutine dump at the moment when everything's hanged. I always thought `TimeoutRule`, too, spawned a separate thread for this reason. You do raise a good point though that due to `CoroutineStart.UNDISPATCHED`, the threading behavior is surprising.

I see these ways to mitigate the problem:
* Start a separate thread for each test and ensure that the test immediately executes on that thread.
* Do run the loop on the main thread, but replace `withTimeout` with custom code that cancels the test and dumps the coroutines.

qwwdfsad: Thanks!

>Start a separate thread for each test and ensure that the test immediately executes on that thread.

Let's consider this one as the last resort if the second option fails

>but replace withTimeout

Wouldn't it be enough to keep `withTimeout` with `invokeOnCancellation(onCancelling = true) { if (isIsATimeoutWeExpecting) dumpCoroutines() } `?

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,46 +278,126 @@ public fun runTest(
...
     testBody: suspend TestScope.() -> Unit
 ): TestResult = asSpecificImplementation().let {
     it.enter()
     createTestResult {
-        runTestCoroutine(it, dispatchTimeout, TestScopeImpl::tryGetCompletionCause, testBody) {
+        /** TODO: moving this [AbstractCoroutine.start] call outside [createTestResult] fails on JS. */
+        it.start(CoroutineStart.UNDISPATCHED, it) {
+            testBody()
+        }
+        /**
```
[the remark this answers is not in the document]
dkhalanskyjb: I don't think we should be concerned about the performance of failing tests unless we're talking orders of magnitude, as failing tests are already outliers. Tests that complete before the first suspension are much more likely: for example, unit tests of the form `runTest { something.viewModelScope.launch { ... }; assertSomething() }` finish without suspensions.

We should check the performance impact of the code you're proposing on some actual code base. I could easily believe it if someone told me that the gain from this is dwarfed by just the overhead of starting a new test. A 20% improvement in total runtime for a common class of unit tests would be great, but if the improvement is something like 0.5%, I wouldn't want to introduce such code. If you want, I will elaborate on the many reasons, but the gist is that it's one more edge case to think about.

kotlinx-coroutines-test/common/src/TestBuilders.kt
dkhalanskyjb: `invokeOnCompletion`? This is not quite enough, because to support cancellation when the test body doesn't suspend and doesn't release the thread, we need to cancel the test coroutine asynchronously. I expanded on the solution you proposed and added a test that shows this problem and tests the thread dumps.

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,46 +278,129 @@ public fun runTest(
...
-        runTestCoroutine(it, dispatchTimeout, TestScopeImpl::tryGetCompletionCause, testBody) {
+        /** TODO: moving this [AbstractCoroutine.start] call outside [createTestResult] fails on JS. */
+        scope.start(CoroutineStart.UNDISPATCHED, scope) {
+            yield()
+            testBody()
+        }
+        var timeoutError: Throwable? = null
+        var cancellationException: CancellationException? = null
+        try {
+            withTimeout(timeout) {
```
dkhalanskyjb, line 317: Is `withTimeout` guaranteed to only throw after the `invokeOnCompletion` handler has finished its work?

qwwdfsad: Yes, all intermediate handlers are invoked synchronously, thus making it impossible for coroutine to finalize its state (-> throw or return to the caller)

kotlinx-coroutines-test/common/src/TestBuilders.kt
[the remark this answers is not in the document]
qwwdfsad: The reasoning works for me, thanks for the explanation.

>but the gist is that it's one more edge case to think about.

I've noticed that all the machinery that follows `testBody()` does [if the test is completed] nothing except for creating system-level jobs/coroutines and immediately completing them. 
I doubt there will be any _significant_ improvement (because the condition covers a really specific subset of tests: failed, not suspended, and `assume`-style ones), and was tempted to think about it as "minor improvement that costs nothing to add". Your concern about this condition being harder to reason about is enough for me to discard the idea.

qwwdfsad (approved):
Great job!

kotlinx-coroutines-test/common/src/TestBuilders.kt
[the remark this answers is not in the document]
dkhalanskyjb: Well, the change is not applicable with the latest revision anyway, as the test body never completes in the undispatched invocation, so not even relevant anymore.

kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -269,46 +277,138 @@ public fun runTest(
...
     dispatchTimeoutMs: Long,
     testBody: suspend TestScope.() -> Unit
-): TestResult = runTest(
-    context = context,
-    dispatchTimeout = dispatchTimeoutMs.milliseconds,
-    testBody = testBody
-)
+): TestResult {
+    if (context[RunningInRunTest] != null)
+        throw IllegalStateException("Calls to `runTest` can't be nested. Please read the docs on `TestResult` for details.")
```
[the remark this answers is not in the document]
dkhalanskyjb: Not really. I think I wrote it this way to be able to place a breakpoint on `throw`, but one can do it with `check` also:
```kotlin
check(...) {
  "message" // breakpoint here
}
```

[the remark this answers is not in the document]
qwwdfsad: IDEA allows you to place a breakpoint into a specific part of such expressions:
image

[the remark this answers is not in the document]
dkhalanskyjb: In theory, yes, I just struggle to make it work sometimes. Line breakpoints are much more reliable.

PR #3607, opened 2023-01-30
Update the code example for Dispatchers.Unconfined

### LINKED ISSUE #3605: Kdoc for Dispatchers.Unconfined is possibly incorrect
**Describe the bug**
The kdoc for `Dispatchers.Unconfined` states the following:
```kotlin
/**
 * [...]
 *
 * ### Event loop
 * Event loop semantics is a purely internal concept and have no guarantees on the order of execution
 * except that all queued coroutines will be executed on the current thread in the lexical scope of the outermost
 * unconfined coroutine.
 *
 * For example, the following code:
 * ```
 * withContext(Dispatchers.Unconfined) {
 *    println(1)
 *    withContext(Dispatchers.Unconfined) { // Nested unconfined
 *        println(2)
 *    }
 *    println(3)
 * }
 * println("Done")
 * ```
 * Can print both "1 2 3" and "1 3 2", this is an implementation detail that can be changed.
 * But it is guaranteed that "Done" will be printed only when both `withContext` are completed.
 */
```

I believe this incorrect and you are in fact guaranteed that it will print "1 2 3". This is because `withContext` actually produces a result so something like this would then be broken:
```kotlin
  val result = withContext(Dispatchers.Unconfined) { 1 + 1}
  println(result)
}
```

Instead I believe the author meant to use `launch` inside of the outer `withContext` which would indeed have an undefined behavior.

**Provide a Reproducer**
I can obviously not prove this but running the snippet from the documentation does produce "1 2 3" as it should.

ansman:
The previous example used `withContext` which is guaranteed to complete before the code after it runs. The updated example uses `launch` which I think was originally intended.

This fixes #3605 and [KT-47032](https://youtrack.jetbrains.com/issue/KT-47032/Incorrect-code-example-for-Dispatchers.Unconfined-in-kotlinx.coroutines-docs)

### DIFF
--- a/kotlinx-coroutines-core/common/src/Dispatchers.common.kt
+++ b/kotlinx-coroutines-core/common/src/Dispatchers.common.kt
@@ -56,7 +56,7 @@ public expect object Dispatchers {
      * ```
      * withContext(Dispatchers.Unconfined) {
      *    println(1)
-     *    withContext(Dispatchers.Unconfined) { // Nested unconfined
+     *    launch(Dispatchers.Unconfined) { // Nested unconfined
      *        println(2)
      *    }
      *    println(3)


dkhalanskyjb (approved):
Thanks!

Issue #3608, opened 2023-01-30
FlowAsPublisher with 1 element doesn't complete after 1 is requested. 


rawilder:
**Describe the bug**

FlowAsPublisher with exactly 1 element suspends and never completes when 1 item is requested.

**Provide a Reproducer**

```kotlin
flow { emit("string1") }
	.asPublisher().subscribe(subscriber)

flow { emit("string2") }
	.onCompletion { println("done") }
	.collect {
	    println(it)
	}

val subscriber = object : org.reactivestreams.Subscriber<String> {
    var subscription: org.reactivestreams.Subscription? = null
    override fun onSubscribe(s: org.reactivestreams.Subscription?) {
        subscription = s
        subscription?.request(1)
    }

    override fun onNext(t: String?) {
        println(t)
    }

    override fun onError(t: Throwable?) {
        t?.printStackTrace()
    }

    override fun onComplete() {
        println("done")
    }
}
```

Actual
```
string1
string2
done
```

Expected
```
string1
done
string2
done
```

Maybe it is an irreconcilable difference between the patterns. `flow { emit("string1") }` is a publisher of one item that should complete after emitting the one item. The implementation of the FlowSubscription is suspending after collecting 1 item for a request of 1: https://github.com/Kotlin/kotlinx.coroutines/blob/1.6.4/reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt#L240.

It feels wrong that to complete a FlowAsPublisher of exactly 1 element you must `request(2)` or `request(1)` twice.

rawilder:
Suggested solution, check for suspension before onNext, that way the collect block will end if the element is the last one. This puts the collect block into a slightly weird state, as it is technically collecting the n+1 element even though it was not requested, but if it hangs forever regardless it may not matter. 

- move suspension check above `onNext`
- change `decrementAndGet` to `getAndDecrement` so it gets the last iteration's value

```kotlin
    private suspend fun consumeFlow() {
        flow.collect { value ->
             // Suspend if needed before emitting the next value
            if (requested.getAndDecrement() <= 0) {
                suspendCancellableCoroutine<Unit> {
                    producer.value = it
                }
            } else {
                // check for cancellation if we don't suspend
                coroutineContext.ensureActive()
            }
            // Emit the value
            subscriber.onNext(value)
        }
    }
```

dkhalanskyjb:
[...]

PR #3609, opened 2023-01-30
Allow FlowAsPublisher to complete if collecting last element

rawilder:
These are the changes as I proposed in https://github.com/Kotlin/kotlinx.coroutines/issues/3608

By moving the suspension check above the `onNext` call, we allow the collect block to complete if it is the last element in the collection. I left `ensureActive` at the bottom to so that `testCancellationIsNotReported` was passing. 

I added a test `testSingleRequestOnSingleElementPublisherCompletes` which has one element, subscribes to receive exactly one element, and expects it completes. 

~Leaving as draft while I verify non FlowAsPublisher tests.~ all `jvmTest` passed

### DIFF
--- a/reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt
+++ b/reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt
@@ -234,17 +234,16 @@ public class FlowSubscription<T>(
      */
     private suspend fun consumeFlow() {
         flow.collect { value ->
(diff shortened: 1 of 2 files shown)


rawilder:
Converting back to draft while I review a potential TCK test implementation for FlowAsPublisher

rawilder:
See discussion in issue for reason for close: https://github.com/Kotlin/kotlinx.coroutines/issues/3608

PR #3615, opened 2023-02-01
Update the wording for Dispatchers.Unconfined

ansman:
In my recent PR (#3607) I changed the example for `Dispatchers.Unconfined` but the rest of the documentation still mentions two `withContext` which isn't correct any longer.

### DIFF
--- a/kotlinx-coroutines-core/common/src/Dispatchers.common.kt
+++ b/kotlinx-coroutines-core/common/src/Dispatchers.common.kt
@@ -64,7 +64,7 @@ public expect object Dispatchers {
      * println("Done")
      * ```
      * Can print both "1 2 3" and "1 3 2". This is an implementation detail that can be changed.
-     * However, it is guaranteed that "Done" will be printed only when both `withContext` calls are completed.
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -64,7 +64,7 @@ public expect object Dispatchers {
      * println("Done")
      * ```
      * Can print both "1 2 3" and "1 3 2". This is an implementation detail that can be changed.
-     * However, it is guaranteed that "Done" will be printed only when both `withContext` calls are completed.
+     * However, it is guaranteed that "Done" will only be printed once the `withContext` call is completed.
```
dkhalanskyjb, line 67:
```suggestion
     * However, it is guaranteed that "Done" will only be printed once both the `withContext` call and the inner `launch` are completed.
```
I think this reflects the point better: the order is not guaranteed, but `Dispatchers.Unconfined` will finish doing its thing by the time you leave the outermost invocation.

PR #3617, opened 2023-02-03
TestTimeSource: Expose comparable timemarks

hfhbd:
Kotlin 1.8 added `ComparableTimeMarks` and `AbstractLongTimeSource` implements this interface, so it could be exposed too.


dkhalanskyjb (approved):
Thanks! Added the minus operation to the test and merged manually in c176901a7d5b4a962747ed829ab42cb71134f2ad.

PR #3624, opened 2023-02-13
Remove dead-code in the LockFreeLinkedList implementation after new c…

qwwdfsad:
…hannels implementation

* Slightly speedup the build without unnecessary stress-tests
* Get rid of dead-code that was used only by channels, so it's easier to reason about JobSupport implementation
* Slightly cut dex size in the face of new channels regression as LFLL is no longer shared between JobSupport and Channel


qwwdfsad:
I've deliberately avoided adjusting the comments to LFLL as it's doomed to be removed and definitely won't be modified

dkhalanskyjb (approved):

PR #3631, opened 2023-02-20
Documentation: fix OutOfMemoryError for the runnable samples 

nikpachoo:
See [KT-51515](https://youtrack.jetbrains.com/issue/KT-51515) for the details.

### DIFF
--- a/docs/topics/cancellation-and-timeouts.md
+++ b/docs/topics/cancellation-and-timeouts.md
@@ -381,7 +381,7 @@ resource inside the block that needs closing or release outside of the block.
 
 For example, here we imitate a closeable resource with the `Resource` class that simply keeps track of how many times 
 it was created by incrementing the `acquired` counter and decrementing the counter in its `close` function.
-Now let us let us create a lot of coroutines, each of which creates a `Resource` at the end of the `withTimeout` block
+Now let us create a lot of coroutines, each of which creates a `Resource` at the end of the `withTimeout` block
 and releases the resource outside the block. We add a small delay so that it is more likely that the timeout occurs
 right when the `withTimeout` block is already finished, which will cause a resource leak.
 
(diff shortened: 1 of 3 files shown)


docs/topics/coroutines-basics.md
```diff
@@ -250,14 +250,14 @@ Done
...
-100000 distinct coroutines that each wait 5 seconds and then print a period
+10000 distinct coroutines that each wait 5 seconds and then print a period
 ('.') while consuming very little memory:
 
 ```kotlin
 import kotlinx.coroutines.*
 
 fun main() = runBlocking {
-    repeat(100_000) { // launch a lot of coroutines
+    repeat(10_000) { // launch a lot of coroutines
```
dkhalanskyjb, line 260: This example is non-runnable, so there's no issue with `100_000` here. In this case, we intentionally highlight the difference from threads, and it's more important to have a whooping 100_000 coroutines here than to be able to use https://play.kotlinlang.org/.

dkhalanskyjb (approved):
Thanks!

docs/topics/coroutines-basics.md
AaronFregeau: Sorry, I'm not sure I agree here. Should we at least change the wording, or warn that the playground doesn't handle this? 

The reason I filed the bug [KT-51515](https://youtrack.jetbrains.com/issue/KT-51515) over a year ago is because I was amazed when I saw this example, and immediately popped into the playground to see it in action. But of course it crashed and it leaves a bad impression about trusting the documentation.

nikpachoo: > Should we at least change the wording, or warn that the playground doesn't handle this?

I agree, we should at least add a warning. @danil-pavlov, could you help us with that?

PR #3634, opened 2023-02-21
Release reusability token when suspendCancellableCoroutineReusable's …

### LINKED ISSUE #3613: Infinite loop in spin-lock implementation if CancellableContunuation reusability
Cherry-picked internal report, so not much details are here.
I'll provide a detailed explanation along with test and a fix in a follow-up PR.

TL;DR it is caused by an exceptional path (i.e. `Mutex.lock(owner)`) along with acquiring reusability token

qwwdfsad:
…block throws an exception

Otherwise, continuation instance is left in REUSABLE_CLAIMED state that asynchronous resumer awaits in an infinite spin-loop, potentially causing deadlock with 100% CPU consumption. Originally, the bug was reproduced on old (pre-#3020) implementation where this very pattern was encountered: it was possible to fail owner's invariant check right in the supplied 'block'. This is no longer the case, so the situation is emulated manually (but still is possible in production environments, e.g. when OOM is thrown). Also, suspendCancellableCoroutineReusable is removed from obsolete BroadcastChannel implementation.

Fixes #3613

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
@@ -245,8 +245,6 @@ internal open class BufferedChannel<E>(
      * In case of coroutine cancellation, the element may be undelivered --
      * the [onUndeliveredElement] feature is unsupported in this implementation.
      *
-     * Note that this implementation always invokes [suspendCancellableCoroutineReusable],
-     * as we do not care about broadcasts performance -- they are already deprecated.
      */
     internal open suspend fun sendBroadcast(element: E): Boolean = suspendCancellableCoroutine { cont ->
         check(onUndeliveredElement == null) {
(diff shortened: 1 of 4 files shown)


qwwdfsad:
Also filed https://github.com/Kotlin/kotlinx.coroutines/issues/3635 to eventually address the issue in a more systematic and graceful manner

kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
```diff
@@ -248,7 +248,7 @@ internal open class BufferedChannel<E>(
      * Note that this implementation always invokes [suspendCancellableCoroutineReusable],
      * as we do not care about broadcasts performance -- they are already deprecated.
```
dkhalanskyjb, line 249:
```suggestion
```

dkhalanskyjb: Also, what did this comment mean? `suspendCancellableCoroutineReusable` is a *slower* internal version of `suspendCancellableCoroutine` that we use internally as a performance de-optimization, and now that `BroadcastChannel` is deprecated, we're free to make it fast, right?

dkhalanskyjb (approved):

kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
qwwdfsad: I believe this is just a miswording, asked Nikita to verify just in case.

qwwdfsad: The idea was "it doesn't have non-suspending fast-path"

PR #3641, opened 2023-02-25
Implement internal API for running current event loop

### LINKED ISSUE #3439: Internal API to retreive a single CPU-bound task from Dispatchers.Default

qwwdfsad:
Fixes #3439


kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -138,23 +155,41 @@ internal class WorkQueue {
...
+    }
+
+    private fun pollWithMode(/* Only blocking OR only CPU */ onlyBlocking: Boolean): Task? {
         val start = consumerIndex.value
         var end = producerIndex.value
-
-        while (start != end && blockingTasksInBuffer.value > 0) {
-            val task = tryExtractBlockingTask(--end)
+        // CPU or (BLOCKING & hasBlocking)
+        val shouldProceed = !onlyBlocking || blockingTasksInBuffer.value > 0
```
qwwdfsad, line 177: `shouldProceed` isn't updated deliberately, I find making it as part of `while` a bit more compact

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -107,25 +112,37 @@ internal class WorkQueue {
      *
      * Returns [NOTHING_TO_STEAL] if queue has nothing to steal, [TASK_STOLEN] if at least task was stolen
      * or positive value of how many nanoseconds should pass until the head of this queue will be available to steal.
+     *
+     * [StealingMode] controls what tasks to steal:
+     * * [STEAL_ANY] is default mode for scheduler, task from the head (in FIFO order) is stolen
+     * * [STEAL_BLOCKING_ONLY] is mode for stealing *an arbitrary* blocking task which is used by scheduler when helping in Dispatchers.IO mode
```
dkhalanskyjb, line 118:
```suggestion
     * * [STEAL_BLOCKING_ONLY] is mode for stealing *an arbitrary* blocking task, which is used by the scheduler when helping in Dispatchers.IO mode
```
Otherwise, it means that the task is used by the scheduler.

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -16,6 +16,11 @@ internal const val MASK = BUFFER_CAPACITY - 1 // 128 by default
 internal const val TASK_STOLEN = -1L
 internal const val NOTHING_TO_STEAL = -2L
 
+internal typealias StealingMode = Int
+internal const val STEAL_ANY: StealingMode = -1
+internal const val STEAL_CPU_ONLY: StealingMode = 0
+internal const val STEAL_BLOCKING_ONLY: StealingMode = 1
```
dkhalanskyjb, line 22: Potential simplification, up to you whether to use it:

```suggestion
internal const val STEAL_ANY: StealingMode = 3
internal const val STEAL_CPU_ONLY: StealingMode = 2
internal const val STEAL_BLOCKING_ONLY: StealingMode = 1
```
This could simplify the code a bit with bitmasks.

```kotlin
// maybe inline
val Task.mode: Int
  get() = if (isBlocking) STEAL_BLOCKING_ONLY else STEAL_CPU_ONLY
```
Then,
```kotlin
            if (lastScheduled.isBlocking) {
                if (stealingMode == STEAL_CPU_ONLY) {
                    return NOTHING_TO_STEAL
                }
            } else if (stealingMode == STEAL_BLOCKING_ONLY) {
                return NOTHING_TO_STEAL
            }
```
becomes
```kotlin
if ((lastScheduled.mode and stealingMode) == 0)
  return NOTHING_TO_STEAL
```

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -138,23 +155,41 @@ internal class WorkQueue {
...
+    }
+
+    fun pollCpu(): Task? {
+        while (true) { // Poll the slot
+            val lastScheduled = lastScheduledTask.value ?: break
+            if (lastScheduled.isBlocking) break
+            if (lastScheduledTask.compareAndSet(lastScheduled, null)) {
+                return lastScheduled
+            } // Failed -> someone else stole it
+        }
```
dkhalanskyjb, line 168: Why not move this code to `pollWithMode` as well? Only one line would need to change to
```kotlin
            if (lastScheduled.isBlocking == onlyBlocking) break
```

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -138,23 +155,41 @@ internal class WorkQueue {
...
             if (task != null) {
                 return task
             }
         }
         return null
     }
 
-    private fun tryExtractBlockingTask(index: Int): Task? {
+    private fun tryExtractFromTheMiddle(index: Int, onlyBlocking: Boolean): Task? {
+        if (onlyBlocking && blockingTasksInBuffer.value == 0) return null
```
dkhalanskyjb, line 188: I don't get it. Why not do this outside the function, in the caller? Then, we could leave the loop immediately (like it was done before the change), and `shouldProceed` wouldn't be needed.

kotlinx-coroutines-core/jvm/src/EventLoop.kt
```diff
@@ -45,6 +49,80 @@ internal actual fun createEventLoop(): EventLoop = BlockingEventLoop(Thread.curr
...
+ * thus blocking at least one thread. To avoid deadlocks and starvation during large hierarchical decompositions,
+ * 'runDecomposedTaskAndJoinIt' should not only block, but also **help** executing the task or other tasks
+ * until an arbitrary condition is satisfied.
+ *
+ * See #3439 for additional details.
+ *
+ * ### Limitations and caveats
+ *
+ * - Executes tasks in-place, thus potentially leaking irrelevant thread-locals from the current thread
+ * - Is not 100% effective, because the caller should somehow "wait" (or do other work) during [Long] returned nanoseconds
```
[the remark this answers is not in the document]
qwwdfsad: Right, it was my main concern while discussing this API.

I wish we had something like `FJP.awaitJoin`, but it doesn't really seem doable in a reasonable timeframe

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
qwwdfsad: Good suggestion, applied

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
```diff
@@ -107,54 +115,75 @@ internal class WorkQueue {
...
         }
 
+        // Failed to poll the slot, scan the queue
         val start = consumerIndex.value
         var end = producerIndex.value
-
-        while (start != end && blockingTasksInBuffer.value > 0) {
-            val task = tryExtractBlockingTask(--end)
+        // Bail out if there is no blocking work for us
+        if (onlyBlocking && blockingTasksInBuffer.value == 0) return null
```
dkhalanskyjb, line 172: It is still a mystery to me why we don't do it on every iteration any longer. Could you drop a comment in the code?

dkhalanskyjb: This particular place still has the change.

dkhalanskyjb: Also, I would be okay with it if it had a corresponding comment and, preferably, was in a separate commit (this one is a bit loaded with changes already).

dkhalanskyjb (approved):

kotlinx-coroutines-core/jvm/src/scheduling/WorkQueue.kt
qwwdfsad: Oh, I probably have to stop pushing any changes around Friday's evenings :sweat_smile:

dkhalanskyjb: Alternatively, you could embrace the flow, go for it, and publish a release today.

PR #3647, opened 2023-02-27
Deprecate BroadcastChannel and the corresponding API

### LINKED ISSUE #2680: Consider keeping BroadcastChannel in a deprecated state for longer
I have just read the news that Broadcast Channel is going to be completely removed with next release.
Why? I have build my [unexBot](https://github.com/Artraxon/unexBot) and [tihibot](https://github.com/Artraxon/tihibot) using the Channel API **because** it was simpler and less prone the change. The Flow API was and is still constantly changing and I simply do not need the added functionality.

I understand that this is an experimental API, but If you leave an "experimental" API out there for 4 years, It should be expected that people are going to use it. And two major versions (effectively 2 years), of which one is also unmature, so 1 year in the end, for swapping out ALL Channel usage is completely reckless. What is really the point in removing it? Mark it as deprecated and refer to the alternatives and then leave it at that for some years, it does not do any harm, only removing it does.

The people that are building many of the open-source libraries for kotlin are not working full-time on there open-source projects. I'm not. I already started refactoring these bots to move the Framework that I've written into a separate library, but now I don't see the point because it's going to be broken anyway in a year and I do not have the time to fix it.

qwwdfsad:
Fixes #2680

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
@@ -3,7 +3,7 @@
  */
 @file:JvmMultifileClass
 @file:JvmName("ChannelsKt")
-@file:Suppress("DEPRECATION_ERROR")
+@file:Suppress("DEPRECATION_ERROR", "DEPRECATION")
 @file:OptIn(ExperimentalContracts::class)
 
 package kotlinx.coroutines.channels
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -3,7 +3,7 @@
  */
 @file:JvmMultifileClass
 @file:JvmName("ChannelsKt")
-@file:Suppress("DEPRECATION_ERROR")
+@file:Suppress("DEPRECATION_ERROR", "DEPRECATION")
```
dkhalanskyjb, line 6: I don't like this `Suppress`, it could obscure some real issues eventually. I checked, and the compiler seems pleased if we just add the `Suppress` to each deprecated method separately: https://gist.github.com/dkhalanskyjb/de271eff5aa3286fd680769e080696f8

dkhalanskyjb (approved):

PR #3649, opened 2023-02-28
KT-51515 cleanup

elizarov:
See https://youtrack.jetbrains.com/issue/KT-51515
Small fixes to clean up #3631

* Fix comment about 100k corouties in cancellation and timeouts section (it is now 10k!)
* Make "Coroutines are light-weight" example runnable by reducing the number of coroutines to 50k, also change the wording about what happens with threads


dkhalanskyjb (approved):

danil-pavlov (changes requested):
I suggest not making it a note because in this case, you'll get two notes one after another and visually overload the text

danil-pavlov (approved):

PR #3650, opened 2023-02-28
chore: technical limits of the playground

danil-pavlov:


dkhalanskyjb:
See https://github.com/Kotlin/kotlinx.coroutines/pull/3649, looks like a better solution.

danil-pavlov:
@dkhalanskyjb I was following [this discussion](https://github.com/Kotlin/kotlinx.coroutines/pull/3631#discussion_r1112873231), but if you decided to go with 50k and the playground can handle that, let's just close my PR

PR #3663, opened 2023-03-06
Fix CLI property ignored in version resolution

Badya:
Cherry pick from `kotlin-community/dev`


qwwdfsad:
Could you please elaborate on what it is needed for?

Badya:
> Could you please elaborate on what it is needed for?

Without this change, when we pass gradle property via CLI, e.g. `kotlin_version` for K2 tests on CI, it will ignore the CLI property because it reads `gradle.properties` directly from the file instead of using a gradle property resolution mechanism.

qwwdfsad (approved):

PR #3671, opened 2023-03-10
Update JavaFx to 17.0.2 in order to support M1

qwwdfsad:


dkhalanskyjb (approved):
Sure, don't see why not.

PR #3672, opened 2023-03-10
Fix `limitedParallelism` doing dispatches when it has no tasks

dkhalanskyjb:
To support fairness w.r.t. the tasks that want to run on the wrapped dispatcher outside of its view, `LimitedDispatcher` redispatches the work loop periodically. This can happen even when there are no more tasks in `LimitedDispatcher`'s queue. Typically, this behavior is completely benign: after being redistpatched, the new runner just checks the queue, sees that there is nothing there, and leaves. However, with closable dispatchers, this affects correctness, as this redispatch may happen even after all the dispatched tasks were run and the dispatcher was closed.

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
@@ -7,7 +7,6 @@ package kotlinx.coroutines.internal
 import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import kotlin.coroutines.*
-import kotlin.jvm.*
 
 /**
  * The result of .limitedParallelism(x) call, a dispatcher
@@ -27,7 +26,7 @@ import kotlin.jvm.*
 internal class LimitedDispatcher(
     private val dispatcher: CoroutineDispatcher,
     private val parallelism: Int
-) : CoroutineDispatcher(), Runnable, Delay by (dispatcher as? Delay ?: DefaultDelay) {
+) : CoroutineDispatcher(), Delay by (dispatcher as? Delay ?: DefaultDelay) {
 
     // Atomic is necessary here for the sake of K/N memory ordering,
     // there is no need in atomic operations for this property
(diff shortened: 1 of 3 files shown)


dkhalanskyjb:
Testing this is difficult. The problematic behavior only happens when there are no more tasks to run when the fairness dispatch wants to proceed, but before that, the worker had a steady supply of tasks and did not get fired. With a hardcoded `16` this could be done, but I don't think it's worth the effort. As for stress tests, for me, it takes several minutes to reproduce the problem.

kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
```diff
@@ -45,61 +44,37 @@ internal class LimitedDispatcher(
...
-                    return
-                }
-                continue
-            }
-
-            synchronized(workerAllocationLock) {
-                runningWorkers.decrementAndGet()
-                if (queue.size == 0) return
-                runningWorkers.incrementAndGet()
-                fairnessCounter = 0
```
dkhalanskyjb, line 72: The behavior of this line of code was lost in the process. It can be restored but this wouldn't look as nice. Is it right that the gains here are marginal?

kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
```diff
@@ -45,61 +44,37 @@ internal class LimitedDispatcher(
...
-        while (true) {
-            val task = queue.removeFirstOrNull()
-            if (task != null) {
-                try {
-                    task.run()
-                } catch (e: Throwable) {
-                    handleCoroutineException(EmptyCoroutineContext, e)
-                }
-                // 16 is our out-of-thin-air constant to emulate fairness. Used in JS dispatchers as well
-                if (++fairnessCounter >= 16 && dispatcher.isDispatchNeeded(this)) {
```
qwwdfsad, line 59: Is it correct that this PR addresses the following issue in this snippet:

* `run` executes task (`:54`)
* Underlying dispatcher is closed
* `run` finishes, unluckily enough overflows `fairnessCounter`, goes to redispatch and crashes?

dkhalanskyjb: That's one aspect of it, yes. Consider also the case where an external `dispatch` or `dispatchYield` happened (for example, from some callback), scheduled a task, waited for a millennia, during which this task and every other one got executed and the dispatcher got closed, and then proceeded to allocate a worker and run it.

qwwdfsad (approved):
Great job with figuring that one out!

qwwdfsad (approved):

PR #3674, opened 2023-03-12
Include runBlocking{} in the context sample code

toboche:
The text in the "Dispatchers and threads﻿" and "Unconfined vs confined dispatcher" sections explicitly mentions the example code having  `runBlocking` call somewhere but the sample code does not show the call. I'd propose to include the call to  `runBlocking` so that the readers don't have to guess where the actual  `runBlocking` call is done.


dkhalanskyjb (changes requested):
Sure, this makes sense. There's also the snippet with `val a = async` that mentions `runBlocking` but doesn't include one.

PR #3675, opened 2023-03-13
Get rid of 'previous-compilation-data.bin' in META-INF

qwwdfsad:


integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
```diff
@@ -0,0 +1,70 @@
...
+        )
+    }
+
+    private fun JarFile.checkMetaInfStructure(expected: Set<String>) {
+        val actual = HashSet<String>()
+        for (e in entries()) {
+            if (e.isDirectory() || !e.realName.contains("META-INF")) {
+                continue
+            }
+            val partialName = e.realName.substringAfter("META-INF/")
```
dkhalanskyjb, line 58: Historical note: the path separators here are always forward slashes, as required by the zip file spec (https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT, 4.4.17.1)

integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
```
dkhalanskyjb, line 2: Sadly, I don't know what would be right here, but I'm sure it's not this. My bet is "Copyright 2023".

dkhalanskyjb (approved):

Issue #3677, opened 2023-03-13
Decide on each particular machinery of debug mode and provide an additional feature flag for each


qwwdfsad:
Currently, when coroutines debug mode is enabled, the following machinery is enabled:

* Every coroutine created is supplied with `CoroutineId`. This one might be useful for local debugging and logging, though its general applicability (i.e. for running applications of IDEA scale, the purpose of numeric id is unclear)
* Stacktrace population for `JobCancellationException` is enabled, which is known to be a performance-hitter (we also know it from IDEA experience)
* `Thread.setName` with coroutine id is invoked. This seemed to be the legacy of earlier days when we had little to no tooling around coroutines debugging. It has three downsides: it crashes ART consistently (#2234), it might dominate coroutine-specific workloads (https://github.com/Kotlin/kotlinx.coroutines/issues/2234#issuecomment-1204922620), and it potentially interferes with code that leverages thread names.
* Stacktrace recovery is enabled. I like this one and have no intention of tweaking it :)

qwwdfsad:
My very first proposal is to completely get rid of `setName` -- it seems to cause more harm than good.

If you have any additional input about `CancellationException` stacktraces (or their absence) or `CoroutineId` -- please feel free to share, whether you are actively using them or, contrary, see no value in it.

qwwdfsad:
UDP: Fleet disables stacktrace recovery because of `setName` overhead

Issue #3611, opened 2023-01-31
Feature request: allow stacktrace recovery in production


ZacSweers:
**Use case**

I know we could do a round-about way via excluding the embedded proguard rules that come with coroutines, but that’s fairly cumbersome and I wonder if this could be done a separate way. The reason I ask is because we want traces from production and our bug reporting tool can retrace these (we do the same with RxDogTag, of which I’m the author). This has been immensely beneficial to us in tracking and fixing bugs that happen in production, as otherwise the stacktraces sent back are unustable.

I recognize this comes at a performance cost and understand why coroutines defaults to not allowing this with its proguard rules, but it is also currently making this decision for consumers and I think it would be valuable to be able to enable it.

**The Shape of the API**

I don't have a clear idea of how it would work with coroutines, but in RxDogTag we basically gate installation/init of its capturing behavior via global boolean flag and don't automatically remove its machinery via embedded proguard rules.

**Prior Art**

https://github.com/uber/rxdogtag

dkhalanskyjb:
Duplicate of #1998.

PR #3678, opened 2023-03-13
Get rid of Thread.name augmentation with CoroutineId when debug mode …

### LINKED ISSUE #2234: Android ART Native Crash sometimes occurs when changing a threads name
```

*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 0 >>> com.- <<<

backtrace:
  #00  pc 00000000000830f0  /apex/com.android.runtime/lib64/bionic/libc.so (abort+160)
  #00  pc 00000000004b977c  /apex/com.android.runtime/lib64/libart.so (art::Runtime::Abort(char const*)+2280)
  #00  pc 000000000000b458  /system/lib64/libbase.so (android::base::LogMessage::~LogMessage()+580)
  #00  pc 00000000005108a0  /apex/com.android.runtime/lib64/libart.so (art::ThreadSuspendByPeerWarning(art::Thread*, android::base::LogSeverity, char const*, _jobject*)+656)
  #00  pc 000000000050fa80  /apex/com.android.runtime/lib64/libart.so (art::ThreadList::SuspendThreadByPeer(_jobject*, bool, art::SuspendReason, bool*)+1120)
  #00  pc 000000000042fa88  /apex/com.android.runtime/lib64/libart.so (art::Thread_setNativeName(_JNIEnv*, _jobject*, _jstring*)+1076)
  #00  pc 00000000000ba7e8  /system/framework/arm64/boot.oat (art_jni_trampoline+152)
  #00  pc 00000000001a5468  /system/framework/arm64/boot.oat (java.lang.Thread.setName+120)
  #00  pc 00000000002ff26c  /data/app/com.-==/oat/arm64/base.odex (kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.setIndexInArray+284)
  #00  pc 0000000000137334  /apex/com.android.runtime/lib64/libart.so (art_quick_invoke_stub+548)
  #00  pc 0000000000145fec  /apex/com.android.runtime/lib64/libart.so (art::ArtMethod::Invoke(art::Thread*, unsigned int*, unsigned int, art::JValue*, char const*)+244)
  #00  pc 00000000002e3948  /apex/com.android.runtime/lib64/libart.so (art::interpreter::ArtInterpreterToCompiledCodeBridge(art::Thread*, art::ArtMethod*, art::ShadowFrame*, unsigned short, art::JValue*)+384)
  #00  pc 00000000002deba8  /apex/com.android.runtime/lib64/libart.so (bool art::interpreter::DoCall<false, false>(art::ArtMethod*, art::Thread*, art::ShadowFrame&, art::Instruction const*, unsigned short, art::JValue*)+892)
  #00  pc 00000000005a728c  /apex/com.android.runtime/lib64/libart.so (MterpInvokeVirtualQuick+616)
  #00  pc 0000000000135594  /apex/com.android.runtime/lib64/libart.so (mterp_op_invoke_virtual_quick+20)
  #00  pc 00000000017d7a5e  /data/app/com.-==/oat/arm64/base.vdex (kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.tryTerminateWorker+174)
  #00  pc 00000000002b4c5c  /apex/com.android.runtime/lib64/libart.so (art::interpreter::Execute(art::Thread*, art::CodeItemDataAccessor const&, art::ShadowFrame&, art::JValue, bool, bool) (.llvm.4579747295037710581)+240)
  #00  pc 0000000000592688  /apex/com.android.runtime/lib64/libart.so (artQuickToInterpreterBridge+1032)
  #00  pc 0000000000140468  /apex/com.android.runtime/lib64/libart.so (art_quick_to_interpreter_bridge+88)
  #00  pc 00000000002fcd94  /data/app/com.-==/oat/arm64/base.odex (kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.park+180)
  #00  pc 00000000002fe314  /data/app/com.-==/oat/arm64/base.odex (kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.tryPark+628)
  #00  pc 00000000002fd984  /data/app/com-==/oat/arm64/base.odex (kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker+228)
  #00  pc 00000000002ff128  /data/app/com.-==/oat/arm64/base.odex (kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run+40)
  #00  pc 0000000000137334  /apex/com.android.runtime/lib64/libart.so (art_quick_invoke_stub+548)
  #00  pc 0000000000145fec  /apex/com.android.runtime/lib64/libart.so (art::ArtMethod::Invoke(art::Thread*, unsigned int*, unsigned int, art::JValue*, char const*)+244)
  #00  pc 00000000004b103c  /apex/com.android.runtime/lib64/libart.so (art::(anonymous namespace)::InvokeWithArgArray(art::ScopedObjectAccessAlreadyRunnable const&, art::ArtMethod*, art::(anonymous namespace)::ArgArray*, art::JValue*, char const*)+104)
  #00  pc 00000000004b2150  /apex/com.android.runtime/lib64/libart.so (art::InvokeVirtualOrInterfaceWithJValues(art::ScopedObjectAccessAlreadyRunnable const&, _jobject*, _jmethodID*, jvalue const*)+416)
  #00  pc 00000000004f2b18  /apex/com.android.runtime/lib64/libart.so (art::Thread::CreateCallback(void*)+1176)
  #00  pc 00000000000e6890  /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_start(void*)+36)
  #00  pc 0000000000084b6c  /apex/com.android.runtime/lib64/bionic/libc.so (__start_thread+64)
```

Looking at the the art source it seems like this timeout when trying to set the thread name may be the culprit 
https://android.googlesource.com/platform/art/+/master/runtime/thread_list.cc#970

It's not really clear to me whether this timeout would be triggered based on a specific state of the Thread or if it's just a race condition that is sometimes triggered.

It seems like the thread name changing is mostly cosmetic? I wonder if an option to disable or reduce thread name changes might be a quick solution--as much as I would hate giving that up for debug purposes :(

qwwdfsad:
…is enabled

See #3677 and #2234 for the rationale

Fixes #2234

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
+++ b/kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
@@ -284,27 +284,7 @@ private const val DEBUG_THREAD_NAME_SEPARATOR = " @"
 @IgnoreJreRequirement // desugared hashcode implementation
 internal data class CoroutineId(
     val id: Long
-) : ThreadContextElement<String>, AbstractCoroutineContextElement(CoroutineId) {
+) : AbstractCoroutineContextElement(CoroutineId) {
     companion object Key : CoroutineContext.Key<CoroutineId>
     override fun toString(): String = "CoroutineId($id)"
-
-    override fun updateThreadContext(context: CoroutineContext): String {
-        val coroutineName = context[CoroutineName]?.name ?: "coroutine"
-        val currentThread = Thread.currentThread()
-        val oldName = currentThread.name
-        var lastIndex = oldName.lastIndexOf(DEBUG_THREAD_NAME_SEPARATOR)
-        if (lastIndex < 0) lastIndex = oldName.length
-        currentThread.name = buildString(lastIndex + coroutineName.length + 10) {
-            append(oldName.substring(0, lastIndex))
-            append(DEBUG_THREAD_NAME_SEPARATOR)
-            append(coroutineName)
-            append('#')
-            append(id)
-        }
-        return oldName
-    }
-
-    override fun restoreThreadContext(context: CoroutineContext, oldState: String) {
-        Thread.currentThread().name = oldState
-    }
 }
--- a/kotlinx-coroutines-core/jvm/src/Debug.kt
+++ b/kotlinx-coroutines-core/jvm/src/Debug.kt
@@ -17,8 +17,6 @@ import kotlin.internal.InlineOnly
  * ### Debugging facilities
  *
  * In debug mode every coroutine is assigned a unique consecutive identifier.
- * Every thread that executes a coroutine has its name modified to include the name and identifier of
- * the currently running coroutine.
  *
  * Enable debugging facilities with "`kotlinx.coroutines.debug`" ([DEBUG_PROPERTY_NAME]) system property,
  * use the following values:
@@ -32,6 +30,11 @@ import kotlin.internal.InlineOnly
  *
  * Debugging facilities are implemented by [newCoroutineContext][CoroutineScope.newCoroutineContext] function that
  * is used in all coroutine builders to create context of a new coroutine.
+ *
+ * ### Legacy note
+ *
+ * Previously, every thread that executed a coroutine had its name modified to include the name and identifier of the currently running coroutine.
+ * This behaviour was removed as part of #3677 and #2234.
  */
 public const val DEBUG_PROPERTY_NAME: String = "kotlinx.coroutines.debug"
 


qwwdfsad:
@dkhalanskyjb WDYT? Do you have any objections or comments?

This PR is draft because it has a bunch of tests failing, mostly guide-related once where coroutine id is part of the output.
I also deliberately avoid adding a feature flag to revert the behaviour -- that way we'll see if anybody actually ever relied on this behaviour and will be able to make much more weighted decision (instead of the maintenance of such flag nearly forever)

dkhalanskyjb:
* When I google `"coroutinename"`, most of the top results do mention/request naming threads. Plenty of tutorials and internal lore exist around the existing behavior of the debug mode.
* This behavior seems to be also used for debugging: https://grep.app/search?q=Log%5C.d.%2AThread%5C.current&regexp=true&filter[lang][0]=Kotlin
* The way I read the issue about the ART crash, it seems to happen in production but not during debugging, so it should not interfere with the purpose of the debug mode.
* Performance is typically expected to suffer a bit during debugging.

I don't understand why the debug mode is enabled in production, and from the discussion in #2234, our users don't, too. Something somewhere introduces the debug mode behavior, most of which is typically undesirable in production, not just this part. That's the actual root of the issue, the way I see it. As or `Thread.name`, yes, if we look at it as a feature for production builds, it's really problematic, but as debug functionality, I think it could be nice.

elizarov:
One thing worries me here. If we had ideal working coroutine debugging tools, we would not need Thread.name augmentation feature at all. But we don't. In particular, if I run the code with coroutines with under debugger and just press "pause", I don't see any coroutine information in the debugger. Currently, the augmented coroutine name would be my only way to figure out what's going on. Would not it be safe to leave it under a flag? 

2. If I run the code in production with coroutines debugger agent, then I can ask it for the

qwwdfsad:
>I don't understand why the debug mode is enabled in production

I tend to think completely the opposite. We should eliminate as many barriers as possible for folks to keep debug mode enabled (the same way a lot of folks have `-ea` in production) -- performance, non-trivial behaviours, and cumbersome scenarios.

The reason it is not enabled by default is straightforward -- it changes the semantics of the application: thread names, exception identities, slightly slowing down the performance etc. and it is not acceptable as **the** default for a sufficiently large framework.

Yet it would be nice to eliminate as many barriers as possible for an average user to have better diagnostics, observability and additional info as possible; coroutines are already hard to debug, and that's the reason we have requests like #3611 and `DebugProbes` improvements (that are enabled by default in **all** IDEA products).

For me, it seems like changing thread's name is not carrying its own weight and I'm yet to find the use of this behaviour for an actual app in the wild.

Roman has a good observation about the actual debugger though, it would be nice to keep this behaviour unless IDEA's debugger **automatically** augments thread's names in its tooltips. Taking into account current state of IDE team priorities, It's unlikely to happen in an observable timeframe though

dkhalanskyjb:
Debugging is when you dive into code to find a bug. At that point, you disable compiler optimizations (where applicable), enable the (often stripped) debug symbols like function names, and don't care even if your code runs twice as slow as it does in production, as you want all the help you can get. For example, it's not inconceivable to record the complete execution of a program when debugging to obtain a consistent reproducer (https://rr-project.org/), or to instrument memory reads and writes (https://valgrind.org/, https://github.com/google/sanitizers/wiki/AddressSanitizer). In production, however, a ≈50x slowdown is often out of the question.

Clearly, diagnostics are often useful in production, despite their cost. After all, there's the whole sprawling telemetry industry, and its role is not limited to just building advertisement profiles, and telemetry is often heavier than, say, stacktrace recovery. The idea behind https://github.com/Kotlin/kotlinx.coroutines/issues/3677 also seems solid to me: yes, what we have now for debugging can also be useful for diagnostics. However, there's a spectrum between "everything runs as fast as it can, but we get zero visibility into what's going on" and "everything is instrumented, the full logs are stored, and even the slightest anomalies are detected and reported," and my argument is making it feasible to enable debug mode in production, essentially jamming debugging and observing into a single point of that spectrum, is not a worthy goal.

PR #3682, opened 2023-03-14
Stabilize Channel.invokeOnClose

### LINKED ISSUE #3358: Consider stabilizing `Channel.invokeOnClose`
It was added in 0.24 and made experimental in 1.0 later that year. Both of those events were 4 years ago. Time to stabilize?

qwwdfsad:
CompletionHandler is not user deliberately, as its contract requires some additional refinement along with `onCancelling` handler stabilization. Note that replacing functional type with the very same typealias is backwards-compatible in the current state of linkage, so we are not giving up any future opportunities.

Also, fix behavioral mismatch: `CancellationException` is supplied (and always has been) to `invokeOnClose` when a channel was cancelled normally instead of `null` as documentation stated. This behaviour is aligned with other cancellation handlers and also allows handler to distinguish whether the channel was closed or cancelled.

Fixes #3358

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -105,8 +105,17 @@ public interface SendChannel<in E> {
      * If the channel is closed already, the handler is invoked immediately.
      *
      * The meaning of `cause` that is passed to the handler:
-     * * `null` if the channel was closed or cancelled without the corresponding argument
-     * * the cause of `close` or `cancel` otherwise.
+     * - `null` if the channel was closed normally without the corresponding argument.
+     * - Instance of [CancellationException] if the channel was cancelled normally without the corresponding argument.
+     * - The cause of `close` or `cancel` otherwise.
+     *
+     * ### Execution context and exception safety
+     *
+     * The [handler] is executed in the context of closing or cancelling operation
+     * as the very last action of channel's lifecycle, meaning that improper
+     * (e.g. throwing or hanging) implementations of handler cannot interfere
+     * with the channel's final state.
+     * Unhandled exceptions from [handler] are propagated to the operation's caller.
      *
      * Example of usage (exception handling is omitted):
      *
@@ -117,21 +126,21 @@ public interface SendChannel<in E> {
      * }
      *
      * val uiUpdater = launch(Dispatchers.Main, parent = UILifecycle) {
-     *    events.consume {}
-     *    events.cancel()
+     *    events.consume { ... handle events ... }
      * }
-     *
+     * // Stop callback as soon as the channel is closed or cancelled
      * events.invokeOnClose { callbackBasedApi.stop() }
      * ```
      *
-     * **Note: This is an experimental api.** This function may change its semantics, parameters or return type in the future.
+     * **Stability note.** This function constitutes a stable API surface with the only exception
+     * about [IllegalStateException] being thrown when multiple handlers are registered.
+     * There is a possibility that this restriction will be lifted in the future.
      *
-     * @throws UnsupportedOperationException if the underlying channel doesn't support [invokeOnClose].
+     * @throws UnsupportedOperationException if the underlying channel does not support [invokeOnClose].
      * Implementation note: currently, [invokeOnClose] is unsupported only by Rx-like integrations
      *
      * @throws IllegalStateException if another handler was already registered
      */
-    @ExperimentalCoroutinesApi
     public fun invokeOnClose(handler: (cause: Throwable?) -> Unit)
 
     /**
--- a/kotlinx-coroutines-core/common/test/channels/BasicOperationsTest.kt
+++ b/kotlinx-coroutines-core/common/test/channels/BasicOperationsTest.kt
@@ -85,6 +85,45 @@ class BasicOperationsTest : TestBase() {
         }
     }
 
+    @Test
+    fun testCancelledChannelInvokeOnClose() {
+        val ch = Channel<Int>()
+        ch.invokeOnClose { assertIs<CancellationException>(it) }
+        ch.cancel()
+    }
+
+    @Test
+    fun testCancelledChannelWithCauseInvokeOnClose() {
+        val ch = Channel<Int>()
+        ch.invokeOnClose { assertIs<TimeoutCancellationException>(it) }
+        ch.cancel(TimeoutCancellationException(""))
+    }
+
+    @Test
+    fun testThrowingInvokeOnClose() = runTest {
+        val channel = Channel<Int>()
+        channel.invokeOnClose {
+            assertNull(it)
+            expect(3)
+            throw TestException()
+        }
+
+        launch {
+            try {
+                expect(2)
+                channel.close()
+            } catch (e: TestException) {
+                expect(4)
+            }
+        }
+        expect(1)
+        yield()
+        assertTrue(channel.isClosedForReceive)
+        assertTrue(channel.isClosedForSend)
+        assertFalse(channel.close())
+        finish(5)
+    }
+
     @Suppress("ReplaceAssertBooleanWithAssertEquality")
     private suspend fun testReceiveCatching(kind: TestChannelKind) = coroutineScope {
         reset()
@@ -124,7 +163,7 @@ class BasicOperationsTest : TestBase() {
         channel.trySend(2)
             .onSuccess { expectUnreached() }
             .onClosed {
-                assertTrue { it is  ClosedSendChannelException}
+                assertTrue { it is ClosedSendChannelException }
                 if (!kind.isConflated) {
                     assertEquals(42, channel.receive())
                 }


kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -105,8 +105,17 @@ public interface SendChannel<in E> {
      * If the channel is closed already, the handler is invoked immediately.
      *
      * The meaning of `cause` that is passed to the handler:
-     * * `null` if the channel was closed or cancelled without the corresponding argument
-     * * the cause of `close` or `cancel` otherwise.
+     * - `null` if the channel was closed normally without the corresponding argument.
+     * - Instance of [CancellationException] if the channel was cancelled normally without the corresponding argument.
```
qwwdfsad, line 109: "If it isn't tested, it doesn't work" case in point

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -117,21 +126,21 @@ public interface SendChannel<in E> {
      * }
      *
      * val uiUpdater = launch(Dispatchers.Main, parent = UILifecycle) {
-     *    events.consume {}
-     *    events.cancel()
+     *    events.consume { ... handle events ... }
      * }
-     *
+     * // Stop callback as soon as the channel is closed or cancelled
```
dkhalanskyjb, line 131:
```suggestion
     * // Stop the callback as soon as the channel is closed or cancelled
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -117,21 +126,21 @@ public interface SendChannel<in E> {
...
      * events.invokeOnClose { callbackBasedApi.stop() }
      * ```
      *
-     * **Note: This is an experimental api.** This function may change its semantics, parameters or return type in the future.
+     * **Stability note.** This function constitutes a stable API surface with the only exception
+     * about [IllegalStateException] being thrown when multiple handlers are registered.
```
dkhalanskyjb, line 136:
```suggestion
     * **Stability note.** This function constitutes a stable API surface, with the only exception being
     * that an [IllegalStateException] is thrown when multiple handlers are registered.
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -117,21 +126,21 @@ public interface SendChannel<in E> {
...
+     * There is a possibility that this restriction will be lifted in the future.
```
dkhalanskyjb, line 137: Why so formal? I thought we were going for a bit of a hip and easy-to-read feeling with our documentation.
```suggestion
     * This restriction could be lifted in the future.
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -105,8 +105,17 @@ public interface SendChannel<in E> {
...
+     * - `null` if the channel was closed normally without the corresponding argument.
+     * - Instance of [CancellationException] if the channel was cancelled normally without the corresponding argument.
+     * - The cause of `close` or `cancel` otherwise.
+     *
+     * ### Execution context and exception safety
+     *
+     * The [handler] is executed in the context of closing or cancelling operation
+     * as the very last action of channel's lifecycle, meaning that improper
+     * (e.g. throwing or hanging) implementations of handler cannot interfere
+     * with the channel's final state.
```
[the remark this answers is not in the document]
qwwdfsad: >Could you rephrase?

I'm struggling to find a better phrasing that reflects the idea. It seems like you got this right -- `invokeOnClose` is the very last action of the channel's state transitions. The same applies to our `Job` -- `invokeOn*` is invoked strictly _after_ the final state, when a job cannot make any new transitions (e.g. adding a new child, hang etc.).

>Otherwise, the usage example is not generally correct

Could you please elaborate on why? If you mean the scenario "events are being emitted when the channel is already closed", then it's perfectly okay:

* Callback API is _typically_ non-suspending, meaning that the only way to send an element is `trySend` (that forces one to handle exceptions as well)
* If it's suspending, any asynchronous closing/cancelling operation will trigger an exception from `send`, so the same story applied -- exceptions should be handled.
* The very last one -- `close`/`cancel` _typically_ are quite fast operations (e.g. one doesn't have to wait for all "children" to complete), so the actual time window where "redundant" work happens is reasonably small (compare to the scenario when the callback is first stopped and then channel's state is finalized).

There is a small note "exception handling is omitted". Do you think it's worth expanding and explicitly mentioning that `close` is asynchronous, `stop` action is the very last and thus exceptions should be properly handled?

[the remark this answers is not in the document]
dkhalanskyjb: Oh, so I didn't get this right, it's the opposite. Maybe something like this?
"The [handler] is executed as part of the closing or cancelling operation, and only after the channel reaches its final state. This means that if the handler throws an exception or hangs, the channel will still be successfully closed or cancelled." This also means that calls to channel methods can start to fail even before the handler runs to completion.

> There is a small note "exception handling is omitted". Do you think it's worth expanding

Yes, I think proper exception handling should be provided in the example.

> Could you please elaborate on why? If you mean the scenario "events are being emitted when the channel is already closed", then it's perfectly okay

Agreed, *if* there is exception handling already. However, the code in the example would be correct even without the exception handling if
* `invokeOnClose` was guaranteed to finish before the channel starts to decline new operations,
* and `callbackBasedApi.stop()` guaranteed that, after it was called, no more emissions happened.

> The very last one -- `close`/`cancel` typically are quite fast operations (e.g. one doesn't have to wait for all "children" to complete),

Makes sense, but I wouldn't bet one way or another when it came to user-supplied callbacks, as they could be of either the "notify about" or the "fine-grained lifecycle control" sort (and from the sample code, it's easy to assume it's the latter). For example, such callbacks could conceivably support use cases like "send one last very important element from the handler," implying that the channel is not allowed to close until the handler finishes, etc. I think we need to explicitly drive home that it's not the case here.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -105,33 +105,45 @@ public interface SendChannel<in E> {
...
+     *       .onClosed { ... channel is already closed, but callback hasn't stopped yet ... }
      * }
      *
      * val uiUpdater = launch(Dispatchers.Main, parent = UILifecycle) {
-     *    events.consume {}
-     *    events.cancel()
+     *    events.consume { ... handle events ... }
      * }
-     *
+     * // Stop the callback as soon as the channel is closed or cancelled
```
dkhalanskyjb, line 130:
```suggestion
     * // Stop the callback after the channel is closed or cancelled
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -105,33 +105,45 @@ public interface SendChannel<in E> {
...
+     * Unhandled exceptions from [handler] are propagated to the closing or cancelling operation's caller.
+     *
+     * Example of usage:
      * ```
-     * val events = Channel(UNLIMITED)
+     * val events = Channel<Event>(UNLIMITED)
      * callbackBasedApi.registerCallback { event ->
      *   events.trySend(event)
+     *       .onFailure { ... buffer overflown, trySend is unsuccessful ... }
+     *       .onClosed { ... channel is already closed, but callback hasn't stopped yet ... }
```
dkhalanskyjb, line 124:
```suggestion
     *       .onClosed { ... channel is already closed, but the callback hasn't stopped yet ... }
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -105,33 +105,45 @@ public interface SendChannel<in E> {
...
+     *    events.consume { ... handle events ... }
      * }
-     *
+     * // Stop the callback as soon as the channel is closed or cancelled
      * events.invokeOnClose { callbackBasedApi.stop() }
      * ```
+     * Note that `invokeOnClose` is invoked **after** the channel is in its final state,
+     * meaning that [handler] always observes already closed/cancelled channel and
+     * that callback's code should be ready to handle the situation when callback
+     * is yet active, but the channel is already closed.
```
dkhalanskyjb, line 136: I agree that this is already excessive and we would be repeating ourselves.

kotlinx-coroutines-core/common/src/channels/Channel.kt
dkhalanskyjb, line 124:
```suggestion
     *       .onFailure { /* buffer overflown, trySend is unsuccessful */ }
     *       .onClosed { /* channel is already closed, but callback hasn't stopped yet */ }
```

Mixing text and code leads to a festive feel, with some punctuation and words in different colors.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -105,33 +105,45 @@ public interface SendChannel<in E> {
...
      * callbackBasedApi.registerCallback { event ->
      *   events.trySend(event)
+     *       .onFailure { ... buffer overflown, trySend is unsuccessful ... }
+     *       .onClosed { ... channel is already closed, but callback hasn't stopped yet ... }
      * }
      *
      * val uiUpdater = launch(Dispatchers.Main, parent = UILifecycle) {
-     *    events.consume {}
-     *    events.cancel()
+     *    events.consume { ... handle events ... }
```
dkhalanskyjb, line 128:
```suggestion
     *    events.consume { /* handle events */ }
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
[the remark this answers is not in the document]
dkhalanskyjb: We can't be sure that we write documentation that everyone understands perfectly. We are sending mixed signals at times: see "Stop the callback as soon as [...]", for example, which one could easily interpret as "the callback will be stopped as soon as [...]", don't you agree? When there's no one to ask, people infer from examples, copy-pasting them, adapting, and seeing if everything works, so I think that examples shouldn't gloss over error handling.

I like the new `onClosed` line, it says unambiguously that no, you can't rely on the callback immediately stopping, you have to handle this.

Also, do we need this line?
> `     *       .onFailure { ... buffer overflown, trySend is unsuccessful ... }`

Can this ever happen with `UNLIMITED`?

dkhalanskyjb (approved):

ifuatgucluer:
Thanks

22 Mar 2023 Çar 16:05 tarihinde Dmitry Khalanskiy ***@***.***>
şunu yazdı:

ifuatgucluer:
good job

PR #3686, opened 2023-03-21
Reduce contention on worker's lock in CoroutineScheduler

### LINKED ISSUE #3652: Avoid synchronized section in CoroutineScheduler during internal pool growth
We had multiple freeze reports from IDEA, all with the following stacktrace:

```
	at kotlinx.coroutines.scheduling.CoroutineScheduler.createNewWorker(CoroutineScheduler.kt:1007)
		at kotlinx.coroutines.scheduling.CoroutineScheduler.tryCreateWorker(CoroutineScheduler.kt:439)
		at kotlinx.coroutines.scheduling.CoroutineScheduler.tryCreateWorker$default(CoroutineScheduler.kt:430)
		at kotlinx.coroutines.scheduling.CoroutineScheduler.signalCpuWork(CoroutineScheduler.kt:426)
		at kotlinx.coroutines.scheduling.CoroutineScheduler.dispatch(CoroutineScheduler.kt:398)
		at kotlinx.coroutines.scheduling.CoroutineScheduler.dispatch$default(CoroutineScheduler.kt:382)
		at kotlinx.coroutines.scheduling.SchedulerCoroutineDispatcher.dispatch(Dispatcher.kt:97)
```

During the startup phase, IDEA aggressively launches hundreds of coroutines, and some of them are run from the main thread;
Due to the fact that the growing mechanism uses coarse-grained lock as the easiest way to avoid thread overprovision, that during contention might cause sub-ms pauses

qwwdfsad:
The previous implementation was prone to a non-trivial contention that caused EDT freezes and, potentially, Android's ANRs.

Two root causes were identified:

1) Thread() constructor that has a non-trivial complexity along with JVM upcalls and is significantly slower than any other regular allocation 2) Thread.start() is on itself a JVM upcall that ends up on a global JVM lock[s]

Thread.start() is now invoked when the lock is released to reduce contention. The first root cause is not addressed as optimistic thread allocation may lead to a potential CPU waste due to how optimistically lock-less detection is and because Thread.start() is an order of magnitude slower anyway

Fixes #3652

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
@@ -28,11 +28,12 @@ internal class ResizableAtomicArray<T>(initialLength: Int) {
         val curLen = curArray.length()
         if (index < curLen) {
             curArray[index] = value
-        } else {
-            val newArray = AtomicReferenceArray<T>((index + 1).coerceAtLeast(2 * curLen))
-            for (i in 0 until curLen) newArray[i] = curArray[i]
-            newArray[index] = value
-            array = newArray // copy done
+            return
         }
+        // It would be nice to copy array in batch instead of 1 by 1, but it seems like Java has no API for that
+        val newArray = AtomicReferenceArray<T>((index + 1).coerceAtLeast(2 * curLen))
+        for (i in 0 until curLen) newArray[i] = curArray[i]
+        newArray[index] = value
+        array = newArray // copy done
     }
 }
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -252,16 +252,21 @@ internal class CoroutineScheduler(
 
     /**
      * State of worker threads.
-     * [workers] is array of lazily created workers up to [maxPoolSize] workers.
+     * [workers] is a dynamic array of lazily created workers up to [maxPoolSize] workers.
      * [createdWorkers] is count of already created workers (worker with index lesser than [createdWorkers] exists).
-     * [blockingTasks] is count of pending (either in the queue or being executed) tasks
+     * [blockingTasks] is count of pending (either in the queue or being executed) blocking tasks.
+     *
+     * Workers array is also used as a lock for workers creation and termination sequence.
      *
      * **NOTE**: `workers[0]` is always `null` (never used, works as sentinel value), so
      * workers are 1-indexed, code path in [Worker.trySteal] is a bit faster and index swap during termination
-     * works properly
+     * works properly.
+     *
+     * Initial size is `Dispatchers.Default` size * 2 to prevent unnecessary resizes for slightly or steadily loaded
+     * applications.
      */
     @JvmField
-    val workers = ResizableAtomicArray<Worker>(corePoolSize + 1)
+    val workers = ResizableAtomicArray<Worker>((corePoolSize + 1) * 2)
 
     /**
      * The `Long` value describing the state of workers in this pool.
@@ -456,12 +461,14 @@ internal class CoroutineScheduler(
         }
     }
 
-    /*
+    /**
      * Returns the number of CPU workers after this function (including new worker) or
      * 0 if no worker was created.
      */
     private fun createNewWorker(): Int {
-        synchronized(workers) {
+        var worker: Worker
+        // kotlin.s to have contracts working
+        val result = kotlin.synchronized(workers) {
             // Make sure we're not trying to resurrect terminated scheduler
             if (isTerminated) return -1
             val state = controlState.value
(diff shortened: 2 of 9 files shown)


kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -456,12 +461,14 @@ internal class CoroutineScheduler(
...
-    /*
+    /**
      * Returns the number of CPU workers after this function (including new worker) or
      * 0 if no worker was created.
      */
     private fun createNewWorker(): Int {
-        synchronized(workers) {
+        var worker: Worker
+        // kotlin.s to have contracts working
+        val result = kotlin.synchronized(workers) {
```
qwwdfsad, line 471: This one is totally bonkers.

There is no way to mark `expect` declaration with contracts, but if I mark actual `kx.c.synchronized`  with contacts, compiler start reporting "default value is unused" when compiling common code, triggering failure due to `-Werror`

dkhalanskyjb: "Default value is unused" seems to be a correct warning that we could just fix. I see two ways to do that:
* Make `synchronized` have the contract on all platforms (undesirable, as the IDE still won't reflect the warnings corresponding to the contracts),
* Add a wrapper function with the contract (example: https://gist.github.com/dkhalanskyjb/43a6668954cd7ee781473339c5a03c32). Can be done like in my patch or by adding a `synchronizedWithContract` and using that where applicable and keeping `synchronized` as is.

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
```diff
@@ -28,11 +28,12 @@ internal class ResizableAtomicArray<T>(initialLength: Int) {
...
-            for (i in 0 until curLen) newArray[i] = curArray[i]
-            newArray[index] = value
-            array = newArray // copy done
+            return
         }
+        // It would be nice to copy array in batch instead of 1 by 1, but it seems like Java has no API for that
+        val newArray = AtomicReferenceArray<T>((index + 1).coerceAtLeast(2 * curLen))
+        for (i in 0 until curLen) newArray[i] = curArray[i]
+        newArray[index] = value
+        array = newArray // copy done
```
dkhalanskyjb, line 37: I'm amused by how our styles are **completely** opposite in this aspect. If I just saw the green code somewhere, I'd be very tempted to edit it, and the result would be the red code.

Just like you say when you see the opposite style by me, "minor, up to you".

dkhalanskyjb: Sure, I don't mind. At one point we'll probably have to establish a common style to avoid flip-flopping, but not today.

I think I also should elaborate on my preference, though; I don't remember if I already did that, but probably not.
You try to reduce nesting, but I find nesting very useful for judging control flow at a glance.
```kotlin
if (...) {
  ...
} else {
  ...
}
...
```
If we squint, don't read the code, and just go through the control flow structures, we can still get a sense of what's going on... unless there's a `return` somewhere in those `if` statements, breaking all intuition. Such rogue `return` statements break the natural control flow that indentation demonstrates—just like `goto` or `throw`, for example, and so I reserve them for exceptional cases. I try to structure code in such a way that squinting still doesn't obscure the control flow in the common case.

For my style, the case we're discussing is cut-and-clear: the *normal* way for this function to execute is to go to the first branch. Exiting there is no emergency. If I squint, I see a resize *typically* happening, but this doesn't reflect reality.

The most semantically clear version to me is not even the red code but
```kotlin
        val curLen = curArray.length()
        if (index >= curLen) {
            val newArray = AtomicReferenceArray<T>((index + 1).coerceAtLeast(2 * curLen))
            for (i in 0 until curLen) newArray[i] = curArray[i]
            array = newArray
        }
        curArray[index] = value
```
It reflects the actual idea of "we're assigning a value, but if the index is large, we also resize".

Before introducing such an edit, I'd need to check the code to be sure that nothing relies on a happens-before induced by the volatile read and write of the array, but I can't imagine any issues, as we can pretend that this `set` is split in two operations of resizing and then assigning the new value, with resizing supposedly happening only occasionally.

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -479,12 +486,14 @@ internal class CoroutineScheduler(
...
+            worker = Worker(newIndex)
             workers.setSynchronized(newIndex, worker)
             require(newIndex == incrementCreatedWorkers())
-            worker.start()
-            return cpuWorkers + 1
+            cpuWorkers + 1
         }
+
+        worker.start() // Start worker when the lock is released to reduce contention, see #3652
+        return result
```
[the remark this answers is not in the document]
qwwdfsad: Good point. I've initially named it `cpuWorkers` but didn't like the name shadowing. Point-free style is even better here

kotlinx-coroutines-core/jvm/src/internal/ResizableAtomicArray.kt
qwwdfsad: Thanks for the explanation! I'll try to internalize it for a while.

>The most semantically clear version

I second this one, it seems to do trick -- both redundant nesting is gone and the overall "intent" (what is a normal code path and what is not) is much more clear.

dkhalanskyjb (approved):

PR #3703, opened 2023-04-05
Enable jvmLincheckTest Gradle task back

FenstonSingel:
This commit fixes the buildscript part that disabled Kover for jvmLincheckTest Gradle task. Previous implementation of this part disabled jvmLincheckTest altogether because `enabled` property that was being set to false actually belonged to DefaultTask class (inherited by jvmLincheckTest) instead of Kover API.


qwwdfsad (approved):
Nice catch!

PR #3704, opened 2023-04-06
Fix link to Dispatchers.IO in Readme

nikpachoo:


dkhalanskyjb:
Yes and no. This is the correct fix, but this section is autogenerated by the `knit` tool. If we change this line without fixing the tool itself, we'll simply have build issues and, the next time we run the `knit` tool, the issue will return.

lukellmann:
The link is correct again with 1.7.1: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-i-o.html

PR #3711, opened 2023-04-12
Fix collecting SharedFlow into a mutable collection

### LINKED ISSUE #3706: `SharedFlow.toList(mutableList)`: difference behavior between 1.6.4 vs 1.7.0-Beta
Sometimes, I use this method to write simple test on SharedFlow.
Version 1.6.4 works well.
Since 1.7.0-Beta, this method is deprecated and the implementation seem be wrong

```kotlin
val acc = MutableList()
val job = scope.launch(start = Undispatched) {
  sharedFlow.toList(acc)
}

// later...

job.cancel()
assertContentEquals(
 expected = ...,
 actual = acc
)
```

IMO, the correct impl. is `SharedFlow.toList(acc) = (this as Flow).toList(acc)`

dkhalanskyjb:
Fixes #3706

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
@@ -144,8 +144,17 @@ public inline fun <T> SharedFlow<T>.retryWhen(noinline predicate: suspend FlowCo
     level = DeprecationLevel.WARNING
 )
 @InlineOnly
-public suspend inline fun <T> SharedFlow<T>.toList(destination: MutableList<T> = ArrayList()): List<T> =
+public suspend inline fun <T> SharedFlow<T>.toList(): List<T> =
     (this as Flow<T>).toList()
+
+/**
+ * @suppress
+ */
+@InlineOnly
+public suspend inline fun <T> SharedFlow<T>.toList(destination: MutableList<T>): Nothing {
+    (this as Flow<T>).toList(destination)
+    throw IllegalStateException("this code is supposed to be unreachable")
+}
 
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -144,9 +144,18 @@ public inline fun <T> SharedFlow<T>.retryWhen(noinline predicate: suspend FlowCo
...
 @InlineOnly
-public suspend inline fun <T> SharedFlow<T>.toList(destination: MutableList<T> = ArrayList()): List<T> =
+public suspend inline fun <T> SharedFlow<T>.toList(): List<T> =
     (this as Flow<T>).toList()
 
+/**
+ * @suppress
+ */
+@InlineOnly
+public suspend inline fun <T> SharedFlow<T>.toList(destination: MutableList<T>): Nothing {
```
qwwdfsad, line 154: Should this one be in lint file? 
Would be nice to have a test from the issue as well, as the reference for the future

dkhalanskyjb: Not sure, but I guess it *is* for linting. It works the same as `toList`, but will warn about unreachable code below.

qwwdfsad: True.
I am rather concerned that [in coroutines] `Lint` is something "don't use this, rather use that", though this works for me as well.

It's weird though that public declaration that we know people use is marked with `@suppress` and has no documentation

PR #3724, opened 2023-04-25
Fix MutexCancellationStressTest flakiness

qwwdfsad:


dkhalanskyjb (approved):
The sponsor of this PR review is the amazing Difftastic tool: https://difftastic.wilfred.me.uk/ Instead of the atrocious results produced by Github or the default `git diff` or the IntelliJ IDEA diff that is actually much better but still off, I get the completely sensible
2023-04-26T14:09:20,859584239+02:00

qwwdfsad:
My bad for messing with the diff though, was stressing the test with hand-rolled loop and this is a leftover

PR #3727, opened 2023-04-26
Enable -Xjvm-default=all for benchmarks

udalov:
And remove usages of JvmDefault, which is going to be deprecated with error in KT-54746.


qwwdfsad (approved):
Let's ensure the build passes and it's good to go

woainikk (approved):

qwwdfsad:
Build has failed due to nodejs flakiness, everything else is ok

PR #3730, opened 2023-05-01
Fix non-linearizability in `BufferedChannel.expandBuffer()`

ndkoval:
`expandBuffer()` was incorrectly processing removed segments.


ndkoval:
Before merging, we need to check that nightly tests succeed. 
https://teamcity.jetbrains.com/project/KotlinTools_KotlinxCoroutines?branch=fix-nonlinearizability-in-buffered-channel-expand-buffer&mode=builds

qwwdfsad:
It's timed out unfortunately

Issue #1631, opened 2019-10-22
recoverStackTrace breaks exception message


orange-buffalo:
Consider the following test case:
```kotlin
class ExceptionTest {
    @Test
    fun `should preserve exception message`() {
        Assertions.assertThatThrownBy {
            runBlocking {
                withContext(newFixedThreadPoolContext(10, "my-context")) {
                    throw MyTokenException("42")
                }
            }
        }.isInstanceOf(MyTokenException::class.java).hasMessage("Token 42 is not valid")
    }
}

class MyTokenException(token: String) : RuntimeException("Token $token is not valid")
```

It is failing on message assertion:
```
Expecting message to be:
  <"Token 42 is not valid">
but was:
  <"Token Token 42 is not valid is not valid">
```

It looks like `recoverStackTrace / tryCopyException` incorrectly recovers the exception. #1040 mentions that similar problem is fixed in 1.2.0, but this test fails on 1.3.2.

qwwdfsad:
I am not sure we can do better automatically without reflectively hacking Throwable class (that probably will produce a shameful warning on Java 9+), I'll try to figure it out.

To understand why it happens, I recommend you to get familiar with [stacktrace recovery machinery](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/debugging.md#stacktrace-recovery-machinery):

The main idea is to reflectively create a copy of the exception, preserving as much information from the original exception as possible. Usually, it is a message and a cause, so we reflectively call a constructor with `String` parameter and, if possible, with `Throwable` (or call `initCause`, whatever) and we cannot potentially know whether this `String` parameter maps directly to `Throwable.message` or not.

As a workaround, you can fix this problem by extending `CopyableThrowable` in your `MyTokenException` class

monosoul:
Having the same issue here. I can set property `kotlinx.coroutines.stacktrace.recovery` to `false` to run tests properly, but the problem here is that it's a system property. And if I run tests in parallel, setting this property in a single test would definitely affect other tests.

qwwdfsad:
@monosoul do you mean broken exception message or stacktrace recovery in general?

I don't think we can provide something besides system property, because stacktrace recovery was designed as **static**  property in the first place. Otherwise, everyone `if (STACKTRACE_RECOVERY_ENABLED)` becomes non-foldable by the compiler, the implementation should be aware that this property can change its value and Android minifier will have a hard time removing stacktrace recovery machinery

[...]

PR #3731, opened 2023-05-01
Properly recover exceptions when they are constructed from 'Throwable…

### LINKED ISSUE #3714: recoverStackTrace causes nested exceptions
**Describe the bug**

When throwing an exception in a coroutine scope it sometimes becomes nested one level too deep in the enclosing exception.

For example, if I throw a: `TestException(IllegalStateException("test"))` (a TestException with a cause of IllegalStateException with a message), then the coroutine scope will throw an exception that is: `TestException(TestException(IllegalStateException("test")))`.

**Provide a Reproducer**

The following test will fail as a result of this bug:

```
@Test
fun testNestedException() = runTest {
    val result = runCatching {
        coroutineScope<Unit> {
            throw NestedException(IllegalStateException("ERROR"))
        }
    }
    val ex = result.exceptionOrNull() ?: error("Expected to fail")
    assertTrue(ex is NestedException)
    assertTrue(ex.cause is IllegalStateException)
    assertEquals("ERROR", (ex.cause as IllegalStateException).message)
}

class NestedException : RuntimeException {
    constructor(message: String) : super(message)
    constructor(cause: Throwable) : super(cause)
    constructor() : super()
}
```

I can get this test to not fail by changing the order of the constructors in the class to:

```
class NestedException : RuntimeException {
    constructor(cause: Throwable) : super(cause)
    constructor(message: String) : super(message)
    constructor() : super()
}
```
But I don't think this is actually a good solution as it shouldn't matter the order of the constructors in my exception, it should work the same way regardless.

I think the issue is at [this line](https://github.com/Kotlin/kotlinx.coroutines/blob/3fc1c20a87a689c09f062afc9e3ac347829d29f7/kotlinx-coroutines-core/jvm/src/internal/ExceptionsConstructor.kt#L30) where the wrong constructor is being selected (the string constructor instead of the cause constructor).

qwwdfsad:
…(cause)' constructor.

And restore the original behaviour. After #1631 this constructor's recovery mechanism was broken because 'Throwable(cause)' changes the message to be equal to 'cause.toString()' which isn't equal to the original message.

Also, make reflective constructor choice independable from source-code order

Fixes #3714


kotlinx-coroutines-core/jvm/src/internal/StackTraceRecovery.kt
```diff
@@ -84,9 +84,6 @@ private fun <E : Throwable> recoverFromStackFrame(exception: E, continuation: Co
 
 private fun <E : Throwable> tryCopyAndVerify(exception: E): E? {
```
dkhalanskyjb, line 85: This method is now exactly the same as just `tryCopyException`.

kotlinx-coroutines-core/jvm/src/internal/ExceptionsConstructor.kt
```diff
@@ -32,13 +32,29 @@ internal fun <E : Throwable> tryCopyException(exception: E): E? {
...
+        // (m) -> 2
+        // (c) -> 1
+        // () -> 0
+        val params = it.parameterTypes // cloned array
+        when (params.size) {
+            2 -> 3
+            1 -> if (params[0] == String::class.java) 2 else 1
+            0 -> 0
+            else -> -1
+        }
```
dkhalanskyjb, line 55: I agree with the change, but I'd like to propose a stylistic edit: https://gist.github.com/dkhalanskyjb/2c2b4362920a01aafb3955216fd0e501 It makes the code more straightforward, in my view, but leads to calls to `safeCtor` even for constructors that don't get chosen in the end.

qwwdfsad: Sure, works for me.
Calls to `createConstructor` are cached, so it should not be a problem.

I've applied your patch **and** removed `inline` from `safeCtor` to reduce the size of the classfile (-25%)

dkhalanskyjb (approved):

PR #3737, opened 2023-05-03
Executor.invoke suspend function.

odedniv:
Basically a conversion of Executor.execute to a suspend function.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -107,6 +107,13 @@ public fun Executor.asCoroutineDispatcher(): CoroutineDispatcher =
 public fun CoroutineDispatcher.asExecutor(): Executor =
     (this as? ExecutorCoroutineDispatcher)?.executor ?: DispatcherExecutor(this)
 
+/** Executes the [Runnable] and suspends until the execution completes. */
+public inline suspend operator fun Executor.invoke(runnable: Runnable) {
+    withContext(asCoroutineDispatcher()) {
+        launch { runnable.run() }
+    }
+}
+
 private class DispatcherExecutor(@JvmField val dispatcher: CoroutineDispatcher) : Executor {
     override fun execute(block: Runnable) {
         if (dispatcher.isDispatchNeeded(EmptyCoroutineContext)) {
--- a/kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
@@ -122,4 +122,16 @@ class ExecutorsTest : TestBase() {
         dispatcher.close()
         check(executorService.isShutdown)
     }
+
+    @Test
+    fun testInvoke() {
+        val executor = Executors.newSingleThreadExecutor { r -> Thread(r, "TestExecutor") }
+        runBlocking {
+            executor {
+                expect(1)
+                checkThreadName("TestExecutor")
+            }
+            finish(2)
+        }
+    }
 }


odedniv:
Took the name of `CoroutineDispatcher.invoke()`.

dkhalanskyjb:
Sorry, I don't get it. How is it different from just `Executor.execute`, and why do you need it? Both accept non-suspend functions, and both don't wait for the result.

odedniv:
`Executor.execute` is technically an asynchronous call (as it can executes in some other thread), an equivalent to `CoroutineDispatcher.dispatch`.

`Executor.invoke` is equivalent to `CoroutineDispatcher.invoke`, which _is_ a suspend function that waits for the `execute` to finish. With `invoke` the Kotlin caller can decide whether to wait for the execution to finish or launch it asynchronously.

This is the equivalent of:
```kt
executor.execute {
  runnable.run()
  // <-- Do something here after the runnable finished.
}
```

Could also be implemented as:

```kt
suspendCoroutine { continuation ->
  executor.execute {
    try {
      runnable.run()
    } catch (e: Throwable) {
      continuation.resumeWithException(e)
      return@execute
    }
    continuation.resume(Unit)
  }
}
```

dkhalanskyjb:
Ok, I see.

Your code could be simplified: there's no need to `launch` there.
```kotlin
suspend inline operator fun Executor.invoke(runnable: Runnable) {
    withContext(asCoroutineDispatcher()) {
        runnable.run()
    }
}
```
At this point, you could as well pass the `CoroutineScope` to the runnable, because why not, it's already there:
```kotlin
suspend operator fun Executor.invoke(runnable: CoroutineScope.() -> Unit) =
    withContext(asCoroutineDispatcher(), runnable)
```
Make it a bit more general, to allow it to return arbitrary results:
```kotlin
suspend operator fun <T> Executor.invoke(runnable: CoroutineScope.() -> T) : T =
    withContext(asCoroutineDispatcher(), runnable)
```
… and use the existing function that you mentioned several times that has exactly the same definition:
```kotlin
suspend operator fun <T> Executor.invoke(runnable: CoroutineScope.() -> T) : T =
    asCoroutineDispatcher()(runnable)
```
At this point, `runnable` can even itself be `suspend`:
```kotlin
suspend operator fun <T> Executor.invoke(runnable: suspend CoroutineScope.() -> T) : T =
    asCoroutineDispatcher()(runnable)
```
So, the function you're proposing is precisely the composition of `asCoroutineDispatcher()` and `CoroutineDispatcher.invoke`, but less general. Is everything correct?

If so, I don't think this is useful in a general-purpose library: we want people to use `CoroutineDispatcher` whenever possible, and just converting an executor to a dispatcher once via `asCoroutineDispatcher()` and then using its `invoke` has better ergonomics (it supports `suspend` functions, passes `CoroutineScope`, etc) and leads to code that more uniformly uses coroutines.

kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -107,6 +107,13 @@ public fun Executor.asCoroutineDispatcher(): CoroutineDispatcher =
 public fun CoroutineDispatcher.asExecutor(): Executor =
     (this as? ExecutorCoroutineDispatcher)?.executor ?: DispatcherExecutor(this)
 
+/** Executes the [Runnable] and suspends until the execution completes. */
+public inline suspend operator fun Executor.invoke(runnable: Runnable) {
+    withContext(asCoroutineDispatcher()) {
```
bcmedeiros, line 112: isn't this going to create a new dispatcher for each invocation?

odedniv:
Sorry I thought I closed this already.

I agree with the review, it's very easy to convert to dispatcher and this would just encourage using executors.

kotlinx-coroutines-core/jvm/src/Executors.kt
odedniv: Although I closed this PR I'm curious, is that an issue?

dkhalanskyjb: It's always better to avoid running code several times when one time suffices. `asCoroutineDispatcher` is a small memory allocation, followed by a reflective method call if the `Executor` is a `ScheduledThreadPoolExecutor`. When it comes to reflective calls, your mileage may vary w.r.t. how slow they are, but in any case, I doubt that this would show up in a profiler in the overwhelming majority of cases.

PR #3741, opened 2023-05-07
Fix documented exception type for `Flow.single()`

lukellmann:
[`require()`](https://github.com/lukellmann/kotlinx.coroutines/blob/d17e875275ff4ffbf5cf1a09e3fdbbf7d34acb83/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt#L58) throws `IllegalArgumentException`, not `IllegalStateException`.


dkhalanskyjb (approved):
Thanks!

PR #3742, opened 2023-05-08
Try once more to restore binary compatibility in `runTest`

### LINKED ISSUE #3673: 1.7.0-Beta: NoSuchMethodError: void kotlinx.coroutines.test.TestBuildersKt.runTest$default(kotlinx.coroutines.test.TestScope, long, kotlin.jvm.functions.Function2, int, java.lang.Object)
**Describe the bug**

When updating a project to use `kotlinx.coroutines` `1.7.0-Beta`, existing tests fail with a `NoSuchMethodError` for a `runTest` that's being called internally by the Compose test utilities.

The full stack trace looks like the following (but fail for every Compose test):

```
com.alexvanyo.composelife.resourcestate.ResourceStateComposableTests > collect_as_state_is_correct FAILED
    java.lang.NoSuchMethodError: 'void kotlinx.coroutines.test.TestBuildersKt.runTest$default(kotlinx.coroutines.test.TestScope, long, kotlin.jvm.functions.Function2, int, java.lang.Object)'
        at androidx.compose.ui.test.AndroidComposeUiTestEnvironment.withTestCoroutines(ComposeUiTest.android.kt:342)
        at androidx.compose.ui.test.AndroidComposeUiTestEnvironment.access$withTestCoroutines(ComposeUiTest.android.kt:218)
        at androidx.compose.ui.test.AndroidComposeUiTestEnvironment$runTest$1$1$1.invoke(ComposeUiTest.android.kt:294)
        at androidx.compose.ui.test.junit4.IdlingStrategy.withStrategy(IdlingStrategy.android.kt:52)
        at androidx.compose.ui.test.AndroidComposeUiTestEnvironment$runTest$1$1.invoke(ComposeUiTest.android.kt:293)
        at androidx.compose.ui.test.junit4.IdlingResourceRegistry.withRegistry(IdlingResourceRegistry.jvm.kt:157)
        at androidx.compose.ui.test.AndroidComposeUiTestEnvironment$runTest$1.invoke(ComposeUiTest.android.kt:292)
        at androidx.compose.ui.test.junit4.ComposeRootRegistry.withRegistry(ComposeRootRegistry.android.kt:146)
        at androidx.compose.ui.test.AndroidComposeUiTestEnvironment.runTest(ComposeUiTest.android.kt:291)
        at androidx.compose.ui.test.ComposeUiTest_androidKt.runAndroidComposeUiTest(ComposeUiTest.android.kt:107)
        at androidx.compose.ui.test.ComposeUiTest_androidKt.runComposeUiTest(ComposeUiTest.android.kt:63)
        at com.alexvanyo.composelife.resourcestate.ResourceStateComposableTests.collect_as_state_is_correct(ResourceStateComposableTests.kt:39)
```

**Provide a Reproducer**

Reproducing PR: https://github.com/alexvanyo/composelife/pull/838

It looks like the update to `runTest` in `1.7.0-Beta` might not have been binary compatible?

dkhalanskyjb:
Fixes #3673

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -578,8 +578,8 @@ internal expect fun dumpCoroutines()
 @JvmName("runTest\$default")
 @Suppress("DEPRECATION", "UNUSED_PARAMETER")
 public fun TestScope.runTestLegacy(
-    dispatchTimeoutMs: Long?,
+    dispatchTimeoutMs: Long,
     testBody: suspend TestScope.() -> Unit,
-    unused1: Int?,
+    unused1: Int,
     unused2: Any?,
-): TestResult = runTest(dispatchTimeoutMs ?: 60_000, testBody)
+): TestResult = runTest(dispatchTimeoutMs, testBody)
--- a/kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
+++ b/kotlinx-coroutines-test/api/kotlinx-coroutines-test.api
@@ -22,7 +22,7 @@ public final class kotlinx/coroutines/test/TestBuildersKt {
 	public static final fun runTest (Lkotlinx/coroutines/test/TestScope;JLkotlin/jvm/functions/Function2;)V
 	public static synthetic fun runTest$default (Lkotlin/coroutines/CoroutineContext;JLkotlin/jvm/functions/Function2;ILjava/lang/Object;)V
 	public static synthetic fun runTest$default (Lkotlinx/coroutines/test/TestCoroutineScope;JLkotlin/jvm/functions/Function2;ILjava/lang/Object;)V
-	public static final synthetic fun runTest$default (Lkotlinx/coroutines/test/TestScope;Ljava/lang/Long;Lkotlin/jvm/functions/Function2;Ljava/lang/Integer;Ljava/lang/Object;)V
+	public static final synthetic fun runTest$default (Lkotlinx/coroutines/test/TestScope;JLkotlin/jvm/functions/Function2;ILjava/lang/Object;)V
 	public static final fun runTest-8Mi8wO0 (Lkotlin/coroutines/CoroutineContext;JLkotlin/jvm/functions/Function2;)V
 	public static final fun runTest-8Mi8wO0 (Lkotlinx/coroutines/test/TestScope;JLkotlin/jvm/functions/Function2;)V
 	public static synthetic fun runTest-8Mi8wO0$default (Lkotlin/coroutines/CoroutineContext;JLkotlin/jvm/functions/Function2;ILjava/lang/Object;)V


kotlinx-coroutines-test/common/src/TestBuilders.kt
```diff
@@ -578,8 +578,8 @@ internal expect fun dumpCoroutines()
...
 @Suppress("DEPRECATION", "UNUSED_PARAMETER")
 public fun TestScope.runTestLegacy(
-    dispatchTimeoutMs: Long?,
+    dispatchTimeoutMs: Long,
     testBody: suspend TestScope.() -> Unit,
-    unused1: Int?,
+    unused1: Int,
     unused2: Any?,
-): TestResult = runTest(dispatchTimeoutMs ?: 60_000, testBody)
+): TestResult = runTest(dispatchTimeoutMs, testBody)
```
ephemient, line 585:
```suggestion
): TestResult = runTest(if (unused1 and 1 != 0) DEFAULT_DISPATCH_TIMEOUT_MS else dispatchTimeoutMs, testBody)
```

qwwdfsad (approved):
Good job.

I haven't dived deep into the restored method's semantics though, it would be nice to manually check compose's library normally operates before the release

dkhalanskyjb:
I looked into two reproducers provided to us: https://github.com/Kotlin/kotlinx.coroutines/issues/3673#issuecomment-1537441738 and https://github.com/Kotlin/kotlinx.coroutines/issues/3673#issue-1619361190. On both, the issue reproduces on 1.7.0, but with a locally-published version with this patch, one project starts to build successfully, while the other changes the errors from `NoSuchMethodError` to some business-logic-related ones. I checked, and both do actually call the newly-provided method.

kotlinx-coroutines-test/common/src/TestBuilders.kt
dkhalanskyjb: I think there's no need: if somebody did actually provide `dispatchTimeoutMs`, the `$default` overload wouldn't be chosen.

eygraber:
Won't this still be an issue because of https://github.com/Kotlin/kotlinx.coroutines/pull/3742#issuecomment-1544320961

dkhalanskyjb:
@eygraber I don't understand your approach there and don't think it would work anyway, but the likely reason for `StackOverflowError` is that what you wrote was essentially `fun runTest() = runTest()`, which, yes, would cause infinite recursion and a stack overflow. Here https://github.com/Kotlin/kotlinx.coroutines/pull/3742/files, you can see that there is no recursion: `runTestLegacy` (with the `JvmName` being what it needed to be for binary compatibility) calls `runTest`, an entirely different function.

kotlinx-coroutines-test/common/src/TestBuilders.kt
ephemient: Isn't that the wrong way around? We're in this function because they didn't supply all the parameters, therefore we need to fill them in.

dkhalanskyjb: Oh, yes, you're right, it *is* the other way around. Let me think about it.

eygraber:
My point was that it seems like something is also expecting this function to be present:

```
java.lang.NoSuchMethodError: 'void kotlinx.coroutines.test.TestBuildersKt.runTest(kotlinx.coroutines.test.TestScope, long, kotlin.jvm.functions.Function2)'
```

Note that it returns void, and has a different signature than the function that was restored in this PR.

kotlinx-coroutines-test/common/src/TestBuilders.kt
dkhalanskyjb: You're fully correct, by decompiling this method in 1.6.4, we get
```java
   // $FF: synthetic method
   public static void runTest$default(TestScope var0, long var1, Function2 var3, int var4, Object var5) {
      if ((var4 & 1) != 0) {
         var1 = 60000L;
      }

      TestBuildersKt.runTest(var0, var1, var3);
   }
```
Thanks for saving us from one more potential bugfix release!

eygraber:
Tried with 1.7.1 and it works ¯\_(ツ)_/¯

PR #3746, opened 2023-05-09
Formalize the internal IDEA dependencies for binary compatibility

dkhalanskyjb:


kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
```diff
@@ -234,17 +236,17 @@ internal object DebugProbesImpl {
...
     /*
      * Internal (JVM-public) method to be used by IDEA debugger in the future (not used as of 1.4-M3).
      * It is equivalent to [dumpCoroutinesInfo], but returns serializable (and thus less typed) objects.
      */
-    public fun dumpDebuggerInfo(): List<DebuggerInfo> =
+    fun dumpDebuggerInfo(): List<DebuggerInfo> =
         dumpCoroutinesInfoImpl { owner, context -> DebuggerInfo(owner.info, context) }
 
-    public fun dumpCoroutines(out: PrintStream): Unit = synchronized(out) {
+    internal fun dumpCoroutines(out: PrintStream): Unit = synchronized(out) {
```
qwwdfsad, line 249: The probability of this frame appearing in user's dumps is quite high, would be nice to avoid mangling here with `@JvmName("dumpCoroutines")`

kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
```diff
@@ -888,6 +888,16 @@ public final class kotlinx/coroutines/debug/internal/DebugCoroutineInfo {
 	public final fun lastObservedStackTrace ()Ljava/util/List;
 }
 
+public final class kotlinx/coroutines/debug/internal/DebugProbesImpl {
```
qwwdfsad, line 891: Have you inferred it by looking at IDEA's code or just by looking at the comments?

qwwdfsad:
Friendly reminder

dkhalanskyjb:
Ah, ok. I misinterpreted the flow of the conversation. After the question about IDEA's side, I thought the review would go, "Well, here's the code where this is used; please go ahead, check it, and update since you're at it already."

qwwdfsad:
Happens :)

We have a full-list here for a different task: https://youtrack.jetbrains.com/issue/KTIJ-24102/Change-coroutine-debugger-implementation-to-use-a-new-package#focus=Comments-27-6989405.0-0

dkhalanskyjb:
Thanks for the list!
I updated the PR in accordance. Here are the things that I couldn't get to work:

These aren't fields, they have getters.
```
kotlinx.coroutines.debug.internal.DebugCoroutineInfo
	field context
	field sequenceNumber
	field lastObservedFrame
	field lastObservedThread
kotlinx.coroutines.CancellableContinuationImpl
	field delegate
	field context
```

I don't know how these names are generated and how to make the compatibility validator know about them:
```
kotlinx.coroutines.debug.internal.DebugProbesImpl
	method isInstalled$kotlinx_coroutines_debug()
	method isInstalled$kotlinx_coroutines_core()
```

There are no such fields. Maybe they are hidden somewhere up the inheritance chain.
```
kotlinx.coroutines.StandaloneCoroutine
	field _state
	field context
```

I think this is from `DispatchedCoroutine`. This is an `atomic` field, so the validator says `get_decision$FU`. I suppose this is from atomicfu, but I don't understand whether this is expected.
```
kotlinx.coroutines.CancellableContinuationImpl
	field _decision
```

Simply doesn't show up in the `.api` file for some reason:
```
kotlinx.coroutines.DispatchedContinuation
	field continuation
```

Not in the library at all:
```
kotlin.coroutines.CombinedContext
	method get()
kotlin.coroutines.jvm.internal.DebugMetadataKt
	method getStackTraceElement()
	method getSpilledVariableFieldMapping()
```

Also, pardon my French, but bloody hell!

qwwdfsad:
Thanks!

>These aren't fields, they have getters.

Right, but unfortunately, the underlying *private* field is used instead.

>I don't know how these names are generated 

`internal` names that are not classes/interfaces are mangled, so it's impossible to call them from Java (`$` is not a valid identifier in Java). The default mangling strategy is `name + $ + module name`

>how to make the compatibility validator know about them

No way, neither there should be, because they are internal.

> Maybe they are hidden somewhere up the inheritance chain.

One is `JobSupport._state`, the other one is the underlying field for the getter. Hint: CTRL+F12 pressed twice enables "Inherited members" overview in IDEA.

>Simply doesn't show up in the .api file for some reason:

Explicitly excluded here: https://github.com/Kotlin/kotlinx.coroutines/blob/master/build.gradle#L136

>Also, pardon my French, but bloody hell!

This all is a very educational story about project management and responsibilities in general, but this margin is too narrow to contain it :)

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -218,13 +218,16 @@ private const val SUSPENDED = 1
 private const val RESUMED = 2
 
 // Used by withContext when context dispatcher changes
-internal class DispatchedCoroutine<in T>(
+@PublishedApi
```
qwwdfsad, line 221: Do you mind explicitly writing "Used exclusively by IDEA debugger, see KTIJ-24102" in all such places (both `@PublishedAPI` and actual declarations used)

whyoleg: May be it make sense also to introduce internal annotation like `UsedByIDEADebugger` with description instead of just copy pasting the same message?

PR #3763, opened 2023-05-26
fix missing reference to already defined isClosedForSend

lucasmontano:
The `// check if already closed for send` comment is unnecessary, there's already a property defined as `isClosedForSend`. This will also prevent bugs related to multiple sources of truth related to which makes it closed for sending.

### DIFF
--- a/reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxObservable.kt
@@ -120,8 +120,7 @@ private class RxObservableCoroutine<T : Any>(
 
     // assert: mutex.isLocked()
     private fun doLockedNext(elem: T): Throwable? {
-        // check if already closed for send
-        if (!isActive) {
+        if (isClosedForSend) {
             doLockedSignalCompleted(completionCause, completionCauseHandled)
             return getCancellationException()
         }


anddreluis2 (commented):
LGTM

dkhalanskyjb:
There are several usages of `!isActive` in the code that are synonymous with `isClosedForSend`, so to fix the problem of "multiple sources of truth," one must change them all. Additionally, the code is duplicated for the `rx3` artifact and would need to change there too for consistency.

That said, I don't think it's worth even a tiny effort. "Multiple sources of truth" is not a problem at all in this particular case: this implementation is quite involved, and any changes that could affect *what it means for this `Observable` to be closed* would be so sweeping that the whole class would need to be rethought from the bottom up.

dkhalanskyjb:
Also, it's strange how much activity there is under this PR that performs some minor internal refactoring. Very rarely does even a PR with a much-requested functionality or a fix for a nasty bug get five thumbs up and also an independent review. If anyone who's interested in this PR is reading this and is not a bot, I'd love to know what in particular resonated with you!

lucasmontano:
@dkhalanskyjb I'm not a bot :) 
I was recording a video about how to double think before leaving unnecessary code comments and that's why I have only changed where it was explicitly used. 

Hope that's fine and sorry for any inconvenience.

PR #3766, opened 2023-05-31
Bump Lincheck version to 2.18.1.

ndkoval:
The previous version 2.17 contains a bug that leads to a memory leak, resulting in OutOfMemoryError


qwwdfsad:
I've rerun the build, `MutexLincheckTest` has failed :(

qwwdfsad (approved):

PR #3767, opened 2023-06-01
Add a warning that actors are part of an obsolete API

jakzal:
It's not clear from the documentation that Actors are part of an obsolete API.

re https://github.com/Kotlin/kotlinx.coroutines/issues/87

### DIFF
--- a/docs/topics/shared-mutable-state-and-concurrency.md
+++ b/docs/topics/shared-mutable-state-and-concurrency.md
@@ -490,6 +490,11 @@ have to switch to a different context at all.
 >
 {type="note"}
 
+> The Actor API will become obsolete in future updates with introduction of complex actors. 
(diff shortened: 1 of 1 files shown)


qwwdfsad (commented):
Thanks!

We'd prefer not to mention the issue and overall direction we are not working towards though. Independently on #87 state, current actors are obsolete and underdesigned

jakzal:
Thanks for taking a look. 

I just wanted to save others from disappointment who like myself might read the docs and get excited by the prospect of having built-in support for actors. Only to learn later the API is obsolete and not worth using. 🤷

qwwdfsad:
Thanks for the feedback!

It indeed looks odd, instead, I've filed https://github.com/Kotlin/kotlinx.coroutines/issues/3785 to avoid such situations

PR #3769, opened 2023-06-01
Fix newSingleThreadContext awaiting cancelled scheduled coroutines

### LINKED ISSUE #3768: newSingleThreadContext on Native can't close until timeouts finish
Reported in the Kotlin Slack originally.

Simplified reproducer:
```kotlin
    @Test
    fun timeoutAffectsClosing() = runBlocking {
        val disp: CloseableCoroutineDispatcher = newSingleThreadContext("my-test")
        val result = withContext(disp) {
            // Disabling the inner timeout will make it work correctly
            withTimeout(30.seconds) {
                "Hello"
            }
        }
        println(result) // Print "Hello"
        disp.close()
        println("Dispatcher closed: $disp") // takes 30 seconds to get here
    }
```

dkhalanskyjb:
Before the change, the Worker is not notified about its work being cancelled, due to no API being present for that.
We work around the issue by checking every 100 milliseconds whether cancellation happened.

Fixes #3768

### DIFF
--- a/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
+++ b/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
@@ -12,11 +12,6 @@ import kotlin.native.concurrent.*
 import kotlin.time.*
 import kotlin.time.Duration.Companion.milliseconds
 
-@ExperimentalCoroutinesApi
-public actual fun newSingleThreadContext(name: String): CloseableCoroutineDispatcher {
-    return WorkerDispatcher(name)
-}
-
 public actual fun newFixedThreadPoolContext(nThreads: Int, name: String): CloseableCoroutineDispatcher {
     require(nThreads >= 1) { "Expected at least one thread, but got: $nThreads" }
     return MultiWorkerDispatcher(name, nThreads)
--- a/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
+++ b/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
@@ -9,6 +9,8 @@ import kotlinx.coroutines.channels.*
 import kotlinx.coroutines.internal.*
 import kotlin.coroutines.*
 import kotlin.native.concurrent.*
+import kotlin.time.*
+import kotlin.time.Duration.Companion.milliseconds
 
 @ExperimentalCoroutinesApi
 public actual fun newSingleThreadContext(name: String): CloseableCoroutineDispatcher {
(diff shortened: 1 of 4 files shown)


dkhalanskyjb:
Ah, ok, I get it, it's because this dispatcher is reused for the `DefaultExecutor`.

dkhalanskyjb:
Maybe it makes sense to have `WorkerDispatcher` be only for the `DefaultExecutor` and have `newSingleThreadContext` be `newFixedThreadPoolContext(1)`, like on the JVM.

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -52,10 +55,24 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
             override fun dispose() {
                 disposableHolder.value = null
             }
+
+            fun isDisposed() = disposableHolder.value == null
+        }
+
+        fun Worker.runAfterDelay(block: DisposableBlock, targetMoment: TimeMark) {
+            if (block.isDisposed()) return
+            val durationUntilTarget = -targetMoment.elapsedNow()
```
qwwdfsad, line 64: Was a bit confused at the beginning when looking at documentation only, have filed https://youtrack.jetbrains.com/issue/KT-59044/Improve-various-aspects-of-TimeSource-documentation

qwwdfsad: Probably we can ignore "infinity" problem for now -- such check would be untestable anyway

qwwdfsad (approved):
Nice one!

I'm onboard with making `WorkerDispatcher` available only for `DefaultExecutor`

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -52,10 +55,24 @@ internal class WorkerDispatcher(name: String) : CloseableCoroutineDispatcher(),
...
+            if (durationUntilTarget > quantum) {
+                executeAfter(quantum.inWholeMicroseconds) { runAfterDelay(block, targetMoment) }
+            } else {
+                executeAfter(durationUntilTarget.inWholeMicroseconds, block)
+            }
         }
 
         val disposableBlock = DisposableBlock(block)
-        worker.executeAfter(timeMillis.toMicrosSafe(), disposableBlock)
+        val targetMoment = TimeSource.Monotonic.markNow() + timeMillis.milliseconds
```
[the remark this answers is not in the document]
qwwdfsad: Still unused

qwwdfsad (changes requested):
Our build started flaking consistently after the change, could you please take a look?
Otherwise LGTM

dkhalanskyjb:
The first build failure was on me, an API dump changed, which I later fixed, but the second one seems to be just Windows doing its thing, no?

dkhalanskyjb:
Ah, no, I see now that something's off.

dkhalanskyjb:
Who knew? This was an actual bug. Turns out `executeAfter` doesn't crashes on negative numbers.

qwwdfsad (approved):

PR #3771, opened 2023-06-02
Update BlockHound version to be compatible with JDK 17

### LINKED ISSUE #3701: Tests for kotlinx-coroutines-debug module fail on JDK 17 due to outdated BlockHound version
When running on JDK 17 (and potentially some earlier versions as well), BlockHound 1.0.2.RELEASE cannot process JDK class files with a new class file version, leading to failure of several tests for the `kotlinx-coroutines-debug` module.

Bumping BlockHound version to 1.0.7.RELEASE solves the inability to process class files, but discovers the problem of changed JVMTI behavior (see corresponding [BlockHound issue](https://github.com/reactor/BlockHound/issues/33) and [JDK issue](https://bugs.openjdk.org/browse/JDK-8192936)). Since the corresponding BlockHound issue is not solved at the moment of writing this, the following workaround needs to be added to `kotlinx-coroutines-debug` Gradle buildscript:

```groovy
tasks.withType(Test).configureEach {
    if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_13)) {
        jvmArgs += ["-XX:+AllowRedefinitionToAddDeleteMethods"]
    }
}
```

qwwdfsad:
* Also, exclude integration from docs, it shouldn't be there

Fixes #3701

### DIFF
--- a/kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
+++ b/kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 @file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-debug/src/CoroutinesBlockHoundIntegration.kt
```diff
@@ -10,6 +10,9 @@ import kotlinx.coroutines.scheduling.*
 import reactor.blockhound.*
 import reactor.blockhound.integration.*
 
+/**
+ * @suppress
```
qwwdfsad, line 14: To get rid of this one: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug/-coroutines-block-hound-integration/

dkhalanskyjb (approved):

PR #3778, opened 2023-06-13
Copy CopyableThreadContextElement when switching context with flowOn.

wanyingd1996:
Hi, coroutine owners, this patch is needed as I'm using the thread context element to propagate thread local changes across context switching with flow. I have added a test case that demonstrate my use case. Please take a look when you have time, thanks!


qwwdfsad:
Could you please rebase your changes onto `develop` instead of a `master`?

qwwdfsad:
oops, messed up with a remote branch, will patch it up

qwwdfsad (approved):
Thanks for the fix!

PR #3788, opened 2023-06-21
OptIn for ExperimentalNativeApi

qurbonzoda:
The API requires opt-in since in 1.9.0.
In 1.9.20 the opt-in requirement level of the annotation will be raised to ERROR.


qwwdfsad (changes requested):
Please mention the corresponding YT ticket you are working on that is blocked with this PR.
Also, the build won't pass in the current develop with `Unresolved reference: ExperimentalNativeApi` -- there is no such declaration in 1.8.20, and, unfortunately, 1.9.0 is not here yet.
I suggest you use approach similar to this one: https://github.com/Kotlin/kotlinx.coroutines/pull/3751

qurbonzoda:
Done, PTAL.

qwwdfsad (approved):
Thanks

PR #3791, opened 2023-06-23
Update user projects config: adapt build script to new TeamCity variables

woainikk:
KTI-1227


qwwdfsad (approved):
Nice one with the documentation!

PR #3798, opened 2023-06-29
Migrate from the deprecated native atomics

mvicsokolova:
This commit is required for the aggregate build that checks the change of deprecation level for native atomics (https://jetbrains.team/p/kt/reviews/10650/timeline). Usages of native atomics from `kotlin.native.concurrent` package are replaced with new atomics from `kotlin.concurrent` because old native atomics will be deprecated with error.

NOTE: this commit is not present in develop branch, because new `kotlin.concurrent` atomics are only available since Kotlin 1.9.0. IT SHOULD BE REMOVED after kotlinx.coroutines updates Kotlin version to 1.9.0.

[Here is the build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_kotlinxcoroutines?branch=update-native-atomics&buildTypeTab=overview&mode=builds) that checked this branch against the 1.9.20-dev Kotlin version. This TC config may be used to check that changes from the given branch won't break the aggregate build.

### DIFF
--- a/benchmarks/src/jmh/kotlin/benchmarks/scheduler/ForkJoinBenchmark.kt
+++ b/benchmarks/src/jmh/kotlin/benchmarks/scheduler/ForkJoinBenchmark.kt
@@ -110,8 +110,13 @@ open class ForkJoinBenchmark : ParametrizedDispatcherBase() {
         }
     }
 
-    class RecursiveAction(val coefficients: LongArray, val start: Int, val end: Int, @Volatile var result: Double = 0.0,
-                          parent: RecursiveAction? = null) : CountedCompleter<Double>(parent) {
+    class RecursiveAction(
+        val coefficients: LongArray,
+        val start: Int,
+        val end: Int,
+        @Volatile @OptIn(ExperimentalStdlibApi::class) var result: Double = 0.0,
+        parent: RecursiveAction? = null
+    ) : CountedCompleter<Double>(parent) {
 
         private var first: ForkJoinTask<Double>? = null
         private var second: ForkJoinTask<Double>? = null
--- a/benchmarks/src/jmh/kotlin/benchmarks/scheduler/ForkJoinBenchmark.kt
+++ b/benchmarks/src/jmh/kotlin/benchmarks/scheduler/ForkJoinBenchmark.kt
@@ -110,13 +110,8 @@ open class ForkJoinBenchmark : ParametrizedDispatcherBase() {
         }
     }
 
-    class RecursiveAction(
-        val coefficients: LongArray,
-        val start: Int,
-        val end: Int,
-        @Volatile @OptIn(ExperimentalStdlibApi::class) var result: Double = 0.0,
-        parent: RecursiveAction? = null
-    ) : CountedCompleter<Double>(parent) {
+    class RecursiveAction(val coefficients: LongArray, val start: Int, val end: Int, @Volatile var result: Double = 0.0,
+                          parent: RecursiveAction? = null) : CountedCompleter<Double>(parent) {
 
         private var first: ForkJoinTask<Double>? = null
         private var second: ForkJoinTask<Double>? = null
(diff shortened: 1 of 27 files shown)


dkhalanskyjb:
I don't have any context for this, so I'll need some explanation.

1. Do I understand correctly that in 1.9.0, in the space of a single release, old atomics get deprecated and force everyone to immediately migrate to the new ones? Now that's harsh! Usually, we do it differently: first, the deprecated API causes warnings, and only later, when people have had enough time to migrate, the warning becomes an error. Was the speed of this change discussed and approved?
2. I don't know how `kotlin-community/dev` is built. Will it always be built with version 1.9.0+? Currently, the build does not pass with the old version, because, as you said, the new atomics are unavailable there.
3. Can this not be tweaked somehow creatively so that the build passes both on the current version and on 1.9.0+? For example, won't it work with `import kotlin.concurrent.*; import kotlin.native.concurrent.*` (or maybe if their order is switched)? Then we can just merge it to `develop`, dropping a comment to the import list to explain that the order and form of the imports are significant.

mvicsokolova:
Hi!)
1. No no, the deprecation was not that fast. Old atomics were deprecated with WARNING in 1.9.0, and now I've prepared the MR (https://jetbrains.team/p/kt/reviews/10650/timeline) for 1.9.20 where deprecation level is changed for ERROR. And this breaks the aggregate build, because there are usages of old atomics in coroutines.
2. `kotlin-community/dev` branch is used in aggregate builds, and uses 1.9.20-dev Kotlin currently. And having Kotlin set to 1.8.20 in all `kotlin-community/dev` branches is wrong. I asked our release managers to update Kotlin version in gradle.properties of `kotlin-community/dev` branches for all the libraries to 1.9.20-dev-*. So, after that the new atomics will be available.

mvicsokolova:
3. I tried adding imports of both kotlin.concurrent.* and kotlin.native.concurrent.*, this will work for develop where Kotlin is 1.8.20 and atomics will be resolved to kotlin.native.concurrent, because they are only present in that package. But for Kotlin 1.9.20-dev atomics are still resolved to the old kotlin.native.concurrent package and the build fails with deprecation error. But I'll check changing the order of packages.

dkhalanskyjb:
My only concern is that a branch should build with `./gradlew clean build` when you check it out. When I do it with `kotlin-community/dev` currently, the build seems to go through fine, but with this PR, it will no longer do so.

> I asked our release managers to update Kotlin version in gradle.properties of kotlin-community/dev branches for all the libraries to 1.9.20-dev-*.

Why can't we do it ourselves? Is it not just about updating the Kotlin version in a couple of places? For example, what's preventing us from upgrading Kotlin to something like 1.9.0-RC in this same PR?

qwwdfsad:
What should we do with this PR? Seems kind of abandoned

mvicsokolova:
I'll create a PR with the necessary changes in develop: in this case just swapping package imports (kotlin.concurrent and kotlin.native.concurrent) should help.

mvicsokolova:
So, here is the update on this PR: 
in this MR (https://jetbrains.team/p/kt/reviews/10650/timeline) atomics from kotlin.native.concurrent are deprecated with error for 1.9.20 -> the aggregate build fails.

In coroutines we need to migrate from old atomics to the new atomics from kotlin.concurrent package, but those are only available since 1.9.0

In order to add this commit to develop first I tried to import both kotlin.concurrent.* and kotllin.native.concurrent packages, thus the build of the develop branch passes but the aggregate still fails, because the old kotlin.native.concurrent.AtomicReference is resolved.

Is there a better solution for this, then just adding this commit to `kotlin-community`?

mvicsokolova:
I've reverted this [commit](https://github.com/Kotlin/kotlinx.coroutines/commit/6fb30e2420041b12ecee59b9d420e4a219b4b2df), we don't need it in this PR

qwwdfsad (approved):

woainikk (approved):

PR #3801, opened 2023-07-05
3789: Update flow.timeout example to re-throw

### LINKED ISSUE #3789: Flow.timeout documentation suggests generic `catch {}` operator.
The API documentation for `Flow<T>.timeout` suggests using a `catch{}` without rethrowing:

```
flow {
    emit(1)
    delay(100)
    emit(2)
    delay(100)
    emit(3)
    delay(1000)
    emit(4)
}.timeout(100.milliseconds).catch {
    emit(-1) // Item to emit on timeout
}.onEach {
    delay(300) // This will not cause a timeout
}
```

This will emit -1 for any and all upstream exceptions (admittedly none other possible in this fizz buzz example). Can we update the documentation to be more targeted?

Such as this.

```
flow {
    emit(1)
    delay(100)
    emit(2)
    delay(100)
    emit(3)
    delay(1000)
    emit(4)
}.timeout(100.milliseconds).catch { exception ->
    if (exception is TimeoutCancellationException) {
      // Catch the TimeoutCancellationException emitted above.
      emit(-1) // Item to emit on timeout
    } else {
      throw exception
    }
}.onEach {
    delay(300) // This will not cause a timeout
}
```

steve-the-edwards:
Fixes #3789

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -203,7 +203,7 @@ public fun <T> Flow<T>.debounce(timeout: (T) -> Duration): Flow<T> =
         timeout(emittedItem).toDelayMillis()
     }
 
-private fun <T> Flow<T>.debounceInternal(timeoutMillisSelector: (T) -> Long) : Flow<T> =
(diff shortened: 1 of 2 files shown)


steve-the-edwards:
cc @pablobaxter

qwwdfsad (approved):
Thanks!

PR #3804, opened 2023-07-12
Fix compilation with newer Kotlin version

nikitabobko:
The compilation started failing after KT-59739


nikitabobko:
Main MR in kotlin: https://jetbrains.team/p/kt/reviews/11038/timeline

Issue #3806, opened 2023-07-13
Feature Request: Introduce scanNotNull Function to Flow Extension


UdaraWanasinghe:
**Description:**
I would like to request the addition of a new function, `scanNotNull`, to the Flow extension in the Kotlin Coroutines library. The `scanNotNull` function would allow folding of flow values while emitting only the intermediate non-null results.

**Expected Behavior:**
The `scanNotNull` function would perform a fold operation on the flow, similar to `scan` or `runningFold`. However, it would skip emitting any intermediate results that are `null`. Only non-null results would be emitted in the resulting flow.

**Motivation:**
I utilized Kotlin flows to manage UI states by converting UI actions to corresponding states. A dedicated action flow collects the necessary data for performing operations, where each action includes the required data. I employed the `scan` function to accumulate the necessary data for the operation. Upon collecting all data points, the operation is performed and translated into the corresponding state. Throughout the data collection process, intermediate results are communicated to the UI via events. However, not all actions are updating the data. I want them to filter out inside the `scan` function. See usage here: [https://github.com/UdaraWanasinghe/webp-android/example/src/main/java/com/aureusapps/android/webpandroid/example/models/CodecViewModel.kt](https://github.com/UdaraWanasinghe/webp-android/blob/6f05448722ec762d15f5e9e1d9f4943ac2181016/example/src/main/java/com/aureusapps/android/webpandroid/example/models/CodecViewModel.kt#L49)

dkhalanskyjb:
Isn't your use case more straightforwardly solved via `scan` + `distinctUntilChanged`?

With your operator:
```kotlin
flow.scanNotNull(initialData) { data, value ->
  if (value.requiresChangingData()) {
    data.update(value)
  } else {
    null
  }
}
```

Without it:
```kotlin
flow.scanNotNull(initialData) { data, value ->
  if (value.requiresChangingData()) {
    data.update(value)
  } else {
    data
  }
}.distinctUntilChanged()
```

or, if `update` is smart enough to leave the object untouched when no update is needed, just
```kotlin
flow.scanNotNull(initialData) { data, value -> data.update(value) }.distinctUntilChanged()
```

qwwdfsad:
There is not much demand since the original report, so closing as won't fix as there is a clear semantically-meaningful workaround

PR #3807, opened 2023-07-13
Add "scanNotNull" function to Flow extension

UdaraWanasinghe:
Adds feature #3806
This adds the "scanNotNull" function as an extension to the Flow class. The "scanNotNull" function allows folding the flow's values using a provided operation and emitting every intermediate non-null result. It skips emitting any intermediate results that are null, ensuring only non-null results are included in the resulting flow.


qwwdfsad:
While the issue remains open, there does not seem to be a demand for this (or for the corresponding stdlib counterpart) feature, so we'll not proceed with this one

PR #3815, opened 2023-07-19
Make annotations on expect declarations comply with new compiler restriction

merfemor:
In KT-58551 we require all annotations from expect declaration to be present on actual.

This commit fixes the following violations:

1. `DefaultDelay` has `@PublishedApi` only on expect. But it is not used from public inline functions, so we can safely remove annotation.

2. expect `IgnoreJreRequirement` has `AnnotationTarget.TYPE` which is not present in actual. This is because Java target TYPE doesn't correspond to Kotlin target TYPE. Since `IgnoreJreRequirement` is internal, we can safely change targets on expect and make them equal to actual.

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineContext.common.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineContext.common.kt
@@ -20,7 +20,7 @@ public expect fun CoroutineScope.newCoroutineContext(context: CoroutineContext):
 @InternalCoroutinesApi
 public expect fun CoroutineContext.newCoroutineContext(addedContext: CoroutineContext): CoroutineContext
 
-@PublishedApi
+@PublishedApi // to have unmangled name when using from other modules via suppress
 @Suppress("PropertyName")
 internal expect val DefaultDelay: Delay
 
--- a/kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
+++ b/kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
@@ -40,7 +40,7 @@ internal expect suspend inline fun recoverAndThrow(exception: Throwable): Nothin
  * The opposite of [recoverStackTrace].
  * It is guaranteed that `unwrap(recoverStackTrace(e)) === e`
  */
-@PublishedApi // published for the multiplatform implementation of kotlinx-coroutines-test
+@PublishedApi // Used from kotlinx-coroutines-test and reactor modules via suppress, not part of ABI
 internal expect fun <E: Throwable> unwrap(exception: E): E
 
 internal expect class StackTraceElement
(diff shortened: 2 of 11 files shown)


kotlinx-coroutines-core/common/src/CoroutineContext.common.kt
```diff
@@ -20,7 +20,6 @@ public expect fun CoroutineScope.newCoroutineContext(context: CoroutineContext):
 @InternalCoroutinesApi
 public expect fun CoroutineContext.newCoroutineContext(addedContext: CoroutineContext): CoroutineContext
 
-@PublishedApi
```
qwwdfsad, line 23: Please leave it on actuals as well.
It is used not for the sake of inlining but in order to access it with the help of `@Suppress` from a different KLib. `@PublishedApi` guarantees a stable unmangled name here

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/CoroutineContext.common.kt
```diff
@@ -20,7 +20,7 @@ public expect fun CoroutineScope.newCoroutineContext(context: CoroutineContext):
...
+@PublishedApi // to have unmangled name when using from other modules
```
qwwdfsad, line 23:
```suggestion
@PublishedApi // to have unmangled name when using from other modules via suppress
```

kotlinx-coroutines-core/js/src/CoroutineContext.kt
```diff
@@ -33,6 +33,7 @@ private fun isJsdom() = jsTypeOf(navigator) != UNDEFINED &&
     jsTypeOf(navigator.userAgent.match) != UNDEFINED &&
     navigator.userAgent.match("\\bjsdom\\b")
 
+@PublishedApi
```
qwwdfsad, line 36:
```suggestion
@PublishedApi // Used from kotlinx-coroutines-test via suppress, not part of ABI
```

kotlinx-coroutines-core/common/src/internal/StackTraceRecovery.common.kt
```diff
@@ -40,6 +40,7 @@ internal expect suspend inline fun recoverAndThrow(exception: Throwable): Nothin
  * The opposite of [recoverStackTrace].
  * It is guaranteed that `unwrap(recoverStackTrace(e)) === e`
  */
+@PublishedApi // published for the multiplatform implementation of kotlinx-coroutines-test
```
qwwdfsad, line 43:
```suggestion
@PublishedApi // Used from kotlinx-coroutines-test and reactor modules via suppress, not part of ABI
```

qwwdfsad (commented):
Please merge the suggestions and it's good to go

mshishkina (approved):

PR #3816, opened 2023-07-19
Clarify documentation of Mutex.lock() behavior

obermillerk:
The previous documentation of the `Mutex.lock()` method was confusing to me. It seemed to imply that the call would suspend the entire time that the lock was held, when in fact it just suspends until the lock is acquired.


qwwdfsad (approved):
Thanks1

qwwdfsad (approved):
Thanks1

qwwdfsad (approved):
Thanks!

Issue #3658, opened 2023-03-06
CancellationException can be thrown not only to indicate cancellation


dkhalanskyjb:
https://betterprogramming.pub/the-silent-killer-thats-crashing-your-coroutines-9171d1e8f79b raises a good point: `CancellationException` should indicate that the coroutine in which it is raised is cancelled, but it's not always actually the case:

```kotlin
    fun testFoo() {
        val scope = CoroutineScope(Dispatchers.IO)
        val f = scope.async {
            delay(1000)
        }
        scope.cancel()
        var result = false
        runBlocking {
            launch {
                f.await() // throws CancellationException, silently stopping the `launch` coroutine
                result = true
            }.join()
        }
        check(result) { "issue" }
    }
```

Here, the programmer didn't make the mistake of catching a `CancellationException` and rethrowing it somewhere else, but still, the `CancellationException` leaked from a different scope.

dkhalanskyjb:
Need to dive deep into the implementation to see if this is possible, but it looks like modifying `Deferred` to throw a non-`CancellationException` may not be a good solution.

Consider this simplified scenario:
```kotlin
runBlocking {
  val result = async(Dispatchers.Default) {
    delay(5.seconds)
  }
  launch(Dispatchers.Unconfined) {
    result.await()
  }
  delay(1.seconds)
  cancel() // cancel the common scope
}
```
Here, we possibly have a race: the global scope is cancelled, which causes the coroutines in `async` and `launch` also to be cancelled. However, it can happen that this sequence of events happens before `launch` gets cancelled:
1. `async` gets cancelled,
2. the `Deferred` learns about the cancellation exception,
3. `await` finishes with a *non*-`CancellationException` in the `launch`,
4. which causes the `launch` to fail with a *non*-`CancellationException`,
5. with which the entire `runBlocking` fails!

So, the code can fail with one exception or rarely with another. This is just one simplified example of a broader class of races that we could introduce.

dovchinnikov:
`await` (probably another function since this one is taken) should return `kotlin.Result`
- if `await` resumes with CE, then it means the current coroutine was cancelled;
- if `await` resumes with `Result`, then `async` was completed before the current coroutine was cancelled, and it represents a value/CE/failure of the `async`.

dovchinnikov:
The above could be implemented as an extension function: 
```
suspend fun Deferred<T>.awaitResult() {
  return try {
    Result.success(await())
  }
[...]

PR #3817, opened 2023-07-20
JobSupport.kt: fix GH!3658

revintec:
https://github.com/Kotlin/kotlinx.coroutines/issues/3658


qwwdfsad (commented):
Thanks for the effort and great job with figuring out JobSupport internals.

Yet I'm afraid we cannot really accept the PR -- it is a breaking change with unknown scope, impact and consequences.

Such change requires a careful evaluation, vast documentation changes, extensive testing and user-project battle-testing, potentially with introduction of a global switch or any other deprecation mechanism

PR #3822, opened 2023-07-24
Fix the IDEA debugger

dkhalanskyjb:
PR #3746 introduced a bug due to which the IDEA coroutine debugger didn't run. See
https://youtrack.jetbrains.com/issue/KTIJ-26327

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
@@ -35,7 +35,9 @@ internal object DebugProbesImpl {
      * This internal method is used by the IDEA debugger under the JVM name
      * "isInstalled$kotlinx_coroutines_debug" and must be kept binary-compatible, see KTIJ-24102
      */
-    val isInstalled: Boolean get() = installations.value > 0
+    val isInstalled: Boolean
+        @JvmName("isInstalled\$kotlinx_coroutines_debug")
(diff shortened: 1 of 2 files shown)


dkhalanskyjb:
When I do `git checkout version-1.7.1`, `git checkout version-1.7.2`, and `git checkout fix-idea-debugger` and start some test, 1.7.1 and `fix-idea-debugger` behave seemingly identically, but `version-1.7.2` does not display any coroutines, instead replying "No coroutine information found." I think this means the fix did the trick.

kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
```diff
@@ -35,7 +35,9 @@ internal object DebugProbesImpl {
      * This internal method is used by the IDEA debugger under the JVM name
      * "isInstalled$kotlinx_coroutines_debug" and must be kept binary-compatible, see KTIJ-24102
      */
-    val isInstalled: Boolean get() = installations.value > 0
+    val isInstalled: Boolean
+        @JvmName("isInstalled\$kotlinx_coroutines_debug")
```
qwwdfsad, line 39:
```suggestion
        // IDEA depended on "internal val isInstalled", thus the mangling. Public + JvmName in order to make this getter part of the ABI
        @JvmName("isInstalled\$kotlinx_coroutines_debug")
```

qwwdfsad (approved):
Otherwise looks good

PR #3823, opened 2023-07-24
Rename ClosableCoroutineDispatcher to fix per-file compilation for JS…

JSMonk:
… target


qwwdfsad:
It would be nice to provide a bit more detailed explanation or mention a YT ticket

PR #3825, opened 2023-07-25
Conditionally remove native targets that are removed in 1.9.20

mvicsokolova:
This commit conditionally removes `iosArm32`, `watchosX86`  that are removed in 1.9.20.

This commit is necessary for the aggreagate builds.

A similar commit was added earlier to atomicfu: https://github.com/Kotlin/kotlinx-atomicfu/commit/7697fff37a68007a4c112c3321ef07f763990d37

### DIFF
--- a/buildSrc/src/main/kotlin/KotlinVersion.kt
+++ b/buildSrc/src/main/kotlin/KotlinVersion.kt
@@ -1,0 +1,14 @@
+@file:JvmName("KotlinVersion")
+
+fun isKotlinVersionAtLeast(kotlinVersion: String, atLeastMajor: Int, atLeastMinor: Int, atLeastPatch: Int): Boolean {
+    val (major, minor) = kotlinVersion
+        .split('.')
+        .take(2)
+        .map { it.toInt() }
+    val patch = kotlinVersion.substringAfterLast('.').substringBefore('-').toInt()
+    return when {
+        major > atLeastMajor -> true
+        major < atLeastMajor -> false
+        else -> (minor == atLeastMinor && patch >= atLeastPatch) || minor > atLeastMinor
+    }
+}
--- a/gradle/compile-native-multiplatform.gradle
+++ b/gradle/compile-native-multiplatform.gradle
@@ -1,9 +1,16 @@
+import static KotlinVersion.isKotlinVersionAtLeast
+
 /*
  * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 project.ext.nativeMainSets = []
 project.ext.nativeTestSets = []
+
+// TODO: this block should be removed when Kotlin version is updated to 1.9.20
+// Right now it is used for conditional removal of native targets that will be removed in 1.9.20 (iosArm32, watchosX86)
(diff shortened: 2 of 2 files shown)


buildSrc/src/main/kotlin/KotlinVersion.kt
```diff
@@ -0,0 +1,14 @@
+@file:JvmName("KotlinVersion")
+
+fun isKotlinVersionAtLeast(kotlinVersion: String, atLeastMajor: Int, atLeastMinor: Int, atLeastPatch: Int): Boolean {
```
mvicsokolova, line 3: Please correct me, if we already have a similar function somewhere, I just didn't find it.

qwwdfsad: We have `kotlin.KotlinVersion.CURRENT.isAtLeast()`

gradle/compile-native-multiplatform.gradle
```diff
@@ -1,10 +1,17 @@
+import static KotlinVersion.isKotlinVersionAtLeast
+
 /*
  * Copyright 2016-2021 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
  */
 
 project.ext.nativeMainSets = []
 project.ext.nativeTestSets = []
 
+// TODO: this block should be removed when Kotlin version is updated to 1.9.20
```
qwwdfsad, line 10: I suggest filing an issue for coroutines so we won't forget about it

buildSrc/src/main/kotlin/KotlinVersion.kt
mvicsokolova: Fixed

gradle/compile-native-multiplatform.gradle
mvicsokolova: https://github.com/Kotlin/kotlinx.coroutines/issues/3828

gradle/compile-native-multiplatform.gradle
```diff
@@ -1,10 +1,17 @@
...
+// Right now it is used for conditional removal of native targets that will be removed in 1.9.20 (iosArm32, watchosX86)
+// and is necessary for testing of Kotlin dev builds.
+def enableDeprecatedNativeTargets = !KotlinVersion.CURRENT.isAtLeast(1, 9, 20)
```
lukellmann, line 13: doesn't this test the version of gradle's embedded kotlin and not the kotlin version with which coroutines are built?

mvicsokolova: Mm, Im not sure about that: 
KotlinVersion.CURRENT `Returns the current version of the Kotlin standard library.`

qwwdfsad: It's easy to check -- update Kotlin version in our build and see if this property changes its value

qwwdfsad: Seems to be the way indeed; sorry for the confusion

qwwdfsad (approved):

mshishkina:
Commit is merged into kotlin-community/dev

PR #3826, opened 2023-07-25
Fix NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS compilation error

nikitabobko:
The compilation error appeared after KT-59665

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/TestBase.kt
+++ b/kotlinx-coroutines-core/jvm/test/TestBase.kt
@@ -54,6 +54,7 @@ public actual typealias TestResult = Unit
  * }
  * ```
  */
+@Suppress("NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS") // Counterpart for ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS
 public actual open class TestBase(private var disableOutCheck: Boolean)  {
 
     actual constructor(): this(false)
--- a/kotlinx-coroutines-core/jvm/test/TestBase.kt
+++ b/kotlinx-coroutines-core/jvm/test/TestBase.kt
@@ -54,7 +54,7 @@ public actual typealias TestResult = Unit
  * }
  * ```
  */
-@Suppress("NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS") // Counterpart for ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS
+@Suppress("NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS") // Counterpart for @Suppress("ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS")
 public actual open class TestBase(private var disableOutCheck: Boolean)  {
 
     actual constructor(): this(false)


nikitabobko:
MR in Kotlin: https://jetbrains.team/p/kt/reviews/11039/timeline

kotlinx-coroutines-core/jvm/test/TestBase.kt
```diff
@@ -54,6 +54,7 @@ public actual typealias TestResult = Unit
  * }
  * ```
  */
+@Suppress("NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS") // Counterpart for ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS
```
qwwdfsad, line 57: Could you please elaborate on how we should proceed further with this suppress?

If this case is no longer supported (or was unsupported but not explicitly reported), shouldn't we have any kind of migration plan?

nikitabobko: This suppress is a counterpart for `@Suppress("ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS")`. You can see that you already have a `@Suppress("ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS")` inside `public actual open class TestBase`.

If you didn't have `@Suppress("ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS")` then `NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS` wouldn't be reported

nikitabobko: Unfortunately, I had to rework the way we report `ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS`. Before my changes, it was a separate checker.  Now it's checked in the expect-actual matcher and it's considered a common expect-actual mismatch incompatibility

And since such types of incompatibilities are reported two times:
```kotlin
// Common
expect class Bar {
    fun foo(): Any
}

// Platform
actual class Bar { // NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS (first time)
    actual fun foo(): String { // ACTUAL_WITHOUT_EXPECT (second time)
        return ""
    }
}
```

Now you need to suppress `ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS` two times

nikitabobko: Probably it will be more clear If I change "Counterpart for ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS" comment to "Counterpart for @Suppress("ACTUAL_FUNCTION_WITH_DEFAULT_ARGUMENTS")"

nikitabobko: The full compilation error of `NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS` also makes it clear:
```
e: file:///home/bobko/jb/test-proj/kotlinx.coroutines/kotlinx-coroutines-core/jvm/test/TestBase.kt:57 Actual class 'TestBase' has no corresponding members for expected class members:

    public expect fun error(message: Any, cause: Throwable?): Nothing

    The following declaration is incompatible because actual function cannot have default argument values, they should be declared in the expected function:
        public actual fun error(message: Any, cause: Throwable? = null): Nothing
```

qwwdfsad (approved):

nikitabobko:
The review in Kotlin is done.

@qwwdfsad can you please merge it?

@mshishkina can you please cherry-pick it to `kotlin-community/dev`?

PR #3837, opened 2023-08-04
Explain potential memory leak that nulling BufferedChannelIterator.co…

qwwdfsad:
…ntinuation out prevents

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
@@ -1690,8 +1690,11 @@ internal open class BufferedChannel<E>(
         }
 
         fun tryResumeHasNextOnClosedChannel() {
-            // Read the current continuation and clean
-            // the corresponding field to avoid memory leaks.
+            /*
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
```diff
@@ -1690,8 +1690,12 @@ internal open class BufferedChannel<E>(
         }
 
         fun tryResumeHasNextOnClosedChannel() {
-            // Read the current continuation and clean
-            // the corresponding field to avoid memory leaks.
+            /*
```
ndkoval, line 1693: Please use `//` for comments in the code for consistency with the rest of the file.

ndkoval: You mention multiple waiters in this comment, but it is always a single one. I suggest the following text instead; please feel free to modify it as you wish:

```
Read the current continuation of the suspended `hasNext()` call and clean the corresponding field to avoid memory leaks. While this nulling out is unnecessary, it eliminates memory leaks (through the continuation) if the channel iterator accidentally remains GC-reachable after the channel is closed. 
```

ndkoval (changes requested):

PR #3847, opened 2023-08-11
Fix expect/actual mismatched member scope for `open expect` compilati…

nikitabobko:
…on errors

Those are compilation errors introduced in Kotlin 1.9.20 in scope of https://youtrack.jetbrains.com/issue/KT-22841

### DIFF
--- a/kotlinx-coroutines-core/concurrent/src/CompletionHandler.kt
+++ b/kotlinx-coroutines-core/concurrent/src/CompletionHandler.kt
@@ -6,12 +6,20 @@ package kotlinx.coroutines
 
 import kotlinx.coroutines.internal.*
 
+// fixme replace the suppress with AllowDifferentMembersInActual once stdlib is updated to 1.9.20 https://github.com/Kotlin/kotlinx.coroutines/issues/3846
+// New 'CompletionHandler` supertype is added compared to the expect declaration.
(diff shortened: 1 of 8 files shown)


nikitabobko:
Don't review it yet. The PR is in progress

nikitabobko:
Main review in kotlin: https://jetbrains.team/p/kt/reviews/11603/timeline

qwwdfsad:
>Don't review it yet. The PR is in progress.

You can use "Open draft pull request" and promote it to the PR when it's ready: 
image

nikitabobko:
Ready to review

qwwdfsad (commented):
Is it correct that each of the ~10 suppresses should be replaced with `AllowDifferentMembersInActual` in 1.9.20?

Feels like there are a bit too much of suppresses

nikitabobko:
> Is it correct that each of the ~10 suppresses should be replaced with AllowDifferentMembersInActual in 1.9.20?

Yes, it's correct

> Feels like there are a bit too much of suppresses

Yes, that's true. Unfortunately, we don't have any consistent design ready for the non-final expect class https://youtrack.jetbrains.com/issue/KT-22841

That's why we have no choice but to prohibit different member scopes for non-final `open` `expect class` and its appropriate `actual class`. We discussed that it will break a lot of code during the design meeting

qwwdfsad (approved):

PR #3850, opened 2023-08-14
Introduce workaround (in fact, fix the previously incorrect expect-ac…

qwwdfsad:
…tual mapping) for CancellationException.

See KT-61168 for the root cause & KT-22004 for the future solution

### DIFF
--- a/kotlinx-coroutines-core/native/src/Exceptions.kt
+++ b/kotlinx-coroutines-core/native/src/Exceptions.kt
@@ -15,6 +15,11 @@ import kotlinx.coroutines.internal.SuppressSupportingThrowableImpl
  */
 public actual typealias CancellationException = kotlin.coroutines.cancellation.CancellationException
 
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+@kotlin.internal.LowPriorityInOverloadResolution
+public actual fun CancellationException(message: String?, cause: Throwable?): CancellationException =
+    CancellationException(message, cause)
+
 /**
  * Thrown by cancellable suspending functions if the [Job] of the coroutine is cancelled or completed
  * without cause, or with a cause or exception that is not [CancellationException]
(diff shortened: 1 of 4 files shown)


kotlinx-coroutines-core/native/src/Exceptions.kt
```diff
@@ -15,6 +15,11 @@ import kotlinx.coroutines.internal.SuppressSupportingThrowableImpl
  */
 public actual typealias CancellationException = kotlin.coroutines.cancellation.CancellationException
 
+@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+@kotlin.internal.LowPriorityInOverloadResolution
+public actual fun CancellationException(message: String?, cause: Throwable?): CancellationException =
+    CancellationException(message, cause)
```
dkhalanskyjb, line 21: Is this `LowPriorityInOverloadResolution` just to make this definition work, or is it also so that platform-specific code chooses the other overload?
* If it's for the definition to work, won't a fully-qualified name work better? Like `= kotlin.coroutines.cancellation.CancellationException(message, cause)`.
* If it's for choosing the other overload, maybe we can ensure this also happens on K2 by marking this function as `inline`? Is this possible?

qwwdfsad: To make the declaration work (otherwise, it's conflicting overloads on the declaration level) and, later, the choice of the overload from our/users' code.

>maybe we can ensure this also happens on K2 by marking this function as inline?

It's a bit unclear to me what you've meant, could you please elaborate on that?

dkhalanskyjb: > It's a bit unclear to me what you've meant, could you please elaborate on that?

If I remember correctly, with K2, in common code, the overload that gets chosen in the end is the common one, so in this case, the `expect` function. If that's undesirable, can't we mark this function as `inline` instead so its calls always get converted to the ones we want?

qwwdfsad: I see, thanks.

>the overload that gets chosen in the end is the common one

Unfortunately, it will be the case only for the code in `kotlinx-coroutines-core` sourceset and not for our users, so I don't think it worth it

dkhalanskyjb (approved):
I can't say that I fully got it, but this seems fine.

dkhalanskyjb:
Ok, now I got it. It _is_ as good as it gets, I think.

Issue #2580, opened 2021-03-13
statefulTransform for flow


phiSgr:
Sorry for not being able to provide a real life use-case as I don't get to write Kotlin at work.

---

In this [SO question](https://stackoverflow.com/questions/66586852/kotlin-flow-missing-many-trivial-functions-like-any-distinct/66587779), OP asks for a `distinct` method for `Flow`.

As of opening this issue, the answer given is:

```kotlin
fun <T> Flow<T>.distinct(): Flow<T> = with(mutableSetOf<T>()) {
    this@distinct.filter{ add(it) }
}
```

It is subtly wrong, as the flow has one shared mutable state, the mutable set, that is reused over multiple collections.

---

I suggest the following HOF to be added
This is inspired by [statefulMapConcat](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/statefulMapConcat.html) from Akka Streams.

```kotlin
fun <T> Flow<T>.statefulTransform(
    transformSupplier: () -> suspend FlowCollector<R>.(value: T) -> Unit
): Flow<T> = flow {
    val transform = transformSupplier()
    collect { value ->
        return@collect transform(value)
    }
}
```

Then `distinct` can be written as

```kotlin
fun <T> Flow<T>.distinct(): Flow<T> = statefulTransform<T, T> {
    with(mutableSetOf<T>()) {
        { if (add(it)) emit (it) }
    }
}
```

BTW I cannot make `@BuilderInference` work with this.

qwwdfsad:
That's a great idea, we definitely will be looking towards that.
It fits perfectly with our idea of orthogonal and composable operators that can be used to express any other transformation, similar to `transform`, `transformWhile` and so on.

It seems that not only `distinct`, but also `distinctUntilChanged`, `scan`, `runningReduce` and others, which is a good sign.


Regarding inference, there is really inconvenient issue that may appear to be the blocker of the whole design: https://youtrack.jetbrains.com/issue/KT-45657

He-Pin:
@qwwdfsad @phiSgr Hi, A `statefulMap` is just got merged into akka and I think this would be a nice addition too.
refs: https://github.com/akka/akka/issues/31077

Is there any update about the issue @qwwdfsad linked? thanks.

He-Pin:
@qwwdfsad any update on this?

JakeWharton:
[...]

PR #3867, opened 2023-08-29
+flow Add Flow.statefulMap operator

He-Pin:
refs: https://github.com/Kotlin/kotlinx.coroutines/issues/2580

As the original Akka's `statefulMapConcat` can lost the state when the stream/flow is completed, so later, a `statefulMap` is introduced for optional emiting value when the stream/flow is completed.

This PR introduces:

1. Flow#statefulMap
2. Flow#statefulTransform

Difference in statefulMap

I made the `onCompletion` as the second parameter, which is different as it is in Akka, because I think most of the time, the optional last emition may not be needed, and can make use of the trailing lambda.

With the `onCompletion` then this operator can be to implement the `grouped` function without losing data, the default behavior aline with `mapAccumlate`.

Early draft, feedback is welcome, thanks.

### DIFF
--- a/kotlinx-coroutines-core/common/test/flow/operators/StatefulMapTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/StatefulMapTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
+import kotlin.test.*
+
+class StatefulMapTest : TestBase() {
+
+    private fun <T> Flow<T>.zipWithIndex(): Flow<Pair<T, Long>> = statefulMap(0L) { index, value ->
+        return@statefulMap Pair(index + 1L, Pair(value, index))
+    }
+
+    @Test
+    fun testStatefulMap() = runTest {
+        val flow = flow {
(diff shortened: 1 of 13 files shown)


kotlinx-coroutines-core/common/test/flow/operators/StatefulTransformTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private fun <T> Flow<T>.distinct(): Flow<T?> = statefulTransform {
+        val set = mutableSetOf<T>();
+        { value ->
+            if (set.add(value)) {
+                emit(value);
+            }
+        }
+    }
```
He-Pin, line 76: Use it to implement `distinct`

kotlinx-coroutines-core/common/test/flow/operators/StatefulMapTest.kt
```diff
@@ -0,0 +1,115 @@
...
+            if (acc.size < size) {
+                acc.add(value)
+            }
+            if (acc.size == size) {
+                val result = acc.toList()
+                acc.clear()
+                return@statefulMap Pair(acc, result)
+            }
+            return@statefulMap Pair(acc, emptyList<T>())
+        }.filterNot(List<T>::isEmpty)
```
He-Pin, line 60: Use it to impement `grouped`.

He-Pin:
wrong target branch, close now , https://github.com/Kotlin/kotlinx.coroutines/pull/3868

PR #3868, opened 2023-08-30
Stateful map and transform

He-Pin:
refs: https://github.com/Kotlin/kotlinx.coroutines/issues/2580

As the original Akka's statefulMapConcat can lost the state when the stream/flow is completed, so later, a statefulMap is introduced for optional emiting value when the stream/flow is completed.

This PR introduces:

- Flow#statefulMap
- Flow#statefulTransform

Difference in statefulMap

I made the onCompletion as the second parameter, which is different as it is in Akka, because I think most of the time, the optional last emition may not be needed, and can make use of the trailing lambda.

With the onCompletion then this operator can be to implement the grouped function without losing data, the default behavior aline with mapAccumlate.

Early draft, feedback is welcome, thanks.

And As the `statefulTransform` can lost state, which may not be a good idea to introduce ? or we can leave it with document mentioned about that?

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt
@@ -43,6 +43,16 @@ public inline fun <T, R> Flow<T>.transform(
     }
 }
 
+public inline fun <T, R> Flow<T>.statefulTransform(
+    @BuilderInference crossinline transformCreator: suspend () -> (suspend FlowCollector<R>.(value: T) -> Unit)
+): Flow<R> = flow {
+    val transform = transformCreator()
+    collect { value ->
+        // kludge, without it Unit will be returned and TCE won't kick in, KT-28938
+        return@collect transform(value)
+    }
+}
+
 // For internal operator implementation
 @PublishedApi
 internal inline fun <T, R> Flow<T>.unsafeTransform(
--- a/kotlinx-coroutines-core/common/test/flow/operators/StatefulMapTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/StatefulMapTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2016-2019 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.coroutines.flow
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
+import kotlin.test.*
+
+class StatefulMapTest : TestBase() {
+
+    private fun <T> Flow<T>.zipWithIndex(): Flow<Pair<T, Long>> = statefulMap(0L) { index, value ->
+        return@statefulMap Pair(index + 1L, Pair(value, index))
+    }
+
+    @Test
+    fun testStatefulMap() = runTest {
+        val flow = flow {
+            emit(1)
+            emit(2)
+            emit(3)
+        }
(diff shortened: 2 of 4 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt
```diff
@@ -43,6 +43,16 @@ public inline fun <T, R> Flow<T>.transform(
...
+public inline fun <T, R> Flow<T>.statefulTransform(
+    @BuilderInference crossinline transformCreator: suspend () -> (suspend FlowCollector<R>.(value: T) -> Unit)
+): Flow<R> = flow {
+    val transform = transformCreator()
+    collect { value ->
+        // kludge, without it Unit will be returned and TCE won't kick in, KT-28938
+        return@collect transform(value)
+    }
+}
+
```
He-Pin, line 55: The document is still missing, will update it later

kotlinx-coroutines-core/common/test/flow/operators/StatefulMapTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    fun testGrouped() = runTest {
+        val flow = flow {
+            emit(1)
+            emit(2)
+            emit(3)
+            emit(4)
+            emit(5)
+        }
+        assertEquals(listOf(listOf(1, 2), listOf(3, 4), listOf(5)), flow.grouped(2).toList())
+    }
```
He-Pin, line 72: With the help of `onCompletion`, the last element `5` is keeped as `List(5)`.

kotlinx-coroutines-core/common/test/flow/operators/StatefulTransformTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    fun testGrouped() = runTest {
+        val flow = flow {
+            emit(1)
+            emit(2)
+            emit(3)
+            emit(4)
+            emit(5)
+        }
+        assertEquals(listOf(listOf(1, 2), listOf(3, 4)), flow.grouped(2).toList())
+    }
```
He-Pin, line 67: As the `statefulTransform` don't know the termination, so the last element 5 which was keeped in mutable list is lost, and we only get `List(1,2), List(3,4)` in this case.

qwwdfsad (commented):
Thanks for the effort!

With stateful transformations in their current form, we are heavily approaching the readability boundary and the limit of curly braces in a single declaration.

E.g. the example from the tests

```
fun <T> Flow<T>.zipWithIndex(): Flow<Pair<T, Long>> = statefulMap({ 0L }) { index, value ->
    return@statefulMap Pair(index + 1L, Pair(value, index))
}
```

can be rewritten using already existing mechanisms:

```
fun <T> Flow<T>.zipWithIndex(): Flow<Pair<T, Long>> = flow {
    var state = 0L
    collect {
        emit(Pair(it, state++))
    }
}
```

While the proposed version is slightly shorter, it's arguably less readable and _not_ less error-prone: it's still easy to accidentally capture the wrong value.
Unnamed lambda-generator of the default value, as well as using pair for the control flow seem to obscure the readability while bringing not that many benefits: it doesn't 
encapsulate a non-trivial implementation detail, it doesn't help readability much, it introduces one more way to do the same thing, and chaining such operators in place 
does not help the declarativity of the operators chain.
Apart from that, it introduces a new pattern (`operator({ state }) { _op_}`) into the core library, and this is the territory where we want to be extra-safe.

All in all, we are not ready to accept the CR in its current API shape

He-Pin:
@qwwdfsad Thanks for the review, what about the `statefulTransform` one?

PR #3871, opened 2023-08-31
Support expected/actual deprecation level limit from KT-60523

ejller:
Required changes for https://youtrack.jetbrains.com/issue/KT-60523


kotlinx-coroutines-core/common/src/Dispatchers.common.kt
```diff
@@ -74,4 +74,33 @@ public expect object Dispatchers {
...
+     * In the following example
+     * ```
+     * // 100 threads for MySQL connection
+     * val myMysqlDbDispatcher = Dispatchers.IO.limitedParallelism(100)
+     * // 60 threads for MongoDB connection
+     * val myMongoDbDispatcher = Dispatchers.IO.limitedParallelism(60)
+     * ```
+     * the system may have up to `64 + 100 + 60` threads dedicated to blocking tasks during peak loads,
+     * but during its steady state there is only a small number of threads shared
+     * among `Dispatchers.IO`, `myMysqlDbDispatcher` and `myMongoDbDispatcher`
```
qwwdfsad, line 103: @dkhalanskyjb I would like to double-check it with you.

The problem is, due to KT-60523 and the absence of `expect` refinements, we have to lift `Dispatchers.IO` to the `common` members.
It doesn't make sense on JS though, but it seems okay to either throw an exception "you are doing something wrong here" or return a `Default`. Both options are fine with me, I'm inclined to think "exception" is a safer bet (that we can change later).

Also, to avoid documentation twisting, I'd suggest just adding a paragraph "JS implementation: ..." without changing anything else.

WDYT?

ejller: I'm concerned that a js member access error will be thrown at runtime. Seems it might be unexpected for an user, especially if it's code from common modules that runs on the js platform

qwwdfsad: That's one way to look at that.

The other one is that such a scenario implies two actions taken:

* A developer has used `Dispatchers.IO` with pretty-specific semantics in the place where they shouldn't have done so
* This code was never run on the QA/developer/staging machine, so the exception manifested itself on a user's side

The rationale behind the exception is to signal the developer, "hey, you are using something with pretty tailored semantics on a platform where it's not supported, are you sure?" with a clear workaround `Dispatchers.IoOnJvmAndNativeAndDefaultOnJs` with a bit prettier name

dkhalanskyjb: First of all, I think the KT-60523 itself is a problem. I think it's completely reasonable to have a good enough common-denominator solution in common code that's deprecated in platform code when a better solution exists, or some granular common API that introduces concepts that are not distinguishable on every platform. Case in point: `Default` vs `IO`. Even if we introduce `Dispatchers.IO` to JS, deprecating its `actual` version would, I think, be a good decision: it just doesn't make sense to call `Dispatchers.IO` for the platform JS code. So, I am against the new language limitation.

To answer the question directly:

* Who would benefit from `IO = Default`: anyone who uses `Dispatchers.IO` in common code to indicate "I'd like to use the thread context where I/O is supposed to happen," which on JS, is just the one thread where everything happens.
* Who would benefit from throwing an exception: a person who expects some specific behavior from `Dispatchers.IO` (what behavior? multithreading?) and would introduce some bug if this behavior was not there. I don't see this as a viable concern: what's the worst that could happen? A deadlock due to some particularly structured common code? There are plenty of ways to be burned by JS being single-threaded, `Dispatchers.IO` is, I think, comparatively insignificant.

My vote is for `IO = Default`.

PR #3872, opened 2023-08-31
Mute "expect/actual classes are experimental" warning

nikitabobko:
This warning is introduced in Kotlin 1.9. KT-61573 If I don't mute this warning then the build fails because of `-Werror`


nikitabobko:
Main MR in Kotlin https://jetbrains.team/p/kt/reviews/11969/timeline

woainikk (approved):

qwwdfsad (approved):

PR #3877, opened 2023-09-04
Create Dispatcher.Default threads with the same context classloader a…

### LINKED ISSUE #3832: On JVM worker threads retain context class loader
**Describe the bug**

If a thread happens to start from a thread with different context class loader, then the class loader is retained for an unforeseeable time (at least for `IDLE_WORKER_KEEP_ALIVE_NS`, may be longer under load). In multi-loader environment this causes a leak.

**Provide a Reproducer**

```

class MyClassLoader : ClassLoader()

fun main() {
  val custom = MyClassLoader()
  runBlocking {
    val current = Thread.currentThread().contextClassLoader
    try {
      Thread.currentThread().contextClassLoader = custom
      launch(Dispatchers.Default) {} // spawn new thread
    }
    finally {
      Thread.currentThread().contextClassLoader = current
    }
  }
  println("A place for breakpoint")
}
```

image

qwwdfsad:
…s the dispatcher itself

In order to properly operate in modularized on a classloader level environments with the absence of other workarounds (i.e. supplying application-specific thread factory)

Fixes #3832

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
+++ b/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
@@ -135,6 +135,12 @@ internal actual object DefaultExecutor : EventLoopImplBase(), Runnable {
     private fun createThreadSync(): Thread {
         return _thread ?: Thread(this, THREAD_NAME).apply {
             _thread = this
+            /*
+             * `DefaultExecutor` is a global singleton that creates its thread lazily.
+             * To isolate the classloaders properly, we are inherting the context classloader from
+             * the singleton itself instead of using parent' thread one
+             * in order not to accidentally capture temporary application classloader.
+             */
             contextClassLoader = this@DefaultExecutor.javaClass.classLoader
             isDaemon = true
             start()
--- a/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
+++ b/kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
@@ -135,6 +135,7 @@ internal actual object DefaultExecutor : EventLoopImplBase(), Runnable {
     private fun createThreadSync(): Thread {
         return _thread ?: Thread(this, THREAD_NAME).apply {
             _thread = this
+            contextClassLoader = this@DefaultExecutor.javaClass.classLoader
             isDaemon = true
             start()
         }
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -593,6 +593,13 @@ internal class CoroutineScheduler(
     internal inner class Worker private constructor() : Thread() {
         init {
             isDaemon = true
+            /*
+             * `Dispatchers.Default` is used as *the* dispatcher in the containerized environments,
+             * isolated by their own classloaders. Workers are populated lazily, thus we are inheriting
+             * `Dispatchers.Default` context class loader here instead of using parent' thread one
+             * in order not to accidentally capture temporary application classloader.
+             */
+            contextClassLoader = this@CoroutineScheduler.javaClass.classLoader
         }
 
         // guarded by scheduler lock, index in workers array, 0 when not in array (terminated)


qwwdfsad:
Unfortunately, no `FieldWalker` test for that because Java prohibits reflective access to `java.*` classes

dkhalanskyjb:
```sh
$ git grep isDaemon
kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt:            isDaemon = true
kotlinx-coroutines-core/jvm/src/ThreadPoolDispatcher.kt:        t.isDaemon = true
kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt:        weakRefCleanerThread = thread(isDaemon = true, name = "Coroutines Debugger Cleaner") {
kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt:            isDaemon = true
kotlinx-coroutines-core/jvm/test/knit/ClosedAfterGuideTestExecutor.kt:        isDaemon = true
kotlinx-coroutines-debug/src/junit/CoroutinesTimeoutImpl.kt:    val testThread = Thread(testResult, "Timeout test thread").apply { isDaemon = true }
```
We have several other places where we create daemon threads; isn't the fix also needed for them? What makes `Dispatchers.Default` special? Sure, it *is* a global singleton, but so is `DebugProbesImpl.weakRefCleanerThread`, and so can be, depending on the user's code, `newFixedThreadPoolContext`. What if someone accidentally creates the `DefaultExecutor` thread in an environment with its own classloader?

qwwdfsad:
>What if someone accidentally creates the DefaultExecutor thread in an environment with its own classloader?

That's totally okay and a supposed use pattern. E.g. having `kotlinx-coroutines` dependency in Gradle plugin typically means it's loaded by a custom classloader.

The problem is not about daemon threads or classloaders per see, but about a pretty specific case:

1) The target in question should be a globally accessible application-level singleton. Control of everything else and their semantics (e.g. user-defined `newFixedThreadPoolContext`) is users' consideration
2) The target should create its own threads lazily from a thread different from where the target itself was initialized.

```
// T1, main app classloader
Dispatchers.Default.doSomething() 
```
Loads and initializes Dispatchers.Default object, its classloader is app classloader

```
// T2, app plugin classloader which is loaded and unloaded dynamically as a child of app classloader
Dispatchers.Default.dispatch(...) 
```

Here, due to the nature of our dispatcher, a new thread is spawned. By default, new thread inherits `context classloader` from the enclosing thread, which is, in our case, `app plugin classloader`.

After that, we have a situation where `Dispatchers.Default` that is meant to be shared across all "plugins" retains a reference to the child classloaders and prevents their unloading.

>isn't the fix also needed for them?

Nope:

* `newFixedThreadPoolContext` is a Java's executor with a supplied thread factory. We correctly inherit the semantics of `java.lang.Executors.newScheduledThreadPool`.
* `DefaultExecutor` -- starts its thread in the same context where it is initiated (shutdown is used only in our own tests)
* `CoroutinesTimeoutImpl` the thread is bound to the same context rule is


Does it make sense? Does it answer your question?

dkhalanskyjb:
> `DefaultExecutor` -- starts its thread in the same context where it is initiated

By "the same context", do you mean it's in the same thread? If so, I disagree.

The thread starts when the `thread` variable is accessed, which may happen after `DefaultExecutor` already finished initialization. Running tests in a debugger, I see that the thread only gets created when `dispatch` or `invokeOnTimeout` is called on the `DefaultExecutor`.

The initialization of `DefaultExecutor` may happen much earlier. For example, it's possible to ensure its initialization by calling `Dispatchers.IO.limitedParallelism(100)`. Then, by definition of `LimitedDispatcher`, it will access and initialize `DefaultDelay`. We clearly can call `Dispatchers.IO.limitedParallelism(100)` in another thread before any dispatches actually happen.

Also, what about `weakRefCleanerThread`? It does seem to tick both of the boxes you listed.

qwwdfsad:
>If so, I disagree.

You are right, this is my oversight, I was quite sure that `ensureStarted` is invoked immediately after the initialization. Thanks for the example with initialization, this is indeed the erroneous pattern. 

>Also, what about weakRefCleanerThread? It does seem to tick both of the boxes you listed.

I'm not really sure about this one.
On the one hand, this one has another property that it has to be launched (`DebugProbes.install()`) and shut down (`DebugProbes.uninstall()`) explicitly.

On the other hand, I can craft an example that leads to the leak, though it should be really obscure. (Initialize in one place, install in the other, then do not uninstall expecting that the debugger will be shared across all the application).
I would prefer avoid touching it,

kotlinx-coroutines-core/jvm/src/DefaultExecutor.kt
```diff
@@ -135,6 +135,7 @@ internal actual object DefaultExecutor : EventLoopImplBase(), Runnable {
     private fun createThreadSync(): Thread {
         return _thread ?: Thread(this, THREAD_NAME).apply {
             _thread = this
+            contextClassLoader = this@DefaultExecutor.javaClass.classLoader
```
dkhalanskyjb, line 138: I think this also needs a comment.

dkhalanskyjb (approved):

Issue #3754, opened 2023-05-14
The calling context impacts whether a mutex or a semaphore are locked on Kotlin/JS


CLOVIS-AI:
My code never interacts with a mutex in any other way than through `withLock`. My understanding is that it is not possible for the code to fail with "`Mutex is locked by <UNLOCKED>`". Yet, on Kotlin/JS, it does, while the same code runs fine on Kotlin/JVM.

## Describe the bug

Code that only ever uses `withLock` fails with "`Mutex is locked by <UNLOCKED>`" on Kotlin/JS, but runs with no issues on Kotlin/JVM.

To reproduce, use:
- Kotlin 1.8.20
- KotlinX.Coroutines 1.6.4 or 1.7.0
- Arrow Core 1.2.0-RC

Add the imports:
```kotlin
import arrow.core.raise.nullable
import kotlinx.coroutines.sync.Mutex
import kotlinx.coroutines.sync.Semaphore
import kotlinx.coroutines.sync.withLock
import kotlinx.coroutines.sync.withPermit
import kotlinx.coroutines.test.runTest
import kotlin.test.assertEquals
```

### 1. Mutex

The following code crashes on Kotlin/JS but runs on Kotlin/JVM.

```kotlin
runTest {
	val lock = Mutex()
	val data = HashMap<Int, Unit>()

	assertEquals(
		null,
		nullable {
			val counter = lock.withLock("user") { data[2] }

			if (counter != null)
				Unit
			else
				raise(null)
		},
	)
}
```

<details><summary>Stacktrace for 1.6.4</summary>
<p>

```
IllegalStateException: Mutex is locked by <UNLOCKED> but expected user
	at protoOf.unlock_uksyr8(/tmp/_karma_webpack_490441/commons.js:62756)
	at protoOf.doResume_5yljmg(/tmp/_karma_webpack_490441/commons.js:3846)
	at protoOf.invoke_qflhgo(/tmp/_karma_webpack_490441/commons.js:3801)
	at slambda.$testBody(/tmp/_karma_webpack_490441/commons.js:3935)
	at protoOf.doResume_5yljmg(/tmp/_karma_webpack_490441/commons.js:64987)
	at protoOf.invoke_tr8wvu(/tmp/_karma_webpack_490441/commons.js:64973)
	at <global>.l(/tmp/_karma_webpack_490441/commons.js:65020)
[...]

PR #3881, opened 2023-09-06
Workaround for KT-58685

CLOVIS-AI:
Since at least Kotlin 1.8.0, the Kotlin/JS IR compiler miscompiles a specific combination of a caller of an `inline` function containing a `finally` block: the `finally` block is called twice. This is rare, but can happen when a user calls `Mutex.withLock` or `Semaphore.withPermit`.

I originally reported this issue to the library (https://github.com/Kotlin/kotlinx.coroutines/issues/3754) without any idea how to reproduce it. @dkhalanskyjb managed to create a minimal reproducer and created a ticket against the Kotlin compiler ([KT-58685](https://youtrack.jetbrains.com/issue/KT-58685)). After discussing on the Kotlin Slack ([thread](https://kotlinlang.slack.com/archives/C0B8L3U69/p1693863899538249)), @rjaros found the workaround contained in this PR. Thanks to @JSMonk as well for attempting to fix this on the compiler side.

I added a unit test that reproduces the bug on the JS platform for both `Mutex.withLock` and `Semaphore.withPermit`. I then applied the workaround to both sources, ensuring it fixes both tests. Until a proper fix is found for the compiler, this should ensure users of the library do not have to deal with this. I believe the workaround has no impact on the library, though of course I'm open to your feedback.

For a more real-life example, this has been blocking the release of the [Pedestal](https://gitlab.com/opensavvy/pedestal) library for a long time now ([#101](https://gitlab.com/opensavvy/pedestal/-/issues/101)) because it appears in our cache implementation.

### DIFF
--- a/kotlinx-coroutines-core/common/src/sync/Mutex.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Mutex.kt
@@ -122,12 +122,18 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
         callsInPlace(action, InvocationKind.EXACTLY_ONCE)
     }
 
+    // Cannot use 'finally' in this function because of KT-58685
+    // See kotlinx.coroutines.sync.MutexTest.testWithLockJsMiscompilation
+
     lock(owner)
-    try {
-        return action()
-    } finally {
+    val result = try {
+        action()
+    } catch (e: Throwable) {
         unlock(owner)
+        throw e
     }
+    unlock(owner)
+    return result
 }
 
 
--- a/kotlinx-coroutines-core/common/test/sync/MutexTest.kt
+++ b/kotlinx-coroutines-core/common/test/sync/MutexTest.kt
@@ -148,4 +148,20 @@ class MutexTest : TestBase() {
         assertFailsWith<IllegalStateException> { mutex.lock(owner) }
         assertFailsWith<IllegalStateException> { select { mutex.onLock(owner) {} } }
     }
+
+    @Test
+    fun testWithLockJsMiscompilation() = runTest {
+        // This is a reproducer for KT-58685
+        // On Kotlin/JS IR, the compiler miscompiles calls to 'unlock' in an inlined finally
+        // This is visible on the withLock function
+        // Until the compiler bug is fixed, this test case checks that we do not suffer from it
+        val mutex = Mutex()
+        assertFailsWith<IndexOutOfBoundsException> {
+            try {
+                mutex.withLock { null } ?: throw IndexOutOfBoundsException() // should throw…
+            } catch (e: Exception) {
+                throw e // …but instead fails here
+            }
+        }
+    }
 }
--- a/kotlinx-coroutines-core/common/src/sync/Semaphore.kt
+++ b/kotlinx-coroutines-core/common/src/sync/Semaphore.kt
@@ -83,12 +83,18 @@ public suspend inline fun <T> Semaphore.withPermit(action: () -> T): T {
         callsInPlace(action, InvocationKind.EXACTLY_ONCE)
     }
 
+    // Cannot use 'finally' in this function because of KT-58685
+    // See kotlinx.coroutines.sync.SemaphoreTest.testWithPermitJsMiscompilation
+
     acquire()
-    try {
-        return action()
-    } finally {
+    val result = try {
+        action()
+    } catch (e: Throwable) {
         release()
+        throw e
     }
+    release()
+    return result
 }
 
 @Suppress("UNCHECKED_CAST")
--- a/kotlinx-coroutines-core/common/test/sync/SemaphoreTest.kt
+++ b/kotlinx-coroutines-core/common/test/sync/SemaphoreTest.kt
@@ -168,4 +168,20 @@ class SemaphoreTest : TestBase() {
         assertFailsWith<IllegalArgumentException> { Semaphore(1, -1) }
         assertFailsWith<IllegalArgumentException> { Semaphore(1, 2) }
     }
+
+    @Test
+    fun testWithPermitJsMiscompilation() = runTest {
+        // This is a reproducer for KT-58685
+        // On Kotlin/JS IR, the compiler miscompiles calls to 'release' in an inlined finally
+        // This is visible on the withPermit function
+        // Until the compiler bug is fixed, this test case checks that we do not suffer from it
+        val semaphore = Semaphore(1)
+        assertFailsWith<IndexOutOfBoundsException> {
+            try {
+                semaphore.withPermit { null } ?: throw IndexOutOfBoundsException() // should throw…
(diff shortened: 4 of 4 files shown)


kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -123,11 +123,14 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
-        return action()
-    } finally {
+    val result = try {
+        action()
+    } catch (e: Throwable) {
         unlock(owner)
+        throw e
     }
+    unlock(owner)
+    return result
```
dkhalanskyjb, line 133: I'm sure you'll agree that this is an unnatural way to write this. You shouldn't trust me to remember not to "fix" the style of code accidentally unless you drop a comment!

kotlinx-coroutines-core/common/test/sync/MutexTest.kt
```diff
@@ -148,4 +148,20 @@ class MutexTest : TestBase() {
...
+        // Until the compiler bug is fixed, this test case checks that we do not suffer from it
+        val mutex = Mutex()
+        assertFailsWith<IndexOutOfBoundsException> {
+            try {
+                mutex.withLock { null } ?: throw IndexOutOfBoundsException() // should throw…
+            } catch (e: Exception) {
+                throw e // …but instead fails here
+            }
+        }
+    }
```
dkhalanskyjb, line 166: I think we need a complementary test, like `testJsMiscompilationIsStillHappening`, that would check that with an alternative, naturally-written definition of `withLock`, anomalous behavior *does* happen. This way, when the bad behavior stops, the test will fail and we'll know that the workaround is no longer needed. This will also ensure that `testWithLockJsMiscompilation` doesn't just work accidentally, we'll have a clear distinction between "this version clearly works" and "that version clearly doesn't." Does this make sense?

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/sync/Mutex.kt
CLOVIS-AI: I agree, that's why I added the test. I'm not sure at all what to write in the comment other than "see test […] for an explanation of why it's an issue"

kotlinx-coroutines-core/common/test/sync/MutexTest.kt
CLOVIS-AI: If we add this, won't it means that whenever it is fixed on the compiler side, it will break CI until someone rolls back this PR?

I'll try to do this using the reproducer from the issue.

CLOVIS-AI: Actually, thinking back on this, I'm not sure it's so easy. This function is inline. The bug happens at the call site.

Therefore, even if the KotlinX.Coroutines team uses a compiler that doesn't have the issue anymore, a user of the library could still be using an older compiler that has the issue, right?

kotlinx-coroutines-core/common/src/sync/Mutex.kt
CLOVIS-AI: Added a comment.

CLOVIS-AI:
Bump @dkhalanskyjb, in case my previous messages got lost somewhere (if they didn't, and you're just busy at the moment, sorry for the ping)

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -122,12 +122,18 @@ public suspend inline fun <T> Mutex.withLock(owner: Any? = null, action: () -> T
...
     lock(owner)
-    try {
-        return action()
-    } finally {
+    val result = try {
+        action()
+    } catch (e: Throwable) {
         unlock(owner)
+        throw e
     }
```
dkhalanskyjb, line 134: To recreate the behavior of `finally`, if `unlock` inside `catch` throws an exception, this exception must add `e` as a suppressed one. Could you also write the corresponding test?

kotlinx-coroutines-core/common/test/sync/MutexTest.kt
dkhalanskyjb: Yeah, the point about CI makes sense, given that the issue is already fixed on the compiler's side, so we don't need a notification.

Regarding the old compiler + new library version: we don't provide guarantees as to what happens when using a library compiled with a newer compiler than your code uses. Generally, it's error-prone.

So, let's not add the tests. I wrote one just to make sure the tests that you added *do* correctly show that there's no problem, and sure enough, your tests do invoke the miscompilation if we keep the old form of `withLock`/`withPermit`. Nice work!

```kotlin
class SemaphoreTest: TestBase() {
    @Test
    fun testWithPermitJsMiscompilation() = runTest {
        // This is a reproducer for KT-58685
        // On Kotlin/JS IR, the compiler miscompiles calls to 'release' in an inlined finally
        // This is visible on the withPermit function
        // Until the compiler bug is fixed, this test case checks that we do suffer from it
        val semaphore = Semaphore(1)
        assertFailsWith<IllegalStateException> {
            try {
                semaphore.withPermitMiscompiled { null } ?: throw IndexOutOfBoundsException() // should throw…
            } catch (e: Exception) {
                throw e // …but instead fails here
            }
        }
    }
}

suspend inline fun <T> Semaphore.withPermitMiscompiled(action: () -> T): T {
    acquire()
    try {
        return action()
    } finally {
        release()
    }
}
```

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/sync/Mutex.kt
CLOVIS-AI: Hi, sorry for the time it took me to get back to this.

I'm not sure exactly what you mean here, something like this?
```kotlin
try {
    action()
} catch (e: Throwable) {
    try {
        unlock(owner)
    } catch (suppressed: Throwable) {
        e.addSuppressed(suppressed)
    }
    throw e
}
```

dkhalanskyjb: Oh, sorry, I was mistaken. There's no such behavior currently, and on second thought, I don't think there should be.

dkhalanskyjb (approved):
Good job, thank you!

lppedd:
Haven't followed the entire discussion, but since [KT-58685](https://youtrack.jetbrains.com/issue/KT-58685) is now fixed, and will probably be available in 1.9.21, what's the purpose if this workaround? To keep it working in previous releases?

dkhalanskyjb:
Yes, for a while. In practice, some people use the newer versions of the library without upgrading the compiler.

Amejonah1200:
This "workaround" breaks non-local returns, as reported in #3985 !

PR #3890, opened 2023-09-19
Reduce the overall time spent on stress tests in stress mode

qwwdfsad:
For now, it takes up to 8 hours, gobbling our TC agent pool and failing by timeout


kotlinx-coroutines-core/jvm/test/channels/BroadcastChannelMultiReceiveStressTest.kt
```diff
@@ -155,4 +160,10 @@ class BroadcastChannelMultiReceiveStressTest(
             if (stop) break
         }
     }
+
+    @Suppress("UNUSED_PARAMETER")
+    private fun println(debugMessage: String) {
+        // Uncomment for local debugging
+        // kotlin.io.println(debugMessage)
```
dkhalanskyjb, line 167: Uncommenting this leads to a crash. The correct version is `println(debugMessage as Any?)`.

PR #3891, opened 2023-09-19
Suppress forgotten error about actual classifier scope mismatch

dsavvinov:
Code red in IDE on IU-232.9921.47, Kotlin IDE Plugin 232-1.9.20-Beta-release-242

Would be nice to merge it, because IDE Integration tests in IJ check exactly this file and they're failing at the moment.

### DIFF
--- a/kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
+++ b/kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
@@ -372,6 +372,9 @@ public actual open class LockFreeLinkedListHead : LockFreeLinkedListNode() {
 
     // optimization: because head is never removed, we don't have to read _next.value to check these:
     override val isRemoved: Boolean get() = false
+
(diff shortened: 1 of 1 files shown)


qwwdfsad:
@nikitabobko could you please take a look? I'm surprised the build passes (e.g. the code compiles with -Werror) while IDE fails and want to double check here

nikitabobko:
The first thing to notice: you can see that the same warning is already suppressed in the supermethod `actual LockFreeLinkedListNode.nextIfRemoved`.

Different results in IDE and compiler happen because when we analyze common code, the K1 compiler sees platform declarations as well, but the K1 IDE doesn't see platform declarations. And when we compare member scopes you get one result or another, because of that.

In K2, we will fix that, because IDE and compiler will be consistent. They both won't see platform declarations during common code analysis.

I suggest accepting Dima's PR because it suppresses the error that will happen in K2.

nikitabobko (approved):

nikitabobko:
> I suggest accepting Dima's PR because it suppresses the error that will happen in K2.

Or if it's not the blocker for IDE tests then we can reject this PR until we really need it for K2

qwwdfsad (approved):

PR #3900, opened 2023-10-02
Replace `!==` with `!=` when comparing `Int?` with `null` in `ChannelsLincheckTest`

lunakoly:
^KT-62143 Fixed


lunakoly:
Closing this PR, because it was decided to discuss this case at the meeting on Oct 9

PR #3907, opened 2023-10-05
Update SafeCollector.common.kt

hoangchungk53qx1:
Tweak and clarify Flow exception transparency is violated


kotlinx-coroutines-core/common/src/flow/internal/SafeCollector.common.kt
```diff
@@ -67,6 +67,8 @@ internal fun SafeCollector<*>.checkContext(currentContext: CoroutineContext) {
         if (emissionParentJob !== collectJob) {
             error(
                 "Flow invariant is violated:\n" +
+                        "\t\tCold flows doesn't allow to emit() concurrently.\n" +
```
kevincianfarini, line 70: A `channelFlow` is still a cold flow, it's just one that has an underlying channel created when collection starts. "Cold flow don't allow..." is misleading.

PR #3909, opened 2023-10-06
Ensure `runTest` unsubscribes from the exception handler

### LINKED ISSUE #3897: A group of tests with mixed usage of `runTest` can cause silent failures
**Describe the bug**

When using `runTest` it is possible for a callback to remain in the `ExceptionCollector` after the test completes. This can then cause future test runs to pass with a swallowed exception in certain cases.

**Provide a Reproducer**

This behavior was found when migrating a large code base to use `runTest` for all unit tests involving coroutines with version `1.7.0+` of the coroutines test library. Currently there is a mix of coroutine tests that use `runTest`, `runBlocking`, or neither. To help find silent failures in tests that are not yet using `runTest` an `UncaughtExceptionHandler` is used so the tests fail. When running all tests we were unexpectedly finding a different number of test failures each time. Below is a simplified sequence of tests that reproduces the issue with explanation. The final test is expected to fail but it currently passes:

1. The first test uses `runTest` to cause the `ExceptionCollector` to become enabled. This test is expected to pass and it does.
2. The second test uses `runBlocking` and throws an uncaught exception in a coroutine. An unprocessed exception is added to the `ExceptionCollector` and the test also fails as expected due to the `UncaughtExceptionHandler`.
3. This runs the same test from step 1 again. This time though the test fails as expected due to an `UncaughtExceptionsBeforeTest` exception. However, what is not expected is after this test fails an exception callback remains in the `ExceptionCollector`.
4. This runs the same test from step 2 again except now it passes when _it is expected to fail_. Because of the callback still in the `ExceptionCollector` it seems the `UncaughtExceptionHandler` is never called since the exception is considered handled.

```
@TestMethodOrder(value = MethodOrderer.MethodName::class)
class CoroutineTests {
    private var defaultUncaughtExceptionHandler: Thread.UncaughtExceptionHandler? = null
    private val testDispatcher: TestDispatcher = UnconfinedTestDispatcher()

    @BeforeEach
    fun beforeEach() {
        defaultUncaughtExceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
        Thread.setDefaultUncaughtExceptionHandler { _, e -> throw e }
        Dispatchers.setMain(testDispatcher)
    }

    @AfterEach
    fun afterEach() {
        Thread.setDefaultUncaughtExceptionHandler(defaultUncaughtExceptionHandler)
        Dispatchers.resetMain()
    }

    @Test
    fun `1 use runTest so ExceptionCollector becomes enabled`() = runTest(testDispatcher) {
        assertTrue(true)
    }

    @Test
    fun `2 use runBlocking and throw uncaught exception that is added to unprocessed exceptions`() = runBlocking {
        CoroutineScope(testDispatcher).launch { throw Exception("Fail") }.join()
    }

    @Test
    fun `3 use runTest again and UncaughtExceptionsBeforeTest is thrown`() = runTest(testDispatcher) {
        assertTrue(true)
    }

    @Test
    fun `4 use runBlocking and throw uncaught exception that is now ignored when it should not be`() = runBlocking {
        CoroutineScope(testDispatcher).launch { throw Exception("Fail") }.join()
    }
}
```

dkhalanskyjb:
Fixes #3897

### DIFF
--- a/kotlinx-coroutines-test/common/test/RunTestTest.kt
+++ b/kotlinx-coroutines-test/common/test/RunTestTest.kt
@@ -413,6 +413,8 @@ class RunTestTest {
      * Tests that [runTest] cleans up the exception handler even if it threw on initialization.
      *
      * This test must be run manually, because it writes garbage to the log.
+     *
+     * The JVM-only source set contains a test equivalent to this one that isn't ignored.
      */
     @Test
     @Ignore
--- a/kotlinx-coroutines-test/common/test/RunTestTest.kt
+++ b/kotlinx-coroutines-test/common/test/RunTestTest.kt
@@ -408,4 +408,67 @@ class RunTestTest {
     fun testCoroutineCompletingWithoutDispatch() = runTest(timeout = Duration.INFINITE) {
         launch(Dispatchers.Default) { delay(100) }
     }
+
+    /**
+     * Tests that [runTest] cleans up the exception handler even if it threw on initialization.
+     *
+     * This test must be run manually, because it writes garbage to the log.
+     */
+    @Test
+    @Ignore
+    fun testExceptionCaptorCleanedUpOnPreliminaryExit(): TestResult = testResultChain({
+        // step 1: installing the exception handler
+        println("step 1")
+        runTest { }
+    }, {
+        it.getOrThrow()
+        // step 2: throwing an uncaught exception to be caught by the exception-handling system
+        println("step 2")
+        createTestResult {
+            launch(NonCancellable) { throw TestException("A") }
+        }
+    }, {
+        it.getOrThrow()
+        // step 3: trying to run a test should immediately fail, even before entering the test body
+        println("step 3")
+        try {
+            runTest {
+                fail("unreached")
+            }
+            fail("unreached")
+        } catch (e: UncaughtExceptionsBeforeTest) {
+            val cause = e.suppressedExceptions.single()
+            assertIs<TestException>(cause)
+            assertEquals("A", cause.message)
+        }
+        // step 4: trying to run a test again should not fail with an exception
+        println("step 4")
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-test/common/test/RunTestTest.kt
```diff
@@ -408,4 +408,67 @@ class RunTestTest {
...
     }
+
+    /**
+     * Tests that [runTest] cleans up the exception handler even if it threw on initialization.
+     *
+     * This test must be run manually, because it writes garbage to the log.
+     */
+    @Test
+    @Ignore
+    fun testExceptionCaptorCleanedUpOnPreliminaryExit(): TestResult = testResultChain({
```
qwwdfsad, line 419: If you see fit, you can write a JVM-only test that invokes `Thread.setDefaultUncaughtExceptionHandler` to avoid stderr cluttering

qwwdfsad (approved):

kotlinx-coroutines-test/common/test/RunTestTest.kt
dkhalanskyjb: Good idea, thanks!

dkhalanskyjb:
> :kotlinx-coroutines-core:koverVerify
> Execution failed for task ':kotlinx-coroutines-core:koverVerify'. kotlinx.kover.gradle.plugin.commons.KoverVerificationException: Rule violated: lines covered percentage is 84.970700, but expected minimum is 85

I'm surprised: this failure happened after I *added* a test, not to mention that it's a separate module.

qwwdfsad:
@dkhalanskyjb is it good to go?

dkhalanskyjb:
Yeah.

PR #3921, opened 2023-10-23
Properly round `Duration` instances to milliseconds

### LINKED ISSUE #3920: `delay(Duration)` improperly coerces nanoseconds and microseconds
**Describe the bug**

I am currently unsure if this is desired behavior, so if it is, please let me know.

When invoking a delay with a `Duration`, the value of that duration is coerced to milliseconds and then delegated to `delay(Long)`. It is not documented behavior that delaying with a duration loses granularity for nanoseconds or microseconds. For example, delaying with `998.75909ms` gets coerced to a call of `998ms` when delegating to `delay(Long)`. 

The offending code is the following: 

```kt
internal fun Duration.toDelayMillis(): Long =
    if (this > Duration.ZERO) inWholeMilliseconds.coerceAtLeast(1) else 0
```

The symptom of this issue is that when working on implementing a cron-like API for delay based job scheduling, I was having jobs invoked up to 600us before they should have been invoked. The end result was that jobs scheduled to occur at the beginning of a second in a certain time zone would invoke early once and then as expected afterwards in quick succession. 

The above function `Duration.toDelaymillis` should likely coerce durations which have a nanosecond component up `duration.inWholeMilliseconds + 1`. 

Locally this implementation is working well for me. 

```kt
internal fun Duration.myToMillis(): Long = if (this > Duration.ZERO) {
    val millis = inWholeMilliseconds
    if (millis * 1_000_000 < inWholeNanoseconds) millis + 1 else millis
} else 0
```

**Provide a Reproducer**

```kt
    @Test fun foo() = runTest(timeout = 1.days) {
            while (true) {
                val expectedDuration = 998.milliseconds + 902752.nanoseconds
                val actualDuration = measureTime { delay(expectedDuration) }
                val difference = actualDuration - expectedDuration
                assertTrue(difference.isPositive()) // fails
            }
        }
    }

```

kevincianfarini:
Prior to this commit Durations used for delays or timeouts lost their nanosecond granularity when being converted to a millisecond Long value. This effectively meant that delays could resume prior to when they were scheduled to do so.

This commit solves this by rounding a Duration with nanosecond components up to the next largest millisecond.

Closes #3920

### DIFF
--- a/kotlinx-coroutines-core/common/src/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/Delay.kt
@@ -7,6 +7,7 @@ package kotlinx.coroutines
 import kotlinx.coroutines.selects.*
 import kotlin.coroutines.*
 import kotlin.time.*
+import kotlin.time.Duration.Companion.nanoseconds
 
 /**
  * This dispatcher _feature_ is implemented by [CoroutineDispatcher] implementations that natively support
@@ -106,7 +107,7 @@ internal interface DelayWithTimeoutDiagnostics : Delay {
 public suspend fun awaitCancellation(): Nothing = suspendCancellableCoroutine {}
 
 /**
- * Delays coroutine for a given time without blocking a thread and resumes it after a specified time.
+ * Delays coroutine for at least the given time without blocking a thread and resumes it after a specified time.
  * If the given [timeMillis] is non-positive, this function returns immediately.
  *
  * This suspending function is cancellable.
@@ -133,7 +134,7 @@ public suspend fun delay(timeMillis: Long) {
 }
 
 /**
- * Delays coroutine for a given [duration] without blocking a thread and resumes it after the specified time.
+ * Delays coroutine for at least the given [duration] without blocking a thread and resumes it after the specified time.
  * If the given [duration] is non-positive, this function returns immediately.
  *
  * This suspending function is cancellable.
@@ -154,8 +155,10 @@ public suspend fun delay(duration: Duration): Unit = delay(duration.toDelayMilli
 internal val CoroutineContext.delay: Delay get() = get(ContinuationInterceptor) as? Delay ?: DefaultDelay
 
 /**
- * Convert this duration to its millisecond value.
- * Positive durations are coerced at least `1`.
+ * Convert this duration to its millisecond value. Durations which have a nanosecond component less than
+ * a single millisecond will be rounded up to the next largest millisecond.
  */
-internal fun Duration.toDelayMillis(): Long =
-    if (this > Duration.ZERO) inWholeMilliseconds.coerceAtLeast(1) else 0
+internal fun Duration.toDelayMillis(): Long = when (isPositive()) {
+    true -> plus(999_999L.nanoseconds).inWholeMilliseconds
+    false -> 0L
+}
--- a/kotlinx-coroutines-core/common/test/DurationToMillisTest.kt
+++ b/kotlinx-coroutines-core/common/test/DurationToMillisTest.kt
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+package kotlinx.coroutines
+
+import kotlin.test.*
+import kotlin.time.*
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
+
+class DurationToMillisTest {
+
+    @Test
+    fun testNegativeDurationCoercedToZeroMillis() {
+        assertEquals(0L, (-1).seconds.toDelayMillis())
+    }
+
+    @Test
+    fun testZeroDurationCoercedToZeroMillis() {
+        assertEquals(0L, 0.seconds.toDelayMillis())
+    }
+
+    @Test
+    fun testOneNanosecondCoercedToOneMillisecond() {
+        assertEquals(1L, 1.nanoseconds.toDelayMillis())
+    }
+
+    @Test
+    fun testOneSecondCoercedTo1000Milliseconds() {
+        assertEquals(1_000L, 1.seconds.toDelayMillis())
+    }
+
+    @Test
+    fun testMixedComponentDurationRoundedUpToNextMillisecond() {
+        assertEquals(999L, (998.milliseconds + 75909.nanoseconds).toDelayMillis())
+    }
+
+    @Test
+    fun testOneExtraNanosecondRoundedUpToNextMillisecond() {
+        assertEquals(999L, (998.milliseconds + 1.nanoseconds).toDelayMillis())
+    }
+
+    @Test
+    fun testInfiniteDurationCoercedToLongMaxValue() {
+        assertEquals(Long.MAX_VALUE, Duration.INFINITE.toDelayMillis())
+    }
+
+    @Test
+    fun testNegativeInfiniteDurationCoercedToZero() {
+        assertEquals(0L, (-Duration.INFINITE).toDelayMillis())
+    }
+
+    @Test
+    fun testNanosecondOffByOneInfinityDoesNotOverflow() {
+        assertEquals(Long.MAX_VALUE / 1_000_000, (Long.MAX_VALUE - 1L).nanoseconds.toDelayMillis())
+    }
+
+    @Test
+    fun testMillisecondOffByOneInfinityDoesNotIncrement() {
+        assertEquals((Long.MAX_VALUE / 2) - 1, ((Long.MAX_VALUE / 2) - 1).milliseconds.toDelayMillis())
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -154,8 +154,15 @@ public suspend fun delay(duration: Duration): Unit = delay(duration.toDelayMilli
...
- * Convert this duration to its millisecond value.
- * Positive durations are coerced at least `1`.
+ * Convert this duration to its millisecond value. Durations which have a nanosecond component less than
+ * a single millisecond will be rounded up to the next largest millisecond.
  */
-internal fun Duration.toDelayMillis(): Long =
-    if (this > Duration.ZERO) inWholeMilliseconds.coerceAtLeast(1) else 0
+internal fun Duration.toDelayMillis(): Long {
+    val millis = inWholeMilliseconds
+    val nanosecondsInMillisecond = 1_000_000
```
kevincianfarini, line 162: Is there a better constant I can use for this rather than a random magic number?

kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -133,7 +133,7 @@ public suspend fun delay(timeMillis: Long) {
 }
 
 /**
- * Delays coroutine for a given [duration] without blocking a thread and resumes it after the specified time.
+ * Delays coroutine for at least the given [duration] without blocking a thread and resumes it after the specified time.
```
kevincianfarini, line 136: I felt that this was an important distinction which felt natural to make here. Please let me know if I should revert.

kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -154,8 +154,15 @@ public suspend fun delay(duration: Duration): Unit = delay(duration.toDelayMilli
...
 
 /**
- * Convert this duration to its millisecond value.
- * Positive durations are coerced at least `1`.
+ * Convert this duration to its millisecond value. Durations which have a nanosecond component less than
+ * a single millisecond will be rounded up to the next largest millisecond.
  */
-internal fun Duration.toDelayMillis(): Long =
-    if (this > Duration.ZERO) inWholeMilliseconds.coerceAtLeast(1) else 0
+internal fun Duration.toDelayMillis(): Long {
```
denis-bezrukov, line 160: it will fail on attempt to do `Duration.INFINITE.toDelayMillis()` as the `Duration.INFINITE.inWholeMilliseconds` returns Long.MAX_VALUE. Adding 1 to it will overflow and return Long.MIN_VALUE. Another place when it may overflow is `millis * nanosecondsInMillisecond` (for any millis duration above Long.MAX_VALUE / 1000_000). 

Probably the easiest way to round it up is to defer all the math to kotlin.time
```
(this + 999999.nanoseconds).inWholeMilliseconds
```

kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -154,8 +154,15 @@ public suspend fun delay(duration: Duration): Unit = delay(duration.toDelayMilli
...
+    val millis = inWholeMilliseconds
+    val nanosecondsInMillisecond = 1_000_000
+    return when {
+        this <= Duration.ZERO -> 0L
+        millis * nanosecondsInMillisecond < inWholeNanoseconds -> millis + 1
```
JakeWharton, line 165: It might be more efficient to check `isFinite() && this != millis.milliseconds` since it doesn't involve any math operations (especially on JS where they're slow). Plus I suspect you're not handling overflow with the current code correctly.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/Delay.kt
kevincianfarini: Perhaps it's best to do modulo arithmetic?

kotlinx-coroutines-core/common/src/Delay.kt
kevincianfarini: Switched to use modulo arithmetic.

denis-bezrukov: It will now fail for any large (but finite) number, just because of the precision error when attempting to access `inWholeNanoseconds`

i.e. this one should fail
```
@Test fun large_but_finite() {
    val milliseconds = Long.MAX_VALUE / 10
    assertEquals(
        expected = milliseconds, 
        actual = milliseconds.milliseconds.toDelayMillis(),
    )
}
```
because 
```
Long.MAX_VALUE                | 9223372036854775807
Long.MAX_VALUE/10             | 922337203685477580
In nanos (absolute precision) | 922337203685477580000000 | mod1000000=0
In nanos (coerced to 64bits)  | 9223372036854775807      | mod1000000=775807

```

kevincianfarini: I see. Good catch. Will delegate to kotlin.time arithmetic.

kevincianfarini: Updated.

kotlinx-coroutines-core/common/src/Delay.kt
```diff
@@ -154,8 +155,18 @@ public suspend fun delay(duration: Duration): Unit = delay(duration.toDelayMilli
...
+ * Convert this duration to its millisecond value. Durations which have a nanosecond component less than
+ * a single millisecond will be rounded up to the next largest millisecond.
  */
-internal fun Duration.toDelayMillis(): Long =
-    if (this > Duration.ZERO) inWholeMilliseconds.coerceAtLeast(1) else 0
+internal fun Duration.toDelayMillis(): Long {
+    val millis = inWholeMilliseconds
+    return when {
+        !isPositive() -> 0L
+
```
dkhalanskyjb, line 165: Style: in this library, we avoid empty lines inside functions.

kotlinx-coroutines-core/common/src/Delay.kt
dkhalanskyjb: I agree with @denis-bezrukov:
> `(this + 999999.nanoseconds).inWholeMilliseconds`

seems to be the cleanest and, I suspect, also the most efficient solution available to us without getting access to the internals of `Duration`. Here, in realistic cases, it's just a single addition and a conversion to milliseconds.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/Delay.kt
kevincianfarini: Pushed. Thank you two for the review.

kotlinx-coroutines-core/common/test/DurationToMillisTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test fun nanosecond_off_by_one_infinity_does_not_overflow() = assertEquals(
+        expected = Long.MAX_VALUE / 1_000_000,
+        actual = (Long.MAX_VALUE - 1L).nanoseconds.toDelayMillis(),
+    )
+
+    @Test fun millisecond_off_by_one_infinity_does_not_overflow() = assertEquals(
+        expected = Long.MAX_VALUE, // INFINITY duration.
+        actual = (Long.MAX_VALUE - 1L).milliseconds.toDelayMillis(),
+    )
```
kevincianfarini, line 62: I actually don't think this is correct now that I look at it again. This duration has no nanosecond sub component and therefore shouldn't be rounded.

kotlinx-coroutines-core/common/test/DurationToMillisTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
+
+class DurationToMillisTest {
+
+    @Test fun negative_duration_coerced_to_zero_millis() = assertEquals(
+        expected = 0L,
+        actual = (-1).seconds.toDelayMillis(),
+    )
```
dkhalanskyjb, line 17: Style: please look at how all the other tests are structured in this library.
* Test names: we use camel case, and also, all test names begin with `test`.
* Indentation is different: `@Test` should be on a separate line, function calls are typically on a single line, etc.
* Explicit parameter names here don't add much and are mostly noise.

kevincianfarini:
Hi! If there's any other desired changes I'm happy to make them.

dkhalanskyjb:
Please rebase your changes on top of `develop` and set that as the base of the PR. See https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs

kevincianfarini:
Done. Sorry, I should have checked the contribution guide first.

kevincianfarini:
Do we think there's a chance this will be merged prior to the next version release? 😄

dkhalanskyjb:
Yes.

dkhalanskyjb (approved):
:thumb-up:

kevincianfarini:
Nice! Thank you.

PR #3922, opened 2023-10-25
Generalize SharedFlow.onSubscription operator - support more general output flow type

sebek64:
- this can be useful in many scenarios, but in general, it allows one to use type hierarchy instead of composition for type-safe code
 - that is especially nice with sealed class hierarchies

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
@@ -398,14 +398,14 @@ private class ReadonlyStateFlow<T>(
  *
  * The receiver of the [action] is [FlowCollector], so `onSubscription` can emit additional elements.
  */
-public fun <T> SharedFlow<T>.onSubscription(action: suspend FlowCollector<T>.() -> Unit): SharedFlow<T> =
+public fun <T : R, R> SharedFlow<T>.onSubscription(action: suspend FlowCollector<R>.() -> Unit): SharedFlow<R> =
     SubscribedSharedFlow(this, action)
 
-private class SubscribedSharedFlow<T>(
+private class SubscribedSharedFlow<T : R, R>(
     private val sharedFlow: SharedFlow<T>,
-    private val action: suspend FlowCollector<T>.() -> Unit
-) : SharedFlow<T> by sharedFlow {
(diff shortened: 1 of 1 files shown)


dkhalanskyjb:
This can easily be achieved using type casts where needed:
```kotlin
        val myFlow = MutableStateFlow<Int>(1)
        val result: SharedFlow<Number> = (myFlow as SharedFlow<Number>).onSubscription {
            emit(2)
        }
```

In any case, it's not clear what are these "many scenarios" where this is useful. Having `T` everywhere is strictly more powerful than having `R`, but if you need `R` explicitly, for example, due to a `FlowCollector` type, then `T` should automatically coerce to the proper type. If you have some specific code that's cumbersome to write that would be fixed by your PR, please file an issue.

sebek64:
Ok, explicit type cast didn't come to my mind. So instead of this code
```
sealed interface TopLevel

data object StreamStart : TopLevel

data class StreamItem(
...
) : TopLevel

val someSharedFlow = MutableSharedFlow<StreamItem>

someSharedFlow
    .onSubscription {
        emit(StreamStart)
    }
```
I can write
```
...
(someSharedFlow as SharedFlow<TopLevel>)
    .onSubscription {
        emit(StreamStart)
    }
```
Thank you for the hint. Anyway, I believe that if the operator would work without explicit cast and it will just smartly cast the output based on emits (similarly to e.g. transform operator), the user experience would be better.

dkhalanskyjb:
Two points:

1. You can also do
```kotlin
        someSharedFlow
            .onSubscription<TopLevel> {
                emit(StreamStart)
            }
```

3. The code you're proposing is *not* fixed by your PR:
```kotlin
        fun <T : R, R> SharedFlow<T>.myOnSubscription(action: suspend FlowCollector<R>.() -> Unit): SharedFlow<R> = TODO()

        val someSharedFlow = MutableSharedFlow<StreamItem>()

        someSharedFlow.myOnSubscription {
            emit(StreamStart)
        }
```

The receiver in the block is still recognized as `FlowCollector<StreamItem>`, leading to a compilation error.

sebek64:
Ok, thanks, that also works, and is much nicer. We can consider this topic as closed.

PR #3924, opened 2023-11-02
Ensure Dispatchers.Main != Dispatchers.Main.immediate on Android

### LINKED ISSUE #3545: `flowOn(Dispatchers.Main)` does not change upstream dispatcher
```kotlin
    flowOf(1)
      .onEach {
        Timber.d("timer: $it ${Thread.currentThread().name} ${currentCoroutineContext()[ContinuationInterceptor]}")
      }
      .flowOn(Dispatchers.Main)
      .onEach {
        Timber.d("timer [1]: $it ${Thread.currentThread().name} ${currentCoroutineContext()[ContinuationInterceptor]}")
      }
      .launchIn(viewModelScope)

// viewModelScope's context is SupervisorJob() + Dispatchers.Main.immediate
```

Console
```
timer: 1 main Dispatchers.Main.immediate
timer [1]: 1 main Dispatchers.Main.immediate
```

### LINKED ISSUE #3963: Unintuitive dispatching behavior in some rare cases with Android's `Dispatchers.Main.immediate`
See https://github.com/Kotlin/kotlinx.coroutines/pull/3924#discussion_r1381428140

This code produces `1`, `2`, `3`:

```kotlin
@Test
fun foo() {
    GlobalScope.launch(Dispatchers.IO) {
        withContext(Dispatchers.Main.immediate) {
            println("1")
            launch(Dispatchers.Main) {
                println("2")
            }
            withContext(Dispatchers.Main) {
                println("3")
            }
        }
    }
}
```

This code, however, produces `1`, `3`, `2`:

```kotlin
@Test
fun foo() {
    GlobalScope.launch(Dispatchers.Main) { // <-- note the different dispatcher
        withContext(Dispatchers.Main.immediate) {
            println("1")
            launch(Dispatchers.Main) {
                println("2")
            }
            withContext(Dispatchers.Main) {
                println("3")
            }
        }
    }
}
```

This behavior may be surprising. For example, consider this pseudo-code:

```kotlin
fun updateUi() = scope.launch(Dispatchers.Main) { }

suspend fun readUiElement(): Int = withContext(Dispatchers.Main) { 1 }

fun update() = viewModelScope.launch {
  updateUi()
  cache = readUiElement()
}
```

Depending on whether `update` is called from the main thread, the block in `updateUi` will happen either earlier or later than `readUiElement`, even though visually, it seems like `updateUi` should happen earlier.

Notably, all the other implementations of `Dispatchers.Main` that we provide don't suffer from the same problem but instead consistently output `1`, `2`, `3` in both scenarios.

dkhalanskyjb:
Fixes #3545
Fixes #3963

### DIFF
--- a/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
+++ b/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
@@ -6,7 +6,7 @@ package kotlinx.coroutines
 
 import kotlin.test.*
 
-class ImmediateDispatcherTest : TestBase() {
+class ImmediateDispatcherTest : MainDispatcherTestBase(Dispatchers.Main) {
 
     @Test
     fun testImmediate() = runTest {
--- a/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
+++ b/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
@@ -11,6 +11,7 @@ class ImmediateDispatcherTest : MainDispatcherTestBase() {
 
     /** Tests that entering [MainCoroutineDispatcher.immediate] takes priority even outside [Dispatchers.Main]. */
     @Test
+    @Ignore // TODO: started failing for some reason
     fun testImmediate() = runTest {
         expect(1)
         val job = launch { expect(3) }
--- a/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
+++ b/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
@@ -9,13 +9,12 @@ import kotlin.test.*
 
 class ImmediateDispatcherTest : MainDispatcherTestBase() {
 
-    /** Tests that entering [MainCoroutineDispatcher.immediate] takes priority even outside [Dispatchers.Main]. */
+    /** Tests that [MainCoroutineDispatcher.immediate] doesn't require dispatches from the test context. */
     @Test
-    @Ignore // TODO: started failing for some reason
     fun testImmediate() = runTest {
         expect(1)
         val job = launch { expect(3) }
-        assertTrue(Dispatchers.Main.immediate.isDispatchNeeded(currentCoroutineContext()))
+        assertFalse(Dispatchers.Main.immediate.isDispatchNeeded(currentCoroutineContext()))
         withContext(Dispatchers.Main.immediate) {
             expect(2)
         }
--- a/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
+++ b/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
@@ -4,14 +4,17 @@
 
 package kotlinx.coroutines
 
+import kotlin.coroutines.*
 import kotlin.test.*
 
-class ImmediateDispatcherTest : MainDispatcherTestBase(Dispatchers.Main) {
+class ImmediateDispatcherTest : MainDispatcherTestBase() {
 
+    /** Tests that entering [MainCoroutineDispatcher.immediate] takes priority even outside [Dispatchers.Main]. */
     @Test
     fun testImmediate() = runTest {
         expect(1)
         val job = launch { expect(3) }
+        assertTrue(Dispatchers.Main.immediate.isDispatchNeeded(currentCoroutineContext()))
         withContext(Dispatchers.Main.immediate) {
             expect(2)
         }
@@ -28,5 +31,11 @@ class ImmediateDispatcherTest : MainDispatcherTestBase(Dispatchers.Main) {
         }
         job.join()
         finish(4)
+    }
+
+    override fun isMainThread(): Boolean? = null
+
+    override fun scheduleOnMainQueue(block: () -> Unit) {
+        Dispatchers.Default.dispatch(EmptyCoroutineContext, Runnable { block() })
     }
 }
--- a/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
+++ b/kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
@@ -7,7 +7,7 @@ package kotlinx.coroutines
 import kotlin.coroutines.*
 import kotlin.test.*
 
-class ImmediateDispatcherTest : MainDispatcherTestBase() {
+class ImmediateDispatcherTest : MainDispatcherTestBase.WithRealTimeDelay() {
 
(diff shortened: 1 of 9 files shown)


ui/kotlinx-coroutines-android/test/HandlerDispatcherTest.kt
```diff
@@ -132,4 +132,11 @@ class HandlerDispatcherTest : TestBase() {
         mainLooper.scheduler.advanceBy(51, TimeUnit.MILLISECONDS)
         finish(5)
     }
+
+    @Test
+    fun testHandlerDispatcherNotEqualToImmediate() {
```
qwwdfsad, line 137: This change changes the behaviour of the following code:
```
@Test
    fun foo() = runBlocking<Unit> {
        withContext(Dispatchers.Main) {
            println("1")
            withContext(Dispatchers.Main.immediate) {
                println("2")
                launch(Dispatchers.Main) {
                    println("4")
                }
                withContext(Dispatchers.Main) {
                    println("3")
                }
            }
        }
    }
```

It seems like this is expected (at least from the issue description), and I would rather expect the current (after your change) behaviour, but still pointing it out just in case

qwwdfsad (approved):

ui/kotlinx-coroutines-android/test/HandlerDispatcherTest.kt
dkhalanskyjb: Oh. Didn't realize this. Then, we'll have to weigh the pros and cons.

I checked all the other `Dispatchers.Main.immediate` implementations, and they all use object identity for equality checks, so this change is also needed simply for consistency.

Also, before this change, the behavior of
```kotlin
            withContext(Dispatchers.Main.immediate) {
                println("2")
                launch(Dispatchers.Main) {
                    println("4")
                }
                withContext(Dispatchers.Main) {
                    println("3")
                }
            }
```
was `2, 4, 3` if it was not called from the main thread, but `2, 3, 4` it it was. Now, it's consistently `2, 4, 3`. I think it's more intuitive that the calling thread doesn't affect the ordering.

Also, yes, I, too, would expect the proposed behavior:

```kotlin
fun updateUi() = scope.launch(Dispatchers.Main) { }

suspend fun readUiElement(): Int = withContext(Dispatchers.Main) { 1 }

fun update() = viewModelScope.launch {
  updateUi()
  cache = readUiElement()
}
```

I think it's reasonable to expect that `readUiElement` detects the results of `updateUi`. Currently, for this to work reliably, I think a `yield` would need to be inserted between the two lines.

... That said, of course, this will subtly break someone's code that relied on the existing implicit ordering.

I'd say the pros outweigh the cons. Most people would be better served by having the execution order roughly the same as the order of function invocations.

qwwdfsad: Then it's worth doing the following:

* Fix this behaviour with tests explicitly (in all integrations)
* Maybe document it around `immediate` property
* Rephrase the issue (or create a new one) in a bit more clear manner so we'll reference it in the changelog properly

It's okay to do it both in and out of the scope of this issue. WDYT?

dkhalanskyjb:
From the latest build failure:
```
e: integration-testing/smokeTest/src/commonMain/kotlin/Sample.kt:4:5 Cannot access class 'kotlin.coroutines.CoroutineContext'. Check your module classpath for missing or conflicting dependencies
e: integration-testing/smokeTest/src/commonMain/kotlin/Sample.kt:6:9 Unresolved reference: println
e: integration-testing/smokeTest/src/commonMain/kotlin/Sample.kt:8:5 Unresolved reference: println
e: org.jetbrains.kotlin.util.KotlinFrontEndException: Exception while analyzing expression in (3,25) in integration-testing/smokeTest/src/commonMain/kotlin/Sample.kt
Attachments:
causeThrowable
java.lang.AssertionError: Built-in class kotlin.Any is not found
```
and
```
Execution failed for task ':kotlinx-coroutines-swing:koverVerify'. kotlinx.kover.gradle.plugin.commons.KoverVerificationException: Rule violated: lines covered percentage is 53.571400, but expected minimum is 70
```

Oh, mighty computing gods, please forgive me for all my sins! I repent twice, thrice, and however many times is needed. I knew I shouldn't have written that compile-time VM with the C++ templates at the university! I implore you, please, spare me and allow me to proceed with this humble pull request. Amen.

qwwdfsad:
Behold, thou shalt not ignore the tests, lest thy build shell not pass!

The first failure is penned by those who have sinned. Thou shalt not seek it -- the sorrow shalt thou find in Gradle

qwwdfsad (approved):
Nice one

kotlinx-coroutines-core/js/test/ImmediateDispatcherTest.kt
```diff
@@ -4,14 +4,17 @@
...
 
+import kotlin.coroutines.*
 import kotlin.test.*
 
-class ImmediateDispatcherTest : TestBase() {
+class ImmediateDispatcherTest : MainDispatcherTestBase.WithRealTimeDelay() {
 
+    /** Tests that [MainCoroutineDispatcher.immediate] doesn't require dispatches from the test context. */
     @Test
     fun testImmediate() = runTest {
```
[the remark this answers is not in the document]
dkhalanskyjb: This tests JS-specific behavior: the `Main.immediate` dispatcher on JS won't require a dispatch even from `Dispatchers.Default`.

ui/kotlinx-coroutines-android/src/HandlerDispatcher.kt
```diff
@@ -172,8 +169,10 @@ internal class HandlerContext private constructor(
         if (invokeImmediately) "$str.immediate" else str
     }
 
-    override fun equals(other: Any?): Boolean = other is HandlerContext && other.handler === handler
-    override fun hashCode(): Int = System.identityHashCode(handler)
+    override fun equals(other: Any?): Boolean =
+        other is HandlerContext && other.handler === handler && other.invokeImmediately == invokeImmediately
+    // inlining `Boolean.hashCode()` for Android compatibility, as requested by Animal Sniffer
+    override fun hashCode(): Int = System.identityHashCode(handler) xor if (invokeImmediately) 1231 else 1237
```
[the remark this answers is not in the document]
dkhalanskyjb: Worth noting that choosing some other constant for the polynomial hash could lead us to a much better situation like only a handful of collisions being possible at all. I don't think it's worth it to search for the required constants, though, as no one will ever be sad that their `HashSet<MainCoroutineDispatcher>` on Android has a `2 * 10^(-10)` collision chance for every pair.

PR #3927, opened 2023-11-06
Rename `inner` to `innerJob` in exception-handling.md to fix highlighting

SebastianAigner:
Because `inner` is a soft keyword in Kotlin, it's wrongly highlighted in the playground. Renaming the variable to `innerJob` makes it consistent with the other `job` variable that is in the same snippet, and removes this odd highlighting.

Screenshot 2023-11-06 at 15 44 40


dkhalanskyjb (approved):
I checked, and nowhere in this document is `inner` referred to by name, so this change is valid.

PR #3935, opened 2023-11-08
Remove reference to actors from coroutines guide

SebastianAigner:
Actors are marked as `@ObsoleteCoroutinesApi`, so people learning about coroutines most likely don't need to learn about them here.

### DIFF
--- a/docs/topics/exception-handling.md
+++ b/docs/topics/exception-handling.md
@@ -9,7 +9,7 @@ coroutine throw an exception.
 
 ## Exception propagation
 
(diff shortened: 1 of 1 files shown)


qwwdfsad:
@SebastianAigner do you need a review here or we are waiting for more changes? (noticed there is no reviewer)

SebastianAigner:
@qwwdfsad review would be appreciated!

qwwdfsad (approved):
Thanks!

We also have `actor` mentions in `coroutine-guide-ui`

PR #3943, opened 2023-11-17
fix MDCContext lost updates made in coroutine after suspension and restoration

f1qwase:
**Issue:**
When using `withContext(MDCContext()) { ... }`, I observed that the MDC context set inside the block is lost after coroutine suspension. Upon investigation, I discovered that this behavior occurs because MDCContext() always restores the original outer MDC context, ignoring any changes made within the block.

**Proposed Solution:**
To address this issue, I propose updating the MDC contextMap with the actual MDC Context on each coroutine suspension. This modification ensures that the MDC context changes made within the block persist beyond coroutine suspension.


qwwdfsad:
This is an expected behaviour -- the coroutine context element cannot be mutated, and updates are not propagated, as stated in the `MDCContext` documentation.

Mutating it as `var` is not an option, as it's immediately prone to data-races

PR #3946, opened 2023-11-21
Introduce jvmBenchmarks into kotlinx-coroutines-core

qwwdfsad:
It allows us to benchmark `internal` API unavailable otherwise and allows us to backport K2 changes properly.

Only benchmarks that somehow leverage `internal` API are moved in the new source-set

### DIFF
--- a/kotlinx-coroutines-core/jvmBenchmark/kotlin/kotlinx/coroutines/SemaphoreBenchmark.kt
+++ b/kotlinx-coroutines-core/jvmBenchmark/kotlin/kotlinx/coroutines/SemaphoreBenchmark.kt
@@ -79,7 +79,7 @@ open class SemaphoreBenchmark {
 }
 
 enum class SemaphoreBenchDispatcherCreator(val create: (parallelism: Int) -> CoroutineDispatcher) {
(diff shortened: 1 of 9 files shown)


kotlinx-coroutines-core/jvmBenchmark/kotlin/kotlinx/coroutines/SemaphoreBenchmark.kt
```diff
@@ -81,7 +80,6 @@ open class SemaphoreBenchmark {
 
 enum class SemaphoreBenchDispatcherCreator(val create: (parallelism: Int) -> CoroutineDispatcher) {
     //    FORK_JOIN({ parallelism -> ForkJoinPool(parallelism).asCoroutineDispatcher() }),
```
dkhalanskyjb, line 82: Why is this not uncommented, but similar code in `ChannelProducerConsumerBenchmark.kt` is?

qwwdfsad: No real reason, uncommented as well

dkhalanskyjb (approved):

PR #3948, opened 2023-11-21
Fix kotlinx-coroutines-debug having the wrong module-info

### LINKED ISSUE #3944: `kotlinx-coroutine-debug` jar seems incorrectly packaged exposing `bytebuddy` module instead of `kotlinx.coroutine.debug` module
**Describe the bug**

`kotlinx-coroutine-debug` jar exposes `bytebuddy` module instead of `kotlinx.coroutine.debug` module.
[`module-info.java` of `kotlinx.coroutine.debug`](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-debug/src/module-info.java) seems to be ignored.

```bash
$ titouan.bion@NVC00231 ~/Downloads curl -s https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-debug/1.7.3/kotlinx-coroutines-debug-1.7.3.jar -o kotlinx-coroutines-debug-1.7.3.jar && jar --describe-module --file kotlinx-coroutines-debug-1.7.3.jar --release 9
releases: 9

net.bytebuddy jar:file:///Users/titouan.bion/Downloads/kotlinx-coroutines-debug-1.7.3.jar!/META-INF/versions/9/module-info.class
exports net.bytebuddy
exports net.bytebuddy.agent.builder
exports net.bytebuddy.asm
exports net.bytebuddy.build
exports net.bytebuddy.description
exports net.bytebuddy.description.annotation
exports net.bytebuddy.description.enumeration
exports net.bytebuddy.description.field
exports net.bytebuddy.description.method
exports net.bytebuddy.description.modifier
exports net.bytebuddy.description.type
exports net.bytebuddy.dynamic
exports net.bytebuddy.dynamic.loading
exports net.bytebuddy.dynamic.scaffold
exports net.bytebuddy.dynamic.scaffold.inline
exports net.bytebuddy.dynamic.scaffold.subclass
exports net.bytebuddy.implementation
exports net.bytebuddy.implementation.attribute
exports net.bytebuddy.implementation.auxiliary
exports net.bytebuddy.implementation.bind
exports net.bytebuddy.implementation.bind.annotation
exports net.bytebuddy.implementation.bytecode
exports net.bytebuddy.implementation.bytecode.assign
exports net.bytebuddy.implementation.bytecode.assign.primitive
exports net.bytebuddy.implementation.bytecode.assign.reference
exports net.bytebuddy.implementation.bytecode.collection
exports net.bytebuddy.implementation.bytecode.constant
exports net.bytebuddy.implementation.bytecode.member
exports net.bytebuddy.jar.asm
exports net.bytebuddy.jar.asm.commons
exports net.bytebuddy.jar.asm.signature
exports net.bytebuddy.matcher
exports net.bytebuddy.pool
exports net.bytebuddy.utility
exports net.bytebuddy.utility.privilege
exports net.bytebuddy.utility.visitor
requires java.base mandated
requires java.instrument static
requires jdk.unsupported static
requires net.bytebuddy.agent static
```

For example for `kotlinx-coroutines-core-jvm`, it is `kotlinx.coroutines.core` module, as expected:
```
$ titouan.bion@NVC00231 ~/Downloads curl -s https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core-jvm/1.7.3/kotlinx-coroutines-core-jvm-1.7.3.jar -o kotlinx-coroutines-core-jvm-1.7.3.jar && jar --describe-module --file kotlinx-coroutines-core-jvm-1.7.3.jar --release 9
releases: 9

kotlinx.coroutines.core jar:file:///Users/titouan.bion/Downloads/kotlinx-coroutines-core-jvm-1.7.3.jar!/META-INF/versions/9/module-info.class
exports kotlinx.coroutines
exports kotlinx.coroutines.channels
exports kotlinx.coroutines.debug
exports kotlinx.coroutines.debug.internal
exports kotlinx.coroutines.flow
exports kotlinx.coroutines.flow.internal
exports kotlinx.coroutines.future
exports kotlinx.coroutines.internal
exports kotlinx.coroutines.intrinsics
exports kotlinx.coroutines.scheduling
exports kotlinx.coroutines.selects
exports kotlinx.coroutines.stream
exports kotlinx.coroutines.sync
exports kotlinx.coroutines.time
requires java.base mandated
requires java.instrument static
requires jdk.unsupported static
requires kotlin.stdlib transitive
uses kotlinx.coroutines.CoroutineExceptionHandler
uses kotlinx.coroutines.internal.MainDispatcherFactory
```

dkhalanskyjb:
Fixes #3944


kotlinx-coroutines-debug/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+            "Multi-Release" to "true"
+        ))
+    }
+}
+
+tasks.getByName("publishMavenPublicationToMavenLocal") {
+    dependsOn(shadowJarWithCorrectModuleInfo)
+}
+
+VersionFile.configure(project, shadowJarWithCorrectModuleInfo.get())
```
qwwdfsad, line 92: Thank you for going the extra mile with rewriting the build script into kts!

While we are here, would you mind improving it even more?

The current approach breaks [configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html).

AFAIK, the recommended approach is the following:

0) Generally, avoid materializing tasks during the configuration phase (e.g. when you are receiving the task of a particular type in an arbitrary place of code)

1) This line should be `VersionFile.configure(project, shadowJarWithCorrectModuleInfo)`

2) `VersionFile.configure` should configure the given task provider: `jarProvider.Configure { dependsOn(...) `

3) The other callsite (version file convention) should do the similar trick:

```
tasks.withType<>.configureEach { ... }
```

kotlinx-coroutines-debug/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+
+configurations.all {
```
qwwdfsad, line 94: Is it correct that this block removes everything from the output that depends on the default `ShadowJar` task, and replaces everything that contains the dependency on `Jar` task with `ShadowJarWithCorrectModuleInfo`?

qwwdfsad: I'm a bit lost here as the `removeIf` covers all artifacts that depends on our jars and the replacement can also replace artifacts in more than one task

qwwdfsad (changes requested):

buildSrc/src/main/kotlin/VersionFile.kt
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import org.gradle.api.*
+import org.gradle.api.tasks.*
+import org.gradle.kotlin.dsl.*
+
+object VersionFile {
```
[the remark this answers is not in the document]
qwwdfsad: This comment should've been written by me in the first place; your change is rather the trigger "now 'git blame' won't help as well"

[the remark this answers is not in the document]
dkhalanskyjb: > Is `configure` part of `VersionFile`

As opposed to what alternative?

> What are the reason between splitting `VersionFile` and `version-file-conventions` into two files

`VersionFile` is "what to do", `conventions` is "go do it."

The `conventions` file is applies `VersionFile` globally to every subproject. `kotlinx-coroutines-debug` calls `VersionFile.configure` on its own, as it needs special treatment. If not for that, we could inline `VersionFile` in `conventions`.

kotlinx-coroutines-debug/build.gradle.kts
dkhalanskyjb: Sure, I understood this issue, I just didn't know the way to implement it in the `VersionFile` in a lazy manner. I'll try the `jarProvider.Configure` approach.

kotlinx-coroutines-debug/build.gradle.kts
dkhalanskyjb: This block has the following effect on the configurations that have any outgoing artifacts to begin with:
```
':kotlinx-coroutines-debug:apiElements' (before): [jar]
':kotlinx-coroutines-debug:apiElements' (after): [endResultJar]
':kotlinx-coroutines-debug:archives' (before): [jar]
':kotlinx-coroutines-debug:archives' (after): [endResultJar]
':kotlinx-coroutines-debug:mainSourceElements' (before): [DecoratingPublishArtifact_Decorated resources:directory::null, DecoratingPublishArtifact_Decorated java:directory::null, DecoratingPublishArtifact_Decorated src:directory::null]
':kotlinx-coroutines-debug:mainSourceElements' (after): [DecoratingPublishArtifact_Decorated resources:directory::null, DecoratingPublishArtifact_Decorated java:directory::null, DecoratingPublishArtifact_Decorated src:directory::null]
':kotlinx-coroutines-debug:runtimeElements' (before): [jar]
':kotlinx-coroutines-debug:runtimeElements' (after): [endResultJar]
':kotlinx-coroutines-debug:shadow' (before): [shadowJar]
':kotlinx-coroutines-debug:shadow' (after): []
':kotlinx-coroutines-debug:shadowRuntimeElements' (before): [shadowJar]
':kotlinx-coroutines-debug:shadowRuntimeElements' (after): []
':kotlinx-coroutines-debug:testResultsElementsForTest' (before): [DecoratingPublishArtifact_Decorated binary:directory::null]
':kotlinx-coroutines-debug:testResultsElementsForTest' (after): [DecoratingPublishArtifact_Decorated binary:directory::null]
```

kotlinx-coroutines-debug/build.gradle.kts
dkhalanskyjb: > Thank you for going the extra mile with rewriting the build script into kts!

It's nothing, I just couldn't figure out how to do it otherwise.

buildSrc/src/main/kotlin/VersionFile.kt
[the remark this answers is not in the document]
qwwdfsad: >As opposed to what alternative?

Keeping it on top-level. But it doesn't seem right for a variety of reasons.

>VersionFile is "what to do", conventions is "go do it."
> If not for that, ... 

Works for me, thanks

kotlinx-coroutines-debug/build.gradle.kts
dkhalanskyjb: I think I fixed this. I verified that the version files do get published for the JVM jars, and all configuration is very indirect and complicated, just the way we want it.

buildSrc/src/main/kotlin/VersionFile.kt
```diff
@@ -0,0 +1,29 @@
...
+ * for the better toolability. See #2941
+ */
+object VersionFile {
+    fun registerVersionFileTask(project: Project): TaskProvider<Task> = with(project) {
+        tasks.register("versionFileTask") {
+            val name = project.name.replace('-', '_')
+            val versionFile = project.layout.buildDirectory.file("$name.version")
+            outputs.file(versionFile)
+            doLast {
+                versionFile.get().asFile.writeText(project.version.toString())
```
ArchangelX360, line 19: Use of `project` in task execution phase breaks configuration cache compatibility.
Assign the `version` to a variable outside of the `doLast` block to avoid that, just like you did with `name`.

Probably, `with(project)` will also break it (unsure), it's redundant here your could to `project.tasks.register`.

kotlinx-coroutines-debug/build.gradle.kts
```diff
@@ -0,0 +1,112 @@
...
+        attributes(mapOf(
+            "Premain-Class" to "kotlinx.coroutines.debug.AgentPremain",
+            "Can-Redefine-Classes" to "true",
+            "Multi-Release" to "true"
+        ))
+    }
+    VersionFile.fromVersionFile(this, versionFileTask)
+}
+
+tasks.getByName("publishMavenPublicationToMavenLocal") {
```
ArchangelX360, line 91:
```suggestion
tasks.named("publishMavenPublicationToMavenLocal") {
```

[Configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html), `getByName` will force-create the task, while `named` won't.

kotlinx-coroutines-debug/build.gradle.kts
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import com.github.jengelman.gradle.plugins.shadow.tasks.*
+
+apply(plugin = "com.github.johnrengelman.shadow")
+
+// apply plugin to use autocomplete for Kover DSL
+apply(plugin = "org.jetbrains.kotlinx.kover")
```
ArchangelX360, line 10:
```suggestion
plugins {
   id("com.github.johnrengelman.shadow")
   id("org.jetbrains.kotlinx.kover") // apply plugin to use autocomplete for Kover DSL
}
```

I don't remember where I read this in the docs, but if I recall correctly, configuration optimisation can be done only if plugins are declared in the newer `plugins` block, instead of legacy `apply` block.

ArchangelX360 (commented):
I left some tips on fixing some configuration avoidance breakage and some config cache breakage. These are informative, and should probably not block this MR, but it's mainly to share good practise and speed up the build.

kotlinx-coroutines-debug/build.gradle.kts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// This is required for BlockHound tests to work, see https://github.com/Kotlin/kotlinx.coroutines/issues/3701
+tasks.withType<Test>().configureEach {
+    if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_13)) {
+        jvmArgs("-XX:+AllowRedefinitionToAddDeleteMethods")
+    }
+}
+
+val jar by tasks.getting(Jar::class) {
```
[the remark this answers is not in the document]
dkhalanskyjb: Thank you, the Archangel, but how do I make it work in this part?
```kotlin
    if (outgoing.artifacts.buildDependencies.getDependencies(null).contains(jar)) {
        outgoing.artifacts.removeIf { it.buildDependencies.getDependencies(null).contains(jar) }
```
I feel that, because of this rule, I need the exact value of the task in any case whenever I access any configuration.

kotlinx-coroutines-debug/build.gradle.kts
```diff
@@ -0,0 +1,112 @@
...
+    enabled = false
+}
+
+val shadowJar by tasks.getting(ShadowJar::class) {
```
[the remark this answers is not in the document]
dkhalanskyjb: Thank you, the Archangel, but here I have the same issues as with the previous comment.

buildSrc/src/main/kotlin/version-file-conventions.gradle.kts
```diff
@@ -0,0 +1,17 @@
+import org.gradle.api.tasks.bundling.*
+
+/* `kotlinx-coroutines-debug` configures `VersionFile` on its own when the corresponding task is created. */
+val invalidModules = listOf("kotlinx-coroutines-debug")
+
+configure(subprojects.filter {
```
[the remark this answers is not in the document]
dkhalanskyjb: Thank you, the Archangel. The wisdom of this advice is indisputable. Yet our land is too full of sin to be ready for such a revelation. The project is ripe with the filthy touch of subproject iteration, and making one place holy will but incite more chaos. The whole city must be burned to crisps before it rises as a temple (if you catch my meaning).

Though if you have quick access to some specific links to the required reading, sure, please share them. Maybe I'm wrong, and this could conveniently be done one iteration over subprojects at a time, then there's no reason not to do that.

kotlinx-coroutines-debug/build.gradle.kts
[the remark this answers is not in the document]
ArchangelX360: For now, I'll suggest to use `.get()` in the `configuration.all` block, so that it restrict the "issue" to the place where we struggle to find a solution while places that could already be solved are solved. E.g. tasks that depends on `jar` or `shadowJar` tasks will already stop eagerly configure them, which is conceptually already a win.

That said, I think there might be a way to write that `configuration.all` block in a lazier way that would not require a an early `get`, but I'll have to think. I'll comment here if I found a way.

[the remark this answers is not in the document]
dkhalanskyjb: 🙏

buildSrc/src/main/kotlin/version-file-conventions.gradle.kts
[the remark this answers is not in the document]
ArchangelX360: Hahahahaha.

I did not look into details yet, I may not have time today, but maybe if I'm bored this weekend, we never know!
But indeed, feel free to ignore my comment for now, it was mainly informative and we probably should nuke all of these cross-configuration at once.

dkhalanskyjb:
Using advice from @shanshin, I redid the whole approach to just pack the `module-info.class` into `shadowJar` in `doLast`.

qwwdfsad:
Good to go, great job!

PR #3953, opened 2023-11-23
Backport of K2 commits that should land in develop

qwwdfsad:


kotlinx-coroutines-test/common/src/TestCoroutineDispatchers.kt
```diff
@@ -93,7 +93,7 @@ private class UnconfinedTestDispatcherImpl(
 
     override fun isDispatchNeeded(context: CoroutineContext): Boolean = false
 
-    @Suppress("INVISIBLE_MEMBER")
+    @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
```
dkhalanskyjb, line 96: Here's a thought: won't the IDE complain that the suppression is redundant? Let's guard against accidentally removing it. Maybe something like this: https://github.com/Kotlin/kotlinx.coroutines/compare/develop...kotlin-community/k2/dev3#diff-ee02d7e5615b49f17adcbbb0cf3359a30222e91ff6a916c2cbc13043d0d79c59R84

If you agree, let's replace the commit that introduces this with the amended one: https://github.com/Kotlin/kotlinx.coroutines/commit/1030449bd84d2b5c9848f6664d85a62bbe95da0d

qwwdfsad: Done

dkhalanskyjb (approved):

PR #3957, opened 2023-11-29
Fix warnings in Native

dkhalanskyjb:


qwwdfsad (approved):

buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts
```diff
@@ -10,7 +10,7 @@ configure(subprojects) {
     if (name in sourceless) return@configure
     apply(plugin = "kotlinx-atomicfu")
     tasks.withType<KotlinCompile<*>>().configureEach {
-        val isMainTaskName = name == "compileKotlin" || name == "compileKotlinJvm"
+        val isMainTaskName = name.startsWith("compileKotlin")
```
qwwdfsad, line 13: I assume it does not match any test-related tasks

PR #3930, opened 2023-11-06
Change Gradle configuration to make sure "source" button is visible in all API docs (#3929)

SebastianAigner:
Fixes #3929

PR #3960, opened 2023-11-30
Update Gradle Dokka configuration to make sure `(source)` button is visible in all API docs

### LINKED ISSUE #3929: Add "source" button to API docs
Currently, there is no convenient way to navigate from the API docs to the actual code in the repository.

Compare other libraries, where Dokka links to the actual source / implementation of symbols: https://hoc081098.github.io/FlowExt/docs/0.x/-flow-ext/com.hoc081098.flowext/-delay-strategy/index.html

Screenshot 2023-11-06 at 16 03 05

This makes it harder for people who might want to contribute documentation improvements, clarifications, and the like. It'd be great if the button could be added!

whyoleg:
Fully fixes #3929 

After #3930 was merged there is no more `(source)` for MPP modules (core and test) because they are using different convention for sources.

In this PR dokka configuration for MPP modules is aligned with how sources are configured + it now doesn't use hardcoded names of sourceSets (previously, f.e. `jdk8` and `concurrent` were not configured, and so no `(source)` link were there)

BTW, I found, that there is some issue, that functions which are in `jdk8` can't resolve classes from `common` (in this case there should be `Flow`:
<details><summary>Screenshot</summary>
image
</details>
I will try to address this in another PR later, or create an issue at Dokka side, depending on the results of investigation

### DIFF
--- a/gradle/dokka.gradle.kts
+++ b/gradle/dokka.gradle.kts
@@ -37,14 +37,31 @@ tasks.withType(DokkaTaskPartial::class).configureEach {
     }
 }
 
-// Custom configuration for MPP modules
-tasks.withType(DokkaTaskPartial::class).configureEach {
-    dokkaSourceSets.configureEach {
-        sourceLink {
-            val relPath = rootProject.projectDir.toPath().relativize(projectDir.toPath())
-            localDirectory.set(projectDir.resolve("src"))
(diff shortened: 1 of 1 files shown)


qwwdfsad (approved):
Thanks!

PR #3965, opened 2023-12-01
Version 1.8.0-RC

dkhalanskyjb:


CHANGES.md
```diff
@@ -1,5 +1,15 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.8.0-RC
+
+* Implement the library for the Web Assembly (Wasm) (#3713). Thanks @igoriakovlev!
```
whyoleg, line 5: I think it's better to mention that only wasm-js is supported (as f.e. kotlin.serialization supports both wasm-js and wasm-wasi)

CHANGES.md
```diff
@@ -1,5 +1,15 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.8.0-RC
```
qwwdfsad, line 3: A few things are not mentioned: 

* #3897
* New (or, at least, specified) `SharedFlow` contracts
* Major Kotlin version update
* Update of Atomicfu that enables K/N transformation and decreases footprint of coroutines-heavy code

dkhalanskyjb: > Major Kotlin version update

Right now, it says

> * Kotlin version is updated to 1.9.21.

How should I additionally highlight it? "Before, it was 1.8.20?" or something like that?

> enables K/N transformation and decreases footprint of coroutines-heavy code

Is this a single Native-only change, or is the footprint decreased on other platforms as well?

qwwdfsad: >How should I additionally highlight it?

Maybe something like "Major Kotlin version update: ..."?
We typically don't do this, but here we have both transition to `1.9.x` *and* update to 1.9.20 where multiplatform (and all around its metadata etc.) is stable

>Is this a single Native-only change

Only K/N. Now all boxes (e.g. `AtomicInt` which is a reference type over an `Int`) is de-boxed where appropriate

qwwdfsad (approved):
Nice!

PR #3969, opened 2023-12-04
Simplify Dokka source link configuration

whyoleg:
Based on discussion: https://github.com/Kotlin/kotlinx-datetime/pull/323#discussion_r1412147614

Checked locally - content is identical for both approaches

### DIFF
--- a/gradle/dokka.gradle.kts
+++ b/gradle/dokka.gradle.kts
@@ -37,31 +37,12 @@ tasks.withType(DokkaTaskPartial::class).configureEach {
     }
 }
 
-fun GradleDokkaSourceSetBuilder.makeLinkMapping(projectDir: File) {
-    sourceLink {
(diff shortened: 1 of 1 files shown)


whyoleg:
Yeah, I've also compared the content locally. This is expected, but unobvious change :)
This is coming from calling `Style::toString()` somewhere in Dokka code...
This specific string is for [CommentTable](https://github.com/Kotlin/dokka/blob/17ad866ff2c85cdebc87fbca648432dd758b1ee1/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/pages/ContentNodes.kt#L433) style. And as it's an `object` we have hashCode in `toString`.

Overall this specific `Style` object should not be present in HTML at all, but just used for some logic branching...
So, it's safe change, I believe some time in the future we will cleanup such things.

whyoleg:
Sorry, accidentally pushed another changes to wrong branch - force pushed back to previous commit

PR #3970, opened 2023-12-04
Fix flakiness of MainDispatcherTestBase.testWithTimeoutContextDelayTi…

qwwdfsad:
…meout

* On our virtualized Windows CI everything is slow and sub-seconds delays got racy on a regular basis, leading to flaky builds
* Replace withContext(Dispatcher.Main) with launch(Dispatcher.Main, ATOMIC) {}.join() in order to avoid cancellation between dispatch and start race
* An alternative way is to guard it with isJavaAndWindows that effectively disables such test on Windows

### DIFF
--- a/kotlinx-coroutines-core/common/test/MainDispatcherTestBase.kt
+++ b/kotlinx-coroutines-core/common/test/MainDispatcherTestBase.kt
@@ -211,12 +211,12 @@ abstract class MainDispatcherTestBase: TestBase() {
             expect(1)
             assertFailsWith<TimeoutCancellationException> {
                 withTimeout(300) {
-                    withContext(Dispatchers.Main) {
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/test/MainDispatcherTestBase.kt
```diff
@@ -211,12 +211,12 @@ abstract class MainDispatcherTestBase: TestBase() {
             expect(1)
             assertFailsWith<TimeoutCancellationException> {
                 withTimeout(300) {
-                    withContext(Dispatchers.Main) {
+                    launch(Dispatchers.Main, start = CoroutineStart.ATOMIC) {
```
dkhalanskyjb, line 214: I'd add a comment, as the form of the code is unintuitive.

dkhalanskyjb (approved):

PR #3973, opened 2023-12-05
Update Dokka to 1.9.10

whyoleg:
Depends on knit release: https://github.com/Kotlin/kotlinx-knit/pull/59

`dokka`/`knit`/`knitCheck` tasks works fine locally with fixed `knit` published to `mavenLocal`

I've also done some minimal manual QA of produced result - haven't spotted any issues, only some style changes


IgnatBeresnev (approved):
I wonder why (if) it works now if only knit and the Dokka version in it were updated 0_o

qwwdfsad (approved):
Thanks!

Issue #3979, opened 2023-12-08
The asFlow() function doesn’t support back pressure management. If an ObservableSource is emitting a considerable volume of output at a high speed, the observable.asFlow() can block the thread, causing crashes.


Ericliu001:
**Describe the bug**

The asFlow() function doesn’t support back pressure management. If an ObservableSource is emitting a considerable volume of output at a high speed, the observable.asFlow() can block the thread, causing crashes.

This blocking behavior deviates from the subscription behavior in Rx, which never blocks, catching users by surprises.

The source code can be found at https://github.com/Kotlin/kotlinx.coroutines/blob/master/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt#L75
As we can see, the trySendBlocking() function is called inside onNext() on main thread, and let’s keep diving into trySendBlocking()

By default, trySendBlocking() invokes trySend() , which is lightweight, non-blocking; however, when trySend() does not return success, the code falls back to invoke send() , which is a blocking operation, resulting in unexpected App Not Responding crashes.

trySend() returns errors when the ChannelFlow is overflowing, and the default buffer size is 64 today, which could not handle rapid emissions from the ObservableSource.

[The danger of asFlow() when converting Rx to coroutines](https://eric-liu-developer.medium.com/the-danger-of-asflow-when-converting-rx-to-coroutines-42c115f1e846)

PR #3980, opened 2023-12-09
fix asFlow blocking thread

Ericliu001:
Fixes the issue https://github.com/Kotlin/kotlinx.coroutines/issues/3979 

Changes the ChannelFlow's buffer to UNLIMITED when asFlow() is called

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Builders.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Builders.kt
@@ -304,7 +304,7 @@ public fun <T> channelFlow(@BuilderInference block: suspend ProducerScope<T>.() 
  * > `awaitClose` block can be called at any time due to asynchronous nature of cancellation, even
  * > concurrently with the call of the callback.
  */
-public fun <T> callbackFlow(@BuilderInference block: suspend ProducerScope<T>.() -> Unit): Flow<T> = CallbackFlowBuilder(block)
+public fun <T> callbackFlow(capacity: Int = BUFFERED, @BuilderInference block: suspend ProducerScope<T>.() -> Unit): Flow<T> = CallbackFlowBuilder(capacity = capacity, block = block)
 
 // ChannelFlow implementation that is the first in the chain of flow operations and introduces (builds) a flow
 private open class ChannelFlowBuilder<T>(
--- a/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
+++ b/reactive/kotlinx-coroutines-rx2/src/RxConvert.kt
@@ -72,11 +72,17 @@ public fun <T : Any> Deferred<T>.asSingle(context: CoroutineContext): Single<T> 
  * resulting flow to specify a user-defined value and to control what happens when data is produced faster
  * than consumed, i.e. to control the back-pressure behavior. Check [callbackFlow] for more details.
  */
-public fun <T: Any> ObservableSource<T>.asFlow(): Flow<T> = callbackFlow {
+public fun <T : Any> ObservableSource<T>.asFlow(): Flow<T> = callbackFlow(capacity = Channel.UNLIMITED) {
     val disposableRef = AtomicReference<Disposable>()
     val observer = object : Observer<T> {
-        override fun onComplete() { close() }
-        override fun onSubscribe(d: Disposable) { if (!disposableRef.compareAndSet(null, d)) d.dispose() }
+        override fun onComplete() {
+            close()
+        }
+
+        override fun onSubscribe(d: Disposable) {
+            if (!disposableRef.compareAndSet(null, d)) d.dispose()
+        }
+
         override fun onNext(t: T) {
             /*
              * Channel was closed by the downstream, so the exception (if any)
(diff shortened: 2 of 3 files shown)


dkhalanskyjb:
This PR does not respect the contribution guidelines (https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs), but more importantly, it's also not the change we'd like to introduce, as this can easily be expressed already by applying `buffer()`.

Ericliu001:
> This PR does not respect the contribution guidelines (https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs), but more importantly, it's also not the change we'd like to introduce, as this can easily be expressed already by applying `buffer()`.

Having to apply `buffer()` as the consumer of the API requires prior knowledge of this potential blocking behavior, i believe the framework should have buffer(UNLIMITED) by default. 

Shall we still integrate the UNLIMITED buffer somewhere?

qwwdfsad:
Among other things, `UNLIMITED` buffer means unlimited queue size and unlimited memory consumption that we would very much like to avoid

Ericliu001:
> Among other things, `UNLIMITED` buffer means unlimited buffer size and unlimited memory consumption that we would very much like to avoid.

I think in this situation it's very "choosing your poison": either blocking the thread or consuming more memory. In Rx, it has chosen the later https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/subjects/BehaviorSubject.java 

```
 if (emitting) {
                        AppendOnlyLinkedArrayList<Object> q = queue;
                        if (q == null) {
                            q = new AppendOnlyLinkedArrayList<>(4);
                            queue = q;
                        }
                        q.add(value);
                        return;
                    }
```

psteiger:
Related to #3282

dkhalanskyjb:
> Having to apply buffer() as the consumer of the API requires prior knowledge of this potential blocking behavior

Likewise, if there was no blocking behavior, your message could be rephrased as "having to apply `buffer()` as the consumer of the API requires prior knowledge of this potential unbounded memory consumption." In any case, someone will miss something by not reading the documentation.

> dangerous, especially on single-threaded, immediate-dispatchers contexts.

What's good about these contexts is that usually, the problem is visible immediately during development and can be fixed. On the other hand, an OOM can be hidden, silently sneaking into production.

psteiger:
@dkhalanskyjb I think it's fair to notice what I believe is an inconsistency in the way kotlinx.coroutines handles `Observable` <> coroutines interop, then: 

`ObservableSource<T>.collect { }` already uses an `UNLIMITED` channel + `trySend` under the hood, which is what I would also expect for `asFlow().collect { }` (and which I believe is a saner and safer default than `BUFFERED` + `trySendBlocking`:

```
public suspend inline fun <T> ObservableSource<T>.collect(action: (T) -> Unit): Unit =
    toChannel().consumeEach(action)

@PublishedApi
internal fun <T> ObservableSource<T>.toChannel(): ReceiveChannel<T> {
    val channel = SubscriptionChannel<T>()
    subscribe(channel)
    return channel
}

private class SubscriptionChannel<T> :
    BufferedChannel<T>(capacity = Channel.UNLIMITED), Observer<T>, MaybeObserver<T>
{
  // ...
    override fun onNext(t: T & Any) {
        trySend(t) // Safe to ignore return value here, expectedly racing with cancellation
    }
  // ...
}
```

dkhalanskyjb:
@psteiger, here are the incompatible things:
* The way of kotlinx-coroutines' `Flow`, where we mostly use channels with limited buffers,
* And the way RxJava (in particular, `ObservableSource`) works, which is either to use unlimited stack space and process all emissions immediately or to use unbounded buffers.

There are two intended use cases for `kotlinx-coroutines-rx2`: either someone knows RxJava well but needs to interface with coroutines, or they know `kotlinx-coroutines` well and need to interface with RxJava.

When converting an RxJava entity to a kotlinx-coroutines entity, we must introduce *some* inconsistency: either between a typical `Flow` and the one from RxJava, or between an RxJava entity and the `Flow` received from it. We chose the latter approach: when converting to a kotlinx-coroutines entity, you now subscribe to how we do things here. That's the behavior that the people well-versed in `kotlinx-coroutines` would expect. `ObservableSource.collect`, on the other hand, is not fully in the kotlinx-coroutines world, it's not a converter but an extension function on `ObservableSource`, so it's kept close to what RxJava folks would want.

Luckily, as stated in https://github.com/Kotlin/kotlinx.coroutines/issues/3979#issuecomment-1850154108, the inconsistency between `Flow` and RxJava is easy to mitigate. If you want to change your failure mode, you have a way to do so.

psteiger:
@dkhalanskyjb Thanks for sharing the design details, and overall I find them reasonable. But I still maintain the position for some change here (either in impl, or in docs).

>  If you want to change your failure mode, you have a way to do so.

No one wants to fail with a deadlock, but unfortunately it's far **too easy** to use `ObservableSource.asFlow()` in specific contexts without realizing the risk. It's a default that is dangerous in specific contexts; versus some other potential default that is safe against deadlocks in all contexts.

If the end decision is to keep the implementation, I'd call for an emphasized warning on `asFlow` docs on which particular cases `asFlow` usage can be dangerous.

Mustafaubaid4:
Thanks w

PR #3986, opened 2023-12-13
Ensure that non-local returns work properly in inline functions

### LINKED ISSUE #3985: Workround #3881 breaks non-local returns in 1.8.0-RC
I recently upgraded the lib of one of my apps from `1.7.3` to `1.8.0-RC`. I saw that something wasn't working a day later in my Grafana. After some investigations and deductions, I figured out that this bump caused functions not being executed, something had to do with locks.

After some time, I very professionally (using lots of `println`s) debugged the whole code related to that feature. I found the bug out:

```kt
fun test() {
  mutex.withLock {
    return
  }
}
```

This doesn't release the lock. Why? Because someone decided to remove `finally` of that damn thing in #3881 .

image
image
image

https://kotlinlang.org/docs/inline-functions.html this even shows how to make a lock function which works. This will break a lot of code which relies on of having a finally out there.

dkhalanskyjb:
Also, ensure that the workaround for KT-58685 still works. It turns out, it was enough to move the `return` a bit to fix the issue.

Fixes #3985


dkhalanskyjb:
I intentionally chose the parent commit of this so that the miscompilation still happens in our test environment. After the CI passes, I'll rebase it on top of the current develop.

qwwdfsad (approved):

Issue #3968, opened 2023-12-02
1.8.0-RC Could not find "org.jetbrains.kotlinx:atomicfu-cinterop-interop"


martinbonnin:
Trying to build an `iosArm64` target with version `1.8.0-RC`, I get hit by this error:

```
> Task :compileKotlinIosArm64 FAILED
e: Could not find "org.jetbrains.kotlinx:atomicfu-cinterop-interop" in [/Users/mbonnin/git/test-coroutines, /Users/mbonnin/.konan/klib, /Users/mbonnin/.konan/kotlin-native-prebuilt-macos-aarch64-1.9.21/klib/common, /Users/mbonnin/.konan/kotlin-native-prebuilt-macos-aarch64-1.9.21/klib/platform/ios_arm64]
```

Reproducer [here](https://github.com/martinbonnin/test-coroutines). Am I missing something?

martinbonnin:
[Explicitely adding `atomicfu` to the dependencies](https://github.com/martinbonnin/test-coroutines/compare/main...fix?body=&expand=1&title=explicitely+import+atomicfu) fixes the problem but sounds like it might not be needed?

PS: also tried disabling konan caches like in [KT-44884](https://youtrack.jetbrains.com/issue/KT-44884/) but didn't seem to help much

hoangchungk53qx1:
try this
https://github.com/Kotlin/kotlinx.coroutines/issues/3305

dkhalanskyjb:
Citing @mvicsokolova's words:

---

The problem is on kotlinx.atomicfu side: the new version of atomicfu-gradle-plugin does not provide coroutines with a transitive dependency to atomicfu anymore. Though it's required because coroutines use inline declarations from kotlinx-atomicfu.The WA for now is adding atomicfu implementation dependency for the native sourceSets:

```kotlin
val nativeMain by creating {
  dependencies {
    implementation("org.jetbrains.kotlinx:atomicfu:0.23.1")
  }
}
```

And I'll prepare the new release of atomicfu where this dependency is provided by atomicfu-gradle-plugin.

PR #3996, opened 2023-12-21
Work around #3968, start running integration tests for Native

dkhalanskyjb:
I think we should keep #3968 open until it's fixed properly.


dkhalanskyjb:
Now that I think about it, nothing should change for the end user when we fix this and remove the workaround, so maybe we should close #3968 after all.

Issue #3770, opened 2023-06-01
Gradually introduce SubclassOptInRequired into our codebase


qwwdfsad:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-subclass-opt-in-required/

Historically, we have multiple interfaces that are not designed for external implementation, , but are not `sealed` for various reasons (including historical ones):

* `Job` should not be extended externally 
* `Deferred` and `CompletableDeferred` should not be extended externally 
* `Channel` _probably_ should not be extended externally 
     * It is not mentioned in the documentation, and it's impossible to fulfil the select's contract with a hand-rolled implementation
     * Yet it is possible to implement it with a delegation, potentially closing the possibility of adding new abstract members to the interface
---
* `Flow`, `SharedFlow`, and `StateFlow` are unstable/experimental for inheritance 
* `CloseableCoroutineDispatcher` is unstable for inheritance
* `CancellableContinuation` is an interface not ready to be extended

It would be nice to adopt the `SubclassOptInRequired` (available from 1.9.0) in the library to provide explicit programmatic warnings/errors and to prevent issues like https://github.com/Kotlin/kotlinx.coroutines/issues/3729 

For "prohibited to implement" interfaces we can introduce the new annotation with a warning (or maybe even an error) and, in the next major release, make the annotation `internal` itself, so it would be much harder to opt-in

PR #4010, opened 2024-01-09
Rework CompletionHandler to avoid subclassing a functional type

dkhalanskyjb:
Workaround for KT-64075


dkhalanskyjb:
Status: it seems like the fix should be on the side of Lincheck. We probably won't have to change the approach taken in this PR.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -201,6 +201,11 @@ public interface CancellableContinuation<in T> : Continuation<T> {
     public fun resume(value: T, onCancellation: ((cause: Throwable) -> Unit)?)
 }
 
+internal fun <T> CancellableContinuation<T>.invokeOnCancellation(handler: CancelHandler) = when (this) {
+    is CancellableContinuationImpl -> invokeOnCancellationInternal(handler)
+    else -> invokeOnCancellation(handler::invoke)
```
qwwdfsad, line 206: Could you please elaborate on what code path hits this branch?

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -628,14 +627,17 @@ private object Active : NotCompleted {
  * Base class for all [CancellableContinuation.invokeOnCancellation] handlers to avoid an extra instance
  * on JVM, yet support JS where you cannot extend from a functional type.
  */
-internal abstract class CancelHandler : CancelHandlerBase(), NotCompleted
+@InternalCoroutinesApi
```
qwwdfsad, line 630: Do we need this annotation on already `internal` declaration?

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
dkhalanskyjb: The code path where someone implemented their own `CancellableContinuation`, against all odds. I *could* throw an exception here instead just to be mean, but why?

qwwdfsad: In a few places, we already rely on the fact that `CancellableContinuation` can be casted to `CancellableContinuationImpl` and, I can bet that, our other codepaths are neither ready nor tested for customly implemented `CC` instances. I would prefer fail fast here.

Also, added `CancellableContinuation` to https://github.com/Kotlin/kotlinx.coroutines/issues/3770

dkhalanskyjb: If we don't support third-party implementations, this would mean we can also remove `invokeOnCompletion(onCancelling, invokeImmediately, handler)`. Except ktor apparently uses it: https://grep.app/search?q=invokeOnCompletion%28onCancelling

qwwdfsad: Not sure I follow, could you please elaborate?
One is the implementation of the interface that e.g. doesn't work with channels and mutex, and the other is the use of an internal API (with no public alternatives).

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -361,7 +361,8 @@ internal fun Job.invokeOnCompletion(
     handler: InternalCompletionHandler
 ): DisposableHandle = when (this) {
     is JobSupport -> invokeOnCompletionInternal(onCancelling, invokeImmediately, handler)
-    else -> invokeOnCompletion(onCancelling, invokeImmediately, handler::invoke)
+    is NonCancellable -> NonDisposableHandle
+    else -> throw UnsupportedOperationException("third-party implementation of Job is not supported")
```
qwwdfsad, line 365: This one might reasonably fail.
Could you please see if the pattern from https://github.com/Kotlin/kotlinx.coroutines/issues/2423 keeps working?

dkhalanskyjb: It doesn't, the CI showed this. Reverted the change to `Job`.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
dkhalanskyjb: It was about `Job`, and it turns out that branch is still needed, so the point is moot.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -625,19 +621,40 @@ private object Active : NotCompleted {
...
  */
-internal abstract class CancelHandler : CancelHandlerBase(), NotCompleted
-
-// Wrapper for lambdas, for the performance sake CancelHandler can be subclassed directly
-private class InvokeOnCancel( // Clashes with InvokeOnCancellation
-    private val handler: CompletionHandler
-) : CancelHandler() {
-    override fun invoke(cause: Throwable?) {
-        handler.invoke(cause)
+internal interface CancelHandler: NotCompleted {
```
qwwdfsad, line 633: Nice doc 👍

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -625,19 +621,40 @@ private object Active : NotCompleted {
...
+    /**
+     * Signal cancellation.
```
qwwdfsad, line 635: Probably it's "Signal**s**" (as per our general rule that is not explicitly stated anywhere)

kotlinx-coroutines-core/common/src/CompletionHandler.common.kt
```diff
@@ -1,43 +1,69 @@
...
+ * Functions used for this must be fast, non-blocking, and thread-safe.
+ * This handler can be invoked concurrently with the surrounding code.
+ * There is no guarantee on the execution context in which the function is invoked.
+ *
  * **Note**: This type is a part of internal machinery that supports parent-child hierarchies
  * and allows for implementation of suspending functions that wait on the Job's state.
  * This type should not be used in general application code.
- * Implementations of `CompletionHandler` must be fast and _lock-free_.
  */
+// TODO: deprecate. This doesn't seem better than a simple function type.
```
qwwdfsad, line 30: Should we?

I see this type alias (or maybe even I would interpret it as an opaque type if it was possible) as a perfect fit -- it is a functional type, but with very specific semantics that is a property of this "type", not of the callsite (`invokeOn*` and coroutine builders), and the documentation to this type is a perfect place to spell it out. 

WDYT? As usual, this particular endpoint might be not the most important one, but it's nice to align our mental models here

kotlinx-coroutines-core/common/src/CompletionHandler.common.kt
```diff
@@ -1,43 +1,69 @@
...
+ * The reason this abstraction exists is to allow providing a readable [toString] in the list of completion handlers
+ * as seen from the debugger.
+ * Use [UserSupplied] to create an instance from a lambda.
+ * We can't avoid defining a separate type, because on JS, you can't inherit from a function type.
+ *
+ * @see CancelHandler for a very similar interface, but used for handling cancellation and not completion.
+ */
+internal interface InternalCompletionHandler {
+    /**
+     * Signal completion.
```
qwwdfsad, line 45: The same here

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
qwwdfsad, line 633:
```suggestion
internal interface CancelHandler : NotCompleted {
```

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
```diff
@@ -625,19 +621,40 @@ private object Active : NotCompleted {
...
+     *
+     * This function:
+     * - Does not throw any exceptions.
+     *   Violating this rule in an implementation leads to [handleUncaughtCoroutineException] being called with a
+     *   [CompletionHandlerException] wrapping the thrown exception.
+     * - Is fast, non-blocking, and thread-safe.
+     * - Can be invoked concurrently with the surrounding code.
+     * - Can be invoked from any context.
+     */
+    fun invoke(cause: Throwable?)
```
qwwdfsad, line 645: Would be nice to mention the `cause` semantics and when it's a null?

qwwdfsad (approved):
Great job, thank you for taking care of this nuisance.

Looks good, all my remarks are trivialities about the doc

kotlinx-coroutines-core/common/src/CompletionHandler.common.kt
dkhalanskyjb: Here's my reasoning behind the `TODO`.

tl;dr: this function's contract is defined by the places that accept the function, not the other way around, so these other places should define what they expect. Also, the latest commit that fixed the explanation of what the argument means, and this type looks even more unnatural now.

Long version:

I know of two use cases for binding an arbitrary function to a name:
* Passing it somewhere so that that other place calls it,
* Storing it somewhere so that someone calls it later.

For the second use case, introducing a type alias is clearly beneficial: when you have a `var handler: CompletionHandler`, there is *some* contract that should connect the parties. If I see a `CompletionHandler`, I may only invoke it so or so, and return, it promises me this and that, and vice versa. But what we have here is purely the first use case.

> but with very specific semantics that is a property of this "type", not of the callsite

This surprises me greatly: I think it's exactly the opposite.

The contract on the handler is exactly the combination of the contracts of the functions that accept the handler. Here's the deciding factor for me: if the behavior of `invokeOnCompletion` changed and the contract on `CompletionHandler` could somehow be relaxed, we wouldn't define a `RelaxedCompletionHandler` (supertype of `CompletionHandler`), we would just change the documentation of `CompletionHandler`. So, `CompletionHandler` is a "fake" type, a product of circumstances, only useful for passing it to the specific call sites.

kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt
dkhalanskyjb: Done. Now the abstraction of `CompletionHandler` looks even more strange.

kotlinx-coroutines-core/common/src/CompletionHandler.common.kt
qwwdfsad: Thanks for the elaborate answer.

>we would just change the documentation of CompletionHandler. 
>The contract on the handler is exactly the combination of the contracts of the functions that accept the handler.

Indeed. It appears I just think about this one inside out -- its definition/contract is driven bu the need of its "clients" (`invokeOnCompletion`), yet the moment the contract is established -- it's the property of *`CompletionHandler`* that it does not throw, not the vise versa -- "in general, CompletionHandler is whatever you want it to be, but it shouldn't throw if passed to Job.invokeOn...()"

qwwdfsad: It's quite close. I don't see it fit e.g. for an arbitrary function, but here we have a multitude of functions that all accept the same kind of parameter with the same kind of constraints. These constraints are significant enough to document, and broad enough to get tired of repeating it (NB: here, I focus not on "repeating the text" but on "repeating the conceptual complexity of the entity accepting this parameter and draining our complexity budget").

It's just lightweight and specific enough so it doesn't deserve a dedicated nominal type.

(please note that this discussion is not a blocker of this PR by any means)

dkhalanskyjb: > here we have a multitude of functions that all accept the same kind of parameter with the same kind of constraints

Let's consider two scenarios separately:
* There's a `CompletionHandler` and, separately, a `CancellationHandler`;
* The two types of handlers stay merged into one, just because they have the same type.

I'm confident the second option is a no-go because the constraints and the behavior are actually different between `invokeOnCompletion` and `invokeOnCancellation`. Knowing just the `CancellationHandler` type, you can't make a judgement about how it will behave, you need to know which of the two endpoints ends up actually receiving the handler.

The first option also seems strange to me: then, the contract on `CompletionHandler` would simply completely mirror the contract on `invokeOnCompletion`, and I don't see the benefits of introducing a separate entry point when we can just say "will be passed to `invokeOnCompletion` and must satisfy the same requirements" and in fact, *must* either do so or direct people to the documentation of `CompletionHandler` explicitly, because no one realistically is going to notice, "Aha, this function accepts not just any lambda but one with a special type, let me read the docs."

PR #4018, opened 2024-01-21
Extract example output from comment to text block

zUniQueX:
Recently I've come across the flow docs and noticed one example output isn't displayed in the online version.

In the source file the output is present but hidden inside an XML comment.

I've extracted the output to a text block to display it in the online docs.


qwwdfsad (approved):
Thanks!

PR #4020, opened 2024-01-23
Fix "Type mismatch: inferred type is String? but String was expected"

woainikk:


dkhalanskyjb:
Historical note: the behavior for which we need this workaround was introduced in https://github.com/jetbrains/kotlin/commit/c9e328528e08eff5e0bd5742d41b4580178d59aa

dkhalanskyjb (approved):

PR #4028, opened 2024-01-30
Disable DebugProbes.enableCreationStackTraces by default

### LINKED ISSUE #3783: Set the default value of DebugProbes.enableCreationStackTraces to false
`DebugProbes.enableCreationStackTraces` is the property of the debugger, indicating that the stacktrace of the coroutine *creation* site has to be captured.

This is a performance-sensitive operation (e.g. `fillInStackTrace` is typically *the* heaviest operation in the coroutines lifecycle), incurring significant (double digit) overhead on *production* systems with little to no benefits. On the other side, the creation stacktrace itself is not asynchronous -- it doesn't attempt to recover the call site stacktrace, meaning that coroutines launched from a `suspend` function are unlikely to have a meaningful creation stacktrace.

Most of the known probes API consumers disable it; moreover, even our documentation recommends doing so.

My suggestion is to switch it off by default and leave as a legacy and/or advanced option

qwwdfsad:
* Adjust tests and the documentation
* Rationale: this option implies significant and non-trivial overhead yet its utility is unclear (e.g. we got a multitude of signals that this option brings no use and none that it's useful even when asked explicitly)

Fixes #3783


qwwdfsad:
I've tested it locally -- IDEA debugger keeps working with this change.
The next step would be deprecating this flag

kotlinx-coroutines-debug/src/DebugProbes.kt
```diff
@@ -65,12 +65,11 @@ public object DebugProbes {
 
     /**
      * Whether coroutine creation stack traces should be captured.
-     * When enabled, for each created coroutine a stack trace of the current
-     * thread is captured and attached to the coroutine.
+     * When enabled, for each created coroutine a stack trace of the current thread is captured and attached to the coroutine.
      * This option can be useful during local debug sessions, but is recommended
-     * to be disabled in production environments to avoid stack trace dumping overhead.
+     * to be disabled in production environments to avoid performance overhead of keeping debug probes enabled.
```
dkhalanskyjb, line 70: I understood the previous version better. With `enableCreationStackTraces`, you copy a bunch of stackframe descriptions with many strings around *on every coroutine creation*, which is expensive, ok, this makes sense. What does the new text mean? This option changes the behavior *while* the debug probes are enabled, it doesn't force them to be enabled longer than needed, right?

dkhalanskyjb (commented):
Here are some more changes that I think are in line with the spirit of the PR: https://gist.github.com/dkhalanskyjb/e04623b28fc274db1c52f5ea2d491ce7

qwwdfsad:
Sure, mind pushing them?

dkhalanskyjb (approved):

PR #4033, opened 2024-02-06
Instruct Proguard to keep all the coroutine service implementations

### LINKED ISSUE #4025: Proguard will remove SwingDispatcherFactory (MainDispatcherFactory) on JVM
**Describe the bug**

If you use Proguard on JVM (e.g. as part of a Compose/Desktop application as per [compose docs](https://github.com/JetBrains/compose-multiplatform/tree/master/tutorials/Native_distributions_and_local_execution#minification--obfuscation)), then the MainDispatcherFactory is removed. When the application runs, it crashes because there is no main dispatcher.

The following Proguard rules are required:
```
-keep class kotlinx.coroutines.internal.MainDispatcherFactory { *; }
-keep class kotlinx.coroutines.swing.SwingDispatcherFactory { *; }

```

Proguard on Desktop/JVM doesn't have automatic rule consumption like Android, so documenting the requirement e.g. here https://github.com/Kotlin/kotlinx.coroutines#r8-and-proguard is critical.

**Provide a Reproducer**

This can be reproduced from this branch of our code:
https://github.com/UstadMobile/UstadMobile/tree/dev-desktop-conveyor

Remove the above lines from app-desktop/compose-desktop.pro , and the crash will be observed.

Alternatively:
1. Create a new Compose JVM/Desktop app
2.  Enable obfuscation as per [compose docs](https://github.com/JetBrains/compose-multiplatform/tree/master/tutorials/Native_distributions_and_local_execution#minification--obfuscation)) 
3. Run it.

dkhalanskyjb:
With this change, `kotlinx-coroutines-android` no longer needs to specify for Proguard and the older R8 versions that the Main dispatcher and the uncaught exception handler must be kept; instead, the rules shipped with the coroutines library do that for every implementation now.

Fixes #4025


dkhalanskyjb:
This is a draft because I haven't yet thoroughly tested it's working.

dkhalanskyjb:
Maybe I didn't actually fix the issue or achieve anything here. https://github.com/Kotlin/kotlinx.coroutines/issues/4025#issue-2103585503 mentions that

> Proguard on Desktop/JVM doesn't have automatic rule consumption like Android

Interesting, I'll need to look into this. I don't understand the difference from Proguard's point of view.

dkhalanskyjb:
Without Proguard supporting this in the Gradle plugin outside Android, we'd better not touch this needlessly.

PR #4036, opened 2024-02-07
Remove the java.lang.management dependency from the debug probes

kjnsn:
This appears to be unused, and prevents debug probes from being imported on android

### DIFF
--- a/kotlinx-coroutines-debug/src/DebugProbes.kt
+++ b/kotlinx-coroutines-debug/src/DebugProbes.kt
@@ -5,7 +5,6 @@ package kotlinx.coroutines.debug
 import kotlinx.coroutines.*
 import kotlinx.coroutines.debug.internal.*
 import java.io.*
-import java.lang.management.*
 import kotlin.coroutines.*
 
 /**


dkhalanskyjb:
It's not that simple, unfortunately. This `import` only affects the name resolution in the file, it doesn't prevent the debug probes from working on Android on its own.

kjnsn:
It looks like the `import` doesn't affect name resolution in the file, as the build and tests worked correctly. 

Should the documentation be updated in the README? 

```
Android runtime does not support Instrument API necessary for kotlinx-coroutines-debug to function, triggering java.lang.NoClassDefFoundError: Failed resolution of: Ljava/lang/management/ManagementFactory;, and it is not possible to use coroutine debugger along with Android emulator.
```

qwwdfsad:
>It looks like the import doesn't affect name resolution in the file, as the build and tests worked correctly.

This import affects links resolution in the KDoc. The build won't fail because we don't have Dokka `failOnWarning` mode enabled (which is a completely different story)

PR #4038, opened 2024-02-08
Ensure that flow operators propagate the cancellation exceptions

### LINKED ISSUE #4035: Flow hangs even when using terminal operators
**Describe the bug**

A flow that uses `take(1)`, `onCompletion { if (it == null) awaitCancellation() }` and `first()` can hang forever instead of returning the first value.

Interestingly, removing `take(1)` will cause the flow to behave as expected.

**Provide a Reproducer**
Here is such an example:
```kotlin
flowOf(1)
    .take(1)
    .onCompletion { if (it == null) awaitCancellation() }
    .first()
```

This code is expected to return `1`, instead if suspends forever. If you remove `take(1)` it returns `1` as expected.

dkhalanskyjb:
Before this change, it could happen that some size-limiting operators upstream swallowed the requests to limit the flow size emitted by the operators downstream.

This could cause `onCompletion` calls between these operators to incorrectly report that the flow was not in fact limited by the downstream operators.

Additionally, in the presence of additional size-limiting operators in the chain, `first` and `single` and their variants could exhibit incorrect behavior where emitting a value from `onCompletion` would overwrite their output.

Fixes #4035

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/internal/FlowExceptions.common.kt
+++ b/kotlinx-coroutines-core/common/src/flow/internal/FlowExceptions.common.kt
@@ -9,11 +9,11 @@ import kotlinx.coroutines.flow.*
  * This exception can be safely ignored by non-terminal flow operator if and only if it was caught by its owner
  * (see usages of [checkOwnership]).
  */
-internal expect class AbortFlowException(owner: FlowCollector<*>) : CancellationException {
-    public val owner: FlowCollector<*>
+internal expect class AbortFlowException(owner: Any) : CancellationException {
+    val owner: Any
 }
 
-internal fun AbortFlowException.checkOwnership(owner: FlowCollector<*>) {
+internal fun AbortFlowException.checkOwnership(owner: Any) {
     if (this.owner !== owner) throw this
 }
 
--- a/kotlinx-coroutines-core/common/src/flow/internal/FlowExceptions.common.kt
+++ b/kotlinx-coroutines-core/common/src/flow/internal/FlowExceptions.common.kt
@@ -4,10 +4,11 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
 
 /**
- * This exception is thrown when operator need no more elements from the flow.
- * This exception should never escape outside of operator's implementation.
+ * This exception is thrown when an operator needs no more elements from the flow.
+ * The operator should never allow this exception to be thrown past its own boundary.
  * This exception can be safely ignored by non-terminal flow operator if and only if it was caught by its owner
  * (see usages of [checkOwnership]).
+ * Therefore, the [owner] parameter must be unique for every invocation of every operator.
  */
 internal expect class AbortFlowException(owner: Any) : CancellationException {
     val owner: Any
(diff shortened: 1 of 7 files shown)


qwwdfsad:
Could you please double-check my understanding?

At first glance, the behavioural difference in this snippet 
```
flowOf(1)
    .take(1)
    .onEach {  } // 1
    .onCompletion { println("??? $it") }
    .first()
```
with `1` being commented out or not just doesn't make any sense. I tried to figure it out by just looking at the commit message, and it was unclear, thus decided to understand the root cause before looking at the solution to avoid spoiling the ~fun~ review.

It seems to me that the root cause is the following:

1) `take` uses `FlowCollector` that it receives in an argument as an ownership token, and it checks the ownership in `catch` block
2) The problem is that `FlowCollector` is not unique to the `take` itself and, in fact, depends on the downstream operator. It might be a collector from `onEach`, or it might be an `onCompletion` collector.
3) `onCompletion` collector, which is `collectWhile` collector, throws an `AbortFlowException` and uses itself as an ownership token.
4) According to `2`, ownership check in `take` might pass or might not pass, leading to the observed behaviour

So the root cause is straightforward -- we cannot use `FlowCollector` from an argument as an ownership token

kotlinx-coroutines-core/common/src/flow/internal/FlowExceptions.common.kt
```diff
@@ -9,11 +9,11 @@ import kotlinx.coroutines.flow.*
  * This exception can be safely ignored by non-terminal flow operator if and only if it was caught by its owner
  * (see usages of [checkOwnership]).
  */
-internal expect class AbortFlowException(owner: FlowCollector<*>) : CancellationException {
-    public val owner: FlowCollector<*>
+internal expect class AbortFlowException(owner: Any) : CancellationException {
+    val owner: Any
```
qwwdfsad, line 13: Maybe it's worth imposing a few restrictions on `owner` in the documentation.

kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
```diff
@@ -46,6 +46,7 @@ public fun <T> Flow<T>.dropWhile(predicate: suspend (T) -> Boolean): Flow<T> = f
 public fun <T> Flow<T>.take(count: Int): Flow<T> {
     require(count > 0) { "Requested element count $count should be positive" }
     return flow {
+        val ownershipMarker = Any()
```
qwwdfsad, line 49: Micro-optimization: we definitely cannot use `this` flow instance, and we cannot use `FlowCollector` instance.

Maybe we can use something like `currentCoroutineContext` or the current continuation instance? 
Don't give it too much thought though, if it's not obvious, then probably it's not worth doing it here

dkhalanskyjb: For me, it's not obvious. How does one obtain the current continuation instance without `suspendCoroutine` (which inpacts the state machine of the code)?

qwwdfsad (approved):
Nice one!

PR #4042, opened 2024-02-12
Small documentation fixes

dkhalanskyjb:


kotlinx-coroutines-core/common/src/Deferred.kt
```diff
@@ -33,17 +33,37 @@ import kotlinx.coroutines.selects.*
...
+     * if the [Deferred] completed successfully, [await] will return the same value every time;
+     * if the [Deferred] completed exceptionally, [await] will rethrow the same exception.
      *
-     * This function can be used in [select] invocation with [onAwait] clause.
-     * Use [isCompleted] to check for completion of this deferred value without waiting.
+     * This suspending function is itself cancellable: if the [Job] of the current coroutine is cancelled or completed
+     * while this suspending function is waiting, this function immediately resumes with [CancellationException].
+     *
+     * This means that [await] can throw [CancellationException] in two cases:
+     * * if the of the coroutine in which [await] got cancelled,
```
lukellmann, line 47:
```suggestion
     * * if the coroutine in which [await] was called got cancelled,
```
or something like this

kotlinx-coroutines-core/jvm/src/Builders.kt
```diff
@@ -10,8 +10,22 @@ import kotlin.coroutines.*
...
+ * ```
+ * suspend fun loadConfiguration() {
+ *     // DO NOT DO THIS:
+ *     val data = runBlocking { // <- redundant and blocks the thread, do not do that
+ *         fetchConfigurationData() // suspending function
+ *     }
+ * ```
+ *
+ * Here, instead of releasing the thread on which `loadConfiguration` runs if `fetchConfigurationData` suspends, it will
+ * block, potentially leading to thread starvation issues.
```
lukellmann, line 28: should this update also be applied to the [`concurrent` and `native`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html) source sets?

kotlinx-coroutines-core/common/src/Deferred.kt
dkhalanskyjb: Yes, thanks!

kotlinx-coroutines-core/jvm/src/Builders.kt
dkhalanskyjb: Done.

kotlinx-coroutines-core/common/src/Deferred.kt
```diff
@@ -33,17 +33,37 @@ import kotlinx.coroutines.selects.*
...
-     * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
-     * immediately resumes with [CancellationException].
      * There is a **prompt cancellation guarantee**. If the job was cancelled while this function was
      * suspended, it will not resume successfully. See [suspendCancellableCoroutine] documentation for low-level details.
      *
-     * This function can be used in [select] invocation with [onAwait] clause.
-     * Use [isCompleted] to check for completion of this deferred value without waiting.
+     * This function can be used in [select] invocations with an [onAwait] clause.
+     * Use [isCompleted] to check for completion of this deferred value without waiting, and
+     * [join] to wait for completion without returning the result.
```
qwwdfsad, line 66: Probably worth improving `onAwait` as well -- remove the failure part and reference `await`

kotlinx-coroutines-core/common/src/Await.kt
```diff
@@ -11,11 +11,10 @@ import kotlin.coroutines.*
  * This function is **not** equivalent to `deferreds.map { it.await() }` which fails only when it sequentially
  * gets to wait for the failing deferred, while this `awaitAll` fails immediately as soon as any of the deferreds fail.
  *
- * This suspending function is cancellable.
- * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting,
- * this function immediately resumes with [CancellationException].
- * There is a **prompt cancellation guarantee**. If the job was cancelled while this function was
- * suspended, it will not resume successfully. See [suspendCancellableCoroutine] documentation for low-level details.
+ * This suspending function is cancellable: if the [Job] of the current coroutine is cancelled or completed while this
```
qwwdfsad, line 14: W.r.t. `This suspending function is cancellable`, this particular part:

>if the [Job] of the current coroutine is cancelled **or completed** while this suspending function is waiting, this function immediately resumes with [CancellationException].

Maybe I'm missing something, but isn't this scenario (Job is completed, but in its body the code is suspended and is waiting) impossible?

kotlinx-coroutines-core/common/src/Deferred.kt
```diff
@@ -33,17 +33,37 @@ import kotlinx.coroutines.selects.*
...
-     * returning the resulting value or throwing the corresponding exception if the deferred was cancelled.
+     * Awaits for completion of this value without blocking the thread and returns the resulting value or throws
+     * the exception if the deferred was cancelled.
      *
-     * This suspending function is cancellable.
-     * If the [Job] of the current coroutine is cancelled or completed while this suspending function is waiting, this function
-     * immediately resumes with [CancellationException].
-     * There is a **prompt cancellation guarantee**. If the job was cancelled while this function was
-     * suspended, it will not resume successfully. See [suspendCancellableCoroutine] documentation for low-level details.
+     * Unless cancelled, [await] will return the same result on each invocation:
```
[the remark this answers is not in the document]
dkhalanskyjb: No, I mean the `await` invocation. Is this incorrect? I'd assume that once `Deferred` has a state, it's final… though maybe during deferred's cancellation `await` throws `CancellationException`, but if a failure happens, then another exception is thrown? I'll check this.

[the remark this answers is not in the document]
qwwdfsad: It is correct, I might misread it.

Maybe worth phrasing it as `unless the caller/calling coroutine is cancelled`?

kotlinx-coroutines-core/common/src/Await.kt
dkhalanskyjb: Yeah, I simply kept it the way it was without thinking, but now I have no idea what the text was trying to convey. I replaced `is cancelled or completed while` with `is cancelled while` throughout the project, even outside the prompt cancellation explanations.

qwwdfsad (approved):
👍

PR #4052, opened 2024-02-23
Remove usage of `kotlin.random.Random` form `CoroutineScheduler.Worker` init

### LINKED ISSUE #4051: Noticeable upfront cost of Dispatchers.Default
**Describe the problem**

Creation of the `CoroutineScheduler` requires initialization of `kotlin.random.Random`, which, in the case of `kotlin-stdlib-jdk8` is used, requires dynamic class-loading of `android.os.BUNDLE` (see: https://youtrack.jetbrains.com/issue/KT-44089)

However, `Dispatchers.Default` is often used early in the application runtime when startup logic uses coroutines, and additional class-loading delays are unfortunate.
See [github: "main" /Dispatchers.Default/](https://github.com/search?q=%22main%28%22+%2FDispatchers.Default%2F&type=code) for examples.

<details>
  <summary>Profiler result</summary>
  
Screenshot 2024-02-23 at 23 26 13

</details>

**Provide a Reproducer**

```kotlin
// FILE: main.kt


fun main() {
    val dispatcher = Dispatchers.Default
    runBlocking {
        delay(1000)
    }

    subCall(dispatcher)
}

suspend fun mainImpl() {
    println("done")
}
```
```kotlin
// FILE: other.kt
import kotlinx.coroutines.CoroutineDispatcher

fun subCall(dispatcher: CoroutineDispatcher) {
    // This runBlocking is causing a long kotlin.Random.<clinit>
    runBlocking(dispatcher) {
        mainImpl()
    }
}
```

<details>
  <summary>build.gradle.kts</summary>

  ```kotlin
// FILE: build.gradle.kts

plugins {
    kotlin("jvm") version "1.9.21"
}

group = "org.example"
version = "1.0-SNAPSHOT"

repositories {
    mavenCentral()
}

dependencies {
    implementation(kotlin("stdlib-jdk8"))
    testImplementation("org.jetbrains.kotlin:kotlin-test")

    // dependency on Kotlinx-coroutines
    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.0")
}
kotlin {
    jvmToolchain(21)
}

```

</details>

### PR DESCRIPTION
**Pull request by semoro**
During the creation of `CoroutineScheduler.Worker`, we need to initialize its embedded random number generator. To avoid additional class-loading costs (#4051), `rngState` is now directly initialized from the current nanoTime (that is used as seed).

Also, a potential bug is fixed, where `nextInt` will always return zero due to incorrect initialization of rngState with zero.
This may happen during the creation of a worker with thread id = `1595972770` (JDK >=8), or unpredictable if fallback thread-local random is used (android SDK <34 or JDK <7), approximate probability is 2.4E-10

Fixes #4051

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -657,20 +657,17 @@ internal class CoroutineScheduler(
         var nextParkedWorker: Any? = NOT_IN_STACK
 
         /*
-         * The delay until at least one task in other worker queues will  become stealable.
+         * The delay until at least one task in other worker queues will become stealable.
          */
         private var minDelayUntilStealableTaskNs = 0L
 
         /**
          * The state of embedded Marsaglia xorshift random number generator, used for work-stealing purposes.
          * It is initialized with a seed.
-         *
-         * @see nextInt
          */
         private var rngState: Int = run {
-            // Initialize with a seed from the least significant integer portion of the nanoTime to ensure initial randomness.
+            // This could've been Random.nextInt(), but we are shaving an extra initialization cost, see #4051
             val seed = System.nanoTime().toInt()
-
             // rngState shouldn't be zero, as required for the xorshift algorithm
             if (seed != 0) return@run seed
             42
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -8,7 +8,6 @@ import java.util.concurrent.*
 import java.util.concurrent.locks.*
 import kotlin.jvm.internal.Ref.ObjectRef
 import kotlin.math.*
-import kotlin.random.*
 
 /**
  * Coroutine scheduler (pool of shared threads) which primary target is to distribute dispatched coroutines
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by semoro** [comment]
Unfortunately, I couldn't provide any practical tests for the aforementioned bug since it requires random events to happen. 

Also, it is unclear to me if I can add any performance tests that will highlight the `Random` usage problem (#4051) either :(

File: kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt [hunk 1]
```diff
@@ -658,11 +657,21 @@ internal class CoroutineScheduler(
...
+        /**
+         * The state of embedded Marsaglia xorshift random number generator, used for work-stealing purposes.
+         * It is initialized with a seed.
+         */
+        private var rngState: Int = run {
+            // This could've been Random.nextInt(), but we are shaving an extra initialization cost, see #4051
+            val seed = System.nanoTime().toInt()
+            // rngState shouldn't be zero, as required for the xorshift algorithm
+            if (seed != 0) return@run seed
+            42
```
dkhalanskyjb, line 673: It's more conventional to return `4`: https://xkcd.com/221/ `42` is the answer to a question, but `rngState` is not giving us an answer of any kind, so this constant is not really applicable.

dkhalanskyjb (approved):
Thanks for fixing the bug! I don't understand the performance benefits of this PR: I'd assume that the initialization costs are amortized, so this doesn't actually win any time if `Random` is used anywhere else, and the cost of initializing `Random` seems to be dwarfed completely by the time it takes to initialize the JVM itself.

Using `System.nanoTime` here shouldn't hurt anything, though, so I won't debate the change.

PR #4054, opened 2024-02-29
Potential workaround for #3820

### LINKED ISSUE #3820: NullPointerException when setting StateFlow value
**Describe the bug**

After upgrading from version 1.6.4 to 1.7.1 (we have since bumped to 1.7.2) we started seeing `NullPointerException` crashes when updating a `StateFlow` value. I am not able to reproduce this locally, we are only seeing this through Firebase reports.

This is happening on Android with `org.jetbrains.kotlinx:kotlinx-coroutines-android`. It's happening on many different devices and Android versions, so it isn't specific to one manufacturer/version.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Class.isInterface()' on a null object reference
       at java.lang.Class.isAssignableFrom(Class.java:589)
       at java.lang.Class.isInstance(Class.java:542)
       at java.util.concurrent.atomic.AtomicReferenceFieldUpdater$AtomicReferenceFieldUpdaterImpl.accessCheck(AtomicReferenceFieldUpdater.java:389)
       at java.util.concurrent.atomic.AtomicReferenceFieldUpdater$AtomicReferenceFieldUpdaterImpl.get(AtomicReferenceFieldUpdater.java:447)
       at kotlinx.coroutines.flow.StateFlowSlot.makePending(StateFlowSlot.java:59)
       at kotlinx.coroutines.flow.StateFlowImpl.updateState(StateFlow.kt:349)
       at kotlinx.coroutines.flow.StateFlowImpl.setValue(StateFlow.kt:316)
```

**Provide a Reproducer**

We have a lot of flows, and only one specific one is crashing. The only unique thing about the flow that is causing crashes is that it holds an `Enum`, but I don't know if that's causing it.

The `StateFlow` is nullable, but the value set on it after initialization is never null. The code below isn't our production code, but is how our code looks like

```
enum class SomeEnum {
    Apple,
    Orange
}

class Class {
    val flow: StateFlow<SomeEnum?> get() = _mutableFlow
    private val _mutableFlow: MutableStateFlow<SomeEnum?> = MutableStateFlow(null)
    
    fun updateValue(value: SomeEnum) {
        _mutableFlow.value = value
    }
}
```

qwwdfsad:
Replace the specific place where ARFU gets misexecuted by specific Android toolchain

Fixes #3820

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/Concurrent.common.kt
+++ b/kotlinx-coroutines-core/common/src/internal/Concurrent.common.kt
@@ -21,7 +21,7 @@ internal expect fun <E> identitySet(expectedSize: Int): MutableSet<E>
 internal expect annotation class BenignDataRace()
 
 // Used **only** as a workaround for #3820 in StateFlow. Do not use anywhere else
-internal expect class AtomicReference<T>(value: T) {
+internal expect class WorkaroundAtomicReference<T>(value: T) {
     public fun get(): T
     public fun set(value: T)
     public fun getAndSet(value: T): T
@@ -29,11 +29,11 @@ internal expect class AtomicReference<T>(value: T) {
 }
 
 @Suppress("UNUSED_PARAMETER", "EXTENSION_SHADOWED_BY_MEMBER")
-internal var <T> AtomicReference<T>.value: T
+internal var <T> WorkaroundAtomicReference<T>.value: T
     get() = this.get()
     set(value) = this.set(value)
 
-internal inline fun <T> AtomicReference<T>.loop(action: AtomicReference<T>.(value: T) -> Unit) {
+internal inline fun <T> WorkaroundAtomicReference<T>.loop(action: WorkaroundAtomicReference<T>.(value: T) -> Unit) {
     while (true) {
         action(value)
     }
--- a/kotlinx-coroutines-core/common/src/internal/Concurrent.common.kt
+++ b/kotlinx-coroutines-core/common/src/internal/Concurrent.common.kt
@@ -19,3 +19,22 @@ internal expect fun <E> identitySet(expectedSize: Int): MutableSet<E>
 @OptionalExpectation
 @Target(AnnotationTarget.FIELD)
 internal expect annotation class BenignDataRace()
+
+// Used **only** as a workaround for #3820 in StateFlow. Do not use anywhere else
+internal expect class AtomicReference<T>(value: T) {
+    public fun get(): T
+    public fun set(value: T)
+    public fun getAndSet(value: T): T
+    public fun compareAndSet(expected: T, value: T): Boolean
+}
+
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/common/src/internal/Concurrent.common.kt
```diff
@@ -19,3 +19,22 @@ internal expect fun <E> identitySet(expectedSize: Int): MutableSet<E>
 @OptionalExpectation
 @Target(AnnotationTarget.FIELD)
 internal expect annotation class BenignDataRace()
+
+// Used **only** as a workaround for #3820 in StateFlow. Do not use anywhere else
```
dkhalanskyjb, line 23: `@OptIn(StopDoNotUse::class)`? `class BadAtomicReferenceForWorkarounds`? An arbitrary version of me down the line can't be trusted not to use this `AtomicReference` accidentally, and I don't read non-KDoc comments above every entity I use.

dkhalanskyjb (approved):
Let's hope this works!

rossbacher:
Is there any eta for a merge/release or a snapshot? We can actually see this in our beta release and could verify a fix that way.

qwwdfsad:
@rossbacher we do plan to release it around the next week or two. 
About a potential snapshot -- @dkhalanskyjb WDYT?

dkhalanskyjb:
Sure, we can publish 1.8.1-Beta in the next few days. Maybe even today.

Also, maybe we should document the procedure for publishing our libraries locally. There should be no reason to wait for a new release just to try out new changes.

rossbacher:
I managed to figure out how to build and publish a fork (1.8.0 + this PR) to our internal artifactory on Friday. We did have this out on beta since then and so far no crashes (where we had about 2 crashes on the same # sessions on the previous beta that just had 1.8.0) so it looks like as if it is fixed. 🤞

rossbacher:
> Also, maybe we should document the procedure for publishing our libraries locally. There should be no reason to wait for a new release just to try out new changes.

This would be great, my initial approach was to build it and manually publish the artifacts to our artifactory (with manually patched .pom files from the previous release, but I realized that at least two artifacts we need (`core` and `bom`) did not seem to get built when you just run the `build` target. 

In the end I just ended up (on my fork) changing https://github.com/Kotlin/kotlinx.coroutines/blob/master/buildSrc/src/main/kotlin/Publishing.kt#L60 to our repo URL and setting my credentials for `libs.sonatype.user` and `libs.sonatype.password` and running the `publish` target, which worked like a charm.

aasitnikov:
We've rolled out an update that includes a fix from this PR a week ago. Yesterday, we had 1,400 crashes from 4.5 million users on the previous version without a fix, and zero crashes from 80,000 users on the current version with the fix. So far, it looks like this workaround is working.

PR #4057, opened 2024-03-05
Update Dokka to 1.9.20

whyoleg:
No issues comparing to Dokka 1.9.10. knit tasks works


dkhalanskyjb:
The changes seem beneficial; only one place caught my eye. In `kotlinx-coroutines-debug/build/dokka/htmlPartial/kotlinx.coroutines.debug/-debug-probes/print-job.html`, in the old version, there's an `<a href` for the `Job` reference, but with the new version, there's some `<dokka-template-command` instead. Is this expected?

whyoleg:
`htmlPartial` tasks outputs are not relevant and could change from release to release, as they are like intermediate `partial` result of Dokka for single module in multi-module projects.
So yes, something could change there. Most likely in this case, resolving of the links was changed (and most likely fixed) as per https://github.com/Kotlin/dokka/issues/2272 - as in `kotlinx.coroutines` there is exists the same package in both `core` and `debug` artefact.

If you still do want to check the difference, it will be more correct to check `multi-module` result which is produced in `build/dokka/htmlMultiModule`

dkhalanskyjb:
Oh, I see, thanks! Didn't know about these fine details. Looking briefly at `dokka/htmlMultiModule`, I do indeed see nothing suspicious.

Issue #1814, opened 2020-02-18
Provide a `runCatching` that does handle a `CancellationException` but re-throws it instead.


Anton-Spaans-Intrepid:
### Problem Description
The Kotlin StdLib contains a function named `runCatching`. It tries the provided lambda and catches any `Throwable` that is not caught or thrown by the lambda.

However, using `runCatching` in Coroutines can have a somewhat hidden (and unexpected) side-effect of it catching `CancellationException`s and not re-throwing them. This could interfere with the Coroutines' Structured Concurrency, with its cancellation handling.

### Possible Solution
E.g. create a new function in kotlinx.coroutines that looks something like this 
(not sure about the name 😄):
```
public inline suspend fun <R> runSuspendCatching(block: () -> R): Result<R> {
    return try {
        Result.success(block())
    } catch(c: CancellationException) {
        throw c
    } catch (e: Throwable) {
        Result.failure(e)
    }
}
```

The `runSuspendCatching` will catch any exception and wrap it in a `Result`, except for `CancellationException`s that will be re-thrown instead. This will prevent unwanted interference with the cancellation handling of Coroutines.

### Notes
- The `suspend` keyword is added to prevent this new function from being called in regular non-suspending/blocking code.
- When code calls the StdLib's `runCatching` in a Coroutine/suspendable-context, a lint warning should warn the developer that handling cancellations may be compromised and `runSuspendCatching` should be used instead.

qwwdfsad:
Thanks for the proposal!

Could you please elaborate why is it unexpected behaviour? E.g. why one may expect to use `runCatching` that actually rethrows **some** of the exceptions? What are the use-cases and why `runSuspendCatching` is preferred over something more conventional?

Anton-Spaans-Intrepid:
@qwwdfsad 

If the developer uses the StdLib `runCatching { ... }` and the `runCatching` wraps a suspendable piece of code (i.e. it could throw a `CancellationException`), the `runCatching` should re-throw any `CancellationException`. If not the suspendable caller of `runCatching` may not get cancelled and the Structured Concurrency of Coroutines breaks down.

fvasco:
@Anton-Spaans-Intrepid
Do you think that the current `runCathing` function should rethrow any `InterruptedException`?

Anton-Spaans-Intrepid:
@fvasco Good question; maybe :-) 
But this is about CancellationExceptions when using runCatching in a Coroutine.

[...]

PR #4059, opened 2024-03-08
Add cooperativeCatch as coroutine-safe alternative to runCatching

mattshoe:
## What is this?
This PR adds 2 new methods:
 1. cooperativeCatch
 2. cooperativeMap

Both of these function the same way that `runCatching` and `mapCatching`.
However, these implementations obey cooperative cancellation by immediately propagating `CancellationException` instead of encapsulating it, as `runCatching` and `mapCatching` do.

## Why do we want it?
Firstly, it is not immediately obvious to developers that `runCatching` may actually swallow `CancellationException` and cause memory leaks when a suspend function is invoked inside `runCatching`. It is perfectly reasonable to assume that devs would think a language-provided feature would play nicely with another language-provided feature

Secondly, many devs strongly prefer the functional style of `runCatching` over a typical `try/catch` block. And they simply cannot use their preferred functional style of exception handling with the current state of coroutines. This PR gives devs the power to preserve their preferred programming style.

Discussion on this topic can be found in this issue: https://github.com/Kotlin/kotlinx.coroutines/issues/1814

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -147,6 +147,11 @@ public final class kotlinx/coroutines/CompletionHandlerException : java/lang/Run
 	public fun <init> (Ljava/lang/String;Ljava/lang/Throwable;)V
 }
 
+public final class kotlinx/coroutines/CooperativeCatchKt {
+	public static final fun cooperativeCatch (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
+	public static final fun cooperativeMap (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
+}
+
 public abstract interface class kotlinx/coroutines/CopyableThreadContextElement : kotlinx/coroutines/ThreadContextElement {
 	public abstract fun copyForChild ()Lkotlinx/coroutines/CopyableThreadContextElement;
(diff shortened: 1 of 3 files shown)


dkhalanskyjb:
> Firstly, it is not immediately obvious to developers that runCatching may actually swallow CancellationException

And this PR doesn't help with that. You still have to know about the new functions to invoke them.

> Secondly, many devs strongly prefer the functional style of runCatching over a typical try/catch block.

`runCatching` is only occasionally useful, regardless of the preferred style, because in Java's (and, by proxy, Kotlin's) model of error handling, unexpected errors manifest themselves not as runtime panics (Rust- or Haskell-style) but as exceptions. `runCatching` catches an out-of-memory exception just as gladly as it catches an exception that can be meaningfully handled. In the `try`-`catch` style, this means that catching a `Throwable` is a huge anti-pattern, only to be done during foundational work. Using `runCatching` because of "style" falls in the same category. Please see https://elizarov.medium.com/kotlin-and-exceptions-8062f589d07 for an elaborate description of error handling in Kotlin.

Unfortunately, this problem is especially acute with `CancellationException`: in most cases, you must rethrow it, but in some cases, you must process it as you would any other. See https://github.com/Kotlin/kotlinx.coroutines/issues/1814#issuecomment-1943224856 for an explanation. Without clearing up the underlying conceptual issues first, `runCatching`-like functions with special behavior around `CancellationException` will only add to the confusion. And if someone wants to have these wrappers even despite the fact that they are error-prone in some cases, then they can introduce them in their own code, as they are straightforward to write.

In short, these wrappers only hide some of the complexity, but don't help to deal with it.

PR #4060, opened 2024-03-11
Version 1.8.1 beta

dkhalanskyjb:


CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.8.1-Beta
+
+* Remove the `@ExperimentalTime` annotation from usages of `TimeSource` (#4046). Thanks, @hfhbd!
+* Attempt a workaround for an Android bug that caused an occasional `NullPointerException` when setting the `StateFlow` value on old Android devices (#3820).
+* Small tweaks.
```
qwwdfsad, line 7: #4051 is missing.

dkhalanskyjb: I couldn't think of a way to phrase it as a notable change to the end user. Do you have suggestions?

qwwdfsad: "Slightly reduced initialization cost of Dispatchers.Default and Dispatchers.IO"?

dkhalanskyjb: * Doesn't seem honest: the costs are still there, but are moved elsewhere. By this logic, we could wrap many things in `lazy` and say that the initialization costs are reduced.
* Even if we end decide to take this formulation, does this not fall into the "small tweaks" category?

"In some scenarios, `Dispatchers.Default` and `Dispatchers.IO` may now be available more quickly"?

dkhalanskyjb: Here's an idea: "No longer use `kotlin.random.Random` as part of `Dispatchers.Default` and `Dispatchers.IO` initialization." Those interested in when exactly `kotlin.random.Random` gets initialized should find it on-point, and those not interested won't be misled.

qwwdfsad (approved):

PR #4065, opened 2024-03-14
Make `BufferedChannelIterator#hasNext` idempotent

gitpaxultek:
The rationale in favor of the change includes:
* in the `1.6.x` series `Channel().iterator().hasNext()` was idempotent
* despite the lack of requirement for the `Iterator#hasNext` method implementations to be idempotent, it is generally safer to make them such. The same applies to `ChannelIterator#hasNext`

Here is an example where it matters. Consider the following code snippet:

```
val chunk = mutableListOf<Int>()
while(iter.hasNext()) {
  val v = iter.next()
  chunk.add(v)
  if (chunk.size >= MAX_BATCH_SIZE || !iter.hasNext()) {
    yield(chunk.toList())
    chunk.clear()
  }
}
```

We address two concerns with one block of code - we run the if-block on reaching the chunk size limit or on the last chunk.

If `#hasNext` isn't idempotent, I see a couple of options. The first solution is, the code would essentially have to execute the content of the if-block in two places - in the loop and after the loop:

```
val chunk = mutableListOf<Int>()
while(iter.hasNext()) {
  val v = iter.next()
  chunk.add(v)
  if (chunk.size >= MAX_BATCH_SIZE) {
    yield(chunk.toList())
    chunk.clear()
  }
}

if (chunk.isNotEmpty) {
    yield(chunk.toList())
}
```

Alternatively, we can do memoization of `#hasNext` ourselves e.g.:

```
val chunk = mutableListOf<Int>()
var hasNextMemo = iter.hasNext()
while(hasNextMemo) {
  val v = iter.next()
  chunk.add(v)
  hasNextMemo = iter.hasNext()
  if (chunk.size >= MAX_BATCH_SIZE || !hasNextMemo) {
    yield(chunk.toList())
    chunk.clear()
  }
}
```

Either way, the lack of idempotency of `#hasNext` spills the concern of knowing the state of the iterator over into the caller's code.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
@@ -1580,8 +1580,8 @@ internal open class BufferedChannel<E>(
      * [CancellableContinuation] and [SelectInstance].
      *
      * Roughly, [hasNext] is a [receive] sibling, while [next] simply
-     * returns the already retrieved element. From the implementation
-     * side, [receiveResult] stores the element retrieved by [hasNext]
+     * returns the already retrieved element and [hasNext] being idempotent.
+     * From the implementation side, [receiveResult] stores the element retrieved by [hasNext]
      * (or a special [CHANNEL_CLOSED] token if the channel is closed).
      *
      * The [invoke] function is a [CancelHandler] implementation,
@@ -1614,8 +1614,10 @@ internal open class BufferedChannel<E>(
         private var continuation: CancellableContinuationImpl<Boolean>? = null
 
         // `hasNext()` is just a special receive operation.
-        override suspend fun hasNext(): Boolean =
-            receiveImpl( // <-- this is an inline function
+        override suspend fun hasNext(): Boolean {
+            return if (this.receiveResult != NO_RECEIVE_RESULT && this.receiveResult != CHANNEL_CLOSED) {
+                true
+            } else receiveImpl( // <-- this is an inline function
                 // Do not create a continuation until it is required;
                 // it is created later via [onNoWaiterSuspend], if needed.
                 waiter = null,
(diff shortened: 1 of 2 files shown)


dkhalanskyjb:
> The first solution is, the code would essentially have to execute the content of the if-block in two places - in the loop and after the loop

Yep, that's the best one. This way, you aren't wasting time on checking "is this the end already?" on every iteration, which even looks awkward. Also, this version can be adapted to the more idiomatic `for`-loop more easily:
```kotlin
import kotlinx.coroutines.channels.Channel

const val MAX_BATCH_SIZE = 100

suspend fun main() {
    val channel = Channel<Int>()
    val chunk = mutableListOf<Int>()
    flow {
        for (v in channel) {
            chunk.add(v)
            if (chunk.size >= MAX_BATCH_SIZE) {
                emit(chunk.toList())
                chunk.clear()
            }
        }
        if (chunk.isNotEmpty()) {
            emit(chunk.toList())
        }
    }
}
```

kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
```diff
@@ -1614,8 +1614,10 @@ internal open class BufferedChannel<E>(
         private var continuation: CancellableContinuationImpl<Boolean>? = null
 
         // `hasNext()` is just a special receive operation.
-        override suspend fun hasNext(): Boolean =
-            receiveImpl( // <-- this is an inline function
+        override suspend fun hasNext(): Boolean {
+            return if (this.receiveResult != NO_RECEIVE_RESULT && this.receiveResult != CHANNEL_CLOSED) {
```
dkhalanskyjb, line 1618:
```suggestion
            return if (this.receiveResult !== NO_RECEIVE_RESULT && this.receiveResult !== CHANNEL_CLOSED) {
```

kotlinx-coroutines-core/common/test/channels/BufferedChannelTest.kt
```diff
@@ -5,6 +5,40 @@ import kotlinx.coroutines.*
 import kotlin.test.*
 
 class BufferedChannelTest : TestBase() {
+    @Test
+    fun testIterHasNextIsIdempotant() = runTest {
```
dkhalanskyjb, line 9:
```suggestion
    fun testIteratorHasNextIsIdempotent() = runTest {
```

dkhalanskyjb:
Thanks!

PR #4072, opened 2024-03-18
Fix `Flow.timeout` swallowing the channel closure exception

### LINKED ISSUE #4071: Callbackflow close exception misbehavior with timeout
**Describe the bug**

When `callbackflow` is chained with `timeout` directly, and closed with an exception, the exception won't be sent to downstream. 

 Coroutines version 1.8.0. Kotlin version 1.9.21.

**Provide a Reproducer**

```
val flow = callbackFlow<Int> {
    close(IllegalStateException("Test Exception"))
}.timeout(5.seconds)
try {
    flow.collect()
} catch (e: Exception) {
    println("exception got: $e")
}
```
I expect this to print:
`exception got: java.lang.IllegalStateException: Test Exception`

But no exception is caught.

However, if I add any operator between callbackflow and timeout, it works as expected, 
```
val flow = callbackFlow<Int> {
    close(IllegalStateException("Test Exception"))
}.catch{throw it}.timeout(5.seconds)
try {
    flow.collect()
} catch (e: Exception) {
    println("exception got: $e")
}
```

dkhalanskyjb:
Fixes #4071

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -394,6 +394,7 @@ private fun <T> Flow<T>.timeoutInternal(
             value.onSuccess {
                 downStream.emit(it)
             }.onClosed {
+                it?.let { throw it }
                 return@onReceiveCatching false
             }
             return@onReceiveCatching true
--- a/kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
@@ -237,6 +237,17 @@ class TimeoutTest : TestBase() {
         testImmediateTimeout(-1.seconds)
     }
 
+    @Test
+    fun testClosing() = runTest {
+        assertFailsWith<TestException> {
+            channelFlow<Int> { close(TestException()) }
+                .timeout(Duration.INFINITE)
+                .collect {
+                    expectUnreached()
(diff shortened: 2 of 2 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -394,6 +394,7 @@ private fun <T> Flow<T>.timeoutInternal(
             value.onSuccess {
                 downStream.emit(it)
             }.onClosed {
+                it?.let { throw it }
```
qwwdfsad, line 397: Is it correct that the reason this issue manifested itself is our fusing optimization?

qwwdfsad: Oh, it boils down to the following:
```
val job = Job()
val produce = produce<Int>(job) { close(TestException()) }
(produce as Job).join()
job.isCancelled // False
```

Which makes sense for `produce` specifically (when it's used as a standalone coroutine builder), but not much when used as part of the flow machinery

qwwdfsad (approved):

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
qwwdfsad: Also, kind of confusing semantics-wise: `close` with a cause **fails** the channel, and when a coroutine fails, the channel fails as well. Yet the channel failure on its now does not represent failure.

PR #4081, opened 2024-04-02
fix: remove `sampleStart` and `sampleEnd` comments from example in coroutine-context-and-dispatchers.md

PetrakovichVictoria:
https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#jumping-between-threads

UPD: removed sampleStart and sampleEnd comments.

Fix this sample not being runnable:
Screenshot 2024-04-02 at 13 00 09

### DIFF
--- a/docs/topics/coroutine-context-and-dispatchers.md
+++ b/docs/topics/coroutine-context-and-dispatchers.md
@@ -227,7 +227,6 @@ import kotlinx.coroutines.*
 fun log(msg: String) = println("[${Thread.currentThread().name}] $msg")
 
 fun main() {
-//sampleStart
     newSingleThreadContext("Ctx1").use { ctx1 ->
         newSingleThreadContext("Ctx2").use { ctx2 ->
             runBlocking(ctx1) {
@@ -239,10 +238,8 @@ fun main() {
             }
         }
     }
-//sampleEnd
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
 
 > You can get the full code [here](../../kotlinx-coroutines-core/jvm/test/guide/example-context-04.kt).
 >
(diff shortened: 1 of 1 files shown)


danil-pavlov (approved):

dkhalanskyjb:
This example isn't running correctly, but fails with
```
Exception in thread "main" java.security.AccessControlException: Access control exception due to security reasons in web playground: 
 access denied ("java.lang.RuntimePermission" "modifyThread")
 at java.lang.SecurityManager.checkPermission (:-1) 
```

PetrakovichVictoria:
I don't know if the problem is in the coroutines here or in some restrictions in the playground. Maybe it's just worth deleting //sampleStart and sampleEnd comments and making this example non-runnable.

dkhalanskyjb:
This example is being run as part of our tests (https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/test/guide/test/DispatcherGuideTest.kt#L37-L44); so, these are probably just the playground restrictions. Also, making the example non-runnable would remove the test automatically, and I don't think we want that.

qwwdfsad:
It's https://youtrack.jetbrains.com/issue/KTL-108 unfortunately

PetrakovichVictoria:
> This example is being run as part of our tests

I also launched it before I created this PR: locally, this example passes w/o problems. 

> Also, making the example non-runnable would remove the test automatically, and I don't think we want that.

1. Having sampleStart and sampleEnd shows an initial will for this example to be runnable. 
2. Almost all other examples from this page work (except [here](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#coroutine-scope) because these ones show some thought in general). So, having this one example non-runnable seems strange twice. 
3. For me, it's not that much important will it be runnable or not, but I suppose, if we want to leave it non-runnable, then `//sampleStart` and `//sampleEnd` should be removed (and the runnability marker too). I can do this if you wish.

PetrakovichVictoria:
I don't see any changes after running knit locally. Removed runnability and sample start/end comments.

dkhalanskyjb:
Thanks!

PR #4082, opened 2024-04-02
chore: reference to The Hitchhiker's Guide to the Galaxy

PetrakovichVictoria:
At the end of the book, they multiply 6 * 7 :)


PetrakovichVictoria:
It's an ideal place for such a reference, as we play with this 42 along the way) I wasn't able to sustain😅

danil-pavlov (approved):

qwwdfsad (approved):
Thanks for all the fish!

Issue #4088, opened 2024-04-05
Remove @ExperimentalCoroutinesApi from CancellableContinuation.resume(...)


swankjesse:
I’d like to use this overload of `CancellableContinuation.resume()` in OkHttp’s coroutines extension.

```
    @ExperimentalCoroutinesApi // since 1.2.0
    public fun resume(value: T, onCancellation: ((cause: Throwable) -> Unit)?)
```

Unfortunately, this functino is experimental. There’s no stable alternative that combines continuations with safe resource management. I’m going to annotate our own API with `@ExperimentalCoroutinesApi` to pass the responsibility to our callers.

I’d like for you to consider removing the experimental annotation from this function. It hasn’t changed since 2020. It solves an important very nicely!

qwwdfsad:
We've been sitting on that for a while for quite a specific reason: its signature should've been `(value: T, cause: Throwable) -> Unit`, otherwise, when serving its main purpose, it is a capturing lambda that always allocates, even though there are no reasons real reasons to.
Changing that, there is a variance conflict -- continuations are contravariant (`in T`) and functional types are covariant (`out T`), which forces us to fallback to `@UnsafeVariance` and ensure that it's sound.


There was an extended attempt (https://github.com/Kotlin/kotlinx.coroutines/pull/3093) to change that and also receive a `CoroutineContext` as the third parameter, though this part remains questionable, and we never figured it out.

Taking into account it's experimental for so long if we are going to change the signature and deprecate it, this version is going to stay basically forever

swankjesse:
Thanks for the explanation. I hadn’t expected covariance!

PR #4090, opened 2024-04-05
Provide `resume` and `tryResume` that pass the value to the callback

dkhalanskyjb:
Fixes some concerns in #4088


dkhalanskyjb:
This is clearly far from optimal, and also, there are no tests, nor documentation. Requesting a review anyway to discuss if the approach is viable.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -198,6 +198,8 @@ public interface CancellableContinuation<in T> : Continuation<T> {
      */
     @ExperimentalCoroutinesApi // since 1.2.0
     public fun resume(value: T, onCancellation: ((cause: Throwable) -> Unit)?)
+
+    public fun <R: T> resume(value: R, onCancellation: ((cause: Throwable, value: R) -> Unit)?)
```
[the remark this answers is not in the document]
dkhalanskyjb: It's not a trick: we're notifying the type system that the type of `onCancellation`'s value is decided not by the view of the continuation, but primarily by the type of `value` passed to the function. This type bound promises that `onCancellation` will receive exactly the `value` that this same function receives, and not just any value that `CancellableContinuation` could be resumed with *in theory*.

Without this information, `onCancellation` would have to be ready to accept an arbitrary *supertype* of `T` (effectively, *any value*): because of the `in` variance of `T` in `CancellableContinuation`, `resume` could be called on a more specialized view of `CancellableContinuation`, that is, if we have `CancellableContinuation<Any>`, the call could be `CancellableContinuation<Int>.resume`, and also, in parallel, the continuation could be actually resumed with a `Double`. We know this can't happen, but with the bound, the type system knows that, too.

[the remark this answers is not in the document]
qwwdfsad: Thanks for the explanation, that's what I've meant :) "Trick" is a substitute for a "a nice and smart solution to our specific issue"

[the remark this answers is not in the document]
dkhalanskyjb: One more: whether we want to break source compatibility in most cases or almost none of them. The latter can be achieved with `onCancellation: ((R).(cause: Throwable) -> Unit)?` and nicely supports the pattern `cont.resume(response) { close() }`, but it can also silently change the semantics of existing code.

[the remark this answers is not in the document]
dkhalanskyjb: > It's not the selects performance I worry about, but the general applicability

I looked into whether we can gain some performance anyway, and the best I could manage included a very non-idiomatic way of structuring code to avoid one extra allocation per `resume`. Here: https://github.com/Kotlin/kotlinx.coroutines/commit/0e3a633243ce37f625eeb0bc3651951fc82a9b9d

kotlinx-coroutines-core/common/src/sync/Mutex.kt
```diff
@@ -140,8 +143,9 @@ internal open class MutexImpl(locked: Boolean) : SemaphoreImpl(1, if (locked) 1
             { unlock(owner) }
         }
 
-    override val isLocked: Boolean get() =
-        availablePermits == 0
+    override val isLocked: Boolean
+        get() =
+            availablePermits == 0
```
dkhalanskyjb, line 148: Autoformatting did something strange.

kotlinx-coroutines-core/common/src/selects/Select.kt
```diff
@@ -844,7 +854,8 @@ internal open class SelectImplementation<R>(
 }
 
 private fun CancellableContinuation<Unit>.tryResume(onCancellation: ((cause: Throwable) -> Unit)?): Boolean {
```
whyoleg, line 856: may be it makes sense to change `OnCancellationConstructor` to return lambda which is compatible with new `tryResume` to not allocate additional lambda here?

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -168,36 +173,59 @@ public interface CancellableContinuation<in T> : Continuation<T> {
     @ExperimentalCoroutinesApi
     public fun CoroutineDispatcher.resumeUndispatchedWithException(exception: Throwable)
 
+    /** @suppress */
+    @Deprecated(
+        "Use the overload that also accepts the `value` and the coroutine context in lambda",
+        level = DeprecationLevel.WARNING
```
qwwdfsad, line 179:
```suggestion
        level = DeprecationLevel.WARNING,
        replaceWith = ReplaceWith("resume(value) { cause, _, _ -> onCancellation(cause) }")
```

qwwdfsad: I checked it on few examples, works as expected

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -168,36 +173,59 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
+     *
+     * The [value] can be rejected in two cases (in both of which [onCancellation] will be called):
+     * - Cancellation happened before the handler was resumed;
+     * - The continuation was resumed successfully (before cancellation), but the coroutine's job was cancelled before
+     *   it had a chance to run in its dispatcher, and so the suspended function threw an exception instead of returning
+     *   this value.
      *
      * The installed [onCancellation] handler should not throw any exceptions.
-     * If it does, they will get caught, wrapped into a [CompletionHandlerException] and
+     * If it does, they will get caught, wrapped into a [CompletionHandlerException], and
```
qwwdfsad, line 194: This is effectively an internal class with no documentation. Probably backticked version will do

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -168,36 +173,59 @@ public interface CancellableContinuation<in T> : Continuation<T> {
...
      *
-     * This function shall be used when resuming with a resource that must be closed by
-     * code that called the corresponding suspending function, for example:
+     * With this version of [resume], it's possible to pass resources that can not simply be left for the garbage
+     * collector (like file handles, sockets, etc.) and need to be closed explicitly:
      *
      * ```
-     * continuation.resume(resource) {
-     *     resource.close()
+     * continuation.resume(resourceToResumeWith) { cause, resourceToClose, context ->
```
qwwdfsad, line 202: The proposed code will be immediately highlighted.
```suggestion
     * continuation.resume(resourceToResumeWith) { _, resourceToClose, _ ->
```

qwwdfsad (approved):
Nice one!

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
dkhalanskyjb: Yes, it's creating a link to 404: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellable-continuation/invoke-on-cancellation.html The reason is https://github.com/Kotlin/dokka/issues/3601, but there is no reason not to work around it here.

Issue #3887, opened 2023-09-15
Replace usages of DCLL to CLL in JobSupport and friends


qwwdfsad:
This is a subtask required for #3886

Work outline:

* [ ] Get rid of DCSS (double compare single swap) primitive in `JobSupport` state transitions
* [ ] Prototype the solution where DCSS is replaced with a straightforward usage of CLL
* [ ] Optimize CLL footprint to be on par with DCSS in `JobSupport`-specific scenarios:
    * [ ]  Ensure that all `CompletionHandler` still can subclass an internal CLL piece in order to save a node allocation if appropriate
    * [ ] Ensure that an additional `DisposableHandle` is not allocated on a hot-path of adding a child to a job hierarchy

dkhalanskyjb:
Do the words DCLL and CLL mean "Doubly- [Concurrent] -Linked List" and "Concurrent Linked List"?

PR #4095, opened 2024-04-11
Rewrite the implementation of the linked list for `JobSupport`

### LINKED ISSUE #3886: Remove concurrent doubly linked list from kotlinx.coroutines codebase
We have `LockFreeLinkedListNode` and co. based on the  ["Lock-Free and Practical Doubly Linked List-Based Deques Using Single-Word Compare-and-Swap"](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.140.4693&rep=rep1&type=pdf) paper.

The implementation has a long-standing tail of issues:

1. The paper itself is known for having non-linearizability issues that trigger various failures when the data structure is stressed enough
2. DCLL is too generic: it allows all operations a trivial double-linked list (bi-directional iterations, mid-section removals etc.), which makes the reasoning and the maintenance of the concurrent invariants a tough task. Attempts to bisect a compact subset of only required invariants all failed.
3. The implementation is slower than it could be: any mutating operation implies at least 4 CASes; any added element corresponds to a separate object with multiple atomic fields (prev, next, removal marker)
4. Due to 2 & 3, the correct implementation is bloated, which contributes ~10% of _optimized_ DEX size of `kotlinx.coroutines`.

The proposed solution is straightforward -- get rid of DCLL and replace it with recently added FADD-based `ConcurrentLinkedList` that semaphore, mutex and channels leverage

dkhalanskyjb:
Fixes #3886

One of the steps for #3887

A draft because
* it's undertested
* and underdocumented,
* it approximately doubles the memory consumption of `JobSupport`.


dkhalanskyjb:
@qwwdfsad, could you help with optimizing the memory consumption? Not requesting a proper review yet because I know already the code is messy, I just expect it to get even messier during memory consumption optimization.

dkhalanskyjb:
For some reason, tests hang after a rebase, exclusively on Wasm/JS.

dkhalanskyjb:
After some discussions with `@igoriakovlev`, we pinpointed this to the following behavior that occurs on Wasm/JS with Kotlin 2.0.0 but not with other platforms or Wasm/JS with Kotlin 1.9.24: https://pl.kotl.in/OtO03tnT-
The issue occurs on https://github.com/Kotlin/kotlinx.coroutines/blob/70c992653d1a97b594ebf6995d2e46f0620cdf65/kotlinx-coroutines-core/common/src/internal/LockFreeLinkedList.common.kt#L157 Here, CAS never succeeds on Wasm/JS, since identity comparisons on value classes are prohibited and may return undefined results. The straightforward "fix" is to make `BrokenForSomeElements` a non-`value` class, but we're currently negotiating about having value classes work with CAS consistently.

This shouldn't block further work on this PR, we can work around this at the last moment if needed.

dkhalanskyjb:
It's not just value classes; apparently, all types that can be boxed aren't guaranteed to do something sensible with `===`, including primitive types.

PR #4097, opened 2024-04-11
Remove w3c usages from wasmJs implementation

igoriakovlev:
In a matter of removing w3c declarations from WasmJs stdlib we need to remove its usages from the coroutines.

### DIFF
--- a/kotlinx-coroutines-core/wasmJs/src/JSDispatcher.kt
+++ b/kotlinx-coroutines-core/wasmJs/src/JSDispatcher.kt
@@ -1,9 +1,10 @@
 package kotlinx.coroutines
 
-import org.w3c.dom.Window
 import kotlin.js.*
 
-public actual typealias W3CWindow = Window
+public actual abstract external class W3CWindow {
+    fun clearTimeout(handle: Int)
+}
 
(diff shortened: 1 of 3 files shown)


dkhalanskyjb:
Is there a place where we can read about the rationale of these changes to the stdlib?

qwwdfsad:
Public issue: https://youtrack.jetbrains.com/issue/KT-54299/Extract-org.w3c-declarations-to-separate-library-from-K-Wasm-Stdlib

qwwdfsad (approved):

dkhalanskyjb (approved):

kotlinx-coroutines-core/wasmJs/src/JSDispatcher.kt
```diff
@@ -1,9 +1,10 @@
 package kotlinx.coroutines
 
-import org.w3c.dom.Window
 import kotlin.js.*
 
-public actual typealias W3CWindow = Window
+public actual abstract external class W3CWindow {
```
ilya-g, line 5: Does it need to be public? it doesn't seem to be used in public API

PR #4098, opened 2024-04-12
Limited parallelism improvements

qwwdfsad:
While I figure out stabilization details, these are pretty straightforward changes that can be merged into minor releases

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
@@ -132,8 +132,8 @@ public abstract class CoroutineDispatcher :
      * For direct dispatchers, it is recommended to override this method
      * and provide a domain-specific implementation or to throw an [UnsupportedOperationException].
      *
-     * Implementations of this method are allowed to return `this` if the current dispatcher can satisfy the requested parallelism,
-     * for example when the maximal parallelism provided by the dispatcher is less than the requested one.
+     * Implementations of this method are allowed to return `this` if the current dispatcher already satisfies the parallelism requirement.
+     * For example, `Dispatchers.Main.limitedParallelism(1)` returns `Dispatchers.Main`, because the main dispatcher is already single-threaded.
      *
      * @throws IllegalArgumentException if the given [parallelism] is non-positive
      * @throws UnsupportedOperationException if the current dispatcher does not support limited parallelism views
--- a/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
@@ -69,7 +69,7 @@ public abstract class CoroutineDispatcher :
      * no more than [parallelism] coroutines are executed at the same time.
      *
      * This method does not impose restrictions on the number of views or the total sum of parallelism values,
-     * each view controls its own parallelism independently with the guarantee that effective parallelism
+     * each view controls its own parallelism independently with the guarantee that the effective parallelism
      * of all views cannot exceed the actual parallelism of the original dispatcher.
      *
      * The resulting dispatcher does not guarantee that the coroutines will always be dispatched on the same
@@ -94,7 +94,7 @@ public abstract class CoroutineDispatcher :
      * and at most 4 threads can exist in the system.
      *
      * Note that this example was structured in such a way that it illustrates the parallelism guarantees.
-     * In practice, it is usually better to use [Dispatchers.IO] or [Dispatchers.Default] instead of creating a
+     * In practice, it is usually better to use `Dispatchers.IO` or [Dispatchers.Default] instead of creating a
      * `backgroundDispatcher`.
      *
      * ### `limitedParallelism(1)` pattern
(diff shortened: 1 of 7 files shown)


kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -65,37 +65,76 @@ public abstract class CoroutineDispatcher :
 
     /**
      * Creates a view of the current dispatcher that limits the parallelism to the given [value][parallelism].
-     * The resulting view uses the original dispatcher for execution, but with the guarantee that
+     * The resulting view uses the original dispatcher for execution but with the guarantee that
      * no more than [parallelism] coroutines are executed at the same time.
      *
      * This method does not impose restrictions on the number of views or the total sum of parallelism values,
-     * each view controls its own parallelism independently with the guarantee that the effective parallelism
+     * each view controls its own parallelism independently with the guarantee that effective parallelism
```
dkhalanskyjb, line 72: Why? I think the article is needed here.

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -65,37 +65,76 @@ public abstract class CoroutineDispatcher :
...
+     * }
+     * ```
+     * Note that there is no guarantee that the underlying system thread will always be the same.
+     *
+     * ### Dispatchers.IO
+     *
+     * `Dispatcher.IO` is considered _elastic_ for the purposes of limited parallelism -- the sum of
+     * views is not restricted by the capacity of `Dispatchers.IO`.
+     * See `Dispatchers.IO` documentation for more details.
+     *
```
dkhalanskyjb, line 124: I'm not sure this section belongs here, but this idea *is* missing from one other place: `newFixedThreadPoolContext`.

Who cares if `Dispatchers.IO` is elastic? Those who are afraid to hit the ceiling. In my view, those are:

* People who want to use `limitedParallelism` on `Dispatchers.IO`. They are covered already by the `Dispatchers.IO` documentation, as they will need to look at that documentation to see the if the default number of threads is enough for them.
* People who believe `Dispatchers.IO` isn't suitable for them, because they already know that it doesn't have a lot of threads. What code do these people have? It's probably something like this:
```kotlin
val dispatcher = newFixedThreadPoolContext(nThreads = 100, name = "")
```
We have a recommendation to replace these things with `limitedParallelism`, but we could certainly elaborate on the exact usage pattern and why it's safe.

WDYT?

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
qwwdfsad: That's an omission, I've tried to rephrase this part multiple times and failed to find a better option

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
qwwdfsad: >Who cares if Dispatchers.IO is elastic?

I would say that for users who are looking for `limitedParallelism`, it's worth explicitly mentioning "hey, here is the common pattern that you might want to be aware of. It's briefly mentioned, see Dispatchers.IO" which seems a reasonable thing to do.

W.r.t. `newFixedThreadPoolContext` -- great idea, I'm on it

qwwdfsad: I tried to be brief there, PTAL

kotlinx-coroutines-core/concurrent/src/MultithreadedDispatchers.common.kt
```diff
@@ -19,7 +19,7 @@ import kotlin.jvm.*
  * associated native resources (threads or native workers). It should not be allocated in place,
  * should be closed at the end of its lifecycle, and has non-trivial memory and CPU footprint.
  * If you do not need a separate thread pool, but only have to limit effective parallelism of the dispatcher,
- * it is recommended to use [CoroutineDispatcher.limitedParallelism] instead.
+ * it is recommended to use `Dispatchers.IO.limitedParallelism(1)` or `Dispatchers.Default.limitedParallelism(1)` instead.
```
dkhalanskyjb, line 22: An interactive link to `limitedParallelism` could help, I think.

kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -65,37 +65,78 @@ public abstract class CoroutineDispatcher :
...
+     * ### Restrictions and implementation details
+     *
+     * The default implementation of `limitedParallelism` does not support direct dispatchers,
+     * such as executing the given runnable in place during [dispatch] calls.
+     * Any dispatcher that may return `false` from [isDispatchNeeded] is considered direct.
+     * For direct dispatchers, it is recommended to override this method
+     * and provide a domain-specific implementation or to throw an [UnsupportedOperationException].
+     *
+     * Implementations of this method are allowed to return `this` if the current dispatchercan satisfy the requested parallelism,
+     * for example when the maximal parallelsm provided by the dispatcher is less than the requested one.
```
dkhalanskyjb, line 136:
```suggestion
     * Implementations of this method are allowed to return `this` if the current dispatcher can satisfy the requested parallelism,
     * for example when the maximal parallelism provided by the dispatcher is less than the requested one.
```

dkhalanskyjb, line 136: Or to rephrase this entirely,
```suggestion
     * Implementations of this method are allowed to return `this` if the current dispatcher already satisfies the parallelism requirement.
     * For example, `Dispatchers.Main.limitedParallelism(1)` returns `Dispatchers.Main`, because the main dispatcher is already single-threaded.
```

PR #4099, opened 2024-04-15
Fix documentation about the default timeout for runTest

OliverO2:
The default timeout for runTest changed from 10 seconds to 60 seconds in 9a98eabed91c71cb2da9b0fe061139632e08990b. This PR makes the README reflect that change.

### DIFF
--- a/kotlinx-coroutines-test/README.md
+++ b/kotlinx-coroutines-test/README.md
@@ -107,7 +107,7 @@ on Kotlin/JS. The main differences are the following:
 
 * **The calls to `delay` are automatically skipped**, preserving the relative execution order of the tasks. This way,
   it's possible to make tests finish more-or-less immediately.
-* **The execution times out after 10 seconds**, cancelling the test coroutine to prevent tests from hanging forever 
+* **The execution times out after 60 seconds**, cancelling the test coroutine to prevent tests from hanging forever 
   and eating up the CI resources.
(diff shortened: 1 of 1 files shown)


dkhalanskyjb:
> This PR makes the README reflect that change

Good catch, thanks!

> includes documentation for the added system property

The idea was that the system property should very rarely be needed. If that's true, mentioning it in the `runTest` documentation should be enough. Do you think the need is widespread enough to be mentioned in the introductory overview that is the README?

OliverO2:
I'd also assume that the property would be used in rare cases (slow CI, many similar `runTest` invocations, JVM only). I don't mind if we'd waive mentioning it in the README.

Possible downside: If people run into timeout problems they would most probably consider adding the `timeout` parameter (hinted at by the README and assisted by the IDE). Will they read the `runTest` documentation on that occasion?

Having said that, it's no big deal for me: If you'd prefer, I can just drop that section on the property.

dkhalanskyjb:
> Will they read the `runTest` documentation on that occasion?

Fair, but will they remember to reread the README that they skimmed when they initially learned about the coroutines test framework?

> Having said that, it's no big deal for me: If you'd prefer, I can just drop that section on the property.

Yes, I think it's better to keep the README approachable and free of technical details.

dkhalanskyjb:
Thanks!

OliverO2:
My pleasure!

PR #4101, opened 2024-04-15
MultiplexFlow

odedniv:
This is useful when the source allows only a single subscription, but the data is needed by multiple users.

It's extremely hard to get this right, so I feel having a kotlinx implementation would be very useful.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/MultiplexFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/MultiplexFlow.kt
@@ -43,9 +43,9 @@ public class MultiplexFlow<K, V> internal constructor(private val multiplexer: M
      * `getAll`.
      */
     public operator fun get(vararg keys: K): Flow<V> = flow {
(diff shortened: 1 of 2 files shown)


odedniv:
Half way through the tests I realized I should probably get some feedback on the idea, so once I get that feedback I can continue implementing the missing test cases.

qwwdfsad:
Please start with an issue to explain the problem being solved and gather some feedback.

Then, we'll be able to see how it fits in `kotlinx.coroutines`, what is the others' feedback and will proceed with PR (see also: [`CONTRIBUTING.md`](https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md)).

Closing as "not enough use-cases",  we'll reopen once we decide it should be added

odedniv:
Created #4103. For future reference please ignore the last state of this PR in favor of the latest state of https://github.com/Kotlin/kotlinx.coroutines/compare/master...odedniv:kotlinx.coroutines:multiplex.

PR #4102, opened 2024-04-16
fix link to `Thread.uncaughtException Handler`: give the link to the interface itself; fix the number of DefaultDispatcher-worker in the output

PetrakovichVictoria:
1. fix the number of DefaultDispatcher-worker in the output: the output always gives the number 1 for worker, not 2:
Screenshot 2024-04-16 at 14 14 42

2. give the link to the `Thread.uncaughtException Handler` interface itself:
Now it will be:
Screenshot 2024-04-16 at 14 16 55

It was:
Screenshot 2024-04-16 at 14 16 47

### DIFF
--- a/docs/topics/exception-handling.md
+++ b/docs/topics/exception-handling.md
@@ -60,7 +60,7 @@ The output of this code is (with [debug](https://github.com/Kotlin/kotlinx.corou
 
 ```text
 Throwing exception from launch
-Exception in thread "DefaultDispatcher-worker-2 @coroutine#2" java.lang.IndexOutOfBoundsException
+Exception in thread "DefaultDispatcher-worker-1 @coroutine#2" java.lang.IndexOutOfBoundsException
 Joined failed job
 Throwing exception from async
 Caught ArithmeticException
@@ -73,7 +73,7 @@ Caught ArithmeticException
 It is possible to customize the default behavior of printing **uncaught** exceptions to the console.
 [CoroutineExceptionHandler] context element on a _root_ coroutine can be used as a generic `catch` block for
 this root coroutine and all its children where custom exception handling may take place.
-It is similar to [`Thread.uncaughtExceptionHandler`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler)).
+It is similar to [`Thread.uncaughtExceptionHandler`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.UncaughtExceptionHandler.html).
 You cannot recover from the exception in the `CoroutineExceptionHandler`. The coroutine had already completed
 with the corresponding exception when the handler is called. Normally, the handler is used to
 log the exception, show some kind of error message, terminate, and/or restart the application.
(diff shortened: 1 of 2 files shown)


PetrakovichVictoria:
@qwwdfsad fixed

docs/topics/exception-handling.md
```diff
@@ -73,7 +73,7 @@ Caught ArithmeticException
 It is possible to customize the default behavior of printing **uncaught** exceptions to the console.
 [CoroutineExceptionHandler] context element on a _root_ coroutine can be used as a generic `catch` block for
 this root coroutine and all its children where custom exception handling may take place.
-It is similar to [`Thread.uncaughtExceptionHandler`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setUncaughtExceptionHandler(java.lang.Thread.UncaughtExceptionHandler)).
+It is similar to [`Thread.uncaughtExceptionHandler`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.UncaughtExceptionHandler.html).
```
dkhalanskyjb, line 76: I think the link to the setter itself is more useful, so it's enough to fix it:
```suggestion
It is similar to [`Thread.uncaughtExceptionHandler`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setUncaughtExceptionHandler-java.lang.Thread.UncaughtExceptionHandler-).
```

kotlinx-coroutines-core/jvm/test/guide/test/ExceptionsGuideTest.kt
```diff
@@ -9,7 +9,7 @@ class ExceptionsGuideTest {
     fun testExampleExceptions01() {
         test("ExampleExceptions01") { kotlinx.coroutines.guide.exampleExceptions01.main() }.verifyExceptions(
             "Throwing exception from launch",
-            "Exception in thread \"DefaultDispatcher-worker-2 @coroutine#2\" java.lang.IndexOutOfBoundsException",
+            "Exception in thread \"DefaultDispatcher-worker-1 @coroutine#2\" java.lang.IndexOutOfBoundsException",
```
dkhalanskyjb, line 12: This is expected to vary from library version to library version (and, in more complex scenarios, from execution to execution) and isn't important: the number describes which of the many threads that were available managed to snatch the task and run it.

PetrakovichVictoria: I understand that this is a minor thing; just wanted to align outputs because I wasn't able to obtain another result, and I didn't want to make readers pay their attention to this difference. Can we leave it with my corrections, or should I revert them?

dkhalanskyjb: If you feel strongly about it, we can keep it, but be aware that this number may change arbitrarily at any point.

PetrakovichVictoria: > If you feel strongly about it

No, I don't, but if it actually doesn't matter if there's 1 or 2 – I'd offer to merge this PR as is.

PR #4106, opened 2024-04-18
Add optional name parameter to .limitedParallelism

### LINKED ISSUE #4023: Ability to specify a limited dispatcher name
**Use case**

We have specific dispatchers, and in coroutine dumps they all look like `LimitedDispatcher@xxxxxxxx`, and it's not immediately clear where the dispatcher comes from.

**The Shape of the API**

Make `kotlinx.coroutines.CoroutineDispatcher.limitedParallelism` accept `debugName: String?`. 

Alternatively, make an extension function `fun CoroutineDispatcher.namedDispatcher(debugName: String): CoroutineDispatcher` which will return a dispatcher, which delegates to `this` and has its `toString` overridden. It can return `this` as is when debug is off.

**Prior Art**

`kotlinx.coroutines.scheduling.LimitingDispatcher` has `name`.

qwwdfsad:
Fixes #4023

### DIFF
--- a/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
+++ b/kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
@@ -170,7 +170,9 @@ public abstract class kotlinx/coroutines/CoroutineDispatcher : kotlin/coroutines
 	public fun get (Lkotlin/coroutines/CoroutineContext$Key;)Lkotlin/coroutines/CoroutineContext$Element;
 	public final fun interceptContinuation (Lkotlin/coroutines/Continuation;)Lkotlin/coroutines/Continuation;
 	public fun isDispatchNeeded (Lkotlin/coroutines/CoroutineContext;)Z
-	public fun limitedParallelism (I)Lkotlinx/coroutines/CoroutineDispatcher;
+	public synthetic fun limitedParallelism (I)Lkotlinx/coroutines/CoroutineDispatcher;
+	public fun limitedParallelism (ILjava/lang/String;)Lkotlinx/coroutines/CoroutineDispatcher;
+	public static synthetic fun limitedParallelism$default (Lkotlinx/coroutines/CoroutineDispatcher;ILjava/lang/String;ILjava/lang/Object;)Lkotlinx/coroutines/CoroutineDispatcher;
 	public fun minusKey (Lkotlin/coroutines/CoroutineContext$Key;)Lkotlin/coroutines/CoroutineContext;
 	public final fun plus (Lkotlinx/coroutines/CoroutineDispatcher;)Lkotlinx/coroutines/CoroutineDispatcher;
 	public final fun releaseInterceptedContinuation (Lkotlin/coroutines/Continuation;)V
@@ -502,7 +504,7 @@ public class kotlinx/coroutines/JobSupport : kotlinx/coroutines/ChildJob, kotlin
 public abstract class kotlinx/coroutines/MainCoroutineDispatcher : kotlinx/coroutines/CoroutineDispatcher {
 	public fun <init> ()V
 	public abstract fun getImmediate ()Lkotlinx/coroutines/MainCoroutineDispatcher;
-	public fun limitedParallelism (I)Lkotlinx/coroutines/CoroutineDispatcher;
+	public fun limitedParallelism (ILjava/lang/String;)Lkotlinx/coroutines/CoroutineDispatcher;
 	public fun toString ()Ljava/lang/String;
 	protected final fun toStringInternalImpl ()Ljava/lang/String;
 }
--- a/kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
@@ -37,7 +37,7 @@ internal class LimitedDispatcher(
     @ExperimentalCoroutinesApi
     override fun limitedParallelism(parallelism: Int, name: String?): CoroutineDispatcher {
         parallelism.checkParallelism()
-        if (parallelism >= this.parallelism) return this
+        if (parallelism >= this.parallelism) return namedOrThis(name)
         return super.limitedParallelism(parallelism, name)
     }
 
@@ -129,3 +129,8 @@ internal class LimitedDispatcher(
 }
 
 internal fun Int.checkParallelism() = require(this >= 1) { "Expected positive parallelism level, but got $this" }
+
+internal fun CoroutineDispatcher.namedOrThis(name: String?): CoroutineDispatcher {
+    if (name != null) return NamedDispatcher(this, name)
+    return this
+}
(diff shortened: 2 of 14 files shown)


qwwdfsad:
Conceptually, the PR is ready.

Practically, I would like to polish and figure some binary compatibility details before merging it and ensure it is a reasonable thing to do in a non-major release

kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
```diff
@@ -502,7 +504,7 @@ public class kotlinx/coroutines/JobSupport : kotlinx/coroutines/ChildJob, kotlin
 public abstract class kotlinx/coroutines/MainCoroutineDispatcher : kotlinx/coroutines/CoroutineDispatcher {
 	public fun <init> ()V
 	public abstract fun getImmediate ()Lkotlinx/coroutines/MainCoroutineDispatcher;
-	public fun limitedParallelism (I)Lkotlinx/coroutines/CoroutineDispatcher;
```
qwwdfsad, line 505: FTW: This change is okay for JVM (https://docs.oracle.com/javase/specs/jvms/se9/html/jvms-6.html#jvms-6.5.invokevirtual), for K/N I'm figuring it out

qwwdfsad: It's also binary compatible for klibs, all good

kotlinx-coroutines-core/jvm/test/DispatchersToStringTest.kt
```diff
@@ -18,5 +18,10 @@ class DispatchersToStringTest {
         assertEquals("Dispatchers.Default.limitedParallelism(2)", Dispatchers.Default.limitedParallelism(2).toString())
         // Not overridden at all, limited parallelism returns `this`
         assertEquals("DefaultExecutor", (DefaultDelay as CoroutineDispatcher).limitedParallelism(42).toString())
+
+        assertEquals("filesDispatcher", Dispatchers.IO.limitedParallelism(1, "filesDispatcher").toString())
+        assertEquals("json", Dispatchers.Default.limitedParallelism(2, "json").toString())
+        // Not overridden at all, limited parallelism returns `this`
+        assertEquals("DefaultExecutor", (DefaultDelay as CoroutineDispatcher).limitedParallelism(42, "ignored").toString())
```
dkhalanskyjb, line 25: What are the reasons to take the non-orthogonal path here? I think it's much more intuitive to have names that are unconditionally applied.

qwwdfsad: My line of thinking is the following:

* The implementation allows returning `this` which is a nice property to have -- you don't have to care about the underlying/given pool capacity if not necessary
    * Also, mistakes are much more forgiving. E.g. now for cases [like that](https://github.com/mamoe/mirai/blob/dev/mirai-core-utils/src/commonTest/kotlin/net/mamoe/mirai/utils/testFramework/AssertNoCoroutineSuspension.kt#L49) nothing really happens, with the name forcing to return something else it's suddenly more subtle
* I don't really like the idea of being able to give dispatchers names when no new dispatcher is effectively created -- the probability of accidental aliasing just increases with not that many benefits
* Returning `LimitedDispatcher` for the sole purpose of naming is a suboptimal idea -- it slightly changes the observable behaviour and introduces some overhead. It kind of forces us to introduce one more implementation (`NamedDispatcher`?) and maintain it

dkhalanskyjb: > The implementation allows returning this which is a nice property to have

This can be preserved in the case of `parallelism >= dispatcher.parallelism && newName != null && newName != dispatcher.name`.

> I don't really like the idea of being able to give dispatchers names when no new dispatcher is effectively created -- the probability of accidental aliasing just increases with not that many benefits

Two bureaucratic benefits come to mind immediately:
* Principle of least surprise: specifically, if the user did provide a name for some reason, it's surprising to see it ignored.
* Principle of orthogonality: conceptually, now, the two arguments to `limitedParallelism` are linked together, while they could not be that.

The second concern is practical, I think. Let's look at the issue statement:

> We have specific dispatchers, and in coroutine dumps they all look like `LimitedDispatcher@xxxxxxxx`, and it's not immediately clear where the dispatcher comes from.

I see two possible cases:
1. `"Dispatchers.Default.limitedParallelism(4)"` is already good enough. Then why would we add a new `name` parameter at all?
2. `"Dispatchers.Default.limitedParallelism(4)"` is *not* good enough, and something like `IndexBuilding` is needed instead. In this case, why does the name revert to the not-good-enough `"Dispatchers.Default.limitedParallelism(4)"` option when there are few CPU cores?

As to the accidental aliasing issue: what dangers do you see in it?

> Returning LimitedDispatcher for the sole purpose of naming is a suboptimal idea -- it slightly changes the observable behaviour and introduces some overhead. It kind of forces us to introduce one more implementation (NamedDispatcher?) and maintain it

That's true. Why is this an issue, though? Is an extra call stack frame per `dispatch` a significant problem?

qwwdfsad: Note: discussed it internally and the rationale will be reflected in the public design doc

kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
```diff
@@ -34,10 +35,10 @@ internal class LimitedDispatcher(
     private val workerAllocationLock = SynchronizedObject()
 
     @ExperimentalCoroutinesApi
-    override fun limitedParallelism(parallelism: Int): CoroutineDispatcher {
+    override fun limitedParallelism(parallelism: Int, name: String?): CoroutineDispatcher {
         parallelism.checkParallelism()
-        if (parallelism >= this.parallelism) return this
-        return super.limitedParallelism(parallelism)
+        if (parallelism >= this.parallelism) return namedOrThis(name)
+        return super.limitedParallelism(parallelism, name)
```
dkhalanskyjb, line 41: (Minor) Since `checkParallelism` was already called, we could avoid calling `super` here, instead directly constructing a `LimitedDispatcher`.

dkhalanskyjb (approved):
`LimitedDispatcher` having its own `name` instead of always delegating this behavior to a `NamedDispatcher` is a nice idea.

There's still potential for tiny optimizations: for example, we could also override `limitedParallelism` on `NamedDispatcher` to allow in-depth inspection of `dispatcher` (throwing away the `NamedDispatcher` wrapper if it's no longer needed), or we could join nested `NamedDispatcher`s into one, etc.

I'm not sure, though, that this would be useful to anyone, and certainly not to a point where this would block adding this useful change.

kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
qwwdfsad: In things like that, I prefer leaving the double-check, but ensuring that the code is reused and has less entry point

PR #4109, opened 2024-04-22
Fix the ticker channel example giving wrong results on the website

dkhalanskyjb:
https://play.kotlinlang.org/ of the example that's being changed here is currently unreliable: the last line is occasionally `null`.

By increasing all time intervals twofold, we reduce the impact of the CPU scheduling in the constrained environment. With this change, the results are consistent across dozens of runs: <https://pl.kotl.in/uCa-60j43>

Originally reported by `@PetrakovichVictoria`

### DIFF
--- a/docs/topics/channels.md
+++ b/docs/topics/channels.md
@@ -576,25 +576,25 @@ import kotlinx.coroutines.channels.*
 
 //sampleStart
 fun main() = runBlocking<Unit> {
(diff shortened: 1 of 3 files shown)


qwwdfsad:
`knitCheck` fails

qwwdfsad (approved):

PR #4110, opened 2024-04-22
Fix increased memory consumption in JobSupport

dkhalanskyjb:
Fixup to 65ef6ea4b4cf200856db1614e0b4ab98e08fe6b0: that commit made `JobSupport` consume more memory by making the information about whether a callback should be invoked when the job enters the cancelling state be represented as a boolean flag and not part of the object class hierarchy.

Now that flag is a virtual function instead.


qwwdfsad (approved):
Thanks for spotting!

PR #4111, opened 2024-04-22
Fix broken API reference links to the Project Reactor Javadoc

dkhalanskyjb:
See <https://github.com/reactor/reactor-core/issues/3794>

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,6 +14,7 @@ lincheck_version=2.18.1
 dokka_version=1.9.20
 byte_buddy_version=1.10.9
 reactor_version=3.4.1
+reactor_docs_version=3.4.5
 reactive_streams_version=1.0.3
 rxjava2_version=2.2.8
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -14,6 +14,7 @@ lincheck_version=2.18.1
 dokka_version=1.9.20
 byte_buddy_version=1.10.9
 reactor_version=3.4.1
+reactor_docs_version=3.4.5
```
qwwdfsad, line 17: Shouldn't we update reactor version instead?

dkhalanskyjb: I thought we were trying to depend on the earliest version possible, so as not to force the users to upgrade to the new Reactor just because of our integration. There are some repositories with 3.3 and 3.4 versions of Reactor mentioned as dependencies.

qwwdfsad (approved):

PR #4115, opened 2024-04-24
Introduce SubclassOptInRequired to the codebase

dkhalanskyjb:
The initial stage of #3770

Marked:
* Job,
* Deferred and CompletableDeferred,
* Flow, SharedFlow, StateFlow,
* CloseableCoroutineDispatcher,
* CancellableContinuation,
* All of their `public` implementors.

### DIFF
--- a/kotlinx-coroutines-core/common/src/Annotations.kt
+++ b/kotlinx-coroutines-core/common/src/Annotations.kt
@@ -90,3 +90,15 @@ public annotation class ObsoleteCoroutinesApi
             "so stable API could be provided instead"
 )
 public annotation class InternalCoroutinesApi
+
+/**
+ * Marks declarations that cannot be safely inherited from.
+ */
+@Target(AnnotationTarget.CLASS)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING, message =
+    "This is a kotlinx.coroutines API that is not intended to be inherited from. " +
+        "Either new methods may be added in the future, which would break the inheritance, " +
+        "or correctly inheriting from it requires fulfilling contracts that may change in the future."
+)
+public annotation class BrittleForInheritanceCoroutinesApi
--- a/kotlinx-coroutines-core/common/src/Annotations.kt
+++ b/kotlinx-coroutines-core/common/src/Annotations.kt
@@ -97,8 +97,19 @@ public annotation class InternalCoroutinesApi
 @Target(AnnotationTarget.CLASS)
 @RequiresOptIn(
     level = RequiresOptIn.Level.WARNING, message =
-    "This is a kotlinx.coroutines API that is not intended to be inherited from. " +
+    "Inheriting from this kotlinx.coroutines API is unstable. " +
         "Either new methods may be added in the future, which would break the inheritance, " +
         "or correctly inheriting from it requires fulfilling contracts that may change in the future."
 )
 public annotation class BrittleForInheritanceCoroutinesApi
+
+/**
+ * Marks declarations that cannot be safely inherited from.
+ */
+@Target(AnnotationTarget.CLASS)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING, message =
+    "This is a kotlinx.coroutines API that is not intended to be inherited from, " +
+        "as the library may handle predefined instances of this in a special manner."
+)
+public annotation class UnsuitableForInheritanceCoroutinesApi
--- a/kotlinx-coroutines-core/common/src/Annotations.kt
+++ b/kotlinx-coroutines-core/common/src/Annotations.kt
@@ -111,7 +111,8 @@ public annotation class ExperimentalForInheritanceCoroutinesApi
     level = RequiresOptIn.Level.WARNING, message =
     "This is a kotlinx.coroutines API that is not intended to be inherited from, " +
         "as the library may handle predefined instances of this in a special manner. " +
+        "This will be an error in a future release. " +
         "If you need to inherit from this, please describe your use case in " +
-        "https://github.com/Kotlin/kotlinx.coroutines/issues, so that we can provide a stable API for inheritance."
+        "https://github.com/Kotlin/kotlinx.coroutines/issues, so that we can provide a stable API for inheritance. "
 )
 public annotation class InternalForInheritanceCoroutinesApi
--- a/kotlinx-coroutines-core/common/src/Annotations.kt
+++ b/kotlinx-coroutines-core/common/src/Annotations.kt
@@ -101,7 +101,7 @@ public annotation class InternalCoroutinesApi
         "Either new methods may be added in the future, which would break the inheritance, " +
         "or correctly inheriting from it requires fulfilling contracts that may change in the future."
 )
-public annotation class BrittleForInheritanceCoroutinesApi
+public annotation class ExperimentalForInheritanceCoroutinesApi
 
 /**
  * Marks declarations that cannot be safely inherited from.
@@ -110,6 +110,8 @@ public annotation class BrittleForInheritanceCoroutinesApi
 @RequiresOptIn(
     level = RequiresOptIn.Level.WARNING, message =
     "This is a kotlinx.coroutines API that is not intended to be inherited from, " +
-        "as the library may handle predefined instances of this in a special manner."
+        "as the library may handle predefined instances of this in a special manner. " +
+        "If you need to inherit from this, please describe your use case in " +
+        "https://github.com/Kotlin/kotlinx.coroutines/issues, so that we can provide a stable API for inheritance."
 )
-public annotation class UnsuitableForInheritanceCoroutinesApi
+public annotation class InternalForInheritanceCoroutinesApi
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -41,6 +41,7 @@ import kotlin.coroutines.intrinsics.*
  *    +-----------+
  * ```
  */
+@SubclassOptInRequired(BrittleForInheritanceCoroutinesApi::class)
 public interface CancellableContinuation<in T> : Continuation<T> {
     /**
      * Returns `true` when this continuation is active -- it has not completed or cancelled yet.
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -42,7 +42,7 @@ import kotlin.coroutines.intrinsics.*
  * ```
  */
 @OptIn(ExperimentalSubclassOptIn::class)
-@SubclassOptInRequired(UnsuitableForInheritanceCoroutinesApi::class)
+@SubclassOptInRequired(InternalForInheritanceCoroutinesApi::class)
 public interface CancellableContinuation<in T> : Continuation<T> {
     /**
      * Returns `true` when this continuation is active -- it has not completed or cancelled yet.
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -41,6 +41,7 @@ import kotlin.coroutines.intrinsics.*
  *    +-----------+
  * ```
  */
+@OptIn(ExperimentalSubclassOptIn::class)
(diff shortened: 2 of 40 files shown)


qwwdfsad:
Postponed until 2.0 because this feature doesn't work as intended now

dkhalanskyjb:
No idea why the `testCoroutinesTimeoutInheritanceWithNoTimeoutInDerived` test could fail as a result of these changes.

dkhalanskyjb:
Maybe a fluke. After a restart, the tests passed.

qwwdfsad (commented):
It feels like we have a mix of concepts here.

We have `SubclassOptInRequired` that is "Hey, we are not sure if we won't add new methods/change the contract. Be wary!" (examples: `SharedFlow`, maybe `Deferred`), and there are "These were never meant to be extended, but they technically might be because of the language restrictions or our omission" (`Job`, `CancellableContinuation`, `ChannelFlow`).

It feels off to have both covered with `Brittle*` with `WARNING` level.

For the later, I suggest using `SubclassOptInRequired(InternalCoroutinesApi::class)`, for the former -- keep it as is, but maybe rename it to something more straightforward -- `ExperimentalForInheritanceCoroutinesApi` (or `UnstableFor*`).

kotlinx-coroutines-core/common/src/flow/Flow.kt
```diff
@@ -173,6 +173,8 @@ import kotlin.coroutines.*
  * These implementations ensure that the context preservation property is not violated, and prevent most
  * of the developer mistakes related to concurrency, inconsistent flow dispatchers, and cancellation.
  */
+@OptIn(ExperimentalSubclassOptIn::class)
+@SubclassOptInRequired(BrittleForInheritanceCoroutinesApi::class)
```
qwwdfsad, line 177: Frankly, taking into account that `Flow` is **the** way to program and that it has been here for a long time, I doubt we ever can change that, and this warning might be just a headache to opt-in. I suggest we might give up on that

kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -13,6 +13,8 @@ import kotlin.coroutines.*
  * This class is generally used as a bridge between coroutine-based API and
  * asynchronous API that requires an instance of the [Executor].
  */
+@OptIn(ExperimentalSubclassOptIn::class)
+@SubclassOptInRequired(BrittleForInheritanceCoroutinesApi::class)
```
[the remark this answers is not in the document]
dkhalanskyjb: With `CloseableCoroutineDispatcher` being its typealias, I see no way around it.

kotlinx-coroutines-core/common/src/flow/Flow.kt
dkhalanskyjb: People who want custom `Flow` implementation should use `AbstractFlow` instead, so I think it's okay to discourage inheriting from `Flow`.

dkhalanskyjb:
Let's look at it in a case-by-case basis:

* `Job` is not an internal API, so I think it would be surprising to have `This is an internal kotlinx.coroutines API` when you try to inherit from it. We could extend the description of `InternalCoroutinesApi` with `If you get this issue when trying to inherit, this means that inheriting is specifically banned` or something like that, but I think this would overcomplicate the meaning of the annotation.
* `Deferred` and `CompletableDeferred` inherit from `Job`, so must be guarded by at least that strict a warning, which means they are also in the "must never be inherited from" category.
* `Flow`, `SharedFlow`, `StateFlow` are probably also interfaces that must never be inherited from directly: I don't see what one could gain from implementing `Flow` and not inheriting from `AbstractFlow`. If we think `SharedFlow` and `StateFlow` are useful things to extend, we probably must also provide the corresponding abstract classes.
* `CloseableCoroutineDispatcher` is the one clear-cut thing in this list: it's reasonable to extend it, but not stable.
* `CancellableContinuation` is in the same basket as `Job`.

To follow the spirit of your suggestion, I introduced another annotation that strongly states that this shouldn't be inherited from and applied it to `Job`, `CancellableContinuation`, and `Deferred`.

kotlinx-coroutines-core/common/src/Annotations.kt
```diff
@@ -90,3 +90,26 @@ public annotation class ObsoleteCoroutinesApi
...
+public annotation class BrittleForInheritanceCoroutinesApi
+
+/**
+ * Marks declarations that cannot be safely inherited from.
+ */
+@Target(AnnotationTarget.CLASS)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING, message =
+    "This is a kotlinx.coroutines API that is not intended to be inherited from, " +
+        "as the library may handle predefined instances of this in a special manner."
```
qwwdfsad, line 113: I would suggest keeping it consistent: `InternalForInheritance` (or `InternalForImplementation`) and adding the corresponding disclaimer to the message: `No compatibility guarantees are provided. It is recommended to report your use-case of internal API to kotlinx.coroutines issue tracker, so stable API could be provided instead`

kotlinx-coroutines-core/common/src/Annotations.kt
```diff
@@ -90,3 +90,26 @@ public annotation class ObsoleteCoroutinesApi
...
+ * Marks declarations that cannot be safely inherited from.
+ */
+@Target(AnnotationTarget.CLASS)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING, message =
+    "Inheriting from this kotlinx.coroutines API is unstable. " +
+        "Either new methods may be added in the future, which would break the inheritance, " +
+        "or correctly inheriting from it requires fulfilling contracts that may change in the future."
+)
+public annotation class BrittleForInheritanceCoroutinesApi
```
qwwdfsad, line 104: We now have three terms in our glossary -- experimental, brittle and unstable.

I suggest sticking with `experimental` as the most straightforward one -- most of our "Not stable for inheritance" mean "as new methods might be added to this interface in the future"

kotlinx-coroutines-core/common/src/flow/Flow.kt
qwwdfsad: I still argue that `Flow` is neither brittle nor experimental.

In my mental model, it's either `Delicate` (because of the invariants that must be preserved) or nothing (because it's quite hard to rationalize it otherwise). WDYT?

qwwdfsad (changes requested):

kotlinx-coroutines-core/common/src/flow/Flow.kt
dkhalanskyjb: In my mental model, it's certainly brittle, because it's a mistake to try to implement it without using `AbstractFlow`. It's not `Delicate` because we don't give realistic means of fulfilling the contracts in custom `Flow` instances that don't inherit from `AbstractFlow`.

kotlinx-coroutines-core/common/src/Annotations.kt
```diff
@@ -90,3 +90,28 @@ public annotation class ObsoleteCoroutinesApi
...
+ */
+@Target(AnnotationTarget.CLASS)
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING, message =
+    "This is a kotlinx.coroutines API that is not intended to be inherited from, " +
+        "as the library may handle predefined instances of this in a special manner. " +
+        "If you need to inherit from this, please describe your use case in " +
+        "https://github.com/Kotlin/kotlinx.coroutines/issues, so that we can provide a stable API for inheritance."
+)
+public annotation class InternalForInheritanceCoroutinesApi
```
whyoleg, line 117: `InternalCoroutinesApi` has `ERROR` level - may be it makes sense to do the same here?

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -41,6 +41,8 @@ import kotlin.coroutines.intrinsics.*
  *    +-----------+
  * ```
  */
+@OptIn(ExperimentalSubclassOptIn::class)
```
whyoleg, line 44: just in case, may be it could be easier to just `optIn` globally for `ExperimentalSubclassOptIn` via `compilerOptions`?

kotlinx-coroutines-core/common/src/Job.kt
```diff
@@ -404,6 +401,7 @@ public fun interface DisposableHandle {
  */
 @InternalCoroutinesApi
 @Deprecated(level = DeprecationLevel.ERROR, message = "This is internal API and may be removed in the future releases")
+@OptIn(InternalForInheritanceCoroutinesApi::class)
```
whyoleg, line 404: just curious: why not to propagate optIn here? Overall, it's still has a lot of annotations, which screams "don't use this", but still (same for other `Job` APIs like `ParentJob` and `JobSupport`)

kotlinx-coroutines-core/common/src/JobSupport.kt
```diff
@@ -19,6 +19,7 @@ import kotlin.jvm.*
  * @param active when `true` the job is created in _active_ state, when `false` in _new_ state. See [Job] for details.
  * @suppress **This is unstable API and it is subject to change.**
  */
+@OptIn(InternalForInheritanceCoroutinesApi::class)
```
whyoleg, line 22: is there a reason why `JobSupport` isn't annotated with `InternalCoroutinesApi` at all?

whyoleg (commented):
mostly just questions, feel free to ignore, if the questions are too silly :)

kotlinx-coroutines-core/common/src/JobSupport.kt
dkhalanskyjb: Yes: it's deprecated with an `ERROR`.

kotlinx-coroutines-core/common/src/Job.kt
dkhalanskyjb: To use a `Deprecated(ERROR)` API, you must suppress an error, which already proves your dedication and bravery beyond any doubt.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
dkhalanskyjb: It would be easier, but the way it is now, having to opt-in is helping us: in every place where there's an `ExperimentalSubclassOptIn` required, we explicitly decide whether to opt-in or to propagate.

kotlinx-coroutines-core/common/src/Annotations.kt
dkhalanskyjb: It does, but I think we should delay making this an `ERROR`.

Imagine a situation: you're not trying to do anything fancy in your code, you just got confused and accidentally overrode something you shouldn't have, and now upgrading the library prevents you from using your code that you know used to work. If you got confused, it means you don't understand that part of your codebase all that well and will need to allocate some time to fixing it. Warnings are more gentle.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
whyoleg: yeah, overall, I understand the difference and agree. But the case with `ExperimentalSubclassOptIn` is IMO a little more specific, as propagating it, will mean that on every usage of annotated class (not only inheriting) the warning will be issues about `ExperimentalSubclassOptIn` feature :)
So propagating this specific optIn doesn't make any sense to me. That's why I mentioned the `global` optIn

dkhalanskyjb: Oh, no, in this case, by "propagating (to the inheritor)" I mean adding `SubclassOptInRequired`. Sometimes, it's okay to inherit from an inheritor, but not from the base class directly, and sometimes, inheriting an inheritor is also undesirable.

kotlinx-coroutines-core/common/src/flow/Flow.kt
dkhalanskyjb: `by`-delegation to `Flow` does happen occasionally (https://grep.app/search?q=%20%3A%20Flow%3C%5B%5E%3E%5D%2B%3E%20by&regexp=true), and the following code does emit a warning:
```kotlin
public class MyFlow: Flow<Int> by flowOf(1, 2, 3) {
}
```

So, `Flow` probably shouldn't be marked as `InternalForInheritance`, at least, as we may want to make it difficult to opt in to that eventually, but this benign pattern should work without issues.

qwwdfsad (approved):

PR #4120, opened 2024-05-02
Fix WasmJS Promise.await() never resumes if promise rejected from JS with non-Kotlin exception

shepeliev:


dkhalanskyjb (changes requested):
Ok, thanks, looks good. Could you base the PR on top of `develop` instead of `master` (see https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md) and also port this fix to `kotlinx-coroutines-core/js/`, where this issue also exists?

shepeliev:
It seems that the behavior of evaluating JS exceptions has changed since Kotlin 2.0.0 was released. So the WasmJS doesn't have this issue anymore. When a promise rejects with a non-Kotlin exception, `await` resumes with a `JsException` without getting stuck.

On the other hand, JS target really still has the issue. I've added a couple of tests, ported the solution to JS target and reverted the changes to WasmJS `Promise.await()` since it now appears to work as expected.

dkhalanskyjb (approved):
Good job, thanks!

PR #4123, opened 2024-05-08
Adopt AutoCloseable from standard library for CloseableCoroutineDispa…

qwwdfsad:
…tcher

* It's stable in 2.0
* For JVM, use Java's AutoCloseable as on Kotlin 1.9.x there is no JVM counterpart


qwwdfsad:
I won't merge this PR though until we update on Kotlin 2.0, so JVM counterpart is handled in a less ad-hoc manner

dkhalanskyjb (approved):

PR #4127, opened 2024-05-10
Introduce basic Flow<T>.chunked operator

### LINKED ISSUE #1290: Flow.chunked operator with size limit

qwwdfsad:
Fixes #1290


qwwdfsad:
A few things about this PR:

* It's a long overdue we introduce `chunked` operator, and whatever we do, `chunked(n)` is likely to be here as the shorthand syntax
* It also might bootstrap the discussion and break out analysis paralysis
* `transform` version (`chunked(n, transform)`) is not added even though the stdlib has one -- we have a strong preference for proper separation of concerns. The documentation sample provides a clear substitute
* Meanwhile, I'll experiment with duration overload, but it's less straightforward


API shape brain dump (eventually I'll move it towards a better-structured issue, but not there yet):

There exists a whole bunch of possibilities, use cases and things full-blown `chunked` family might/should address

* Kotlin's `windowed` -- the same, but buffers are overlapped
* Reactive `windowed` -- the same as `chunked`, but it yields `Flow<Flow<T>>`. #2989
* Time-based chunking: 
    * The most straightforward and common -- keep the buffer open for the given duration, then emit it if it's not empty. Start the new buffer
    * A whole variety of supplemental things -- with time skips, predicate-based buffering
* Batching: group while you can (or the limit exceeded), emit it and wait the next element.

The problem with time is that it requires the actual concurrency, while the size-based version does not. I'm still exploring the possibility of something like `chunkedTransform { /* decide whether it's time-based or not */ }`, but not much of ideas for now.

kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
```diff
@@ -129,3 +129,41 @@ public fun <T> Flow<T>.runningReduce(operation: suspend (accumulator: T, value:
...
+            if (acc.size == size) {
+                emit(acc)
+                // Cleanup, but don't allocate -- it might've been the case this is the last element
+                result = null
+            }
+        }
+        val acc = result ?: return@flow
+        if (acc.isNotEmpty()) {
+            emit(acc)
+        }
```
dkhalanskyjb, line 167: A non-null `result` is always non-empty, right?
```suggestion
        result?.let { emit(it) }
```

kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
```diff
@@ -129,3 +129,38 @@ public fun <T> Flow<T>.runningReduce(operation: suspend (accumulator: T, value:
         emit(accumulator as T)
     }
 }
+
+/**
+ * Splits the given flow into a flow of non-overlapping lists each not exceeding the given [size] but never empty.
+ * The last emitted list may have fewer elements than the given size.
```
dovchinnikov, line 135: Please name the parameter `maxSize` then

qwwdfsad: That was my original intent, but stdlib disagrees on that: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/chunked.html

A nice thing to consider tho

dkhalanskyjb (approved):

PR #4129, opened 2024-05-14
Update to Kotlin 2.0-RC3

qwwdfsad:
Apply all workarounds from k2 branch


kotlinx-coroutines-test/js/src/internal/JsPromiseInterfaceForTesting.kt
```diff
@@ -0,0 +1,19 @@
+package kotlinx.coroutines.test.internal
+
+/* This is a declaration of JS's `Promise<Unit>`. We need to keep it a separate class, because
```
lppedd, line 3: Worth linking to https://youtrack.jetbrains.com/issue/KT-60561 ?

I just hope it gets fixed soon tbh, this workaround means everyone is going to have their own promise types, a mess.

dkhalanskyjb (changes requested):
Compilation fails. I think the reason is that the Kotlin version in `integration-testing` should also be updated.

kotlinx-coroutines-test/js/src/internal/JsPromiseInterfaceForTesting.kt
lppedd: Well it's not really just for testing. I have some typealiases to `Promise<T>` in my KMP project and I can work with them in K1. In K2 I have to remove the typealias or have to create an intermediate class, just like `JsPromiseInterfaceForTesting`.

dkhalanskyjb: `typealias Future<T> = Promise<T>` should still work. Is your use case covered by it?

lppedd: It works because you have the type parameter in `Future`, but something like `typealias HostValue = Promise<Int>` wont' work. This plus KT-61043 makes working with typealiases a lot more cumbersome.

dkhalanskyjb (approved):

PR #4135, opened 2024-05-27
Promote CoroutineDispatcher.limitedParallelism to stable

### LINKED ISSUE #3864: Stabilize CoroutineDispatcher.limitedParallelism
In general, we received mostly positive feedback about the API shape, and there is no reason to keep it experimental any further.

Few things to address during stabilization:

* [x] Ensure that doc explicitly states that it limits parallelism and not the number of threads (e.g. `limitedParallelism(1)` does not re-use the same thread all the time)
* [x] Ensure that `CoroutineContext` in `dispatch` is either transparent or that it can be transparent to the underlying dispatcher to control scheduling tweaks
* [x] Document `isDispatchNeeded` and unconfined-like behaviour for limited parallelism
* [x] Mention `limitedParallelism(1)` properties
* [x] Consider https://github.com/Kotlin/kotlinx.coroutines/issues/4023

qwwdfsad:
Fixes #3864

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
@@ -140,7 +140,6 @@ public abstract class CoroutineDispatcher :
      * @throws IllegalArgumentException if the given [parallelism] is non-positive
      * @throws UnsupportedOperationException if the current dispatcher does not support limited parallelism views
      */
-    @ExperimentalCoroutinesApi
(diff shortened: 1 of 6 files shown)


qwwdfsad:
Design notes will follow the moment I'll unblock you with everything release-related

dkhalanskyjb (approved):
https://github.com/Kotlin/kotlinx.coroutines/pull/4132 raises a good concern about a version of `limitedParallelism` with soft and hard limits, but it looks like the API form we have now will be useful in any case.

PR #4144, opened 2024-05-30
Clarify documentation of `launchIn`

rohandhruva:
This line of the `launchIn` documentation is a bit confusing: I think it applies to all general usage of `launchIn`, which left me wondering why the returned `Job` is ignored. 

This is a really small tweak to make this clear. As an alternate, I think we can also just remove the line altogether and avoid any confusion. 

See also: https://kotlinlang.slack.com/archives/CRJCTR5PD/p1717093849474799


qwwdfsad (approved):
Thanks!

Issue #4133, opened 2024-05-26
Improve the API reference


globsterg:
**Describe the bug**

`kotlinx-datetime` was given new and updated documentation... `kotlinx-corotoutines` could also benefit from the same treatment.

**Provide a Reproducer**

Going through the list from `kotlinx-datetime` (https://github.com/Kotlin/kotlinx-datetime/issues/347):

* Extensively document what the entity represents, ~whether it follows ISO-8601~, how the entity can be acquired, ~and what time scale it comes from~. [Example](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-swing/kotlinx.coroutines.swing/-swing-dispatcher/)
* Provide a minimum of how the entity can be used. [Example](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.intrinsics/start-coroutine-cancellable.html)
* Thoroughly explain the allowed and expected contract and behaviours and system-specifics. [Example](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx3/kotlinx.coroutines.rx3/-scheduler-coroutine-dispatcher/)
* If applicable, provide contexts in which the entity is expected/recommended to be used (e.g. our README shines at that) and how it interacts with other (core) entities. [Example](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/test-time-source.html)
* Provide a basic usage example. [Example](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-semaphore/)
* If applicable, provide details about conversion rules, gotchas and recommendations. [Example](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.future/as-completable-future.html) (it's not mentioned what happens when cancellation goes the other way)
* Where appropriate, information about conversions and interactions with native types (e.g. ~Instant <-> Java Instant~ CoroutineDispatcher <-> Java Executor conversion)
* Ensure that information used in our README is, in fact, deducible from the core entities documentation

globsterg:
If you allow it, I would like to contribute to this myself. I am implementing my own programming language and need to learn more about the real-world asynchronous runtimes before deciding on what the concurrency story will look like, and a deep dive into an existing framework would help me tremendously.

Basically, it's a trade offer.

[...]

PR #4148, opened 2024-06-04
Extend the KDoc for some channel APIs

globsterg:
One of the pieces of https://github.com/Kotlin/kotlinx.coroutines/issues/4133.

Each file can be reviewed independently (and changes to other files will be added later). I'm keeping this a single pull request to avoid spamming.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
@@ -144,7 +144,9 @@ public inline fun <E, R> ReceiveChannel<E>.consume(block: ReceiveChannel<E>.() -
  *         if (it == 42) {
  *             println("Found the answer")
  *             return@launch
- *         } else it.cleanup()
+ *         } else {
+ *             it.cleanup()
+ *         }
  *     }
  * }
  * ```
--- a/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
@@ -62,7 +62,9 @@ public fun <E : Any> ReceiveChannel<E>.onReceiveOrNull(): SelectClause1<E?> {
  * ```
  * suspend fun <E> ReceiveChannel<E>.consumeFirst(): E =
  *     consume { return receive() }
+ * 
  * fun Int.cleanup() { println("cleaning up $this") }
+ *
  * val channel = Channel<Int>(10, onUndeliveredElement = Int::cleanup)
  * // Launch a procedure that creates values
  * launch(Dispatchers.Default) {
--- a/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
@@ -49,46 +49,43 @@ public fun <E : Any> ReceiveChannel<E>.onReceiveOrNull(): SelectClause1<E?> {
 }
 
 /**
- * Executes the [block] and then [cancels][ReceiveChannel.cancel] the channel,
- * ensuring that all the elements that were sent are processed by either [block] or [ReceiveChannel.cancel].
+ * Executes the [block] and then [cancels][ReceiveChannel.cancel] the channel.
  *
  * It is guaranteed that, after invoking this operation, the channel will be [cancelled][ReceiveChannel.cancel], so
  * the operation is _terminal_.
- * If the [block] finishes with an exception, that exception will be used for cancelling the channel.
+ * If the [block] finishes with an exception, that exception will be used for cancelling the channel and rethrown.
  *
- * This function is useful for building more complex terminal operators while ensuring that no elements will be lost.
+ * This function is useful for building more complex terminal operators while ensuring that the producers stop sending
+ * new elements to the channel.
+ *
  * Example:
- *
  * ```
  * suspend fun <E> ReceiveChannel<E>.consumeFirst(): E =
- *     consume { return receive() }
- * 
- * fun Int.cleanup() { println("cleaning up $this") }
- *
- * val channel = Channel<Int>(10, onUndeliveredElement = Int::cleanup)
- * // Launch a procedure that creates values
- * launch(Dispatchers.Default) {
- *     repeat(10) {
- *         val sendResult = channel.trySend(it)
- *         if (sendResult.isFailure) {
- *             print("in the producer: ")
- *             it.cleanup()
- *         }
- *         yield()
+ *    consume { return receive() }
+ * // Launch a coroutine that constantly sends new values
+ * val channel = produce(Dispatchers.Default) {
+ *     var i = 0
+ *     while (true) {
+ *         // Will fail with a `CancellationException`
+ *         // after `consumeFirst` finishes.
+ *         send(i++)
  *     }
  * }
  * // Grab the first value and discard everything else
- * launch(Dispatchers.Default) {
- *     val firstElement = channel.consumeFirst()
- *     println("received $firstElement")
- * }
+ * val firstElement = channel.consumeFirst()
+ * check(firstElement == 0)
+ * // *Note*: some elements could be lost in the channel!
  * ```
  *
- * In this example, all ten values created by the producer coroutine will be processed: one by `consumeFirst`,
- * and the other ones by `Int.cleanup`, invoked either by [ReceiveChannel.cancel] inside [consume] or by the
- * producer itself when it observes failure.
- * In any case, exactly nine elements will go through a cleanup in this example.
- * If `consumeFirst` is implemented as `for (e in this) { return e }` instead, the cleanup does not happen.
+ * In this example, the channel will get closed, and the producer coroutine will finish its work after the first
+ * element is obtained.
+ * If `consumeFirst` was implemented as `for (e in this) { return e }` instead, the producer coroutine would be active
+ * until it was cancelled some other way.
+ *
+ * [consume] does not guarantee that new elements will not enter the channel after [block] finishes executing, so
+ * some channel elements may be lost.
+ * Use the `onUndeliveredElement` parameter of a manually created [Channel] to define what should happen with these
+ * elements during [ReceiveChannel.cancel].
  */
 public inline fun <E, R> ReceiveChannel<E>.consume(block: ReceiveChannel<E>.() -> R): R {
     contract {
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -49,10 +49,44 @@ public fun <E : Any> ReceiveChannel<E>.onReceiveOrNull(): SelectClause1<E?> {
...
+ *     }
+ * }
+ * // Grab the first value and discard everything else
+ * launch(Dispatchers.Default) {
+ *     val firstElement = channel.consumeFirst()
+ *     println("received $firstElement")
+ * }
+ * ```
+ *
+ * In this example, all ten values created by the producer coroutine will be processed: one by `consumeFirst`,
```
qwwdfsad, line 85: Reading this, it seems like the example is mixing two concepts: one is about consuming, and the other about undelivered clean up (which we don't consider as "processed", it has a different meaning).

Maybe something more straightforward?
Like the following (outline, haven't tried to compile it):
```
val producer = produce {
    var currentElement = 0
    try {
        while(true) {
            send(currentElement++)
        }
    } finally {
        println("Cancelled after: $currentElement")
    }
}

consumeFirst { ... }
```

It's deterministic, more trivial and straightforward. WDYT?

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -70,23 +104,83 @@ public inline fun <E, R> ReceiveChannel<E>.consume(block: ReceiveChannel<E>.() -
...
  *
  * The operation is _terminal_.
  * This function [consumes][ReceiveChannel.consume] all elements of the original [ReceiveChannel].
+ *
+ * This function is useful in cases when this channel is only expected to have a single consumer that decides when
+ * the producer may stop and ensures that the elements that were sent do get processed.
+ * Example:
+ *
+ * ```
+ * fun Int.cleanup() { println("cleaning up $this") }
```
qwwdfsad, line 127: Ditto about `onUndeliveredElement` and `cleanup`, it doesn't really seem relevant to these functions

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -70,23 +104,83 @@ public inline fun <E, R> ReceiveChannel<E>.consume(block: ReceiveChannel<E>.() -
...
+ *         }
+ *     }
+ * }
+ * // Launch the exclusive consumer
+ * launch(Dispatchers.Default) {
+ *     channel.consumeEach {
+ *         if (it == 42) {
+ *             println("Found the answer")
+ *             return@launch
+ *         } else it.cleanup()
```
qwwdfsad, line 145: Please wrap it in braces

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -70,23 +104,83 @@ public inline fun <E, R> ReceiveChannel<E>.consume(block: ReceiveChannel<E>.() -
...
         for (e in this) action(e)
     }
 
 /**
- * Returns a [List] containing all elements.
+ * Returns a [List] containing all the elements sent to this channel, preserving their order.
+ *
+ * This function will attempt to receive elements and put them into the list until the channel is
+ * [closed][SendChannel.close].
+ * Calling [toList] without closing the channel is always incorrect:
```
qwwdfsad, line 164: Let's make it a bit less ambigous? "Calling toList on channels that are not closed" or something like that

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -70,23 +104,83 @@ public inline fun <E, R> ReceiveChannel<E>.consume(block: ReceiveChannel<E>.() -
...
+ * If the channel is [closed][SendChannel.close] with a cause, [toList] will rethrow that cause.
  *
  * The operation is _terminal_.
  * This function [consumes][ReceiveChannel.consume] all elements of the original [ReceiveChannel].
+ *
+ * Example:
+ * ```
+ * val values = listOf(1, 5, 2, 9, 3, 3, 1)
+ * val channel = Channel<Int>()
+ * GlobalScope.launch {
```
qwwdfsad, line 178: Let's use `produce` for that? 
Esp. here, where `finally` block is required otherwise

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -19,25 +19,43 @@ public interface ProducerScope<in E> : CoroutineScope, SendChannel<E> {
...
- *
  * Example of usage:
  * ```
  * val callbackEventsStream = produce {
  *     val disposable = registerChannelInCallback(channel)
  *     awaitClose { disposable.dispose() }
  * }
  * ```
+ *
+ * Internally, [awaitClose] is implemented using [SendChannel.invokeOnClose].
```
qwwdfsad, line 45: Nice catch!

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -19,25 +19,43 @@ public interface ProducerScope<in E> : CoroutineScope, SendChannel<E> {
...
+ * Currently, every channel can have at most one [SendChannel.invokeOnClose] handler.
+ * This means that calling [awaitClose] several times in a row or combining it with other [SendChannel.invokeOnClose]
+ * invocations is prohibited.
+ * An [IllegalStateException] will be thrown if this rule is broken.
+ *
+ * **Pitfall**: when used in [produce], if the channel is [cancelled][ReceiveChannel.cancel], [awaitClose] can either
```
qwwdfsad, line 51: Could you please share a reproducer for that?

The documentation here looks like a bug "In deterministic scenarios, the behaviour here is unspecified and racy"

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -58,35 +76,151 @@ public suspend fun ProducerScope<*>.awaitClose(block: () -> Unit = {}) {
...
+ * The channel is [closed][SendChannel.close] when the coroutine completes.
+ *
+ * ```
+ * val values = listOf(1, 2, 3, 4)
+ * val channel = produce<Int> {
+ *     for (value in values) {
+ *         send(value)
+ *     }
+ * }
+ * check(channel.toList() == values)
```
qwwdfsad, line 96: That's a nice example for `toList` as well!

qwwdfsad (changes requested):
well done!

kotlinx-coroutines-core/common/src/channels/Produce.kt
globsterg: Yes, here is the reproducer: https://github.com/Kotlin/kotlinx.coroutines/issues/4149

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
globsterg: I found `produce` too high-level and thought that using it may detract from the point of the sample. Is it wrong? Is `produce` common knowledge and part of any coroutine user's toolkit?

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
globsterg: Yes, good point. The point of `consume` seems to be just notifying the producer that there are no more consumers. I noticed that `consume` is a reliable way not to miss any elements and became too fixated on that, but this really is a just side effect.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: `produce` is more high-level, but it's also the idiomatic way to create a producer coroutine, so it's worth mentioning. It's true that it may overwhelm the reader, but we can probably work around this by adding extensive comments.

kotlinx-coroutines-core/common/src/channels/Produce.kt
```diff
@@ -58,35 +76,187 @@ public suspend fun ProducerScope<*>.awaitClose(block: () -> Unit = {}) {
...
+ *         send(it)
+ *         println("Sent $it")
+ *     }
+ * }
+ * channel.cancel() // no elements can be received after this!
+ * ```
+ *
+ * There is no way to recover these lost elements.
+ * If this is unsuitable, please create a [Channel] manually and pass the `onUndeliveredElement` callback to the
+ * constructor.
```
qwwdfsad, line 183: This example is irrelevant to `produce` documentation, it's `Channel` doc that should explain that

qwwdfsad (approved):
Thanks! All good except one redundant example

kotlinx-coroutines-core/common/src/channels/Produce.kt
dkhalanskyjb: I beg to differ. `produce` calls `Channel` function, passing a `null` to `onUndeliveredElement`. This is *`produce`'s* implementation detail (that we must document) and not the `Channel`'s.

dkhalanskyjb: Or do you mean the snippet below? It's not an example, it's an explanation of how to approximate the `produce` above with an explicitly created channel. Would it work for you if I explain this in text explicitly?

qwwdfsad: I mean the snippet itself, yes. I would say it's sufficient to refer to `onUndeliveredElement` where it has a more comprehensive explanation and self-sufficient example

dkhalanskyjb: Non-obvious things this snippet highlights:

* Where the `onUndeliveredParameter` is (you can't refer to it directly, the user has to chase the docs)
* `finally` with the `close` in the coroutine

qwwdfsad: >Where the onUndeliveredParameter is (you can't refer to it directly, the user has to chase the docs)

* "If this is unsuitable, please create a [Channel] manually and pass the `onUndeliveredElement` callback to the constructor: [Channel(onUndeliveredElement = ...)]]Channel]" probably will suffice.

I realize that using it might be less obvious, but still believe it's not the produce's documentation job to explain that

PR #4157, opened 2024-06-18
Turn off optimization mode when compiling tests for K/N

homuroll:
Optimization was turned on due to performance problems. However, with new K/N memory manager these performance problems seem to be fixed. On the other hand, compiling in optimized mode significantly increases the compiler's memory consumption, so it's better to turn it off.
In particular, we (in K/N team) are having some problems with updating to LLVM 16 due to increased memory consumption (it's still being investigated what causes it), more details here: https://youtrack.jetbrains.com/issue/KT-68418/Native-updating-to-LLVM-16-makes-kotlinx.coroutines-build-in-Aggregate-unstable. This commit workarounds it.

### DIFF
--- a/kotlinx-coroutines-core/build.gradle.kts
+++ b/kotlinx-coroutines-core/build.gradle.kts
@@ -95,13 +95,8 @@ kotlin {
      * All new MM targets are build with optimize = true to have stress tests properly run.
      */
     targets.withType(KotlinNativeTargetWithTests::class).configureEach {
-        binaries.getTest(DEBUG).apply {
-            optimized = true
-        }
-
         binaries.test("workerTest", listOf(DEBUG)) {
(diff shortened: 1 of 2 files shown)


qwwdfsad (approved):
There is a tiny degradation in Native test times and compilation speed improvement.

Waiting for the stats from Igor and it's good to go

homuroll:
Some numbers:
- 2nd run of `./gradlew clean :kotlinx-coroutines-core:linkDebugTestMacosArm64` sped up from ~69 sec to ~19 sec
- but the running time of the test binary increased from ~38 sec to ~49.5 sec
The difference between the optimized binary and the non-optimized one is noticeable but I believe it's not that bad as it's used to be and won't be a problem.

PR #4160, opened 2024-06-19
Fix opt-in usage error shown in IntelliJ IDEA for some of the project files

KirpichenkovPavel:
Problem:
IntelliJ IDEA reports opt-in usage errors despite the fact that opt-ins are configured for all the compilations correctly. The reported error doesn't result in broken compilations or tests.

Motivation for the change:
The problem is not caused by the K2 IDE analysis mode, but it affects the K2 IDE quality gates.
See: https://youtrack.jetbrains.com/issue/KT-68642

Explanation:
Before the release of KGP 2.0.0 there was no clear way to set up compiler arguments for particular Kotlin multiplatform source sets: ones that participate in multiple compilations but are not a default source set for any compilation. Examples of such source sets are shared test source sets (they are not compiled into Kotlin metadata) and the unique `jdk8` source set used in `kotlinx-coroutines`. The 2.0.0 release provided a way to set up defaults for all multiplatform source sets in the whole Gradle project. They do behave as defaults and can still be overwritten by compilations.

Solution:
Use the new (experimental) Kotlin Gradle Plugin DSL to set up default compiler arguments and opt-ins for all multiplatform source sets. To keep arguments consistent between source sets and compilations, they have been extracted and shared between the multiplatform convention plugin and the compilation convention plugin.

### DIFF
--- a/buildSrc/src/main/kotlin/GlobalKotlinCompilerOptions.kt
+++ b/buildSrc/src/main/kotlin/GlobalKotlinCompilerOptions.kt
@@ -1,10 +1,9 @@
 import org.jetbrains.kotlin.gradle.dsl.KotlinCommonCompilerOptions
 
 internal fun KotlinCommonCompilerOptions.configureGlobalKotlinArgumentsAndOptIns() {
-    freeCompilerArgs.addAll("-progressive", "-Xexpect-actual-classes")
+    freeCompilerArgs.addAll("-progressive")
     optIn.addAll(
         "kotlin.experimental.ExperimentalTypeInference",
-        "kotlin.ExperimentalMultiplatform",
         // our own opt-ins that we don't want to bother with in our own code:
         "kotlinx.coroutines.DelicateCoroutinesApi",
         "kotlinx.coroutines.ExperimentalCoroutinesApi",
--- a/buildSrc/src/main/kotlin/GlobalKotlinCompilerOptions.kt
+++ b/buildSrc/src/main/kotlin/GlobalKotlinCompilerOptions.kt
@@ -1,0 +1,15 @@
+import org.jetbrains.kotlin.gradle.dsl.KotlinCommonCompilerOptions
+
+internal fun KotlinCommonCompilerOptions.configureGlobalKotlinArgumentsAndOptIns() {
+    freeCompilerArgs.addAll("-progressive", "-Xexpect-actual-classes")
+    optIn.addAll(
+        "kotlin.experimental.ExperimentalTypeInference",
+        "kotlin.ExperimentalMultiplatform",
+        // our own opt-ins that we don't want to bother with in our own code:
+        "kotlinx.coroutines.DelicateCoroutinesApi",
+        "kotlinx.coroutines.ExperimentalCoroutinesApi",
+        "kotlinx.coroutines.ObsoleteCoroutinesApi",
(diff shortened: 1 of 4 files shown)


KirpichenkovPavel:
Sure. You are absolutely right that the IDE doesn't use flags from platform compilations for shared multiplatform source sets. In the CLI environment there is no question which flags to use. Once a compilation task is launched, we know exaclty the flags the compilation uses. But for shared source sets this has a consequence: multiple compilations with potntially different flags can include the very same source set. In other words, the flags for the same source sets are different depending on the compilation! To make things more difficult, some shared source sets take part in metadata compilations with their own set of flags. The IDE has to represent potentially conflicting flags from all of these compilations. AFAIK, currently the IDE doesn't  use any logic for merging arguments from different compilations and just applies the compilation arguments to their default source set. This approach is not ideal, of course, and leads to issues such as [KT-68642](https://youtrack.jetbrains.com/issue/KT-68642) I mentioned before. [KT-57292](https://youtrack.jetbrains.com/issue/KT-57292) tries to address the issue, and while it doesn't add any argument merging rules, it does provide a way to at least set up arguments in an IDE-friendly way. I used [this explanation](https://youtrack.jetbrains.com/issue/KT-57292/Rework-configuration-of-compiler-settings-in-MPP-Projects#focus=Comments-27-9366553.0-0) to set them up for coroutines. This is the best I can do right now, since the overall argument importing logic is not under my control.

globsterg:
Oh, interesting, thank you! Sounds like the logic on the side of the IDE should be "if every compilation of a source set has an opt-in flag, this source set as a whole has an opt-in flag," but if an opt-in truly is defined for every viable compilation, the user can just specify that the subproject as a whole (or an individual source-set) has these opt-ins. Makes sense.

1. In this case, maybe [configure-compilation-conventions.gradle.kts](https://github.com/Kotlin/kotlinx.coroutines/pull/4160/files#diff-31e53ce14d22bfc07895e3280e2d6ed4947f974c5e7621912a61e42d151afb13) shouldn't mention these opt-ins, and instead, the [kotlin-jvm-conventions.gradle.kts](https://github.com/Kotlin/kotlinx.coroutines/blob/master/buildSrc/src/main/kotlin/kotlin-jvm-conventions.gradle.kts) file should contain this logic? According to https://github.com/Kotlin/kotlinx.coroutines/blob/1bffe67a32d9d0285320f5b23fa94bc2b5f2b92e/build.gradle.kts#L103-L113, all subprojects use either `kotlin-jvm` or `kotlin-multiplatform` convention file.
2. I wonder if there is a way to fix these flags as well: https://github.com/Kotlin/kotlinx.coroutines/pull/4160/files#diff-31e53ce14d22bfc07895e3280e2d6ed4947f974c5e7621912a61e42d151afb13R32-R38 Looking through the issues you linked, I found https://youtrack.jetbrains.com/issue/KT-63930/Convenient-way-to-apply-opt-ins-and-specify-compiler-arguments-in-source-set-hierarchies, which seems to describe exactly the problem at hand, but that issue is still open.

KirpichenkovPavel:
1. I've moved the opt-ins to the JVM part, thanks for the suggestion
2. Yes, looks like the issue is still open. I don't know a solution either, sorry

qwwdfsad (approved):
Thanks for the detailed explanation, all good

PR #4164, opened 2024-06-30
Fix links to external documentation in Timeout

Sejsel:
Currently, it ends up rendering wrong in the [API reference](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html).

image


qwwdfsad (approved):
Thanks!

PR #4165, opened 2024-07-02
Fix race condition in `ListenableFutureTest.testFutureIsDoneAfterChildrenCompleted`.

chaoren:
The `foo` coroutine could be cancelled before it's started, which would immediately trigger the `expect(3)` in `foo.invokeOnCompletion`. This could happen before `expect(2)` and fail the test.

### DIFF
--- a/integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
+++ b/integration/kotlinx-coroutines-guava/test/ListenableFutureTest.kt
@@ -627,16 +627,15 @@ class ListenableFutureTest : TestBase() {
     fun testFutureIsDoneAfterChildrenCompleted() = runTest {
         expect(1)
         val testException = TestException()
+        val latch = CountDownLatch(1)
         // Don't propagate exception to the test and use different dispatchers as we are going to block test thread.
         val future = future(context = NonCancellable + Dispatchers.Default) {
-            val foo = async {
+            val foo = async(start = CoroutineStart.UNDISPATCHED) {
                 try {
                     delay(Long.MAX_VALUE)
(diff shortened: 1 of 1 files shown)


dkhalanskyjb (changes requested):
Why do you consider this a problem? I think that the test should pass either way, whether `foo` starts or not. For example, if we replace the future's body with this:
```kotlin
            val bar = async<Int> { throw testException }
            try {
                bar.await()
            } finally {
                val foo = async {
                    try {
                        delay(Long.MAX_VALUE)
                        42
                    } finally {
                        withContext(NonCancellable) {
                            delay(200)
                        }
                    }
                }
                foo.invokeOnCompletion {
                    expect(3)
                }
            }
```
then `foo` never starts, but the test still passes.

chaoren:
If `foo` is cancelled before it's started, then `foo.invokeOnCompletion` is immediately triggered and `expect(3)` happens, which could happen before `expect(2)`. The only thing making sure that `expect(3)` appears after `expect(2)` is the `delay(200)` in `foo`.

chaoren:
I've noticed that the test is consistently flaky if you run it on a machine with very few cores.

dkhalanskyjb:
Oh, okay, got it, thanks. Could you make this dependency explicit, then, by adding a `CountDownLatch(1)`? `withContext(NonCancellable) { delay(200) }` would be replaced with awaiting the latch, and we'd count down the latch after `expect(2)`.

chaoren:
Done.

dkhalanskyjb (approved):
Well done, thanks!

PR #4166, opened 2024-07-03
Enable lazier initialization of Dispatchers.IO

metalhead8816:
Currently any reference to Dispatchers class including many calls to Dispatchers.Default, force Dispatchers.IO to be initialized. This incurs several system property reads and other expensive operations even if Dispatchers.IO is never used.

Swap to a get backed val and definition in new IoDispatcher file to allow separate initialization in a smaller class clinit.


dkhalanskyjb (changes requested):
I understand the `get()` part, but why did you move the object to a different file? I don't think objects are initialized until they are first accessed.
```kotlin
object X {
    init {
        println("X initialized")
    }
}

fun main() {
    println("ok")
}
```
prints `ok`, without `X initialized` (https://pl.kotl.in/5BCrlXilO).

metalhead8816:
You are correct. Looks like I cross some stack trace graphs between my attempts to fix this. Updated based on your comments.

dkhalanskyjb (approved):

PR #4171, opened 2024-07-08
Add missing `@OptIn`s for `InternalForInheritanceCoroutinesApi`.

chaoren:


dkhalanskyjb (approved):
Thanks!

Yeah, we really should re-enable the "warnings as errors" flag that got turned off when we migrated to Kotlin 2.0.

PR #4173, opened 2024-07-09
Fix stack overflow when calling asDeferred on a Future many times

### LINKED ISSUE #4156: Calling CompletionStage.asDeferred() on the same future many times causes stack overflows
**Describe the bug**

Given a shared `CompletableFuture` object `future`:

1. Attach other transformations on `future`, e.g. `val didRun = AtomicBoolean(); future.whenComplete { _, _ -> didRun.set(true) }`
2. Call `future.asDeferred()` many times. Generally, 10,000 times is sufficient to trigger the failure. The resulting `Deferred<*>` don't have to be used.
3. Complete `future` exceptionally, e.g. `future.completeExceptionally(RuntimeException())`.

Actual: Some of the `Deferred<*>` results would not complete. Some of the extra transformations in (1) do not run.

The exception thrown from the completed `Deferred<*>` objects *may* contain a suppressed `StackOverflowError`. In testing, in most of the cases there is no indication that anything failed, and the StackOverflowError can be observed by setting a breakpoint inside CompletableFuture's internal machinery.

Expected: all the `Deferred<*>` results should complete exceptionally, and additional transforms attached in (1) should run.

**Why would you do this?**

The above condition can happen when you use an asynchronous cache, e.g. Caffeine.newBuilder()...buildAsync().

In such use cases, an underlying `CompletableFuture`-based cache would return a cached (and shared) `CompletableFuture` object, and code that is using the cache would call `CompletionStage.asDeferred()` to consume the cached future asynchronously.

In exceptional conditions, e.g. when the backend computation gated behind the cache slows to a crawl and then starts failing, the above described condition can happen - a single `CompletableFuture` is returned to a bunch of cache consumers while everybody is waiting for the computation to complete. If the incoming request rate is high enough, you can easily end up with 10,000 waiters on the same `CompletableFuture`.

Then, when backend computation finally fails, this bug is triggered.

Note: this bug is particular insidious because it is very hard to trigger normally, and when it does, caches can behave very unexpectedly.

In the case of Caffeine, because some transforms fail to run (1 above), Caffeine does not get notified that the `CompletableFuture` fail and keeps returning the same failed `CompletableFuture` to future callers.

**Why not CompletionStage.await()**

Because the `CompletableFuture` is shared between multiple callers. Using `CompletionStage.await()` will allow one caller's cancellation to break all other concurrent callers waiting on the same future.

**Provide a Reproducer**

This is a `*.main.kts`. Run with `kotlin <something>.main.kts <num calls>`.

```kotlin
@file:Repository("https://repo1.maven.org/maven2")
@file:DependsOn("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.8.1")

import kotlinx.coroutines.future.asDeferred
import java.util.concurrent.CompletableFuture
import java.util.concurrent.atomic.AtomicBoolean

val numCalls = args[0].toInt()

val future = CompletableFuture<Unit>()

// Step 1
val didRun = AtomicBoolean(false)
future.whenComplete { _, _ -> didRun.set(true) }

// Step 2
val deferreds = (0 until numCalls).map { future.asDeferred() }

// Step 3
future.completeExceptionally(RuntimeException())

// There are no thread pool and executors involved above. All executions should complete on step 3.
runBlocking {
    // But just in case, wait a few seconds.
    delay(2.seconds)

    println("Expected additional transformation to run. Actual: ${didRun.get()}")
    println("Expected ${deferreds.size} Deferred<Unit> to complete exceptionally. Actual: ${deferreds.filter { it.isCompleted }.size}")
}
```

Versions:

* Mac OS X Sonoma 14.4.1, M3 Pro.
* java -version

```
openjdk version "21.0.3" 2024-04-16
OpenJDK Runtime Environment Homebrew (build 21.0.3)
OpenJDK 64-Bit Server VM Homebrew (build 21.0.3, mixed mode, sharing)
```

* kotlin -version: `Kotlin version 2.0.0-release-341 (JRE 21.0.3)`
* kotlinx-coroutines-core: 1.8.1 (above)

Results:

* `$ kotlin test.main.kts 100`: no bug

```
$ JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.3/libexec/openjdk.jdk/Contents/Home kotlin test.main.kts 100
Expected additional transformation to run. Actual: true
Expected 100 Deferred<Unit> to complete exceptionally. Actual: 100
```
* `$ kotlin test.main.kts 10000`: bug, with various failure modes

```
$ JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.3/libexec/openjdk.jdk/Contents/Home kotlin test.main.kts 10000
Expected additional transformation to run. Actual: true
Expected 10000 Deferred<Unit> to complete exceptionally. Actual: 8023
$ JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.3/libexec/openjdk.jdk/Contents/Home kotlin test.main.kts 10000
Expected additional transformation to run. Actual: true
Expected 10000 Deferred<Unit> to complete exceptionally. Actual: 2970
$ JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.3/libexec/openjdk.jdk/Contents/Home kotlin test.main.kts 10000
Expected additional transformation to run. Actual: true
Expected 10000 Deferred<Unit> to complete exceptionally. Actual: 9894
$ JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.3/libexec/openjdk.jdk/Contents/Home kotlin test.main.kts 10000
Expected additional transformation to run. Actual: true
Expected 10000 Deferred<Unit> to complete exceptionally. Actual: 9995
$ JAVA_HOME=/opt/homebrew/Cellar/openjdk@21/21.0.3/libexec/openjdk.jdk/Contents/Home kotlin test.main.kts 10000
Expected additional transformation to run. Actual: false
Expected 10000 Deferred<Unit> to complete exceptionally. Actual: 8075
```

dkhalanskyjb:
Fixes #4156

### DIFF
--- a/kotlinx-coroutines-core/jdk8/src/future/Future.kt
+++ b/kotlinx-coroutines-core/jdk8/src/future/Future.kt
@@ -138,7 +138,7 @@ public fun <T> CompletionStage<T>.asDeferred(): Deferred<T> {
             handleCoroutineException(EmptyCoroutineContext, e)
         }
     }
-    result.cancelFutureOnCompletion(future)
+    result.invokeOnCompletion(handler = CancelFutureOnCompletion(future))
     return result
 }
 
@@ -189,4 +189,19 @@ private class ContinuationHandler<T>(
             cont.resumeWithException((exception as? CompletionException)?.cause ?: exception)
         }
     }
+}
+
+private class CancelFutureOnCompletion(
+    private val future: Future<*>
+) : JobNode() {
+    override fun invoke(cause: Throwable?) {
+        // Don't interrupt when cancelling future on completion, because no one is going to reset this
+        // interruption flag and it will cause spurious failures elsewhere.
+        // We do not cancel the future if it's already completed in some way,
+        // because `cancel` on a completed future won't change the state but is not guaranteed to behave well
+        // on reentrancy. See https://github.com/Kotlin/kotlinx.coroutines/issues/4156
+        if (cause != null && !future.isDone) future.cancel(false)
+    }
+
+    override val onCancelling get() = false
 }
--- a/kotlinx-coroutines-core/jdk8/src/future/Future.kt
+++ b/kotlinx-coroutines-core/jdk8/src/future/Future.kt
@@ -194,6 +194,8 @@ private class ContinuationHandler<T>(
 private class CancelFutureOnCompletion(
     private val future: Future<*>
 ) : JobNode() {
+    override val onCancelling get() = false
+
     override fun invoke(cause: Throwable?) {
         // Don't interrupt when cancelling future on completion, because no one is going to reset this
         // interruption flag and it will cause spurious failures elsewhere.
@@ -202,6 +204,4 @@ private class CancelFutureOnCompletion(
         // on reentrancy. See https://github.com/Kotlin/kotlinx.coroutines/issues/4156
         if (cause != null && !future.isDone) future.cancel(false)
     }
-
-    override val onCancelling get() = false
 }
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -205,7 +205,7 @@ private class CancelFutureOnCancel(private val future: Future<*>) : CancelHandle
     override fun invoke(cause: Throwable?) {
         // Don't interrupt when cancelling future on completion, because no one is going to reset this
         // interruption flag and it will cause spurious failures elsewhere
-        if (cause != null) future.cancel(false)
+        future.cancel(false)
     }
     override fun toString() = "CancelFutureOnCancel[$future]"
 }
--- a/kotlinx-coroutines-core/jvm/src/Executors.kt
+++ b/kotlinx-coroutines-core/jvm/src/Executors.kt
@@ -1,6 +1,5 @@
 package kotlinx.coroutines
 
-import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.internal.*
 import java.io.Closeable
 import java.util.concurrent.*
@@ -145,7 +144,7 @@ internal class ExecutorCoroutineDispatcherImpl(override val executor: Executor) 
         )
         // If everything went fine and the scheduling attempt was not rejected -- use it
         if (future != null) {
-            continuation.cancelFutureOnCancellation(future)
+            continuation.invokeOnCancellation(CancelFutureOnCancel(future))
             return
         }
         // Otherwise fallback to default executor
@@ -200,4 +199,13 @@ private class DisposableFutureHandle(private val future: Future<*>) : Disposable
         future.cancel(false)
     }
     override fun toString(): String = "DisposableFutureHandle[$future]"
+}
+
+private class CancelFutureOnCancel(private val future: Future<*>) : CancelHandler {
+    override fun invoke(cause: Throwable?) {
+        // Don't interrupt when cancelling future on completion, because no one is going to reset this
+        // interruption flag and it will cause spurious failures elsewhere
+        if (cause != null) future.cancel(false)
+    }
+    override fun toString() = "CancelFutureOnCancel[$future]"
 }
--- a/kotlinx-coroutines-core/jvm/src/Future.kt
+++ b/kotlinx-coroutines-core/jvm/src/Future.kt
@@ -12,10 +12,15 @@ import java.util.concurrent.*
  * invokeOnCancellation { if (it != null) future.cancel(false) }
  * ```
  */
+@Deprecated(
+    "This function does not do what its name implies: it will not cancel the future if just cancel() was called.",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("this.invokeOnCancellation { future.cancel(false) }")
+)
 public fun CancellableContinuation<*>.cancelFutureOnCancellation(future: Future<*>): Unit =
-    invokeOnCancellation(handler = CancelFutureOnCancel(future))
+    invokeOnCancellation(handler = PublicCancelFutureOnCancel(future))
 
-private class CancelFutureOnCancel(private val future: Future<*>) : CancelHandler {
+private class PublicCancelFutureOnCancel(private val future: Future<*>) : CancelHandler {
     override fun invoke(cause: Throwable?) {
         // Don't interrupt when cancelling future on completion, because no one is going to reset this
         // interruption flag and it will cause spurious failures elsewhere
--- a/kotlinx-coroutines-core/jvm/src/Future.kt
+++ b/kotlinx-coroutines-core/jvm/src/Future.kt
@@ -9,36 +9,11 @@ import java.util.concurrent.*
  * Cancels a specified [future] when this job is cancelled.
  * This is a shortcut for the following code with slightly more efficient implementation (one fewer object created).
  * ```
- * invokeOnCompletion { if (it != null) future.cancel(false) }
- * ```
- *
- * @suppress **This an internal API and should not be used from general code.**
- */
-@InternalCoroutinesApi
-public fun Job.cancelFutureOnCompletion(future: Future<*>): DisposableHandle =
-    invokeOnCompletion(handler = CancelFutureOnCompletion(future))
-
-/**
- * Cancels a specified [future] when this job is cancelled.
- * This is a shortcut for the following code with slightly more efficient implementation (one fewer object created).
- * ```
  * invokeOnCancellation { if (it != null) future.cancel(false) }
  * ```
  */
 public fun CancellableContinuation<*>.cancelFutureOnCancellation(future: Future<*>): Unit =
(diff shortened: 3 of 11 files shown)


kotlinx-coroutines-core/jdk8/src/future/Future.kt
```diff
@@ -190,3 +190,18 @@ private class ContinuationHandler<T>(
...
+private class CancelFutureOnCompletion(
+    private val future: Future<*>
+) : JobNode() {
+    override fun invoke(cause: Throwable?) {
+        // Don't interrupt when cancelling future on completion, because no one is going to reset this
+        // interruption flag and it will cause spurious failures elsewhere.
+        // We do not cancel the future if it's already completed in some way,
+        // because `cancel` on a completed future won't change the state but is not guaranteed to behave well
+        // on reentrancy. See https://github.com/Kotlin/kotlinx.coroutines/issues/4156
+        if (cause != null && !future.isDone) future.cancel(false)
```
zuevmaxim, line 203: > but is not guaranteed to behave well on reentrancy
Can you please mention what exactly goes wrong? Something like `Reentrant cancellation of a future may cause repeated cancellation of it's dependants`

dkhalanskyjb: I don't think the documentation says anything specific, unfortunately, so what exactly goes wrong seems to be an implementation detail. In the issue linked, the problem was that `future.cancel` shared the job of calling the remaining `handle` callbacks, which called `future.cancel`, which shared the job, and so on recursively in the same thread.

kotlinx-coroutines-core/jvm/src/Future.kt
```diff
@@ -5,42 +5,22 @@ package kotlinx.coroutines
...
-        // interruption flag and it will cause spurious failures elsewhere
-        if (cause != null) future.cancel(false)
-    }
-
-    override val onCancelling get() = false
-}
+    invokeOnCancellation(handler = PublicCancelFutureOnCancel(future))
 
-private class CancelFutureOnCancel(private val future: Future<*>) : CancelHandler {
+private class PublicCancelFutureOnCancel(private val future: Future<*>) : CancelHandler {
```
zuevmaxim, line 23: The name is unclear to me. Should it be named `CancelFutureOnCancelWithCause`? Should we update `toString` by the way?

kotlinx-coroutines-core/jvm/src/Future.kt
```diff
@@ -5,42 +5,22 @@ package kotlinx.coroutines
...
  * ```
  * invokeOnCancellation { if (it != null) future.cancel(false) }
  * ```
  */
+@Deprecated(
+    "This function does not do what its name implies: it will not cancel the future if just cancel() was called.",
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("this.invokeOnCancellation { future.cancel(false) }")
+)
 public fun CancellableContinuation<*>.cancelFutureOnCancellation(future: Future<*>): Unit =
```
zuevmaxim, line 20: There are no tests for this method. Should we introduce one? I will become clearer which cases this method does not cover

kotlinx-coroutines-core/jvm/src/Future.kt
```diff
@@ -5,42 +5,22 @@ package kotlinx.coroutines
...
-
 /**
  * Cancels a specified [future] when this job is cancelled.
  * This is a shortcut for the following code with slightly more efficient implementation (one fewer object created).
  * ```
  * invokeOnCancellation { if (it != null) future.cancel(false) }
  * ```
  */
+@Deprecated(
+    "This function does not do what its name implies: it will not cancel the future if just cancel() was called.",
```
zuevmaxim, line 16: > it will not cancel the future if just cancel() was called

Can you please specify the difference between the case when the method does work and does not work? For example, `The future will be cancelled only when the job is cancelled, but not the Continuation`

kotlinx-coroutines-core/jvm/src/Future.kt
dkhalanskyjb: I would call it `CancelFutureOnCancel`, but then, the class names would be in conflict, so I had to rename it.

kotlinx-coroutines-core/jvm/src/Executors.kt
```diff
@@ -201,3 +200,12 @@ private class DisposableFutureHandle(private val future: Future<*>) : Disposable
     }
     override fun toString(): String = "DisposableFutureHandle[$future]"
 }
+
+private class CancelFutureOnCancel(private val future: Future<*>) : CancelHandler {
+    override fun invoke(cause: Throwable?) {
+        // Don't interrupt when cancelling future on completion, because no one is going to reset this
+        // interruption flag and it will cause spurious failures elsewhere
+        future.cancel(false)
```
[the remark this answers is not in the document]
dkhalanskyjb: I'm pretty sure we can't, but could you double-check it? Ensuring that the code is correct is typically shared equally between the author and the reviewer in this library, we tend to dive deep.

kotlinx-coroutines-core/jvm/src/Future.kt
dkhalanskyjb, line 16: @zuevmaxim,
```suggestion
    "This function does not do what its name implies: it will not cancel the future if the continuation was cancelled without a cause specified (e.g., via `cancel()` with no argument).",
```
?

kotlinx-coroutines-core/jvm/src/Future.kt
dkhalanskyjb: We could add some tests, but I don't think they would realistically help us catch regressions: the function is simple, and we're unlikely to ever change it again now that it's deprecated.

kotlinx-coroutines-core/jvm/src/Future.kt
zuevmaxim: Looks good

kotlinx-coroutines-core/jvm/src/Future.kt
dkhalanskyjb: Sure, `Deprecated` sounds good.

kotlinx-coroutines-core/jvm/src/Executors.kt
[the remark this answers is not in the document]
zuevmaxim: I see no callbacks to be cancelled by the `future`, so I cannot come up with an example when there might be a problem

zuevmaxim (approved):

kotlinx-coroutines-core/jdk8/src/future/Future.kt
```diff
@@ -190,3 +190,18 @@ private class ContinuationHandler<T>(
...
+    }
+
+    override val onCancelling get() = false
+}
```
qwwdfsad, line 207: AFAIR we consistently leave a trailing newline and there is an IDEA setting to do so

kotlinx-coroutines-core/jdk8/src/future/Future.kt
```diff
@@ -190,3 +190,18 @@ private class ContinuationHandler<T>(
...
+    override fun invoke(cause: Throwable?) {
+        // Don't interrupt when cancelling future on completion, because no one is going to reset this
+        // interruption flag and it will cause spurious failures elsewhere.
+        // We do not cancel the future if it's already completed in some way,
+        // because `cancel` on a completed future won't change the state but is not guaranteed to behave well
+        // on reentrancy. See https://github.com/Kotlin/kotlinx.coroutines/issues/4156
+        if (cause != null && !future.isDone) future.cancel(false)
+    }
+
+    override val onCancelling get() = false
```
qwwdfsad, line 206: Code-style: properties before functions (though it seems we missed this during the original review)

kotlinx-coroutines-core/jvm/src/Future.kt
```diff
@@ -5,46 +5,26 @@ package kotlinx.coroutines
...
-    invokeOnCompletion(handler = CancelFutureOnCompletion(future))
-
 /**
  * Cancels a specified [future] when this job is cancelled.
  * This is a shortcut for the following code with slightly more efficient implementation (one fewer object created).
  * ```
  * invokeOnCancellation { if (it != null) future.cancel(false) }
  * ```
  */
+@Deprecated(
```
qwwdfsad, line 15: Please drop a comment since what version it's deprecated as we do in other places

qwwdfsad (approved):

kotlinx-coroutines-core/jdk8/src/future/Future.kt
dkhalanskyjb: The point of this rule is unclear to me, but ok, changed this throughout the codebase.

kotlinx-coroutines-core/jvm/src/Future.kt
lukellmann: It seems like this change wasn't actually applied?

PR #4188, opened 2024-07-22
Correct the link for coroutines-guide.html

lamtrinhdev:
Correct the link for coroutines-guide.html


qwwdfsad:
Thanks! It is, in fact, obsolete documentation. Superceded with #4189

PR #4189, opened 2024-07-22
Remove obsolete npm folder from the doc

qwwdfsad:


dkhalanskyjb:
This does look obsolete: the last published version of https://www.npmjs.com/package/kotlinx-coroutines-core is 1.7.3. But why did it stop being published? Was this a conscious decision? Our changelog doesn't say anything about this, it seems. We also reworked our Gradle scripts entirely between 1.7.3 and 1.8.0, so this could just be a mistake on our part (which no one noticed?..) What is going on?

qwwdfsad:
Starting from Kotlin 1.9 (where JS IR became stable), there is no longer npm publication all across the ecosystem -- we no longer produce JS sources, only klibs

dkhalanskyjb (approved):

PR #4195, opened 2024-07-28
Fix documentation of runBlocking

Eaterminer:
kotlinx-coroutines-core/jvm/src/Builders.kt documentation looked rather strange and had incorrect grammar, replaced with kotlinx-coroutines-core/native/src/Builders.kt documentation for runBlocking.

This should be correct. If I have made a mistake, I apologize in advance.


dkhalanskyjb:
Yes, this is correct, thanks!

PR #4200, opened 2024-07-30
Fix too many `the` and `it` on some comments

ParkJong-Hun:
I think that the double use of `the` and `it` in this part in English is grammatically incorrect, except for the intentional emphasis.  
And, it doesn't seems to be something that needed to be emphasized.
So, I didn't think it was an intentional mistake, so I fixed it.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/Migration.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Migration.kt
@@ -90,7 +90,7 @@ public fun <T> Flow<T>.publishOn(context: CoroutineContext): Flow<T> = noImpl()
  *        }
  * }
  * ```
- * Opposed to subscribeOn, it it **possible** to use multiple `flowOn` operators in the one flow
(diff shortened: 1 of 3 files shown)


kotlinx-coroutines-core/common/src/flow/Migration.kt
```diff
@@ -90,7 +90,7 @@ public fun <T> Flow<T>.publishOn(context: CoroutineContext): Flow<T> = noImpl()
  *        }
  * }
  * ```
- * Opposed to subscribeOn, it it **possible** to use multiple `flowOn` operators in the one flow
+ * Opposed to subscribeOn, it **possible** to use multiple `flowOn` operators in the one flow
```
dkhalanskyjb, line 93: Probably what was meant is, "it is."

ParkJong-Hun: Thanks! you're right

PR #4203, opened 2024-08-01
Remove redundant `html_version` gradle property

turansky:
Looks like property for [kotlinx.html](https://github.com/Kotlin/kotlinx.html), which isn't used in project


qwwdfsad:
Thanks!

PR #4207, opened 2024-08-08
Rerun apiDump

zuevmaxim:
Refresh the API dump without code changes. Fix the order of items in the dump after updating to a newer version of the compatibility-validator. Version updates without dump refreshing were possible due to https://github.com/Kotlin/binary-compatibility-validator/issues/260


dkhalanskyjb:
When you merge it, please include some context in the commit message: why do we have to do that? How did it come to this? Otherwise, seeing just "Rerun apiDump" in `git blame` will not be very helpful. https://github.com/Kotlin/binary-compatibility-validator/issues/260 may be worth linking to.

dkhalanskyjb (approved):

PR #4223, opened 2024-09-09
WA: Disable JS legacy transformation manually until it is completely removed from the atomicfu plugin.

mvicsokolova:
Fixes KT-71203

Through an oversight `atomicfu-gradle-plugin` kept tranformJs option (corresponding to the legacy JS transformation) enabled by default, and with no *.js files to transform it just copied the compileJs output directory, which led to an error described in [KT-71203](https://youtrack.jetbrains.com/issue/KT-71203).

I suggest to manually disable JS transformation in all subprojects, which apply atomicfu plugin (I couldn't do it more gracefully in the root build script 😔 ).

 This fix can be removed as soon as JS transformation is removed completely in atomicfu library. I plan to prepare a corresponding commit and provide it with the next kotlinx-atomicfu release.

### DIFF
--- a/buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts
@@ -1,19 +1,9 @@
 import org.jetbrains.kotlin.gradle.tasks.*
-import kotlin.reflect.*
-import kotlin.reflect.full.*
 
 configure(subprojects) {
     val project = this
     if (name in sourceless) return@configure
-    apply(plugin = "org.jetbrains.kotlinx.atomicfu")
-    // Workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
-    afterEvaluate {
-        extensions.findByName("atomicfu")?.let { atomicfuExtension ->
-            atomicfuExtension::class.memberProperties.firstOrNull { it is KMutableProperty<*> && it.name == "transformJs" }?.let {
-                (it as KMutableProperty<*>).setter.call(atomicfuExtension, false)
-            }
-        }
-    }
+    apply(plugin = "atomicfu-conventions")
     tasks.withType<KotlinCompilationTask<*>>().configureEach {
         val isMainTaskName = name.startsWith("compileKotlin")
         compilerOptions {
--- a/buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts
@@ -1,9 +1,19 @@
 import org.jetbrains.kotlin.gradle.tasks.*
+import kotlin.reflect.*
+import kotlin.reflect.full.*
 
 configure(subprojects) {
     val project = this
     if (name in sourceless) return@configure
     apply(plugin = "org.jetbrains.kotlinx.atomicfu")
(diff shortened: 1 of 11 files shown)


reactive/kotlinx-coroutines-reactive/build.gradle.kts
```diff
@@ -49,3 +49,7 @@ kover {
         }
     }
 }
+
+atomicfu {
+    transformJs = false
+}
```
dkhalanskyjb, line 55: But `kotlinx-coroutines-reactive` doesn't build any JS artifacts. Of the subprojects listed, only `-core`, `-test`, and `test-utils` do. Is this really needed?

kotlinx-coroutines-core/build.gradle.kts
```diff
@@ -306,3 +306,7 @@ artifacts {
 tasks.named("dokkaHtmlPartial") {
     dependsOn(jvmJar)
 }
+
+atomicfu {
+    transformJs = false
+}
```
dkhalanskyjb, line 312: Can't this be put into `buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts`? The `atomicfu` plugin is applied there, and it would be nice to keep the related logic close together.

kotlinx-coroutines-test/build.gradle.kts
```diff
@@ -22,3 +22,7 @@ kotlin {
         }
     }
 }
+
+atomicfu {
```
dkhalanskyjb, line 26:
```suggestion
// workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
atomicfu {
```
?

mvicsokolova: Added this comment 🙏

kotlinx-coroutines-core/build.gradle.kts
mvicsokolova: I applied this config in `configure-compilation-conventions` in the last commit)

buildSrc/src/main/kotlin/configure-compilation-conventions.gradle.kts
```diff
@@ -1,9 +1,19 @@
 import org.jetbrains.kotlin.gradle.tasks.*
+import kotlin.reflect.*
+import kotlin.reflect.full.*
 
 configure(subprojects) {
     val project = this
     if (name in sourceless) return@configure
     apply(plugin = "org.jetbrains.kotlinx.atomicfu")
+    // Workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
+    afterEvaluate {
```
mvicsokolova, line 10: The problem is that `AtomicFUPluginExtnesion` type is not accessible in the root build script yet, and seems that the only way to dynamically set the property in this extension class is via reflection 🥲

dkhalanskyjb: How about something like this? https://github.com/Kotlin/kotlinx.coroutines/commit/60c070771a7fdcd2b364fcc19f4b272f2ce9211f

mvicsokolova: Cool! Thanks, that's much better and solves the problem)

dkhalanskyjb (approved):

PR #4236, opened 2024-09-26
Add API reference badge to README.md

qwwdfsad:
A bit easier and more convenient to find quickly


dkhalanskyjb (approved):
When we added the badge to `kotlinx-datetime`, there were questions about consistent badge order across libraries: https://github.com/Kotlin/kotlinx-datetime/pull/256#discussion_r1136094023.

https://github.com/Kotlin/kotlinx-datetime and https://github.com/Kotlin/kotlinx.serialization are now inconsistent with one another for some reason, but both suggest that the API reference goes above Slack but after both Kotlin and Maven Central.

I myself don't have a preference for any particular order of badges (and am not bothered by the variance across different libraries: our README.md is already fully free-form, it seems), so if you think the last line is the best place to put the badge, I don't mind.

qwwdfsad:
Thanks for pointing it out. I'm completely OK with any reasonable order -- if anyone feels we benefit from consistent order across the board, then why not.

Put it before Slack based on a subjective feeling it's slightly better :)

PR #4238, opened 2024-09-27
Replace a usage of deprecated getTimeNanos() in K/N

qurbonzoda:
The function was deprecated with warning in Kotlin 1.9. In Kotlin 2.1 the deprecation level will be raised to error.

See KT-71628

### DIFF
--- a/kotlinx-coroutines-core/native/src/EventLoop.kt
+++ b/kotlinx-coroutines-core/native/src/EventLoop.kt
@@ -4,7 +4,7 @@ package kotlinx.coroutines
 
 import kotlin.coroutines.*
 import kotlin.native.concurrent.*
-import kotlin.system.*
+import kotlin.time.*
 
 internal actual abstract class EventLoopImplPlatform : EventLoop() {
 
@@ -26,5 +26,6 @@ internal class EventLoopImpl: EventLoopImplBase() {
 
 internal actual fun createEventLoop(): EventLoop = EventLoopImpl()
 
-@Suppress("DEPRECATION")
-internal actual fun nanoTime(): Long = getTimeNanos()
+private val startingPoint = TimeSource.Monotonic.markNow()
+
+internal actual fun nanoTime(): Long = (TimeSource.Monotonic.markNow() - startingPoint).inWholeNanoseconds


kotlinx-coroutines-core/native/src/EventLoop.kt
```diff
@@ -26,5 +26,5 @@ internal class EventLoopImpl: EventLoopImplBase() {
 
 internal actual fun createEventLoop(): EventLoop = EventLoopImpl()
 
-@Suppress("DEPRECATION")
+@Suppress("DEPRECATION_ERROR")
```
qurbonzoda, line 29: The change can be merged to develop if we suppress both `DEPRECATION` and `DEPRECATION_ERROR`.
But, it would seem weird.

woainikk (approved):

kotlinx-coroutines-core/native/src/EventLoop.kt
qurbonzoda: This also underscores the need to refactor the usages of `nanoTime()` to use the Duration API instead. Having the kotlinx-datetime `Clock` and `Instant` in stdlib would simplify this migration.

qurbonzoda: > Can we replace it with the proposed replacement instead?

There is not direct replacement for the `getTime` currently.

qurbonzoda: It might make sense to leave `getTimeX` deprecated with warning until a replacement is occurs, though.

JakeWharton: Neither `Clock` nor `Instant` would be monotonic, though.

Maybe make the actualization on more granular groups of targets (linux, macos, mingw, androidNative) and invoke OS-level APIs to get the monotonic nano time. They're all just one- or two-liners.

qurbonzoda: I have replaced the K/N `getTimeNanos()` usage with `TimeSource.Monotonic.markNow()`.
Subsequently, now `ComparableTimeMark` is used to designate a point of time instead of `Long`.
PTAL.

kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -428,53 +430,53 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
...
-                     * Carefully update [DelayedTaskQueue.timeNow] so that it does not sweep past first's tasks time
+                     * Carefully update [DelayedTaskQueue.timeMark] so that it does not sweep past first's tasks time
                      * and only goes forward in time. We cannot let it go backwards in time or invariant can be
                      * violated for tasks that were already scheduled.
                      */
-                    val firstTime = firstTask.nanoTime
+                    val firstTime = firstTask.timeMark
                     // compute min(now, firstTime) using a wrap-safe check
-                    val minTime = if (firstTime - now >= 0) now else firstTime
+                    val minTime = if (firstTime - now >= Duration.ZERO) now else firstTime
```
qurbonzoda, line 468: I am not sure is such wrap-safe checks are still relevant.
Is it safe to transform such expressions to `firstTime >= now` or `minOf(firstTime, now)`?

kotlinx-coroutines-core/jvm/test/VirtualTimeSource.kt
```diff
@@ -20,40 +24,38 @@ internal inline fun withVirtualTimeSource(log: PrintStream? = null, block: () ->
...
-
 private class ThreadStatus {
     @Volatile @JvmField
-    var parkedTill = NOT_PARKED
+    var parkedTill: ComparableTimeMark? = null
     @Volatile @JvmField
     var permit = false
     var registered = 0
-    override fun toString(): String = "parkedTill = ${TimeUnit.NANOSECONDS.toMillis(parkedTill)} ms, permit = $permit"
+    override fun toString(): String = "parkedTill = $parkedTill, permit = $permit"
```
qurbonzoda, line 33: Could you please tell how important is it to present the `parkedTill` value in milliseconds ?
It will require storing the `VirtualTimeSource.zero` in `ThreadStatus`.

kotlinx-coroutines-core/native/src/EventLoop.kt
```diff
@@ -14,8 +14,8 @@ internal actual abstract class EventLoopImplPlatform : EventLoop() {
         current.executeAfter(0L, {})// send an empty task to unpark the waiting event loop
     }
 
-    protected actual fun reschedule(now: Long, delayedTask: EventLoopImplBase.DelayedTask) {
-        DefaultExecutor.invokeOnTimeout(now, delayedTask, EmptyCoroutineContext)
+    protected actual fun reschedule(now: ComparableTimeMark, delayedTask: EventLoopImplBase.DelayedTask) {
+        DefaultExecutor.invokeOnTimeout((now - markNow()).inWholeMilliseconds, delayedTask, EmptyCoroutineContext)
```
qurbonzoda, line 18: Could you please confirm that a "point of time" value was wrongly passed to a function that expects a "duration of time"? Additionally, the former was in nanos, while the later in milllis.

kotlinx-coroutines-core/common/src/EventLoop.common.kt
```diff
@@ -487,48 +489,48 @@ internal abstract class EventLoopImplBase: EventLoopImplPlatform(), Delay {
...
     internal class DelayedTaskQueue(
-        @JvmField var timeNow: Long
+        @JvmField var timeMark: ComparableTimeMark
     ) : ThreadSafeHeap<DelayedTask>()
 }
 
 internal expect fun createEventLoop(): EventLoop
 
-internal expect fun nanoTime(): Long
+internal expect fun markNow(): ComparableTimeMark
```
whyoleg, line 533: just curious: is it possible to use [ValueTimeMark](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-time-source/-monotonic/-value-time-mark/) which is returned by [TimeSource.Monotonic#markNow](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-time-source/-monotonic/) to avoid boxing? And does it make sense?

qurbonzoda: The issue is that only `TimeSource.Monotonic` returns `ValueTimeMark`. For testing purposes, the project uses `TestTimeSource` to generate predictable time marks, which are not `ValueTimeMark`.

kotlinx-coroutines-core/jvm/test/guide/example-cancel-02.kt
```diff
@@ -2,17 +2,18 @@
 package kotlinx.coroutines.guide.exampleCancel02
 
 import kotlinx.coroutines.*
+import kotlin.time.Duration.Companion.milliseconds
 
 fun main() = runBlocking {
-    val startTime = currentTimeMillis()
+    val startTime = markNow()
```
qurbonzoda, line 8: Are these tests supposed to be executed with an adjustable time source?
Maybe we should just use `TimeSource.Monotonic` instead?

qurbonzoda: Hmm, it seems to be generated by Knit, which replaces `System.currentTimeMillis()` with `currentTimeMillis()`: https://github.com/Kotlin/kotlinx-knit/blob/d461cde5558a8f9a9338f80fda29607fd3779b3b/src/Knit.kt#L205

docs/topics/cancellation-and-timeouts.md
```diff
@@ -149,18 +150,19 @@ Replace `while (i < 5)` in the previous example with `while (isActive)` and reru
 
 ```kotlin
 import kotlinx.coroutines.*
 
 fun main() = runBlocking {
 //sampleStart
-    val startTime = System.currentTimeMillis()
+    val startTime = markNow()
```
dkhalanskyjb, line 157: How will this compile? Is `markNow` in the scope? Also, `knitCheck` fails, so `./gradlew knit` should be run.

kotlinx-coroutines-core/native/src/EventLoop.kt
dkhalanskyjb: Yes, thanks for noticing!

kotlinx-coroutines-core/jvm/src/AbstractTimeSource.kt
```diff
@@ -28,12 +28,8 @@ internal inline fun mockTimeSource(source: AbstractTimeSource?) {
 }
 
 @InlineOnly
-internal inline fun currentTimeMillis(): Long =
-    timeSource?.currentTimeMillis() ?: System.currentTimeMillis()
-
-@InlineOnly
-internal actual inline fun nanoTime(): Long =
-    timeSource?.nanoTime() ?: System.nanoTime()
```
dkhalanskyjb, line 36: This approach seems too invasive to me, and together with the added boxing throughout our implementation, it looks like it can lead to some observable performance degradation.

Here's an alternative proposal: `internal fun nanoTime(): Long = (TimeSource.Monotonic.markNow() - startingPoint).inWholeNanoseconds; private val startingPoint: ValueTimeMark = TimeSource.Monotonic.markNow()`. This way, `startingPoint` will get initialized once per run, and the rest of the program can be kept the way it is.

qurbonzoda: Perfect. Please take a look.

dkhalanskyjb (approved):

kotlinx-coroutines-core/native/src/EventLoop.kt
qurbonzoda: I will fix this in a separate PR

docs/topics/cancellation-and-timeouts.md
qurbonzoda: Reverted

kotlinx-coroutines-core/jvm/test/VirtualTimeSource.kt
qurbonzoda: Reverted

kotlinx-coroutines-core/native/src/EventLoop.kt
```diff
@@ -26,5 +26,6 @@ internal class EventLoopImpl: EventLoopImplBase() {
 
 internal actual fun createEventLoop(): EventLoop = EventLoopImpl()
 
-@Suppress("DEPRECATION")
-internal actual fun nanoTime(): Long = getTimeNanos()
+private val startingPoint = TimeSource.Monotonic.markNow()
+
+internal actual fun nanoTime(): Long = (TimeSource.Monotonic.markNow() - startingPoint).inWholeNanoseconds
```
ilya-g, line 31: Tip: You could use `startingPoint.elapsedNow` instead of subtracting marks

Issue #4124, opened 2024-05-08
`kotlinx.coroutines.debug`'s `module-info.java` is incorrect preventing any project using JPMS to use debug probes


ArchangelX360:
Any usage of the `kotlinx.coroutines.debug` package, for example `import kotlinx.coroutines.debug.DebugProbes` results in:
```
[JAVA_MODULE_DOES_NOT_EXPORT_PACKAGE] Symbol is declared in module 'kotlinx.coroutines.debug' which does not export package 'kotlinx.coroutines.debug'
 ```
The split package between `kotlinx.coroutines.core` and `kotlinx.coroutines.debug` of package `kotlinx.coroutines.debug` must be solved and cannot be hacked around by commenting its `exports` statement in `kotlinx.coroutines.debug`'s `module-info.java` 😢.

---

To continue the investigation, using custom module loading code, we merged `kotlin.stdlib`, `kotlinx.coroutines.core` and `kotlinx.coroutines.debug` modules at runtime:
- `kotlinx.coroutines.core` and `kotlinx.coroutines.debug` to mitigate the package split
- `kotlin.stdlib` to mitigate the circular dependency between `kotlinx.coroutines.debug` and `kotlin.stdlib` after bytebuddy injection at runtime
  - `kotlin.stdlib` ---  (due to injected code) --> `kotlinx.coroutines.debug` -> `kotlinx.stdlib`

That provided us with a few more information about what is wrong with the current `module-info.java`:

1. some statements are missing (probably coming from the shadowing of bytebuddy):
```java
requires com.sun.jna;
exports kotlinx.coroutines.repackaged.net.bytebuddy.agent to com.sun.jna;
```

2. these redundant (bytebuddy being shadowed in the same module) `requires` crashes at runtime:
```
requires net.bytebuddy.agent;
requires net.bytebuddy;
```

3. when used in production, the current declaration and structure of `kotlinx-coroutines-debug` jar forces users to bring in jars to their runtime module path to fulfill:
```java
requires org.junit.jupiter.api;
requires org.junit.platform.commons;
```
JUnit is not something one would want in production runtime module path 😅...

---

I did not find any suitable workaround to unblock ourselves. The only solution I see is to:
[...]

PR #4247, opened 2024-10-10
Fix 'kotlinx.coroutines.debug' & 'kotlinx.coroutines.debug.internal' split package problem

sellmair:
https://youtrack.jetbrains.com/issue/IDEA-359879/coroutines-debugging-Investigate-split-package-problem
https://github.com/Kotlin/kotlinx.coroutines/issues/4124

### DIFF
--- a/integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
+++ b/integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
@@ -1,0 +1,54 @@
+@file:OptIn(ExperimentalCoroutinesApi::class)
+
+import org.junit.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.debug.*
+import org.junit.Ignore
+import org.junit.Test
+import java.io.*
+import java.lang.IllegalStateException
+import kotlin.test.*
+
+class DynamicAttachDebugJpmsTest {
+
+    /**
+     * This test is disabled because:
+     * Dynamic Attach with JPMS is not yet supported.
+     *
+     * Here is the state of experiments:
+     * When launching this test with additional workarounds like
+     * ```
+     *     jvmArgs("--add-exports=kotlinx.coroutines.debug/kotlinx.coroutines.repackaged.net.bytebuddy=com.sun.jna")
+     *     jvmArgs("--add-exports=kotlinx.coroutines.debug/kotlinx.coroutines.repackaged.net.bytebuddy.agent=com.sun.jna")
+     *```
+     *
+     * Then we see issues like
+     *
+     * ```
+     * Caused by: java.lang.IllegalStateException: The Byte Buddy agent is not loaded or this method is not called via the system class loader
+     * 	at kotlinx.coroutines.debug/kotlinx.coroutines.repackaged.net.bytebuddy.agent.Installer.getInstrumentation(Installer.java:61)
+     * 	... 54 more
+     * 	```
+     */
+    @Ignore("shaded byte-buddy does not work with JPMS")
+    @Test
+    fun testAgentDumpsCoroutines() =
+        DebugProbes.withDebugProbes {
+            runBlocking {
+                val baos = ByteArrayOutputStream()
+                DebugProbes.dumpCoroutines(PrintStream(baos))
+                // if the agent works, then dumps should contain something,
+                // at least the fact that this test is running.
+                Assert.assertTrue(baos.toString().contains("testAgentDumpsCoroutines"))
+            }
+        }
+
+    @Test
+    fun testAgentIsNotInstalled() {
+        assertEquals(false, DebugProbes.isInstalled)
+        assertFailsWith<IllegalStateException> {
+            DebugProbes.dumpCoroutines(PrintStream(ByteArrayOutputStream()))
+        }
+    }
+
+}
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
@@ -7,10 +7,12 @@ package kotlinx.coroutines.debug.internal
  * Otherwise, access to `AgentPremain.isInstalledStatically` triggers the load of its internal `ClassFileTransformer`
  * that is not available on Android.
  *
- * Usage Note: Fleet (Reflection): FleetDebugProbes
- * Usage Note: Android (Hard Coded, ignored for Leak Detection)
- * Usage Note: IntelliJ (Suppress KotlinInternalInJava): CoroutineDumpState
+ * The entity (despite being internal) has usages in the following products
+ * - Fleet (Reflection): FleetDebugProbes
+ * - Android (Hard Coded, ignored for Leak Detection)
+ * - IntelliJ (Suppress KotlinInternalInJava): CoroutineDumpState
  */
+@PublishedApi
 internal object AgentInstallationType {
     internal var isInstalledStatically = false
 }
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
@@ -3,9 +3,13 @@ package kotlinx.coroutines.debug.internal
 /**
  * Object used to differentiate between agent installed statically or dynamically.
  * This is done in a separate object so [DebugProbesImpl] can check for static installation
- * without having to depend on [kotlinx.coroutines.debug.AgentPremain], which is not compatible with Android.
+ * without having to depend on [AgentPremain], which is not compatible with Android.
  * Otherwise, access to `AgentPremain.isInstalledStatically` triggers the load of its internal `ClassFileTransformer`
  * that is not available on Android.
+ *
+ * Usage Note: Fleet (Reflection): FleetDebugProbes
+ * Usage Note: Android (Hard Coded, ignored for Leak Detection)
+ * Usage Note: IntelliJ (Suppress KotlinInternalInJava): CoroutineDumpState
  */
 internal object AgentInstallationType {
     internal var isInstalledStatically = false
--- a/integration-testing/.gitignore
+++ b/integration-testing/.gitignore
@@ -1,0 +1,2 @@
+.kotlin
+kotlin-js-store
--- a/integration-testing/README.md
+++ b/integration-testing/README.md
@@ -8,6 +8,7 @@ The tests are the following:
 * `coreAgentTest` checks that `kotlinx-coroutines-core` can be run as a Java agent.
 * `debugAgentTest` checks that the coroutine debugger can be run as a Java agent.
 * `debugDynamicAgentTest` checks that `kotlinx-coroutines-debug` agent can self-attach dynamically to JVM as a standalone dependency.
+* `debugDynamicAgentJpmsTest` checks that `kotlinx-coroutines-debug` agent can self-attach dynamically to JVM as a standalone dependency  (with JPMS)
 * `smokeTest` builds the multiplatform test project that depends on coroutines.
 
 The `integration-testing` project is expected to be in a subdirectory of the main `kotlinx.coroutines` project.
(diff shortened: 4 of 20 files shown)


dkhalanskyjb:
I've looked through https://grep.app/search?q=kotlinx.coroutines.debug.AgentPremain, searching for places where people maybe rely on the current fully qualified name programmatically, and this is the only thing I've found: https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/b6916aa28808b31d5cda83da5da4e381595bd0b6/src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/InitializeIntelliJPlatformPluginTask.kt#L148 . This was added in the commit https://github.com/JetBrains/intellij-platform-gradle-plugin/commit/a759cecfe7e68aa0360e25586cb2e2ed2f517b16 . Looks like this change could break coroutine debugging for IDEA plugin authors? Could you investigate?

integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.IllegalStateException
+import kotlin.test.*
+
+class DynamicAttachDebugJpmsTest {
+
+    /**
+     * Using:
+     *
+     *     jvmArgs("--add-exports=kotlinx.coroutines.debug/kotlinx.coroutines.repackaged.net.bytebuddy=com.sun.jna")
+     *     jvmArgs("--add-exports=kotlinx.coroutines.debug/kotlinx.coroutines.repackaged.net.bytebuddy.agent=com.sun.jna")
```
dkhalanskyjb, line 18: What does this mean?

integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+     *
+     *
+     * Caused by: java.lang.IllegalStateException: The Byte Buddy agent is not loaded or this method is not called via the system class loader
+     * 	at kotlinx.coroutines.debug/kotlinx.coroutines.repackaged.net.bytebuddy.agent.Installer.getInstrumentation(Installer.java:61)
+     * 	... 54 more
```
dkhalanskyjb, line 23: And this?

dkhalanskyjb (commented):
The idea behind the changes is clear and seems sound. The only questions I have are for comments, the build scripts (but I'm still reading those), and the possibility of this causing breakage (which may be fine if we communicate this properly).

sellmair:
> I've looked through https://grep.app/search?q=kotlinx.coroutines.debug.AgentPremain, searching for places where people maybe rely on the current fully qualified name programmatically, and this is the only thing I've found: https://github.com/JetBrains/intellij-platform-gradle-plugin/blob/b6916aa28808b31d5cda83da5da4e381595bd0b6/src/main/kotlin/org/jetbrains/intellij/platform/gradle/tasks/InitializeIntelliJPlatformPluginTask.kt#L148 . This was added in the commit [JetBrains/intellij-platform-gradle-plugin@a759cec](https://github.com/JetBrains/intellij-platform-gradle-plugin/commit/a759cecfe7e68aa0360e25586cb2e2ed2f517b16) . Looks like this change could break coroutine debugging for IDEA plugin authors? Could you investigate?


@hsz would you mind taking a look?

kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
```diff
@@ -3,9 +3,13 @@ package kotlinx.coroutines.debug.internal
...
  * Object used to differentiate between agent installed statically or dynamically.
  * This is done in a separate object so [DebugProbesImpl] can check for static installation
- * without having to depend on [kotlinx.coroutines.debug.AgentPremain], which is not compatible with Android.
+ * without having to depend on [AgentPremain], which is not compatible with Android.
  * Otherwise, access to `AgentPremain.isInstalledStatically` triggers the load of its internal `ClassFileTransformer`
  * that is not available on Android.
+ *
+ * Usage Note: Fleet (Reflection): FleetDebugProbes
+ * Usage Note: Android (Hard Coded, ignored for Leak Detection)
+ * Usage Note: IntelliJ (Suppress KotlinInternalInJava): CoroutineDumpState
```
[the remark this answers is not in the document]
sellmair: I tried to add small notes on internal APIs which are forcefully used by clients (by reflection or ignoring internal visibility). Ss those clients are 'not to be broken', I think its worth marking the code somehow as "internal, but breached'

integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
sellmair: You see that this test is Ignored: I could not get JPMS and Dynamic Debug Probes to be happy with each other. I suspect that the shading of byte-buddy is a problem here. 

This feature was not working previously, but I thought it's nice to keep the test and comment the 
- conditions under which this was tested
- the issue we were facing

sellmair:
I agreed with @hsz that I will track the kotlinx coroutines release: 
Once a version of coroutines, which includes this change, lands in IntelliJ, I will notify the team and they will introduce the change 👍

kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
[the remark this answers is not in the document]
dkhalanskyjb: Ok, makes sense. Though a comment would not stop us: who reads them? Better to mark this as `@PublishedApi` so that this appears in the binary compatibility validator dump. Also, it would help a lot if it were reworded, like "This object must not be renamed or moved, because this would break binary compatibility for the following clients forcefully access it:", and then the list. Even with zero context, this gives a clear and actionable impression.

kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
```diff
@@ -3,9 +3,13 @@ package kotlinx.coroutines.debug.internal
...
  */
 internal object AgentInstallationType {
     internal var isInstalledStatically = false
```
dkhalanskyjb, line 15: No sense to mark this as `internal`, as we want this in the BCV dump.

integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
dkhalanskyjb: Let me double-check my understanding: do you mean that, if these JVM runtime arguments are added, the test will pass, but without them, it won't? Why not introduce a separate source set where these arguments are passed, so that we know this workaround works reliably?

Also, there could be a separate test that shows that there still *is* this problem; something like `assertFailsWith<IllegalStateException> { test body }`. With the additional comment like "we want this test to fail", we'll be able to immediately learn if something changes and this usage becomes possible without workarounds.

dkhalanskyjb (changes requested):
Please also add descriptions of the new tests to `integration-testing/README.md`

sellmair:
> Please also add descriptions of the new tests to `integration-testing/README.md`

Done ✅

kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
dkhalanskyjb: `isInstalledStatically` is used from the listed products, rights? Then it, too, should be in the API dump, so not `internal`.

integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+            runBlocking {
+                val baos = ByteArrayOutputStream()
+                DebugProbes.dumpCoroutines(PrintStream(baos))
+                // if the agent works, then dumps should contain something,
+                // at least the fact that this test is running.
+                Assert.assertTrue(baos.toString().contains("testAgentDumpsCoroutines"))
+            }
+        }
+
+    @Test()
```
dkhalanskyjb, line 38:
```suggestion
    @Test
```

integration-testing/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugTest.kt
```diff
@@ -0,0 +1,54 @@
...
+            runBlocking {
+                val baos = ByteArrayOutputStream()
+                DebugProbes.dumpCoroutines(PrintStream(baos))
+                // if the agent works, then dumps should contain something,
+                // at least the fact that this test is running.
+                Assert.assertTrue(baos.toString().contains("testAgentDumpsCoroutines"))
+            }
+        }
+
+    @Test()
```
dkhalanskyjb, line 46:
```suggestion
    @Test
```

integration-testing/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugTest.kt
```diff
@@ -0,0 +1,54 @@
+@file:OptIn(ExperimentalCoroutinesApi::class)
```
dkhalanskyjb, line 1: It looks like the whole `integration-testing/src/debugDynamicAgentJpmsTest` directory is not even used at all and is just a copy of `integration-testing/jpmsTest/src`?

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/jvm/src/debug/internal/AgentInstallationType.kt
sellmair:
```kotlin
@PublishedApi
internal object AgentInstallationType {
    var isInstalledStatically = false
}
```

Results in 
```
public final class kotlinx.coroutines.debug.internal.AgentInstallationType {
  public static final kotlinx.coroutines.debug.internal.AgentInstallationType INSTANCE;
  public final boolean isInstalledStatically();
  public final void setInstalledStatically(boolean);
  static {};
}

```

sellmair: However the original code 
```kotlin
@PublishedApi
internal object AgentInstallationType {
    internal var isInstalledStatically = false
}

```

Results in 
```
public final class kotlinx.coroutines.debug.internal.AgentInstallationType {
  public static final kotlinx.coroutines.debug.internal.AgentInstallationType INSTANCE;
  public final boolean isInstalledStatically$kotlinx_coroutines_core();
  public final void setInstalledStatically$kotlinx_coroutines_core(boolean);
  static {};
}

```

sellmair: Therefore, making it public will break current reflection users
Code example from Fleet
```
  override fun setInstalledStatically(installed: Boolean) {
    c.getDeclaredMethod("setInstalledStatically\$kotlinx_coroutines_core", Boolean::class.java).invoke(instance, installed)
  }
  override fun isInstalledStatically(): Boolean {
    return c.getDeclaredMethod("isInstalledStatically\$kotlinx_coroutines_core").invoke(instance) as Boolean
  }
```

integration-testing/jpmsTest/src/debugDynamicAgentJpmsTest/kotlin/DynamicAttachDebugJpmsTest.kt
sellmair: As far as I am aware, Titouan Bion is already working on finding ways of making this test pass!

dkhalanskyjb (approved):
Thank you! This issue was really annoying, it's great that you fixed it!

Issue #4212, opened 2024-08-12
Introduce Flow.all/any/none operators


CLOVIS-AI:
**Use case**

I have a complex `Flow` that possesses many elements. I have a business rule that is, literally, to do something if any of them satisfies a condition.

If I were using `List` or `Set`, I would use `any { theCondition(it) }`. However, `Flow` doesn't seem to have `any`.

There is an old issue (#2239) that asks for this feature, and is closed because of alternative implementations:
> ```kotlin
> suspend fun <T : Any> Flow<T>.any(predicate: suspend (T) -> Boolean): Boolean {
>     return this.firstOrNull { predicate(it) } != null
> }
>
> suspend fun <T> Flow<T>.all(predicate: suspend (T) -> Boolean): Boolean {
>     return this.count { !predicate(it) } == 0
> }
> ```

I dislike this solution because:
- The first one doesn't work with a `Flow<Foo?>` because `firstOrNull` is only available on non-nullable types.
- Both options obfuscate the operation I am actually attempting to perform.

I believe it is worth having `any`/`all`/`none` directly in the library because the proposed implementation have downsides.

**The Shape of the API**

```kotlin
suspend fun <T> Flow<T>.any(predicate: suspend (T) -> Boolean): Boolean
suspend fun <T> Flow<T>.all(predicate: suspend (T) -> Boolean): Boolean
suspend fun <T> Flow<T>.none(predicate: suspend (T) -> Boolean): Boolean
```

**Prior Art**

These three functions are already available on `Iterable` and `Sequence`. The behavior should be the same.

`all` and `none` can use the `count` operator under the hood, since it already shortcuts. I believe this is a possible implementation of `any`, though I haven't tested it yet:

```kotlin
suspend fun <T> Flow<T>.any(predicate: suspend (T) -> Boolean) = this
    .transformWhile {
        if (predicate(it)) {
            emit(true)
            false // ONE element MATCHES, no need to go further
        } else {
            true // continue
        }
    }
    .onEmpty { emit(false) }
    .first()
```

I can submit a PR if you think this is worth pursuing.

CLOVIS-AI:
Simplified implementation:

```kotlin
suspend fun <T> Flow<T>.any(predicate: (T) -> Boolean) = this
    .filter { predicate(it) }
    .map { true }
    .onEmpty { emit(false) }
    .first()
```

dkhalanskyjb:
[...]

PR #4249, opened 2024-10-14
Introduce Flow.any, Flow.all, Flow.none

CLOVIS-AI:
As discussed in #4212.

This is my first proper contribution to Coroutines, so please be nice :)

At this stage, my goals are:
• Optimal API surface
• Correctness

I have not made any attempts to improve performance. In particular, I wouldn't be surprised if `any` could be improved. I believe `all` and `none` are optimal.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt
@@ -164,3 +164,46 @@ public fun <T> Flow<T>.chunked(size: Int): Flow<List<T>> {
         result?.let { emit(it) }
     }
 }
+
+/**
+ * Returns `true` if at least one element matches the given [predicate].
+ *
+ * This operation is *terminal*.
+ *
+ * @see Iterable.any
+ * @see Sequence.any
+ */
+@ExperimentalCoroutinesApi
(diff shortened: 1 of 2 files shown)


CLOVIS-AI:
I took inspiration from other tests, which seem to use `expect(Int)` and `finish(Int)`, but I'm not completely sure what they're for nor how to use them, so I didn't. Please let me know if they are necessary.

dkhalanskyjb (changes requested):
`expect` and `finish` are used to describe when exactly which code is run. If there are 100 steps, the chain of calls must be `expect(1)`, `expect(2)`, `expect(3)` ... `finish(100)` and nothing else. If the test is broken and `expect(5)` runs before `expect(4)` did, the test fails.

This isn't useful for the tests you've already written. It could potentially be useful for tests that are missing, those that show short-circuiting behavior: `flow { emit(1); fail("unreached") }.any { it == 1 }` should finish without an error. Or maybe not.

qwwdfsad:
@CLOVIS-AI would you like to continue this PR or should we handle it from here?

CLOVIS-AI:
I would love to, but sadly my schedule doesn't allow me too… Feel free to make the changes you want if I'm too slow :)

dkhalanskyjb:
Closed in favor of #4304

Issue #4034, opened 2024-02-06
flatMapLatest + combine + first causes 


ansman:
**Describe the bug**

Firstly, it's possible this is expected behavior but it certainly doesn't seem that way.

We have a flow that we collect using `first()`. This flow is a combination of multiple upstream flows but they all emit their items synchronously. The downstream flow limits the number of emissions to a single element then uses `awaitCompletion()` to avoid terminating the flow. This is the flow we call `first()` on.

Under some circumstances, we are seeing the `first()` call resuming _after_ the scope has been cancelled which is causing a crash for us.

If either the `flatMapLatest` or `combine` is removed, the code behaves as expected, although `first()` never resumes at these `x

I know there are no real guarantees that code won't be running while canceled. But there are few things that makes me think that this doesn't apply in this case:
- There is no non-cooperative suspension as far as I know.
- Only a single thread is involved.

**Provide a Reproducer**
This reproduces it on an Android device, but it's possible it can reproed on the JVM too.
```kotlin
// This must be run on the main thread
val scope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
println("Before launch")
scope.launch {
    flowOf(listOf(1, 2, 3))
        .flatMapLatest { numbers ->
            combine(numbers.map { flowOf(it) }) { it.sum() }
        }
        .onEach { println("Got item") }
        .first()
    println("After first, scope is active: ${scope.isActive}")
}
println("After launch")
scope.cancel()
```
I expect this to print:
```
Before launch
Got item
After first, scope is active: true
After launch
```

Instead I'm seeing this:
```
Before launch
Got item
After launch
After first, scope is active: false
```

dkhalanskyjb:
Could not reproduce with coroutines 1.7.3 using Android instrumented tests. The output for me is:
```
Before launch
Got item
After launch
```
Due to cancellation, the "After first" `println` didn't execute.

ansman:
Yes, that's right. It does _not_ repro in an instrumented test for us either. But it does when you put that code in `Activity.onCreate` it does. I'll attach a complete project.

ansman:
Here is a complete example. Run it and see the output in logcat.
[Coroutinesbug.zip](https://github.com/Kotlin/kotlinx.coroutines/files/14195972/Coroutinesbug.zip)

dkhalanskyjb:
Ok, reproduced.

ansman:
[...]

PR #4254, opened 2024-10-21
Rethrow AbortFlowException upon cancellation (#4034)

jxdabc:
collectWhile catches AbortFlowException even if the CoroutineScope is cancelled while waiting on child tasks, causing it to fail to respond to cancel().

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
@@ -5,6 +5,7 @@ package kotlinx.coroutines.flow
 
 import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.internal.*
+import kotlin.coroutines.*
 import kotlin.jvm.*
 import kotlinx.coroutines.flow.flow as safeFlow
 import kotlinx.coroutines.flow.internal.unsafeFlow as flow
@@ -133,5 +134,7 @@ internal suspend inline fun <T> Flow<T>.collectWhile(crossinline predicate: susp
         collect(collector)
     } catch (e: AbortFlowException) {
         e.checkOwnership(collector)
+        // The task might have been cancelled before AbortFlowException was thrown.
+        coroutineContext.ensureActive()
     }
 }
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
```diff
@@ -133,5 +134,8 @@ internal suspend inline fun <T> Flow<T>.collectWhile(crossinline predicate: susp
         collect(collector)
     } catch (e: AbortFlowException) {
         e.checkOwnership(collector)
+        // If thrown in a CoroutineScope waiting on child tasks, this exception can override a later CancellationException
+        // from cancel(). Rethrow if it does.
+        if (!coroutineContext.isActive) { throw e }
```
dkhalanskyjb, line 139: Why not just `ensureActive()`?

kotlinx-coroutines-core/common/test/flow/terminal/FirstTest.kt
```diff
@@ -173,4 +175,19 @@ class FirstTest : TestBase() {
 
         assertFailsWith<CancellationException> { flow.first() }
     }
-}
+
+    @Test
+    fun test() = runTest {
```
dkhalanskyjb, line 180: The name of the test is not informative.

kotlinx-coroutines-core/common/test/flow/terminal/FirstTest.kt
```diff
@@ -173,4 +175,19 @@ class FirstTest : TestBase() {
...
+                coroutineScope {
+                    emitAll(produce(start = UNDISPATCHED) {
+                        channel.send(Unit)
+                    })
+                }
+            }.first()
+            expectUnreached()
+        }
+        job.cancel()
+    }
```
dkhalanskyjb, line 192: In this scenario, it's worth it to use `expect` and `finish` to assert the order of execution. The rules:
* `expect(1)` is the first invocation in any test.
* `expect(n + 1)` can only be called after `expect(n)`.
* `finish(n + 1)` can only be called after `expect(n)`.
* Each test that uses an `expect` must call `finish`.
* After `finish`, no extra `expect` or `finish` can be called.

With this, we can ensure that indeed, the test checks what we want it to check: first, cancellation happens, then, `first` throws a `CancellationException`.

dkhalanskyjb (changes requested):

jxdabc:
comments resolved

dkhalanskyjb (approved):

dkhalanskyjb:
Thanks!

PR #4258, opened 2024-10-24
Remove star import of the packages, which contain conflicting declarations

mvicsokolova:
Star imports of `java.util.concurrent.atomic.*` and `kotlin.concurrent.*` will lead to the compilation error: once Atomic classes are introduced in `kotlin.concurrent` package, they will conflict with Java atomic classes, which have the same names.

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/jdk8/future/FutureTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/jdk8/future/FutureTest.kt
@@ -10,7 +10,7 @@ import java.util.concurrent.*
 import java.util.concurrent.atomic.*
 import java.util.concurrent.locks.*
 import java.util.function.*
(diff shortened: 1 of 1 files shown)


mvicsokolova:
And I'll need this commit in `kotlin-community/dev` branches too, when we merge it)

dkhalanskyjb (approved):
`git grep -lF 'kotlin.concurrent.*' | xargs git grep -lF 'java.util.concurrent.atomic.*'` confirms that only this file has both of the star imports.

Normally, I'd ask for a comment near the import (so as to avoid accidentally reverting the change), but in this case, the stakes don't seem high: in the worst case, the CI will fail, but we'll likely upgrade to the new atomics before we touch the file for any reason, so even that is far-fetched.

We can merge this once the CI passes.

PR #4260, opened 2024-10-25
Improve the yield() documentation

dkhalanskyjb:


qwwdfsad (approved):

kotlinx-coroutines-core/common/src/Yield.kt
```diff
@@ -4,23 +4,143 @@ import kotlinx.coroutines.internal.*
 import kotlin.coroutines.intrinsics.*
 
 /**
- * Yields the thread (or thread pool) of the current coroutine dispatcher
- * to other coroutines on the same dispatcher to run if possible.
+ * Suspends this coroutine and immediately schedules it for further execution.
  *
+ * A coroutine run uninterrupted on a thread until the coroutine *suspend*,
```
lukellmann, line 9:
``` suggestion
 * A coroutine runs uninterrupted on a thread until the coroutine *suspends*,
```

kotlinx-coroutines-core/common/src/Yield.kt
```diff
@@ -4,23 +4,143 @@ import kotlinx.coroutines.internal.*
...
+ * giving other coroutines a chance to use that thread for their own computations.
+ * Normally, coroutines suspend whenever they wait for something to happen:
+ * for example, trying to receive a value from a channel that's currently empty will suspend.
+ * Sometimes, a coroutine does not need to wait for anything,
+ * but we still want it to give other coroutines a chance to run.
+ * Calling [yield] has this effect:
+ *
+ * ```
+ * fun updateProgressBar(value: Int, marker: String) {
```
lukellmann, line 18: Is the `value` parameter necessary for this example? Removing the parameter (and not passing `it`) seems simpler and allows you to focus on the behavior that this example is intended to illustrate.

PR #4268, opened 2024-11-06
Update CancellableContinuation documentation

qwwdfsad:


kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
  * the specified cancel cause.
  *
- * An instance of `CancellableContinuation` is created by the [suspendCancellableCoroutine] function.
+ * ### Usage
+ *
+ * An instance of `CancellableContinuation` can only be obtained by the [suspendCancellableCoroutine] function.
+ * The interface itself is public for use and private for implementation.
+ *
+ * A typical use of this function is to suspend a coroutine while waiting for a result
+ * from a callback or external source of values:
```
dkhalanskyjb, line 23:
```suggestion
 * A typical usage of this function is to suspend a coroutine while waiting for a result
 * from a callback or an external source of values that also support cancelling.
 * If the callback does not support cancelling, consider using [suspendCoroutine].
 *
 * Example:
```
I saw a question like, "What do I pass to `invokeOnCancellation` if my callback can't be unregistered?"

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ *     }
+ *     // Cancel the computation if the continuation itself was cancelled because a caller of 'await' is cancelled
+ *     c.invokeOnCancellation { future.cancel(true) }
+ * }
+ * ```
+ *
+ * ### Thread-safety
+ *
+ * Instances of [CancellableContinuation] are thread-safe and can be safely shared across multiple threads.
+ * [CancellableContinuation] allows concurrent invocations of [cancel] and [resume] pair, guaranteing
```
dkhalanskyjb, line 45:
```suggestion
 * [CancellableContinuation] allows concurrent invocations of the [cancel] and [resume] pair, guaranteeing
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * that only one of these operations will succeed.
+ * Concurrent invocations of [resume] methods lead to a [IllegalStateException] and considered a programmatic error.
```
dkhalanskyjb, line 47:
```suggestion
 * Concurrent invocations of [resume] methods lead to a [IllegalStateException] and are considered a programmatic error.
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -45,18 +111,23 @@ import kotlin.coroutines.intrinsics.*
...
+     * Returns `true` when this continuation was completed -- [resumed][Continuation.resumeWith] or
+     * [cancelled][CancellableContinuation.cancel].
+     *
+     * This state implies that [isActive] is `false`.
      */
     public val isCompleted: Boolean
 
     /**
-     * Returns `true` if this continuation was [cancelled][cancel].
+     * Returns `true` if this continuation was [cancelled][CancellableContinuation.cancel].
```
dkhalanskyjb, line 130: Why is this needed? Do we need to introduce such a fix throughout the documentation of every class/interface referring to its other members? I hope not: if just writing `[cancel]` is problematic, to me, this looks like a bug in our documentation tooling.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -45,18 +111,23 @@ import kotlin.coroutines.intrinsics.*
 @SubclassOptInRequired(InternalForInheritanceCoroutinesApi::class)
 public interface CancellableContinuation<in T> : Continuation<T> {
     /**
-     * Returns `true` when this continuation is active -- it has not completed or cancelled yet.
+     * Returns `true` when this continuation is active -- it was created,
+     * but not yet [resumed][Continuation.resumeWith] or [cancelled][CancellableContinuation.cancel].
+     *
+     * This state implies that [isCompleted] and [isCancelled] are `false`.
```
dkhalanskyjb, line 117:
```suggestion
     * This state implies that [isCompleted] and [isCancelled] are `false`,
     * but this can change immediately after the invocation because of parallel calls to [cancel] and [resume].
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -124,6 +195,7 @@ public interface CancellableContinuation<in T> : Continuation<T> {
     /**
      * Cancels this continuation with an optional cancellation `cause`. The result is `true` if this continuation was
      * cancelled as a result of this invocation, and `false` otherwise.
+     * [cancel] might return `false` when the continuation was either [resumed][resume] or already [cancelled][cancel].
```
dkhalanskyjb, line 198: But here, it's just `[cancel]` and `[resume`]?..

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * ### Thread-safety
+ *
+ * Instances of [CancellableContinuation] are thread-safe and can be safely shared across multiple threads.
+ * [CancellableContinuation] allows concurrent invocations of [cancel] and [resume] pair, guaranteing
+ * that only one of these operations will succeed.
+ * Concurrent invocations of [resume] methods lead to a [IllegalStateException] and considered a programmatic error.
+ *
+ * ### Prompt cancellation guarantee
+ *
+ * Cancellable continuation provides a **prompt cancellation guarantee**.
```
dkhalanskyjb, line 51:
```suggestion
 * A cancellable continuation provides a **prompt cancellation guarantee**.
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ *
+ * The cancellation of the coroutine's job is generally asynchronous with respect to the suspended coroutine.
+ * The suspended coroutine is resumed with a call to its [Continuation.resumeWith] member function or to the
+ * [resume][Continuation.resume] extension function.
+ * However, when the coroutine is resumed, it does not immediately start executing but is passed to its
+ * [CoroutineDispatcher] to schedule its execution when the dispatcher's resources become available for execution.
+ * The job's cancellation can happen before, after, and concurrently with the call to `resume`. In any
+ * case, prompt cancellation guarantees that the coroutine will not resume its code successfully.
+ *
+ * If the coroutine was resumed with an exception (for example, using [Continuation.resumeWithException] extension
```
dkhalanskyjb, line 64:
```suggestion
 * If the coroutine was resumed with an exception (for example, using the [Continuation.resumeWithException] extension
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * function) and cancelled, then the exception thrown by the `suspendCancellableCoroutine` function is determined
+ * by what happened first: exceptional resume or cancellation.
+ *
+ * ### Resuming with a closeable resource
+ *
+ * [CancellableContinuation] provides the capability for working with values that represent a resource that should be
```
dkhalanskyjb, line 70:
```suggestion
 * [CancellableContinuation] provides the capability to work with values that represent a resource that should be
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * closed. For that, it proved `resume(value: R, onCancellation: ((cause: Throwable, value: R, context: CoroutineContext) -> Unit)`
```
dkhalanskyjb, line 71:
```suggestion
 * closed. For that, it provides the `resume(value: R, onCancellation: ((cause: Throwable, value: R, context: CoroutineContext) -> Unit)`
```
?

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * function that guarantees either the given `value` will be successfully returned from the corresponding `suspend` function
```
dkhalanskyjb, line 72:
```suggestion
 * function that guarantees that either the given `value` will be successfully returned from the corresponding `suspend` function
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * ```
+ * continuation.resume(resourceToResumeWith) { _, resourceToClose, _
+ *     // Will be invoked if the continuation is cancelled while being dispatched
+ *     resourceToClose.close()
+ * }
+ * ```
+ *
+ * #### Continuation states
+ *
+ * Cancellable continuation has three observable states:
```
dkhalanskyjb, line 84:
```suggestion
 * A cancellable continuation has three observable states:
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
  *
- * Invocation of [cancel] transitions this continuation from _active_ to _cancelled_ state, while
- * invocation of [Continuation.resume] or [Continuation.resumeWithException] transitions it from _active_ to _resumed_ state.
- *
- * A [cancelled][isCancelled] continuation implies that it is [completed][isCompleted].
+ * For a detailed description of each state, see the corresponding properties' documentation.
  *
- * Invocation of [Continuation.resume] or [Continuation.resumeWithException] in _resumed_ state produces an [IllegalStateException],
- * but is ignored in _cancelled_ state.
+ * A successful invocation of [cancel] transitions continuation from _active_ to _cancelled_ state, while
```
dkhalanskyjb, line 94:
```suggestion
 * A successful invocation of [cancel] transitions the continuation from an _active_ to a _cancelled_ state, while
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * invocation of [Continuation.resume] or [Continuation.resumeWithException] transitions it from _active_ to _resumed_ state.
```
dkhalanskyjb, line 95:
```suggestion
 * an invocation of [Continuation.resume] or [Continuation.resumeWithException] transitions it from an _active_ to a _resumed_ state.
```

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
```diff
@@ -4,31 +4,97 @@ import kotlinx.coroutines.internal.*
...
+ * }
+ * ```
+ *
+ * ### Thread-safety
+ *
+ * Instances of [CancellableContinuation] are thread-safe and can be safely shared across multiple threads.
+ * [CancellableContinuation] allows concurrent invocations of [cancel] and [resume] pair, guaranteing
+ * that only one of these operations will succeed.
+ * Concurrent invocations of [resume] methods lead to a [IllegalStateException] and considered a programmatic error.
+ *
```
dkhalanskyjb, line 48:
```suggestion
 * Concurrent invocations of [cancel] methods is permitted, and at most one of them succeeds.
 *
```
I'd consider rewriting this section not in terms of `cancel`/`resume` pairs but `n * cancel + 1 * resume` parallel calls (as that's more accurate) if this doesn't make the documentation less readable.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
qwwdfsad: >or an external source of values that also support cancelling.

This seems to be misleading -- the source might not support cancellation, but our suspending function does.
Using `suspendCoroutine` removes cancellation via Job completely. (Also, support**s** or you've meant something else?)

qwwdfsad: Rephrased it, PTAL

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
qwwdfsad: I wanted to be more explicit, I can revert that if you believe it's important.
I.e. it could've been some top-level `cancel`, too much of a context to process at once

dkhalanskyjb: If it's just a stylistic choice (that no one except the ones reading pure KDoc text will see?), it's not important to me either way, but if this is needed for the link to correctly resolve, that's alarming.

kotlinx-coroutines-core/common/src/CancellableContinuation.kt
dkhalanskyjb: Cancelling the coroutine without cancelling the callback actually causes confusion (I'm not sure that I can find a link to one of the relevant discussions quickly, though): "Hey, I've cancelled the request almost immediately, how come it still goes through?" There's clearly some expectation that coroutine cancellation is magic that automatically deals with blocking calls and unregistering callbacks. We can't do much about this, but at least the docs should expose the brutal truth to those who actually read them.

Don't let that block the PR, though. If you agree this is worth including and can think of correct phrasing that highlights that it's fully up to the programmer to integrate cancellation / what happens when cancellation is only supported partially, please go ahead; if not, we can look into this later.

dkhalanskyjb (approved):

PR #4269, opened 2024-11-06
Improve CoroutineScope extensions documentation

qwwdfsad:


kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -117,13 +123,30 @@ public fun MainScope(): CoroutineScope = ContextScope(SupervisorJob() + Dispatch
 /**
  * Returns `true` when the current [Job] is still active (has not completed and was not cancelled yet).
  *
- * Check this property in long-running computation loops to support cancellation:
+ * A coroutine cancallation [is cooperative](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative)
```
dkhalanskyjb, line 126:
```suggestion
 * Coroutine cancallation [is cooperative](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative)
```

The article changes the meaning from "coroutine cancellation (as a concept; in general; as a set of rules and behaviors) is cooperative" to "each instance of coroutine cancellation happens in a cooperative manner". Both are true, but I guess the first one is what was meant.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -117,13 +123,30 @@ public fun MainScope(): CoroutineScope = ContextScope(SupervisorJob() + Dispatch
...
+ * and normally, the cancellation machinery happens when the coroutine *suspends*, for example,
+ * when trying to read from a [channel][kotlinx.coroutines.channels.Channel] that is empty.
+ *
+ * Sometimes, a coroutine does not need to do any of such operations, but still wants to be a cooperative
```
dkhalanskyjb, line 130:
```suggestion
 * Sometimes, a coroutine does not need to do any of such operations, but still wants to be cooperative
```

This is a funny typo!

> cooperative (noun)
> an enterprise or organization owned by and operated for the benefit of those using its services

Who doesn't want to be one?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -117,13 +123,30 @@ public fun MainScope(): CoroutineScope = ContextScope(SupervisorJob() + Dispatch
 /**
  * Returns `true` when the current [Job] is still active (has not completed and was not cancelled yet).
  *
- * Check this property in long-running computation loops to support cancellation:
+ * A coroutine cancallation [is cooperative](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative)
+ * and normally, the cancellation machinery happens when the coroutine *suspends*, for example,
```
dkhalanskyjb, line 127:
```suggestion
 * and normally, it's checked if a coroutine is cancelled when it *suspends*, for example,
```
I think that's more specific without being misleading.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb, line 130:
```suggestion
 * Sometimes, a coroutine does not need to perform suspending operations, but still wants to be a cooperative
```
More specific?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -117,13 +123,30 @@ public fun MainScope(): CoroutineScope = ContextScope(SupervisorJob() + Dispatch
...
+ * when trying to read from a [channel][kotlinx.coroutines.channels.Channel] that is empty.
+ *
+ * Sometimes, a coroutine does not need to do any of such operations, but still wants to be a cooperative
+ * and respect cancellation.
+ *
+ * [isActive] property is inteded to be used for scenarios like this:
```
dkhalanskyjb, line 133:
```suggestion
 * The [isActive] property is inteded to be used for scenarios like this:
```
[— There's a uniquely named property that you want.
— Which property is that?
—] The `isActive` property.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -292,24 +315,41 @@ public fun CoroutineScope.cancel(cause: CancellationException? = null) {
 public fun CoroutineScope.cancel(message: String, cause: Throwable? = null): Unit = cancel(CancellationException(message, cause))
 
 /**
- * Ensures that current scope is [active][CoroutineScope.isActive].
+ * Ensures that current scope is [active][CoroutineScope.isActive] and throws the [CancellationException] that was the
+ * scope's cancellation cause otherwise.
```
dkhalanskyjb, line 319:
```suggestion
 * Throws the [CancellationException] that was the scope's cancellation cause if the scope is no longer [active][CoroutineScope.isActive].
```
Otherwise, it may leave the impression that "ensuring the scope is active" is some procedure that's not just throwing an exception.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -292,24 +315,41 @@ public fun CoroutineScope.cancel(cause: CancellationException? = null) {
...
  *
- * If the job is no longer active, throws [CancellationException].
- * If the job was cancelled, thrown exception contains the original cancellation cause.
- * This function does not do anything if there is no [Job] in the scope's [coroutineContext][CoroutineScope.coroutineContext].
+ * A coroutine cancallation [is cooperative](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative)
+ * and normally, the cancellation machinery happens when the coroutine *suspends*, for example,
+ * when trying to read from a [channel][kotlinx.coroutines.channels.Channel] that is empty.
+ *
+ * Sometimes, a coroutine does not need to do any of such operations, but still wants to be a cooperative
+ * and respect cancellation.
```
dkhalanskyjb, line 326:
```suggestion
 * Coroutine cancallation [is cooperative](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative)
 * and normally, it's checked if a coroutine is cancelled when it *suspends*, for example,
 * when trying to read from a [channel][kotlinx.coroutines.channels.Channel] that is empty.
 *
 * Sometimes, a coroutine does not need to perform suspending operations, but still wants to be cooperative
 * and respect cancellation.
```
The changes also proposed for `isActive`, in bulk.

dkhalanskyjb (approved):
I see some potential to improve wording in minor ways, but in general, the new suggestions are a significant improvement, thanks!

Issue #4209, opened 2024-08-12
Dispatcher failures may leave coroutines uncompleted


dkhalanskyjb:
**Describe the bug**

While reviewing https://github.com/Kotlin/kotlinx.coroutines/pull/4181, it became evident that we don't properly handle failures in coroutine dispatchers, and this can surface in ways other than just strange-looking exceptions.

**Provide a Reproducer**

```kotlin
val dispatcher = newSingleThreadContext("unreliable friend")
runTest {
    launch(dispatcher, start = CoroutineStart.UNDISPATCHED) {
        try {
            println("This code runs...")
            suspendCancellableCoroutine<Int> { cont ->
                // we launch a separate thread and wait a bit,
                // because we want the coroutine to actually suspend
                // and go through a dispatch.
                launch(Dispatchers.Default) {
                    delay(100)
                    // close the dispatcher, now it will throw on `dispatch`
                    dispatcher.close()
                    // try dispatching the coroutine
                    cont.resume(3)
                }
            }
        } catch (e: Throwable) {
            println("Caught $e")
            throw e
        } finally {
            println("... therefore, this code must run.")
        }
    }
}
```

This code will hang after printing `This code runs...`, as the launched coroutine never finishes.

dkhalanskyjb:
It's unclear what to do in this scenario, though. Here are some options:

* Try to fail the whole program by calling `handleCoroutineException`. We don't call it a mechanism of last-resort exception propagation for nothing.
* On the JVM, [`Executor.asCoroutineDispatcher`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/as-coroutine-dispatcher.html) tries cancelling the task and redispatching it to `Dispatchers.IO` if the executor rejected it. We can't do literally that in common code, as there is no `Dispatchers.IO` on platforms with no threads, but if there are no threads, we can just use `DefaultExecutor`. The problem is, what if finalizers contain some code that must run on a specific thread? If `Dispatchers.Main` fails, it's usually a contract violation to try to run its code on `Dispatchers.IO`.
* We can pretend that the coroutine that couldn't get dispatched did finish with an error. This way, finalizers won't be run (which is a contract violation), but at least the whole coroutine hierarchy won't deadlock because of an incorrect dispatch.
* Something else?

iseki0:
What's the current status? I usually meet coroutines hang-up, especially when I create many and many coroutines. 
[...]

PR #4272, opened 2024-11-11
Improve handling of DispatchException in an undispatched case

qwwdfsad:


qwwdfsad:
Looking at https://github.com/Kotlin/kotlinx.coroutines/issues/4209#issuecomment-2324575057, I have yet to see an elegant, generalized solution to our problem that is neither a breaking change nor a contract breach

kotlinx-coroutines-core/common/src/intrinsics/Undispatched.kt
```diff
@@ -38,54 +38,66 @@ internal fun <R, T> (suspend (R) -> T).startCoroutineUndispatched(receiver: R, c
...
     }
 }
+
+private fun ScopeCoroutine<*>.notOwnTimeout(cause: Throwable): Boolean {
+    return cause !is TimeoutCancellationException || cause.coroutine !== this
+}
+
+private fun ScopeCoroutine<*>.dispatchException(e: DispatchException) {
+    makeCompleting(CompletedExceptionally(e.cause))
+    throw throw recoverStackTrace(e.cause, uCont)
```
dkhalanskyjb, line 102: One `throw` should be enough.

kotlinx-coroutines-core/common/src/intrinsics/Undispatched.kt
```diff
@@ -38,54 +38,66 @@ internal fun <R, T> (suspend (R) -> T).startCoroutineUndispatched(receiver: R, c
...
     } else {
         state.unboxState()
     }
 }
+
+private fun ScopeCoroutine<*>.notOwnTimeout(cause: Throwable): Boolean {
+    return cause !is TimeoutCancellationException || cause.coroutine !== this
+}
+
+private fun ScopeCoroutine<*>.dispatchException(e: DispatchException) {
```
dkhalanskyjb, line 100: I think this function should be inlined. It's not obvious that it calls `makeCompleting`.

kotlinx-coroutines-core/jvm/test/FailFastOnStartTest.kt
```diff
@@ -4,9 +4,17 @@ package kotlinx.coroutines
...
 import kotlinx.coroutines.testing.*
 import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
 import org.junit.*
 import org.junit.Test
 import org.junit.rules.*
+import org.junit.runners.model.TestTimedOutException
+import java.util.concurrent.TimeoutException
```
dkhalanskyjb, line 14: Are these imports used?

kotlinx-coroutines-core/common/src/intrinsics/Undispatched.kt
```diff
@@ -38,54 +38,66 @@ internal fun <R, T> (suspend (R) -> T).startCoroutineUndispatched(receiver: R, c
...
+    makeCompleting(CompletedExceptionally(e.cause))
```
dkhalanskyjb, line 101: Why not `makeCompletingOnce`, to highlight the aspects of this codepath that are different from the happy path?

kotlinx-coroutines-core/jvm/test/FailFastOnStartTest.kt
```diff
@@ -81,4 +89,24 @@ class FailFastOnStartTest : TestBase() {
     fun testAsyncNonChild() = runTest(expected = ::mainException) {
         async<Int>(Job() + Dispatchers.Main) { fail() }
     }
+
+    @Test
+    fun testFlowOn()  {
+        // See #4142, this test ensures that `coroutineScope { produce(failingDispatcher, ATOMIC) }`
+        // rethrows an exception. It does not help with the completion of such a coroutine though.
+        // Hack to avoid waiting for test completion
```
dkhalanskyjb, line 97: Where's the hack?

kotlinx-coroutines-core/jvm/test/FailFastOnStartTest.kt
```diff
@@ -81,4 +89,24 @@ class FailFastOnStartTest : TestBase() {
     fun testAsyncNonChild() = runTest(expected = ::mainException) {
         async<Int>(Job() + Dispatchers.Main) { fail() }
     }
+
+    @Test
+    fun testFlowOn()  {
```
dkhalanskyjb, line 94:
```suggestion
    fun testFlowOn() {
```

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/intrinsics/Undispatched.kt
qwwdfsad: I wanted to be extra-sure the exception is thrown

kotlinx-coroutines-core/common/src/intrinsics/Undispatched.kt
qwwdfsad: I am, in general, for the opposite approach -- get rid of all "very uncommon" cases from a code path that is supposed to be followed. I renamed the function and changed return type to be `Nothing` for the readability sake

kotlinx-coroutines-core/jvm/test/FailFastOnStartTest.kt
qwwdfsad: Nope. Fixed

kotlinx-coroutines-core/jvm/test/FailFastOnStartTest.kt
qwwdfsad: Made the comment more straightforward

dkhalanskyjb (approved):

PR #4283, opened 2024-11-29
Update atomicfu version to 0.26.1

mvicsokolova:


dkhalanskyjb:
@mvicsokolova, does the `kotlinx.atomicfu.enableJsIrTransformation=false` flag in `gradle.properties` still have some meaning?

PR #4284, opened 2024-11-29
Update Kotlin to 2.1.0

qwwdfsad:


kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -153,6 +153,7 @@ public suspend fun <T> withContext(
         // FAST PATH #1 -- new context is the same as the old one
         if (newContext === oldContext) {
             val coroutine = ScopeCoroutine(newContext, uCont)
+            @Suppress("LEAKED_IN_PLACE_LAMBDA") // Contract is preserved, invoked immediately or throws
```
qwwdfsad, line 156: This one (an one on the contract definition) is annoying, I'll report it to the team

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -161,11 +162,14 @@ public suspend fun <T> withContext(
...
             // There are changes in the context, so this thread needs to be updated
             withCoroutineContext(coroutine.context, null) {
+                @Suppress("LEAKED_IN_PLACE_LAMBDA") // Contract is preserved, invoked immediately or throws
                 return@sc coroutine.startUndispatchedOrReturn(coroutine, block)
             }
         }
         // SLOW PATH -- use new dispatcher
         val coroutine = DispatchedCoroutine(newContext, uCont)
+        // Contract is preserved, invoked immediately or throws
+        @Suppress("LEAKED_IN_PLACE_LAMBDA")
```
dkhalanskyjb, line 172: I'd prefer that the comment is to the right of `Suppress`: typically, the comment above a line with a suppression refers to the line itself.

gradle.properties
```diff
@@ -1,8 +1,8 @@
 # Kotlin
 version=1.9.0-SNAPSHOT
 group=org.jetbrains.kotlinx
-kotlin_version=2.0.0
-kotlin_language_version=2.0
+kotlin_version=2.1.0
+kotlin_language_version=2.1
```
dkhalanskyjb, line 5: Does this line do anything useful today other than accidentally disabling the "warnings are errors" flag?

dkhalanskyjb (approved):

gradle.properties
qwwdfsad: AFAIR release team uses it, but not sure. Will figure out separately

dkhalanskyjb: The release team may override this property, this I get, but do we need to keep it in gradle.properties?

qwwdfsad: At least for the sake of having "release team" branch and mainline having the same configuration and catching the same warnings/errors.

PR #4294, opened 2024-12-07
Revisit all @Transient usages in coroutines

### LINKED ISSUE #4293: Revisit @Transient-marked properties and ensure they are of nullable type
Otherwise, none of the code that using non-nullable `@Transient` is sound in the face of Java serialization.

See also: https://github.com/Kotlin/kotlinx.coroutines/pull/4291

qwwdfsad:
Fixes #4293

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/flow/internal/FlowExceptions.kt
+++ b/kotlinx-coroutines-core/jvm/src/flow/internal/FlowExceptions.kt
@@ -2,6 +2,10 @@ package kotlinx.coroutines.flow.internal
 
 import kotlinx.coroutines.*
 
+/**
+ * Implementation note: `owner` is an internal marked that is used ONLY for identity checks by coroutines machinery,
+ * and it's never exposed, thus it's safe to have it both `@Transient` and non-nullable.
+ */
 internal actual class AbortFlowException actual constructor(
     @JvmField @Transient actual val owner: Any
 ) : CancellationException("Flow was aborted, no more elements needed") {
--- a/kotlinx-coroutines-core/jvm/src/Exceptions.kt
+++ b/kotlinx-coroutines-core/jvm/src/Exceptions.kt
@@ -24,8 +24,14 @@ public actual fun CancellationException(message: String?, cause: Throwable?) : C
 internal actual class JobCancellationException public actual constructor(
     message: String,
     cause: Throwable?,
-    @JvmField @Transient internal actual val job: Job
+    job: Job
 ) : CancellationException(message), CopyableThrowable<JobCancellationException> {
+
+    @Transient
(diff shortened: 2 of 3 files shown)


qwwdfsad:
See also: https://youtrack.jetbrains.com/issue/KT-73762/Warn-about-Transient-being-not-sound-to-use-with-non-nullable-types

qwwdfsad:
In general, I cannot say that "trust me, this is safe" is, well, actually safe.

The problem with serialization is that it's impossible to reason about it in the bigger picture. We have our invariants, they are inherently *local* to the execution, and serializing-deserializing some of the things basically erases these invariants.

kotlinx-coroutines-core/jvm/src/flow/internal/FlowExceptions.kt
```diff
@@ -2,6 +2,10 @@ package kotlinx.coroutines.flow.internal
 
 import kotlinx.coroutines.*
 
+/**
+ * Implementation note: `owner` is an internal marked that is used ONLY for identity checks by coroutines machinery,
+ * and it's never exposed, thus it's safe to have it both `@Transient` and non-nullable.
```
dkhalanskyjb, line 7: It seems possible to maliciously deserialize and use an `AbortFlowException` that would fail in ways that legal instances wouldn't fail. I'm not sure why not just mark `owner` as nullable: I don't think anything breaks if we have `@JvmField @Transient val owner: Any? = owner` where a constructor only accepts a non-nullable version. Why not do that instead?

dkhalanskyjb (approved):
I have one minor remark about the contract of `AbortFlowException`, but I don't think it's a realistic concern; it's mostly a question of purity that can be ignored.

kotlinx-coroutines-core/jvm/src/flow/internal/FlowExceptions.kt
qwwdfsad: I thought about that as well and decided not to do it: our general contract is that `owner` should be unique, which nullable type will immediately defeat -- and also will open the door *for us* to introduce new bugs (also, additionally cognitive load on handling nulls).

All of that, for the sake of the really malicious actor that probably can work around their ways anyway

dkhalanskyjb: > also will open the door for us to introduce new bugs

I don't see the risk if

> a constructor only accepts a non-nullable version

qwwdfsad: That's harded to follow in the code itself, the concept of "potential malicious actor" starts poisoning the signatures and makes readers/operators authors/etc. to read through that

PR #4298, opened 2024-12-16
Temporarily disable Native IR transformations, the fix will be provided in 2.1.10

mvicsokolova:
The corresponding issue: [KT-73865](https://youtrack.jetbrains.com/issue/KT-73865/Incorrect-type-is-generated-for-irPropertyReference-during-K-N-transformation)

I'll provide the fix for this issue in `2.1.10`, if the release of coroutines is not planned with Kotlin 2.1.0, then we can leave this PR unmerged.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -47,4 +47,5 @@ kotlinx.atomicfu.enableJvmIrTransformation=true
 # When the flag below is set to `true`, AtomicFU cannot process
 # usages of `moveForward` in `ConcurrentLinkedList.kt` correctly.
 kotlinx.atomicfu.enableJsIrTransformation=false
-kotlinx.atomicfu.enableNativeIrTransformation=true
+# Temporarily disable Native IR transformation, the fix will be provided in 2.1.10: KT-73865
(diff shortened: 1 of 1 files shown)


qwwdfsad:
Along with that, it would be nice to understand (from the YT ticket) what went wrong in our acceptance criteria -- AFAIR coroutines with AFU are both part of nightlies and aggregates

dkhalanskyjb:
From an internal discussion, the risk is: in a future Kotlin version, the Kotlin/Native compiler may fail to compile code that uses `kotlinx.coroutines` that was compiled with Native atomicfu transformations using Kotlin 2.1.0. Our direct users (Kotlin programmers) would have to migrate to a newer `kotlinx.coroutines` version.

My hunch is that while the risk is valid, the performance drop for the end users is a more severe issue than the possibility that someone would have to upgrade to newer `kotlinx.coroutines` (since we don't break binary compatibility) from some short list of affected versions, so I vote for not merging this change.

qwwdfsad:
agreed

PR #4304, opened 2024-12-19
Introduce Flow.any, Flow.all, Flow.none

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/4212

Authored by `@CLOVIS-AI`


qwwdfsad (approved):
Nice one!

kotlinx-coroutines-core/common/src/flow/terminal/Logic.kt
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @see Iterable.any
+ * @see Sequence.any
+ */
+public suspend fun <T> Flow<T>.any(predicate: suspend (T) -> Boolean): Boolean {
+    var found = false
+    collectWhile {
+        predicate(it).also { satisfies ->
+            if (satisfies) found = true
+        }.let(Boolean::not)
```
JakeWharton, line 39: Can we do something that is easier to read like:

```suggestion
        val satisfies = predicate(it)
        if (satisfies) found = true
        !satisfies
```

kotlinx-coroutines-core/common/src/flow/terminal/Logic.kt
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @see Iterable.all
+ * @see Sequence.all
+ */
+public suspend fun <T> Flow<T>.all(predicate: suspend (T) -> Boolean): Boolean {
+    var foundCounterExample = false
+    collectWhile {
+        predicate(it).also { satisfies ->
+            if (!satisfies) foundCounterExample = true
+        }
```
dkhalanskyjb, line 76:
```suggestion
        val satisfies = predicate(it)
        if (!satisfies) foundCounterExample = true
        satisfies
```

dkhalanskyjb:
Merged manually in https://github.com/Kotlin/kotlinx.coroutines/commit/c6daacbe385d599af2ba2e494e9eb823398dcb7e to preserve authorship

PR #4314, opened 2024-12-24
resolve conflicts of module-info.class on kotlinx-coroutine-debug

RyuNen344:
- From version 1.10.0, kotlinx-coroutines-debug no longer bundles bytebuddy and has stopped using its custom shadowJar task.
  - https://github.com/Kotlin/kotlinx.coroutines/pull/4266
- It now conflicts with build-logic that were configured globally for Java 9 module system.
  - https://github.com/Kotlin/kotlinx.coroutines/blob/ee92d16c4b48345648dcd8bb15f11ab9c3747f67/buildSrc/src/main/kotlin/Java9Modularity.kt#L137-L146
- This conflict of module-info.class causes issues in programs that use ASM-based tools like Jacoco, so it needs to be fixed.

### DIFF
--- a/kotlinx-coroutines-debug/build.gradle.kts
+++ b/kotlinx-coroutines-debug/build.gradle.kts
@@ -1,5 +1,4 @@
 import org.gradle.api.JavaVersion
-import org.gradle.api.file.DuplicatesStrategy
 import org.gradle.api.tasks.bundling.Jar
 import org.gradle.api.tasks.testing.Test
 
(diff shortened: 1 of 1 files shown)


dkhalanskyjb (approved):
I'm not sure why this causes problems for some tools, but indeed, the current configuration has a problem: https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-debug/1.10.1/kotlinx-coroutines-debug-1.10.1.jar contains two files with the path `META-INF/versions/9/module-info.class`, and with your fix, only one such file remains. Thanks!

PR #4320, opened 2025-01-03
Specify Automatic-Module-Name for kotlinx-coroutines-core

### LINKED ISSUE #3842: Ambiguous module reference: kotlinx.coroutines.core
**Describe the bug**

I'm trying to upgrade to v1.7.3 and add the modules to a module-info file but it seems that `kotlinx-coroutines-core` and `kotlinx-coroutines-core-jvm` cause a problem because they have the same module name (I think)

image

My pom dependencies:

```xml
        <dependency>
            <groupId>org.jetbrains.kotlinx</groupId>
            <artifactId>kotlinx-coroutines-core</artifactId>
        </dependency>
        <dependency>
            <groupId>org.jetbrains.kotlinx</groupId>
            <artifactId>kotlinx-coroutines-jdk9</artifactId>
        </dependency>
```
This seems to happen because `kotlinx-coroutines-core-jvm` is a dependency of `kotlinx-coroutines-jdk9` and it's module name is `kotlinx-coroutines-core`. At the same time `kotlinx-coroutines-core` has an implied module name which is its package name. 
Am I importing these dependencies wrong?

fzhinkin:
Open questions are:
- [ ] The proper module name. Perhaps, `kotlinx.coroutines.core.metadata` will be better.
- [ ] Should we continue suggesting to use `kotlinx-coroutines-core` dependency in the readme file? I don't see any immediate benefits, but I see a downside: `-core` dependency adds an extra dep resolution step.

Fixes #3842

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Add dependencies (you can also add other modules that you need):
 ```xml
 <dependency>
     <groupId>org.jetbrains.kotlinx</groupId>
-    <artifactId>kotlinx-coroutines-core</artifactId>
+    <artifactId>kotlinx-coroutines-core-jvm</artifactId>
     <version>1.10.1</version>
 </dependency>
 ```
--- a/kotlinx-coroutines-core/build.gradle.kts
+++ b/kotlinx-coroutines-core/build.gradle.kts
@@ -1,5 +1,3 @@
-import org.gradle.api.tasks.testing.*
-import org.gradle.kotlin.dsl.*
 import org.jetbrains.kotlin.gradle.plugin.mpp.*
 import org.jetbrains.kotlin.gradle.targets.native.tasks.*
 import org.jetbrains.kotlin.gradle.tasks.*
@@ -182,15 +180,23 @@ val jvmJar by tasks.getting(Jar::class) { setupManifest(this) }
  * This manifest contains reference to AgentPremain that belongs to
  * kotlinx-coroutines-core-jvm, but our resolving machinery guarantees that
  * any JVM project that depends on -core artifact also depends on -core-jvm one.
+ *
+ * To avoid a conflict with a JPMS module provided by kotlinx-coroutines-core-jvm,
+ * an explicit automatic module name has to be specified in the manifest.
  */
-val allMetadataJar by tasks.getting(Jar::class) { setupManifest(this) }
+val allMetadataJar by tasks.getting(Jar::class) {
+    setupManifest(this, "kotlinx.coroutines.core.trampoline")
+}
 
-fun setupManifest(jar: Jar) {
+fun setupManifest(jar: Jar, autoModuleName: String? = null) {
     jar.manifest {
-        attributes(mapOf(
+        attributes(
             "Premain-Class" to "kotlinx.coroutines.debug.internal.AgentPremain",
-            "Can-Retransform-Classes" to "true",
-        ))
+            "Can-Retransform-Classes" to "true"
+        )
+        autoModuleName?.also {
+            attributes("Automatic-Module-Name" to it)
+        }
     }
 }
 


JakeWharton:
> The proper module name.

Definitely would vote for "common" or "metadata" over "trampoline".

README.md
```diff
@@ -85,7 +85,7 @@ Add dependencies (you can also add other modules that you need):
 ```xml
 <dependency>
     <groupId>org.jetbrains.kotlinx</groupId>
-    <artifactId>kotlinx-coroutines-core</artifactId>
+    <artifactId>kotlinx-coroutines-core-jvm</artifactId>
```
dkhalanskyjb, line 88: In my opinion, `-jvm` is visual noise in this case and doesn't actually mean anything. Of course it's JVM, we're mentioning a Kotlin project in a Maven file. I understand that there's an extra dependency resolution step, but does it have any non-negligible effect?

dkhalanskyjb (commented):
I can confirm that the reproducer provided at https://github.com/Kotlin/kotlinx.coroutines/issues/3842 no longer has the described issue after these changes.

My understanding of Java modules is surface-level, so please double-check my understanding. Right now, both `kotlinx-coroutines-core` and `kotlinx-coroutines-core-jvm` have the Java module name name `kotlinx.coroutines.core`; `core-jvm` specifies it explicitly, and `-core` just gets it automatically. If someone depends on `core`, they also depend on `core-jvm`, so any mentions of `kotlinx.coroutines.core` in Java module specifications will now correctly resolve to `core-jvm` (and not to both), but in addition to that, `kotlinx.coroutines.core.trampoline` will refer to a module containing nothing of use, so no one has any reason to write `kotlinx.coroutines.core.trampoline` in their code, ever. Right?

If so, the name I'd prefer would be scary and highlight the uselessness of the module. Something like `kotlinx.coroutines.core.artifact_disambiguating_module`.

README.md
fzhinkin: > I understand that there's an extra dependency resolution step, but does it have any non-negligible effect?

Just trying to make the world _better_ by saving ~130KB of networking traffic for each resolution when the artifact is not cached locally yet. :)

dkhalanskyjb:
@fzhinkin, hi! Are we doing this?

fzhinkin:
@dkhalanskyjb, sorry, I missed your question. Not now, I'm still about to figure out how to proceed with it for all our libraries.

README.md
fzhinkin: Okay, let's put the discussion about a recommended artifact name aside and solve the problem with auto-module names first.

I'll revert the README change.

fzhinkin:
The same issue existed for `kotlinx-coroutines-test`, specified auto-module-name explicitly for it too.

kotlinx-coroutines-test/build.gradle.kts
```diff
@@ -21,3 +21,9 @@ kotlin {
         }
     }
 }
+
+val allMetadataJar by tasks.getting(Jar::class) {
+    manifest {
+        attributes("Automatic-Module-Name" to "kotlinx.coroutines.test.artifact_disambiguating_module")
+    }
+}
```
dkhalanskyjb: Since this logic affects all multiplatform projects, would `kotlin-multiplatform-conventions` be the better place to put it in?

fzhinkin: Yep, makes total sense. Moved it there.

kotlinx-coroutines-core/build.gradle.kts
```diff
@@ -161,15 +161,16 @@ val jvmJar by tasks.getting(Jar::class) { setupManifest(this) }
...
-val allMetadataJar by tasks.getting(Jar::class) { setupManifest(this) }
+val allMetadataJar by tasks.getting(Jar::class) {
+    setupManifest(this)
+}
 
 fun setupManifest(jar: Jar) {
     jar.manifest {
         attributes(
-            mapOf(
+
```
dkhalanskyjb: Minor: an extra empty line.
In general, it looks like the changes to this file can be reverted now, can't they?

fzhinkin: Reverted

fzhinkin:
> [fzhinkin](https://github.com/fzhinkin) [force-pushed](https://github.com/Kotlin/kotlinx.coroutines/compare/3675e11214ddcce18ae250a30f661e16702f9c34..dccbea2065f6bb0912eb250b16d2aa75755003fd) the 3842-resolve-module-conflict-name  branch from [3675e11](https://github.com/Kotlin/kotlinx.coroutines/commit/3675e11214ddcce18ae250a30f661e16702f9c34) to [dccbea2](https://github.com/Kotlin/kotlinx.coroutines/commit/dccbea2065f6bb0912eb250b16d2aa75755003fd) [now](https://github.com/Kotlin/kotlinx.coroutines/pull/4320#event-24089122117)

^ rebased the branch to resolve conflicts with `develop`.

dkhalanskyjb:
Can this be merged?

PR #4321, opened 2025-01-05
Fix TestScope factory KDoc formatting

PhilGlass:
The [docs for the `TestScope` factory function](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-test-scope.html) are supposed to include a nested list, but the indentation is wrong so the KDoc is a bit confusing.

Before:

Screenshot 2025-01-05 at 13 10 15

After:

Screenshot 2025-01-05 at 13 10 22


dkhalanskyjb (approved):
Thanks! Could you target the `master` branch instead? From https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md :

> Documentation in markdown files can be updated directly in the master branch, unless the documentation is in the source code, and the patch changes line numbers.

PhilGlass:
@dkhalanskyjb Done. FWIW I read that section and wasn't sure whether it applied to KDoc changes - it mentions markdown _files_, but this PR touches a Kotlin file with some KDoc/markdown in it. Perhaps worth clarifying.

PR #4323, opened 2025-01-07
Optimize dumpCoroutinesInfoAsJsonAndReferences

mvicsokolova:
I've removed `trimIndent` and `toStringRepr` calls, made during the composition of coroutine dump json.

The motivation: Intellij IDEA Debugger supports the Coroutine View, which can show the hierarchy of all the available coroutines. This View struggled of sever performance issues ([IDEA-335303](https://youtrack.jetbrains.com/issue/IDEA-335303)). There were reasons for that on the debugger side, like fetching additional information for each coroutine. But other than that, the invocation of `dumpCoroutinesInfoAsJsonAndReferences` took significant amount of time.

I made some dumb performance measurements on the dumps of ~5000 coroutines, the absolute time is irrelevant ofc, just % of time of the parent call are:

1. Original version, most of time is taken by `trimIndent`:
Screenshot 2025-01-07 at 16 56 44

2. Without `trimIndent` invocation:
Screenshot 2025-01-07 at 16 57 09

3. Without `toStringRepr` invocation as well, it finally makes sense and shows `dumpCoroutineInfos` as the next significant call:
 
Screenshot 2025-01-07 at 16 50 38

It's not necessary to call `trimIndent` on every piece of this JSON, so that it could be parsed on the debugger side. So, I suggest to optimize this code.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
@@ -184,14 +184,14 @@ internal object DebugProbesImpl {
         val coroutinesInfoAsJson = ArrayList<String>(size)
         for (info in coroutinesInfo) {
             val context = info.context
-            val name = context[CoroutineName.Key]?.name
-            val dispatcher = context[CoroutineDispatcher.Key]
+            val name = context[CoroutineName.Key]?.name?.toStringRepr()
+            val dispatcher = context[CoroutineDispatcher.Key]?.toStringRepr()
             coroutinesInfoAsJson.add(
                 """
                 {
-                    "name": "$name",
+                    "name": $name,
                     "id": ${context[CoroutineId.Key]?.id},
-                    "dispatcher": "$dispatcher",
+                    "dispatcher": $dispatcher,
                     "sequenceNumber": ${info.sequenceNumber},
                     "state": "${info.state}"
                 } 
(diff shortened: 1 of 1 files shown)


fzhinkin:
@mvicsokolova, is there a way to improve `trimIndent` performance? It might be fine to abstain from calling it here, but I believe that it will be beneficial for all other projects relying on that function.

mvicsokolova:
Yes, I'll create a ticket for trimIndent optimization in kotlin-stdlib 👍🏼

mvicsokolova:
I'm not sure for now though, what's the better way to deal with special symbols in the coroutine name, than the current `toStringRepr`.

dkhalanskyjb:
Yes, `repr()` is needed to ensure that JSON is valid. Not all hope is lost, though. I'm not a performance optimization expert, but there are a few ideas I'd try.
* Usually, there are *very* few dispatchers in the program. This presents an optimization opportunity. If you cache the string representations of dispatchers in an `IdentityHashMap<CoroutineDispatcher, String>`, the number of `repr()` calls should drastically decrease (at least by about a half for a large number of *named* coroutines, but realistically, even more).
* If this doesn't help, maybe you could first check if the string contains any special symbols (99% of them won't), and if not, skip char-by-char string copying entirely.
* Do we need a pretty JSON at all? If we're not trimming indents, why not write the JSON string in a single line? The size of the resulting string would be literally half of what it is.

dkhalanskyjb (changes requested):
If we are going to merge this, at least `trimIndent`'s functionality should be recreated by deindenting the string (with an additional comment explaining that the abscence of `trimIndent` is intentional); otherwise, we get redundant whitespace in the result.

mvicsokolova:
I'll close this PR and provide the optimization for `dumpCoroutinesAsJson` in the separate one.

Issue #4322, opened 2025-01-06
Calling `Flow.stateIn` with a cancelled scope suspends forever


francescotescari:
**Describe the bug**

Calling `Flow.stateIn` with a cancelled scope suspends forever. The expected behavior IMHO is for `stateIn` to rethrow the cancellation exception of the scope, similar to how `scope.async { }.await()` and `CompleteableDeferred(scope.job)` behave on a cancelled scope. This behavior happens not only if the scope is already cancelled, but also if it gets cancelled concurrently with `stateIn`.

The cause of the issue is that `stateIn` awaits a `CompleteableDeferred` that is completed exclusively by a coroutine launched (non-atomically) in the collecting (possibly cancelled) scope. 

A possible fix is to bind the `CompletableDeferred` [here](https://github.com/Kotlin/kotlinx.coroutines/blob/ee92d16c4b48345648dcd8bb15f11ab9c3747f67/kotlinx-coroutines-core/common/src/flow/operators/Share.kt#L323) to the job of the collecting scope (with `CompleteableDeferred(scope.coroutineContext[Job])`). 

**Provide a Reproducer**

```kotlin
import kotlinx.coroutines.*
import kotlinx.coroutines.flow.*
import kotlin.coroutines.*

suspend fun main() {
    val flow = flowOf(1, 2, 3)
    val cancelledScope = CoroutineScope(EmptyCoroutineContext).apply { cancel() }
    println("Awaiting stateIn...")
    val stateFlow = flow.stateIn(cancelledScope) // Suspends forever
    println("Done!") // Never printed
}

// prints "Awaiting stateIn..." and hangs forever
```

dkhalanskyjb:
I agree both that this is a bug and that your fix is the way to go. Would you like to make a PR? If so, could you also make `stateIn` fail with a `NoSuchElementException("Expected at least one element")` if the flow collection completes without emitting a single element?

This was a surprisingly difficult decision. Below are my thoughts.

----

There's one more bug/feature/behavior of the same form:

```kotlin
flow { }.stateIn(CoroutineScope(EmptyCoroutineContext)) // hangs
```

We can think of `stateIn` as `shareIn` that also keeps track of the "current" emission (the same correspondence as `StateFlow` itself has with `SharedFlow`). Let's see how `shareIn` deals with these issues.

If the scope is cancelled when we attempt to wait for the initial value, it hangs:
```kotlin
[...]

PR #4327, opened 2025-01-10
fix: handle cancelled scope and empty flow in `Flow.stateIn`

francescotescari:
Fix for https://github.com/Kotlin/kotlinx.coroutines/issues/4322

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
@@ -320,7 +320,7 @@ public fun <T> Flow<T>.stateIn(
  */
 public suspend fun <T> Flow<T>.stateIn(scope: CoroutineScope): StateFlow<T> {
     val config = configureSharing(1)
-    val result = CompletableDeferred<StateFlow<T>>()
+    val result = CompletableDeferred<StateFlow<T>>(scope.coroutineContext[Job])
     scope.launchSharingDeferred(config.context, config.upstream, result)
     return result.await()
 }
--- a/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
@@ -317,6 +317,7 @@ public fun <T> Flow<T>.stateIn(
  * with multiple downstream subscribers. See the [StateFlow] documentation for the general concepts of state flows.
  *
  * @param scope the coroutine scope in which sharing is started.
+ * @throws NoSuchElementException if the upstream flow does not emit any value.
  */
 public suspend fun <T> Flow<T>.stateIn(scope: CoroutineScope): StateFlow<T> {
     val config = configureSharing(1)
@@ -339,6 +340,9 @@ private fun <T> CoroutineScope.launchSharingDeferred(
                         result.complete(ReadonlyStateFlow(it, coroutineContext.job))
                     }
                 }
+            }
+            if (state == null) {
+                result.completeExceptionally(NoSuchElementException("Flow is empty"))
             }
         } catch (e: Throwable) {
             // Notify the waiter that the flow has failed
--- a/kotlinx-coroutines-core/common/test/flow/sharing/StateInTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/sharing/StateInTest.kt
@@ -98,4 +98,12 @@ class StateInTest : TestBase() {
             flow.stateIn(cancelledScope)
         }
     }
+
+    @Test
+    fun testThrowsNoSuchElementExceptionOnEmptyFlow() = runTest {
+        val flow = emptyFlow<Any>()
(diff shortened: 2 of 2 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Share.kt
```diff
@@ -340,6 +341,9 @@ private fun <T> CoroutineScope.launchSharingDeferred(
                     }
                 }
             }
+            if (state == null) {
+                result.completeExceptionally(NoSuchElementException("Flow is empty"))
```
francescotescari, line 345: Notice: I think we shouldn't cancel the collecting scope here, as there is nothing wrong with the flow collection itself, the exception should be limited to `stateIn` IMHO.

dkhalanskyjb: Yes, good point, let's not cancel the collecting scope. This is tricky to do ensure cleanly, though! Here's a slightly convoluted solution:
```kotlin
    /* If `scope`'s job is cancelled, both `deferredParentJob` and the deferred will be cancelled.
    If the deferred is cancelled, this will not cancel `deferredParentJob` (as it's a `SupervisorJob`),
    and `CoroutineExceptionHandler` present in `scope` (if any) won't get invoked,
    as `Deferred` values are themselves responsible for propagating their exceptions. */
    val deferredParentJob = SupervisorJob(scope.coroutineContext[Job])
    val result = CompletableDeferred<StateFlow<T>>(deferredParentJob)
    scope.launchSharingDeferred(config.context, config.upstream, result)
    return try {
        result.await()
    } finally {
        deferredParentJob.cancel()
    }
```
then the modified test is:
```kotlin
    @Test
    fun testThrowsNoSuchElementExceptionOnEmptyFlow() = runTest {
        val flow = emptyFlow<Any>()
        assertFailsWith<NoSuchElementException> {
            flow.stateIn(this)
        }
        assertEquals(true, coroutineContext[Job]?.isActive)
        coroutineContext.cancelChildren()
    }
```

(Note this `runTest` is not `kotlinx.coroutines.test.runTest`!) If the main coroutine of `runTest` gets cancelled, the whole test fails; also, if there are any uncaught exceptions in child coroutines of `runTest`, they do get reported at the end of the test. So, this test does check that `NoSuchElementException` doesn't propagate upwards.

The only other option I see is to rely on our internal low-level `job.invokeOnCompletion` function to avoid the parent-child relationship between `scope` and `result` altogether. After implementing that option, I don't like how it recreates parts of child cancellation logic. This looks like a weird micro-optimization.

francescotescari: Another simpler (not elegant) trick is to make `CompletableDeferred` of type `CompletableDeferred<Result<StateFlow<T>>>` and complete the deferred successfully but with `Result.failure` and unpack with `getOrThrow` it `stateIn` side.

dkhalanskyjb: Good idea!

francescotescari: Sorry for the delay, addressed it now. 
LMK if it looks good to you 🙏

kotlinx-coroutines-core/common/test/flow/sharing/StateInTest.kt
```diff
@@ -88,4 +89,23 @@ class StateInTest : TestBase() {
...
+    }
+
+    @Test
+    fun testThrowsNoSuchElementExceptionOnEmptyFlow() = runTest {
+        val flow = emptyFlow<Any>()
+        assertFailsWith<NoSuchElementException> {
+            flow.stateIn(this)
+        }
+        // Ensure that the collecting scope is not cancelled by the NoSuchElementException
+        assertEquals(true, coroutineContext[Job]?.isActive)
```
dkhalanskyjb, line 109:
```suggestion
        assertTrue(coroutineContext[Job]?.isActive)
```

kotlinx-coroutines-core/common/test/flow/sharing/StateInTest.kt
```diff
@@ -88,4 +89,23 @@ class StateInTest : TestBase() {
...
+    }
+
+    @Test
+    fun testThrowsNoSuchElementExceptionOnEmptyFlow() = runTest {
+        val flow = emptyFlow<Any>()
+        assertFailsWith<NoSuchElementException> {
+            flow.stateIn(this)
+        }
+        // Ensure that the collecting scope is not cancelled by the NoSuchElementException
+        assertTrue(coroutineContext[Job]?.isActive)
```
dkhalanskyjb, line 109:
```suggestion
        assertEquals(true, coroutineContext[Job]?.isActive)
```

dkhalanskyjb:
Thank you both for the bug report and a well-thought-out fix!

PR #4277, opened 2024-11-21
Fix DefaultExecutor having a separate thread and being used for cleanup work and running Dispatchers.Unconfined

dkhalanskyjb:
Fixes #4262

PR #4330, opened 2025-01-14
Reach target parallelism in .limitedParallelism of failing dispatchers

dkhalanskyjb:
A tiny self-contained piece of https://github.com/Kotlin/kotlinx.coroutines/pull/4277

### DIFF
--- a/kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
+++ b/kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
@@ -117,21 +117,29 @@ internal class LimitedDispatcher(
      */
     private inner class Worker(private var currentTask: Runnable) : Runnable {
         override fun run() {
-            var fairnessCounter = 0
-            while (true) {
-                try {
-                    currentTask.run()
-                } catch (e: Throwable) {
-                    handleCoroutineException(EmptyCoroutineContext, e)
+            try {
+                var fairnessCounter = 0
+                while (true) {
+                    try {
+                        currentTask.run()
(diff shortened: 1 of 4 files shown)


qwwdfsad (approved):

kotlinx-coroutines-core/common/src/internal/LimitedDispatcher.kt
```diff
@@ -65,7 +65,17 @@ internal class LimitedDispatcher(
...
+        try {
+            startWorker(Worker(task))
+        } catch (e: Throwable) {
+            /* If we failed to start a worker, we should decrement the counter.
+            The queue is in an inconsistent state--it's non-empty despite the target parallelism not having been
+            reached--but at least a properly functioning worker will have a chance to correct this if some future
+            dispatch does succeed.
+            If we don't decrement the counter, it will be impossible to ever reach the target parallelism again. */
+            runningWorkers.decrementAndGet()
+            throw e
```
JakeWharton, line 77: This seems to basically be a `finally` block. Any reason not to use that, instead of `catch`+re`throw`?

dkhalanskyjb: The reason to use `catch`/`throw` is that we don't want to deallocate the worker if `startWorker` finishes normally, as this means the worker is actually started. The same holds true below: if `return` is used, this means we've already deallocated the worker and don't need to do that again.

JakeWharton: Ah, right right. Thanks for clarifying.

PR #4331, opened 2025-01-14
Introduce guidance functions explaining 'launch' and 'async' usage

dkhalanskyjb:
Fixes KTIJ-13441


qwwdfsad (approved):
That's a very nice addition!

I am a bit concerned whether it might somehow obstruct refactorings and/or typing flow, maybe it's worth passing it through the RC cycle

PR #4336, opened 2025-01-17
Restore allWarningsAsErrors in main source-sets

qwwdfsad:


kotlinx-coroutines-core/native/src/Builders.kt
```diff
@@ -64,7 +66,7 @@ public actual fun <T> runBlocking(context: CoroutineContext, block: suspend Coro
     var completed = false
     ThreadLocalKeepAlive.addCheck { !completed }
     try {
-        @Suppress("LEAKED_IN_PLACE_LAMBDA") // Contract is preserved, invoked immediately or throws
+
```
dkhalanskyjb, line 69: Is this newline a typo?

reactive/kotlinx-coroutines-reactive/src/Publish.kt
```diff
@@ -316,6 +316,7 @@ public class PublisherCoroutine<in T>(
         signalCompleted(cause, handled)
     }
 
+    @Deprecated("Since 1.2.0, binary compatibility with versions <= 1.1.x", level = DeprecationLevel.HIDDEN)
```
dkhalanskyjb, line 319: Binary compatibility is not the only reason why we have this method. If we remove it, the `Subscription` interface is no longer implemented. In fact, I'd argue it's a bug that this code causes a warning: https://youtrack.jetbrains.com/issue/KT-74697/Overriding-a-method-thats-both-deprecated-and-non-deprecated-should-not-cause-warnings

kotlinx-coroutines-core/native/src/Builders.kt
qwwdfsad: Yep, fixed

reactive/kotlinx-coroutines-reactive/src/Publish.kt
qwwdfsad: Nice catch!

I propose to leave it as is here because for users everything keeps working (i.e. `PublisherCoroutine,cancel` is not exposed) and we do not provide any new documentation here

dkhalanskyjb: I don't object to `@Deprecated`, as long as the error message honestly states the reason and doesn't mislead.

dkhalanskyjb: Now that I think about it more, `@Deprecated` is slightly worse than a suppression, because if https://youtrack.jetbrains.com/issue/KT-74697 does get fixed, we'll be notified about it via the "redundant suppression" message, whereas the deprecation is with us to stay.

dkhalanskyjb (changes requested):

dkhalanskyjb (approved):

PR #4344, opened 2025-02-05
Simplify kotlinx-coroutines-core build script

abdulowork:
This split in source sets and compilations seems to have been used to test that the core can compile against the 1.6 JVM target, but the coroutines jvm jar ships class files for jvm 1.8+ since 1.5.0.

We need this cleanup to promote KotlinCompilation.source API to an error deprecation.

I tested manually that there is no impact on the resulting jar:

Screenshot 2025-02-05 at 1 31 17 PM


dkhalanskyjb:
Cool, thanks! I didn't know Kotlin builds were reproducible!

PR #4353, opened 2025-02-12
Build setup cleanup

qwwdfsad:


buildSrc/src/main/kotlin/atomicfu-conventions.gradle.kts
```diff
@@ -1,8 +0,0 @@
-plugins {
-    id("org.jetbrains.kotlinx.atomicfu")
-}
-
-// Workaround for KT-71203. Can be removed after https://github.com/Kotlin/kotlinx-atomicfu/issues/431
```
dkhalanskyjb, line 5: I was surprised to see this workaround removed, because the linked issue was not closed, but it looks like the workaround is not needed anyway: https://youtrack.jetbrains.com/issue/KT-71203/kotlinx-coroutines-core-compilation-fails-with-the-latest-2.1.-compiler#focus=Comments-27-11140112.0-0

dkhalanskyjb (approved):

PR #4361, opened 2025-02-21
Enable '-jvm-default=disable' explicitly to prevent API dump changes

udalov:
The default mode is changing from 'disable' to 'enable' in KT-71768. However, core libraries will migrate to the new '-jvm-default=enable' mode explicitly, once they update to Kotlin 2.2, see KT-72051.

### DIFF
--- a/buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
@@ -15,7 +15,10 @@ kotlin {
     jvm {
         compilations.all {
             compileTaskProvider.configure {
-                compilerOptions.jvmTarget = JvmTarget.JVM_1_8
+                compilerOptions {
+                    jvmTarget = JvmTarget.JVM_1_8
+                    freeCompilerArgs.addAll("-jvm-default=disable")
+                }
             }
         }
     }


udalov:
This should not be merged earlier than [this change](https://kotlin.jetbrains.space/p/kotlin/reviews/1393/timeline) that resolves [KT-71768](https://youtrack.jetbrains.com/issue/KT-71768).

All changes in user projects:
* kotlinx.collections.immutable: https://github.com/Kotlin/kotlinx.collections.immutable/pull/207
* kotlinx.coroutines: https://github.com/Kotlin/kotlinx.coroutines/pull/4361
* kotlinx-datetime: https://github.com/Kotlin/kotlinx-datetime/pull/492
* Dokka: https://code.jetbrains.team/p/kct/repositories/dokka_kct/reviews/25/timeline

udalov:
@dkhalanskyjb Please approve so that I can merge it together with the [MR](https://kotlin.jetbrains.space/p/kotlin/reviews/1393/timeline) in Kotlin.

dkhalanskyjb (approved):
If this is just a formality, sure. If not, I don't understand why not just introduce API dump changes, given that they look benign.

udalov:
Sure, I just thought that keeping the `kotlin-community/k2/dev` branch as close to `develop` as possible is convenient.

When the time comes to update the compiler dependency in the libraries to 2.2, you will either need to update API dumps (preferably), or enable `-jvm-default=disable` in the same way (in case there are any problems).

PR #4365, opened 2025-02-25
Fix the incorrect documentation for Flow.conflate

### LINKED ISSUE #4352: ChannelFlow instances do not prefer conflation over buffering when fused
**Describe the bug**

The [`Flow.conflate()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/conflate.html) documentation states that: 

> Adjacent applications of conflate/[buffer](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/buffer.html), [channelFlow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/channel-flow.html), [flowOn](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flow-on.html) and [produceIn](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/produce-in.html) are always fused so that only one properly configured channel is used for execution. **Conflation takes precedence over buffer() calls with any other capacity.**

This is not the behavior I've observed when using a callbackFlow within my library. 

While working on a library I publish, I wanted to ensure that a flow I return from my public API is always conflated. It does not make sense for events to be buffered because, even though I was using a `callbackFlow`, the items within that flow describe _state_ and not _events_. I thought I had achieved this behavior by applying the `Flow.conflate` operator to the flow I return, however I was able to disprove that this unconditionally always configures the underlying channel to be conflated in some unit tests. In my experience, adjacent applications of the `Flow.buffer` operator would reconfigure the underlying channel's capacity, but _would not_ reconfigure the buffer overflow strategy. 

Instead, I would expect calls to `Flow.conflate` to _always_ take precedence over other fused operators as the documentation leads me to believe.  

I am hoping that this isn't just treated as unclear documentation. From my perspective, I don't see much utility in allowing Flow consumers to buffer a flow which has previously been conflated. 

**Provide a Reproducer**

```kt
// The flow that's returned from my public API. 
val original = channelFlow { awaitCancellation() }.conflate()

// The flow that's altered by the API consumer. 
val altered = original.buffer(3, onBufferOverflow = BufferOverflow.SUSPEND)
```

The above snippet will configure the underlying channel to have a capacity of three and a buffer overflow strategy of `DROP_OLDEST`. It seems like partial retention of the conflated policy is successful since `DROP_OLDEST` is retained, however the final configured capacity is inconsistent with what's documented.

dkhalanskyjb:
Fixes #4352

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Context.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Context.kt
@@ -82,7 +82,7 @@ import kotlin.jvm.*
  * which effectively requests a buffer of any size. Multiple requests with a specified buffer
  * size produce a buffer with the sum of the requested buffer sizes.
  *
- * A `buffer` call with a non-default value of the [onBufferOverflow] parameter overrides all immediately preceding
+ * A `buffer` call with a non-[SUSPEND] value of the [onBufferOverflow] parameter overrides all immediately preceding
  * buffering operators, because it never suspends its upstream, and thus no upstream buffer would ever be used.
  *
  * ### Conceptual implementation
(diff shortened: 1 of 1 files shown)


qwwdfsad (approved):

dkhalanskyjb:
The flaky test is going to be fixed in an upcoming PR.

PR #4370, opened 2025-03-03
Review branch for #4208

dkhalanskyjb:
To simplify reviewing https://github.com/Kotlin/kotlinx.coroutines/pull/4208, I've rebased the changes provided there into three clearly separated commits. After this PR gets approved, the changes introduced here will be backported to #4208, and we'll merge that.


kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -206,10 +207,124 @@ private class LazyStandaloneCoroutine(
...
+     * when either the corresponding thread is GC'ed or it cleans up its stale entries on other TL accesses.
+     * When such coroutines are promoted to old generation, `ThreadLocalMap`s become bloated and an arbitrary accesses to thread locals
+     * start to consume significant amount of CPU because these maps are open-addressed and cleaned up incrementally on each access.
+     * (You can read more about this effect as "GC nepotism").
+     *
+     * To avoid that, we attempt to narrow down the lifetime of this thread local as much as possible:
+     * - It's never accessed when we are sure there are no thread context elements
+     * - It's cleaned up via [ThreadLocal.remove] as soon as the coroutine is suspended or finished.
+     */
+    private val threadStateToRecover = commonThreadLocal<Pair<CoroutineContext, Any?>?>(Symbol("UndispatchedCoroutine"))
```
qwwdfsad, line 247: Is it a legal pattern to use in this place?
K/N thread locals are name-bound, which means that one `UndispatchedCoroutine` can overwrite the other `UndispatchedCoroutine` thread local.

So something like 
```
withContext(..forcePopulationOfThreadStateRecover...) {
    withContext(..doItAgain..) {}
}
```
might actually overwrite the state. Is it the case or am I missing something?

qwwdfsad (commented):
Tracking the history of changes would be hard, taking that git considers all that as removal-addition of files. Not sure if we can do anything with that though

dkhalanskyjb:
A heads-up: after looking at @qwwdfsad's comment, I was wondering why no test caught this problem if it's a real one and went to investigate. I've noticed that some tests were not ported from the JVM to other targets and decided to try to move them. However, seemingly unrelated tests failed: for some reason, outside the JVM, thread context elements don't get cleaned up.

This fully deterministic, single-threaded test fails on Native, but passes on the JVM:
```kotlin
class Oh {
    @Test
    fun smokeTest() {
        runBlocking {
            withContext(C()) {
                yield()
            }
            assertEquals(2, restores)
        }
    }

    private var restores = 0

    inner class C(): ThreadContextElement<Unit>, CoroutineContext.Key<C> {
        override fun updateThreadContext(context: CoroutineContext) {
        }

        override fun restoreThreadContext(context: CoroutineContext, oldState: Unit) {
            restores++
        }

        override val key: CoroutineContext.Key<*>
            get() = this
    }
}
```
Curiously, I don't think it's the problem reported by @qwwdfsad, as this persists even when I change the implementation of thread-local variables on Native to account for the problem. I don't yet understand what's going on here.

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -206,10 +207,124 @@ private class LazyStandaloneCoroutine(
...
+     *     // COROUTINE_SUSPENDED is returned
+     * } finally {
+     *     thisCoroutine().clearThreadContext() // Concurrently the "smth" could've been already resumed on a different thread
+     *     // and it also calls saveThreadContext and clearThreadContext
+     * }
+     * ```
+     *
+     * Usage note:
+     *
+     * This part of the code is performance-sensitive.
```
qwwdfsad, line 232: Worth noting that it's JVM-specific

domgew:
@dkhalanskyjb @qwwdfsad @zuevmaxim 
Is there any ETA for when this could be available?

murfel:
@domgew definitely not in the next month, more likely in a while, I'm afraid.

whyoleg:
Hey! Are there any plans to finalize this PR? Is any help needed? :)

dkhalanskyjb:
@whyoleg, yes, there are plans to finalize this PR. What's holding it back is a subtle dependency on the JVM-specific behavior in the code of `ThreadContextElement`. Fixing that requires spending several days on rewriting some code in the internal coroutines machinery. So far, an opportunity to do such a deep dive without being sidetracked by the more urgent tasks has not presented itself. Since this task requires knowledge of the `kotlinx.coroutines` internals, I don't expect outside help to arrive here. That said, if someone is feeling ambitious, we will gladly accept the contribution.

domgew:
Are there any news regarding this PR or the feature itself?

It would be quite useful for things like user context, localisation, structured logging (like SLF4J's MDC), ...

PR #4371, opened 2025-03-05
Use -Xjvm-default instead of -jvm-default

udalov:
Apparently this branch is used in user projects to build both 2.1.20 and 2.2.0, and 2.1.20 did not support the new `-jvm-default` (KT-73007).


udalov:
OK, I was told that generally no review is expected for changes into `kotlin-community/k2/dev`, so I'm going to merge it as is. ¯\\_(ツ)_/¯

PR #4372, opened 2025-03-05
Enable '-Xjvm-default=disable' explicitly to prevent API dump changes

udalov:
The default mode is changing from 'disable' to 'enable' in KT-71768. However, core libraries will migrate to the new '-jvm-default=enable' mode explicitly, once they update to Kotlin 2.2, see KT-72051.


udalov:
This change is similar to https://github.com/Kotlin/kotlinx.coroutines/pull/4361, except that it's against the main dev branch (which is apparently easier for everyone in terms of user project / release train maintenance) and it uses `-Xjvm-default` so that the project will be compilable against both 2.1.20 and 2.2.0 with the same resulting binaries.

PR #4377, opened 2025-03-08
fix for image width

LiYing2010:


docs/topics/debug-coroutines-with-idea.md
```diff
@@ -108,12 +108,12 @@ Using IntelliJ IDEA debugger, you can dig deeper into each coroutine to debug yo
 
 If you use `suspend` functions, in the debugger, you might see the "was optimized out" text next to a variable's name:
 
-![Variable "a" was optimized out](variable-optimised-out.png)
+![Variable "a" was optimized out](variable-optimised-out.png){width=480}
```
LiYing2010, line 111: JFYI
this image is a little too big in html page
so I think it will be better to change its width

koshachy (approved):
LGTM

PR #4380, opened 2025-03-09
Add StateFlow<T>.onSubscription

### LINKED ISSUE #4275: Add StateFlow<T>.onSubscription
**Use case**

The current `SharedFlow<T>.onSubscription` implementation returns a `SharedFlow<T>`, which means losing type information when calling it from a `StateFlow<T>`. It would be nice to have a `StateFlow<T>.onSubscription` counterpart for `StateFlow<T>`, thus avoiding the aforementioned issue.

**The Shape of the API**

Unless there are implementation details I'm unaware of, the API could be copy-pasted from the `SharedFlow<T>.onSubscription` implementation and adapted to work with `StateFlow<T>`:

```kotlin
public fun <T> StateFlow<T>.onSubscription(action: suspend FlowCollector<T>.() -> Unit): StateFlow<T> =
    SubscribedStateFlow(this, action)

@OptIn(ExperimentalForInheritanceCoroutinesApi::class)
private class SubscribedStateFlow<T>(
    private val stateFlow: StateFlow<T>,
    private val action: suspend FlowCollector<T>.() -> Unit
) : StateFlow<T> by stateFlow {
    override suspend fun collect(collector: FlowCollector<T>) =
        stateFlow.collect(SubscribedFlowCollector(collector, action))
}
```

**Prior Art**

- [SharedFlow<T>.onSubscription](https://github.com/Kotlin/kotlinx.coroutines/blob/d8d6f8f37978b8e202d93b34f23f101df9c5724d/kotlinx-coroutines-core/common/src/flow/operators/Share.kt#L399)
- [SubscribedSharedFlow](https://github.com/Kotlin/kotlinx.coroutines/blob/d8d6f8f37978b8e202d93b34f23f101df9c5724d/kotlinx-coroutines-core/common/src/flow/operators/Share.kt#L403)

xit0c:
Fixes #4275


kotlinx-coroutines-core/common/src/flow/operators/Share.kt
```diff
@@ -412,6 +412,29 @@ private class SubscribedSharedFlow<T>(
...
+ * (after the subscription is registered).
+ *
+ * The [action] is called before any value is emitted from the upstream
+ * flow to this subscription but after the subscription is established. It is guaranteed that all emissions to
+ * the upstream flow that happen inside or immediately after this `onSubscription` action will be
+ * collected by this subscription.
+ *
+ * The receiver of the [action] is [FlowCollector], so `onSubscription` can emit additional elements.
+ */
+public fun <T> StateFlow<T>.onSubscription(action: suspend FlowCollector<T>.() -> Unit): StateFlow<T> =
```
dkhalanskyjb, line 426: Please also add tests for this. Good inspirations are `SharedFlowTest#testOnSubscription` and `SharedFlowTest#onSubscriptionThrows`, which can be added to `StateFlowTest`. `testOnSubscriptionWithException` is quite complex and doesn't highlight the basic concepts of `onSubscription`, even though it probably tests them well.

dkhalanskyjb, line 426: Please also add `@ExperimentalCoroutinesApi`. We can be fairly certain that this API is benign, but still, better safe than sorry, and also, this gives us the formal permission to publish this in a minor release if needed.

whyoleg: JIC: if `@ExperimentalCoroutinesApi` will be added, then current users who use `SharedFlow.onSubscription` on a variables of type `StateFlow` will start to receive an opt-in warning without a nice ability to use previous overload.

dkhalanskyjb: They could do `.let { it as SharedFlow<T> }.onSubscription`. You raise a good point, though!

dkhalanskyjb (approved):
Well done, thank you!

PR #4388, opened 2025-03-21
Specify explicit return types for some public API functions

KvanTTT:


KvanTTT:
K1 also reports `NO_EXPLICIT_RETURN_TYPE_IN_API_MODE` on a code like that:

```kt
@Deprecated(
    level = DeprecationLevel.ERROR,
    message = "This function is prohibited",
)
public fun Int.<!NO_EXPLICIT_RETURN_TYPE_IN_API_MODE!>deprecatedFunction<!>() = 42
```

If skip the reporting for `DeprecationLevel.ERROR`, than it's a breaking change that it's better to avoid.

dkhalanskyjb:
But we are using K2 already: https://github.com/Kotlin/kotlinx.coroutines/blob/465e29d325841244f3a1aac2e13073bc965f9736/gradle.properties#L4

KvanTTT:
I understand. I mean I fixed a bug with missing `NO_EXPLICIT_RETURN_TYPE_IN_API_MODE` on extension members (see https://youtrack.jetbrains.com/issue/KT-74325/Explicit-API-mode-does-not-enforce-explicit-return-types-for-extension-properties) and it caused the compiler failure on this project because `NO_EXPLICIT_RETURN_TYPE_IN_API_MODE` is reported on extensions as it should be.

After that, I checked if `NO_EXPLICIT_RETURN_TYPE_IN_API_MODE` is reported on members marked by `DerprecationLevel.ERROR` in K1 and it's reported. Currently I would like to avoid breaking this behavior because it could turn some red code into green. Also, I'm not sure if `NO_EXPLICIT_RETURN_TYPE_IN_API_MODE` should consider annotations at all, the return type is missing in public API even despite the annotation.

dkhalanskyjb (approved):
Shouldn't this PR target `develop`? `kotlin-community/k2/dev` doesn't need any approval from the library maintainers.

KvanTTT:
Ok, if I can merged to `kotlin-community/k2/dev` without approvement, I can also create a similar MR to `develop`.

woainikk (approved):

PR #4392, opened 2025-03-24
Fulfill the new requirements for Kotlin User Projects

dkhalanskyjb:


woainikk (approved):

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
```diff
@@ -147,3 +148,36 @@ fun shouldUseLocalMaven(project: Project): Boolean {
     }
     return hasSnapshotDependency || isSnapshotTrainEnabled(project)
 }
+
+/**
+ * Returns a non-null value if the CI needs to override the default behavior of treating warnings as errors.
+ * Then, `true` means that warnings should be treated as errors, `false` means that they should not.
```
hrach, line 154:
```suggestion
 * Then, `enable` means that warnings should be treated as errors, `disable` means that they should not.
```

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
```diff
@@ -147,3 +148,36 @@ fun shouldUseLocalMaven(project: Project): Boolean {
...
+ */
+private fun warningsAreErrorsOverride(project: Project): Boolean? =
+    when (val prop = project.rootProject.properties["kotlin_Werror_override"] as? String) {
```
osipxd, line 157: Probably, should be used API compatible with isolated projects?
```suggestion
    when (val prop = project.providers.gradleProperty("kotlin_Werror_override").orNull) {
```

osipxd: Sure!
Calling `Project.properties` is quite expensive because it collects not only properties from `gradle.properties` and CLI arguments, but also tasks and their fields, extensions, extras, conventions, etc (see https://github.com/gradle/gradle/issues/25053)
`getProperty` and `hasProperty` are better, but they're not compatible with [isolated projects](https://docs.gradle.org/current/userguide/isolated_projects.html#build_logic_constraints) because of the access to `properties` of another project.

Since you're getting the property from the root project, it should be safe to use `providers.gradleProperty` which looks for Gradle properties in the root project. As a bonus, this API is compatible with isolated projects (which I hope will be stable someday) as there is no need in calling methods of the `rootProject`

osipxd: However, as soon as this syntax is used everywhere in this file, this change is indeed out of the scope of the current PR.

woainikk:
@dkhalanskyjb can you please also like this?: 
Implement a Gradle property for passing additional CLI options to the Kotlin compiler
(property naming is not critical but kotlin_additional_cli_options is suggested for consistency). And the passed kotlin_additional_cli_options should be applied as compiler options

dkhalanskyjb:
@woainikk, are there examples (either of other libraries doing this or of this property being used)? What exactly is the content of the property? Is it a space-separated list of strings? A comma-separated list? Something else? Should the options go before or after the usual ones (if that's important)?

woainikk:
@dkhalanskyjb Can you please also add logging to somewhere, to all applied compiler flags were printed to log?

qwwdfsad (approved):
Let's figure overridable `-Werror` separately

PR #4416, opened 2025-04-21
docs: add example for `runTest` to reiterate that runTest waits for children to finish

murfel:
Two previous examples (L63-84) could be misinterpreted as if there's a need to explicitly `join` or generally wait for children coroutines to finish, since they don't do anything after joining the launched coroutine.

The misleading examples (L63-84):
```
 * @Test
 * fun exampleWaitingForAsyncTasks1() = runTest {
 *     // 1
 *     val job = launch {
 *         // 3
 *     }
 *     // 2
 *     job.join() // the main test coroutine suspends here, so the child is executed
 *     // 4
 * }
 *
 * @Test
 * fun exampleWaitingForAsyncTasks2() = runTest {
 *     // 1
 *     launch {
 *         // 3
 *     }
 *     // 2
 *     testScheduler.advanceUntilIdle() // runs the tasks until their queue is empty
 *     // 4
 * }
 * ```

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -85,7 +85,6 @@ public expect class TestResult
  * }
  * ```
  *
- *
  * If the results of children coroutines computations are not needed in the runTest scope,
  * there is no need to wait for children coroutines to finish, they are awaited for automatically
  * by runTest parent coroutine.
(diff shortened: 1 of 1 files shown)


dkhalanskyjb (changes requested):
Makes sense, thanks. Please also add the same piece of documentation to the other `runTest` overload in the same file for consistency.

PR #4417, opened 2025-04-21
docs: fix example for channel, do not assume a single-threaded dispatcher

murfel:
Ditto

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -87,6 +87,7 @@ public interface SendChannel<in E> {
      * ```
      * val channel = Channel<Int>()
      * launch {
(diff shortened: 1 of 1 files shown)


dkhalanskyjb (changes requested):
The example could indeed benefit from more comments. To answer your question: it is guaranteed in all scenarios, because the default argument for the channel's buffer size is `RENDEZVOUS`, which means no buffer gets created. Could you add something like `// a channel without a buffer` to the code?

I don't understand how `delay(100)` helps, though.

murfel:
Discussed that _suspends_ in the docs should be interpreted as _may or may not suspend, but is a suspension point_, and the user shouldn't think or base their code on whether the function will actually suspend, hence it is counterproductive to sprinkle the doc snippets with `yield()`.

PR #4421, opened 2025-04-23
Use providers.gradle property instead of rootProject.properties in the build scripts

dkhalanskyjb:
See <https://github.com/Kotlin/kotlinx.coroutines/pull/4392/files#r2016110128>
for an explanation of the benefits of this.

Additionally, the `integration-testing` no longer sees the `kotlinx.coroutines` having a `SNAPSHOT` version (which is always the case) as a valid reason to start looking at development repositories for packages.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,7 +40,7 @@ allprojects {
     if (deployVersion != null) version = deployVersion
 
     if (isSnapshotTrainEnabled(rootProject)) {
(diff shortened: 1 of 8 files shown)


qwwdfsad (approved):
Nice one! Though I must admit, it hurts to read how `Properties.getProperties()` is implemented and how exactly `gradleProperty` helps

qwwdfsad:
image

Why would they change properties? Properties are properties!

Issue #4205, opened 2024-08-05
Some problem about `addLast` in LockFreeTaskQueue


autorain:
```kotlin
fun addLast(element: E): Int {
        _state.loop { state ->
            if (state and (FROZEN_MASK or CLOSED_MASK) != 0L) return state.addFailReason() // cannot add
            state.withState { head, tail ->
                val mask = this.mask // manually move instance field to local for performance
                // If queue is Single-Consumer then there could be one element beyond head that we cannot overwrite,
                // so we check for full queue with an extra margin of one element
                if ((tail + 2) and mask == head and mask) return ADD_FROZEN // overfull, so do freeze & copy
                // If queue is Multi-Consumer then the consumer could still have not cleared element
                // despite the above check for one free slot.
                if (!singleConsumer && array[tail and mask].value != null) {
                    // There are two options in this situation
                    // 1. Spin-wait until consumer clears the slot
                    // 2. Freeze & resize to avoid spinning
                    // We use heuristic here to avoid memory-overallocation
                    // Freeze & reallocate when queue is small or more than half of the queue is used
                    if (capacity < MIN_ADD_SPIN_CAPACITY || (tail - head) and MAX_CAPACITY_MASK > capacity shr 1) {
                        return ADD_FROZEN
                    }
                    // otherwise spin
                    return@loop
                }
                val newTail = (tail + 1) and MAX_CAPACITY_MASK
                if (_state.compareAndSet(state, state.updateTail(newTail))) {
                    // successfully added
                    array[tail and mask].value = element
                    // could have been frozen & copied before this item was set -- correct it by filling placeholder
                    var cur = this
                    while(true) {
                        if (cur._state.value and FROZEN_MASK == 0L) break // all fine -- not frozen yet
                        cur = cur.next().fillPlaceholder(tail, element) ?: break
                    }
                    return ADD_SUCCESS // added successfully
                }
            }
        }
    }
```
Why use `(tail + 2) and mask == head and mask` not `(tail + 1) and mask == head and mask` ? Could some kind person explain it to me? Thanks a lot.

qwwdfsad:
Hey, there is no real reason.
This `+2` was relevant in the earlier version of LFQ where instead of arraylets, a two-phased copy was used, and there was an additional invariant (the deque operation was two-phased, and there was an intermediate `REMOVE` state).

[...]

PR #4426, opened 2025-04-23
Simplify full-queue check to use single margin

PPLong222:
Update outdated full-queue check logic of `LockFreeTaskQueueCore`. No more need to have 2 margin elements; one is sufficient.

Tested in `LockFreeTaskQueueStressTest` and `LockFreeTaskQueueTest`

See also #4205


PPLong222:
Thanks a lot for taking the time to reply!

I realise this change is quite subtle and that getting to the bottom of it means digging through git history and the queue’s invariants, so please feel free to review it whenever you have the bandwidth. From what I can tell, the patch shouldn’t affect performance or observable behaviour much—it’s mainly about **correctness semantics** and **avoiding confusion** for future readers(or maybe I just have misunderstood something).

The motivation came while I was reading `LockFreeTaskQueueCore`: I couldn’t understand why the full-queue check uses two extra slots instead of the usual single margin you see in a circular buffer. After spending a day tracing the code and trying a variety of stress-test cases, I still couldn’t find a situation where one margin element isn’t enough. 

If I’ve overlooked something, just let me know and I’ll be happy to revise (or close) the PR. Thanks again, and take all the time you need!

PR #4427, opened 2025-04-28
Kotlin UP: Fulfill the new requirements for Kotlin User Projects (#4392)

woainikk:


woainikk:
^^ small fixes is needed, because parameters will be passed in brackets.

dkhalanskyjb:
Note: it's not obvious at this point that this fix is actually necessary; we are discussing this internally.

Issue #4422, opened 2025-04-23
Publisher.asFlow() sometimes fails to propagate thread context elements


dkhalanskyjb:
This issue was discovered analytically, by reading the source code. No one reported it yet.
```kotlin
val threadLocal = ThreadLocal<String>.withInitial { "default" }

@Test
fun testFlowOnThreadContext() = runBlocking {
    val publisher = Publisher<Int> { it ->
        it.onSubscribe(object : Subscription {
            override fun request(n: Long) {
                assertEquals("value", threadLocal.get())
                it.onNext(1)
                it.onComplete()
            }

            override fun cancel() {}
        })
    }

    val context1 = Dispatchers.IO + threadLocal.asContextElement("value")
    val context2 = threadLocal.asContextElement("value")

    // succeeds
    publisher.asFlow().flowOn(context1).collect { }
    // fails with org.junit.ComparisonFailure: expected:<[value]> but was:<[default]>
    publisher.asFlow().flowOn(context2).collect { }
}
```

Issue #4403, opened 2025-04-02
ChannelFlow with unfused flowOn does not update/restore ThreadContextElement after collect


Zincfox:
**Describe the bug**

When collecting a `channelFlow` that was `flowOn`-d a `ThreadContextElement`, while preventing fusion, the associated `ThreadContextElement` is not always properly restored afterwards until the next suspension point.

The specific conditions seem to be:
- Use of `channelFlow {}` (normal `flow {}` works as expected)
- Map the `ChannelFlow` to prevent fusion (works with fusion as expected)
- Use `flowOn` on the resulting mapped `ChannelFlow`
- After `collect {}`ing, check before the next suspension point (tested by adding `yield()` before check, which updates as expected during suspension)

I could not find a matching issue here besides potentially #4121, which I admittedly did not fully understand, but I believe the presence of a `ThreadContextElement` in the parent `CoroutineContext`, bundled with the fact that this does not occur when using a normal `Flow` instead of a `ChannelFlow` or allowing fusion or adding a `yield()` call before observing the result, indicates that this is a different issue with similar symptoms (possibly `ChannelFlow`-specific)?

**Version**

`kotlinx-coroutines`:1.10.1
`kotlinx-coroutines-test`:1.10.1

`build.gradle.kts`:
```kts
plugins {
    kotlin("jvm") version "2.1.10"
}
//[...]
dependencies {
//[...]
    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.1")
    testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.10.1")
//[...]
}
tasks.test {
    useJUnitPlatform()
}
kotlin {
    jvmToolchain(17)
}
```

**Provide a Reproducer**

<details open>

<summary>Minimal reproducer (ThreadLocalElement)</summary>

```kotlin
import kotlinx.coroutines.asContextElement
import kotlinx.coroutines.flow.channelFlow
import kotlinx.coroutines.flow.flowOn
import kotlinx.coroutines.flow.map
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.yield
import kotlin.test.Test
import kotlin.test.assertEquals

class CoroutineThreadLocalTest {

    companion object {

        val threadLocal = ThreadLocal<String?>()
    }

    @Test
    fun testChannelFlowThreadLocal_minimal() = runBlocking(threadLocal.asContextElement(value = "1")) {
[...]

Issue #4121, opened 2024-05-03
ThreadLocal.asContextElement may not be cleaned up when used with Dispatchers.Main.immediate


denis-bezrukov:
I am not sure if it is covered by https://github.com/Kotlin/kotlinx.coroutines/issues/2930, but to me steps look different

Having the following code
```
        val threadLocal = ThreadLocal<Int>()
        val stateFlow = MutableStateFlow<Int>(0)

        viewLifecycleOwner.lifecycleScope.launch {
            stateFlow.collect {
                Log.d("testing", "coroutine A: ${threadLocal.get()}")
            }
        }
        viewLifecycleOwner.lifecycleScope.launch(threadLocal.asContextElement(5)) {
            delay(1000)
            Log.d("testing", "before emit")
            stateFlow.emit(1)
            Log.d("testing", "after emit")
        }
```
It logs 
```
coroutine A: null
before emit
after emit
coroutine A: 5
```

So the coroutine A, that doesn't have a threadLocal as a ContextElement, has an access to it just because of Dispatchers.Main.immediate optimization (the code was invoked directly without dispatching, and somehow context element wasn't cleaned up).
If the first coroutine is launched with `Dispatchers.Main` (not immediate) the issue disappears. 

The reproducer is for android project (can be placed in `onViewCreated` in an empty fragment) just because JVM has no direct dispatcher.

dkhalanskyjb:
Non-Android reproducer with `Dispatchers.Unconfined`:

```kotlin
import kotlinx.coroutines.*
import kotlinx.coroutines.flow.MutableStateFlow
import kotlin.time.Duration.Companion.seconds

fun main() {
    runBlocking {
        val threadLocal = ThreadLocal<Int>()
        val stateFlow = MutableStateFlow<Int>(0)

        launch(Dispatchers.Unconfined) {
            stateFlow.collect {
                println("coroutine A: ${threadLocal.get()}")
            }
        }
        launch(threadLocal.asContextElement(5)) {
            delay(1.seconds)
            println("before emit")
            stateFlow.emit(1)
            println("after emit")
        }
    }
}
```

Similar example using undispatched starting strategy:

```kotlin
import kotlinx.coroutines.*
import kotlin.time.Duration.Companion.seconds

fun main() {
    runBlocking {
        val threadLocal = ThreadLocal<Int>()
        withContext(threadLocal.asContextElement(5)) {
[...]

PR #4431, opened 2025-05-06
Fix flowOn handling of thread context elements

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/4403
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/4422
Fixes some other, similar bugs that weren't reported.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Produce.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Produce.kt
@@ -266,6 +266,7 @@ public fun <E> CoroutineScope.produce(
     produce(context, capacity, BufferOverflow.SUSPEND, start, onCompletion, block)
 
 // Internal version of produce that is maximally flexible, but is not exposed through public API (too many params)
+// (scope + context1).produce(context2) == scope.produce(context1 + context2)
 internal fun <E> CoroutineScope.produce(
     context: CoroutineContext = EmptyCoroutineContext,
     capacity: Int = 0,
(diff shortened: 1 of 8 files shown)


qwwdfsad (approved):
Nice one!

LouisCAD:
Any chance a minor release with these fixes is published? It's been more than 6 months since this was merged… 🫤

chaoren:
Hi, we're still seeing issues with `flowOn` interactions with `ThreadContextElement`. Here's a repro: https://pl.kotl.in/EhosParyg

I think this PR specifically fixed some issues, but introduced other new issues.

fzhinkin:
@chaoren, note that Kotlin playground uses an outdated version of kotlinx-coroutines (1.7.3, while the fix landed 1.11.0).

But some cases are indeed failing w/ 1.11.0 (but there's nothing else I can say here, as I didn't analyze the test yet).

dkhalanskyjb:
@chaoren, your reproducer consists of two parts:
1. The thread local not getting cleaned up when switching to coroutines that don't declare their usage of the thread local, and
2. Getting the wrong thread-local value when switching between coroutines that do declare their usages.

The fix here was for 2). Running your reproducer with 1.11, I don't get any prints indicating errors there! 1) is reproducible, but it is a separate issue that's wider than `flowOn`: https://github.com/Kotlin/kotlinx.coroutines/issues/4121 It's mostly a design issue: if a coroutine does not care about thread locals, do we want to waste time settings the thread-local variable that, supposedly, won't even get read? Please chime in under #4121 if you have opinions on this.

PR #4433, opened 2025-05-09
Document structured concurrency

dkhalanskyjb:
This PR updates the documentation so that it clearly defines the structured concurrency interactions. The updated docs include:
* [x] `CoroutineScope.kt`.
* [x] `supervisorScope`
* [x] `Builders.common.kt`: `launch`, `async`, and the other coroutine builders should document their structured concurrency behaviors.
* [x] The `withTimeout` family of functions.
* [x] `produce`.

Some other functions could benefit from this, but are outside the scope of the PR:
* [ ] Coroutine builders from the reactive integrations.
* [ ] JS+Wasm `promise`.
* [ ] The `future` function.

We should include what's already there in the next release, without trying to complete the change. At the cost of introducing the inconsistency, we'll be able to publish a release sooner, and the feedback on the parts that were changed could inform the future edits.

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineScope.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineScope.kt
@@ -327,7 +327,7 @@ import kotlin.coroutines.intrinsics.*
  * job.join() // finishes normally
  * ```
  *
- * If a coroutine fails with an exception,
+ * If a coroutine fails with a non-[CancellationException] exception,
  * is not a coroutine created with lexically scoped coroutine builders like [coroutineScope] or [withContext],
  * *and* its parent is a normal [Job] (not a [SupervisorJob]),
  * the parent fails with that exception, too (and the same logic applies recursively to the parent of the parent, etc.):
@@ -374,10 +374,13 @@ import kotlin.coroutines.intrinsics.*
  * the first one to fail will be propagated, and the rest will be attached to it as
  * [suppressed exceptions][Throwable.suppressedExceptions].
  *
- * If a coroutine fails with an exception and cannot cancel its parent
+ * If a coroutine fails with a non-[CancellationException] exception and cannot cancel its parent
  * (because its parent is a [SupervisorJob] or there is none at all),
  * the failure is reported through other channels.
  * See [CoroutineExceptionHandler] for details.
+ *
+ * Failing with a [CancellationException] only cancels the coroutine itself and its children.
+ * It does not affect the parent or any other coroutines and is not considered a failure.
  *
  * ### How-to: stop failures of child coroutines from cancelling other coroutines
  *
--- a/kotlinx-coroutines-core/common/src/CoroutineScope.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineScope.kt
@@ -10,102 +10,463 @@ import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
 
 /**
- * Defines a scope for new coroutines. Every **coroutine builder** (like [launch], [async], etc.)
- * is an extension on [CoroutineScope] and inherits its [coroutineContext][CoroutineScope.coroutineContext]
- * to automatically propagate all its elements and cancellation.
+ * A scope in which coroutines run.
  *
- * The best ways to obtain a standalone instance of the scope are [CoroutineScope()] and [MainScope()] factory functions,
- * taking care to cancel these coroutine scopes when they are no longer needed (see the section on custom usage below for
- * explanation and example).
+ * A scope defines a group of coroutines with shared execution properties
+ * (represented as [CoroutineContext.Element] values)
+ * and mutually dependent lifecycles.
  *
- * Additional context elements can be appended to the scope using the [plus][CoroutineScope.plus] operator.
+ * The execution properties of coroutines are defined by the [coroutineContext] property of this interface
+ * and are inherited by all coroutines launched in this scope.
+ * For example, the [CoroutineDispatcher] element of the context defines which threads the coroutines will run on,
+ * with [Dispatchers.Main] meaning the coroutines will run on the main thread of the application.
+ * See a more detailed explanation of the context elements in a separate section below.
  *
- * ### Convention for structured concurrency
+ * The lifecycles of coroutines are governed by a set of rules called "structured concurrency",
+ * meaning that the lifetimes of child coroutines are strictly inside the lifetime of the scope.
+ * If the scope is cancelled, all coroutines in it are cancelled too, and the scope itself
+ * cannot be completed until all its children are completed.
+ * See a more detailed explanation of structured concurrency in a separate section below.
  *
- * Manual implementation of this interface is not recommended, implementation by delegation should be preferred instead.
- * By convention, the [context of a scope][CoroutineScope.coroutineContext] should contain an instance of a
- * [job][Job] to enforce the discipline of **structured concurrency** with propagation of cancellation.
+ * ## Using the coroutine scope
  *
- * Every coroutine builder (like [launch], [async], and others)
- * and every scoping function (like [coroutineScope] and [withContext]) provides _its own_ scope
- * with its own [Job] instance into the inner block of code it runs.
- * By convention, they all wait for all the coroutines inside their block to complete before completing themselves,
- * thus enforcing the structured concurrency. See [Job] documentation for more details.
+ * The methods of this interface are not intended to be called directly.
+ * Instead, a [CoroutineScope] is passed as a receiver to the coroutine builders such as [launch] and [async],
+ * controlling the execution properties and lifetimes of the created coroutines.
  *
- * ### Android usage
+ * ## Coroutine context elements
  *
- * Android has first-party support for coroutine scope in all entities with the lifecycle.
- * See [the corresponding documentation](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope).
+ * A [CoroutineScope] is defined by a set of [CoroutineContext] elements, one of which is typically a [Job],
+ * described in the section on structured concurrency and responsible for managing lifetimes of coroutines.
  *
- * ### Custom usage
+ * Other coroutine context elements include, but are not limited to, the following:
  *
- * `CoroutineScope` should be declared as a property on entities with a well-defined lifecycle that are
- * responsible for launching child coroutines. The corresponding instance of `CoroutineScope` shall be created
- * with either `CoroutineScope()` or `MainScope()`:
+ * - The scheduling policy, represented by a [CoroutineDispatcher] element.
+ *   Some commonly used dispatchers are provided in the [Dispatchers] object.
+ * - [CoroutineExceptionHandler] that defines how failures of child coroutines should be reported whenever
+ *   structured concurrency does not provide a way to propagate the failure to the parent
+ *   (that is, a [SupervisorJob] is used).
+ * - A [CoroutineName] element that can be used to name coroutines for debugging purposes.
+ * - On the JVM, a `ThreadContextElement` ensures that a specific thread-local value gets set on the thread
+ *   that executes the coroutine.
  *
- * - `CoroutineScope()` uses the [context][CoroutineContext] provided to it as a parameter for its coroutines 
- *   and adds a [Job] if one is not provided as part of the context.
- * - `MainScope()` uses [Dispatchers.Main] for its coroutines and has a [SupervisorJob].
+ * ## Obtaining a coroutine scope
  *
- * **The key part of custom usage of `CoroutineScope` is cancelling it at the end of the lifecycle.**
- * The [CoroutineScope.cancel] extension function shall be used when the entity that was launching coroutines
- * is no longer needed. It cancels all the coroutines that might still be running on behalf of it.
+ * Manual implementations of this interface are not recommended.
+ * Instead, there are several structured ways to obtain a [CoroutineScope].
  *
- * For example:
+ * ### Lexical scopes
+ *
+ * [coroutineScope] and [supervisorScope] functions can be used in any `suspend` function to define a scope
+ * lexically, ensuring that all coroutines launched in this scope have completed by the time scope-limiting
+ * function exits.
  *
  * ```
- * class MyUIClass {
- *     val scope = MainScope() // the scope of MyUIClass, uses Dispatchers.Main
- *
- *     fun destroy() { // destroys an instance of MyUIClass
- *         scope.cancel() // cancels all coroutines launched in this scope
- *         // ... do the rest of cleanup here ...
+ * suspend fun doSomething() = coroutineScope { // scope `A`
+ *     repeat(5) { outer ->
+ *         // spawn a new coroutine in the scope `A`
+ *         launch {
+ *             println("Coroutine $outer started")
+ *             coroutineScope { // scope `B`, separate for each `outer` coroutine
+ *                 repeat(5) { inner ->
+ *                     // spawn a new coroutine in the scope `B`
+ *                     launch {
+ *                         println("Coroutine $outer.$inner started")
+ *                         delay(10.milliseconds)
+ *                         println("Coroutine $outer.$inner finished")
+ *                     }
+ *                 }
+ *             }
+ *             // will only exit once all `Coroutine $outer.X finished` messages are printed
+ *             println("Coroutine $outer finished")
+ *         }
  *     }
+ * } // will only exit once all `Coroutine X finished` messages are printed
+ * ```
  *
- *     /*
- *      * Note: if this instance is destroyed or any of the launched coroutines
- *      * in this method throws an exception, then all nested coroutines are cancelled.
- *      */
- *     fun showSomeData() = scope.launch { // launched in the main thread
- *        // ... here we can use suspending functions or coroutine builders with other dispatchers
- *        draw(data) // draw in the main thread
+ * This should be the preferred way to create a scope for coroutines.
+ *
+ * ### `CoroutineScope` constructor function
+ *
+ * When the lifecycle of the scope is not limited lexically
+ * (for example, when coroutines should outlive the function that creates them)
+ * but is tied to the lifecycle of some entity, the [CoroutineScope] constructor function can be used
+ * to define a personal scope for that entity that should be stored as a field there.
+ *
+ * **The key part of using a custom `CoroutineScope` is cancelling it at the end of the lifecycle.**
+ * The [CoroutineScope.cancel] extension function shall be used when the entity launching coroutines
+ * is no longer needed. It cancels all the coroutines that might still be running on its behalf.
+ *
+ * ```
+ * class MyEntity(scope: CoroutineScope? = null): AutoCloseable {
+ *    // careful: do not write `get() =` here by accident!
+ *    private val scope = scope ?: CoroutineScope(SupervisorJob() + CoroutineExceptionHandler { _, e ->
+ *        println("Error in coroutine: $e")
+ *    })
+ *
+ *    fun doSomethingWhileEntityExists() = scope.launch {
+ *        while (true) {
+ *            // do some work
+ *            delay(50.milliseconds)
+ *            println("Doing something")
+ *        }
+ *    }
+ *
+ *    override fun close() {
+ *        // cancel all computations related to this entity
+ *        scope.cancel()
+ *    }
+ * }
+ *
+ * fun main() {
+ *     MyEntity().use { entity ->
+ *         entity.doSomethingWhileEntityExists()
+ *         Thread.sleep(200)
+ *         entity.close()
  *     }
  * }
  * ```
+ *
+ * Usually, a custom [CoroutineScope] should be created with a [SupervisorJob] and
+ * a [CoroutineExceptionHandler] to handle exceptions in child coroutines.
+ * See the documentation for the [CoroutineScope] constructor function for more details.
+ * Also note that `MyEntity` accepts the `scope` parameter that can be used to pass a custom scope for testing.
+ *
+ * Sometimes, authors of coroutine-aware frameworks provide [CoroutineScope] instances like this out of the box.
+ * For example, on Android, all entities with a lifecycle and all `ViewModel` instances expose a [CoroutineScope]:
+ * see [the corresponding documentation](https://developer.android.com/topic/libraries/architecture/coroutines).
+ *
+ * ### Taking another view of an existing scope
+ *
+ * Occasionally, several coroutines need to be launched with the same additional [CoroutineContext] that is not
+ * present in the original scope.
+ * In this case, the [CoroutineScope.plus] operator can be used to create a new view of the existing scope:
+ *
+ * ```
+ * coroutineScope {
+ *     val sameScopeButInUiThread = this + Dispatchers.Main
+ *     sameScopeButInUiThread.launch {
+ *         println("Running on the main thread")
+ *     }
+ *     launch {
+ *         println("This will run using the original dispatcher")
+ *     }
+ *     sameScopeButInUiThread.launch {
+ *         println("And this will also run on the main thread")
+ *     }
+ * }
+ * ```
+ *
+ * The lifecycle of the new scope is the same as the original one, but the context includes new elements.
+ *
+ * ### Application lifecycle scope
+ *
+ * [GlobalScope] is a [CoroutineScope] that has the lifetime of the whole application.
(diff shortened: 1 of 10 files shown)


kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ * ## Using the coroutine scope
  *
- * Every coroutine builder (like [launch], [async], and others)
- * and every scoping function (like [coroutineScope] and [withContext]) provides _its own_ scope
- * with its own [Job] instance into the inner block of code it runs.
- * By convention, they all wait for all the coroutines inside their block to complete before completing themselves,
- * thus enforcing the structured concurrency. See [Job] documentation for more details.
+ * The methods of this interface are not intended to be called directly.
+ * Instead, a [CoroutineScope] is passed as a receiver to the coroutine builders such as [launch] and [async],
+ * controlling the execution properties and lifetimes of the created coroutines.
```
murfel: lifetimes -> lifecycles

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
- * - `CoroutineScope()` uses the [context][CoroutineContext] provided to it as a parameter for its coroutines 
- *   and adds a [Job] if one is not provided as part of the context.
- * - `MainScope()` uses [Dispatchers.Main] for its coroutines and has a [SupervisorJob].
+ * ## Obtaining a coroutine scope
  *
- * **The key part of custom usage of `CoroutineScope` is cancelling it at the end of the lifecycle.**
- * The [CoroutineScope.cancel] extension function shall be used when the entity that was launching coroutines
- * is no longer needed. It cancels all the coroutines that might still be running on behalf of it.
+ * Manual implementations of this interface are not recommended.
+ * Instead, there are several structured ways to obtain a [CoroutineScope].
```
murfel: ~~structured~~

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ * Other coroutine context elements include, but are not limited to, the following:
  *
- * `CoroutineScope` should be declared as a property on entities with a well-defined lifecycle that are
- * responsible for launching child coroutines. The corresponding instance of `CoroutineScope` shall be created
- * with either `CoroutineScope()` or `MainScope()`:
+ * - The scheduling policy, represented by a [CoroutineDispatcher] element.
+ *   Some commonly used dispatchers are provided in the [Dispatchers] object.
+ * - [CoroutineExceptionHandler] that defines how failures of child coroutines should be reported whenever
+ *   structured concurrency does not provide a way to propagate the failure to the parent
+ *   (typically, because the root scope of the ancestry tree is not lexically scoped).
```
murfel: fyi: this is where I had to lookup what "lexically scoped" means, my understanding from the google summary is "where / in which context it is defined / written in the source code"

but this still leaves me wondering what are the specific cases this "not lexically scoped" refers to

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
- * is no longer needed. It cancels all the coroutines that might still be running on behalf of it.
+ * Manual implementations of this interface are not recommended.
+ * Instead, there are several structured ways to obtain a [CoroutineScope].
  *
- * For example:
+ * ### Lexical scopes
+ *
+ * [coroutineScope] and [supervisorScope] functions can be used in any `suspend` function to define a scope
+ * lexically, ensuring that all coroutines launched in this scope have completed by the time scope-limiting
+ * function exits.
```
murfel: Split the sentence in two to simplify reading. Less ideas per sentence.

```suggestion
 * [coroutineScope] and [supervisorScope] functions can be used in any `suspend` function to define a scope
 * lexically. All coroutines launched in the new scope will have completed by the time the scope-limiting
 * function exits.
```

murfel: Maybe it's worth avoiding the new term "scope-limiting function". Maybe "by the time the lexical scope function exits"?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *                 }
+ *             }
+ *             // will only exit once all `Coroutine $outer.X finished` messages are printed
+ *             println("Coroutine $outer finished")
+ *         }
+ *     }
+ * } // will only exit once all `Coroutine X finished` messages are printed
+ * ```
+ *
+ * This should be the preferred way to create a scope for coroutines.
```
murfel: Move it before the example? So that there's no confusion if it refers to this section, or if it's a setup for the next section.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
murfel: I managed to get the gist of "lexically" after reading this section, but I stumbled every time I saw this word.

It only gets clear after the _Lexical scopes_ and _CoroutineScope constructor_ sections.

Possibly define it, avoid using before these sections, or refer to those sections?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *
+ * ### `CoroutineScope` constructor function
+ *
+ * When the lifecycle of the scope is not limited lexically
+ * (for example, when coroutines should outlive the function that creates them)
+ * but is tied to the lifecycle of some entity, the [CoroutineScope] constructor function can be used
+ * to define a personal scope for that entity that should be stored as a field there.
```
murfel:
```suggestion
 * When the lifecycle of the scope is not limited lexically
 * (for example, when coroutines should outlive the function that creates them)
 * but is tied to the lifecycle of some entity, the [CoroutineScope] constructor function can be used
 * to define a personal scope for the entity. This scope should be stored as a field in the entity.
```

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *             // will only exit once all `Coroutine $outer.X finished` messages are printed
+ *             println("Coroutine $outer finished")
+ *         }
+ *     }
+ * } // will only exit once all `Coroutine X finished` messages are printed
+ * ```
+ *
+ * This should be the preferred way to create a scope for coroutines.
+ *
+ * ### `CoroutineScope` constructor function
```
murfel, line 86: Shouldn't it just be called _constructor_, instead of _constructor function_?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ * ```
+ *
+ * The lifecycle of the new scope is the same as the original one, but the context includes new elements.
+ *
+ * ### Application lifecycle scope
+ *
+ * [GlobalScope] is a [CoroutineScope] that has the lifetime of the whole application.
+ * It is convenient for launching top-level coroutines that are not tied to the lifecycle of any entity,
+ * but it is easy to misuse it and create memory leaks or resource leaks when a coroutine actually should be tied
+ * to the lifecycle of some entity.
```
murfel: Simplify by splitting into two, less ideas per sentence.

> However, it is easy to misuse...

(Let me know if you'd like me to point out other occurrences of this in my future reading bouts of this PR.)

(By the way, prompting an LLM with "rewrite this sentence" gives out the same sentence split that what I would suggest, and some stylistically better words that I wouldn't come up with myself. Although it needs some editing to put all the erased terms back in.)

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ * ```
+ * GlobalScope.launch(CoroutineExceptionHandler { _, e ->
+ *     println("Error in coroutine: $e")
+ * }) {
+ *    while (true) {
+ *        println("I will be running forever, you cannot stop me!")
+ *        delay(1.seconds)
+ *    }
+ * }
+ * ```
```
murfel, line 174: I feel like it's more of an example, unrelated to GlobalScope, for "if you define your custom exception handler which ignores exceptions, no one will be able to cancel you", rather than "if you misuse GlobalScope, you're setting yourself up for failure".

murfel: Also it relies on the hidden curriculum of "cancelation is implemented as an exception" which I believe hasn't been covered up to here yet.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *
+ * ### `by`-delegation
+ *
+ * When all else fails and a custom [CoroutineScope] implementation is needed, it is recommended to use
+ * `by`-delegation to implement the interface:
```
murfel: _Fails_ is a very specific word and brings unneeded connotations.

> When no other approaches are applicable

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *     println("Error in coroutine: $e")
+ * }) {
+ *    while (true) {
+ *        println("I will be running forever, you cannot stop me!")
+ *        delay(1.seconds)
+ *    }
+ * }
+ * ```
+ *
+ * ### `by`-delegation
```
murfel, line 176: I'm a bit surprised that I only learn that I should use by-delegration instead of the constructor [function] at the very end of the section. Possibly mention in the beginning that the preferred approach is the by-delegation, and the constructor is only spelled out there for clarity/demonstration?

murfel (commented):
Let me know if it's too nitpicky. Or if you'd prefer the first round to be more high-level, and then you'd be interested in nitpicky to polish it up. Your feedback to my feedback is appreciated 🙂

dkhalanskyjb:
Higher-level overview first is nice, since there's no point in debating the grammar of a section that gets removed as a whole. So, the nitpicking should (initially) be limited to exactly the level above which everything's already well done.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: Would a specific example make it clear?
```suggestion
 *   (typically, because the root scope of the ancestry tree is not lexically scoped,
      that is, not created using builders like [coroutineScope] or [withContext]
      that return the result directly to the caller).
```

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: Why?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: `CoroutineScope` is not a class and does not have a constructor, we only provide a constructor function.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: Splitting removes the focus on the contradiction, which is the whole point. Would this work instead?
```suggestion
 * Although is convenient for launching top-level coroutines that are not tied to the lifecycle of any entity,
 * it is easy to misuse it and create memory leaks or resource leaks when a coroutine actually should be tied
 * to the lifecycle of some entity.
```

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: The example *is* using the constructor function here to obtain a non-custom implementation. The constructor function is better than `by`-delegation, which is in turn better than `override val coroutineContext: CoroutineContext`.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: > if you define your custom exception handler which ignores exceptions, no one will be able to cancel you

That's not true, though. The `CoroutineExceptionHandler` is here just to solidify the good practice of supplying them. Removing the `CoroutineExceptionHandler` does not change the point.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: Less ideas per sentence is good, but is it really okay to have less than one idea per sentence? In my view, it's better to have a longer sentence that describes a single idea instead of having two sentences with 0.5 ideas each.

dkhalanskyjb: > the new term "scope-limiting function". Maybe "by the time the lexical scope function exits"?

"The lexical scope function" is also a new term, though. What makes a function lexical-scope? Fully stated, `coroutineScope` is a function that limits the coroutine scope to the lexical scope of its block, yet this is a mouthful and distracts from the idea that's being conveyed. Do you have an idea of how this could be expressed clearly but also succinctly?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: Could you elaborate?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
dkhalanskyjb: Is it confusing in the rendered form as well? It's expected that this text will be read from kotlinlang.org or from the IDE in the popup window that appears on a mouse-over.

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *     // Await completion of the child coroutine,
+ *     // and therefore the grandchild coroutine too.
+ *     childJob.join()
+ *     check(grandChildFinished)
+ * }
+ * ```
+ *
+ * Such a [CoroutineScope] receiver is provided for [launch], [async], and other coroutine builders,
+ * as well as for scoping functions like [coroutineScope], [supervisorScope], and [withContext].
+ * Each of these [CoroutineScope] instances is tied to the lifecycle of the code block it runs in.
```
murfel, line 295: lexical

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *
+ * ```
+ * coroutineScope {
+ *     val job = coroutineContext[Job]
+ *     val childJob = launch { }
+ *     check(job === childJob.parent)
+ * }
+ * ```
+ *
+ * Because every coroutine has a lifecycle and a [Job], a [CoroutineScope] can be associated with it.
```
murfel: - Job *is* a lifecycle, so "and hence a Job representing it", they are not equal
- the corollary is not clear
- it's more like: every coroutine is represented in code by CoroutineScope, hence CoroutineScope's Job represents the coroutine's lifecycle (should be at the top of the section)

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *
+ * #### Interactions between coroutines
+ *
+ * See the [Job] documentation for a detailed explanation of interactions between [Job] values.
+ * Below is a summary of the most important points for structuring code in day-to-day usage.
+ *
+ * A coroutine cannot reach the final state until all its children have reached their final states.
+ * See the example above.
+ *
+ * If a [CoroutineScope] is cancelled (either explicitly or because it corresponds to some coroutine that failed
```
murfel, line 311: can a coroutine be cancelled? or it's only "scope can be cancelled", "coroutine can fail with an exception"

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *      awaitCancellation()
+ * }
+ * scope.cancel() // comment this out for the line below to hang
+ * job.join() // finishes normally
+ * ```
+ *
+ * If a coroutine fails with a non-[CancellationException] exception,
+ * is not a coroutine created with lexically scoped coroutine builders like [coroutineScope] or [withContext],
+ * *and* its parent is a normal [Job] (not a [SupervisorJob]),
+ * the parent fails with that exception, too (and the same logic applies recursively to the parent of the parent, etc.):
```
murfel: (wording) general note: too long till the punchline, move it to the front:

if a coroutine fails with ..., the parent fails with the same exception if:
- one
- two
- three

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ *                 awaitCancellation()
+ *             }
+ *         }
+ *     }.exceptionOrNull()
+ *     is IllegalStateException
+ * )
+ * // the calling coroutine will *not* be cancelled,
+ * // even though the caller is a parent of the failed `coroutineScope`,
+ * // because `coroutineScope` is a lexically scoped coroutine builder,
+ * // which propagate their exceptions by rethrowing them.
```
murfel: I didn't understand the explanation

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
+ * )
+ * ```
+ *
+ * If several coroutines fail with non-[CancellationException] exceptions,
+ * the first one to fail will be propagated, and the rest will be attached to it as
+ * [suppressed exceptions][Throwable.suppressedExceptions].
+ *
+ * If a coroutine fails with a non-[CancellationException] exception and cannot cancel its parent
+ * (because its parent is a [SupervisorJob] or there is none at all),
+ * the failure is reported through other channels.
```
murfel: through other means, since channel is a term

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
- *        // ... here we can use suspending functions or coroutine builders with other dispatchers
- *        draw(data) // draw in the main thread
+ * ### How-to: prevent a child coroutine from being cancelled
+ *
+ * Sometimes, you may want to run a coroutine even if the parent coroutine is cancelled.
+ * This pattern provides a way to achieve that:
+ *
+ * ```
+ * scope.launch(start = CoroutineStart.ATOMIC) {
+ *     // Do not move `NonCancellable` to the `context` argument of `launch`!
```
murfel, line 426: argument -> parameter

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
- * The best ways to obtain a standalone instance of the scope are [CoroutineScope()] and [MainScope()] factory functions,
- * taking care to cancel these coroutine scopes when they are no longer needed (see the section on custom usage below for
- * explanation and example).
+ * A scope defines a group of coroutines with shared execution properties
+ * (represented as [CoroutineContext.Element] values)
+ * and mutually dependent lifecycles.
  *
- * Additional context elements can be appended to the scope using the [plus][CoroutineScope.plus] operator.
+ * The execution properties of coroutines are defined by the [coroutineContext] property of this interface
+ * and are inherited by all coroutines launched in this scope.
```
murfel: CoroutineContext = Job + execution properties + other stuff

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,481 @@ import kotlin.coroutines.*
...
- * Defines a scope for new coroutines. Every **coroutine builder** (like [launch], [async], etc.)
- * is an extension on [CoroutineScope] and inherits its [coroutineContext][CoroutineScope.coroutineContext]
- * to automatically propagate all its elements and cancellation.
+ * A scope in which coroutines run.
  *
- * The best ways to obtain a standalone instance of the scope are [CoroutineScope()] and [MainScope()] factory functions,
- * taking care to cancel these coroutine scopes when they are no longer needed (see the section on custom usage below for
- * explanation and example).
+ * A scope defines a group of coroutines with shared execution properties
+ * (represented as [CoroutineContext.Element] values)
```
dkhalanskyjb:
```suggestion
 * A scope defines a group of coroutines with shared execution properties
 * (for example, on which threads the coroutines execute or which thread-local variables are tracked in them)
```
Would this clarify the meaning of "execution properties" here and for the rest of the kdoc?

LouisCAD: Late comment 🙈

There's nothing thread-local in a coroutine scope. `CoroutineContext.Element`s are an analogue, but are not thread-bound. Mentioning "thread-local" like this would create confusion with `ThreadLocal` in Java and equivalent in other platforms or libraries.

"are tracked in them" isn't clear to me either.

dkhalanskyjb: What I meant to do here was provide `ThreadContextElement` as an example of coroutine context elements. These represent thread-local variables that are tracked in a coroutine. Does this make sense? Could you suggest a better phrasing?

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,485 @@ import kotlin.coroutines.*
...
+ *     coroutineScope {
+ *         val file1 = async {
+ *             // if this fails, everything else quickly fails too
+ *             downloadFile("http://example.com/file1")
+ *         }
+ *         val file2 = async {
+ *             downloadFile("http://example.com/file2")
+ *         }
+ *         launch(Dispatchers.Main) {
+ *             // create a separate coroutine on the UI thread
```
murfel: Transpose the lines, first the comment, then the action

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,485 @@ import kotlin.coroutines.*
...
+ * and it is enough to specify the relationships between operations declaratively, as shown in the example,
+ * without being overwhelmed by intricate inter-thread communications.
+ *
+ * ### Specifics
+ *
+ * Coroutines and [CoroutineScope] instances have an associated lifecycle.
+ * A runtime representation of a lifecycle in `kotlinx.coroutines` is called a [Job].
+ * [Job] instances form a hierarchy of parent-child relationships,
+ * and the [Job] of every coroutine spawned in a [CoroutineScope] is a child of the [Job] of that scope.
+ * This is often shortened to saying that the coroutine is the scope's child.
```
murfel: Err, I don't like this wording. I prefer: scope/Job/coroutine used interchangeably (think and elaborate)

kotlinx-coroutines-core/common/src/CoroutineScope.kt
```diff
@@ -10,102 +10,485 @@ import kotlin.coroutines.*
...
+ * ```
+ *
+ * ### `by`-delegation
+ *
+ * When the approaches listed above are not applicable and a custom [CoroutineScope] implementation is needed,
+ * it is recommended to use `by`-delegation to implement the interface:
+ *
+ * ```
+ * class MyEntity : CoroutineScope by CoroutineScope(
+ *     SupervisorJob() + Dispatchers.Main + CoroutineExceptionHandler { _, e ->
```
murfel: Why `Dispatchers.Main` though? Because Android?

PR #4434, opened 2025-05-13
Explain that `limitedParallelism(1)` is not a mutex

dkhalanskyjb:


kotlinx-coroutines-core/common/src/CoroutineDispatcher.kt
```diff
@@ -149,6 +150,42 @@ public abstract class CoroutineDispatcher :
...
+     * Coroutine 1 entering...
+     * Coroutine 2 entering...
+     * Coroutine 0 leaving.
+     * Coroutine 1 leaving.
+     * Coroutine 2 leaving.
+     * ```
+     *
+     * This means that coroutines are not guaranteed to run to completion before the dispatcher starts executing
+     * code from another coroutine.
+     * The only guarantee in this example is that two `println` calls in several threads simultaneously.
```
JakeWharton, line 185: Seems to be missing a few words
```suggestion
     * The only guarantee in this example is that two `println` calls will not occur in several threads simultaneously.
```

qwwdfsad (approved):

Issue #3670, opened 2023-03-09
Disallow `Job` passed as context to `withContext`/`launch`/`async`


dovchinnikov:
**What do we have now?**

It's possible to pass `Job` as part of context to the mentioned functions:
```
withContext(Job()) {
    ...
}
cs.launch(Job()) {
    ...
}
cs.async(Job()) {
    ...
}
```

**What should be instead?**

Error should be logged when a `Job` is detected in the added context.

**Why?**

All above cases "replace" the context `Job` and attach the scope `Job` to the passed `Job()` as a child. 
This might be unexpected at best, and plain dangerous because it breaks parent-child relation.

dovchinnikov:
This is continuation of https://github.com/Kotlin/kotlinx.coroutines/issues/814#issuecomment-1457747716

dovchinnikov:
> So, here is the problem: how do we carefully migrate from today's state to this ideal future state? We need to somehow start giving a warning on any code that tries to pass coroutine context with a job as a parameter to those coroutine builder functions, so that we can later turn this warning into an error. But how do we even identify such code? That is the question.

Let's consider correct use cases, and declare the rest incorrect.

1. `NonCancellable`
```
withContext(NonCancellable) {}
```
`NonCancellable` is a singleton, `withContext` can detect and allow this logging a warning with invitation to replace it with `nonCancellable {}` (TBD name).

2. Orphan jobs x `withContext`
```
withContext(Job()) {}
withContext(SupervisorJob()) {}
withContext(CompletableDeferred<Unit>()) {}
```
Can be detected by checking the parent handle. It is essentially the same as 1 and should be handled as such. 
`withContext` does not change the state of the `Job`, so all three calls behave like `nonCancellable {}`

3. Orphan jobs x `launch`/`async`
```
cs.launch(Job()) {} 
cs.launch(SupervisorJob()) {} 
... 
```
`launch`/`async` do change the state of the parent `Job`, so there is a difference between regular and supervisor jobs.
This should be offered to be replaced with `CoroutineScope(cs.coroutineContext + Job()).launch {}` (or supervisor variant), at least this way it's evident that `launch`-ed coroutine is a child of that new scope.

4. Child of current context x `withContext`
```
withContext(Job(parent = currentCoroutineContext().job)) {}
```
The above example will hang the current coroutine because a child is attached, but never completed. 
```
val job = Job(parent = currentCoroutineContext().job)
try {
  withContext(job) {}
} 
[...]

Issue #4329, opened 2025-01-14
Reconsider what happens when `await()` gets cancelled


dkhalanskyjb:
In `kotlinx.coroutines`, we have two types of functions called `await()`:

* Multiple consumers can call `await()` independently: this is the case for [JS Promise](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await.html), [Deferred](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/await.html), the [reactive integrations](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx2/kotlinx.coroutines.rx2/await.html), and [`com.google.android.gms.tasks.Task` by default](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-play-services/kotlinx.coroutines.tasks/await.html).
* If a consumer is who calls `await()` is cancelled, the entity gets cancelled: this is the idea behind bidirectional cancellation in https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.future/await.html and https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-guava/kotlinx.coroutines.guava/await.html, and  [`com.google.android.gms.tasks.Task` if you pass a cancellation token](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-play-services/kotlinx.coroutines.tasks/await.html).

A way to share (for example) a `ListenableFuture` among several consumers is to [convert it to a `Deferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-guava/kotlinx.coroutines.guava/as-deferred.html) and call `await()` on that.

There are a few things that bother me about this.

* `future.await()` behaves differently from `future.asDeferred().await()`: one cancels the computation when `await()` is cancelled, the other one doesn't. It's inconsistent, and not in a way that's unavoidable when translating concepts across ecosystems: if `Future.get` gets interrupted, the whole computation doesn't get aborted, so this seems like `kotlinx.coroutines` invention.
* Why do these two operations that behave significantly differently share the same name? The data structure that is used is orthogonal to whether it's single-shot. If the second kind of `await()` was called something different (for example, `consume()`), we could
[...]

PR #4435, opened 2025-05-13
Introduce deprecated coroutine builder overloads accepting a Job

dkhalanskyjb:
This is a small step towards deprecating the practice of breaking
structured concurrency by passing a `Job` to a coroutine builder.

Thanks to @LouisCAD for the idea!

See https://github.com/Kotlin/kotlinx.coroutines/issues/3670

### DIFF
--- a/kotlinx-coroutines-core/common/src/Guidance.kt
+++ b/kotlinx-coroutines-core/common/src/Guidance.kt
@@ -318,9 +318,10 @@ public fun <T> async(
  * // Guarantees the completion, but not the delivery of the value
  * scope.async(start = CoroutineStart.ATOMIC) {
  *     withContext(NonCancellable) {
- *         // this line will be reached even if the parent is cancelled
+ *         // The actual body of the coroutine.
+ *         // This code will get executed even if the parent is cancelled.
  *     }
- *     // note: the cancellation exception *can* be thrown here,
+ *     // Note: the cancellation exception *can* be thrown here,
  *     // losing the computed value!
  * }
  *


LouisCAD:
Awesome stuff!

kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -136,6 +136,29 @@ private class LazyDeferredCoroutine<T>(
...
+ *     withContext(extraContextWithoutJob) {
+ *         // your code here
+ *     }
+ * }
+ * try {
+ *     deferred.await()
+ * } finally {
+ *     // if `await` fails because the current job is cancelled,
+ *     // also cancel the now-unnecessary computations
+ *     deferred.cancel()
```
lukellmann, line 156: Should it be pointed out that just calling `cancel` on `deferred` won't prevent the current coroutine from completing before the now cancelled coroutine in the other scope completes?

kotlinx-coroutines-core/common/src/Guidance.kt
```diff
@@ -20,6 +20,152 @@ public fun launch(
...
+ *     println("Failure($e)")
+ * }) {
+ *     supervisorScope {
+ *         val coroutines = List(10) {
+ *             launch {
+ *                 delay(10.milliseconds * it)
+ *                 throw IllegalStateException("$it is tired of all this")
+ *             }
+ *         }
+ *         coroutines.joinAll()
```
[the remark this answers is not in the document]
dkhalanskyjb: Yes, it will. The idea was to highlight how `joinAll` will not fail in this scenario, but this piece of documentation may not be a good place for that.

kotlinx-coroutines-core/common/src/Guidance.kt
```diff
@@ -36,3 +182,152 @@ public fun <T> async(
...
+ *
+ * ```
+ * // Guarantees the completion, but not the delivery of the value
+ * scope.async(start = CoroutineStart.ATOMIC) {
+ *     withContext(NonCancellable) {
+ *         // this line will be reached even if the parent is cancelled
+ *     }
+ *     // note: the cancellation exception *can* be thrown here,
+ *     // losing the computed value!
+ * }
```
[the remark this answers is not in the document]
dkhalanskyjb: My comment is misleading, but yes, the value can be lost, just not because of `withContext`. The problem is that if the `Deferred` is cancelled before returning a value, it will no longer be able to finish with a value, even if the computation succeeds.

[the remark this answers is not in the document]
lukellmann: Ok, that's a bit surprising to me. I guess it's because it could have children that haven't completed yet?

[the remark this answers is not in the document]
dkhalanskyjb: Nope, even without any children: https://pl.kotl.in/-ZYf_PR9z
The reason can be seen in the diagram in https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/ : once a `Job` gets cancelled, there is no way back to being `Active` or `Completing` or returning a value.

LouisCAD:
With these changes, `someScope.launch(NonCancellable)` will no longer compile. I've yet to see a tangible use case for that, where the `someScope` is useful (for anything but `Job`, that'd be), but wanted to point it out.

Another construct could be thought out, or an overload taking the `NonCancellable` type could be introduced, if that actually needs to be addressed.

dkhalanskyjb:
@LouisCAD, what is the compilation error? I've just checked, and `launch(NonCancellable)` compiles for me (with a warning, as intended). I consider `launch(NonCancellable)` to be incorrect, with `launch(start = CoroutineStart.ATOMIC) { withContext(NonCancellable) { /* computation */ } }` being a strictly better version of the same.

LouisCAD:
Right, it compiles with just a warning.
For some reason, I thought the deprecation level was error, though I guess that is where it's headed, fast forward into the future.

Came back here to comment about a specific use case I just had.

In the snippet below, I want an `async` that contains its crashes to `await()`/`join()` calls on the resulting `Deferred`, and doesn't bubble it up to its parents.
I can't use `supervisorScope { … }` this time, so I'm using this:

```kotlin
async(SupervisorJob(coroutineContext.job)) { // this (the receiver) is a `CoroutineScope`
    // Some throwing code
}
```

With those changes, it'd trigger a warning, even though it's technically correct.

Addressing #4329 would provide a good alternative, though.

kotlinx-coroutines-core/common/src/Guidance.kt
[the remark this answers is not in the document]
LouisCAD: I think it'd make sense to make an overload to specifically accepts `NonCancellable`. It's doable since `NonCancellable` is not just a `Job` instance, but an object that has its own exclusive type.

That overload could then have the KDoc just for this use case, and could also have a matching `ReplaceWith` clause.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -54,6 +54,25 @@ public fun <T> CoroutineScope.future(
     return coroutine.future
 }
 
+/**
+ * Deprecated version of [future] that accepts a [Job].
+ *
+ * See the documentation for the non-deprecated [future] function to learn about the functionality of this function.
```
murfel, line 60: Is there a way to fully-qualify the non-deprecated `[future]`, so that it's clickable? Currently, it stays on the same page if I click on it in the kdoc in the IDE.

And the deprecated [async] below shows the non-deprecated async kdoc, confusingly.

I noticed that we don't really care about navigability of kdocs. If so, is it because of some technical difficulties outside of kx.coroutines control?

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+
+/**
+ * Deprecated version of [produce] that accepts a [Job].
+ *
+ * See the documentation for the non-deprecated [produce] function to learn about the functionality of this function.
+ * This piece of documentation explains why this overload is deprecated.
+ *
+ * It is incorrect to pass a [Job] as [context] to [produce], [async], or [launch],
+ * because this violates structured concurrency.
+ * The passed [Job] becomes the sole parent of the newly created coroutine, which completely severs the tie between
```
murfel, line 518: "completely severs the tie" -> "removes the parent-child relationship"

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ *   if the [CoroutineScope]'s lifecycle is bound to some component that may not be used after it's destroyed,
+ *   performing computations after the parent [CoroutineScope] is cancelled may lead to crashes.
+ * - For concurrent decomposition of work (when the [CoroutineScope] contains a non-[supervisor][SupervisorJob] job),
+ *   failure of the newly created coroutine also causes the sibling coroutines to fail,
+ *   improving the responsiveness of the program:
+ *   unnecessary computations will not proceed when it's obvious that they are not needed.
+ * - The [CoroutineScope] can only complete when all its children complete.
+ *   If the [CoroutineScope] is lexically scoped (for example, created by [coroutineScope], [supervisorScope],
+ *   or [withContext]), this means that
+ *   the lexical scope will only be exited (and the calling function will finish) once all child coroutines complete.
```
murfel, line 536: Points 1-3 are explanation to why "cancellation of the parent job cancels the children" is useful, and point 4 is an independent one. Consider moving points 1-3 into sub-points.

Then your structure will be: structured concurrency ensures
* A cancelled parents cancel children (A is important because 1, 2, 3)
* B parent waits for children

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ *
+ * ## Possible alternatives
+ *
+ * ### Avoiding cancellation
+ *
+ * Sometimes, it is undesirable for the child coroutine to react to the cancellation of the parent: for example,
+ * some computations have to be performed either way.
+ * `produce(NonCancellable)` or `produce(Job())` can be used to achieve this effect.
```
murfel: It's unclear if this is what you should do, or what you should avoid.

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ *
+ * Alternative approaches that preserve structured concurrency are this:
```
murfel:
```suggestion
 * Here's an alternative approach that preserves structured concurrency:
```

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ * // Guarantees the completion
+ * scope.produce(start = CoroutineStart.ATOMIC) {
+ *     withContext(NonCancellable) {
+ *         // this line will be reached even if the parent is cancelled
+ *     }
+ * }
+ * ```
+ *
+ * This way, the child coroutine is guaranteed to complete,
+ * but the scope is still aware of the child and can track its lifecycle.
```
murfel: "can track its lifecycle" is some new info and new language. It would be more consistent to only refer to the benefits described in the structured concurrency section. Alternatively, add the lifecycle language to the "structured concurrency" section (structured concurrency ensures parent can track child lifecycle so that ...). Some other points need to be rewritten then.

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ *
+ * ### Avoid cancelling other children
+ *
+ * Occasionally, the failure of one child does not mean the work of the other children is also unneeded.
```
murfel: Avoid double negation

```suggestion
 * Occasionally, the failure of one child does not mean that other children should discard their work.
```

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ * `producer(Job()) { failure() }` makes sure that the only effect of `failure()` is to make *this* `producer` finish
```
murfel: Ditto, make it extremely clear that this should not be used.

"While X makes sure ..., it violates structured concurrency and hence should not be used"

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ * ```
+ * val childSupervisorScope = CoroutineScope(
+ *     scope.coroutineContext + SupervisorJob(scope.coroutineContext.job)
+ * )
+ * childSupervisorScope.produce {
+ *     // failures in this coroutine will not affect other children
+ * }
+ * ```
+ *
+ * For a lexically scoped [CoroutineScope], using a [supervisorScope] at the end of the outer scope may help:
```
murfel:
```suggestion
 * For a lexically scoped [CoroutineScope], one of the options is to use a [supervisorScope] at the end of the outer scope:
```

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ *     launch {
+ *         // failures in this coroutine will affect everyone
+ *     }
+ *     supervisorScope {
+ *         val channel = produce {
+ *             // failures in this coroutine
+ *             // are only available through `channel`
+ *         }
+ *     }
+ *     // this line will only be reached when the `produce` coroutine complete
```
murfel: complete**s**

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ * }
+ * // this line will be reached when both `launch` and `produce` complete
+ * ```
+ *
+ * All of these approaches preserve the ability of a parent to cancel the children and to wait for their completion.
```
murfel, line 633: I like this very succinct summary of why we need to strive to preserve structured concurrency.

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ * ### Avoiding both cancelling and being cancelled
+ *
+ * Sometimes, coroutines to be spawned are just completely unrelated to the [CoroutineScope] used as the receiver,
+ * and no structured concurrency mechanisms are needed.
+ *
+ * In that case, [GlobalScope] is the semantically clearer way of expressing opting out of structured concurrency:
+ *
+ * ```
+ * GlobalScope.produce {
+ *     // this is explicitly a rogue computation
```
murfel: I don't like the term _rogue_. It's emotionally charged and doesn't hint at what it means. Possibly, an orphaned coroutine would work better? Did we leak this to outside documentation already, or is there a room for negotiation?

In this particular case no need to decide now, we could just say: this computation doesn't have a parent, and thus cannot be cancelled.

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
```diff
@@ -506,3 +506,140 @@ internal fun <E, R, V> ReceiveChannel<E>.zip(
...
+ * ### Avoiding cancellation
+ *
+ * Sometimes, it is undesirable for the child coroutine to react to the cancellation of the parent: for example,
+ * some computations have to be performed either way.
+ * `produce(NonCancellable)` or `produce(Job())` can be used to achieve this effect.
+ *
+ * Alternative approaches that preserve structured concurrency are this:
+ *
+ * ```
+ * // Guarantees the completion
```
murfel: A bit verbose given L550?
If you decide to keep, then also add it in launch.

kotlinx-coroutines-core/common/src/Guidance.kt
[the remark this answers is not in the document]
murfel: This was also confusing to me. A simple comment on this line `// joining failed supervised coroutines doesn't fail either` would help!

murfel (changes requested):

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
murfel: See a note on lifecycle below, related to this section.

murfel: These two notes are the only non-trivial comments, requesting changes in this section (and other sections referring to this one.)

This section establishes the ground rules, and the following sections should refer to arguments listed here.

The whole narrative is like this:
- Passing a Job explicitly prevents structured concurrency, don't do it, we are about to give you a collection of samples to use instead.
- Defining structured concurrency (parent-child relationship), various implications of parent-child relationship, and what scenarios they prevent or enable.
- List of samples, each sample demonstrates how it preserves structured concurrency.

Currently, the structured concurrency section defines things more loosely and doesn't precisely match with the arguments in the samples.

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
dkhalanskyjb: There isn't any way to affect the resolution here, unfortunately. Not only do we not have control over this, but also, the resolution results may be different between the IDE and the rendered html documents.

In such cases, there's an argument to be made for not leaving a link at all. However, my preferred design for the improved KDoc navigability includes warnings for ambiguous cases. With that in mind, I'm hopeful that one day, after we upgrade to a new Kotlin version, we'll see a hundred "ambiguous reference in KDoc" warnings and will be able to easily audit and fix them.

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
dkhalanskyjb: The target audience for this is people who don't have a solid grasp on structured concurrency yet. To them, "parent-child relationship" may be a phrase devoid of special meaning. The dramatic "severs the tie" phrasing is intended to clearly describe the stakes: one piece of code is completely unaware of the other. The more experienced users can be expected to understand the specifics from the "sole parent" remark.

kotlinx-coroutines-core/common/src/channels/Deprecated.kt
dkhalanskyjb: The more robust separation would be 1,2,3-cancellation, 4-completion, because in 3, the cancellation also goes the other way, from the child to the parent. I don't understand the benefits of such a grouping, though.

In response to your proposal, I did change the structure a bit to flow more naturally, though I'm not sure that it aligns with what you had in mind.

PR #4437, opened 2025-05-15
docs: update `ThreadLocal<T>.asContextElement` documentation

nikitabobko:
The latest code snippet in the KDoc is no longer relevant. Coroutines update and restore the thread-local value on every suspension points.

The code became irrelevant after
727c38f708c355e325326def66c80dfec7f81ebe (related issue: https://github.com/Kotlin/kotlinx.coroutines/issues/985)

> Use `withContext` to update the corresponding thread-local variable to a different value, for example:

The commit drops this paragraph and the attached code snippet. The snippet is useless since the correct usage pattern is already demonstraited in the first code snippet.


qwwdfsad (approved):
Thanks! Better late than never 😃

Issue #4242, opened 2024-10-04
Consider marking `runBlocking` as `@DelicateCoroutinesApi`


joffrey-bion:
From the [runBlocking documentation](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html):

> It is designed to bridge regular blocking code to libraries that are written in suspending style, to be used in main functions and in tests

Part of this is obsolete now. We have `suspend fun main` to get a suspending context right from the entrypoint. We have `kotlinx-coroutines-test` for tests.

This leaves the "briding" use case as the remaining valid case for `runBlocking`, but this is quite general/vague and `runBlocking` is still dangerous in many such cases. We have already seen a few nasty cases where `runBlocking` is synonym of trouble:
* https://github.com/Kotlin/kotlinx.coroutines/issues/4215
* https://github.com/Kotlin/kotlinx.coroutines/issues/3982

It would be worth documenting the cases that are safe for `runBlocking`, and the cases that we know are unsafe and have alternatives. A bit like what was done for `GlobalScope`.

My intuition is that there aren't many cases where `runBlocking` is the "clean" way to go. In that sense, I believe it would be a net positive to mark `runBlocking` as `@DelicateCoroutinesApi`, with an updated doc to explain more details.

qwwdfsad:
That would be a nice idea if we were just to (re-)introduce `run-blocking`.

The problem is, it exists for quite a few years already, is *really* widespread (i.e. single DuckDuckGo app has [83](https://grep.app/search?q=runBlocking&filter[lang][0]=Kotlin&filter[repo][0]=duckduckgo/Android) usages) and already leveraged in coursed, lectures, talks and books.

Marking it as a delicate API will likely create a serious disruption and a lot of immediate opt-ins (potentially module-wide), which will rather undermine the meaning and intent of `@DelicateCoroutinesApi`

joffrey-bion:
This is fair enough. Then here are my suggestions:

* my points about docs still hold. I think the documentation of `runBlocking` should not mention obsolete use cases like `main` functions and tests, it should clearly warn about potential issues with unintentionally nested `runBlocking` calls (currently it's showing an obvious case instead of warning about non-obvious ones), and provide examples of right and wrong usages
[...]

PR #4440, opened 2025-05-21
Add `doBlocking()` utility method for running coroutines.

kevinhwang:
This is a useful utility method used in places where you want to run a coroutine for its side effect.

E.g.,

```kt
fun main() = doBlocking {
  // ...
}
```

```kt
@Test
fun myTest() = doBlocking {
  // ...
}
```

With `runBlocking()`, if the last line of the coroutine block being run doesn't evaluate to `Unit`, e.g.:

```kt
fun main() {
  runBlocking {
    // ...
    Unit
  }
}
```

the compiler will complain about being unable to infer the type variable T of the block. Inserting a "return value" of `Unit` at the end is necessary to tell the compiler the block's return type.

Similarly, if you do

```kt
@Test
fun myTest() = runBlocking {
  // ...
   Unit
}
```

without the explicit `Unit` "return value" at the end, if the last line evaluates to a non-unit type, you'll get a type mismatch error, or the compiler will infer `myTest()`'s return type to be non-Unit.

### DIFF
--- a/kotlinx-coroutines-core/concurrent/src/Builders.concurrent.kt
+++ b/kotlinx-coroutines-core/concurrent/src/Builders.concurrent.kt
@@ -34,6 +34,6 @@ public expect fun <T> runBlocking(context: CoroutineContext = EmptyCoroutineCont
  * fun foobar() = runBlocking { ... }
  * ```
  */
-public fun <T> doBlocking(context: CoroutineContext = EmptyCoroutineContext, block: suspend CoroutineScope.() -> Unit) {
+public fun doBlocking(context: CoroutineContext = EmptyCoroutineContext, block: suspend CoroutineScope.() -> Unit) {
   runBlocking(context, block)
 }
--- a/kotlinx-coroutines-core/concurrent/src/Builders.concurrent.kt
+++ b/kotlinx-coroutines-core/concurrent/src/Builders.concurrent.kt
@@ -22,3 +22,18 @@ import kotlin.coroutines.*
  * block, potentially leading to thread starvation issues.
  */
 public expect fun <T> runBlocking(context: CoroutineContext = EmptyCoroutineContext, block: suspend CoroutineScope.() -> T): T
+
+/**
+ * Runs the given coroutine on the current thread, blocking until completion.
+ *
+ * This is like [runBlocking], but returns [Unit] which is a useful utility to run coroutines for their side effect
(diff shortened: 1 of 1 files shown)


dkhalanskyjb:
The expected way to use `runBlocking` in `main` is
```kotlin
fun main() {
    runBlocking {
        // your code
    }
}
```
> the compiler will complain about being unable to infer the type variable T of the block.

Could you provide a specific example?

kevinhwang:
> The expected way to use `runBlocking` in `main` is

Within Google it's common to define [single-expression functions](https://kotlinlang.org/docs/functions.html#single-expression-functions) using `runBlocking` and `doBlocking` to run a coroutine from a non-coroutine function (e.g., main, or a unit test method) and block on it:

```kotlin
fun main() = doBlocking {
  // ...
}
```

It's a nice short-hand and convenient, leading to one less level of nested braces / scopes to be read. But `runBlocking` causes problems if the block passed to it evaluates to anything other than `Unit`—then type inference will cause the function (e.g., `main`, or the test method) to have a non-`Unit` return type.

> Could you provide a specific example?

Yeah, see https://stackoverflow.com/questions/58875520.

dkhalanskyjb:
> Yeah, see https://stackoverflow.com/questions/58875520.

Could not reproduce. The code *compiles* just fine for me, the tests don't run because of JUnit's logic.

> But `runBlocking` causes problems if the block passed to it evaluates to anything other than `Unit`

Yes. When using `runBlocking` as a statement and not a (value-returning) expression, the single-*expression* form should be avoided.

I understand the point about the single-expression form being more convenient (in fact, our own codebase does use this form non-idiomatically to save an indentation level even when we return `Unit`), and typically, we do try to introduce shorthand forms for common idioms to streamline the process of writing code.

`runBlocking` is special, however. Because of how widely misused `runBlocking` is (https://github.com/Kotlin/kotlinx.coroutines/issues/4242), making using it (or functions similar to it) more streamlined goes against our goals. In the linked issue, we are actually discussing how to make it *less* streamlined.

I can imagine us adding a `withContext` form that doesn't return a value (like `launch` can be thought of as a non-value-returning `async`), for example, to handle this case:
```kotlin
suspend fun main() = withContext(Dispatchers.IO) {
    // something
}
```
I haven't seen this problem in real code yet, though.

PR #4445, opened 2025-05-30
Use cache redirector to download Gradle

kbrgnj:
KTI-1836


dkhalanskyjb:
Replied under https://youtrack.jetbrains.com/issue/KTI-1836

PR #4460, opened 2025-06-18
Convert return in expression body to labeled return in preparation for KT-22786

cypressious:


dkhalanskyjb:
The change in https://github.com/Kotlin/kotlinx.coroutines/pull/4460/commits/35be629f8ceb1318f3391bd921ce1d7be723a534 is already in `develop`.

PR #4464, opened 2025-06-20
Fix deprecation warnings and enable warningsAsErrors for buildSrc

ALikhachev:
K/JS related deprecation warnings are to be promoted to error within KT-68597 in Kotlin 2.2


buildSrc/src/main/kotlin/CacheRedirector.kt
```diff
@@ -100,16 +100,30 @@ private fun Project.checkRedirect(repositories: RepositoryHandler, containerName
...
-        yarnRootExtension?.downloadBaseUrl?.let {
-            yarnRootExtension.downloadBaseUrl = CacheRedirector.maybeRedirect(it)
+        plugins.withType(YarnPlugin::class) {
+            extensions.configure(YarnRootEnvSpec::class.java) {
+                // no API to modify the value in-place keeping it lazy: https://github.com/gradle/gradle/issues/27227
+                val currentValue = downloadBaseUrl.orNull
+                if (currentValue != null) {
+                    downloadBaseUrl = CacheRedirector.maybeRedirect(currentValue)
+                }
+            }
```
dkhalanskyjb: Can this not be written like it was before? That is,
```suggestion
                downloadBaseUrl.orNull?.let {
                    downloadBaseUrl = CacheRedirector.maybeRedirect(it)
                }
            }
```
I'd expect this to be a pure Kotlin construction, without any relation to what Gradle supports or doesn't support.

ALikhachev: Sure, it can be. If you prefer this style – I can update it. The comment above relates to the fact that `orNull` resolves the value of the property eagerly instead of applying a dynamic transformator on it

dkhalanskyjb: I somewhat prefer the current style, but it's not important. I just got confused about what part of the change was important and which accidental.

ALikhachev: I've returned the original style

PR #4467, opened 2025-07-01
Increase Lincheck version to 3.0

ndkoval:


kotlinx-coroutines-core/build.gradle.kts
```diff
@@ -252,6 +252,7 @@ kover {
 
             // lincheck has NPE error on `ManagedStrategyStateHolder` class
             excludedClasses.addAll("org.jetbrains.kotlinx.lincheck.*")
```
dkhalanskyjb, line 254: Historical note: https://repo.maven.apache.org/maven2/org/jetbrains/lincheck/lincheck/3.0/lincheck-3.0.jar does contain classes in both `org.jetbrains.lincheck` and `org.jetbrains.kotlinx.lincheck` packages, so keeping these exclusions is intentional.

kotlinx-coroutines-core/jvm/test/lincheck/ChannelsLincheckTest.kt
```diff
@@ -154,9 +149,6 @@ abstract class ChannelLincheckTestBase(
     // @Operation TODO non-linearizable in BufferedChannel
     open fun isEmpty() = c.isEmpty
 
-    @StateRepresentation
-    fun state() = (c as? BufferedChannel<*>)?.toStringDebug() ?: c.toString()
```
dkhalanskyjb, line 158: With this change, should we also remove `toStringDebug()` itself? If not, then at least the comment to that function should be updated.

dkhalanskyjb (commented):
It's alarming that the CI consistently takes about ~20-30 minutes longer with this change. What could be the reason?

dkhalanskyjb:
Also, on my machine, `./gradlew clean check -Pstress=true` fails. The following tests have problems:
```
    Buffered1BroadcastChannelLincheckTest. modelCheckingTest[jvm]
    Buffered1ChannelLincheckTest. modelCheckingTest[jvm]
    Buffered2BroadcastChannelLincheckTest. modelCheckingTest[jvm]
    Buffered2ChannelLincheckTest. modelCheckingTest[jvm]
    ConflatedBroadcastChannelLincheckTest. modelCheckingTest[jvm]
    ConflatedChannelLincheckTest. modelCheckingTest[jvm]
    RendezvousChannelLincheckTest. modelCheckingTest[jvm]
    Semaphore1LincheckTest. modelCheckingTest[jvm]
    Semaphore2LincheckTest. modelCheckingTest[jvm]
    UnlimitedChannelLincheckTest. modelCheckingTest[jvm]
```
One example of a problem: https://gist.github.com/dkhalanskyjb/528602e1adfdb402b14bcdef11a5ad44

dkhalanskyjb:
Sorry, disregard my latest message: the same happens on `develop` as well. I think I haven't run `clean check -Pstress=true` for some time, and something may have regressed in my configuration.

PR #4468, opened 2025-07-01
Add lint warnings on SharedFlow operations that never complete - SharedFlow.last()

### LINKED ISSUE #3275: SharedFlow last() never returns
Hi,

when using SharedFlows the function last() delivered by Reduce.kt through extensions will never return. It collects the flow and suspends until it is completed, but SharedFlows never complete. 
For intuitive usage last() in SharedFlows should be a shortcut for replayCache.last()

Thanks!

murfel:
Fixes #3275

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
@@ -186,3 +186,15 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
 @InlineOnly
 public suspend inline fun <T> SharedFlow<T>.count(): Int =
     (this as Flow<T>).count()
+
+/**
+ * @suppress
+ */
+@Suppress("DeprecatedCallableAddReplaceWith")
+@Deprecated(
+    level = DeprecationLevel.WARNING,
+    message = "SharedFlow never completes, so this terminal operation never completes. Use 'replayCache.last()' for reading the last emitted item in the replay cache.",
+    replaceWith = ReplaceWith("replayCache.last()")
+)
+@InlineOnly
+public fun <T> SharedFlow<T>.last(): T = noImpl()


kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -186,3 +186,15 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
 @InlineOnly
 public suspend inline fun <T> SharedFlow<T>.count(): Int =
     (this as Flow<T>).count()
+
+/**
+ * @suppress
+ */
+@Suppress("DeprecatedCallableAddReplaceWith")
+@Deprecated(
+    level = DeprecationLevel.WARNING,
```
dkhalanskyjb, line 193: Good call making this a `WARNING` and not an `ERROR`. Wouldn't be nice to break the compilation of someone's code using `last()` instead of `collect()`.

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -186,3 +186,15 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
...
+    message = "SharedFlow never completes, so this terminal operation never completes. Use 'replayCache.last()' for reading the last emitted item in the replay cache.",
+    replaceWith = ReplaceWith("replayCache.last()")
```
dkhalanskyjb: `ReplaceWith` should typically be used for suggesting code that's equivalent to the deprecated one, but more explicitly written to highlight the problem. Example: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/capitalize.html
The reason is that people can perform replacements without looking closely, and the changed behavior can be a surprise. The proper replacement would be something like `(this as Flow<T>).last()`, but I'm not sure it's worth providing, as it is meaningless anyway.

Another problem with this replacement specifically is that `replayCache` is often empty (for example, https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow.html has its size `0` by default), and `last()` throws exceptions on empty lists.

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -186,3 +186,15 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
...
 public suspend inline fun <T> SharedFlow<T>.count(): Int =
     (this as Flow<T>).count()
+
+/**
+ * @suppress
+ */
+@Suppress("DeprecatedCallableAddReplaceWith")
+@Deprecated(
+    level = DeprecationLevel.WARNING,
+    message = "SharedFlow never completes, so this terminal operation never completes. Use 'replayCache.last()' for reading the last emitted item in the replay cache.",
```
dkhalanskyjb: One use case for `sharedFlow.last()` that comes to mind is to subscribe to a shared flow so that it "starts working" (as shared flows can be configured behave differently depending on whether there are subscribers: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-sharing-started/), but doesn't actually do anything. In other words, subscribing for the side effects of subscribing and not for the value. In that scenario, the idiomatic approach is to call `sharedFlow.collect()`. This could be worth highlighting in the replacement message.

dkhalanskyjb: This line is too long: we try to stick to 120 characters per line, and typically, we don't even fill that limit to the brim, instead breaking strings on punctuation (dots, commas...). Example: https://github.com/Kotlin/kotlinx.coroutines/pull/4433/files#diff-c6a5519f3b99094bab5ee5f1797928d8ed5df48460d0840e642bcf7da495a6dbR61-R64 The reason is to avoid large git diffs when adding some new words in the middle of the text and causing a cascade of changes due to reformatting.

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -186,3 +186,15 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
...
+    replaceWith = ReplaceWith("replayCache.last()")
+)
+@InlineOnly
+public fun <T> SharedFlow<T>.last(): T = noImpl()
```
dkhalanskyjb: `noImpl()` does not go well with `DeprecationLevel.WARNING`. It's either `noImpl()` + `ERROR` (so that code doesn't work + doesn't compile) or a proper implementation + `WARNING` (so that code works + compiles). Here, `WARNING` is the better choice, given that some code may already call this without encountering any real problems, so the implementation should be fixed. `count()` just above is a good inspiration.

dkhalanskyjb (changes requested):

dkhalanskyjb:
Also, the build fails with this change. Please take a look.

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -186,3 +186,16 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
 @InlineOnly
 public suspend inline fun <T> SharedFlow<T>.count(): Int =
     (this as Flow<T>).count()
+
+/**
+ * @suppress
+ */
+@Suppress("DeprecatedCallableAddReplaceWith")
```
dkhalanskyjb: The purpose of this annotation is that in `kotlinx.coroutines`, all warnings in the non-test source sets lead to a build failure, so we have to suppress false warnings and can not just ignore them.
That said, the warning suppressed by this no longer exists, so this line does nothing.
Could you please remove this suppression *everywhere in the codebase*?

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
```diff
@@ -178,11 +176,22 @@ public suspend inline fun <T> SharedFlow<T>.toSet(destination: MutableSet<T>): N
...
     message = "SharedFlow never completes, so this terminal operation never completes.",
     level = DeprecationLevel.WARNING
 )
 @InlineOnly
 public suspend inline fun <T> SharedFlow<T>.count(): Int =
     (this as Flow<T>).count()
+
+/**
+ * @suppress
+ */
```
murfel, line 189: What does this comment do, by the way?

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
murfel: I wonder if it is feasible to have an automated check which asserts that every `@Suppress` suppresses an existing warning?

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
murfel: This
image

I looked up all the other usages. I don't understand the semantics of this field (looks like just a comment?), and, in particular, its meaning when there's no text following it.

dkhalanskyjb: It's this: https://kotlinlang.org/docs/kotlin-doc.html#suppress
For example, you won't see the `count` overload defined just above in this list: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/count.html

dkhalanskyjb (changes requested):
This is almost ready to go, but I've noticed it targets the `master` branch. The ongoing development happens in `develop`, and `master` contains the latest release. It does happen occasionally that targeting `master` is the better choice, but not in this case. Please see https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md

kotlinx-coroutines-core/common/src/flow/operators/Lint.kt
dkhalanskyjb: Maybe, but there are some challenges in the way of this.
Who should supply this check? The compiler or the IDE?
- The IDE: in fact, we used to have this inspection at some point, but it no longer works now (I don't know if that's intentional or just a regression). I actually prefer not having this inspection, as it was more of a nuisance: the IDE is published out of sync with the Kotlin compiler, and their knowledge of what is a warning and what isn't can differ. So, you'd get inspections like "this suppression isn't suppressing anything" when the suppression is actually useful, and vice versa.
- The compiler: it could probably emit some diagnostic message like "I do not need this suppression, your code is fine as is", but some projects are configured so that they can be compiled on several versions of the compiler, and it would be unpleasant to have a set of suppresses that are needed on one version but not the other emit diagnostic messages. This is just noise that you can't do anything about. Except maybe add a construct like `Suppress("REDUNDANT_SUPPRESS")`?

murfel: I wish there was a button to run it on demand, at least. That would be useful at least for kotlinx.coroutines.

Also, I wonder how c++ world deals with it.

I searched briefly for an issue on this in YouTrack but wasn't successful. If you happen to ever stumble upon it again, please tune me in!

murfel: While theoretically it's an unsolvable issue, I feel like there's a lot of low-hanging fruits with minimal annoying impact that could be done.

dkhalanskyjb: I haven't actually seen a YouTrack issue, just observed that the behavior disappeared.

> I wish there was a button to run it on demand, at least.

Great point! There is such a button. I guess the inspection was left in, but got disabled. In `Code → Analyze code → Run inspection by name`, there `Redundant suppression` inspection is listed, and it does find the affected code (and a lot of autogenerated code as well).

PR #4472, opened 2025-07-07
Propagate exception to wasm-js and js in `propagateExceptionFinalResort`

### LINKED ISSUE #4451: propagateExceptionFinalResort implementation for wasm-js and js confusing users
Both implementations call `console.error` and pass `exception.toString()`, which is basically a name of Exception class + message. In addition, `console.error` prints a stack trace from a call site (not from the exception!!), which users often consider a stack trace from the exception ([example](https://youtrack.jetbrains.com/issue/KT-77801)).

In [this project](https://github.com/user-attachments/files/20628501/KotlinProject.250602.2.zip), you can find a project with a simple repro and possible improvements implemented via `CoroutineExceptionHandler`.

I think `#1` or any other way to rethrow the exception is the best option since there are standard ways inside JS runtimes to process unhandled exceptions. These might be used directly or by some library intended to handle problems on a page. In the worst case, a browser (runtime) will show the error in a standard and more convenient way.

murfel:
Fixes #4451

### DIFF
--- a/kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
+++ b/kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
@@ -1,0 +1,45 @@
+package kotlinx.coroutines
+
+import kotlinx.coroutines.testing.*
+import kotlin.js.*
+import kotlin.test.*
+
+class PropagateExceptionFinalResortTest : TestBase() {
+    @BeforeTest
+    fun setupExceptionListenerOnly() {
+        // Remove a Node.js's internal listener, which prints the exception to stdout.
+        js("""
+            globalThis.originalListeners = process.listeners('uncaughtException');
+            process.removeAllListeners('uncaughtException');
+        """)
+    }
+
+    @AfterTest
+    fun restoreListeners() {
+        js("""
+            if (globalThis.originalListeners) {
+                process.removeAllListeners('uncaughtException');
+                globalThis.originalListeners.forEach(function(listener) {
+                    process.on('uncaughtException', listener);
+                });
+            }
+        """)
+    }
+
+    @Test
+    fun testThrows() = runTest {
+        js("""
+            globalThis.exceptionCaught = false;
+            process.on('uncaughtException', function(e) {
+                globalThis.exceptionCaught = true;
+            });
+        """)
+        val job = GlobalScope.launch {
+            throw IllegalStateException("My ISE")
+        }
+        job.join()
+        delay(1)
+        val exceptionCaught = js("globalThis.exceptionCaught") as Boolean
+        assertTrue(exceptionCaught)
+    }
+}
--- a/kotlinx-coroutines-core/jsAndWasmJsShared/src/internal/CoroutineExceptionHandlerImpl.kt
+++ b/kotlinx-coroutines-core/jsAndWasmJsShared/src/internal/CoroutineExceptionHandlerImpl.kt
@@ -1,0 +1,14 @@
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.*
+
+internal expect interface JsAny
+
+internal expect fun Throwable.toJsException(): JsAny
+
+internal fun throwAsync(e: JsAny): Unit = js("setTimeout(function () { throw e }, 0)")
+
+internal actual fun propagateExceptionFinalResort(exception: Throwable) {
+    throwAsync(exception.toJsException())
+
+}
--- a/kotlinx-coroutines-core/wasmJs/src/internal/CoroutineExceptionHandlerImpl.kt
+++ b/kotlinx-coroutines-core/wasmJs/src/internal/CoroutineExceptionHandlerImpl.kt
@@ -1,8 +1,17 @@
 package kotlinx.coroutines.internal
 
-import kotlinx.coroutines.*
+internal actual typealias JsAny = kotlin.js.JsAny
 
-internal actual fun propagateExceptionFinalResort(exception: Throwable) {
-    // log exception
-    console.error(exception.toString())
+internal actual fun Throwable.toJsException(): JsAny {
+    return toJsError(message, this::class.simpleName, stackTraceToString())
+}
+
+internal fun toJsError(message: String?, className: String?, stack: String?): JsAny {
+    js("""
+    const error = new Error();
+    error.message = message;
+    error.name = className;
+    error.stack = stack;
+    return error;
+    """)
 }
--- a/kotlinx-coroutines-core/wasmJs/test/PropagateExceptionFinalResortTest.kt
+++ b/kotlinx-coroutines-core/wasmJs/test/PropagateExceptionFinalResortTest.kt
@@ -1,0 +1,28 @@
+package kotlinx.coroutines
+
+import kotlinx.coroutines.testing.TestBase
+import kotlin.test.*
+
+fun addUncaughtExceptionHandler() {
+    js("""
+        globalThis.exceptionCaught = false;
+        process.on('uncaughtException', function(e) {
+            globalThis.exceptionCaught = true;
+        });
+    """)
+}
+
+fun exceptionCaught(): Boolean = js("globalThis.exceptionCaught")
+
+class PropagateExceptionFinalResortTest : TestBase() {
+    @Test
+    fun testThrows() = runTest {
+        addUncaughtExceptionHandler()
+        val job = GlobalScope.launch {
+            throw IllegalStateException("My ISE")
+        }
+        job.join()
+        delay(1) // Give time for the error handler to execute
+        assertTrue(exceptionCaught())
+    }
+}
--- a/kotlinx-coroutines-core/js/src/internal/CoroutineExceptionHandlerImpl.kt
+++ b/kotlinx-coroutines-core/js/src/internal/CoroutineExceptionHandlerImpl.kt
@@ -1,8 +1,7 @@
 package kotlinx.coroutines.internal
 
-import kotlinx.coroutines.*
+import kotlin.js.unsafeCast
 
-internal actual fun propagateExceptionFinalResort(exception: Throwable) {
-    // log exception
-    console.error(exception.toString())
-}
+internal actual external interface JsAny
+
+internal actual fun Throwable.toJsException(): JsAny = this.unsafeCast<JsAny>()


kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,45 @@
...
+            globalThis.exceptionCaught = false;
+            process.on('uncaughtException', function(e) {
+                globalThis.exceptionCaught = true;
+            });
+        """)
+        val job = GlobalScope.launch {
+            throw IllegalStateException("My ISE")
+        }
+        job.join()
+        delay(1)
```
dkhalanskyjb: Why is this needed? Please add a comment to the code.

kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,45 @@
...
+                process.removeAllListeners('uncaughtException');
+                globalThis.originalListeners.forEach(function(listener) {
+                    process.on('uncaughtException', listener);
+                });
+            }
+        """)
+    }
+
+    @Test
+    fun testThrows() = runTest {
```
dkhalanskyjb: I'd expect a more descriptive name, like `testHandlingUncaughtExceptionsInNodeJs`. Maybe even a documentation comment explaining what exactly we are checking here. Given how much JS code there is in this file, it's not trivial to figure out.

kotlinx-coroutines-core/jsAndWasmJsShared/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,14 @@
...
+
+internal expect interface JsAny
+
+internal expect fun Throwable.toJsException(): JsAny
+
+internal fun throwAsync(e: JsAny): Unit = js("setTimeout(function () { throw e }, 0)")
+
+internal actual fun propagateExceptionFinalResort(exception: Throwable) {
+    throwAsync(exception.toJsException())
+
```
dkhalanskyjb: Minor: unnecessary newline.

kotlinx-coroutines-core/jsAndWasmJsShared/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,14 @@
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.*
+
+internal expect interface JsAny
+
+internal expect fun Throwable.toJsException(): JsAny
+
+internal fun throwAsync(e: JsAny): Unit = js("setTimeout(function () { throw e }, 0)")
```
dkhalanskyjb, line 13: Please add a comment explaining the principle behind this solution. It looks mysterious: why set something to run after a timeout of *zero*?

kotlinx-coroutines-core/wasmJs/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -1,8 +1,17 @@
...
 
-import kotlinx.coroutines.*
+internal actual typealias JsAny = kotlin.js.JsAny
 
-internal actual fun propagateExceptionFinalResort(exception: Throwable) {
-    // log exception
-    console.error(exception.toString())
+internal actual fun Throwable.toJsException(): JsAny {
+    return toJsError(message, this::class.simpleName, stackTraceToString())
+}
```
dkhalanskyjb: Minor: can this be an expression body?

kotlinx-coroutines-core/wasmJs/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,28 @@
+package kotlinx.coroutines
+
+import kotlinx.coroutines.testing.TestBase
+import kotlin.test.*
+
+fun addUncaughtExceptionHandler() {
```
dkhalanskyjb: Worth marking such things as `private` to avoid calling this function specific to one test from other tests.

kotlinx-coroutines-core/wasmJs/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,28 @@
...
+    js("""
+        globalThis.exceptionCaught = false;
+        process.on('uncaughtException', function(e) {
+            globalThis.exceptionCaught = true;
+        });
+    """)
+}
+
+fun exceptionCaught(): Boolean = js("globalThis.exceptionCaught")
```
dkhalanskyjb: Minor: stylistically, such helper functions go to the bottom of the file, so that the important thing (the test itself) is front and center when you open the file.

dkhalanskyjb (changes requested):
Makes sense, thanks!

kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
murfel: I feel like test name is too short to include a very specific descriptive name, so I've only added a comment. Do you want me to rename the test name anyway into the next best thing? (I assume with the goal of having a better name should this test fail.)

murfel:
Idea's internal formatter (CMD+Option+L) wants to format all `js("""...""")` blocks like this:

```
js(
    """
    globalThis.originalListeners = process.listeners('uncaughtException');
    process.removeAllListeners('uncaughtException');
"""
)
```

Instead of the simpler this, which is also the way it is written in docs.
```
js("""
    globalThis.originalListeners = process.listeners('uncaughtException');
    process.removeAllListeners('uncaughtException');
""")
```

Should we override the formatter, locally or globally, for this particular function?

dkhalanskyjb:
What the IDE does is not the style used in our style guide: https://kotlinlang.org/docs/coding-conventions.html#strings Worth filing a bug, but I'm not sure who is wrong: the IDE, the style guide, or both.

The important thing, though, is for a project to have a consistent style, the specifics don't matter that much. In `kotlinx.coroutines`, we don't yet have a style for `"""` (grepping `"""`, I see all sorts of formatting), so if you have a preference, please feel free to declare it the official `kotlinx.coroutines` style for `"""` and apply it everywhere.

kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
dkhalanskyjb: Yes, having a rough idea of what failed when you see the list of failed tests after introducing a change is the point.

kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,50 @@
...
+            globalThis.exceptionCaught = false;
+            process.on('uncaughtException', function(e) {
+                globalThis.exceptionCaught = true;
+            });
+        """)
+        val job = GlobalScope.launch {
+            throw IllegalStateException("My ISE")
+        }
+        job.join()
+        delay(1)  // Let the exception be re-thrown and handled.
```
dkhalanskyjb: Minor: an extra space.

kotlinx-coroutines-core/wasmJs/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,33 @@
...
+     * It is checked by setting up an exception handler within Wasm/JS.
+     */
+    @Test
+    fun testThrows() = runTest {
+        addUncaughtExceptionHandler()
+        val job = GlobalScope.launch {
+            throw IllegalStateException("My ISE")
+        }
+        job.join()
+        delay(1)  // Let the exception be re-thrown and handled.
```
dkhalanskyjb: Ditto.

kotlinx-coroutines-core/wasmJs/test/PropagateExceptionFinalResortTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+class PropagateExceptionFinalResortTest : TestBase() {
+    /*
+     * Test that `propagateExceptionFinalResort` re-throws the exception on Wasm/JS.
+     *
+     * It is checked by setting up an exception handler within Wasm/JS.
+     */
+    @Test
+    fun testThrows() = runTest {
+        addUncaughtExceptionHandler()
```
dkhalanskyjb: I wonder if we want to remove it afterward. Given that it doesn't seem like uncaught exceptions do *anything* on Wasm/JS by default, maybe it doesn't matter. Still, avoiding test cross-contamination is worth it so that this doesn't bite us when we forget about this test completely.

kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
murfel: It's just that all the info is in the class name + path (for platform), and naming this accurately would be duplicating those. Had we had a dozen tests, we wouldn't be naming them so verbosely. I've done it anyways for now.

kotlinx-coroutines-core/js/test/PropagateExceptionFinalResortTest.kt
murfel: Oh, double space before comment was intentional. I think I got it from Python.

I wish the IDE reformat action would point this out, too!

Fixed to single space.

dkhalanskyjb: No big deal, this is just not the style adopted in our library.

murfel: Can we enforce it in automated checks?

dkhalanskyjb: I'm open to this (and preventing merging if the style is incorrect), but only if this check doesn't make the CI fail for intermediate commits. It would be annoying to push code just to test something, without ensuring the code is pretty, only to be met with a useless style check complaint. That said, it makes sense to have this check block merging.

fzhinkin: Note that instead of making such a check part of the regular CI pipeline, or a pre-commit hook, it could also be implemented as a GH Action, (I'm not 100% sure about this part, but) that'll report all problems right into the PR.

dkhalanskyjb (changes requested):
This also needs to be documented in all related places. At least the `CoroutineExceptionHandler` documentation is mentioning this.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -83,14 +83,15 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
+ * When no handler is installed, an exception is handled in the following way:
+ * - If the exception is [CancellationException], it is ignored, as these exceptions are used to cancel coroutines.
  * - Otherwise, if there is a [Job] in the context, then [Job.cancel] is invoked.
  * - Otherwise, as a last resort, the exception is processed in a platform-specific manner:
  *   - On JVM, all instances of [CoroutineExceptionHandler] found via [ServiceLoader], as well as
  *     the current thread's [Thread.uncaughtExceptionHandler], are invoked.
  *   - On Native, the whole application crashes with the exception.
- *   - On JS, the exception is logged via the Console API.
+ *   - On JS and Wasm/JS, the exception is propagated into the event loop
+ *     and is processed in a platform-specific way determined by the platform itself.
```
murfel: I need some help with the wording, though. I'm unsure on how we spell Wasm/JS (because there's Wasm/WASI) (if Wasm/JS is ok, would the full spelling be Kotlin/Wasm/JS ? That doesn't look great)

Also, not sure saying "event loop" is a fair representation

dkhalanskyjb: I also don't know what it's called, but it's "Kotlin/Wasm-JavaScript" and "Kotlin/Wasm and WASI", according to https://kotlinlang.org/docs/wasm-overview.html

"Event loop" is okay, but it's better to specify which one: "the JavaScript runtime's event loop" or something similar.

murfel: I'll spell it "JS and Wasm JS". "Kotlin/Wasm-JavaScript" looks a bit overcrowded, and is only mentioned once on the website ([google search](https://www.google.com/search?q=%22Wasm-JavaScript%22+site%3Ahttps%3A%2F%2Fkotlinlang.org%2Fdocs%2F&sca_esv=4ab2dcfe03ad0fc5&ei=GcN4aJPTM_GwhbIP6O7M2QM&ved=0ahUKEwjTw4T4ysOOAxVxWEEAHWg3MzsQ4dUDCBA&uact=5&oq=%22Wasm-JavaScript%22+site%3Ahttps%3A%2F%2Fkotlinlang.org%2Fdocs%2F&gs_lp=Egxnd3Mtd2l6LXNlcnAiMyJXYXNtLUphdmFTY3JpcHQiIHNpdGU6aHR0cHM6Ly9rb3RsaW5sYW5nLm9yZy9kb2NzL0jUElCFB1iLEXADeACQAQCYASigASiqAQExuAEDyAEA-AEBmAIAoAIAmAMAiAYBkgcAoActsgcAuAcAwgcAyAcA&sclient=gws-wiz-serp))

murfel: this table clarified things a bit for me (Kotlin/Wasm is a target platform, while `wasmJs` and `wasmWasi` are targets )

https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-dsl-reference.html#targets

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -83,14 +83,15 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
- * - If exception is [CancellationException], it is ignored, as these exceptions are used to cancel coroutines.
+ * When no handler is installed, an exception is handled in the following way:
+ * - If the exception is [CancellationException], it is ignored, as these exceptions are used to cancel coroutines.
  * - Otherwise, if there is a [Job] in the context, then [Job.cancel] is invoked.
  * - Otherwise, as a last resort, the exception is processed in a platform-specific manner:
  *   - On JVM, all instances of [CoroutineExceptionHandler] found via [ServiceLoader], as well as
  *     the current thread's [Thread.uncaughtExceptionHandler], are invoked.
  *   - On Native, the whole application crashes with the exception.
- *   - On JS, the exception is logged via the Console API.
+ *   - On JS and Wasm JS, the exception is propagated into the JavaScript's runtime event loop
```
dkhalanskyjb:
```suggestion
 *   - On JS and Wasm JS, the exception is propagated into the JavaScript runtime's event loop
```

kotlinx-coroutines-core/jsAndWasmJsShared/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,17 @@
...
+internal expect fun Throwable.toJsException(): JsAny
+
+/*
+ * Schedule an exception to be thrown inside JS or Wasm/JS event loop,
+ * rather than in the current execution branch.
+ */
+internal fun throwAsync(e: JsAny): Unit = js("setTimeout(function () { throw e }, 0)")
+
+internal actual fun propagateExceptionFinalResort(exception: Throwable) {
+    throwAsync(exception.toJsException())
```
bashor, line 16: @murfel @dkhalanskyjb can we throw an exception right from here? May it break anything inside kx-coroutines? 

In my demo project, `throwAsync` was mainly needed to escape kx-coroutines handling and emulate the unhandled exception behavior while using `CoroutineExceptionHandler`.

bashor, line 16: From kotlin 2.2.20-Beta2, in kotlin code, you can just throw a kotlin exception, and everything else will be done by kotlin runtime/compiler, including unwrapping `JsExecption`.
```
throw e
```

But this way, users using old toolchain will get a `WebAssembly.Exception`.

We can try kotlin version at runtime and use a fallback for older versions, but should we?

bashor: cc @murfel @dkhalanskyjb

dkhalanskyjb: No, this method is not allowed to fail. It's called deep inside the coroutine internals, and if it throws anything, things will break.

dkhalanskyjb: We can replace this with just `throw e` when `kotlinx.coroutines` upgrades to Kotlin 2.2.20. Would that work?

PR #4478, opened 2025-07-11
Fix a couple of flaky tests

dkhalanskyjb:


kotlinx-coroutines-debug/test/RunningThreadStackMergeTest.kt
```diff
@@ -9,6 +9,7 @@ import kotlin.test.*
 
 class RunningThreadStackMergeTest : DebugTestBase() {
 
+    private var coroutineThread: Thread? = null
```
qwwdfsad, line 12: Should be volatile as we, at least in one test, publish it from a different thread

qwwdfsad (changes requested):

kotlinx-coroutines-debug/test/RunningThreadStackMergeTest.kt
dkhalanskyjb: I initially had that `@Volatile`, but removed the annotation because of the existing happens-before relationship: `coroutineThread = Thread.currentThread()` ≤ `testMainBlocker.countDown()` ≤ exiting `testMainBlocker.await()`. Is that incorrect?

qwwdfsad: True, I just missed that. If that's the case for all the tests, all good

dkhalanskyjb: Yes, it should be the case, as both the new variable and the `testMainBlocker` are only accessed in two functions bound by this happen-before, the rest of the code just calls those.

PR #4481, opened 2025-07-17
Fix flakiness in `ChannelSelectStressTest`.

chaoren:
It's possible for another sender to close the channel between incrementing `sent` and suspending for `select`. If that happens, the `select` just throws `CancellationException`, cancelling the coroutine and causing an element to be missing at the end of the test.


dkhalanskyjb:
Thanks!

PR #4483, opened 2025-07-18
Fix a flaky check and a TSan violation in `StateFlowStressTest`.

chaoren:
There's no guarantee that any particular collector will ever encounter a value emitted by a particular emitter. It's entirely possible for the value to be overwritten by a different emitter before the collector gets a chance to collect it. It's very unlikely for a collector to miss the second half of all values emitted by a particular emitter, but it is still possible and this causes the test to be flaky.

### DIFF
--- a/kotlinx-coroutines-core/jvm/test/flow/StateFlowStressTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/flow/StateFlowStressTest.kt
@@ -1,9 +1,10 @@
 package kotlinx.coroutines.flow
 
-import kotlinx.coroutines.testing.*
-import kotlinx.coroutines.*
-import org.junit.*
+import java.util.concurrent.atomic.AtomicLongArray
 import kotlin.random.*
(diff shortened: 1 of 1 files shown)


chaoren:
I guess if you want to check that none of the collectors "hanged up" then you could check that `c.max()` is sufficiently high?

dkhalanskyjb:
Yes, good idea. Could you please add that check?

chaoren:
Done. I noticed that the relative values between different emitters doesn't actually matter. They can each emit values at an arbitrary pace, so it doesn't make sense to use `c.max()`. I just changed it to make sure that the collector collected recent enough values from at least one emitter. I also added some changes to fix a TSan violation.

chaoren:
@dkhalanskyjb could you please take a look at the new changes?

dkhalanskyjb (approved):
Sure, I was waiting for a review re-request (top right corner). Thanks!

PR #4484, opened 2025-07-20
Add terrakok/FlowMarbles as stories to the Flow documentation

JSMonk:
The MR is attaching a fantastical interactive explanation (created by @terrakok) of how `Flow` operations work to their documentation.

The result of the MR is presented in the attached video:

https://github.com/user-attachments/assets/f1e1e7ca-6f0b-4e5f-b6b4-b732ed8fc891

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
@@ -26,6 +26,8 @@ fun main() = runBlocking {
 */
 
 /**
+ * @story https://jsmonk.github.io/FlowMarbles/?embedded=true#story/Debounce
+ *
  * Returns a flow that mirrors the original flow, but filters out values
  * that are followed by the newer values within the given [timeout][timeoutMillis].
  * The latest value is always emitted.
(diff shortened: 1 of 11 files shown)


fzhinkin:
@JSMonk, @terrakok nicely done!

There are two aspects here:
1) Correctness of visualizations: it's up to @murfel or @dkhalanskyjb to decide if something has to be changed.
2) Visualizations are currently reside in a separate repository and if we're going to include them into official docs, they should be hosted alongside all other documentation. Are there any specific requirements for that?

kotlinx-coroutines-core/common/src/flow/operators/Delay.kt
```diff
@@ -26,6 +26,8 @@ fun main() = runBlocking {
 */
 
 /**
+ * @story https://jsmonk.github.io/FlowMarbles/?embedded=true#story/Debounce
```
fzhinkin, line 29: Placing a `@story` tag here stops Idea from rendering KDoc past the tag.

terrakok:
I don't mind to copy my repo to the `kotlinx.cotoutines` 👍

PR #4487, opened 2025-07-25
Migrate from BCV to abiValidation

qwwdfsad:


qwwdfsad:
I manually tested that `checkLegacyAbi` triggers failure in KMP and JVM modules if I change ABI.
Now will check the CI (and overwrite with force-push)

fzhinkin:
Note that the validation job is not invoked as a dependency of the `check` task (https://youtrack.jetbrains.com/issue/KT-78525)

qwwdfsad:
Thanks. To catch that, I pushed the last commit 😅

dkhalanskyjb (approved):
Note: the CI failure didn't show this (or I didn't find the error message there), but the `foo` introduced in `reactive/kotlinx-coroutines-reactive/src/ReactiveFlow.kt` also causes the build to fail, which is indeed what we want. Following the instruction in the error message does fix the error.

buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
```diff
@@ -150,3 +160,7 @@ tasks.named("jvmTest", Test::class) {
     }
     project.properties["stressTest"]?.let { systemProperty("stressTest", it) }
 }
+
+tasks.check {
+   dependsOn(tasks.checkLegacyAbi)
+}
```
murfel: newline

qwwdfsad: Sorry, for some reason, EAP IDEA rolled back this particular setting 😞

murfel: I feel like I'm constantly getting newline erasure with pycharm/clion/idea, regardless of EAP...

PR #4489, opened 2025-07-25
Fix flaky shared flow subscription

murfel:
#4488


dkhalanskyjb:
The change makes sense, thanks! However, the failing tests are important. This changes the behavior of other operators non-trivially, and we need to evaluate how big the change is. Could you please classify the tests into those that fail reasonably and those that should ideally still pass? Then, we'll figure out if we need to isolate the change to just `collectLatest` and will have an idea of how big the impact of the change is.

kotlinx-coroutines-core/common/test/flow/operators/BufferTest.kt
```diff
@@ -41,11 +41,134 @@ class BufferTest : TestBase() {
...
+                    expect(2 * (firstBatchSize + (batchNo - 1) * batchSize) + curBatchSize + batchIdx + 2)
+                }
+            }
+        finish(2 * n + 2)
+    }
+
     @Test
     // capacity == -1 to checkBuffer means "no buffer" -- emits / collects are sequentially ordered
     fun testBaseline() =
         checkBuffer(-1) { this }
```
murfel, line 95: Unrelated: Idea gives a warning on `this`: `Flow is constructed but not used`. That's a false positive, right? I'll report it if so.

kotlinx-coroutines-core/common/test/flow/channels/FlowCallbackTest.kt
```diff
@@ -2,9 +2,9 @@
 
 package kotlinx.coroutines.flow
 
-import kotlinx.coroutines.testing.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.testing.*
```
murfel, line 7: Unrelated: L1: Do we need the suppression on Line 1? I don't quite understand how it is relevant here.

murfel:
I aimed to fix all tests but these (conflicting with #4493)
- [ ]      🔴 kotlinx.coroutines.flow.FlatMapLatestTest.testSwitchRendevouzBuffer[jvm]
- [ ]      🔴 kotlinx.coroutines.flow.TransformLatestTest.testSwitchRendezvousBuffer[jvm]
- [ ]      🔴 kotlinx.coroutines.flow.TransformLatestTest.testSwitchBuffer[jvm]

kotlinx-coroutines-core/common/test/flow/operators/ConflateTest.kt
```diff
@@ -20,4 +20,22 @@ class ConflateTest : TestBase() {
...
+//            }
+//        }
+//        val result = flow.flowOn(wrapperDispatcher()).conflate().onEach {
+//            println("Emitting $it")  // prints 1 once and hangs  // todo: ask Dmitry
+//            delay(100)
+//        }.toList()
+//        assertEquals(listOf(1, 10, 20, 30), result)
+//        finish(2)
+//    }
+}
```
dkhalanskyjb: Please revert the changes to this file. We try not to commit commented out code.

dkhalanskyjb (changes requested):

murfel:
@dkhalanskyjb I don't exactly understand why the build failed, could you take a look?

dkhalanskyjb:
The test `kotlinx.coroutines.rx2.ObservableAsFlowTest#testLongRange` hangs.

The chain of events is:
* `kotlinx.coroutines.flow.internal.ChannelFlow#collect` is called on the `channelFlow` of `Observable#asFlow`.
* The dispatcher stays the same, so `CoroutineStart.UNDISPATCHED` is used in the call to `produceImplInternal`.
* The lambda passed to `channelFlow` in `kotlinx.coroutines.rx2.RxConvertKt#asFlow` gets invoked. There, we subscribe to the `Observable`, which is a lazy sequence of more than 64 values.
* `onNext` is repeatedly called by the lazy sequence, populating the channel. At some point, it can't do that anymore, because the channel gets full.
* Because `trySendBlocking` is used to send values into the channel and contains `runBlocking` internally, it parks the current thread...
* ... but that was the only thread we had. We have a deadlock. No coroutines are currently waiting for their turn: the coroutine that was supposed to receive elements from the channel is not suspended, it's in an `UNDISPATCHED` call to the code that blocked without ever suspending, and the code will be block until someone reads from the channel.

murfel:
How do you understand which test hangs from this page?

https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildWindows/5892908

dkhalanskyjb:
Click the red exclamation mark here:

image

Scroll up paste the thread dump at the moment of termination, to the exact moment the build was cancelled, and unfold the `ObservableAsFlowTest` collapsible section:

image

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -109,14 +109,24 @@ public abstract class ChannelFlow<T>(
...
     override suspend fun collect(collector: FlowCollector<T>): Unit =
         coroutineScope {
-            collector.emitAll(produceImpl(this))
+            // If upstream and collect have the same dispatcher, launch the `produce` coroutine undispatched.
+            // This allows the collector to reliably subscribe to the flow before it starts emitting.
+            val current = currentCoroutineContext()[ContinuationInterceptor]
+            val desired = context[ContinuationInterceptor]
+            val start = if (desired == null || desired == current) {
+                CoroutineStart.UNDISPATCHED
+            } else CoroutineStart.ATOMIC
```
qwwdfsad, line 128: Please consider using a consistent code style.

We have some legacy places where one branch is wrapped into braces while another is not, but it's rather confusing than intentional

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -109,14 +109,24 @@ public abstract class ChannelFlow<T>(
...
-        scope.produce(context, produceCapacity, onBufferOverflow, start = CoroutineStart.ATOMIC, block = collectToFun)
+        produceImplInternal(scope, CoroutineStart.ATOMIC)
+
+    internal open fun produceImplInternal(scope: CoroutineScope, start: CoroutineStart): ReceiveChannel<T> =
+        scope.produce(context, produceCapacity, onBufferOverflow, start = start, block = collectToFun)
 
     override suspend fun collect(collector: FlowCollector<T>): Unit =
         coroutineScope {
-            collector.emitAll(produceImpl(this))
+            // If upstream and collect have the same dispatcher, launch the `produce` coroutine undispatched.
```
qwwdfsad, line 122: Previously, `produceImpl` called `markConsumed`, but now it is no longer the case. `ChannelAsFlow.collect` mentions it explicitly.
Probably can be broken by a double collection

qwwdfsad:
```
val channel = Channel<Int>(1).also { it.send(1) }
val flow = channel.consumeAsFlow().buffer()
launch { flow.collect() }
channel.close()
flow.collect()
```

Something like this should fail, but it does not, so I assume this is a regression.
Ditto for collecting non-closing channel, it probably hangs instead of failing fast

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -109,14 +109,24 @@ public abstract class ChannelFlow<T>(
...
      * where the pipeline after the [flowOn] call successfully executes (mostly, its `onCompletion`)
      * handlers, while the pipeline before does not, because it was cancelled during its dispatch.
-     * Thus `onCompletion` and `finally` blocks won't be executed and it may lead to a different kinds of memory leaks.
+     * Thus `onCompletion` and `finally` blocks won't be executed, and it may lead to a different kind of memory leaks.
      */
     public open fun produceImpl(scope: CoroutineScope): ReceiveChannel<T> =
-        scope.produce(context, produceCapacity, onBufferOverflow, start = CoroutineStart.ATOMIC, block = collectToFun)
+        produceImplInternal(scope, CoroutineStart.ATOMIC)
+
+    internal open fun produceImplInternal(scope: CoroutineScope, start: CoroutineStart): ReceiveChannel<T> =
```
qwwdfsad, line 117: Probably shouldn't be open, there is a single implementation of this method

kotlinx-coroutines-core/common/src/flow/internal/ChannelFlow.kt
```diff
@@ -109,14 +109,24 @@ public abstract class ChannelFlow<T>(
...
         coroutineScope {
-            collector.emitAll(produceImpl(this))
+            // If upstream and collect have the same dispatcher, launch the `produce` coroutine undispatched.
+            // This allows the collector to reliably subscribe to the flow before it starts emitting.
+            val current = currentCoroutineContext()[ContinuationInterceptor]
+            val desired = context[ContinuationInterceptor]
+            val start = if (desired == null || desired == current) {
+                CoroutineStart.UNDISPATCHED
+            } else CoroutineStart.ATOMIC
+            collector.emitAll(produceImplInternal(this, start))
```
qwwdfsad, line 129: Another problem:

`produceImpl` was overridden in inheritors, `produceImplInternal` is not, so any behavioural divergence might result in a user-visible regression.

For example, `.merge().conflate()` chain now behaves differently, depending on whether it is `collect`ed or `produceIn`'d because they handle `onBufferOverflow` differently.

qwwdfsad (changes requested):

PR #4511, opened 2025-08-20
propagateExceptionFinalResort-nodejs-reportError

murfel:
#4451

Changelog 
> Needs to be annotated as Breaking changes / Behavior change
> Unhandled exceptions in Kotlin coroutines are now re-thrown into the environment instead of being logged into the console on the `js` and `wasmJS` targets.

I think mentioning that it still only prints to console in browser is not needed, since it, in fact, now throws in browser, only the browser chooses to log it into console.

Also, the logged exception message in browser may change, as it will now be formatted by the browser's preference, rather than by us.


murfel:
@turansky PTAL!

turansky:
LGTM!

murfel:
`reportError` for future reference on different js server runtimes (Bun, Deno)

* The standard https://developer.mozilla.org/en-US/docs/Web/API/Window/reportError
* Kotlin Wrapper https://github.com/JetBrains/kotlin-wrappers/blob/3d7419234a45f7f8a31ea9274f6909ecd63734c6/kotlin-browser/src/webMain/generated/web/errors/reportError.kt
* Deno https://docs.deno.com/api/web/~/reportError
* Bun https://bun.sh/reference/globals/reportError#globals.reportError

kotlinx-coroutines-core/wasmJs/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -14,3 +18,15 @@ internal fun toJsError(message: String?, className: String?, stack: String?): Js
     return error;
     """)
 }
+
+internal actual fun propagateExceptionFinalResort(exception: Throwable) {
+    val jsException = exception.toJsException()
+    // https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-browser/src/webMain/generated/web/errors/reportError.kt
+    if (globalThis.reportError != null) {
+        // Modern browsers, Deno, Bun, etc.
+        globalThis.reportError(jsException)
```
murfel: As I understand, we can only rely on `reportError` once your change in kotlin-wrappers is commited to master? So this PR is dependant on your merge.

https://github.com/JetBrains/kotlin-wrappers/blob/3d7419234a45f7f8a31ea9274f6909ecd63734c6/kotlin-browser/src/webMain/generated/web/errors/reportError.kt

CHANGES.md
```diff
@@ -1,5 +1,8 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.10.3
+* `propagateExceptionFinalResort` now throws instead of logging on the `js` and `wasmJS` targets.
```
murfel, line 4: Needs to be annotated as Breaking changes / Behavior change

CHANGES.md
```diff
@@ -1,5 +1,8 @@
 # Change log for kotlinx.coroutines
 
+## Version 1.10.3
+* Unhandled exceptions in Kotlin coroutines are now re-thrown into the environment instead of being logged into the console on the `js` and `wasmJS` targets.
```
murfel: I think mentioning that it still only prints to console in browser is not needed, since it, in fact, now throws in browser, only the browser chooses to log it into console.

Also, the logged exception message in browser may change, as it will now be formatted by the browser's preference, rather than by us.

murfel: Needs to be annotated as Breaking changes / Behavior change

dkhalanskyjb: We only update changelogs as part of the release procedure: https://github.com/Kotlin/kotlinx.coroutines/blob/master/RELEASE.md This way, several changes can be combined and described in bulk or undo one another.

kotlinx-coroutines-core/wasmJs/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -14,3 +18,15 @@ internal fun toJsError(message: String?, className: String?, stack: String?): Js
     return error;
     """)
 }
+
+internal actual fun propagateExceptionFinalResort(exception: Throwable) {
+    val jsException = exception.toJsException()
+    // https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-browser/src/webMain/generated/web/errors/reportError.kt
```
dkhalanskyjb: Using commit hashes instead of branch names can make such links more reliable. I wanted to say "this link could start returning 404 tomorrow", but in fact, it returns 404 for me today!

PR #4513, opened 2025-08-25
Update Readme to say Kotlin 2.1.0 to match the reality

murfel:
It's been Kotlin 2.1.0 for three patch releases already, but the readme wrongly says 2.0.0.


murfel:
> This is a companion version for the Kotlin `2.1.0` release.

Worth clarifying what a companion version is, something like: It was tested with Kotlin 2.1.0, but it is also expeted to work with the current stable Kotlin version.

PR #4514, opened 2025-08-25
Explicit types arguments instead of reified intersection

bnorm:
Reified intersection type arguments are currently a warning (KT-52469) but will soon become an error (KT-71420).

This change is to unblock the user project build for the merge for KT-71420


murfel:
Thank you!

PR #4517, opened 2025-09-02
Add tests for Channel closed with an exception

murfel:


kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+                }.onFailure {
+                    // empty, but not yet closed
+                    break
+                }.getOrThrow().apply(lst::add)
+            }
+            assertTrue(lst.isEmpty())
+        }
+    }
+
+    /**
```
dkhalanskyjb: This syntax means the comment is a KDoc and refers to the entity directly below it. To make this a normal multiline comment for separating groups of tests, use `/*`.

That said, I think the separation is easy to break accidentally. Why not introduce separate classes to make this more robust?

```kotlin
class ClosedChannelTest: TestBase() {
    /**
     * Properties.
     *
     * Properties stay the same regardless of whether the channel was closed with or without exception.
     */
    class PropertiesTests: TestBase() {
        
    }
}
```

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+import kotlin.test.*
+
+class ClosedChannelTest : TestBase() {
+    /**
+     * Iterator.
+     */
+
+    @Test
+    fun testIteratorClosedHasNextFalse() = runTest {
+        TestChannelKind.entries.forEach { kind ->
```
dkhalanskyjb: Proposal:
```kotlin
    fun TestChannelKind.Companion.runTestForEach(
        testBody: suspend CoroutineScope.(TestChannelKind) -> Unit
    ): TestResult = runTest {
        TestChannelKind.entries.forEach { kind ->
            testBody(kind)
        }
    }
```
May be more convenient, with how often this pattern occurs.

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+            val channel = kind.create<Int>()
+            val iterator = channel.iterator()
+            channel.close(TestException())
+            assertFailsWith<TestException> { (iterator.hasNext()) }
+        }
+    }
+
+    @Test
+    fun testClosedToListOk() = runTest {
```
dkhalanskyjb:
```suggestion
    fun testClosedToListDoesNotThrow() = runTest {
```

dkhalanskyjb: Also, this test is not part of the `Iterator` section.

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    @Test
+    fun testClosedWithExceptionToListThrows() = runTest {
+        TestChannelKind.entries.forEach { kind ->
+            val channel = kind.create<Int>()
+            channel.close(TestException())
+            assertFailsWith<TestException> { channel.toList() }
+        }
+    }
```
dkhalanskyjb: Duplicates `ChannelsTest#testToListOnFailedChannel`.

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+
+    @Test
+    fun testClosedConsumeEachOk() = runTest {
```
dkhalanskyjb: Duplicates `ConsumeTest#testConsumeEachClosesOnSuccess`, and I'd argue that place is a better fit for this test, since `close` is the more fundamental operation.

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+        TestChannelKind.entries.forEach { kind ->
+            val channel = kind.create<Int>()
+            channel.close()
+            channel.consumeEach {  }
+        }
+    }
+
+    @Test
+    fun testClosedWithExceptionConsumeEachThrows() = runTest {
```
dkhalanskyjb: Duplicates `ConsumeTest#testConsumeEachThrowingOnChannelClosing`.

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+     * toList() alternative when a channel might close with an exception.
+     */
+
+    @Test
+    fun testToListCatching() = runTest {
+        TestChannelKind.entries.forEach { kind ->
+            val channel = kind.create<Unit>()
+            channel.close(TestException())
+            val lst = mutableListOf<Unit>()
+            while (true) {
```
dkhalanskyjb: Why is this loop necessary? Only one iteration of it will run. `lst:add` will never run, as won't `getOrThrow()`. The whole test can be simplified to
```kotlin
            channel.close(TestException())
            val result = channel.receiveCatching()
            result.onClosed { // suspends
                assertTrue(it is TestException)
            }
            assertTrue(result.isClosed)
```
This is covered at least by `BasicOperationsTest#testReceiveCatching`

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+                    assertTrue(it is TestException)
+                    break
+                }.getOrThrow().apply(lst::add)
+            }
+            assertTrue(lst.isEmpty())
+        }
+    }
+
+    @Test
+    fun testTryToListCatching() = runTest {
```
dkhalanskyjb: Ditto.

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+            assertTrue(channel.isClosedForReceive)
+        }
+    }
+
+    @Test
+    fun testIsEmpty() = runTest {
+        val channel = Channel<Int>()
+        assertTrue(channel.isEmpty)
+        assertFalse(channel.isClosedForReceive)
+    }
```
dkhalanskyjb: This is unrelated to closing channels.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/test/channels/ClosedChannelTest.kt
murfel: It looks ugly, I'd rather not. See the last commit. Шило на мыло. Revert?

dkhalanskyjb: The approach with `ChannelTestBase` does indeed look too complicated and indirect for such small gains, so yes, I agree it's not a win. That wasn't what I had in mind, though. My thought was to introduce only the `fun TestChannelKind.Companion.runTestForEach` extension in `kotlinx.coroutines.channels` in the test source set (or `fun TestBase.runTestForEachKind` if you prefer).

In any case, this is just a suggestion, please ignore it if you don't find it valuable.

murfel: It's not that now I need to inherit from `ChannelTestBase`. It's actually the way it looks in code, and it will look identical with your approach as well. Reverted.

kotlinx-coroutines-core/common/test/channels/ChannelIteratorTest.kt
```diff
@@ -0,0 +1,26 @@
...
+        }
+    }
+
+    @Test
+    fun testIteratorClosedWithExceptionHasNextThrows() = runTest {
+        TestChannelKind.entries.forEach { kind ->
+            val channel = kind.create<Int>()
+            val iterator = channel.iterator()
+            channel.close(TestException())
+            assertFailsWith<TestException> { (iterator.hasNext()) }
```
dkhalanskyjb:
```suggestion
            assertFailsWith<TestException> { iterator.hasNext() }
```

kotlinx-coroutines-core/common/test/channels/ChannelPropertiesTest.kt
```diff
@@ -0,0 +1,86 @@
+package kotlinx.coroutines.channels
+
+import kotlinx.coroutines.testing.*
+import kotlin.test.*
+
+/**
+ * Properties stay the same regardless of whether the channel was closed with or without exception.
+ */
+class ChannelPropertiesTest : TestBase() {
```
dkhalanskyjb:
```suggestion
class ChannelPropertiesAfterClosingTest : TestBase() {
```
This seems to more accurately reflect the focus.

kotlinx-coroutines-core/common/test/channels/ChannelPropertiesTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        channel.send(1)
+        assertFalse(channel.isEmpty)
+        assertFalse(channel.isClosedForReceive)
+        channel.close(TestException())
+        assertFalse(channel.isEmpty)
+        assertFalse(channel.isClosedForReceive)
+        channel.receive()
+        assertFalse(channel.isEmpty)
+        assertTrue(channel.isClosedForReceive)
+    }
```
dkhalanskyjb: The tests here come in perfect pairs of channels being closed with/without an exception. This calls for either extracting the test logic into separate parameterized functions or checking both cases in each test (using something like `for (exception in listOf(null, TestException())`).

PR #4520, opened 2025-09-10
Add ReceiveChannel.toList(destination)

dkhalanskyjb:
Use case: collecting elements up until the point the channel is closed without losing the elements when `toList` when the exception is thrown.

This function is similar to `Flow<T>.toList(destination)`, which we already have, so the addition makes sense from the point of view of consistency as well.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channels.common.kt
@@ -194,6 +194,43 @@ public suspend fun <E> ReceiveChannel<E>.toList(): List<E> = buildList {
     }
 }
 
+/**
+ * Consumes the elements of this channel into the given [destination] mutable list.
+ *
+ * This function will attempt to receive elements and put them into the list until the channel is
+ * [closed][SendChannel.close].
+ * Calling [toList] on channels that are not eventually closed is always incorrect:
+ * - It will suspend indefinitely if the channel is not closed, but no new elements arrive.
+ * - If new elements do arrive and the channel is not eventually closed, [toList] will use more and more memory
+ *   until exhausting it.
+ *
+ * If the channel is [closed][SendChannel.close] with a cause, [toList] will rethrow that cause.
+ * However, the [destination] list is left in a consistent state containing all the elements received from the channel
+ * up to that point.
+ *
+ * The operation is _terminal_.
+ * This function [consumes][ReceiveChannel.consume] all elements of the original [ReceiveChannel].
+ *
+ * Example:
+ * ```
+ * val values = listOf(1, 5, 2, 9, 3, 3, 1)
+ * // start a new coroutine that creates a channel,
+ * // sends elements to it, and closes it
+ * // once the coroutine's body finishes
+ * val channel = produce {
+ *    values.forEach { send(it) }
+ * }
+ * val destination = mutableListOf<Int>()
+ * channel.toList(destination)
+ * check(destination == values)
+ * ```
+ */
+public suspend inline fun <T> ReceiveChannel<T>.toList(destination: MutableList<T>) {
+    consumeEach {
+        destination.add(it)
+    }
+}
+
 @PublishedApi
 internal fun ReceiveChannel<*>.cancelConsumed(cause: Throwable?) {
     cancel(cause?.let {
--- a/kotlinx-coroutines-core/common/test/channels/ChannelsTest.kt
+++ b/kotlinx-coroutines-core/common/test/channels/ChannelsTest.kt
@@ -85,14 +85,13 @@ class ChannelsTest: TestBase() {
     }
 
     @Test
-    fun testEmptyList() = runTest {
+    fun testEmptyToList() = runTest {
         assertTrue(emptyList<Nothing>().asReceiveChannel().toList().isEmpty())
     }
 
     @Test
     fun testToList() = runTest {
         assertEquals(testList, testList.asReceiveChannel().toList())
-
     }
 
     @Test
@@ -102,6 +101,39 @@ class ChannelsTest: TestBase() {
         assertFailsWith<TestException> {
             channel.toList()
         }
+    }
+
+    @Test
+    fun testEmptyToListWithDestination() = runTest {
+        val initialList = listOf(-1, -2, -3)
+        val destination = initialList.toMutableList()
+        emptyList<Nothing>().asReceiveChannel().toList(destination)
+        assertEquals(initialList, destination)
+    }
+
+    @Test
+    fun testToListWithDestination() = runTest {
+        val initialList = listOf(-1, -2, -3)
+        val destination = initialList.toMutableList()
+        testList.asReceiveChannel().toList(destination)
+        assertEquals(initialList + testList, destination)
+    }
+
+    @Test
+    fun testToListWithDestinationOnFailedChannel() = runTest {
+        val initialList = listOf(-1, -2, -3)
+        val destination = initialList.toMutableList()
+        val channel = Channel<Int>(10)
+        val elementsToSend = (1..5).toList()
+        elementsToSend.forEach {
+            val result = channel.trySend(it)
+            assertTrue(result.isSuccess)
+        }
+        channel.close(TestException())
+        assertFailsWith<TestException> {
+            channel.toList(destination)
+        }
+        assertEquals(initialList + elementsToSend, destination)
     }
 
     private fun <E> Iterable<E>.asReceiveChannel(context: CoroutineContext = Dispatchers.Unconfined): ReceiveChannel<E> =


kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -194,6 +194,43 @@ public suspend fun <E> ReceiveChannel<E>.toList(): List<E> = buildList {
...
+ *
+ * This function will attempt to receive elements and put them into the list until the channel is
+ * [closed][SendChannel.close].
+ * Calling [toList] on channels that are not eventually closed is always incorrect:
+ * - It will suspend indefinitely if the channel is not closed, but no new elements arrive.
+ * - If new elements do arrive and the channel is not eventually closed, [toList] will use more and more memory
+ *   until exhausting it.
+ *
+ * If the channel is [closed][SendChannel.close] with a cause, [toList] will rethrow that cause.
+ * However, the [destination] list is left in a consistent state containing all the elements received from the channel
```
murfel: Since there's little difference between toList(List) and toList(), shall we reference one in the other instead of copy-pasting?

> **Similar to toList(), however, this overload** consumes the elements of this channel into the given [destination] mutable list.
> If the channel is [closed][SendChannel.close] with a cause, [toList] will rethrow that cause, **same as toList**.
> However, **in contrast to toList**, the [destination] list is left in a consistent state containing all the elements received from the channel

And keep both usage examples, of course.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -194,6 +194,43 @@ public suspend fun <E> ReceiveChannel<E>.toList(): List<E> = buildList {
     }
 }
 
+/**
```
murfel, line 202: Re the original toList() docmuentation - shall it say that toList must be used by a single consumer (the implication of consumeEach and the iterator used there)?

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: We don't have a consistent policy regarding this, it's typically a subjective decision. The big upside of the approach you propose is the reduced maintenance burden (if we change the docs, we only need to do it once), but the big downside is that there's no straightforward way of navigating from the documentation for one overload to the one for the other. Hover over a `toList` usage in the IDE and try navigating to the documentation for the other overload, and you'll see what I mean.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -183,16 +186,13 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
  * }
  * check(channel.toList() == values)
  * ```
  */
-public suspend fun <E> ReceiveChannel<E>.toList(): List<E> = buildList {
-    consumeEach {
-        add(it)
-    }
-}
+public suspend inline fun <T> ReceiveChannel<T>.toList(destination: MutableList<T> = ArrayList()): List<T> =
```
murfel: Oh, we can't check that the return value will be used now. Maybe unsplit it back, and do not return List on one of the destination overload?

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: It already says so:
> * The operation is _terminal_.
> * This function [consumes][ReceiveChannel.consume] all elements of the original [ReceiveChannel].

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -201,3 +201,5 @@ internal fun ReceiveChannel<*>.cancelConsumed(cause: Throwable?) {
     })
 }
 
+@Deprecated("Preserving binary compatibility, was stable", level = DeprecationLevel.HIDDEN)
+public suspend fun <T> ReceiveChannel<T>.toList(): List<T> = toList(ArrayList())
```
murfel: Newline police

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: I've attempted this, but cross-referencing overloads clutters the documentation quite a bit, and also, `Flow.toList` already follows the single-overload approach: `public suspend fun <T> Flow<T>.toList(destination: MutableList<T> = ArrayList()): List<T>`, so it's more internally consistent to have a single overload.

I'm wondering how real the risk of calling `toList()` on a channel and forgetting to obtain the resulting list is. Do you have an opinion on this?

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: Yes, but the UX is bad and in its current form people will use it incorrectly with a higher probability.

_Terminal_ doesn't redirect anywhere.

Oh wait, you can technically call it from different coroutines, it just doesn't make much sense. Nevermind then.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: We could in theory check it on `flow` if people do this mistake...

However, having the return value also allows chaining, and that tips the scale for me.

So let's keep your current version.

murfel: I wonder if people mistakenly do `flow.toList()` instead of `flow.collect()` to trigger flow execution for its side effects.

dkhalanskyjb: @sandwwraith, what do you think, is it more idiomatic to have `fun X.toList(): List<T>` + `fun X.toList(destination: MutableList<T>): Unit` or just `fun X.toList(destination: MutableList<T> = ArrayList<T>): List<T>` from the point of view of having an excessive return value?

sandwwraith: I think it is better to have it as in the stdlib now: `fun X.toList(): List<T>` + `@Ignorable fun <C: Collection/MutableList> X.toCollection/MutableList(destination: C): C`. Generic argument in the second function allows you to collect result to `ArrayList` or `HashSet`, for example.

murfel:
Could you also fix the `Channel.toList` docs to avoid saying "all elements" (two entries on this page)
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/to-list.html

dkhalanskyjb:
@murfel, that page is automatically generated from the documentation comments that are edited in this PR.

murfel:
I'm aware.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -162,9 +165,16 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
     }
 
 /**
- * Returns a [List] containing all the elements sent to this channel, preserving their order.
+ * [Consumes][consume] the elements of this channel into a list, preserving their order.
  *
- * This function will attempt to receive elements and put them into the list until the channel is
+ * This is a convenience function equivalent to calling [consumeAsFlow] followed by [kotlinx.coroutines.flow.toList].
+ * It is useful for testing code that uses channels to observe the elements the channel contains at the end of the test.
```
murfel, line 164:
```suggestion
 * It is useful for testing to observe the elements the channel contains at the end of the test.
```

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -162,9 +165,16 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ *
+ * There is no way to recover channel elements if the channel gets closed with an exception
+ * or to apply additional transformations to the elements before building the resulting collection.
+ * Please use [consumeAsFlow] and [kotlinx.coroutines.flow.toCollection] for such advanced use-cases.
```
murfel, line 168:
```suggestion
 * Please use either [consumeTo] or [consumeAsFlow] and [kotlinx.coroutines.flow.toCollection] for such advanced use-cases.
```

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
  * ```
  */
 public suspend fun <E> ReceiveChannel<E>.toList(): List<E> = buildList {
```
murfel, line 198: We forgot about lost elements (`onUndeliveredElement`). It should be at least mentioned in the doc, and possibly `onUndeliveredElement` could be added as a parameter.

`onUndeliveredElement` overcomplicates the signature for no real use-case, but provides completeness. No strong opinion on my end.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -173,8 +183,9 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
  *
  * If the channel is [closed][SendChannel.close] with a cause, [toList] will rethrow that cause.
  *
- * The operation is _terminal_.
- * This function [consumes][ReceiveChannel.consume] all elements of the original [ReceiveChannel].
+ * Since this function is implemented using [consume], the function is _terminal_.
+ * In practice, this means that if adding new elements to the list fails with an exception,
+ * that exception will be used for [cancelling][ReceiveChannel.cancel] the channel and rethrown.
```
murfel: I agree with both statements (L186 and L187-188), but I don't see how one implies the others.

Terminal operation only means the channel will be cancelled after the operation, as defined in lines 54/47.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ *
+ * The intended use case for this function is collecting the remaining elements of a closed channel
+ * and processing them in a single batch.
+ *
+ * Example:
+ * ```
+ * val doContinue = AtomicBoolean(true)
+ *
+ * // Start the sender
+ * val channel = produce {
```
murfel, line 231: Re using produce - could you confirm if our undestanding matches?

In theory we may want to avoid using `@ExperimentalCoroutinesApi` in samples, but the channels are the same level of lower-level abstractions, so if someone uses the channels, it's ok to make them aware of produce.

And also because our `@ExperimentalCoroutinesApi` annotations do not always precisely indicate the state of API.

Same for `AtomicBoolean`, but I can see how it's not as easy to substitute for a stable API.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel, line 93: I suggest we also note explicitly that, despite the semantics and examples implying that a single coroutine calls this function (and other functions based on this one), multiple coroutines *can* invoke this function on the same channel.

And in some cases it can or cannot make sense.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ *
+ * This is a convenience function equivalent to calling [consumeAsFlow]
+ * followed by [kotlinx.coroutines.flow.toCollection].
+ * Please use [consumeAsFlow] directly in scenarios where elements should undergo additional transformations
+ * before being added to the resulting collection.
+ *
+ * [consumeTo] attempts to receive elements and put them into the collection until the channel is
+ * [closed][SendChannel.close].
+ *
+ * If the channel is [closed][SendChannel.close] with a cause, [consumeTo] will rethrow that cause.
```
murfel, line 213:
```suggestion
 * If the channel is [closed][SendChannel.close] with a cause, [consumeTo] will rethrow that cause immediately.
 * The channel could still contain elements that will never be received.
```

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ * However, the elements already received up to that point will remain in the collection.
+ *
+ * Since this function is implemented using [consume], the function is _terminal_.
+ * In practice, this means that if adding new elements to the collection fails with an exception,
+ * that exception will be used for [cancelling][ReceiveChannel.cancel] the channel and rethrown.
+ *
+ * The intended use case for this function is collecting the remaining elements of a closed channel
```
murfel, line 223: Not necessarily "the remaining elements of a **closed** channel".

It is definitely for use when we know that the channel **is or will be closed** at some point.

murfel: Your sample is of this use-case, but there are others.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ *         println("Received ${nextElement.getOrNull()}")
+ *     }
+ *     doContinue.store(false)
+ *     delay(100.milliseconds)
+ *     val remainingElements = mutableListOf<Int>()
+ *     try {
+ *         channel.consumeTo(remainingElements)
+ *     } finally {
+ *         println("Remaining elements: $remainingElements")
+ *     }
```
murfel, line 257: A more closer example would be to have a channel which is can be **closed with a cause**, to separate the usage of this function from just `channel.toList()`

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ *         delay(10.milliseconds)
+ *         if (i == 42) break
+ *     }
+ * }
+ *
+ * // Start the consumer
+ * launch {
+ *     // Read elements until we suddenly decide to stop
+ *     // or until the channel is closed.
+ *     while (Random.nextInt() % 100 != 42) {
```
murfel: Simplify and avoid false connotation with 42 above: `Random.nextInt(100) != 0`

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: Also would be nice to show an example when `consumeTo` is used before the channel is definitely closed.

(This example could approach this by just removing delay on L251)

murfel: I'm a bit wary that you are advocating exclusively for one particular use-case, here in the sample and in the kdoc.

murfel: Let's discuss the intention of this API and samples on Monday!

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -189,11 +200,68 @@ public suspend inline fun <E> ReceiveChannel<E>.consumeEach(action: (E) -> Unit)
...
+ *         val nextElement = channel.receiveCatching()
+ *         if (nextElement.isClosed) return@launch
+ *         println("Received ${nextElement.getOrNull()}")
+ *     }
+ *     doContinue.store(false)
```
murfel, line 250:
```suggestion
  *     delay(100.milliseconds)
  *     doContinue.store(false)
```

murfel: Let the producer 'accidentally' produce more elements, so that the `remainingElements` are definitely not empty (it was always empty when I ran this sample on my mac)

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: We do want to direct the users to `Flow` here.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: This is a good catch. Actually, the text in its current form is simply wrong.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: We have no specific policy, only opinions.

I think it's okay to use `@ExperimentalCoroutinesApi` in samples. If we have a piece of API, then we probably believe it's a good way to solve some problem, even if it's `@Experimental`. When a sample deals with the problem solved by an API, we should stand by our decision to provide the API at all, and also, spread the knowledge about this API existing so that we can gather more feedback.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: Not true: https://pl.kotl.in/U4TKRZ-I2

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: What you're describing is not the intended use case. If you only know the channel will be closed at some point, then `consumeTo` uses an unbound amount of memory, which we try to avoid. A possible use case is sending a bounded number of elements (for example, we know there will be at most 100 elements before the channel gets closed). I'd argue that channels are not the right tool for the job in that case, you should instead build a list on the sender's side and publish it as a whole to the receiver.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: Could you explain what behavior of `onUndeliveredElement` is missing from the description? We do say that the elements that did get delivered may still be lost if the channel is closed with a cause.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
```diff
@@ -242,11 +248,12 @@ public suspend fun <E> ReceiveChannel<E>.toList(): List<E> = buildList {
  * launch {
  *     // Read elements until we suddenly decide to stop
  *     // or until the channel is closed.
- *     while (Random.nextInt() % 100 != 42) {
+ *     while (Random.nextInt() % 100 != 0) {
```
murfel: It's better to teach users to use even adjacent things properly. `Random.nextInt(100)` requires less random bits per invokation and is also genuinely uniformly distributed.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: Ah, found it.

It should be placed (or duplicated) at the end, near the description of its terminal behaviour, since it's logically connected.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: Nevermind, I short circuited (and also misread the docs). Please do add when exactly the exception will be rethrown, though.

> The channel could still contain elements that will never be received.

Elaborate this whole scenario.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: If you have a fairly closed scope for channels in mind, could you elaborate such use-cases in the kdoc? Currently it leaves me puzzled as to why this is the only case presented, even though I can see that I can use it in other ways.

murfel (changes requested):

murfel:
Each `consume`-like function has a very similar description. Could you unify these descriptions? Since they are saying the same thing with different words. Or worse, give slightly different information.

E.g. `consume` and `consumeEach` both end with

`consume`
```
 * [consume] does not guarantee that new elements will not enter the channel after [block] finishes executing, so
 * some channel elements may be lost.
 * Use the `onUndeliveredElement` parameter of a manually created [Channel] to define what should happen with these
 * elements during [ReceiveChannel.cancel].
```

`consumeEach`
```
 * **Pitfall**: even though the name says "each", some elements could be left unprocessed if they are added after
 * this function decided to close the channel.
 * In this case, the elements will simply be lost.
 * If the elements of the channel are resources that must be closed (like file handles, sockets, etc.),
 * an `onUndeliveredElement` must be passed to the [Channel] on construction.
 * It will be called for each element left in the channel at the point of cancellation.
```

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: Yes, it's a good idea to mention `Flow` (which take over most of the typical use cases for `Channel`) in the Channel documentation: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel/

Certainly out of scope of this PR, though.

murfel: Sure, but I still can't verify if that's the only reasonable use-case worth mentioning.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: > Please do add when exactly the exception will be rethrown, though.

What happens there is inherent to the behavior of closing the channel: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-send-channel/close.html

> Elaborate this whole scenario.

The text you quote is incorrect, closing the channel still allows accessing all elements there.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: I don't see how that's connected. *Terminal* only means that the operation cancels the channel, which does get mentioned.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: Yes, I also can't verify that. Somewhere, there may and probably do exist reasonable usages I haven't thought of. This section of the text is descriptive ("here's what this function was designed to do"), not prescriptive ("here's the singular scenario in which you're allowed to call this function"). The purpose of this section is to help people that are unsure if this is the function they want: if this description matches their scenario, they will know that yes, they have found the right one. If not, they may need to dive in deeper and/or keep searching.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: I don't see the point. I haven't heard of this being a pitfall, so it's not something we need to warn people about. This information can also be deduced from the first two paragraphs, so someone trying to debug an accidental usage of `consume` from multiple coroutines already has the information they need.

dkhalanskyjb:
Good job noticing the discrepancy, fixed it.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
dkhalanskyjb: I'd leave the "code that uses channels" phrasing so that people whose code doesn't use channels can easily understand that this is not the function they're looking for.

kotlinx-coroutines-core/common/src/channels/Channels.common.kt
murfel: It's not about warning. It's about confirming that it is indeed possible to use it from multiple coroutines.

murfel: `and before the channel is cancelled`

Issue #4469, opened 2025-07-02
Simplify attach of the coroutines debug agent in the debugger


mvicsokolova:
**Problem**
To debug coroutines the debugger should pass `kotlinx-coroutines-core.jar` found in the project’s classpath as a `-javaagent` vm parameter to the given run configuration.
The debugger does the following:
1. Checks if the given run configuration has coroutines dependency in it’s classpath
2. Retrieves all the `kotlinx-coroutines-core` jars available
3. Chooses the correct version to become the debug agent (the latest version available)

Various bugs appear in this logic from time to time (like incorrect jar version was chosen or run configuration api has changed), which cause attach of the coroutines debug agent to fail or crash the debug session: [IDEA-368446](https://youtrack.jetbrains.com/issue/IDEA-368446/Coroutine-agent-attach-is-broken-after-the-change-of-the-project-run-configuration-type), [IDEA-369170](https://youtrack.jetbrains.com/issue/IDEA-368910/Cant-run-debugging-in-a-simple-project-with-coroutines), [IDEA-278577](https://youtrack.jetbrains.com/issue/IDEA-373038/Cant-start-an-application-in-debug-mode-because-of-java.lang.NoClassDefFoundError-kotlin-Result-in-javaagen)

**Would be great**
if the correct coroutines jar was already added as a `javaagent` by the time the debugger is attached to the run configuration. And by setting some flag in the debugger we would enable actual insertion of debug probes.

But It may be not ok to always add a vm parameter to the run configuration (not in the debug mode).

Could we discuss whether it’s possible and makes sense to remove this attach logic from the debugger? And how this could be implemented and whether any similar agent examples already exist? 👀

dkhalanskyjb:
From what I've seen, it doesn't seem viable.
The Java Agent is the way to obtain the ability to rewrite bytecode of classes without modifying the classloader, and someone must install the Java Agent. For that, *someone* has to register something as the Java Agent. I didn't manage to find a way of doing this without knowing the path to the jar file.

That said, won't this work to find the required jar file reliably?
```kotlin
val jarPath: String? = CoroutineScope::class.java
   .getProtectionDomain()
   .getCodeSource()
   .getLocation()
   .toURI()
   .getPath()
```
You do already obtain some library APIs, right? If so, is it possible to see which jar these APIs come from using this trick?

gorrus:
[...]

PR #4524, opened 2025-09-16
Attach coroutine debug agent safely

mvicsokolova:
Recently there was a bug in the debugger [IDEA-377586](https://youtrack.jetbrains.com/issue/IDEA-377586/Cannot-Debug-Tomcat-in-IntelliJ-IDEA-2025.2?backToIssues=false): the debug session failed at the start because initialization of the coroutine debug agent failed with this error:
```
Exception in thread "main" java.lang.NoClassDefFoundError: kotlin/Result
	at kotlinx.coroutines.debug.internal.AgentPremain.<clinit>(AgentPremain.kt:19)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1160)
	at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.ensureClassInitialized(MethodHandleAccessorFactory.java:300)
	at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.newMethodAccessor(MethodHandleAccessorFactory.java:71)
	at java.base/jdk.internal.reflect.ReflectionFactory.newMethodAccessor(ReflectionFactory.java:159)
	at java.base/java.lang.reflect.Method.acquireMethodAccessor(Method.java:726)
	at java.base/java.lang.reflect.Method.invoke(Method.java:577)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:560)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:572)
```

This happened because debugger applied `kotlinx-coroutines-core` as a `-javaagent` to the run configuration, which did not have `kotlin-stdlib` in it's bootstrap classpath and the Kotlin classes would've only be loaded later.

Ad a WA the debugger just skips run configurations like this and does not try to attach the coroutine debug agent at all.
This issue relates to the main problem of the coroutines agent attach (https://github.com/Kotlin/kotlinx.coroutines/issues/4469), and SPI solution would help in this case too.

For now I suggest to provide some "safe" wrapper around `AgentPremain`, which would do nothing in case `AgentPremain` initialization fails.
It would be bad to fail completely silently, as the user would want to know, what could be done to attach the agent anyway, e.g. add `kotlin-stdlib.jar` to the bootstrap classpath or smth. So maybe we can choose silent failure / clear error message by setting some system property.

Anyway, I think it makes sense to provide some readable clear error message if something went wrong and give a hint how to attach the agent, because the debugger does not help in these cases.

If there may be a better solution, let's discuss please)


kotlinx-coroutines-core/jvm/src/debug/internal/AgentPremain.kt
```diff
@@ -69,3 +69,28 @@ internal object AgentPremain {
...
+ * Initialization of [AgentPremain] may fail, e.g., because it started before the required kotlin-stdlib classes were loaded.
+ * If `kotlinx.coroutines.apply.debug.agent` is set to `false`, the coroutine debug agent will not be applied.
+ */
+@Suppress("unused")
+internal object SafeAgentPremain {
+    @JvmStatic
+    fun premain(args: String?, instrumentation: Instrumentation) {
+        // If this property is disabled, the debug agent will not be applied at all
+        val applyDebugAgent = System.getProperty("kotlinx.coroutines.apply.debug.agent", "true").toBoolean()
+        if (!applyDebugAgent) return
```
dkhalanskyjb: If someone controls the system environment at the point when `premain` is being run, it means they can edit the command line which starts the process, and then, they can simply avoid specifying the Java agent. What's the purpose of adding a new property, then?

kotlinx-coroutines-core/jvm/src/debug/internal/AgentPremain.kt
```diff
@@ -69,3 +69,28 @@ internal object AgentPremain {
...
+
+/**
+ * This class is loaded if and only if kotlinx-coroutines-core was used as a javaagent argument.
+ * It serves as a "safe" wrapper around [AgentPremain] that does not require any kotlin-stdlib classes to be loaded for its initialization.
+ *
+ * Initialization of [AgentPremain] may fail, e.g., because it started before the required kotlin-stdlib classes were loaded.
+ * If `kotlinx.coroutines.apply.debug.agent` is set to `false`, the coroutine debug agent will not be applied.
+ */
+@Suppress("unused")
+internal object SafeAgentPremain {
```
dkhalanskyjb: I don't understand the purpose of this wrapper. What's stopping us from adding the check to `AgentPremain` itself? Won't it be enough to call something like this function as the first thing in `AgentPremain.premain`?
```kotlin
private fun checkIfStdlibIsAvailable() {
  try {
    Result.success(42)
  } catch (_: Throwable) {
    error("kotlinx.coroutines debug agent failed to load " +
        "because the Kotlin standard library is not present in the classpath.")
  }
}
```

kotlinx-coroutines-core/jvm/src/debug/internal/AgentPremain.kt
mvicsokolova: True, we don't need a property, we can just add the hint to remove the `-javaagent` argument in the error message

kotlinx-coroutines-core/jvm/src/debug/internal/AgentPremain.kt
mvicsokolova: This makes sense, yes) This is the first point when we find out that stdlib is not loaded

gorrus: If someone later adds a field of some stdlib type to the AgentPremain class, it will cause problems at class loading time. The wrapper will protect against that.

mvicsokolova: No, actually this does not help, because the error happens when the `AgentPremain` class is trying to initialize, at the moment `AgentPremain.<clinit>` is called. And this check in `premain` is already too late.

dkhalanskyjb: To protect against that, we could implement the check in `AgentPremain` like so:
```kotlin
// This should be the first property, to ensure the check happens first! Add new properties only below!
val dummy = checkIfStdlibIsAvailable()
```

kotlinx-coroutines-core/jvm/src/debug/internal/AgentPremain.kt
mvicsokolova: Removed this extra system property

kotlinx-coroutines-core/jvm/src/debug/internal/AgentPremain.kt
mvicsokolova: And replaced the wrapper class with this dummy property check

gorrus:
please also add a test to ensure that the agent does no harm if stdlib is not available

integration-testing/build.gradle.kts
```diff
@@ -220,6 +220,21 @@ tasks {
         classpath = sourceSet.runtimeClasspath
     }
 
+    // todo assert that this task is failing
+    // In this test coroutine debug agent is attached as a javaagent vm argument
+    // to a pure Java project (safeDebugAgentTest) with no Kotlin stdlib dependency.
+    // In this case the error should be thrown from AgetnPremain class:
+    // "java.lang.IllegalStateException: kotlinx.coroutines debug agent failed to load."
+    create<Test>("safeDebugAgentTest") {
```
mvicsokolova: @dkhalanskyjb hi!)
I need some help with Gradle here 👀 

How could we check, that `:safeDebugAgentTest:run` fails with an error from `AgentPremain?`
Is there smth like `assertFailsWith ` in Gradle?

I've thought of an alternative to run `:safeDebugAgentTest:run` from a Kotlin test: to create a `ProcessBuilder` + try / catch, but maybe there is an easier option?

dkhalanskyjb: I'm no Gradle expert, either. I've seen this approach to checking whether and how a Gradle task fails:
* https://github.com/Kotlin/kotlinx-benchmark/blob/bb2208c76f819056e1fa0e665dee503249d4650e/integration/src/test/kotlin/kotlinx/benchmark/integration/BenchmarkFunctionNameValidationTest.kt the test itself,
* https://github.com/Kotlin/kotlinx-benchmark/blob/bb2208c76f819056e1fa0e665dee503249d4650e/integration/src/test/resources/templates/funny-names the project whose build should fail,
* https://github.com/Kotlin/kotlinx-benchmark/blob/bb2208c76f819056e1fa0e665dee503249d4650e/integration/build.gradle.kts#L15 the required dependency.

I think this should point you in the right direction, but if you encounter any difficulties, please let me know, and I'll look into this more closely.

murfel: also possibly @shanshin could help?

shanshin: Would you like to check the cause of the failure in functional tests or directly in the build?

mvicsokolova: I want to ensure that the `run` task of the project fails (`:safeDebugAgentTest:run`), I guess those examples from the benchmarks should help, I'll try

Thank you)

mvicsokolova: So, I decided to avoid gradle test runners, that would be too complicated. Just forwarded errorOutput of the gradle task and added an error message check.

integration-testing/safeDebugAgentTest/gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=2g
```
mvicsokolova: And here I assumed that the error on TC:  `JVM crashed (This problem is reported because OutOfMemoryError has been found in the build log text)` could be fixed with more heap space, but I'm not sure

dkhalanskyjb: It may be a mistake on the part of TeamCity. The only places in the build log where I see the `OutOfMemoryError` are this message that you quote and `-XX:+HeapDumpOnOutOfMemoryError`, which technically contains the string `OutOfMemoryError` but does not indicate any problems.

mvicsokolova: The reason for the "JVM crash" was throwing exception during agent initiliazation, seems that we cannot do that.
So, I rolled back to the initial solution with the wrapper class, and now it just prints an error message, that the agent was not attached and returns without throwing an error.

integration-testing/build.gradle.kts
```diff
@@ -220,6 +221,47 @@ tasks {
...
+            it.name == "kotlinx-coroutines-core-jvm-$coroutinesVersion.jar"
+        }.singleFile
+
+        commandLine(
+            if (System.getProperty("os.name").toLowerCase().contains("windows"))
+                "gradlew.bat"
+            else
+                "./gradlew",
+            ":safeDebugAgentTest:run",
+            "-Pdebug.agent=${coroutinesDebugJar.absolutePath}"
```
dkhalanskyjb: `kotlinx.coroutines` is used in compiler integration tests, where additional parameters are passed to Gradle to indicate which compiler version to use, among other things. This command-line invocation can occasionally break in that scenario: imagine that `kotlinx.coroutines` uses `2.2.0`, but someone tests `kotlinx.coroutines` with `2.4.0-Beta`. The locally published `kotlinx-coroutines-core` JAR will be compiled by `2.4.0-Beta`, but this invocation will still use 2.2.0, which is a mismatch prohibited by the compiler (more than one major version).

integration-testing/safeDebugAgentTest/gradle.properties
dkhalanskyjb: 1. Still, why is the wrapper class needed? Can't the same logic just be part of the original class?
2. Can this `gradle.properties` be removed, then, or is it still important?

integration-testing/build.gradle.kts
mvicsokolova: Ok, I see. I just used command-line invocation to be able to intercept the output of `:safeDebugAgentTest:run` without failing the gradle task. And how is this scenario handled for smokeTest e.g.? Where does it take the compiler version from?

integration-testing/safeDebugAgentTest/gradle.properties
mvicsokolova: I used the wrapper class to avoid throwing any errors during agent initialization. It doesn't make sense to just print the message with the "dummy property" solution, as the intialization error will be thrown after the check anyway. And then we are back to the problem of the "JVM crash" in the test.
But that's just the problem of the test, we do not change the original behavior of the agent, if we throw an error, we just provide an additional message to the initialization error. 
If we're ok with that, then the test should be rewritten somehow

mvicsokolova: we can remove `gradle.properties`, yes

dkhalanskyjb (approved):
Well done, thank you!

Issue #4491, opened 2025-07-31
Port a bugfix for the coroutine scheduler


dkhalanskyjb:
https://github.com/Kotlin/kotlinx.coroutines/pull/4132/commits/cc1ad3d33dc787a7402f1f442e256e5029e04cd2 describes a bug in the coroutine scheduler. We should look at it, write tests confirming the bug, and port the fix.

LouisCAD:
What are the possible consequences of this bug?

murfel:
@vsalavatov could you provide some context for your commit, e.g. how you discovered the bug? I am also interested in this, as I might work on bringing it in.

qwwdfsad:
The comment in the commit explains the race; it's very similar to https://github.com/Kotlin/kotlinx.coroutines/issues/3660 (also, seems to be fixed with this commit as well, though requires a check).


Taking how the scheduler is implemented [and polished] at this point, all concurrency bugs have the same nature:

Under very specific circumstances, there is a so-called "liveness" miss -- there is a task in some queue, there are idle threads, but these threads are not looking for the task. The moment the scheduler state is disturbed (i.e. another task got sent), the problem disappears. Both bugs (one in comments and one in the issue) are of this nature.

You might want to look at scheduler stress tests as they are written in a very specific manner to catch bugs like that -- the scheduler is loaded and then all threads halt (with a latch or a barrier) and with a timeout we detect lack of progress. Or reset and try again.

vsalavatov:
@murfel, sure! I've had a lot of fun with this bug (literally)! :)

[...]

Issue #3660, opened 2023-03-06
Investigate potential starvation scenario in CoroutineScheduler


qwwdfsad:
Consider the following scenario (where N is the number of cores):

* `N - 1` CPU workers are indefinitely busy [with some work] and do not participate in dispatching activities
* Two workers are parked in the worker's stack
* Two tasks are scheduled internally or externally: blocking and CPU and, in order to compensate, two workers are unparked
* `W1` immediately acquires the last CPU permit and steals the blocking task, gets preempted
* `W2` fails to acquire CPU token, searches for blocking tasks only, fails to find one, goes back to the parked state (via the corresponding scan-addToStack-rescan-park procedure)
* `W1` gets scheduled again, releases the CPU token, starts executing blocking task [for indefinitely long period of time]
* CPU task is kept in the queue, and no one is looking for that even though there is a vacant CPU token

Note that scenario manifests itself in very rare circumstances when no tasks whatsoever make any progress (or maybe all these tasks completed depend on the very last CPU task scheduled), and as soon as any progress happens, the situation is immediately resolved.

We should decide whether we want to address this issue and, if so, come up with a robust mechanism to prevent such scenarios.

One of the potential solution might be to immediately compensate worker if the CPU token released during blocking task execution is the very last CPU token

PR #4528, opened 2025-09-18
Fix a race in CoroutineScheduler tryPark

dkhalanskyjb:
Fix authored by @vsalavatov.

The race could lead to CPU tasks not being executed even when CPU cores are available.

Fixes #4491

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -798,7 +798,20 @@ internal class CoroutineScheduler(
              */
             while (inStack() && workerCtl.value == PARKED) { // Prevent spurious wakeups
                 if (isTerminated || state == WorkerState.TERMINATED) break
-                tryReleaseCpu(WorkerState.PARKING)
+                val hadCpu = tryReleaseCpu(WorkerState.PARKING)
+                if (hadCpu && !globalCpuQueue.isEmpty) {
+                    /*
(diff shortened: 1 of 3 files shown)


qwwdfsad:
Note during the review: quite similar to https://github.com/Kotlin/kotlinx.coroutines/issues/3660

qwwdfsad (changes requested):
I think we shouldn't merge this one (apart from the test that should be marked as `@Ignore`).

#4491 and #3660 are issues of the same equivalence class, and the proposed fix addresses only a very specific, nailed issue.

Consider the same pattern, but manifested in a slightly different manner:

* T_CPU holds the only CPU permit, scans the tasks, doesn't find anything, places itself on a stack
* T_CPU scans again, doesn't find anything again, a thread switch happens at `tryPark()`
* T_B (or several workers in BLOCKING mode) also put themselves on the stack, on top of the T_CPU
// **NB: the diff is here**
* T_B', another blocking worker, schedules a CPU task (dispatched to its local queue), wakes up T_B
// **End of the diff**
* T_B can't acquire a CPU permit, scans blocking queue(s), doesn't find anything, parks
* T_CPU releases the CPU permit, parks there are tasks in the CPU queue, but all workers are parked, so the scheduler won't make progress until there is another dispatch

The problem is basically the same, but the fix does not address it.

The root cause is not a missing check but rather a systemic flaw -- communication between threads (e.g. stack-based parking/unparking) and handover of the CPU permits are not synchronized, which leads to a whole plethora of "park and act" races.

PR #4536, opened 2025-10-01
Depend on root components instead of platform-specific components

timofeys1:
The "-js" dependencies are an unconventional way to specify dependencies, and these dependencies are problematic for testing lenient KMP resolution, which we are doing in KT-77297.

To make sure there is no impact, I published coroutines before and after these changes. Here is the [diff](https://github.com/user-attachments/files/22645076/diff.txt). Now KMP Gradle JS consumers will see `atomicfu` component instead of `atomicfu-js` (which is the convention in KMP resolution), and there used to be some duplication in POM dependencies, which is now gone.

### DIFF
--- a/buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
@@ -53,10 +53,6 @@ kotlin {
         moduleName = project.name
         nodejs()
         compilations["main"]?.dependencies {
-            // In the past coroutines had an api dependency on "atomicfu-js" which made "atomicfu-js" klib come in front
-            // of stdlib in the compilation classpath and affected the "depends" klib manifest ordering. Remove the
-            // compileOnly dependency in the future if the ordering of "depends" doesn't matter.
-            compileOnly("org.jetbrains.kotlinx:atomicfu-js:${version("atomicfu")}")
             api("org.jetbrains.kotlinx:atomicfu:${version("atomicfu")}")
         }
     }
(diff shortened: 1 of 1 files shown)


timofeys1:
I decided to revert `depends` the ordering fix to not introduce an unnecessary workaround. Now there will also be a diff in klibs due to `depends` ordering

murfel:
https://youtrack.jetbrains.com/projects/KT/issues/KT-77297

integration-testing/smokeTest/build.gradle.kts
```diff
@@ -49,17 +49,17 @@ kotlin {
         }
         jsTest {
             dependencies {
-                implementation(kotlin("test-js"))
+                implementation(kotlin("test"))
```
dkhalanskyjb, line 52: Is there a reason not to have a single `kotlin("test")` dependency in `commonMain`, like is done in the other build scripts in this PR?

antohaby: it makes sense to keep things only in commonMain

antohaby: But there is no harm to duplicate it in commonMain and jsTest as `implementation(kotlin("test"))`

antohaby:
@dkhalanskyjb is there a chance to get it merged today/tomorrow?

test-utils is a test module that is not affecting any publications and it tests are green after the change. (particularly about this change I proposed here: https://github.com/Kotlin/kotlinx.coroutines/pull/4551). 

I'd like to merge a fix for 2.3.0-Beta2 and branching is near 😟

dkhalanskyjb:
Fixed this myself and merged manually in https://github.com/Kotlin/kotlinx.coroutines/commit/adce920f1c46218678ceee784c696d57e6eb5c34

PR #4538, opened 2025-10-05
Add joinAll for Array of Job

MukjepScarlet:


dkhalanskyjb:
We would like to encourage using `List<T>` instead of arrays, and this function goes against that.

PR #4539, opened 2025-10-09
ChannelIterator: add an example

murfel:
Add an example, stress that different coroutines *can* iterate over the channel, with their own iterator instance

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -1051,7 +1051,7 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?) 
 /**
  * Iterator for a [ReceiveChannel].
  * Instances of this interface are *not thread-safe*.
- * Each iterator instance should be used by a single coroutine.
+ * Each iterator instance should be created and used by a single coroutine.
  *
  * An example usage:
  *
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -1051,15 +1051,19 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?) 
 /**
  * Iterator for a [ReceiveChannel].
  * Instances of this interface are *not thread-safe*.
- * A coroutine is only allowed to call methods on the iterator instances which it instantiated.
+ * A coroutine is only allowed to call methods on those iterator instances which it instantiated.
  *
  * Typically, an iterator is used indirectly in the `for` loop and is not instantiated explicitly.
  *
  * ```
  * for (element in channel) {
- *     // process element
+ *     // process the element
  * }
  * ```
+ *
+ * If your use-case requires handling the iterator directly,
+ * you must call [hasNext] before each [next].
+ * Refer to [hasNext] and [next] for more details.
  *
  * An example usage:
  *
--- a/kotlinx-coroutines-core/common/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/Channel.kt
@@ -1051,7 +1051,15 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?) 
 /**
  * Iterator for a [ReceiveChannel].
  * Instances of this interface are *not thread-safe*.
- * Each iterator instance should be created and used by a single coroutine.
+ * A coroutine is only allowed to call methods on the iterator instances which it instantiated.
+ *
+ * Typically, an iterator is used indirectly in the `for` loop and is not instantiated explicitly.
+ *
+ * ```
+ * for (element in channel) {
+ *     // process element
+ * }
+ * ```
  *
  * An example usage:
  *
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -1050,7 +1050,36 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?)
 
 /**
  * Iterator for a [ReceiveChannel].
- * Instances of this interface are *not thread-safe* and shall not be used from concurrent coroutines.
+ * Instances of this interface are *not thread-safe*.
+ * Each iterator instance should be used by a single coroutine.
```
dkhalanskyjb: That's not true, it can be passed around different coroutines as long as this doesn't result in multithreaded access.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -1050,7 +1050,36 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?)
 
 /**
  * Iterator for a [ReceiveChannel].
- * Instances of this interface are *not thread-safe* and shall not be used from concurrent coroutines.
+ * Instances of this interface are *not thread-safe*.
+ * Each iterator instance should be created and used by a single coroutine.
+ *
+ * An example usage:
```
dkhalanskyjb, line 1069: Could you describe the sequence of events where this sample would answer a question a user may realistically have? It neither explains the `next()`/`hasNext()` interplay useful in deeply technical cases where you'd manipulate `ChannelIterator` manually, nor does it spell out "hey, this is not the class that you need to use manually, just write `for (... in ...)` instead" if that was the intention.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/channels/Channel.kt
murfel: I'm addressing my own confusion with the original wording - it primed me to think that only one coroutine is allowed to create instances of the iterator.

murfel: Added wording on "don't use directly" and "next/hasNext"

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -1050,7 +1050,48 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?)
 
 /**
  * Iterator for a [ReceiveChannel].
- * Instances of this interface are *not thread-safe* and shall not be used from concurrent coroutines.
+ * Instances of this interface are *not thread-safe*.
+ * A coroutine is only allowed to call methods on those iterator instances which it instantiated.
```
dkhalanskyjb: Still not true. The only limitation is that it's non-thread-safe, but like with *most* other non-thread-safe structures, you can still use it from different threads:
```kotlin
import kotlinx.coroutines.*
import kotlinx.coroutines.sync.*

suspend fun main() {
    val channel = Channel<Int>()
    val iterator: ChannelIterator<Int> = channel.iterator()
    val mutex = Mutex(locked = false)
        repeat(10) { coroutine ->
            launch {
                repeat(10) { iteration ->
                    mutex.withLock {
                        repeat(10) {
                            iterator.hasNext()
                            println("Coroutine $coroutine, iteration $iteration: ${iterator.next()}")
                        }
                    }
                }
            }
        }
        launch {
            repeat(1000) {
                channel.send(it)
            }
        }
    }
    println("Done")
}
```
No one's stopping us from protecting an iterator with an asynchronous mutex, or a synchronous one, or even sending an iterator from one coroutine to another via other channels once every dozen operations, and then operating on it with complete confidence.

This distinction is not me being pedantic. There are often cases when it actually matters which specific thread (or much more rarely, coroutine) runs some code. We *could* have added some logic to the `ChannelIterator` initializer that would check if it's still running on the same coroutine, throwing an exception if it isn't. Some APIs do operate like that. We haven't and aren't planning to. I can't imagine use cases where it makes sense to send some channel iterator to another coroutine, but it's not prohibited in the contract, nor checked in the code.

Even the original formulation is too strict, as it's possible to have safe single-threaded concurrency (on `Dispatchers.Main`, for example). The only problem is with parallelism.

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -1050,7 +1050,49 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?)
 
 /**
  * Iterator for a [ReceiveChannel].
- * Instances of this interface are *not thread-safe* and shall not be used from concurrent coroutines.
+ * An instance of this interface is *not thread-safe*.
+ * Multiple coroutines are allowed to create and use
+ * its own instances of this interface for the same channel.
```
dkhalanskyjb:
```suggestion
 * their own instances of this interface for the same channel.
```

kotlinx-coroutines-core/common/src/channels/Channel.kt
```diff
@@ -1050,7 +1050,49 @@ public inline fun <T> ChannelResult<T>.onClosed(action: (exception: Throwable?)
...
+ *
+ * An example usage:
+ *
+ * ```
+ * val channel = Channel<Int>()
+ * launch {
+ *      channel.send(1)
+ *      channel.send(2)
+ *      channel.send(3)
+ *      channel.close()  // NB: must close for iterators to finish
```
dkhalanskyjb:
```suggestion
 *      channel.close() // NB: must close for iterators to finish
```

dkhalanskyjb (approved):

Issue #4530, opened 2025-09-21
Add `asFlow()` extension for `SharedFlow`


g000sha256:
Currently, `kotlinx.coroutines` provides:
```kotlin
fun <T> MutableSharedFlow<T>.asSharedFlow(): SharedFlow<T>
```
and
```kotlin
fun <T> MutableStateFlow<T>.asStateFlow(): StateFlow<T>
```
to hide mutability.

However, there's no convenient way to hide the "hot flow" nature and return a simple `Flow<T>` that cannot be cast back to the original type. I suggest to add:
```kotlin
fun <T> SharedFlow<T>.asFlow(): Flow<T>
```

This would provide a clean, zero-overhead way to expose hot flows as a plain `Flow` interface.

### Use cases
- `API` design where implementation details (`SharedFlow`/`StateFlow`) should not leak.
- Preventing clients from accessing `replayCache`, `subscriptionCount`, or `value`.
- Consistent return types in interfaces that may use different `Flow` implementations.

Komdosh:
What's wrong with:

```kotlin
val values: Flow<Value> = _mutableSharedFlow.asSharedFlow()
```

g000sha256:
@Komdosh, yes, `SharedFlow` is already a `Flow`, but a simple assignment allows casting back to the original type:
```kotlin
class Repository {
    private val _mutableSharedFlow = MutableSharedFlow<Value>()
    val values: Flow<Value> = _mutableSharedFlow.asSharedFlow()
}

// Client code can break encapsulation:
val sharedFlow = repository.values as SharedFlow<Value> // Success!
// Now the client can access replayCache, subscriptionCount, etc..
```

The proposed `asFlow()` would return a wrapper that prevents such casting, similar to how `asSharedFlow()` prevents casting back to `MutableSharedFlow`:
```kotlin
val values: Flow<Value> = _mutableSharedFlow.asFlow()

// Client code:
val sharedFlow = repository.values as SharedFlow<Value> // ClassCastException!
```
This is about enforcing encapsulation at runtime, not just at compile time. It ensures that implementation details (using `SharedFlow` internally) don't leak to consumers who should only depend on the `Flow` interface.

Komdosh:
So this is an error too?

```
public fun main(){
    val l: List<String> = mutableListOf("0", "1").toList()
    (l as MutableList<String>).removeAt(0)
    println(l) // ["1"]
}
```

g000sha256:
[...]

PR #4541, opened 2025-10-13
Add `asFlow` operator

g000sha256:
See #4530

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Share.kt
@@ -356,7 +356,26 @@ private fun <T> CoroutineScope.launchSharingDeferred(
 // -------------------------------- asFlow --------------------------------
 
 /**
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/flow/operators/Share.kt
```diff
@@ -353,6 +353,15 @@ private fun <T> CoroutineScope.launchSharingDeferred(
     }
 }
 
+// -------------------------------- asFlow --------------------------------
+
+/**
+ * Represents this shared flow as a flow.
```
dkhalanskyjb: Please expand the documentation:
* Differences from simple upcasting should be explained.
* The purpose is not immediately clear.
* A usage example would be nice, as it's not always obvious where to stick `asFlow()`.

g000sha256: ✅

PR #4543, opened 2025-10-15
Cleanup: remove the `BuilderInference` annotation

dkhalanskyjb:
The `BuilderInference` annotation has no effect since Kotlin 2.0 and can be safely removed.


murfel:
Just to confirm, this annotation does not affect the ABI?

dkhalanskyjb:
Yes, the annotation has no effect whatsoever, for other libraries as well, though I already let the stdlib know.

The ABI hasn't changed, or the ABI dump verification would have made the CI fail (the `checkLegacyAbi` Gradle task, which is part of `check`)

murfel:
@dkhalanskyjb great! what about all our other libs?

dkhalanskyjb:
I'm not aware of them using `@BuilderInference`.

Issue #4199, opened 2024-07-29
Consider deprecation cycle for `CoroutineDispatcher.invoke`


JakeWharton:
`CoroutineDispatcher.invoke`, a function used as such:
```kotlin
Dispatchers.IO {
  stuff
}
```
is a shorter version of `withContext` that only works with dispatchers.

Does this function need to exist? It seems to predate structured concurrency and is so terse that it offers no indication as to its underlying behavior. Does it launch a new coroutine? Does it move the current coroutine? Does the supplied block honor structured concurrency?

I personally have a hard time finding a reason for this function to continue to exist.

CLOVIS-AI:
For context, it was originally added in #428 as an experiment. As far as I know, it is very rarely used (if at all) in the ecosystem.

dkhalanskyjb:
Very few usages found:
https://grep.app/search?q=.IO%20%7B&case=true&filter[lang][0]=Kotlin
https://grep.app/search?q=Dispatchers.Default%20%7B&case=true&filter[lang][0]=Kotlin
https://grep.app/search?q=Dispatchers.Main%20%7B&case=true&words=true&filter[lang][0]=Kotlin

I personally find this to be non-idiomatic modern Kotlin. `Thing { }` looks to me like a builder of some sort. So, I'm for this.

qwwdfsad:
We can try and proceed with it in RC2 along with other deprecations

dkhalanskyjb:
I'm not sure what I expected, but the deprecation doesn't look pretty in the IDE!
![Screenshot_20240801_134245](https://github.com/user-attachments/assets/7e350091-29ee-4910-a7ec-7ac4ab24e444)
Given how rare this is, I don't think this should stop us, though.

dkhalanskyjb:
[Auto-replacement also doesn't work](https://github.com/user-attachments/assets/77ff6e6d-0678-42ee-b9c3-f7238e9d701a)

JakeWharton:
Finding some good IntelliJ bugs/feature requests at least!

LouisCAD:
You can say I'm biased, but I like it because it's concise and clear enough once you know what a coroutine dispatcher is.

`withContext` is a lower-level construct, that can also be misused since it accepts the `CoroutineContext` type, which can notoriously(?) contain a `Job`, breaking structured concurrency. 

[...]

PR #4545, opened 2025-10-16
Upgrade to Kotlin 2.2.20

dkhalanskyjb:
* 2.2.20 added new opt-in requirements for using Wasm/JS APIs.
* It also introduced a new `web` source set that is equivalent to `jsAndWasmJsShared` that we used to have before.
* In addition, it now checks the syntax inside the `js()` calls and forbids returning `JsAny` (or even `JsException`). Our current code doesn't compile with those restrictions, so I had to change the `propagateExceptionFinalResort` implementation to avoid repeatedly diving into the JS world and back to Kotlin; instead, there's only one `js()` block now that doesn't return anything.

### DIFF
--- a/kotlinx-coroutines-core/web/src/internal/CoroutineExceptionHandlerImpl.kt
+++ b/kotlinx-coroutines-core/web/src/internal/CoroutineExceptionHandlerImpl.kt
@@ -7,7 +7,7 @@ import kotlin.js.*
  * rather than in the current execution branch.
  */
 @OptIn(ExperimentalWasmJsInterop::class)
-internal fun throwAsyncJsError(message: String?, className: String?, stack: String?) {
+internal fun throwAsyncJsError(message: String?, className: String?, stack: String) {
     js("""
         var error = new Error();
         error.message = message;


README.md
murfel, line 24: This one needs to be updated to 2.2.20 as well.

dkhalanskyjb: The link resolves to 2.2.20 for me, there's no 2.2.0 anymore on play.kotlinlang.org.

LouisCAD:
[You can now](https://youtrack.jetbrains.com/issue/KT-32313/Support-contracts-for-operator-functions) add a contract to `operator fun CoroutineDispatcher.invoke`, just like `withContext` has.

dkhalanskyjb:
@LouisCAD, we should probably just remove `CoroutineDispatcher.invoke` instead: https://github.com/Kotlin/kotlinx.coroutines/issues/4199

kotlinx-coroutines-core/web/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,22 @@
+package kotlinx.coroutines.internal
+
+import kotlin.js.*
+
+/*
+ * Schedule an exception to be thrown inside JS or Wasm/JS event loop,
+ * rather than in the current execution branch.
+ */
+@OptIn(ExperimentalWasmJsInterop::class)
+internal fun throwAsyncJsError(message: String?, className: String?, stack: String?) {
```
murfel: -> `stack: String`, since `stackTraceToString()` returns `String`

kotlinx-coroutines-core/web/src/internal/CoroutineExceptionHandlerImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+        error.message = message;
+        error.name = className;
+        error.stack = stack;
+        setTimeout(function () { throw error }, 0);
+    """)
+}
+
+internal actual fun propagateExceptionFinalResort(exception: Throwable) = with(exception) {
+    throwAsyncJsError(message, this::class.simpleName, stackTraceToString())
+}
```
murfel: newline

murfel:
Approved ✅

PR #4547, opened 2025-10-17
Cleanup: reduce the number of warnings during a build

dkhalanskyjb:


murfel:
Any advice on how to review this?

E.g. which parts were not the obvious fix for a warning?

dkhalanskyjb:
The majority of changed lines follow one of several patterns, so it's enough to understand an edit once, and then the rest of the instances become clear as well.

Examples:
* Removed `@Suppress`: if no new warnings/errors surfaced, then this `@Suppress` was already redundant, as its whole purpose is to hide warnings and errors we disagree with.
* Removed type annotation on a function call: likewise, this is just historical baggage from when Kotlin was worse at inferring types.
* `@Suppress("DEPRECATION")` added to tests that check the behavior of a deprecated API (using which causes a warning) whose correctness we still want to ensure.
* Added `?` to types coming from Java that the compiler now knows can be `null`.

There are a few non-trivial fixes, but most changes are just boilerplate.

murfel:
Could you describe the process for these two? The process itself is partly proof of correctness (and the other part is the clean compilation).

E.g. which `@Suppress` / type annotations did you attempt removing?

>Removed @Suppress: if no new warnings/errors surfaced, then this @Suppress was already redundant, as its whole purpose is to hide warnings and errors we disagree with.

> Removed type annotation on a function call: likewise, this is just historical baggage from when Kotlin was worse at inferring types.

dkhalanskyjb:
Sure.

For `@Suppress`, the process was:
1. Observe a warning like "suppressing errors may break the behavior".
2. Attempt removing the `@Suppress`.
3. If a new error appears that was hidden by the `@Suppress`, restore the `@Suppress`.

Steps 2 and 3 can be replaced with just looking at the code or preliminary knowledge or assumptions, so some `@Suppress` annotations have been left untouched.

For removing type annotations: when I was fixing some warnings in a file, I also went through the suggestion provided by the IDE, considered if applying them is proper, and then I either did that or reported a broken suggestion to the IDE team.

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
murfel: `select<String>` - `<String>` can also be removed

murfel: Not sure how you discovered the other entries for the "Remove explicit type arguments" inspection, by hand or by following warnings. If the latter, not sure why this and other few cases were missed. Possibly also clean them up?

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
```diff
@@ -151,4 +152,4 @@ class AtomicCancellationCommonTest : TestBase() {
         yield() // now yield
         finish(4)
     }
-}
+}
```
murfel: Newline

murfel: There's a checkbox in Idea, "ensure every saved file ends with a line break"

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
```diff
@@ -131,6 +131,7 @@ class AtomicCancellationCommonTest : TestBase() {
         finish(4)
     }
 
+    @Suppress("DEPRECATION")
```
murfel, line 134: Could you add `// Mutex.onLock`, since there's no hint neither on github, nor in Idea after the suppress has been added

kotlinx-coroutines-core/common/test/JobStatesTest.kt
```diff
@@ -11,7 +9,7 @@ import kotlin.test.*
  */
 class JobStatesTest : TestBase() {
     @Test
-    public fun testNormalCompletion() = runTest {
```
murfel, line 14: There's another 15 entries for `public fun` in `common/test`, and most of them don't need the public modifier. Could you get rid of them as well?

kotlinx-coroutines-core/common/test/WithTimeoutDurationTest.kt
```diff
@@ -183,14 +181,15 @@ class WithTimeoutDurationTest : TestBase() {
     @Test
     fun testExceptionFromWithinTimeout() = runTest {
         expect(1)
+        @Suppress("UNREACHABLE_CODE")
```
murfel, line 184: Please add `// expectUnreached()` here and everywhere else

murfel: Do you happen to know why this particular `expectUnreached()` triggered the inspection, but many others didn't?

kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
```diff
@@ -1,7 +1,6 @@
 package kotlinx.coroutines
 
 import kotlinx.coroutines.testing.*
-import org.junit.Test
```
murfel, line 4: Could you confirm if other `import org.junit.Test` and `import org.junit.*` in the project are in fact necessary?

kotlinx-coroutines-test/common/test/RunTestTest.kt
```diff
@@ -50,44 +51,6 @@ class RunTestTest {
         }
     }
 
-    /** Tests that even the dispatch timeout of `0` is fine if all the dispatches go through the same scheduler. */
```
murfel, line 53: Why rearranging the file?

reactive/kotlinx-coroutines-reactor/test/Check.kt
```diff
@@ -5,7 +5,7 @@ import reactor.core.publisher.Mono
 
 fun <T> checkMonoValue(
         mono: Mono<T>,
-        checker: (T) -> Unit
+        checker: (T?) -> Unit
```
murfel, line 8: ?

reactive/kotlinx-coroutines-reactor/test/ReactorContextTest.kt
```diff
@@ -100,9 +100,9 @@ class ReactorContextTest : TestBase() {
 
     private fun bar(): Flow<String> = flux {
         val ctx = reactorContext()
-        (1..3).forEach { send(ctx.getOrDefault(it, "noValue")) }
+        (1..3).forEach { send(ctx.getOrDefault(it, "noValue")!!) }
```
murfel, line 103: ?

ui/kotlinx-coroutines-android/testdata/r8-test-common.pro
```diff
@@ -15,4 +15,4 @@
 }
 
 # We are cheating a bit by not having android.jar on R8's library classpath. Ignore those warnings.
--ignorewarnings
\ No newline at end of file
+-dontwarn
```
murfel, line 18: ?

ui/kotlinx-coroutines-android/testdata/r8-test-rules.pro
```diff
@@ -6,7 +6,7 @@
 -checkdiscard class kotlinx.coroutines.internal.FastServiceLoader
 -checkdiscard class kotlinx.coroutines.DebugKt
 -checkdiscard class kotlinx.coroutines.internal.StackTraceRecoveryKt
--checkdiscard class kotlinx.coroutines.debug.DebugProbesKt
+-checkdiscard class kotlinx.coroutines.debug.internal.DebugProbesKt
```
murfel, line 9: ?

murfel (commented):
Thank you for taking this cleanup on! I was pretty annoyed with all the warnings.

? in my comments mean please provide context/reference/proof for the correctness of the change.

---

I'd also like a prove of completeness of certain replacements. E.g. as if with the `publc` modifier in test, the proof is empty grep in Idea. The proof can come just as a confirmation statement, "grep for `xxx` in directory `yyy` is empty", then I can check it in O(1).

For now I'm not sure when you did or didn't aim for completeness.

I also not sure how you discovered the warnings. Did you use compilation warnings only? If I run "inspect project" in Idea, I see many more warnings, some of them similar to the ones fixed here.

---
Just FYI: here's a context to which changes are easier to check during review

#### Easy to check
Removals
- `@Suppress`
- inferable type parameters
- not-null assertions (!!)
- `public` modifiers from tests
- imports

These removals are easy because if the CI passes after the change, then the change is valid. Hence, I don't need to think about it.

And other things
- Rename the unused parameter in the catch clause to `_`

#### Harder to check
- Adding suppresses (since we don't have a check "this suppress doesn't suppress anything")
- Loosening the type to nullable (adding ?)
- Adding not-null assertion (!!)
- Changes not triggered by inspections (If I undo the change, I don't see any _static_ warning that triggers me)

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
murfel: Here and in other entries of `@Suppress("DEPRECATION")`

kotlinx-coroutines-test/common/test/RunTestTest.kt
dkhalanskyjb: To have one `Suppress` for a group of tests.

reactive/kotlinx-coroutines-reactor/test/Check.kt
dkhalanskyjb: > Added `?` to types coming from Java that the compiler now knows can be null.

reactive/kotlinx-coroutines-reactor/test/ReactorContextTest.kt
dkhalanskyjb: > Added `?` to types coming from Java that the compiler now knows can be null.

ui/kotlinx-coroutines-android/testdata/r8-test-common.pro
dkhalanskyjb: Try running `:kotlinx-coroutines-android:runR8` with and without this change, and the difference should be clear. `ignorewarnings` doesn't suppress the warnings we're not interested in.

ui/kotlinx-coroutines-android/testdata/r8-test-rules.pro
dkhalanskyjb: There's no `kotlinx.coroutines.debug.DebugProbesKt` now, ever since it `DebugProbes.kt` got moved to the `internal` package, so the rule was meaningless before the change.

kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
dkhalanskyjb: `import org.junit.Test` is necessary unless there's `import kotlin.test.Test` (or `import kotlin.test.*`), which are equivalent. `import org.junit.*` may be difficult to get rid of, even if `kotlin.test.Test` is used, as JUnit has some APIs that `kotlin.test` doesn't provide (test parameterization, for one: https://youtrack.jetbrains.com/issue/KT-46899).

kotlinx-coroutines-core/common/test/WithTimeoutDurationTest.kt
dkhalanskyjb: `expectUnreached()` is not some specific construct, it's just a function call. The inspection gets triggered when a value of type `Nothing` gets returned. `Nothing` is a type with no values, and you can only return it by throwing an exception or having an infinite loop in your code. Here, `withTimeout` returns `Nothing`, because its body never produces a normal value, and the compiler can recognize that. If we add a type parameter (`withTimeout<Int>`, for example), the compiler won't be able to do that.

kotlinx-coroutines-core/jvm/test/ExecutorsTest.kt
murfel: Are you aiming for completeness in this PR?

If so, could you confirm if you checked/fixed other similar imports, and so, could you provide a verification (e.g. "grep for `xxx` in `yyy` produces 5 entries all of which are necessary, since they don't have `import kotlin.test.Test`")?

dkhalanskyjb: I was aiming for completeness in fixing the warnings. While editing the files containing the warnings, I also looked through the IDE suggestions (which don't cause warnings) and stylistically outdated forms as well. The latter was not part of the goal of the pull request, it is only a fly-by improvement, using the principle of leaving the file in a better state than I found it. Any pull request from me can contain such small improvements.

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
dkhalanskyjb: Such comments can become outdated when additional APIs used in the test get deprecated. I'd rather not risk misleading the reader. The more reliable method to learn which particular API needs the suppression is to temporarily remove the annotation.

murfel: I see, but without the comment, there's no way to know which suppress the author meant when they wrote that suppress.

(Either way, would be nice to have all suppresses in the codebase either commented, or un-commented. Not necessarily by your PR, of course.)

kotlinx-coroutines-core/jvm/test/exceptions/Stacktraces.kt
```diff
@@ -20,23 +20,23 @@ public fun verifyStackTrace(e: Throwable, vararg traces: String) {
...
+fun String.normalizeStackTrace(): String =
     replace(Regex(":[0-9]+"), "") // remove line numbers
     .replace("kotlinx_coroutines_core_main", "") // yay source sets
     .replace("kotlinx_coroutines_core", "")
     .replace(Regex("@[0-9a-f]+"), "") // remove hex addresses in debug toStrings
     .lines().joinToString("\n") // normalize line separators
 
-public fun String.count(substring: String): Int = split(substring).size - 1
\ No newline at end of file
+fun String.count(substring: String): Int = split(substring).size - 1
```
murfel: Newline

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
dkhalanskyjb: I was following warnings in the files I've touched. I did not actively seek out the places where this inspection triggers, as an inspection is not a nuisance until you actually touch the file, and when you do, sure, fix the inspection, why not.

kotlinx-coroutines-core/common/test/AtomicCancellationCommonTest.kt
dkhalanskyjb: I don't see how the authorial intent matters here. The suppression affects the compilation process only, not the behaviors observable by the user, so rigor here feels like a waste of effort to me.

kotlinx-coroutines-core/common/test/JobStatesTest.kt
dkhalanskyjb: This commit removes the `public ` from the test source sets throughout the project: https://github.com/Kotlin/kotlinx.coroutines/pull/4547/commits/44927b6c0a82e5256b963ed7e72105267333cc5f

kotlinx-coroutines-core/jvm/test/exceptions/Stacktraces.kt
dkhalanskyjb: Fixed.

PR #4549, opened 2025-10-17
Update atomicfu version to 0.30.0-beta

murfel:


dkhalanskyjb:
Let's put this in `develop` so that we catch any regressions there may be while we work on our code normally.

PR #4551, opened 2025-10-21
Update kotlin-test dependency declaration in test-utils

antohaby:
kotlin-test is KMP dependency that is intended to be consumed via root coordinates.

There is no need to depend on kotlin-test-common explicitly and on platfom-specific source sets.

It also fixes potential problems with SourceSet visibility algorithm. i.e. it is curcial to have connected graph of dependencies.
i.e. kotlin-test-common and kotlin-test-js for example is not connected from this algorithm point of view.

### DIFF
--- a/test-utils/build.gradle.kts
+++ b/test-utils/build.gradle.kts
@@ -5,29 +5,14 @@
 kotlin {
     sourceSets {
         commonMain.dependencies {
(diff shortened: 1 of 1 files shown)


dkhalanskyjb:
Duplicates part of https://github.com/Kotlin/kotlinx.coroutines/pull/4536

PR #4552, opened 2025-10-22
Depend on root components instead of platform-specific components

dkhalanskyjb:
The "-js" dependencies are an uncoventional way to specify dependencies and these dependencies are problematic for testing lenient KMP resolution.


dkhalanskyjb:
Merged manually in https://github.com/Kotlin/kotlinx.coroutines/commit/adce920f1c46218678ceee784c696d57e6eb5c34

PR #4554, opened 2025-10-22
Enhance CoroutineDispatcher.invoke with contract

LouisCAD:
Add contract to CoroutineDispatcher.invoke operator, to allow initializing read-only properties declared before, from inside the passed block.

https://youtrack.jetbrains.com/issue/KT-32313/Support-contracts-for-operator-functions

This commit also improve the KDoc to show a usage example, and explain why it's less error-prone than withContext.

### DIFF
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -177,10 +177,25 @@ public suspend fun <T> withContext(
  * completes, and returns the result.
  *
  * This inline function calls [withContext].
+ *
+ * Example usage:
+ * ```
+ * Dispatchers.IO { someFile.exists() } // Checking the file system is blocking I/O.
+ * ```
+ * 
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/Builders.common.kt
```diff
@@ -177,10 +177,25 @@ public suspend fun <T> withContext(
  * completes, and returns the result.
  *
  * This inline function calls [withContext].
+ *
+ * Example usage:
+ * ```
+ * Dispatchers.IO { someFile.exists() } // Checking the file system is blocking I/O.
```
qwwdfsad, line 183:
```suggestion
 * Dispatchers.IO { 
 *     someFile.exists() // Checking the file system is blocking I/O
 * }
```

qwwdfsad (commented):
Thanks! Could you please also add a test for that? (i.e. like `RunBlocking.testContract`)

dkhalanskyjb:
Let's reach a conclusion on #4199 first before going through with this, so that we don't risk doing extra work that will ultimately not be used.

PR #4563, opened 2025-10-29
Move Promise-related functions to the webMain source set

### LINKED ISSUE #4526: Cannot use the `promise` extension function on `CoroutineScope` from the new common web target
I cannot use the `promise` extension function on `CoroutineScope` from the new common web target. This worked from within wasmJsMain but does not if I move the relevant code to webMain. See discussion in Slack: https://kotlinlang.slack.com/archives/C3PQML5NU/p1758108932037429?thread_ts=1758060473.617919&cid=C3PQML5NU

### LINKED ISSUE #4544: Make Promise.await common between k/wasm and k/js sources
**Use case**
With kotlinx-browser:0.5.0 allowing a lot of W3C API usages being made in webMain shared source set (for both k/js and k/wasm), I'd like to have as little k/js and k/wasm specific or custom code as possible. 

The `Promise` type is already common. Many W3C API return a `Promise` and it's convenient to use them in suspend functions. For example:

```kotlin
window.fetch(someUrl).await<Response>()
```

Right now I have such a snippet to simplify the code:

```kotlin
private suspend fun <R : JsAny?> Promise<R>.await(): R = suspendCancellableCoroutine { continuation ->
    this.then(
        onFulfilled = { continuation.resumeWith(Result.success(it)); null },
        onRejected = { continuation.resumeWithException(it.asJsException()); null }
    )
}
```

**The Shape of the API**

The API would look like the existing `Promise.await` but common between k/js and k/wasm. I assume a non-breaking update is possible. 

**Prior Art**
N/A

cc: @bashor @JSMonk

dkhalanskyjb:
Fixes #4544
Fixes #4526

On Wasm/JS, before this commit, we had:

```kotlin
fun <T> CoroutineScope.promise(
    start: CoroutineStart = CoroutineStart.DEFAULT,
    block: suspend CoroutineScope.() -> T
): Promise<JsAny?>

fun <T> Deferred<T>.asPromise(): Promise<JsAny?>

fun <T> Promise<JsAny?>.asDeferred(): Deferred<T>

suspend fun <T> Promise<JsAny?>.await(): T
```

These signatures are either losing type information (`promise`, `asPromise`) or are type-unsafe (`asDeferred`, `await`).

The way an arbitrary `T` is converted into a `JsAny?` is by calling

```kotlin
public actual fun <T : Any> T.toJsReference(): JsReference<T> =
    implementedAsIntrinsic
```

Note the `Any` type bound, it's going to be important.

The opposite direction is taken by just an unchecked type cast:

```kotlin
value as T
```

The *correct* types for the current implementations of the functions we have are:

```kotlin
fun <T> CoroutineScope.promise(
    start: CoroutineStart = CoroutineStart.DEFAULT,
    block: suspend CoroutineScope.() -> T?
): Promise<JsReference<T & Any>?>

fun <T> Deferred<T?>.asPromise(): Promise<JsReference<T & Any>?>

fun <T> Promise<T: JsAny?>.asDeferred(): Deferred<T>

suspend fun <T: JsAny?> Promise<T>.await(): T
```

In principle, since Kotlin/Wasm/JS is experimental, we can break the existing usages if it is necessary.

1. Of course, we don't have `& Any`, which is an issue for this use case. If `block` returns a `null`, then the whole `Promise` resolves to `null`, and if not, it resolves to the proper `JsReference<T>` where `T` is non-`null`.
2. There is a behavior in Kotlin/Wasm/JS that makes the correct types for `await()` and `asDeferred()` incompatible with the current ones: `JsReference<T> as T` works, but does not happen automatically, so where you had `await<String>()` before, now, you will only be able to do `await<JsReference<String>>().get()`.

The first issue can be worked around by having two overloads:

```kotlin
fun <T: Any> CoroutineScope.promise(
    start: CoroutineStart = CoroutineStart.DEFAULT,
    block: suspend CoroutineScope.() -> T?
): Promise<JsReference<T>?>

fun <T: Any> Deferred<T?>.asPromise(): Promise<JsReference<T>?>

// These last two are questionable:

fun <T: Any> CoroutineScope.promise(
    start: CoroutineStart = CoroutineStart.DEFAULT,
    block: suspend CoroutineScope.() -> T
): Promise<JsReference<T>>

fun <T: Any> Deferred<T>.asPromise(): Promise<JsReference<T>>
```

This prevents writing code polymorphic in `T: Any?` that uses these functions, but this limitation may be okay in practice.

The second issue may also be solved by introducing extra overloads, but it doesn't look pretty:

```kotlin
suspend fun <T: JsAny?> Promise<T>.await(): T

// This one is questionable:

suspend fun <T: Any> Promise<JsReference<T>>.await(): T
```

These two signatures are also interchangeable with those that Kotlin/JS had before this commit:

```kotlin
fun <T> Promise<T: JsAny?>.asDeferred(): Deferred<T>

suspend fun <T: JsAny?> Promise<T>.await(): T
```

Since both Kotlin/JS and these functions are stable, we need to preserve their behaviors, which means

These, however, are a challenge:

```kotlin
fun <T: JsAny?> CoroutineScope.promise(
    start: CoroutineStart = CoroutineStart.DEFAULT,
    block: suspend CoroutineScope.() -> T
): Promise<T>

fun <T: JsAny?> Deferred<T>.asPromise(): Promise<T>
```

On Kotlin/JS, `promise` and `asPromise` are polymorphic in `T` over an arbitrary `T`, but Wasm/JS extends the universe of types with Kotlin-specific ones.

The ideal solution would be to provide both on Wasm/JS: have `fun promise` in `webMain` that would only work with `JsReference`, and in `wasmJsMain`, also have a `fun promise` that would accept arbitrary types; the same goes for `Deferred.asPromise`. Then, overload resolution could kick in and choose the most specific overload.

Unfortunately, Kotlin does not allow this:

```kotlin
internal interface A

internal fun <T: A> b(block: () -> T) {
}

internal fun <T> b(block: () -> T) {
}

internal fun c() {
    b { 3 } // does not compile
}
```

We could mark the general implementation in `webMain` with `LowPriorityInOverloadResolution`, but then,
that implementation does not get chosen at all in Kotlin/Wasm/JS code:

```kotlin
internal interface A

@Suppress("INVISIBLE_REFERENCE")
internal fun <T: A> b(block: () -> T): String {
    return ""
}

internal fun <T> b(block: () -> T): T {
    return block()
}

internal fun c() {
    val v: Int = b { 3 } // does compile now
    val s: String = b { object: A {} } // but this doesn't
}
```

The same goes for `Deferred.asPromise`:

```kotlin
fun <T: Any> Deferred<T>.asPromise(): Promise<JsReference<T>> { TODO() }
fun <T: JsAny?> Deferred<T>.asPromise(): Promise<T> { TODO() }

val v = async {
    "OK".toJsReference()
}
v.asPromise() // fails to compile
```

Therefore, any `promise` or `asPromise` implementations in `webMain` necessarily *preclude* the implementations that works with arbitrary Kotlin types in Kotlin/Wasm/JS.

For compatibility with JS, `promise` and `Deferred.asPromise` common to `web` are limited to the JS types, even on Wasm/JS, which can't be fixed by adding extra overloads to Wasm/JS.

On the Wasm/JS side, this is a breaking change.
It can usually be mitigated by sticking a
`JsReference::get` or `Any::toJsReference` to relevant places, but not always.
- For example, if third-party code exposes a `Deferred<String>`, the users will be out of luck trying to obtain a `Promise<JsString>` out of that.
- More annoyingly, `GlobalScope.promise { }` launched for the side effects will no longer compile, as `Unit` is not `JsAny`. Every such lambda now has to end with some `JsAny` value, for example, `null`. This change was introduced throughout our codebase as well, as we are also affected by this breakage.

### DIFF
--- a/kotlinx-coroutines-core/web/test/PromiseTest.kt
+++ b/kotlinx-coroutines-core/web/test/PromiseTest.kt
@@ -59,6 +59,28 @@ class PromiseTestWeb : TestBase() {
     }
 
     @Test
+    fun testCompletedDeferredAsPromiseString() = GlobalScope.promise {
+        val deferred = async(start = CoroutineStart.UNDISPATCHED) {
+            // completed right away
+            "OK".toJsString()
+        }
+        val promise = deferred.asPromise()
+        assertEquals("OK", promise.await().toString())
+        null
+    }
+
+    @Test
+    fun testWaitForDeferredAsPromiseString() = GlobalScope.promise {
+        val deferred = async {
+            // will complete later
+            "OK".toJsString()
+        }
+        val promise = deferred.asPromise()
+        assertEquals("OK", promise.await().toString()) // await yields main thread to deferred coroutine
+        null
+    }
+
+    @Test
     fun testCancellableAwaitPromise() = GlobalScope.promise {
         lateinit var r: (JsAny) -> Unit
         val toAwait = Promise<JsAny?> { resolve, _ -> r = resolve }


kotlinx-coroutines-core/web/src/Promise.kt
```diff
@@ -31,26 +30,26 @@ internal fun promiseGetDeferred(promise: Promise<JsAny?>): JsAny? = js("""{
  * @param start coroutine start option. The default value is [CoroutineStart.DEFAULT].
  * @param block the coroutine code.
  */
-@ExperimentalWasmJsInterop
-public fun <T> CoroutineScope.promise(
+@OptIn(ExperimentalWasmJsInterop::class)
```
dkhalanskyjb, line 33: It's unclear to me if we want to propagate the opt-in here instead. Are these functions *also* part of the experimental Wasm/JS interop, or are they fine, even when `Promise` is not? The description of the annotation https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.js/-experimental-wasm-js-interop/ does not make it clear to me.

The approach taken with `Promise` (which this PR could also replicate when needed) is to have `@ExperimentalWasmJsInterop expect` in `web`, `@ExperimentalWasmJsInterop actual` in `wasmJs`, and `actual` in `js` (without the annotation).

kotlinx-coroutines-core/web/test/PromiseTest.kt
```diff
@@ -4,52 +4,58 @@ import kotlinx.coroutines.testing.*
...
+        assertEquals("OK", promise.await().get())
+        null
     }
 
     @Test
     fun testWaitForDeferredAsPromise() = GlobalScope.promise {
         val deferred = async {
             // will complete later
-            "OK"
+            "OK".toJsReference()
```
JSMonk, line 54: @dkhalanskyjb would it work with `toJsString()`? I would prefer to test conversion into `Js$TYPE_NAME` instead of `toJsReference`

dkhalanskyjb: It would. Added a new pair of tests that check that: https://github.com/Kotlin/kotlinx.coroutines/pull/4563/commits/8316ce2d0998b06d039c050a215a7451641be354

murfel:
Shall we possibly contribute the overload resolution potential use-case to the compiler team?

kotlinx-coroutines-core/web/src/Promise.kt
dkhalanskyjb: Another approach I tried to get to more granular `ExperimentalWasmJsInterop` opt-ins is
```kotlin
// web
@OptionalExpectation
internal expect annotation class ExperimentalWasmJsInteropWasmJsOnly

// wasmJs
internal actual typealias ExperimentalWasmJsInteropWasmJsOnly =
    ExperimentalWasmJsInterop
```

This causes an error:
```
[OPT_IN_MARKER_CAN_ONLY_
BE_USED_AS_ANNOTATION_OR_ARGUMENT_IN_OPT_IN] This class can only be used as an annotation or as an argument to '@OptIn'.
```

PR #4567, opened 2025-11-10
Fix : Dispatcher failures may leave coroutines uncompleted #4209

HarshSingh011:
Prevent coroutine hang by safely handling synchronous dispatcher throws and falling back


HarshSingh011:
@dkhalanskyjb  can you please review my pr as I have tried fixing "Dispatcher failures may leave coroutines uncompleted" so that If still there is some issue I will try to improve it

dkhalanskyjb:
Hello! The approach taken here is a compromise with important downsides to it, which means it needs to at least be discussed first before implementing.

Also, there are numerous problems with the pull request: comments not matching the behavior, unnecessary code duplication, not following the existing layout of functions, pointless tests that don't check anything.

Finally, this looks to me like the output of an AI tool. Please don't send out pull requests generated by the AI. They don't help in the least, but take away the time needed to review them.

PR #4572, opened 2025-11-13
Removed redundant enabling tracking klib-based targets

shanshin:
In the BCV implementation as part of KGP, klib-based targets tracking is enabled by default, so there is no need to explicitly enable it.


shanshin:
The changes are necessary to change the experimental API in BCV.
[The changes](https://jetbrains.team/p/kt/reviews/24436/timeline) and [the current build log](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_Aggregate/819553626).

After merging these changes, the build should be successful.

woainikk:
As the change is already in develop, I will rebase the kotlin-cpmmunity/dev branch onto develop, so the MR can be closed,.

PR #4574, opened 2025-11-18
Add the EXACTLY_ONCE contract to `suspendCancellableContinuation`

dkhalanskyjb:
`suspendCancellableContinuation` invokes its block exactly once. `suspendCoroutineUninterceptedOrReturn`,
which is used for implementing `suspendCancellableContinuation`, already has a contract stating just that,
so for consistency and completeness, we add a contract to `suspendCancellableContinuation` itself as well.

This will let the compiler recognize that `val`-variables can be safely assigned inside the lambda, as they won't be reassigned or left uninitialized (which would be the case if the lambda executed more than once or possibly not execute at all).

There were no use cases actually reported for this.

### DIFF
--- a/kotlinx-coroutines-core/common/test/CancellableContinuationTest.kt
+++ b/kotlinx-coroutines-core/common/test/CancellableContinuationTest.kt
@@ -134,4 +134,15 @@ class CancellableContinuationTest : TestBase() {
         })
         finish(5)
     }
+
+    /** Tests that the compiler recognizes that [suspendCancellableCoroutine] invokes its block exactly once. */
+    @Test
+    fun testSuspendCancellableCoroutineContract() = runTest {
+        val i: Int
+        suspendCancellableCoroutine { cont ->
+            i = 1
+            cont.resume(Unit)
+        }
+        assertEquals(1, i)
+    }
 }
--- a/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
+++ b/kotlinx-coroutines-core/common/src/CancellableContinuation.kt
@@ -1,6 +1,7 @@
 package kotlinx.coroutines
 
 import kotlinx.coroutines.internal.*
+import kotlin.contracts.*
 import kotlin.coroutines.*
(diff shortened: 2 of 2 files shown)


kotlinx-coroutines-core/common/test/CancellableContinuationTest.kt
```diff
@@ -134,4 +134,15 @@ class CancellableContinuationTest : TestBase() {
         })
         finish(5)
     }
+
+    /** Tests that the compiler recognizes that [suspendCancellableCoroutine] invokes its block exactly once. */
```
murfel, line 138: by allowing the `val` variable

(the point of focus could be not immediately clear in future)

kotlinx-coroutines-core/common/test/CancellableContinuationTest.kt
```diff
@@ -134,4 +134,15 @@ class CancellableContinuationTest : TestBase() {
...
+    @Test
+    fun testSuspendCancellableCoroutineContract() = runTest {
+        val i: Int
+        suspendCancellableCoroutine { cont ->
+            i = 1
+            cont.resume(Unit)
+        }
+        assertEquals(1, i)
+    }
 }
```
murfel, line 148: Could you set your editor to add newlines?

dkhalanskyjb: It's not at all straightforward, I'm afraid. Of course, I will make sure to do that if I learn that the lack of newlines in Kotlin files leads to some actual problems.

kotlinx-coroutines-core/common/test/CancellableContinuationTest.kt
dkhalanskyjb: Given that the whole test is five lines, there isn't a lot of room for misinterpreting what the test does, especially given that the KDoc describes that we're testing the `EXACTLY_ONCE` contract.

PR #4573, opened 2025-11-14
test passes of isDispacherNeeded

HarshSingh011:
Fixes a bug where withContext(Dispatchers.IO|Default) could run its block on the caller thread if the caller started the coroutine with CoroutineStart.UNDISPATCHED. withContext now checks whether dispatch is required in the fast path and forces a dispatch when necessary, guaranteeing the block executes on the requested dispatcher.

PR #4577, opened 2025-11-25
Force dispatch for UNDISPATCHED coroutines with explicit withContext …

HarshSingh011:
UNDISPATCHED coroutines previously ignored explicit dispatcher changes in withContext(), continuing execution on the current thread instead of switching to the specified dispatcher. This violated expected behavior when users explicitly requested a thread switch via withContext(Dispatchers.IO).

The fix tracks whether a coroutine's initial undispatched execution has completed and forces dispatch only when an explicit dispatcher is provided (not compound contexts like coroutineContext + Job()). This distinction is made by checking if the context itself is a ContinuationInterceptor.

The trade-off is minimal: 8 additional bytes per Job instance for the tracking flag, which is acceptable for correctness. All existing undispatched behavior for non-dispatcher contexts is preserved

### DIFF
--- a/kotlinx-coroutines-core/common/src/Builders.common.kt
+++ b/kotlinx-coroutines-core/common/src/Builders.common.kt
@@ -1,7 +1,7 @@
 @file:JvmMultifileClass
 @file:JvmName("BuildersKt")
 @file:OptIn(ExperimentalContracts::class)
-@file:Suppress("LEAKED_IN_PLACE_LAMBDA", "WRONG_INVOCATION_KIND")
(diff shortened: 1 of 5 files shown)


HarshSingh011:
Hello, Can you please let me now what is the problem in the code it would be great help to me if you can clearly the issue here

fzhinkin:
@HarshSingh011, the answer was given here: https://github.com/Kotlin/kotlinx.coroutines/pull/4573#issuecomment-3540878363

PR #4578, opened 2025-12-01
Clarify the guarantees on the number of IO threads

dkhalanskyjb:


kotlinx-coroutines-core/jvm/src/Dispatchers.kt
```diff
@@ -57,8 +57,10 @@ public actual object Dispatchers {
      * dispatcher typically does not lead to an actual switching to another thread. In such scenarios,
      * the underlying implementation attempts to keep the execution on the same thread on a best-effort basis.
      *
-     * As a result of thread sharing, more than 64 (default parallelism) threads can be created (but not used)
-     * during operations over IO dispatcher.
+     * Whenever we are talking about the number of IO threads in the pool being limited by some number,
+     * that only means that at most that many threads are going to *actually perform* IO work in parallel.
```
murfel: Possibly make it more formal, by getting rid of "we"

"The number of IO threads is at most N" means that at most N threads...

dkhalanskyjb: I liked your suggestion in principle, but didn't think of a way to clearly incorporate it, so I reworded the KDoc to improve accuracy. Since this is an extensive change, please take a look again.

murfel (approved):

PR #4581, opened 2025-12-03
Update the expected API dump according to the Kotlin compiler changes

andrey-mogilev:
The changes in Kotlin compiler were implemented according to https://youtrack.jetbrains.com/issue/KT-82467/Improve-binary-and-JVM-compatibility-by-generating-bridges-for-abstract-interface-methods.
This is the first part of changes, as the option that enables bridges in interfaces is disabled by default yet.


andrey-mogilev:
We need to update the expected ABI output due to the compiler changes, but it seems it is not currently possible due to use of the ABI test with the older compiler in Quality Gate. So, probably we shall postpone the problem and move the compiler changes under a separate `LanguageFeature` with since=2.4

PR #4585, opened 2025-12-05
feat: rewriting cancellation and timeouts page

daniCsorbaJB:
This PR rewrites the entire Cancellation and timeouts page to address user feedback and to align it structurally with the rest of the kotlinlang.org documentation.

It includes revised explanations, new examples, and additional content.
Changes were reviewed by Dmitry Khalanskiy

Related ticket: [KT-69339](https://youtrack.jetbrains.com/issue/KT-69339)
Related summary ticket for the entire coroutine guide rewrite efforts: [KT-66329](https://youtrack.jetbrains.com/issue/KT-66329)

### DIFF
--- a/docs/topics/cancellation-and-timeouts.md
+++ b/docs/topics/cancellation-and-timeouts.md
@@ -1,506 +1,621 @@
-<!--- TEST_NAME CancellationGuideTest -->
 <contribute-url>https://github.com/Kotlin/kotlinx.coroutines/edit/master/docs/topics/</contribute-url>
 
 [//]: # (title: Cancellation and timeouts)
 
-This section covers coroutine cancellation and timeouts.
+Cancellation lets you stop a coroutine before it completes.
+It stops work that's no longer needed, such as when a user closes a window or navigates away in a user interface while a coroutine is still running.
+You can also use it to release resources early and to stop a coroutine from accessing objects past their disposal.
 
-## Cancelling coroutine execution
-
-In a long-running application, you might need fine-grained control on your background coroutines.
-For example, a user might have closed the page that launched a coroutine, and now its result
-is no longer needed and its operation can be cancelled. 
-The [launch] function returns a [Job] that can be used to cancel the running coroutine:
-
-```kotlin
-import kotlinx.coroutines.*
-
-fun main() = runBlocking {
-//sampleStart
-    val job = launch {
-        repeat(1000) { i ->
-            println("job: I'm sleeping $i ...")
-            delay(500L)
-        }
-    }
-    delay(1300L) // delay a bit
-    println("main: I'm tired of waiting!")
-    job.cancel() // cancels the job
-    job.join() // waits for job's completion 
-    println("main: Now I can quit.")
-//sampleEnd    
-}
-```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
-<!--- KNIT example-cancel-01.kt -->
-> You can get the full code [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/test/guide/example-cancel-01.kt).
+> You can use cancellation to stop long-running coroutines that keep producing values even after other coroutines no longer need them, for example, in [pipelines](channels.md#pipelines).
 >
-{style="note"}
+{style="tip"}
 
-It produces the following output:
+Cancellation works through the [`Job`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/) handle, which represents the lifecycle of a coroutine and its parent-child relationships.
+`Job` allows you to check whether the coroutine is active and allows you to cancel it, along with its children, as defined by [structured concurrency](coroutines-basics.md#coroutine-scope-and-structured-concurrency).
 
-```text
-job: I'm sleeping 0 ...
-job: I'm sleeping 1 ...
-job: I'm sleeping 2 ...
-main: I'm tired of waiting!
-main: Now I can quit.
-```
+## Cancel coroutines
 
-<!--- TEST -->
+A coroutine is canceled when the [`cancel()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/cancel.html) function is invoked on its `Job` handle.
+This handle is returned by [coroutine builder functions](coroutines-basics.md#coroutine-builder-functions) like
+the [`.launch()`](coroutines-basics.md#coroutinescope-launch) function, which returns a `Job`, or the [`.async()`](coroutines-basics.md#coroutinescope-async)
+function, which returns a [`Deferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/).
+`Deferred` is a type that implements `Job` and supports the same cancellation behavior.
 
-As soon as main invokes `job.cancel`, we don't see any output from the other coroutine because it was cancelled. 
-There is also a [Job] extension function [cancelAndJoin] 
-that combines [cancel][Job.cancel] and [join][Job.join] invocations.
+You can call the `cancel()` function manually, or it can be invoked automatically through cancellation propagation when a parent coroutine is canceled.
 
-## Cancellation is cooperative
+When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) at [suspension points](#suspension-points-and-cancellation).
 
-Coroutine cancellation is _cooperative_. A coroutine code has to cooperate to be cancellable.
-All the suspending functions in `kotlinx.coroutines` are _cancellable_. They check for cancellation of 
-coroutine and throw [CancellationException] when cancelled. However, if a coroutine is working in 
-a computation and does not check for cancellation, then it cannot be cancelled, like the following 
-example shows:
-
-```kotlin
-import kotlinx.coroutines.*
-
-fun main() = runBlocking {
-//sampleStart
-    val startTime = System.currentTimeMillis()
-    val job = launch(Dispatchers.Default) {
-        var nextPrintTime = startTime
-        var i = 0
-        while (i < 5) { // computation loop, just wastes CPU
-            // print a message twice a second
-            if (System.currentTimeMillis() >= nextPrintTime) {
-                println("job: I'm sleeping ${i++} ...")
-                nextPrintTime += 500L
-            }
-        }
-    }
-    delay(1300L) // delay a bit
-    println("main: I'm tired of waiting!")
-    job.cancelAndJoin() // cancels the job and waits for its completion
-    println("main: Now I can quit.")
-//sampleEnd    
-}
-```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
-<!--- KNIT example-cancel-02.kt -->
-> You can get the full code [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/test/guide/example-cancel-02.kt).
+> You can use the [`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function to suspend a coroutine until cancellation.
 >
-{style="note"}
+{style="tip"}
 
-Run it to see that it continues to print "I'm sleeping" even after cancellation
-until the job completes by itself after five iterations.
-
-<!--- TEST 
-job: I'm sleeping 0 ...
-job: I'm sleeping 1 ...
-job: I'm sleeping 2 ...
-main: I'm tired of waiting!
-job: I'm sleeping 3 ...
-job: I'm sleeping 4 ...
-main: Now I can quit.
--->
-
-The same problem can be observed by catching a [CancellationException] and not rethrowing it:
+Here's an example that demonstrates how to manually cancel coroutines:
 
 ```kotlin
 import kotlinx.coroutines.*
+import kotlin.time.Duration
 
-fun main() = runBlocking {
 //sampleStart
-    val job = launch(Dispatchers.Default) {
-        repeat(5) { i ->
+suspend fun main() {
+    withContext(Dispatchers.Default) {
+        // Used as a signal that the coroutine has started running
+        val job1Started = CompletableDeferred<Unit>()
+        
+        val job1: Job = launch {
+            
+            println("The coroutine has started")
+
+            // Marks the point where the coroutine has started running
+            job1Started.complete(Unit)
             try {
-                // print a message twice a second
-                println("job: I'm sleeping $i ...")
-                delay(500)
-            } catch (e: Exception) {
-                // log the exception
-                println(e)
+                // Suspends indefinitely
+                // Without cancellation, this call would never return
+                delay(Duration.INFINITE)
+            } catch (e: CancellationException) {
+                println("The coroutine was canceled: $e")
+              
+                // Always rethrow cancellation exceptions!
+                throw e
+            }
+            println("This line will never be executed")
+        }
+      
+        // Waits for job1 to start before canceling it
+        job1Started.await()
+
+        // Cancels the coroutine, so delay() throws a CancellationException
+        job1.cancel()
+
+        // async returns a Deferred handle, which inherits from Job
+        val job2 = async {
+                  // This line may not be printed if the coroutine is canceled before its body starts executing
+                  println("The async coroutine has started")
+
+            try {
+                // Equivalent to delay(Duration.INFINITE)
+                // Suspends until this coroutine is canceled
+                awaitCancellation()
+
+            } catch (e: CancellationException) {
+                println("The async coroutine was canceled")
+                throw e
             }
         }
+        job2.cancel()
     }
-    delay(1300L) // delay a bit
-    println("main: I'm tired of waiting!")
-    job.cancelAndJoin() // cancels the job and waits for its completion
-    println("main: Now I can quit.")
-//sampleEnd    
+    // Coroutine builders such as withContext() or coroutineScope()
+    // wait for all child coroutines to complete, even when they are canceled
+    println("All cancellations have completed")
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="manual-cancellation-example"}
+
+In this example, [`CompletableDeferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/) is used as a signal that the coroutine has started running.
+The coroutine calls `complete()` when it reaches the start point, and `await()` only returns once that `Deferred` is completed. This way, cancellation happens only after the coroutine has started running.
+The coroutine created by `.async()` doesn't have this check, so it may be canceled before it can run the code inside its block.
+
+> Catching `CancellationException` can break the cancellation propagation.
+> If you must catch it, rethrow it to let the cancellation propagate correctly through the coroutine hierarchy.
+>
+> For more information, see [Coroutine exceptions handling](exception-handling.md#cancellation-and-exceptions).
+>
+{style="warning"}
+
+### Cancellation propagation
+
+[Structured concurrency](coroutines-basics.md#coroutine-scope-and-structured-concurrency) ensures that canceling a coroutine also cancels all of its children.
+This prevents child coroutines from continuing work after the parent has already stopped.
+
+Here's an example:
+
+```kotlin
+import kotlinx.coroutines.*
+import kotlin.time.Duration
+
+//sampleStart
+suspend fun main() {
+    withContext(Dispatchers.Default) {
+        // Used as a signal that the child coroutines have started running
+        val childrenLaunched = CompletableDeferred<Unit>()
+
+        // Starts two child coroutines
+        val parentJob = launch {
+            launch {
+                try {
+                    awaitCancellation()
+                } finally {
+                    println("Child coroutine 1 has been canceled")
+                }
+            }
+            launch {
+                try {
+                    awaitCancellation()
+                } finally {
+                    println("Child coroutine 2 has been canceled")
+                }
+            }
+            // Marks that the child coroutines have started running
+            childrenLaunched.complete(Unit)
+        }
+        // Waits for the parent coroutine to signal that it has started
+        // all of its children
+        childrenLaunched.await()
+
+        // Cancels the parent coroutine, which cancels all its children
+        parentJob.cancel()
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true" id="cancellation-propagation-example"}
+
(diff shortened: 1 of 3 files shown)


docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+>
+{style="tip"}
+
+Cancellation works through the [`Job`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/) handle, which represents the lifecycle of a coroutine and its parent-child relationships.
+`Job` allows you to check whether the coroutine is active and allows you to cancel it, along with its children, as defined by [structured concurrency](coroutines-basics.md#coroutine-scope-and-structured-concurrency).
+
+## Cancel coroutines
+
+A coroutine is canceled when the [`cancel()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/cancel.html) function is invoked on its `Job` handle.
+This handle is returned by [coroutine builder functions](coroutines-basics.md#coroutine-builder-functions) like
```
danil-pavlov: I'm a bit confused here, it reads like the return of Job is repeated twice:

>  ... on its `Job` handle. This handle (which is Job?) is returned by coroutine builder functions like the `.launch()` function, which returns a `Job`

danil-pavlov: Also, it's possible to use active voice here, "coroutine builder functions return"

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-main: Now I can quit.
--->
+In this section, you can learn ways to create cancelable coroutines.
 
-The same problem can be observed by catching a [CancellationException] and not rethrowing it:
+### Suspension points and cancellation
+
+When a coroutine is canceled, it continues running until it reaches a point in the code where it may suspend, also known as a _suspension point_.
+If the coroutine suspends there, the suspending function checks whether it has been canceled.
+If it has been canceled, the coroutine stops and throws `CancellationException`.
```
danil-pavlov: Repetition of "it has been canceled", maybe "if it has/if this is the case"

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+    println("All child jobs completed!")
 }
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
-<!--- KNIT example-cancel-03.kt -->
-> You can get the full code [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/test/guide/example-cancel-03.kt).
+{kotlin-runnable="true" id="suspension-points-example"}
+
+> All suspending functions in the `kotlinx.coroutines` library cooperate with cancellation because they use [`suspendCancellableCoroutine()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html) internally, which checks for cancellation when the coroutine suspends.
+> On the other hand, custom suspending functions that use [`suspendCoroutine()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.coroutines/suspend-coroutine.html) don't react to cancellation.
```
danil-pavlov:
```suggestion
> In contrast, custom suspending functions that use [`suspendCoroutine()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.coroutines/suspend-coroutine.html) don't react to cancellation.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-{style="note"}
+{style="tip"}
 
-While catching `Exception` is an anti-pattern, this issue may surface in more subtle ways, like when using the
-[`runCatching`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html) function,
-which does not rethrow [CancellationException].
+### Check for cancellation explicitly
 
-## Making computation code cancellable
+If a coroutine doesn't [suspend](#suspension-points-and-cancellation) for a long time, it doesn't stop when it's canceled unless it explicitly checks for cancellation.
```
danil-pavlov, line 234: Does the coroutine check for cancellation itself?
maybe "... unless there is an explicit check for cancellation"/"unless explicitly checked for cancellation"

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-There are two approaches to making computation code cancellable.
-The first one is periodically invoking a suspending function that checks for cancellation.
-There are the [yield] and [ensureActive]
-functions, which are great choices for that purpose.
-The other one is explicitly checking the cancellation status using [isActive].
-Let us try the latter approach.
+To check for cancellation, use the following APIs:
 
-Replace `while (i < 5)` in the previous example with `while (isActive)` and rerun it. 
+* [`isActive`](#isactive): this property is `false` when the coroutine is canceled.
```
danil-pavlov: I think we can avoid colon in this list:

To check for cancellation, use the following APIs:
* isActive property that is false when the coroutine is canceled.
* ensureActive() function that throws..

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+* [`ensureActive()`](#ensureactive): this function throws `CancellationException` immediately if the coroutine is canceled.
+* [`yield()`](#yield): this function suspends the coroutine, releasing the thread and giving other coroutines a chance to run on it. Suspending the coroutine lets it check for cancellation and throw `CancellationException` if it's canceled.
```
danil-pavlov: I think this description can be shorter since the subsections with details are right below and to avoid repetition

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-## Asynchronous timeout and resources
+### Run non-cancelable blocks
 
-<!-- 
-  NOTE: Don't change this section name. It is being referenced to from within KDoc of withTimeout functions.
--->
+You can prevent cancellation from affecting certain parts of a coroutine.
+To do so, pass [`NonCancellable`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-non-cancellable/) as an argument to the `withContext()` coroutine builder function.
+
+> Avoid using `NonCancellable` with other coroutine builders like `.launch()` or `.async()`. Doing so breaks structured concurrency by severing the parent-child relationship.
```
danil-pavlov:
```suggestion
> Avoid using `NonCancellable` with other coroutine builders like `.launch()` or `.async()`. Doing so breaks structured concurrency by disrupting the parent-child relationship.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+
+suspend fun main() {
+    withContext(Dispatchers.Default) {
+        val childJob = launch {
+            startService()
+            try {
+                awaitCancellation()
+            } finally {
+                withContext(NonCancellable) {
+
```
danil-pavlov:
```suggestion
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-    withTimeout(1300L) {
-        repeat(1000) { i ->
-            println("I'm sleeping $i ...")
-            delay(500L)
+suspend fun main() {
+    withContext(Dispatchers.Default) {
+        val childStarted = CompletableDeferred<Unit>()
+        val childJob = launch {
+            try {
+                // Cancellation triggers a thread interrupt
```
danil-pavlov:
```suggestion
                // Cancellation triggers a thread interruption
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+                    }
+                }
+            } catch (e: CancellationException) {
+                println("Coroutine canceled (Kotlin): $e")
+                throw e
+            }
         }
+        childStarted.await()
+
+        // Cancels the coroutine and interrupts the thread running Thread.sleep()
```
danil-pavlov:
```suggestion
        // Cancels the coroutine and interrupts the thread by running Thread.sleep()
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+        scope.launch {
+            val contents = withContext(Dispatchers.IO) {
+                Files.newBufferedReader(
+                    path, Charset.forName("US-ASCII")
+                ).use {
+                    it.readLines()
+                }
+            }
+            // It's safe to call updateUi here,
+            // because cancellation would have prevented withContext() from returning any values
```
danil-pavlov:
```suggestion
            // In case of cancellation, withContext() wouldn't return any values
```

danil-pavlov: to avoid "would have" construction, plus the comment was getting cut

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
 
-## Run non-cancellable block
-
-Any attempt to use a suspending function in the `finally` block of the previous example causes
-[CancellationException], because the coroutine running this code is cancelled. Usually, this is not a 
-problem, since all well-behaved closing operations (closing a file, cancelling a job, or closing any kind of 
-communication channel) are usually non-blocking and do not involve any suspending functions. However, in the 
-rare case when you need to suspend in a cancelled coroutine you can wrap the corresponding code in
-`withContext(NonCancellable) {...}` using [withContext] function and [NonCancellable] context as the following example shows:
+Use `yield` to allow other coroutines to run on the same thread or thread pool before one coroutine finishes:
```
danil-pavlov:
```suggestion
Use `yield` to allow other coroutines to run on the same thread or thread pool before one of them finishes:
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+            // Repeatedly sorts the list while the coroutine remains active
+            while (isActive) {
+                unsortedList.sort()
+                ++i
             }
+            println(
+                "Stopped sorting the list after $i iterations"
+            )
         }
+        // Sorts the list for 100 milliseconds, then consider it sorted enough
```
danil-pavlov:
```suggestion
        // Sorts the list for 100 milliseconds, then considers it sorted enough
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+            },
+            launch {
+                val mutex = Mutex(locked = true)
+                // Suspends while waiting for a mutex that remains locked indefinitely
+                mutex.lock()
             }
-        }
+        )
+        
+        // Gives the child coroutines time to start and suspend
```
danil-pavlov, line 216:
```suggestion
        // Gives child coroutines time to start and suspend
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-<!--- TEST 
-job: I'm sleeping 0 ...
-job: I'm sleeping 1 ...
-job: I'm sleeping 2 ...
-main: I'm tired of waiting!
-job: I'm sleeping 3 ...
-job: I'm sleeping 4 ...
-main: Now I can quit.
--->
+In this section, you can learn ways to create cancelable coroutines.
```
danil-pavlov:
```suggestion
In this section, you can learn how to create cancelable coroutines.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+{style="warning"}
 
-As soon as main invokes `job.cancel`, we don't see any output from the other coroutine because it was cancelled. 
-There is also a [Job] extension function [cancelAndJoin] 
-that combines [cancel][Job.cancel] and [join][Job.join] invocations.
+### Cancellation propagation
 
-## Cancellation is cooperative
+[Structured concurrency](coroutines-basics.md#coroutine-scope-and-structured-concurrency) ensures that canceling a coroutine also cancels all of its children.
+This prevents child coroutines from continuing work after the parent has already stopped.
```
danil-pavlov:
```suggestion
This prevents child coroutines from working after the parent has already stopped.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
-        var nextPrintTime = startTime
-        var i = 0
-        while (i < 5) { // computation loop, just wastes CPU
-            // print a message twice a second
-            if (System.currentTimeMillis() >= nextPrintTime) {
-                println("job: I'm sleeping ${i++} ...")
-                nextPrintTime += 500L
+suspend fun main() {
+    withContext(Dispatchers.Default) {
+        // Used as a signal that the child coroutines have been launched
```
danil-pavlov, line 123:
```suggestion
        // Used as a signal that child coroutines have been launched
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+            launch {
+                println("Child coroutine 2 has started running")
+                try {
+                    awaitCancellation()
+                } finally {
+                    println("Child coroutine 2 has been canceled")
+                }
+            }
+            // Completes the CompletableDeferred,
+            // signaling that the child coroutines have been launched
```
danil-pavlov, line 145:
```suggestion
            // signaling that child coroutines have been launched
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
 ```
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
-<!--- KNIT example-cancel-02.kt -->
-> You can get the full code [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/test/guide/example-cancel-02.kt).
->
-{style="note"}
+{kotlin-runnable="true" id="cancellation-propagation-example"}
+
+In this example, each child coroutine uses a [`finally` block](exceptions.md#the-finally-block), so the code inside it runs when the coroutine is canceled.
+Here, `CompletableDeferred` signals that the child coroutines are launched before they are canceled, but it doesn't guarantee that they start running. If they are canceled first, nothing is printed.
```
danil-pavlov, line 161:
```suggestion
Here, `CompletableDeferred` signals that child coroutines are launched before they are canceled, but it doesn't guarantee that they start running. If they are canceled first, nothing is printed.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+You can call the `cancel()` function manually, or it can be invoked automatically through cancellation propagation when a parent coroutine is canceled.
 
-In a long-running application, you might need fine-grained control on your background coroutines.
-For example, a user might have closed the page that launched a coroutine, and now its result
-is no longer needed and its operation can be cancelled. 
-The [launch] function returns a [Job] that can be used to cancel the running coroutine:
+When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) the next time it checks for cancellation.
+For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
+
+> You can use the [`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function to suspend a coroutine until cancellation.
```
danil-pavlov:
```suggestion
> You can use the [`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function to suspend a coroutine until it's canceled.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+>
+{style="tip"}
+
+Here's an example that demonstrates how to manually cancel coroutines:
```
danil-pavlov:
```suggestion
Here's an example on how to manually cancel coroutines:
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
...
+      
+        // Waits for job1 to start before canceling it
+        job1Started.await()
+
+        // Cancels the coroutine, so delay() throws a CancellationException
+        job1.cancel()
+
+        // async returns a Deferred handle, which inherits from Job
+        val job2 = async {
+                  // This line may not be printed if the coroutine is canceled before its body starts executing
```
danil-pavlov: This comment is tool long, maybe break into two lines like the rest

docs/topics/cancellation-and-timeouts.md
```diff
@@ -1,506 +1,628 @@
-<!--- TEST_NAME CancellationGuideTest -->
 <contribute-url>https://github.com/Kotlin/kotlinx.coroutines/edit/master/docs/topics/</contribute-url>
 
 [//]: # (title: Cancellation and timeouts)
```
danil-pavlov, line 3: Timeouts play such a small role on this page, I wonder if we should rename to just "Canceling coroutines"

danil-pavlov (changes requested):

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: 🤔 Interesting thought and worth considering — imho it's fine to keep it in. 
While not as prevalent on the page, timeouts are still one of the most common cancellation use cases, so it'd be nice if people can find it in the title and make the mental connection between the two.  

On the other hand, I'd see value in removing it as well. 
Although I'd prefer to simply call it "Cancellation/Coroutine cancellation" if we decide to change it, so the title already conveys that it's something that can happen through propagation as well (not initiated by the user).

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: ah 🤔 — I wanted to achieve two things with this sentence: 
- To let the reader know where the `Job` handle comes from 
- To let the reader know that Deferred is (while not in the strictest sense a `Job`) handle it acts the same way in relation to cancellation because it inherits from it

> Also, it's possible to use active voice here, "coroutine builder functions return"

I think so: what do you think about this version (slightly refocused)

> A coroutine is canceled when the [`cancel()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/cancel.html) function is invoked on its `Job` handle.
> [Coroutine builder functions](coroutines-basics.md#coroutine-builder-functions) such as
[`.launch()`](coroutines-basics.md#coroutinescope-launch) return a `Job`. The [`.async()`](coroutines-basics.md#coroutinescope-async)
function returns a [`Deferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/), which
implements `Job` and supports the same cancellation behavior.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: Changed it to : "If it has"

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: > Does the coroutine check for cancellation itself?

Yes, it's the code in the coroutine (so in turn it's the coroutine that cooperates by checking for the cancellation) better highlighted by this part: 

> This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: nice idea 👍 — restructured it like you suggested

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: I see what you mean 🤔  — It was much shorter, but it was suggested to better highlight the difference between it and ensureActive() in a place where the reader can see both.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: 🤔 I think "severing" more accurately describes what happens here. Disrupt can mean many things; it could mean a temporary disturbance or a slight relationship change. What happens here is that the tree hierarchy is severed/cut between parent and child.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: great suggestion 👍 thank you — implemented it

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: I think "the" helps here to signal that we mean the specific child coroutines in the example

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: same here — I think it helps slightly. Tbh, could work either way, but I think this helps with the user understanding that this is not necessarily a generic solution for child coroutines, just specific to this example.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: I think it's better with the:

“the child coroutines” refers to specific coroutines created just above in the code.
Without “the,” to me it sounds more general, as if it applies to any coroutines, not the ones in this example.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: same here — I think it helps slightly. Tbh, could work either way, but I think this helps with the user understanding that this is not necessarily a generic solution for child coroutines, just specific to this example.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: ty 💯 broke it into two

docs/topics/cancellation-and-timeouts.md
danil-pavlov: I think it doesn't change the meaning, especially with plurals, and in it's better to keep comments concise
but it doesn't mattet that much

docs/topics/cancellation-and-timeouts.md
danil-pavlov: I think this word is out of our doc style, you can use "break" here instead of  "severing", but change the first "breaks" in this sentence to something else, like "breaches, disrupts"

docs/topics/cancellation-and-timeouts.md
danil-pavlov: I think it's clear from the context, the comment is in the middle of the example

danil-pavlov (approved):

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: 👍  break can work — rewrote it like : 

> Avoid using `NonCancellable` with other coroutine builders like `.launch()` or `.async()`. Doing so disrupts structured concurrency by breaking the parent-child relationship.

PR #4588, opened 2025-12-11
Ensure the coroutines debug agent attaches even with no stdlib

dkhalanskyjb:
Alternative for #4565

With this change, we ensure that `AgentPremain` does not touch any of the `kotlin.*` APIs when it's being attached to a process. This means that even if it's attached by a mistake to a pure Java process, it won't cause any harm.

Most of the changes are limited to removing the `runCatching` antipattern, but one sad consequence is the inability to use the convenient `thread { }` function.

An integration test was implemented, checking that a pure Java process completes successfully even with the debug agent attached to it.
The earlier version of the test performed the check even before the Java class got compiled. It was fine, given that the execution never even got to the main class and crashed during the debug agent attach. With the proposed change, however, it had to be significantly modified.

### DIFF
--- a/integration-testing/safeDebugAgentTest/build.gradle.kts
+++ b/integration-testing/safeDebugAgentTest/build.gradle.kts
@@ -20,22 +20,23 @@ tasks.register("attachAgentWithoutKotlinStdlib") {
 
     doLast {
         val agentJar = System.getProperty("coroutines.debug.agent.path")
-        val errorOutputStream = ByteArrayOutputStream()
-        val standardOutputStream = ByteArrayOutputStream()
-
-        project.javaexec {
+        val execResult = project.providers.javaexec {
             mainClass.set("Main")
             classpath = sourceSets.main.get().runtimeClasspath
             jvmArgs = listOf("-javaagent:$agentJar")
-            errorOutput = errorOutputStream
-            standardOutput = standardOutputStream
+            isIgnoreExitValue = true
         }
-
-        check(errorOutputStream.toString().isEmpty()) {
-            "Expected no output in the error stream, but got:\n$errorOutputStream"
+        val exitCode = execResult.result.get().exitValue
+        val stdout = execResult.standardOutput.asText.getOrElse("<not found>")
+        val stderr = execResult.standardError.asText.getOrElse("<not found>")
+        check (exitCode == 0) {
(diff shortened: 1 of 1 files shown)


mvicsokolova:
Thank you! I've tested this change on the debugger side

integration-testing/safeDebugAgentTest/build.gradle.kts
```diff
@@ -15,29 +15,27 @@ application {
...
-    val errorOutput = errorOutputStream.toString()
-    val standardOutput = standardOutputStream.toString()
-    if (!errorOutput.contains(expectedAgentError)) {
-        throw GradleException("':safeDebugAgentTest:runWithExpectedFailure' completed with an unexpected output:\n" + standardOutput + "\n" + errorOutput)
+        check(errorOutputStream.toString().isEmpty()) {
+            "Expected no output in the error stream, but got:\n$errorOutputStream"
+        }
+        check(standardOutputStream.toString().contains("OK!")) {
+            "Expected 'OK!' in the standard output, but got:\n$standardOutputStream"
+        }
```
fzhinkin, line 40:
```suggestion
        val agentJar = System.getProperty("coroutines.debug.agent.path")

        val execResult = project.providers.javaexec {
            mainClass.set("Main")
            classpath = sourceSets.main.get().runtimeClasspath
            jvmArgs = listOf("-javaagent:$agentJar")
            isIgnoreExitValue = true
        }

        val exitCode = execResult.result.get().exitValue
        val stdout = execResult.standardOutput.asText.getOrElse("<not found>")
        val stderr = execResult.standardError.asText.getOrElse("<not found>")

        check (exitCode == 0) {
            "Process execution ended with non-zero exit code: $exitCode\nstdout:\n$stdout\nstderr:\n$stderr"
        }
        check(stderr.isEmpty()) {
            "Expected no output in the error stream, but got:\n$stderr"
        }
        check(stdout.contains("OK!")) {
            "Expected 'OK!' in the standard output, but got:\n$stdout"
        }
```

There were a few issues:
- `Project.javaexec` is deprecated and it's better to use a supported alternative
- With original javaexec settings, any non-zero exit code would result in an exception with "non-zero exit code" as a result, but without any details and without stdout/err contents.

dkhalanskyjb: Applied the suggestion. Github's "Commit suggestion" refused to do that, saying that it can't apply suggestions to removed lines.

Issue #4583, opened 2025-12-04
CoroutineScope cancellation affects MutableSharedFlow in unexpected way


antonvkrutko:
Possible issue with consuming the data from `MutableSharedFlow`. 

Reproduction conditions:
Cancelling the scope in which data is produced and collected and then launching emission and consuming on the new scope, but on the same global instance of `MutableSharedFlow`. After one or several restarts emitted items are no longer delivered to the collector. 

Below is the minimal example to reproduce the issue. However, reproduction of this issue is greatly affected by the work load performed (or time spent) in the collector.

Coroutines versions the issue is tried to be reproduced on: `1.10.2`, `1.8.0`
Environment: JVM, Android. 

Note: correctness of counter values themselves is not important in that example. 

```
import kotlinx.coroutines.*
import kotlinx.coroutines.flow.MutableSharedFlow
import kotlinx.coroutines.flow.launchIn
import kotlinx.coroutines.flow.onEach

var counter = 0L
val mutableSharedFlow = MutableSharedFlow<Long>()

var cs: CoroutineScope? = null
fun start() {
    cs?.cancel()
    cs = CoroutineScope(Dispatchers.IO + SupervisorJob())

    mutableSharedFlow.onEach {
        delay(50)
        println(it)
    }.launchIn(cs!!)

    cs!!.launch {
        while (currentCoroutineContext().isActive) {
            mutableSharedFlow.emit(counter)
            counter++
        }
    }
}

suspend fun main() {
    while (true) {
        start()
        println("restart, counter = $counter")

        delay(3000)
    }
}
```

Output below:

```
....
47
48
49
50
51
52
53
54
restart, counter = 56
restart, counter = 56
restart, counter = 56
restart, counter = 56
restart, counter = 56
... // proper behavior (emissions happening) is never restored after this point
```

dkhalanskyjb:
Thank you for the nice reproducer! I also observe the problematic behavior. Yes, looks like a bug, we'll need to investigate.

dkhalanskyjb:
Minified, single-threaded test illustrating the behavior:
```kotlin
private class SharedFlowResubscriptionStressTest {

    fun testResubscription() = runBlocking {
        val mutableSharedFlow = MutableSharedFlow<Unit>()
        var iteration = 0
        while (true) {
            println("iteration ${++iteration}")
            launch {
                val valueObtained = CompletableDeferred<Unit>()
                val j1 = launch {
                    println("ready to collect")
[...]

PR #4591, opened 2025-12-12
Update internal SharedFlow comments

dkhalanskyjb:
Fixes https://github.com/Kotlin/kotlinx.coroutines/issues/4583

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -542,8 +542,7 @@ internal open class SharedFlowImpl<T>(
         // take into account a special case of sync shared flow that can go past 1st queued emitter
         if (bufferCapacity == 0 && queueSize > 0) newMinCollectorIndex++
         forEachSlotLocked { slot ->
-            @Suppress("ConvertTwoComparisonsToRangeCheck") // Bug in JS backend
-            if (slot.index >= 0 && slot.index < newMinCollectorIndex) newMinCollectorIndex = slot.index
+            if (slot.index in 0..<newMinCollectorIndex) newMinCollectorIndex = slot.index
         }
         assert { newMinCollectorIndex >= minCollectorIndex } // can only grow
         if (newMinCollectorIndex <= minCollectorIndex) return EMPTY_RESUMES // nothing changes
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -613,10 +613,6 @@ internal open class SharedFlowImpl<T>(
          * The value may be larger than [bufferCapacity] if [nCollectors] was 0,
          * but the `minOf(replay)` will clamp this value to a valid one, since `replay <= bufferCapacity`. */
         val newBufferSize1 = (newBufferEndIndex - head).toInt()
-        /** If [nCollectors] was 0, there are no active collector slots,
-         * and the invariant dictates that [minCollectorIndex] should be [bufferEndIndex] (`replayIndex + bufferSize`).
-         * Ensure this directly. */
-        if (nCollectors == 0) newMinCollectorIndex = newBufferEndIndex
         // Compute new replay size -> limit to replay the number of items we need, take into account that it can only grow
         var newReplayIndex = maxOf(replayIndex, newBufferEndIndex - minOf(replay, newBufferSize1))
         // adjustment for synchronous case with cancelled emitter (NO_VALUE)
@@ -624,6 +620,10 @@ internal open class SharedFlowImpl<T>(
             newBufferEndIndex++
             newReplayIndex++
         }
+        /** If [nCollectors] was 0, there are no active collector slots,
+         * and the invariant dictates that [minCollectorIndex] should be [bufferEndIndex] (`replayIndex + bufferSize`).
+         * Ensure this directly. */
+        if (nCollectors == 0) newMinCollectorIndex = newBufferEndIndex
         // Update buffer state
         updateBufferLocked(newReplayIndex, newMinCollectorIndex, newBufferEndIndex, newQueueEndIndex)
         // just in case we've moved all buffered emitters and have NO_VALUE's at the tail now
--- a/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
@@ -58,7 +58,7 @@ import kotlin.jvm.*
  * get values from the buffer without suspending emitters. The buffer space determines how much slow subscribers
  * can lag from the fast ones. When creating a shared flow, additional buffer capacity beyond replay can be reserved
  * using the `extraBufferCapacity` parameter.
- * 
+ *
  * A shared flow with a buffer can be configured to avoid suspension of emitters on buffer overflow using
  * the `onBufferOverflow` parameter, which is equal to one of the entries of the [BufferOverflow] enum. When a strategy other
  * than [SUSPENDED][BufferOverflow.SUSPEND] is configured, emissions to the shared flow never suspend.
@@ -292,11 +292,23 @@ public fun <T> MutableSharedFlow(
 // ------------------------------------ Implementation ------------------------------------
 
 internal class SharedFlowSlot : AbstractSharedFlowSlot<SharedFlowImpl<*>>() {
+    /**
+     * The first index in the buffer that was not yet consumed by this collector,
+     * or -1 if the slot is not allocated to a collector.
+     */
     @JvmField
-    var index = -1L // current "to-be-emitted" index, -1 means the slot is free now
+    var index = -1L
 
+    /**
+     * The current continuation of the collector.
+     *
+     * This field is used by the collector each time when it can not acquire the next value and suspends while
+     * waiting for the next one.
+     * `null` means that either the collector is not suspended but is processing some value,
+     * or that this slot is not allocated to a collector.
+     */
     @JvmField
-    var cont: Continuation<Unit>? = null // collector waiting for new value
+    var cont: Continuation<Unit>? = null
 
     override fun allocateLocked(flow: SharedFlowImpl<*>): Boolean {
         if (index >= 0) return false // not free
@@ -318,7 +330,7 @@ internal open class SharedFlowImpl<T>(
     private val replay: Int,
     private val bufferCapacity: Int,
     private val onBufferOverflow: BufferOverflow
-) : AbstractSharedFlow<SharedFlowSlot>(), MutableSharedFlow<T>, CancellableFlow<T>, FusibleFlow<T> {
+) : AbstractSharedFlow<SharedFlowImpl<*>, SharedFlowSlot>(), MutableSharedFlow<T>, CancellableFlow<T>, FusibleFlow<T> {
     /*
         Logical structure of the buffer
 
@@ -350,7 +362,7 @@ internal open class SharedFlowImpl<T>(
 
     // Stored state
     private var buffer: Array<Any?>? = null // allocated when needed, allocated size always power of two
-    private var replayIndex = 0L // minimal index from which new collector gets values
+    private var replayIndex = 0L // minimal index from which a new collector gets values
     private var minCollectorIndex = 0L // minimal index of active collectors, equal to replayIndex if there are none
     private var bufferSize = 0 // number of buffered values
     private var queueSize = 0 // number of queued emitters
@@ -532,13 +544,27 @@ internal open class SharedFlowImpl<T>(
         return index
     }
 
-    // Is called when a collector disappears or changes index, returns a list of continuations to resume after lock
+    /**
+     * Handle a collector having receiving the next value from the buffer or disappearing.
+     *
+     * As a result of both, the *index* of the collector--that is, the index of the next value in the buffer this
+     * collector intends to consume--changes, so the internal invariants need updating.
+     *
+     * In addition, it may turn out that either
+     * - This collector was the slowest one, and now that it has processed a value in the buffer,
+     *   the buffer has one more free slot to accommodate more values, so new emitters can be resumed.
+     * - This collector was the last one, and now that it's gone, emitters don't need to distribute values among
+     *   collectors anymore and can all continue executing immediately.
+     *
+     * In both cases, an array of continuations of the correct number of waiting emitters is returned
+     * so that they can be resumed.
+     */
     internal fun updateCollectorIndexLocked(oldIndex: Long): Array<Continuation<Unit>?> {
         assert { oldIndex >= minCollectorIndex }
         if (oldIndex > minCollectorIndex) return EMPTY_RESUMES // nothing changes, it was not min
-        // start computing new minimal index of active collectors
+        // start computing the new minimal index of active collectors
         val head = head
-        var newMinCollectorIndex = head + bufferSize
+        var newMinCollectorIndex = head + bufferSize // = this.bufferEndIndex
         // take into account a special case of sync shared flow that can go past 1st queued emitter
         if (bufferCapacity == 0 && queueSize > 0) newMinCollectorIndex++
         forEachSlotLocked { slot ->
@@ -546,21 +572,27 @@ internal open class SharedFlowImpl<T>(
         }
         assert { newMinCollectorIndex >= minCollectorIndex } // can only grow
         if (newMinCollectorIndex <= minCollectorIndex) return EMPTY_RESUMES // nothing changes
-        // Compute new buffer size if we drop items we no longer need and no emitter is resumed:
-        // We must keep all the items from newMinIndex to the end of buffer
-        var newBufferEndIndex = bufferEndIndex // var to grow when waiters are resumed
+        /* The new minimal resumer index has increased, so new emitters can be resumed,
+         * with their values now moving to the buffer. */
+        // Initially, `bufferEndIndex`, but later, grown for every resumed emitter.
+        var newBufferEndIndex = bufferEndIndex
+        // Represents how many emitters we can awaken at most.
         val maxResumeCount = if (nCollectors > 0) {
-            // If we have collectors we can resume up to maxResumeCount waiting emitters
-            // a) queueSize -> that's how many waiting emitters we have
-            // b) bufferCapacity - newBufferSize0 -> that's how many we can afford to resume to add w/o exceeding bufferCapacity
+            // Represents how many elements would be in the buffer if we didn't resume any emitters.
             val newBufferSize0 = (newBufferEndIndex - newMinCollectorIndex).toInt()
+            /** a) We can't awaken more than [queueSize] emitters, as only that many have registered.
+                b) If we awaken more than `bufferCapacity - newBufferSize0` emitters, we'll exceed [bufferCapacity]. */
             minOf(queueSize, bufferCapacity - newBufferSize0)
         } else {
-            // If we don't have collectors anymore we must resume all waiting emitters
-            queueSize // that's how many waiting emitters we have (at most)
+            // If we don't have collectors anymore, we must resume all waiting emitters and drop the buffer.
+            /** [queueSize] is the number of emitters that registered (there may, in fact, be fewer than that). */
+            queueSize
         }
+        /* Walk through at most `maxResumeCount` emitters, moving their values to the buffer,
+        * zeroing out the cells the emitters were occupying, and storing their continuations in an array that
+        * will be returned from this function and processed once the lock gets released. */
         var resumes: Array<Continuation<Unit>?> = EMPTY_RESUMES
-        val newQueueEndIndex = newBufferEndIndex + queueSize
+        val newQueueEndIndex = newBufferEndIndex + queueSize // = this.queueEndIndex
         if (maxResumeCount > 0) { // collect emitters to resume if we have them
             resumes = arrayOfNulls(maxResumeCount)
             var resumeCount = 0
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -10,19 +10,66 @@ import kotlin.jvm.*
...
+ * A common data structure for `StateFlow` and `SharedFlow`.
+ */
+internal abstract class AbstractSharedFlow<This, S : AbstractSharedFlowSlot<This>> : SynchronizedObject() {
+    /**
+     * Array of slots for collectors of the shared flow.
+     *
+     * `null` by default, created on demand.
+     * Each cell is also `null` by default, and the specific slot object is [created][createSlot] on demand.
+     * The whole array being `null` or a cell being `null` is equivalent to the cell not being
+     * [*allocated*][AbstractSharedFlowSlot.allocateLocked]--not to be confused with memory allocation, this means
```
murfel: Allocate/free slot was misleading naming, but at least it was consistent.

The current snapshot of all comments uses: register, inhabit, allocate to mean allocate. 

Consider picking one term to describe comments. Possibly even replace it in code as well, so there's no need for a lengthy explanation.

Here's some options:
inhabit / free
register / deregister (a bit tricky because you can both register a collector and a slot)
inhabited slot / registered slot / occupied slot / assigned slot

(Another reason for this is because "allocate" used in its true sense at least twice in this file, on L81 and L87)

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -31,11 +78,13 @@ internal abstract class AbstractSharedFlow<S : AbstractSharedFlowSlot<*>> : Sync
...
         }
 
+    /** Allocate a new implementation-representation of a collector, but do not register it anywhere yet. */
     protected abstract fun createSlot(): S
 
+    /** Equivalent to [arrayOfNulls]. */
     protected abstract fun createSlotArray(size: Int): Array<S?>
 
-    @Suppress("UNCHECKED_CAST")
+    /** Register a new collector and return its newly allocated slot. A slot may be [created][createSlot] or reused. */
```
murfel: "may be either"

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -75,7 +125,8 @@ internal abstract class AbstractSharedFlow<S : AbstractSharedFlowSlot<*>> : Sync
             subscriptionCount = _subscriptionCount // retrieve under lock if initialized
             // Reset next index oracle if we have no more active collectors for more predictable behavior next time
```
murfel: nit: you removed the other mention/introduction of the term oracle, so now this comment is less clear

murfel: The word oracle was not the best choice, but it was used in the same sense as "hint" in C++ [`map.emplace_hint`](https://en.cppreference.com/w/cpp/container/map/emplace_hint.html). No action, just saying.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
         if (oldIndex > minCollectorIndex) return EMPTY_RESUMES // nothing changes, it was not min
-        // start computing new minimal index of active collectors
+        // start computing the new minimal index of active collectors
         val head = head
-        var newMinCollectorIndex = head + bufferSize
+        var newMinCollectorIndex = head + bufferSize // = this.bufferEndIndex
         // take into account a special case of sync shared flow that can go past 1st queued emitter
         if (bufferCapacity == 0 && queueSize > 0) newMinCollectorIndex++
         forEachSlotLocked { slot ->
-            @Suppress("ConvertTwoComparisonsToRangeCheck") // Bug in JS backend
```
murfel: nit: could you get rid of this now irrelevant suppress in all of the codebase (1 other usage)

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -10,19 +10,66 @@ import kotlin.jvm.*
...
+     * The backing field for [subscriptionCount].
+     *
+     * Will not be initialized until [subscriptionCount] is accessed for the first time.
+     */
+    private var _subscriptionCount: SubscriptionCountStateFlow? = null
+
+    /** A `StateFlow` representing [nCollectors], potentially with some delay. A user-visible API. */
     val subscriptionCount: StateFlow<Int>
         get() = synchronized(this) {
             // allocate under lock in sync with nCollectors variable
```
murfel, line 75: allocate -> initialize

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -31,11 +78,13 @@ internal abstract class AbstractSharedFlow<S : AbstractSharedFlowSlot<*>> : Sync
             }
         }
 
+    /** Allocate a new implementation-representation of a collector, but do not register it anywhere yet. */
```
murfel: I understand the function's meaning but I can't parse "implementation-representation"

murfel: Avoid "allocate" since it's implementation defined where the implementer gets their slots from. There may not be any allocations. Also, I'm not sure if it is common to refer to memory allocations in the JVM world?

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -10,19 +10,66 @@ import kotlin.jvm.*
...
+     */
+    abstract fun freeLocked(flow: F): Array<Continuation<Unit>?>
 }
 
-internal abstract class AbstractSharedFlow<S : AbstractSharedFlowSlot<*>> : SynchronizedObject() {
-    protected var slots: Array<S?>? = null // allocated when needed
+/**
+ * A common data structure for `StateFlow` and `SharedFlow`.
+ */
+internal abstract class AbstractSharedFlow<This, S : AbstractSharedFlowSlot<This>> : SynchronizedObject() {
```
LouisCAD: Is `This` a new Kotlin feature to refer to the type of the instance's subclass?

murfel: It's only a generic parameter name. Could as well be named T. That said, I was also startled when I saw `This`. Possibly rename into something less misleading.

dkhalanskyjb: Nope, you can read it as `T`, I just decided to use a more accurate name.

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -10,19 +10,66 @@ import kotlin.jvm.*
...
-    private var _subscriptionCount: SubscriptionCountStateFlow? = null // init on first need
 
+    /**
+     * A good starting index for looking for the next non-*allocated* slot in [slots].
+     *
+     * It is not guaranteed that this slot will not be *allocated*, nor is it guaranteed that it will be the first
+     * non-*allocated* slot.
+     * This is just a heuristic to have a better guess in common scenarios.
+     */
+    private var nextIndex = 0
```
murfel, line 63: This comment says that `nextIndex` is a "good starting index" / "heuristics" / "guess" but doesn't say how this heuristic is used and what this even means. It does provide two options of what it is definitely not, which is useful, but just would be nice to define the object.

> The next collector will be assigned to the first free slot at `nextIndex` or on the cyclical right of `nextIndex`.

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -10,19 +10,66 @@ import kotlin.jvm.*
 @JvmField
 internal val EMPTY_RESUMES = arrayOfNulls<Continuation<Unit>?>(0)
```
murfel, line 11: First batch of comments. Feel free to address now or wait for the rest to come.

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
murfel: Oh, I just love Github's time travelling machine :)

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -10,19 +10,66 @@ import kotlin.jvm.*
...
+     */
     abstract fun allocateLocked(flow: F): Boolean
-    abstract fun freeLocked(flow: F): Array<Continuation<Unit>?> // returns continuations to resume after lock
+
+    /**
+     * Mark this slot as available for reuse. Only call this under the [flow]'s lock.
+     *
+     * Returns an array of continuations that need to be resumed after the lock is released.
+     * These continuations represent suspended emitters that were waiting for the slowest collector to move on
+     * so that the next value can be placed into the buffer.
```
murfel: for the slowest collector **(assigned to this slot)** to move on

so that the next value -> so that **their values**

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -58,7 +58,7 @@ import kotlin.jvm.*
  * get values from the buffer without suspending emitters. The buffer space determines how much slow subscribers
  * can lag from the fast ones. When creating a shared flow, additional buffer capacity beyond replay can be reserved
  * using the `extraBufferCapacity` parameter.
- *
```
murfel: Could you also fix all other IDE warnings/grammar/style in both files?

Image

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -292,11 +292,23 @@ public fun <T> MutableSharedFlow(
 // ------------------------------------ Implementation ------------------------------------
 
 internal class SharedFlowSlot : AbstractSharedFlowSlot<SharedFlowImpl<*>>() {
+    /**
+     * The first index in the buffer that was not yet consumed by this collector,
```
murfel: nit: that -> which, was not yet -> has not yet been

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -292,11 +292,23 @@ public fun <T> MutableSharedFlow(
...
+     */
     @JvmField
-    var index = -1L // current "to-be-emitted" index, -1 means the slot is free now
+    var index = -1L
 
+    /**
+     * The current continuation of the collector.
+     *
+     * This field is used by the collector each time when it can not acquire the next value and suspends while
+     * waiting for the next one.
```
murfel: Used how? Please formulate more precisely, otherwise it adds more mystery than removes. ("The collector's continuation is saved/written/preserved here" or something)

"cannot", "suspends waiting for the next value to arrive."

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
murfel: head + totalSize is off by 1 character, could you move them?

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -350,7 +362,7 @@ internal open class SharedFlowImpl<T>(
 
     // Stored state
     private var buffer: Array<Any?>? = null // allocated when needed, allocated size always power of two
```
murfel: **is** always **a** power of two

kotlinx-coroutines-core/common/src/flow/internal/AbstractSharedFlow.kt
```diff
@@ -54,7 +103,8 @@ internal abstract class AbstractSharedFlow<S : AbstractSharedFlowSlot<*>> : Sync
                 slot = slots[index] ?: createSlot().also { slots[index] = it }
                 index++
                 if (index >= slots.size) index = 0
-                if ((slot as AbstractSharedFlowSlot<Any>).allocateLocked(this)) break // break when found and allocated free slot
+                @Suppress("UNCHECKED_CAST")
```
murfel, line 110: If you moved this UNCHECKED_CAST, could you move the other one in SharedFlow.kt at L395 and L435.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
murfel: in the **SharedFlow's** buffer

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
         return index
     }
 
-    // Is called when a collector disappears or changes index, returns a list of continuations to resume after lock
+    /**
+     * Handle a collector having receiving the next value from the buffer or disappearing.
```
murfel: ~having~

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
+     * so that they can be resumed.
+     */
     internal fun updateCollectorIndexLocked(oldIndex: Long): Array<Continuation<Unit>?> {
         assert { oldIndex >= minCollectorIndex }
         if (oldIndex > minCollectorIndex) return EMPTY_RESUMES // nothing changes, it was not min
-        // start computing new minimal index of active collectors
+        // start computing the new minimal index of active collectors
         val head = head
-        var newMinCollectorIndex = head + bufferSize
+        var newMinCollectorIndex = head + bufferSize // = this.bufferEndIndex
```
murfel: ==

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
+     *   collectors anymore and can all continue executing immediately.
+     *
+     * In both cases, an array of continuations of the correct number of waiting emitters is returned
+     * so that they can be resumed.
+     */
     internal fun updateCollectorIndexLocked(oldIndex: Long): Array<Continuation<Unit>?> {
         assert { oldIndex >= minCollectorIndex }
         if (oldIndex > minCollectorIndex) return EMPTY_RESUMES // nothing changes, it was not min
-        // start computing new minimal index of active collectors
+        // start computing the new minimal index of active collectors
```
murfel: nit: this comment won't be needed if you make `newMinCollectorIndex` val (`val newMinCollectorIndex = { ... }`)

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
-            @Suppress("ConvertTwoComparisonsToRangeCheck") // Bug in JS backend
-            if (slot.index >= 0 && slot.index < newMinCollectorIndex) newMinCollectorIndex = slot.index
+            if (slot.index in 0..<newMinCollectorIndex) newMinCollectorIndex = slot.index
         }
         assert { newMinCollectorIndex >= minCollectorIndex } // can only grow
         if (newMinCollectorIndex <= minCollectorIndex) return EMPTY_RESUMES // nothing changes
-        // Compute new buffer size if we drop items we no longer need and no emitter is resumed:
-        // We must keep all the items from newMinIndex to the end of buffer
-        var newBufferEndIndex = bufferEndIndex // var to grow when waiters are resumed
+        /* The new minimal resumer index has increased, so new emitters can be resumed,
```
murfel: "resumer" and "minimal resumer index" are new terms, so it is not clear what they refer to. Given that it's only used once, avoid using it at all.

"Buffer space has freed up, so ..."

murfel: Also I don't understand if you're using two different styles of comments/kdoc on purpose or not. /** and /*. If that's on purpose, I struggle to get the semantic difference, in this sea of three types of comments, and it's a bit overwhelming to parse the code with so much semantic encoding.

It's nice to get the kdoc on hover for some complicated local variables, but otherwise it's overwhelming.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
+            // Represents how many elements would be in the buffer if we didn't resume any emitters.
             val newBufferSize0 = (newBufferEndIndex - newMinCollectorIndex).toInt()
+            /** a) We can't awaken more than [queueSize] emitters, as only that many have registered.
+                b) If we awaken more than `bufferCapacity - newBufferSize0` emitters, we'll exceed [bufferCapacity]. */
             minOf(queueSize, bufferCapacity - newBufferSize0)
         } else {
-            // If we don't have collectors anymore we must resume all waiting emitters
-            queueSize // that's how many waiting emitters we have (at most)
+            // If we don't have collectors anymore, we must resume all waiting emitters and drop the buffer.
+            /** [queueSize] is the number of emitters that registered (there may, in fact, be fewer than that). */
```
murfel: Way more emitters registered. This is an upper bound of the number of suspended emitters.

Some of the emitters in the queued emitters section of the buffer array (great naming produces great descriptions) are `NO_VALUE` due to being canceled or having emitted in a previous step.

(I know that you know - I'm just formulating the verbal description that you could rephrase from.)

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
+            queueSize
         }
+        /* Walk through at most `maxResumeCount` emitters, moving their values to the buffer,
+        * zeroing out the cells the emitters were occupying, and storing their continuations in an array that
+        * will be returned from this function and processed once the lock gets released. */
```
murfel: [IDE hint] Long sentences (41 words here) are harder to read, according to the research; consider splitting

A human confirms. Consider making it into a list, or somehow else rewriting it.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
murfel: make -> mark

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
murfel: TODO: is this the number of emiitters that registered though?

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -318,7 +330,7 @@ internal open class SharedFlowImpl<T>(
     private val replay: Int,
     private val bufferCapacity: Int,
     private val onBufferOverflow: BufferOverflow
-) : AbstractSharedFlow<SharedFlowSlot>(), MutableSharedFlow<T>, CancellableFlow<T>, FusibleFlow<T> {
+) : AbstractSharedFlow<SharedFlowImpl<*>, SharedFlowSlot>(), MutableSharedFlow<T>, CancellableFlow<T>, FusibleFlow<T> {
     /*
         Logical structure of the buffer
```
murfel: Suggestion: buffer is a confusing naming and is used to mean 3 different things:
1. the buffer array of buffered values (including the replay cache) and queued emitters
2. the buffer of buffered values **including** the replay cache
3. the buffer of buffered values **excluding** the replay cache

2 and 3 is more or less consistent in the kdoc and code and steer towards 2

But the meanings 1 and 2 are used intermittently throughout the code. Could we rename the buffer in this sense into the **buffer array**? To refer to the code representation.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -578,21 +609,21 @@ internal open class SharedFlowImpl<T>(
                 }
             }
         }
-        // Compute new buffer size -> how many values we now actually have after resume
+        /** Represents how many values are stored in the buffer after processing the emitters.
+         * The value may be larger than [bufferCapacity] if [nCollectors] was 0,
+         * but the `minOf(replay)` will clamp this value to a valid one, since `replay <= bufferCapacity`. */
```
murfel: This value is always non-smaller than [bufferCapacity], and could be larger in another case. The `head` hasn't moved and still points to the original slowest possible collector. The `newBufferEndIndex` has started out as `bufferEndIndex == head + bufferSize` (at max that would be `head + bufferCapacity`) and could only increase.

Also "Represents how many values" -> "The number of values stored in the buffer after processing the emitters."

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -578,21 +609,21 @@ internal open class SharedFlowImpl<T>(
...
         // Compute new replay size -> limit to replay the number of items we need, take into account that it can only grow
         var newReplayIndex = maxOf(replayIndex, newBufferEndIndex - minOf(replay, newBufferSize1))
         // adjustment for synchronous case with cancelled emitter (NO_VALUE)
         if (bufferCapacity == 0 && newReplayIndex < newQueueEndIndex && buffer!!.getBufferAt(newReplayIndex) == NO_VALUE) {
             newBufferEndIndex++
             newReplayIndex++
         }
+        /** If [nCollectors] was 0, there are no active collector slots,
+         * and the invariant dictates that [minCollectorIndex] should be [bufferEndIndex] (`replayIndex + bufferSize`).
+         * Ensure this directly. */
```
murfel: The original comment asserted that this invariant **can in fact be broken**, and thus this assignment is indeed required, and not just some unnecessary duct tape.

murfel (changes requested):

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
dkhalanskyjb: I don't understand this point. The invariant can't be broken, we're under the lock. However, to ensure the invariant stays non-broken, this assignment is required.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -532,36 +544,55 @@ internal open class SharedFlowImpl<T>(
...
+     */
     internal fun updateCollectorIndexLocked(oldIndex: Long): Array<Continuation<Unit>?> {
         assert { oldIndex >= minCollectorIndex }
         if (oldIndex > minCollectorIndex) return EMPTY_RESUMES // nothing changes, it was not min
-        // start computing new minimal index of active collectors
+        // start computing the new minimal index of active collectors
         val head = head
-        var newMinCollectorIndex = head + bufferSize
+        var newMinCollectorIndex = head + bufferSize // = this.bufferEndIndex
         // take into account a special case of sync shared flow that can go past 1st queued emitter
```
murfel: Could you also replace this cryptic "sync shared flow" to "shared flow with no buffer", here and below, to consistently call this situation the same name.

PR #4596, opened 2025-12-23
Clarify the behavior of CoroutineExceptionHandler

### LINKED ISSUE #1157: CoroutineExceptionHandler not called in top-most parent-scope.
I have this example code-snipper:
```
val xHandlerParent = CoroutineExceptionHandler { c, e ->
    println("Parent Handled Crash")
}

val xHandlerLeaf = CoroutineExceptionHandler { c, e ->
    println("Leaf Handled Crash")
}

fun test1() {
    // Exception throw up the call-stack. Expected
    CoroutineScope(Job()).launch {
        delay(1000)
        throw Exception("Some Exception 1")
    }

    Thread.sleep(2000)
}

fun test2() {
    // Exception handled by xHandlerParent: Expected
    CoroutineScope(xHandlerParent).launch {
        delay(1000)
        throw Exception("Some Exception 1")
    }

    Thread.sleep(2000)
}

fun test3() {
    // Exception handled by xHandlerLeaf: Expected?
    // Yes, since there is a top scope with a CoroutineExceptionHandler.
    // No,  since the top-most scope does not have CoroutineExceptionHandler.
    CoroutineScope(Job()).launch(xHandlerLeaf) {
        delay(1000)
        throw Exception("Some Exception 1")
    }

    Thread.sleep(2000)
}

fun test4() {
    // Exception handled by xHandlerLeaf. Unexpected.
    // Shouldn't it be handled by xHandlerParent instead?
    CoroutineScope(xHandlerParent).launch(xHandlerLeaf) {
        delay(1000)
        throw Exception("Some Exception 1")
    }

    Thread.sleep(2000)
}

fun test5() {
    // Exception handled by xHandlerParent: Expected
    CoroutineScope(xHandlerParent).launch {
        withContext(xHandlerLeaf) {
            delay(1000)
            throw Exception("Some Exception 1")
        }
    }

    Thread.sleep(2000)
}

fun test6() {
    // Exception handled by xHandlerParent: Expected?
    // Yes, since there is a top scope with a CoroutineExceptionHandler.
    // No,  since the top-most scope does not have CoroutineExceptionHandler.
    CoroutineScope(Job()).launch(xHandlerParent) {
        launch(xHandlerLeaf) {
            delay(1000)
            throw Exception("Some Exception 1")
        }
    }

    Thread.sleep(2000)
}

fun test7() {
    // Exception thrown up the call-stack. Expected?
    // Yes, since the top-most scope does not have CoroutineExceptionHandler.
    // No, since there is a top scope with a CoroutineExceptionHandler.
    CoroutineScope(Job()).launch {
        withContext(xHandlerLeaf) {
            delay(1000)
            throw Exception("Some Exception 1")
        }
    }

    Thread.sleep(2000)
}
```

My questions are:
- Is the top-most scope always used and if it has `CoroutineExceptionHandler`, it will be used, otherwise, the exception is thrown up the call-stack?
- Or is the top scope that has a `CoroutineExceptionHandler` used instead? Only if no such top scope can be found, the exception is thrown up the call-stack.
- Regardless of the answer to the previous question, why does the example of `test4()` not use the `xHandlerParent`, since that is the `CoroutineExceptionHandler` of the top and top-most scope?

### LINKED ISSUE #4504: Documentation needs adjustments: states async cannot cause uncaught exceptions but async can if fails before it is awaited.
The above states that:  `async` cannot create uncaught exceptions.

However: IF `async` throws an exception BEFORE we called `await` on the deferred object, THEN uncaught exception happens BUT `CoroutineExceptionHandler` does not appear to be invoked. 

#### Code self contained

```
package com.glassthought.sandbox

import kotlinx.coroutines.*
import java.time.Instant
import kotlin.time.Duration
import kotlin.time.Duration.Companion.milliseconds


fun main(): Unit {
  val coroutineExceptionHandler = CoroutineExceptionHandler { _, ex ->
    runBlocking(CoroutineName("CoroutineExceptionHandler")) {
      printlnWithTime("Caught exception in CoroutineExceptionHandler: ${ex::class.simpleName} with message=[${ex.message}].")
    }
  }

  val mainJob = Job()
  val scope = CoroutineScope(mainJob + coroutineExceptionHandler)

  scope.launch(CoroutineName("SpawnedFromMain")) {
    actionWithMsg("foo", { foo(scope) })
  }

  runBlocking {
    actionWithMsg("mainJob.join()", { mainJob.join() })
  }
}

private suspend fun foo(scope: CoroutineScope) {
  val deferred = scope.async(CoroutineName("async-1")) {
    actionWithMsg("throw-exception", {
      throw MyRuntimeException.create("exception-from-async-before-it-got-to-await")
    }, delayDuration = 500.milliseconds)
  }

  printlnWithTime("Just launched co-routines.")
  actionWithMsg(
    "deferred.await()",
    {
      deferred.await()
    },
    delayDuration = 1.seconds
  )
}

suspend fun <T> actionWithMsg(
  actionName: String,
  action: suspend () -> T,
  delayDuration: Duration = Duration.ZERO
): T {
  val hasDelay = delayDuration > Duration.ZERO
  if (hasDelay) {
    printlnWithTime("action=[$actionName] is being delayed for=[${delayDuration.inWholeMilliseconds} ms] before starting.")

    delay(delayDuration)
  }
  printlnWithTime("action=[$actionName] is starting.")

  var result: T


  try {
    result = action()
  } catch (exc: Exception) {
    // We are going to rethrow all exceptions, so CancellationException will also be rethrown.
    // Therefore, we respect: [Cooperative Cancellation](http://www.glassthought.com/notes/3ha01u9931je002miy86vdo)
    if (exc is CancellationException) {
      printlnWithTime("Cancellation Exception - rethrowing.")
    } else {
      printlnWithTime("Finished action=[$actionName], it THREW exception of type=[${exc::class.simpleName}] we are rethrowing it.")
    }
    throw exc
  }

  printlnWithTime("Finished action=[$actionName].")

  return result
}

class MyRuntimeException private constructor(msg: String) : RuntimeException(msg) {
  companion object {
    suspend fun create(msg: String): MyRuntimeException {
      val exc = MyRuntimeException(msg)

      printlnWithTime("throwing exception=[${exc::class.simpleName}] with msg=[${msg}]")

      return exc
    }
  }
}

fun printlnWithTime(msg: String) {
  println("[${Instant.now()}]: " + msg)
}
```
#### Original Code
```kotlin
package com.glassthought.sandbox

import com.glassthought.sandbox.util.out.impl.out
import kotlinx.coroutines.*
import kotlin.time.Duration.Companion.milliseconds


fun main(): Unit {
  val coroutineExceptionHandler = CoroutineExceptionHandler { _, ex ->
    runBlocking(CoroutineName("CoroutineExceptionHandler")) {
      out.error("Caught exception in CoroutineExceptionHandler: ${ex::class.simpleName} with message=[${ex.message}].")
    }
  }

  val mainJob = Job()
  val scope = CoroutineScope(mainJob + coroutineExceptionHandler)

  scope.launch(CoroutineName("SpawnedFromMain")) {
    out.actionWithMsg("foo", { foo(scope) })
  }
  
  runBlocking {
    out.actionWithMsg("mainJob.join()", { mainJob.join() })
  }
}

private suspend fun foo(scope: CoroutineScope) {
  val deferred = scope.async(CoroutineName("async-1")) {
    out.actionWithMsg("throw-exception", {
      throw MyRuntimeException.create("exception-from-async-before-it-got-to-await", out)
    }, delayDuration = 500.milliseconds)
  }

  out.info("Just launched co-routines.")
  out.actionWithMsg(
    "deferred.await()",
    {
      deferred.await()
    },
    delayDuration = 1.seconds
  )
}
```

#### Recorded output of command:
```txt
Picked up JAVA_TOOL_OPTIONS: -Dkotlinx.coroutines.debug
Picked up JAVA_TOOL_OPTIONS: -Dkotlinx.coroutines.debug
[INFO][elapsed:   18ms][①][coroutname:@SpawnedFromMain#1] [->] action=[foo] is starting.
[INFO][elapsed:   18ms][⓶][coroutname:@coroutine#2] [->] action=[mainJob.join()] is starting.
[INFO][elapsed:   31ms][①][coroutname:@SpawnedFromMain#1]    Just launched co-routines.
[INFO][elapsed:   35ms][①][coroutname:@SpawnedFromMain#1]    [🐢] action=[deferred.await()] is being delayed for=[1000 ms] before starting.
[INFO][elapsed:   35ms][⓷][coroutname:@async-1#3] [🐢] action=[throw-exception] is being delayed for=[500 ms] before starting.
[INFO][elapsed:  535ms][⓷][coroutname:@async-1#3] [->] action=[throw-exception] is starting.
[WARN][elapsed:  557ms][⓷][coroutname:@async-1#3]    💥 throwing exception=[MyRuntimeException] with msg=[exception-from-async-before-it-got-to-await]
[WARN][elapsed:  557ms][⓷][coroutname:@async-1#3] [<-][💥] Finished action=[throw-exception], it THREW exception of type=[MyRuntimeException] we are rethrowing it.
[INFO][elapsed:  568ms][①][coroutname:@SpawnedFromMain#1] [<-][🫡] Cancellation Exception - rethrowing.
[INFO][elapsed:  569ms][⓶][coroutname:@coroutine#2] [<-] Finished action=[mainJob.join()].
```

#### Output colored per co-routine:
Image

dkhalanskyjb:
The existing documentation didn't specify in detail which exceptions were considered unpropagated,
and was also using the confusing "uncaught" wording.

Fixes #1157
Fixes #4504

### DIFF
--- a/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
@@ -287,7 +287,7 @@ public interface CoroutineExceptionHandler : CoroutineContext.Element {
     public companion object Key : CoroutineContext.Key<CoroutineExceptionHandler>
 
     /**
-     * Handles an [exception] in the given [context].
+     * Handles an [exception] that occurred in the given [context].
      * It is invoked if a coroutine fails without a clear propagation path,
      * as described in the [CoroutineExceptionHandler] documentation.
      *
--- a/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
@@ -4,15 +4,15 @@ import kotlinx.coroutines.internal.*
 import kotlin.coroutines.*
 
 /**
- * Helper function for coroutine builder implementations to handle uncaught and unexpected exceptions in coroutines,
- * that could not be otherwise handled in a normal way through structured concurrency, saving them to a future, and
+ * Helper function for coroutine builder implementations to handle unpropagated and unexpected exceptions in coroutines,
+ * that could not be otherwise handled normally through structured concurrency or saving them to a `Future`, and
  * cannot be rethrown. This is a last resort handler to prevent lost exceptions.
  *
  * If there is [CoroutineExceptionHandler] in the context, then it is used. If it throws an exception during handling
  * or is absent, all instances of [CoroutineExceptionHandler] found via `ServiceLoader` and
  * `Thread.uncaughtExceptionHandler` are invoked.
  *
- * @suppress **This is internal API and it is subject to change.**
+ * @suppress **This is an internal API and it is subject to change.**
  */
 @InternalCoroutinesApi
 public fun handleCoroutineException(context: CoroutineContext, exception: Throwable) {
@@ -48,7 +48,7 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
  * [handler] is invoked inside coroutine machinery on an unspecified thread.
  * Therefore, it must be thread-safe and finish quickly.
  *
- * Throwing exceptions from this method is discouraged and
+ * Throwing exceptions in [handler] is discouraged and
  * will invoke platform-specific last-resort exception handling,
  * described in the [CoroutineExceptionHandler] interface documentation.
  */
@@ -177,7 +177,7 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  *
  * ```
  * GlobalScope.launch {
- *     throw IllegalStateException("This *is* an unpropagated exception.")
+ *     throw IllegalStateException("This is an **unpropagated exception**.")
  * }
  * ```
  *
--- a/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
@@ -59,7 +59,7 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * In most scenarios, a clear exception propagation path allows processing failures in coroutines: for example,
  * a [coroutineScope] call can rethrow the exception to the caller,
  * and failing coroutines typically [cancel][Job.cancel] their [parent][Job.parent] coroutines.
- * See "Propagation channels recognized by `kotlinx.coroutines`" below for an enumeration of ways an exception in a
+ * See "Propagation paths recognized by `kotlinx.coroutines`" below for an enumeration of ways an exception in a
  * coroutine can get propagated.
  *
  * However, in some cases, a clear propagation path is not available. Example:
@@ -89,7 +89,7 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * by `kotlinx.coroutines` and will invoke last-resort exception handling, potentially crashing the program.
  * See the "Platform-specific last-resort handling of lost exceptions" section for details.
  *
- * ### Propagation channels recognized by `kotlinx.coroutines`
+ * ### Propagation paths recognized by `kotlinx.coroutines`
  *
  * The only exceptions that need to be propagated at all are those with which coroutines finish.
  * If an exception is handled via a `try`/`catch` block inside the coroutine itself,
@@ -108,7 +108,8 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * }
  * ```
  *
- * Exceptions in lexically scoped coroutines are always propagated by being rethrown to the caller:
+ * Exceptions in lexically scoped coroutines (those that, like [coroutineScope], return the result to the caller)
+ * are always propagated by being rethrown to the caller:
  *
  * ```
  * // This function will throw an `IllegalStateException`
@@ -148,7 +149,7 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * }
  * ```
  *
- * When none of the propagation channels listed above applies, an exception can not be propagated.
+ * When none of the propagation paths listed above apply, an exception can not be propagated.
  * Most common examples are coroutines created using the [launch] function on a scope with no [Job] (most notably,
  * [GlobalScope]) or a [SupervisorJob]:
  *
@@ -201,6 +202,7 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  *     try {
  *          // do something
  *     } catch (e: Throwable) {
+ *          if (e is CancellationException) ensureActive()
  *          // handle exception
  *     }
  * }
@@ -252,7 +254,7 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * [CoroutineExceptionHandler].
  * The explanation is that initially, the exception *does* have a propagation path and will get propagated to the
  * parent coroutine using structured concurrency.
- * The parent itself, however, has no viable propagation channel and has to use *its* [CoroutineExceptionHandler].
+ * The parent itself, however, has no viable propagation path and has to use *its* [CoroutineExceptionHandler].
  *
  * Similarly, this [CoroutineExceptionHandler] is redundant and will never be invoked:
  *
--- a/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
     }
 
 /**
- * An optional element in the coroutine context to handle **uncaught** exceptions.
+ * An optional element in the [CoroutineContext] to handle coroutine exceptions without a clear propagation path.
  *
- * Normally, uncaught exceptions can only result from _root_ coroutines created using the [launch][CoroutineScope.launch] builder.
- * All _children_ coroutines (coroutines created in the context of another [Job]) delegate handling of their
- * exceptions to their parent coroutine, which also delegates to the parent, and so on until the root,
- * so the `CoroutineExceptionHandler` installed in their context is never used.
- * Coroutines running with [SupervisorJob] do not propagate exceptions to their parent and are treated like root coroutines.
- * A coroutine that was created using [async][CoroutineScope.async] always catches all its exceptions and represents them
- * in the resulting [Deferred] object, so it cannot result in uncaught exceptions.
+ * `kotlinx.coroutines` ensures that every [exception][Throwable] thrown in a coroutine is handled by some code instead
+ * of being forgotten about.
+ * This helps failures not to go unnoticed, with the program silently entering an incorrect state.
  *
- * ### Handling coroutine exceptions
+ * In most scenarios, a clear exception propagation path allows processing failures in coroutines: for example,
+ * a [coroutineScope] call can rethrow the exception to the caller,
+ * and failing coroutines typically [cancel][Job.cancel] their [parent][Job.parent] coroutines.
+ * See "Propagation channels recognized by `kotlinx.coroutines`" below for an enumeration of ways an exception in a
+ * coroutine can get propagated.
  *
- * `CoroutineExceptionHandler` is a last-resort mechanism for global "catch all" behavior.
- * You cannot recover from the exception in the `CoroutineExceptionHandler`. The coroutine had already completed
- * with the corresponding exception when the handler is called. Normally, the handler is used to
- * log the exception, show some kind of error message, terminate, and/or restart the application.
- *
- * If you need to handle the exception in a specific part of the code, it is recommended to use `try`/`catch` around
- * the corresponding code inside your coroutine. This way you can prevent completion of the coroutine
- * with the exception (exception is now _caught_), retry the operation, and/or take other arbitrary actions:
+ * However, in some cases, a clear propagation path is not available. Example:
  *
  * ```
- * scope.launch { // launch child coroutine in a scope
+ * supervisorScope {
+ *     launch { error("Failure") }
+ * }
+ * ```
+ *
+ * Here, the coroutine created by `launch` fails with the exception `"Failure"`,
+ * and [supervisorScope] does not react to exceptions from its children, as opposed to [coroutineScope].
+ *
+ * In such cases, a [CoroutineExceptionHandler] should be used to process the exceptions:
+ *
+ * ```
+ * withContext(CoroutineExceptionHandler { ctx, ex ->
+ *     println("Exception $ex thrown from coroutine context $ctx")
+ * }) {
+ *     supervisorScope {
+ *         launch { error("Failure") }
+ *     }
+ * }
+ * ```
+ *
+ * Not handling a lost exception with a [CoroutineExceptionHandler] is treated as a programming error
+ * by `kotlinx.coroutines` and will invoke last-resort exception handling, potentially crashing the program.
+ * See the "Platform-specific last-resort handling of lost exceptions" section for details.
+ *
+ * ### Propagation channels recognized by `kotlinx.coroutines`
+ *
+ * The only exceptions that need to be propagated at all are those with which coroutines finish.
+ * If an exception is handled via a `try`/`catch` block inside the coroutine itself,
+ * the coroutine machinery will not even learn about it:
+ *
+ * ```
+ * launch {
+ *     try {
+ *         throw IllegalStateException("""
+ *             This is not an unpropagated exception,
+ *             since it gets caught.
+ *         """)
+ *     } catch (_: IllegalStateException) {
+ *         println("Caught an exception")
+ *     }
+ * }
+ * ```
+ *
+ * Exceptions in lexically scoped coroutines are always propagated by being rethrown to the caller:
+ *
+ * ```
+ * // This function will throw an `IllegalStateException`
+ * coroutineScope {
+ *     throw IllegalStateException("""
+ *         This exception is propagated
+ *         by being rethrown to the caller.
+ *     """)
+ * }
+ * ```
+ *
+ * An exception is considered to be propagated if it's transferred to the parent
+ * through [structured concurrency][CoroutineScope]:
+ *
+ * ```
+ * coroutineScope {
+ *     launch {
+ *         throw IllegalStateException("""
+ *             This exception is propagated
+ *             by `launch` to its parent (`coroutineScope`)
+ *             that is able to process child exceptions.
+ *         """)
+ *     }
+ * }
+ * ```
+ *
+ * Finally, an exception is considered to have been propagated
+ * if the return value of the coroutine builder allows querying the result of the coroutine's execution:
+ *
+ * ```
+ * val deferred = GlobalScope.async {
+ *     throw IllegalStateException("""
+ *         This exception is propagated,
+ *         since calling `await()` on `deferred`
+ *         will rethrow the exception.
+ *     """)
+ * }
+ * ```
+ *
+ * When none of the propagation channels listed above applies, an exception can not be propagated.
+ * Most common examples are coroutines created using the [launch] function on a scope with no [Job] (most notably,
+ * [GlobalScope]) or a [SupervisorJob]:
+ *
+ * ```
+ * supervisorScope {
+ *     launch {
+ *         throw IllegalStateException("""
+ *             This *is* an exception with no propagation path, since
+ *             1. The block of `launch` finishes with it.
+ *             2. `launch` is not lexically scoped,
+ *             3. `supervisorScope` does not handle the failures in children,
+ *             4. `launch` returns a `Job`, which does not allow querying the exception.
+ *         """)
+ *     }
+ * }
+ * ```
+ *
+ * ```
+ * GlobalScope.launch {
+ *     throw IllegalStateException("This *is* an unpropagated exception.")
+ * }
+ * ```
+ *
+ * ### Platform-specific last-resort handling of lost exceptions
+ *
+ * When no [CoroutineExceptionHandler] is present in the [CoroutineContext] of the failing coroutine,
+ * an exception with no propagation path is handled in the following way as the last-resort measure:
+ *
+ * - On JVM, all instances of [CoroutineExceptionHandler] found via `ServiceLoader` and
+ *   the current thread's [Thread.uncaughtExceptionHandler] are invoked.
+ * - On Native, the whole application crashes with the exception.
+ * - On JS and Wasm JS, the exception is propagated into the JavaScript runtime's event loop
+ *   and is processed in a platform-specific way determined by the platform itself.
+ *
+ * ### Recommended patterns for handling coroutine exceptions
+ *
+ * A [CoroutineExceptionHandler] is intended to be a more lenient version of the platform-specific last-resort handling
+ * of coroutine exceptions, allowing one to log exceptions, show an error message, restart the program,
+ * and in general, fail more gracefully.
+ * It is not a replacement for handling exceptions in the normal control flow, and it only gets invoked after the
+ * coroutine has completed and can no longer be resumed.
+ *
+ * If you need to handle the exception in a specific part of the code, it is recommended to use `try`/`catch` around
+ * the corresponding code inside your coroutine instead of relying on a [CoroutineExceptionHandler].
+ * This way, you can prevent completion of the coroutine with the exception,
+ * retry the operation, and/or take arbitrary other actions:
+ *
+ * ```
+ * scope.launch { // launch a child coroutine in a scope
  *     try {
  *          // do something
  *     } catch (e: Throwable) {
@@ -81,19 +206,63 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * }
  * ```
  *
- * ### Uncaught exceptions with no handler
+ * Alternatively, whenever a failure is indeed supposed to terminate a coroutine,
+ * using [async] instead of [launch] and later calling [Deferred.await] instead of [Job.join]
+ * on its result to check if the computation was successful will allow gracefully processing the exception.
  *
- * When no handler is installed, an exception is handled in the following way:
- * - If the exception is [CancellationException], it is ignored, as these exceptions are used to cancel coroutines.
- * - Otherwise, if there is a [Job] in the context, then [Job.cancel] is invoked.
- * - Otherwise, as a last resort, the exception is processed in a platform-specific manner:
- *   - On JVM, all instances of [CoroutineExceptionHandler] found via `ServiceLoader`, as well as
- *     the current thread's `Thread.uncaughtExceptionHandler`, are invoked.
- *   - On Native, the whole application crashes with the exception.
- *   - On JS and Wasm JS, the exception is propagated into the JavaScript runtime's event loop
- *     and is processed in a platform-specific way determined by the platform itself.
+ * ### Pitfalls
  *
- * [CoroutineExceptionHandler] can be invoked from an arbitrary thread.
+ * #### Using a [CoroutineExceptionHandler] does not prevent coroutine failures
+ *
+ * A [CoroutineExceptionHandler] is only called after the coroutine completes if, informally,
+ * the exception has nowhere else to go.
+ * A common pitfall is trying to use a [CoroutineExceptionHandler] to prevent the expected failure of a child coroutine
+ * from also making the parent coroutine fail:
+ *
+ * ```
+ * coroutineScope {
+ *     launch(CoroutineExceptionHandler { ctx, ex ->
+ *         println("This line will not be printed!")
+ *     }) {
+ *         error("This failure will cancel `coroutineScope`!")
+ *     }
+ * }
+ * ```
+ *
+ * A [CoroutineExceptionHandler] has no effect in this scenario: the failure in [launch] will cancel [coroutineScope].
+ * Using `try`/`catch` is the proper way to achieve this.
+ *
+ * See "Recommended patterns for handling coroutine exceptions" above for more details.
+ *
+ * #### Overriding [CoroutineExceptionHandler] in coroutines with exception propagation paths has no effect
+ *
+ * Consider this snippet:
+ *
+ * ```
+ * // Launch the parent coroutine:
+ * GlobalScope.launch(CoroutineExceptionHandler { ctx, e -> println("Outer") }) {
+ *     // Launch the child coroutine:
+ *     launch(CoroutineExceptionHandler { ctx, e -> println("Nested") }) {
+ *         error("Error")
+ *     }
+ * }
+ * ```
+ *
+ * It will print `Outer`, even though the coroutine where the failure originally happened specifies its own
+ * [CoroutineExceptionHandler].
+ * The explanation is that initially, the exception *does* have a propagation path and will get propagated to the
+ * parent coroutine using structured concurrency.
+ * The parent itself, however, has no viable propagation channel and has to use *its* [CoroutineExceptionHandler].
+ *
+ * Similarly, this [CoroutineExceptionHandler] is redundant and will never be invoked:
+ *
+ * ```
+ * GlobalScope.async(CoroutineExceptionHandler { ctx, e -> println("Failure") }) {
+ *     error("Error")
+ * }
+ * ```
+ *
+ * The caller of [async] is responsible for handling the exceptions in the returned [Deferred] value.
  */
 public interface CoroutineExceptionHandler : CoroutineContext.Element {
     /**
@@ -104,6 +273,13 @@ public interface CoroutineExceptionHandler : CoroutineContext.Element {
     /**
      * Handles uncaught [exception] in the given [context]. It is invoked
      * if the coroutine has an uncaught exception.
+     *
+     * [handleException] is invoked inside coroutine machinery in an unspecified thread.
+     * Therefore, it should be thread-safe and finish quickly.
+     *
+     * Throwing exceptions from this method is discouraged and
+     * will invoke platform-specific last-resort exception handling,
+     * described in the [CoroutineExceptionHandler] documentation.
      */
     public fun handleException(context: CoroutineContext, exception: Throwable)
 }
--- a/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
+++ b/kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
@@ -40,9 +40,18 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
 
 /**
  * Creates a [CoroutineExceptionHandler] instance.
- * @param handler a function which handles exception thrown by a coroutine
+ *
+ * When an exception without a propagation path is thrown, [handler] is invoked with the [CoroutineContext]
+ * of the coroutine where the exception occurred, as well as the problematic [Throwable] itself.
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
     }
 
 /**
- * An optional element in the coroutine context to handle **uncaught** exceptions.
+ * An optional element in the [CoroutineContext] to handle coroutine exceptions without a clear propagation path.
```
murfel, line 62: You've somewhat created a bunch of new terms to replace a single old one. It used to be _uncaught exceptions_, now it is _unpropagated exceptions_, and it is introduced indirectly through mentions of a mix of wordings:
* clear propagation paths
* exception propagation paths
* clear exception propagation paths
* propagation channels

To mean "means of handling exceptions which come after try-catch but before CoroutineExceptionHandler".

It definitely needs to be standardised. I suggest either "exception propagation paths" or "propagation channels", or something else. (Drop the "clear". It makes it sound informal, but we do want to signal that it is a formal term).

No strong opinion on which term to use, as both have downsides and I cannot come up with a better one. "exception propagation paths" are too lengthy and "propagation channels" misleadingly mention channels.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
  *
- * Normally, uncaught exceptions can only result from _root_ coroutines created using the [launch][CoroutineScope.launch] builder.
- * All _children_ coroutines (coroutines created in the context of another [Job]) delegate handling of their
- * exceptions to their parent coroutine, which also delegates to the parent, and so on until the root,
- * so the `CoroutineExceptionHandler` installed in their context is never used.
- * Coroutines running with [SupervisorJob] do not propagate exceptions to their parent and are treated like root coroutines.
- * A coroutine that was created using [async][CoroutineScope.async] always catches all its exceptions and represents them
- * in the resulting [Deferred] object, so it cannot result in uncaught exceptions.
+ * `kotlinx.coroutines` ensures that every [exception][Throwable] thrown in a coroutine is handled by some code instead
```
murfel: It feels somehow off to refer to the library name. I like how you refer to the magic _coroutine machinery_ below. Use that?

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
+ * of being forgotten about.
+ * This helps failures not to go unnoticed, with the program silently entering an incorrect state.
```
murfel: Too verbose.
> ensures that every exception thrown in a coroutine is handled.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
  *
```
murfel, line 66: I'd also add an overview of the upcoming exception handling hierarchy:
1. try-catch
2. "exception propagation paths"
3. CoroutineExceptionHandler
4. final resort

The #2 could also benefit from a brief overview (lexically scoped, parent, return value (deferred)).

It would be better to put it in a very brief listicle rather than complete sentences and acknowledge that it doesn't provide the full information yet.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
- * ### Handling coroutine exceptions
+ * In most scenarios, a clear exception propagation path allows processing failures in coroutines: for example,
+ * a [coroutineScope] call can rethrow the exception to the caller,
+ * and failing coroutines typically [cancel][Job.cancel] their [parent][Job.parent] coroutines.
+ * See "Propagation channels recognized by `kotlinx.coroutines`" below for an enumeration of ways an exception in a
```
murfel: I appreciate that you want to cut to the chase as early as possible. Consider doing it even sooner with a tiny example after the first sentence (for those who are familiar but only want to lookup the syntax of adding a CEH), and then you'll be able to keep the narrative linear without forward referencing. Currently it reads, "if your coroutines don't have something vague, for example in this case, use CoroutineExceptionHandler".

You already mention the `supervisorScope` example twice: once in this section, and once at the end of the Propagation channels section.

If you add a quick `supervisorScope` example at the top and have a linear story, it will only appear once later, on the border of the Propagation Channels and CEH sections.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -50,29 +50,154 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
...
+ *             This is not an unpropagated exception,
+ *             since it gets caught.
+ *         """)
+ *     } catch (_: IllegalStateException) {
+ *         println("Caught an exception")
+ *     }
+ * }
+ * ```
+ *
+ * Exceptions in lexically scoped coroutines are always propagated by being rethrown to the caller:
```
murfel: Can we by any chance add a hyperlink to anywhere where "lexically scoped" is defined?

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -81,19 +206,63 @@ public inline fun CoroutineExceptionHandler(crossinline handler: (CoroutineConte
  * }
```
murfel, line 218: rethrow CancellationException

murfel (changes requested):

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: I attempted this and didn't like the result:
```kotlin
  * There are scenarios when there is no clear propagation path that `kotlinx.coroutines` can take to propagate an
  * exception. Example:
  *
  * ```
  * supervisorScope {
  *     launch { error("Failure") }
  * }
  * ```
  *
  * Here, the coroutine created by `launch` fails with the exception `"Failure"`.
  * If instead of [supervisorScope], [coroutineScope] were used, the exception would be propagated to the caller of
  * [coroutineScope]. First, the failed [launch] would call [Job.cancel] on the [coroutineScope], which would lead
  * to [coroutineScope] failing. Then, since [coroutineScope] returns its result to the caller, the exception would
  * be rethrown. `kotlinx.coroutines` considers this a propagation path.
  *
  * However, [supervisorScope] does not react to exceptions from its children, as opposed to [coroutineScope],
  * which means [Job.cancel] would not be called.
  *
  * In such cases, a [CoroutineExceptionHandler] should be used to process the exceptions:
  *
  * ```
  * withContext(CoroutineExceptionHandler { ctx, ex ->
  *     println("Exception $ex thrown from coroutine context $ctx")
  * }) {
  *     supervisorScope {
  *         launch { error("Failure") }
  *     }
  * }
  * ```
  *
  * In most scenarios, a clear exception propagation path allows processing failures in coroutines.
  * Code should normally be structured in a way that ensures the existence of exception propagation paths.
  * See "Propagation paths recognized by `kotlinx.coroutines`" below.
```
* The minor role of `CoroutineExceptionHandler` is not as prominent. "There are scenarios" does not give any indication that these scenarios are unusual, and I didn't find a more natural way of expressing this thought than just starting with the idea that `CoroutineExceptionHandler` is unnecessary in most cases.
* The logic is backwards: to explain why `supervisorScope { launch }` fails to meet the propagation path criteria, we have to introduce them afterwards, overloading the reader.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: Your proposed formulation misses an important nuance. By "handling", people often mean "*appropriately* reacting to", but `kotlinx.coroutines` can't do that. The appropriate action heavily depends on the use case. The extra wording here weakens the "handling" to "*somehow* reacting to". I don't think rewording this would help with verbosity, either, as this is in any case an important point that shouldn't stay implicit.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: The other usages of "coroutine machinery" (`git grep machinery`) indicate that it's the sum of `JobSupport`, `CancellableContinuationImpl`, etc.—a specific implementation of how coroutines interact, suspend, resume, etc. With that in mind, using "the coroutines machinery" here becomes misleading, as not only the low-level implementation of corouttines themselves follows this principle, but other parts of the library as well: `git grep handleCoroutineException`. The mention of `kotlinx.coroutines` here is intended to evoke the feeling of it being a general guiding principle for our APIs.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: "Propagation channels" is a nice catch! We have already discussed it once already under the PR with the structured concurrency documentation, and you raised a valuable point that "channel" is a specific term in concurrent programming. Removed all mentions of channels to prevent confusion.

> Drop the "clear". It makes it sound informal, but we do want to signal that it is a formal term

I don't think that we do at that point. "Clear" only occurs in the first section, whose goal is to provide a high-level intuition of the forces at play. In my view, it is perfectly fine to add extra adjectives/verbosity to aid intuition there. The goal of "clear" is to contrast the propagation paths discussed here with those implied by the business logic. Generally, before it's clear that "propagation paths" is a precise characterization and not just a description of circumstances, "coroutine exceptions without a propagation path" looks meaningless to me, and "coroutine exceptions that `kotlinx.coroutines` can not propagate to where the client code can be expected to observe and react to them" feels detailed enough to obscure the high-level idea.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: I'm wary that this would postpone explaining the general idea of what a `CoroutineExceptionHandler` is and when it's needed. Would be a waste to have a reader go through this list, figure out the exception propagation paths, and then realize that hey, their code is already completely idiomatic, there is no need to be concerned about such low-level details.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
murfel: I only took a prefix of your text. 

"instead of being forgotten about" and "This helps failures not to go unnoticed, with the program silently entering an incorrect state" feel like general discussion on why it's important to handle exceptions, not relaxation of the meaning of "handle".

Let's relax the meaning of "handle" into "reacted to" and also shorten the discussion (which is useful to mention on its own merit)

> `kotlinx.coroutines` ensures that every [exception][Throwable] thrown in a coroutine is reacted to by some code. This helps to draw attention to failures and notice when the program enters an incorrect state.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
murfel: If this is our idiomatic rethrowing, shall we use the same in our tutorial? Instead of `throw e`.

https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancel-coroutines

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
murfel: What about this? This uses "exception propagation paths" as a level of abstraction. Even though you don't know if this applies to you or not if you don't know what is "exception propagation paths", you get the big picture, and now you only need to learn if your current situation has a propagation paths already or not.

```
 * An optional element in the [CoroutineContext] to handle coroutine exceptions without a clear propagation path.
 *
 * Code should normally be structured in a way that ensures the existence 
 * of exception propagation paths.
 * For example, a [coroutineScope] call will rethrow an exception to the caller.
 * In rare scenarios when there is no exception propagation path,
 * a [CoroutineExceptionHandler] should be used to process exceptions.
 * In such cases, not using a [CoroutineExceptionHandler] is considered a programming error
 * and will invoke _last-resort exception handling_ which may crash the program.
 *
 * Here, the coroutine created by `launch` fails with the exception `"Failure"`
 * but [supervisorScope] does not react to exceptions from its children,
 * as opposed to [coroutineScope].
 *
 * ```
 * supervisorScope {
 *     launch { error("Failure") }
 * }
 * ```
 *
 * In such cases, a [CoroutineExceptionHandler] should be used to process the exceptions:
 *
 * ```
 * withContext(CoroutineExceptionHandler { ctx, ex ->
 *     println("Exception $ex thrown from coroutine context $ctx")
 * }) {
 *     supervisorScope {
 *         launch { error("Failure") }
 *     }
 * }
 * ```
 *
 * Read below on general philosophy, exception propagation paths, and _last-resort exception handling_.
 ```

Comment (murfel) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: In fact, it is not re-ordering of sections, it is a replacement for the intro section. The current intro section is hard to digest and get a big picture exactly because it already dives into details of a propagation path for coroutineScope + launch.

I'd say it's more important to give a cleaner big picture without one chunk ("exception propagation paths") then to give a harder-to-digest big picture with a vague idea of that chunk.

File: kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt (hunk 3 above)
Comment (dkhalanskyjb) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: The problem isn't the exception hygiene and error handling best practices, though. It's that asynchronous computations don't have an inherent exception propagation path, the way exceptions can bubble up to the `public static void main` and crash the program if you leave them untouched on their way up.

A more disciplined, math-book-like way to structure this piece of documentation would be to extract a separate section called something like "Error handling in concurrent code" and explain in general terms that for queue-based-executor-based concurrency, some exceptions will inherently have nowhere to go. I was hoping to avoid this heavy methodical approach in favor of a specific example of an "intuitively" bad behavior the existence of `CoroutineExceptionHandler` aims to address.

> Let's relax the meaning of "handle" into "reacted to"

We can't get rid of "handling" as a word entirely, since the interface is called `CoroutineExceptionHandler`. Might as well lean into it.

File: kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt (hunk 5 above)
Comment (dkhalanskyjb) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: The idea behind the `coroutineScope { launch { } }` detour is that the reader will look at it, recognize their code is not affected by any problems `CoroutineExceptionHandler` is designed to solve, and move on.

In your proposed alternative, we risk involving the reader that may not at all be interested in the contents—there is no "happy fast path".

```
* Code should normally be structured in a way that ensures the existence of exception propagation paths.
* In rare scenarios when there is no exception propagation path,
* a [CoroutineExceptionHandler] should be used to process exceptions.
```

The reaction I expect: "Okay, so it turns out, I must structure my code somehow. I guess I'll need to go figure out how to do that. Why am I only learning about this here? The coroutine tutorial didn't mention anything about needing to structure my code in some specific way" Then, the reader ends up in the technical deep-dive into the propagation paths, which, ideally, they shouldn't even need. It takes effort to learn the details and then recognize that unknowingly, you've already been doing everything right.

Comment (murfel) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: We can plug the `coroutineScope` back in:

```
 * An optional element in the [CoroutineContext] to handle coroutine exceptions 
 * without a clear propagation path.
 *
 * Code should normally be structured in a way that ensures the existence 
 * of exception propagation paths.
 * For example, a [coroutineScope] call will rethrow an exception to the caller.
```

Edited my suggested snippet above to include this, and some other minor wordings.

The difference with the committed version is frontloading the form of the text.

File: kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt (hunk 3 above)
Comment (murfel) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: You want to communicate (a 10% of):
1. In an async framework, exception propagation is not as straightforward as in a regular program. 
2. We can't truly **handle** exceptions without a propagation path, we can just react to them.

Just FYI, this quote does neither for me, doesn't even clue at these.
> ... instead of being forgotten about. This helps failures not to go unnoticed, with the program silently entering an incorrect state.

I understand how you went from 1+2 to this by simplifying, but this turned out to be a very erasing transformation.

And for what it is now, it looks just like a general discussion on exceptions.

Comment (dkhalanskyjb) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: 1. I don't want this couple of sentences to communicate that, no. The purpose is to lead up to it. The example of `supervisorScope` and `launch` is supposed to demonstrate the problem in specific terms.
2. Likewise, I don't intend to make this contradiction. The idea is to clarify what exactly `kotlinx.coroutines` seeks to do, and that is that exceptions don't get completely forgotten about.

The ideas that asynchronously thrown exceptions are problematic + not being able to fully handle exceptions are the underlying reasons for this piece of text, not its message.

File: kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt (hunk 5 above)
Comment (dkhalanskyjb) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt]: The outline you provide already reads too abstract to me, I'm wary of losing the reader in these general words. Attempting to pad out the outline to a full text results in a yet longer span of technical information without anything to ground the reader using the concept they may be familiar with.

File: kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt (hunk 7 above)
Comment (dkhalanskyjb) [reply kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt:218]: It's complicated: https://github.com/Kotlin/kotlinx.coroutines/issues/1814#issuecomment-1943224856
In the specific example in the tutorial, anything goes; here, `ensureActive()` is needed; in general, there is no one-size-fits-all solution.

File: kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt [hunk 8]
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
     }
 
 /**
- * An optional element in the coroutine context to handle **uncaught** exceptions.
+ * An optional element in the [CoroutineContext] to handle coroutine exceptions without a clear propagation path.
+ *
+ * This interface is part of the overall strategy through which `kotlinx.coroutines` ensures [exceptions][Throwable]
+ * don't go unnoticed.
+ *
+ * In most scenarios, a clear exception propagation path for processing failures in coroutines does exist.
```
murfel: "there exists"

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ * The only exceptions that need to be propagated at all are those with which coroutines finish.
+ * If an exception is handled via a `try`/`catch` block inside the coroutine itself,
+ * the coroutine machinery will not even learn about it:
+ *
+ * ```
+ * launch {
+ *     try {
+ *         throw IllegalStateException("""
+ *             This is not an unpropagated exception,
+ *             since it gets caught.
```
murfel: Too many negations make it hard to parse. Maybe something like "This exception does not need to be propagated because it gets caught", to have the same form as the examples below ("This exception is propagated by...").

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ *     """)
+ * }
+ * ```
+ *
+ * An exception is considered to be propagated if it's transferred to the parent
+ * through [structured concurrency][CoroutineScope]:
+ *
+ * ```
+ * coroutineScope {
+ *     launch {
```
murfel, line 137: Consider something like `scope.launch { ... }`, so that it's not lexically scoped and does not interfere with the previous example.

murfel: Although I see that it's a bit tricky that it requires asserting that the `scope` must have a CEH installed.

murfel: Then maybe introduce the caller for the line 124, something like a function where `coroutineScope` is called.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ * // This function will throw an `IllegalStateException`
+ * coroutineScope {
+ *     throw IllegalStateException("""
+ *         This exception is propagated
+ *         by being rethrown to the caller.
+ *     """)
+ * }
+ * ```
+ *
+ * An exception is considered to be propagated if it's transferred to the parent
```
murfel, line 132: Simplify and use a single form for all "propagated" mentions in all three examples, or at least this one and the next one: "an exception is propagated if ..."

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ *     throw IllegalStateException("""
+ *         This exception is propagated,
+ *         since calling `await()` on `deferred`
+ *         will rethrow the exception.
+ *     """)
+ * }
+ * ```
  *
- * ### Handling coroutine exceptions
+ * When none of the propagation paths listed above apply, an exception can not be propagated.
```
murfel: "cannot"

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
  *
- * `CoroutineExceptionHandler` is a last-resort mechanism for global "catch all" behavior.
- * You cannot recover from the exception in the `CoroutineExceptionHandler`. The coroutine had already completed
- * with the corresponding exception when the handler is called. Normally, the handler is used to
- * log the exception, show some kind of error message, terminate, and/or restart the application.
+ * ```
+ * supervisorScope {
+ *     launch {
+ *         throw IllegalStateException("""
+ *             This *is* an exception with no propagation path, since
```
murfel: The stressed "is" was unexpected, it is unclear what it is contrasted with (there was no "is not" to contrast it with). I'd stress the **no propagation path** bit. To be consistent with previous examples, the text could be "This exception has **no propagation path**".

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ * ### Platform-specific last-resort handling of lost exceptions
+ *
+ * When no [CoroutineExceptionHandler] is present in the [CoroutineContext] of the failing coroutine,
+ * an exception with no propagation path is handled in the following way as the last-resort measure:
+ *
+ * - On JVM, all instances of [CoroutineExceptionHandler] found via `ServiceLoader` and
+ *   the current thread's `Thread.uncaughtExceptionHandler` are invoked.
+ * - On Native, the whole application crashes with the exception.
+ * - On JS and Wasm JS, the exception is propagated into the JavaScript runtime's event loop
+ *   and is processed in a platform-specific way determined by the platform itself.
```
murfel: We can now say that on web, `reportError` is property-tested and invoked, if exists. Otherwise it is rethrown.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ *
```
murfel: Is Wasm/Wasi omission intentional?

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ *         error("Error")
+ *     }
+ * }
+ * ```
+ *
+ * It will print `Outer`, even though the coroutine where the failure originally happened specifies its own
+ * [CoroutineExceptionHandler].
+ * The explanation is that initially, the exception *does* have a propagation path and will get propagated to the
+ * parent coroutine using structured concurrency.
+ * The parent itself, however, has no viable propagation path and has to use *its* [CoroutineExceptionHandler].
```
murfel: **its own**

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ *
+ * Similarly, this [CoroutineExceptionHandler] is redundant and will never be invoked:
+ *
+ * ```
+ * GlobalScope.async(CoroutineExceptionHandler { ctx, e -> println("Failure") }) {
```
murfel: Failure -> Won't ever be invoke, to be consistent with other examples

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
     /**
-     * Key for [CoroutineExceptionHandler] instance in the coroutine context.
+     * Key for the [CoroutineExceptionHandler] instance in a coroutine context.
      */
     public companion object Key : CoroutineContext.Key<CoroutineExceptionHandler>
 
     /**
-     * Handles uncaught [exception] in the given [context]. It is invoked
-     * if the coroutine has an uncaught exception.
+     * Handles an [exception] in the given [context].
```
murfel: "**in** the given context" can be misinterpreted as if the handler is called from within the provided context.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ * coroutineScope {
+ *     launch(CoroutineExceptionHandler { ctx, ex ->
+ *         println("This line will not be printed!")
+ *     }) {
+ *         error("This failure will cancel `coroutineScope`!")
+ *     }
+ * }
+ * ```
+ *
+ * A [CoroutineExceptionHandler] has no effect in this scenario: the failure in [launch] will cancel [coroutineScope].
```
murfel: Spell out the rules according to which this example works:
> `launch` has a propagation path, so the error is propagated to the parent, and the CEH will never be invoked

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
+ * Using `try`/`catch` is the proper way to achieve this.
```
murfel: Spell out "this" to avoid accidentally referring to the previous sentence and specify where the try-catch should go (there are two options: inside and outside launch).

> To avoid the parent coroutine from being cancelled from a launch exception, wrap the call to launch into a try-catch (and also possibly reconsider your design?)

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
```diff
@@ -40,70 +40,257 @@ internal fun handlerException(originalException: Throwable, thrownException: Thr
...
  * ```
  *
- * ### Uncaught exceptions with no handler
+ * Alternatively, whenever a failure is indeed supposed to terminate a coroutine,
+ * using [async] instead of [launch] and later calling [Deferred.await] instead of [Job.join]
+ * on its result to check if the computation was successful will allow gracefully processing the exception.
+ *
+ * ### Pitfalls
+ *
+ * #### Using a [CoroutineExceptionHandler] does not prevent coroutine failures
```
murfel, line 227: Something is confusing me in this example, but I can't quite pinpoint it. I'll take a look at it again after the next set of changes.

murfel: Possibly, it's that there are a few mistakes/misconceptions combined together.
* A CEH doesn't prevent a coroutine from failing.
* A CEH is not called if there exists a propagation path.
* And possibly something else?

murfel:
This version reads much lighter and less technical, although I am already tainted and cannot make a fair judgement.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: Nope, good catch!

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: Your analysis is spot-on. Indeed, to make a mistake shown in this section, you need to be completely unaware of what coroutine exception handlers do. We can expect people to be that unaware, though! Context: https://github.com/Kotlin/kotlinx.coroutines/pull/1886#discussion_r400394293
Also, believing `CoroutineExceptionHandler` behaves as `catch` does isn't even unreasonable. In languages with effect handlers, you can redefine the behavior of `throw`. Even absent that, it's quite natural to believe an exception handler will at least prevent a coroutine from cancelling its parent: after all, the exception was "handled".

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: > it requires asserting that the scope must have a CEH installed.

This particular example requires asserting that the parent job is a normal one and not a `SupervisorJob`. Manually constructing such a scope would detract from the example and highlights `CoroutineScope()`, whose usage we want to discourage. In any case, it's unclear to me what issue you're describing. The example is the simplest one that illustrates children propagating exceptions to their parents.

kotlinx-coroutines-core/common/src/CoroutineExceptionHandler.kt
dkhalanskyjb: I don't see the value in it, and there are slight semantic differences between the three cases.
* In the first scenario, rethrowing is *the* direct act of propagation.
* In the second scenario, being transferred is like a checkmark on the propagation sheet: it may not be the only one, or even the most important act of propagation, but it suffices.
* In the third scenario, the exception isn't meaningfully propagated in the program-logic sense. Instead, `kotlinx.coroutines` considers its propagation job done if the exception can in theory be queried afterwards somehow.

Issue #4600, opened 2026-01-09
Easier construction of MDCContext


schielek:
**Use case**

Using the MDCContext of the SLF4J extension is cumbersome. It is recommended to use it in the following way:

```kotlin
MDC.put("a", "1")
MDC.put("b", "2")
MDC.put("c", "3")
withContext(MDCContext()) {
    // ...
}
```

The code has two problems:
1. Many lines required for something that could fit on one line
2. The MDC values are leaking into code after the `withContext` block. On backend server applications where threads are reused they could leak into other HTTP calls.

Cleaning up the MDC adds even more boilerplate code:

```kotlin
val closables = listOf(
    MDC.putCloseable("a", "1"),
    MDC.putCloseable("b", "2"),
    MDC.putCloseable("c", "3"),
)
try {
    withContext(MDCContext()) {
        // ...
    }
} finally {
    closables.forEach { it.close() }
}
```

**The Shape of the API**

It would be nice to have a secondary constructor, which takes key values pairs as arguments so I can just write:
```kotlin
withContext(MDCContext("a" to "1", "b" to "2", "c" to "3")) {
    // ...
}
```

As a current workaround I use the following function, but I think a secondary constructor as part of the library would be better:

```kotlin
fun MDCContext(vararg keyValuePairs: Pair<String, String>) = MDCContext(
    (MDC.getCopyOfContextMap() ?: emptyMap()) + keyValuePairs,
)
```

This offers a concise way of setting up values in the MDC without polluting the current thread. Also, it behaves better than the example from above: The MDC is not only cleaned up, but rather restored to its previous state after the `withContext` block.

You can find my PR [here](https://github.com/Kotlin/kotlinx.coroutines/pull/4601).

dkhalanskyjb:
The expected form of what you're trying to achieve is:
```kotlin
withContext(MDCContext(MDC.getCopyOfContextMap() + listOf("a" to "1", "b" to "2", "c" to "3"))) {
}
```
I completely agree that this is quite cumbersome, and we could do better.

The specific API shape that you propose isn't something we'd like to introduce, though. I'll describe the problems and potential solutions in more details as comments under your PR.

vladimirdolzhenko:
Please ping https://youtrack.jetbrains.com/issue/KTIJ-38392/Inspection-to-use-MDC-properly-in-suspend-context when the ticket is implemented

PR #4601, opened 2026-01-09
Add secondary MDCContext constructor for convenience

schielek:
See https://github.com/Kotlin/kotlinx.coroutines/issues/4600

### DIFF
--- a/integration/kotlinx-coroutines-slf4j/test/MDCContextTest.kt
+++ b/integration/kotlinx-coroutines-slf4j/test/MDCContextTest.kt
@@ -177,8 +177,8 @@ class MDCContextTest : TestBase() {
     fun testPrimaryAndSecondaryConstructorHaveConsistentOutcomeForNoInputParametersAndEmptyMDC() = runTest {
         listOf(
             { MDCContext() },
-            { MDCContext(*arrayOf())},
-        ).forEach{ constructor ->
+            { MDCContext(*arrayOf()) },
+        ).forEach { constructor ->
             withContext(constructor()) {
                 val copyOfContextMap = MDC.getCopyOfContextMap()
                 assertNull(copyOfContextMap)
@@ -191,8 +191,8 @@ class MDCContextTest : TestBase() {
         MDC.put("a", "1")
         listOf(
             { MDCContext() },
-            { MDCContext(*arrayOf())},
-        ).forEach{ constructor ->
+            { MDCContext(*arrayOf()) },
+        ).forEach { constructor ->
             withContext(constructor()) {
                 val copyOfContextMap = MDC.getCopyOfContextMap()
                 assertNotNull(copyOfContextMap)


integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
```diff
@@ -86,6 +94,31 @@ public class MDCContext(
...
+     * withContext(MDCContext("kotlin" to "rocks")) {
+     *     println(MDC.get("kotlin")) // prints "rocks"
+     * }
+     * println(MDC.get("kotlin")) // prints "is great"
+     * ```
+     *
+     * @param keyValuePairs key-value-pairs with will be put into the MDC
+     */
+    public constructor(vararg keyValuePairs: Pair<String, String>): this(
+        (MDC.getCopyOfContextMap() ?: emptyMap()) + keyValuePairs,
```
dkhalanskyjb: It's inconsistent to obtain the caller's context map here. The primary constructor takes the context map as a parameter and only has `getCopyOfContextMap` as the default.

Another inconsistency is that passing nothing at all into this constructor when the context is empty results in an empty map, but in the other constructor, it results in the context not being set at all. Compare:
```kotlin
// primary constructor
withContext(MDCContext()) {
    println(MDC.getCopyOfContextMap()) // null
}
// passing nothing to the secondary constructor
withContext(MDCContext(*arrayOf())) {
    println(MDC.getCopyOfContextMap()) // {}
}
```

In general, it shouldn't matter which of the two constructors gets chosen in an `MDCContext()` call with no parameters.

A solution to this could be to introduce a single overload that has both the `contextMap` parameter and the `vararg` arguments. That overload can check if there are no `vararg`s passed + the `contextMap` is `null`, and if so, keep the context map passed to `setCurrent` `null` as well.

All of these corner cases need to be tested and documented as well.

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
```diff
@@ -86,6 +94,31 @@ public class MDCContext(
...
+     * MDC.put("kotlin", "is great")
+     * withContext(MDCContext("kotlin" to "rocks")) {
+     *     println(MDC.get("kotlin")) // prints "rocks"
+     * }
+     * println(MDC.get("kotlin")) // prints "is great"
+     * ```
+     *
+     * @param keyValuePairs key-value-pairs with will be put into the MDC
+     */
+    public constructor(vararg keyValuePairs: Pair<String, String>): this(
```
dkhalanskyjb: The consistent name for the parameter is `pairs`: https://github.com/JetBrains/kotlin/blob/d75477d5337719cc9096feef6b6f703fa0d0558f/libraries/stdlib/src/kotlin/collections/Maps.kt#L670

schielek: done

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
schielek: Good catch. I added tests to ensure consistency.

> introduce a single overload

Can you elaborate what you mean by single overload. I am considering the following options:
1. Having only the primary constructor with additional varargs. (This is probably a bytecode breaking change)
2. Having the the secondary constructor take a contextMap parameter as well. (This cannot be resolved by the compiler unless first argument is specified or named arguments are use)

edit: if the arguments are swapped, pairs first and then contextMap, it can be resolved
edit: changing the order makes the api awkward since in case of conflicting entries I would expect the second parameter to overwrite the first and not the other way around.

schielek: > A solution to this could be to introduce a single overload that has both the contextMap parameter and the vararg arguments. That overload can check if there are no varargs passed + the contextMap is null, and if so, keep the context map passed to setCurrent null as well.

I wonder if passing both, a context map and key-value-pairs, is even a valid use case.
```kotlin
MDCContext(mapOf("a" to "1", "b" to "2"), "c" to "3", "d" to "4")
```
is no different than writing
```kotlin
MDCContext(mapOf("a" to "1", "b" to "2") + listOf("c" to "3", "d" to "4"))
```

dkhalanskyjb: > I wonder if passing both, a context map and key-value-pairs, is even a valid use case.

I also can't think of a realistic piece of code that would need this. Maybe something like `MDCContext(ComponentSpecificMdcContext, "extra" to "value")`? Still, it seemingly doesn't cost anything to add the extra flexibility, which also comes with the extra explicitness of having the `getCopyOfContextMap` call right in the signature.

> Can you elaborate what you mean by single overload. I am considering the following options:

Both options are fine by me!

> Having only the primary constructor with additional varargs. (This is probably a bytecode breaking change)

Yes, this would be a bytecode-breaking change if done naively. From `integration/kotlinx-coroutines-slf4j/api/kotlinx-coroutines-slf4j.api`:

```
public final class kotlinx/coroutines/slf4j/MDCContext : kotlin/coroutines/AbstractCoroutineContextElement, kotlinx/coroutines/ThreadContextElement {
        public fun <init> ()V
        public fun <init> (Ljava/util/Map;)V
```
The constructor is part of the binary signature.

The way to go would be to preserve the current constructor as well, with a `@Deprecated(level = DeprecationLevel.HIDDEN)` annotation.

> Having the the secondary constructor take a contextMap parameter as well. (This cannot be resolved by the compiler unless first argument is specified or named arguments are use)

I expect that this could also be made to work, but would require some compiler-tricking magic. The overload with `vararg` would need to be marked with
```kotlin
@Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
```

PR #4603, opened 2026-01-13
Publish `kotlinx-coroutines` for integration testing in `build/`

dkhalanskyjb:
Before this change, the tests in `integration-testing` were expecting the local Maven repository to contain the artifacts from `kotlinx-coroutines`.
The full build process looked like this:
```sh
./gradlew clean build publishToMavenLocal
cd integration-testing
./gradlew clean build
```
This was problematic:
* Builds could not be run in parallel, as they would interfere with one another via the local Maven repository.
* Moreover, even sequential builds could suffer due to not every file in the local Maven repository being overwritten.

With this change, the build process is:
```sh
./gradlew clean build publishAllPublicationsToIntegrationTestingRepository
cd integration-testing
./gradlew clean build
```

### DIFF
--- a/buildSrc/src/main/kotlin/pub-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/pub-conventions.gradle.kts
@@ -16,8 +16,8 @@ publishing {
     repositories {
         configureMavenPublication(this, project)
         maven {
-            name = "IntegrationTesting"
-            url = uri(project.rootProject.layout.buildDirectory.dir("integration-testing-repository"))
+            name = "BuildLocal"
+            url = uri(project.rootProject.layout.buildDirectory.dir("build-local-repository"))
         }
     }
 
@@ -73,5 +73,5 @@ tasks.register("bintrayUpload") { dependsOn(tasks.matching { it.name == "publish
 
 // Compatibility with old TeamCity configurations that perform `publishToMavenLocal`
 tasks.named("publishToMavenLocal") {
-    dependsOn(tasks.matching { it.name == "publishAllPublicationsToIntegrationTestingRepository" })
+    dependsOn(tasks.matching { it.name == "publishAllPublicationsToBuildLocalRepository" })
 }
--- a/integration-testing/README.md
+++ b/integration-testing/README.md
@@ -14,4 +14,4 @@ The tests are the following:
 
 The `integration-testing` project is expected to be in a subdirectory of the main `kotlinx.coroutines` project.
 
-To run all the available tests: `./gradlew publishAllPublicationsToIntegrationTestingRepository` + `cd integration-testing` + `./gradlew check`.
+To run all the available tests: `./gradlew publishAllPublicationsToBuildLocalRepository` + `cd integration-testing` + `./gradlew check`.
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -73,7 +73,7 @@ allprojects {
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
         maven("https://redirector.kotlinlang.org/maven/dev")
         mavenLocal()
-        maven(rootProject.layout.projectDirectory.dir("../build/integration-testing-repository/"))
+        maven(rootProject.layout.projectDirectory.dir("../build/build-local-repository/"))
     }
 }
 
--- a/KOTLIN_UPGRADE.md
+++ b/KOTLIN_UPGRADE.md
@@ -23,7 +23,7 @@
    Set the compiler version of Kotlin Playground to the same major version as `<version>`,
    click "Copy link", and replace the original link in `README.md` with the new one.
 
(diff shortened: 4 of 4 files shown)


buildSrc/src/main/kotlin/pub-conventions.gradle.kts
```diff
@@ -15,6 +15,10 @@ apply(plugin = "signing")
 publishing {
     repositories {
         configureMavenPublication(this, project)
+        maven {
+            name = "IntegrationTesting"
```
fzhinkin: We can also use this repository for pre-publication artifacts validation and to gather build results from separate steps in CI.

So maybe a more neutral name could considered for this repository? Something like, `BuildLocal`.

integration-testing/README.md
```diff
@@ -14,4 +14,4 @@ The tests are the following:
 
 The `integration-testing` project is expected to be in a subdirectory of the main `kotlinx.coroutines` project.
 
-To run all the available tests: `./gradlew publishToMavenLocal` + `cd integration-testing` + `./gradlew check`.
+To run all the available tests: `./gradlew publishAllPublicationsToIntegrationTestingRepository` + `cd integration-testing` + `./gradlew check`.
```
murfel: These commands don't need `clean`?

buildSrc/src/main/kotlin/pub-conventions.gradle.kts
```diff
@@ -66,3 +70,8 @@ tasks.matching { it.name == "generatePomFileForKotlinMultiplatformPublication" }
 
 // Compatibility with old TeamCity configurations that perform :kotlinx-coroutines-core:bintrayUpload
 tasks.register("bintrayUpload") { dependsOn(tasks.matching { it.name == "publish" }) }
+
+// Compatibility with old TeamCity configurations that perform `publishToMavenLocal`
+tasks.named("publishToMavenLocal") {
+    dependsOn(tasks.matching { it.name == "publishAllPublicationsToIntegrationTestingRepository" })
+}
```
murfel, line 77: Should / could we update the TeamCity config?

murfel:
I don't exactly understand the workflow, so a few questions to clear up my confusion.

> Builds could not be run in parallel, as they would interfere with one another via the local Maven repository.

Which builds of what / how many of them are you referring to?

> Moreover, even sequential builds could suffer due to not every file in the local Maven repository being overwritten.

If you do two sequential builds for the new integration-testing repository, wouldn't they suffer from the same problem?

dkhalanskyjb:
> Which builds of what / how many of them are you referring to?

Scenario:
* You work on `branch-a` in `kotlinx.coroutines`.
* After making some changes, you want to make sure it builds. You start the full build by calling `./gradlew clean build publishToMavenLocal; cd integration-testing; ./gradlew clean check`.
* While the build runs, to avoid downtime, you open another copy of `kotlinx.coroutines` on `branch-b`.
* You start another build.

If `./gradlew clean build publishToMavenLocal` succeeds for both builds, there's a race condition: which version is going to be checked by integration tests? Maybe both will be checked, or maybe just one, the one that happened to finish last, or maybe, the published artifacts will be mixed together arbitrarily.

The more builds you run on parallel on the same computer, the more pronounced the issue. The local Maven repository becomes shared mutable state.

> If you do two sequential builds for the new integration-testing repository, wouldn't they suffer from the same problem?

Not if `clean` is called, purging the new repository, which is entirely contained in the `kotlinx.coroutines` root and is therefore not shared between several checked out copies.

buildSrc/src/main/kotlin/pub-conventions.gradle.kts
dkhalanskyjb: Yes, after merging this PR and waiting for some time to make sure all active branches include this commit.

integration-testing/README.md
dkhalanskyjb: It's not absolutely required. In most cases, you can do without it and save yourself some time spent on recompilation. If there are drastic changes like removing a Gradle module or updating the Kotlin version, yes, you need a `clean`. It also doesn't hurt to call `clean` before reporting a compiler bug when something crashes, because often, the reason is some sort of stale state.

murfel:
Got it, it's to isolate separate branches.

integration-testing/build.gradle.kts
```diff
@@ -73,6 +73,7 @@ allprojects {
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
         maven("https://redirector.kotlinlang.org/maven/dev")
         mavenLocal()
+        maven(rootProject.layout.projectDirectory.dir("../build/build-local-repository/"))
```
murfel, line 76: But all the other projects like `.../java8/...` with a similar setup disappeared, is it intentional?

dkhalanskyjb: Yes, it's a result of the rebase on top of https://github.com/Kotlin/kotlinx.coroutines/commit/63472f5e903de83c979080474e15fb7749a19a45

PR #4604, opened 2026-01-15
Remove the deprecated `kotlinx-coroutines-test` APIs

dkhalanskyjb:
* Remove the APIs that were deprecated with an `ERROR`.
* Advance the deprecations for `runTest` overloads accepting the dispatch timeout to `ERROR`.
* Reword the documentation so that it no longer references the removed APIs as a point of comparison.
* Remove the migration guide from the obsolete APIs.
* Remove the code that was only used by the now-removed APIs and simplify some of the remaining code.

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
+++ b/kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
@@ -172,7 +172,7 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
     }
 
     /**
-     * Checks that the only tasks remaining in the scheduler are cancelled.
+     * Checks that no tasks remain in the event queue.
      */
     internal fun isIdle(): Boolean =
         synchronized(lock) { events.isEmpty }


kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
```diff
@@ -185,10 +174,8 @@ public class TestCoroutineScheduler : AbstractCoroutineContextElement(TestCorout
     /**
      * Checks that the only tasks remaining in the scheduler are cancelled.
      */
-    internal fun isIdle(strict: Boolean = true): Boolean =
-        synchronized(lock) {
-            if (strict) events.isEmpty else events.none { !it.isCancelled() }
-        }
+    internal fun isIdle(): Boolean =
+        synchronized(lock) { events.isEmpty }
```
murfel, line 178: The kdoc also needs a rewording?

murfel, line 178: Also doesn't need the `synchronized` now, since `events` is thread-safe by itself, and, in particular, isEmpty/size is atomic.

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -22,19 +22,6 @@ import kotlin.time.*
  *
  * The usual way to access a [TestScope] is to call [runTest], but it can also be constructed manually, in order to
  * use it to initialize the components that participate in the test.
- *
- * #### Differences from the deprecated [TestCoroutineScope]
```
murfel, line 26: Is it our general policy to remove docs on the deprecated stuff when it is promoted to Error? I can imagine how some projects postpone updating usages while it is in Warning, and are only update once they are forced to do it when it is promoted to Error. It is unfortunate that by that point all the useful details are gone from the kdocs!

I also understand that keeping details on APIs that no longer compile is also undesirable. Just wondering what's the user journey of a procrastinating project would be.

kotlinx-coroutines-test/common/src/TestScope.kt
```diff
@@ -143,22 +130,16 @@ public val TestScope.testTimeSource: TimeSource.WithComparableMarks get() = test
...
- * - A [CoroutineExceptionHandler] is created that makes [TestCoroutineScope.cleanupTestCoroutines] throw if there were
- *   any uncaught exceptions, or forwards the exceptions further in a platform-specific manner if the cleanup was
- *   already performed when an exception happened. Passing a [CoroutineExceptionHandler] is illegal, unless it's an
- *   [UncaughtExceptionCaptor], in which case the behavior is preserved for the time being for backward compatibility.
- *   If you need to have a specific [CoroutineExceptionHandler], please pass it to [launch] on an already-created
- *   [TestCoroutineScope] and share your use case at
- *   [our issue tracker](https://github.com/Kotlin/kotlinx.coroutines/issues).
+ * - A [CoroutineExceptionHandler] is created that makes [runTest] throw after the test is finished if there were
+ *   any uncaught exceptions, or forwards the exceptions further in a platform-specific manner if by the time an
+ *   exception happened, the test has completed. Passing a [CoroutineExceptionHandler] in [context] is unsupported.
```
murfel, line 135: This info is removed from the new kdoc but is kept in the implementation. Is this intentional?
> If you need to have a specific [CoroutineExceptionHandler], please pass it to [launch] on an already-created [TestCoroutineScope] and share your use case at [our issue tracker]

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: > Is it our general policy to remove docs on the deprecated stuff when it is promoted to Error?

I think the docs should get removed when the API is removed/hidden. If you encounter an error, you may be forced to deal with it after having ignored the warnings, and that means finally learning what the API you used to use does.

`TestCoroutineScope` was deprecated with `ERROR` in 1.9.0 (two major releases ago, in 2024), and this PR removes it entirely. The warnings have been there for five years.

kotlinx-coroutines-test/common/src/TestScope.kt
dkhalanskyjb: Yeah, the notice has been there since the initial introduction of the API, and no one ever approached us about this.

kotlinx-coroutines-test/common/src/TestCoroutineScheduler.kt
dkhalanskyjb: In general, such an argument doesn't hold: thread safety/atomicity is not enough to remove a mutex. See:
```
synchronized { list.remove(0); list.add('x') }     |     synchronized { list.isEmpty() }
```
If we remove `synchronized` around `isEmpty`, it may start observing an intermediate state where the list is empty, even if logically, the other operation was replacing one value with another.

In this specific case, not synchronizing `isIdle` does feel safe, but I'd rather not dive too deep nor touch it: it's only called from a deprecated-with-error function and will be removed in the next release anyway.

PR #4606, opened 2026-01-24
Port tests from jvm to concurrent: change

murfel:
to be able to compile

```
git cherry-pick -n c70eb28
```


test-utils/native/src/FieldWalker.kt
```diff
@@ -0,0 +1,9 @@
+package kotlinx.coroutines.testing
+
+actual object FieldWalker {
+    actual fun walk(root: Any?): Set<Any> {
+        return emptySet()
+    }
```
dkhalanskyjb: This implementation of `walk` should be removed.
* It is only applicable in cases when the test checks that some reference is *not* there, but the name is misleading the reader into thinking that this could be used, for example, to ensure some reference *is* there. This point could be mitigated by choosing a better name, like `subsetOfReachableReferences`.
* The test codebase is using `walk` only in places where returning an empty set will surely cause the test to fail: "how many references there are" and "check that there is a reference."

test-utils/native/src/ThreadUtils.kt
```diff
@@ -0,0 +1,28 @@
...
+    return ConcurrentThread(name, block).apply { start() }
+}
+
+actual class ConcurrentThread actual constructor(
+    private val name: String?, private val block: () -> Unit
+) {
+    private var worker: Worker? = null
+    private var future: Future<Unit>? = null
+
+    actual fun start() {
```
dkhalanskyjb, line 20: We should guard against calling `start` twice accidentally. That would be nasty to try and debug.

test-utils/jvm/src/ConcurrentThreadUtils.kt
```diff
@@ -0,0 +1,23 @@
...
+    return ConcurrentThread(name, block).apply { start() }
+}
+
+actual class ConcurrentThread actual constructor(
+    private val name: String?, private val block: () -> Unit
+) {
+    private var thread: Thread? = null
+
+    actual fun start() {
+        thread = thread(name = name, block = block)
```
dkhalanskyjb: Here, too, we'd need a check that the thread has started. In `join` as well: it is semantically incorrect to exit `join()` before the thread finished, so we could make some bugs with this implementation.

One approach could be to manually add the guards, but with slight tweak, we can avoid that for the JVM.

In concurrent code:
```kotlin
expect class ConcurrentThread(
    block: Runnable, name: String
) {
    fun start()
    fun join()
}

fun ConcurrentThread(name: String, block: () -> Unit) = ConcurrentThread({ block() }, name)
```

On the JVM:
```kotlin
actual fun runThread(
    name: String?, block: () -> Unit
): ConcurrentThread {
    return thread(name = name, block = block)
}

actual typealias ConcurrentThread = Thread
```

test-utils/concurrent/src/ConcurrentThreadUtils.concurrent.kt
```diff
@@ -0,0 +1,28 @@
...
+        val elapsed = start.elapsedNow()
+        val remaining = duration - elapsed
+        if (remaining <= Duration.ZERO) return
+        ParkingSupport.park(remaining)
+    }
+}
+
+expect fun runThread(
+    name: String? = null, block: () -> Unit
+): ConcurrentThread
```
dkhalanskyjb: Currently unused.

test-utils/concurrent/src/ConcurrentThreadUtils.concurrent.kt
```diff
@@ -0,0 +1,28 @@
...
+
+expect class ConcurrentThread(
```
dkhalanskyjb: The name `ConcurrentThread` makes sense in the context of the `concurrent` source set, but would likely confuse a stray reader: aren't all threads concurrent? `MultiplatformThread` could be an improvement in that regard.

test-utils/concurrent/src/ConcurrentThreadUtils.concurrent.kt
```diff
@@ -0,0 +1,28 @@
+package kotlinx.coroutines.testing
+
+import kotlinx.atomicfu.locks.*
+import kotlin.time.*
+
+@OptIn(ExperimentalThreadBlockingApi::class)
+fun concurrentSleep(duration: Duration) {
```
dkhalanskyjb: Likewise, "concurrent sleep" doesn't make much sense. Just `threadSleep`/`blockingSleep`, maybe?

kotlinx-coroutines-core/native/test/PoolSizes.kt
```diff
@@ -0,0 +1,4 @@
+package kotlinx.coroutines.internal
+
+internal actual val CONCURRENT_CORE_POOL_SIZE: Int = Platform.getAvailableProcessors().coerceAtLeast(2)
+internal actual val CONCURRENT_MAX_POOL_SIZE: Int = 2048
```
dkhalanskyjb, line 7: Currently, these constants are only used in tests. That means, they should either stay stay in the test source set or be used in the dispatcher implementations instead of the hardcoded values.

kotlinx-coroutines-core/jvm/test/FailFastOnStartJvmTest.kt
```diff
@@ -0,0 +1,33 @@
+@file:Suppress("DeferredResultUnused")
```
dkhalanskyjb: This is intended for `async` calls, of which none are left in this file.

kotlinx-coroutines-core/jvm/src/internal/PoolSizes.kt
```diff
@@ -0,0 +1,7 @@
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.scheduling.CORE_POOL_SIZE
+import kotlinx.coroutines.scheduling.MAX_POOL_SIZE
+
+internal actual val CONCURRENT_CORE_POOL_SIZE: Int = CORE_POOL_SIZE
+internal actual val CONCURRENT_MAX_POOL_SIZE: Int = MAX_POOL_SIZE
```
dkhalanskyjb, line 8: If you decide to use the constants meaningfully in the Native source sets instead of moving them to `test/`, please move them to `kotlinx.coroutines.scheduling` (from `kotlinx.coroutines.internal`) so that this unnecessary indirection disappears.

dkhalanskyjb (changes requested):
First batch of comments.

test-utils/concurrent/src/SynchronizedSet.kt
```diff
@@ -0,0 +1,16 @@
+package kotlinx.coroutines.testing
+
+import kotlinx.coroutines.sync.*
+
+class ConcurrentSynchronizedSet {
+    private val elements = LinkedHashSet<Long>()
+    private val mutex = Mutex()
+
+    suspend fun add(value: Long) = mutex.withLock {
```
dkhalanskyjb: The original was not `suspend`, and since suspensions are an observable event, we'd need to carefully check that this change doesn't affect the behavior of the tests. We could do that, but in my view, it's preferable to avoid such deep thought in an extensive, mechanical PR like this.

test-utils/concurrent/src/ConcurrentCyclicBarrier.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.atomicfu.locks.*
+import kotlin.concurrent.atomics.*
+import kotlin.time.*
+
+// Adapted from kotlinx-atomicfu
+// https://github.com/Kotlin/kotlinx-atomicfu/blob/d09c2b07cd16b0b273bd94edaa4929acd2ec42bc/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/CyclicBarrierTest.kt#L59
+
+private class HandleWrapper(val handle: ParkingHandle) {
+    val woken = AtomicBoolean(false)
+}
```
dkhalanskyjb: Minor, style: private helper classes typically go to the bottom of the file.

test-utils/concurrent/src/ConcurrentCyclicBarrier.kt
```diff
@@ -0,0 +1,72 @@
...
+
+class ConcurrentCyclicBarrier(private val parties: Int) {
```
dkhalanskyjb: The name is incorrect (in `kotlinx.atomicfu`, too): this barrier is not cyclic, since `woken` never gets reset. I don't know if we ever use the fact that the Java barrier is indeed cyclic, so we need to either audit the usages and rename the data structure or make the barrier truly cyclic.

kotlinx-coroutines-core/concurrent/test/FailFastOnStartTest.kt
```diff
@@ -2,25 +2,13 @@
...
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.testing.*
+import kotlin.coroutines.*
 import kotlin.test.*
 
 class FailFastOnStartTest : TestBase() {
-
-    @Rule
-    @JvmField
-    val timeout: Timeout = Timeout.seconds(5)
```
dkhalanskyjb, line 22: Please leave a `TODO` comment mentioning that these tests may hang and would benefit from a timeout mechanism.

kotlinx-coroutines-core/concurrent/test/flow/SharingStressTest.kt
```diff
@@ -1,26 +1,28 @@
...
-    @get:Rule
-    val subscriberDispatcher = ExecutorRule(nSubscribers)
+    val emitterDispatcher = newSingleThreadContext("Emitter")
+
+    val subscriberDispatcher = newFixedThreadPoolContext(nSubscribers, "Subscribers")
+
+    @AfterTest
+    fun tearDown() {
+        emitterDispatcher.close()
+        subscriberDispatcher.close()
```
dkhalanskyjb, line 25: Minor, style: by convention, deinitialization should happen in the opposite order from initialization. Here, it's completely unimportant, but it still catches the eye as a potential mistake.

dkhalanskyjb (changes requested):
Another batch of comments. A general concern is that the `Concurrent*` naming used throughout the PR is unfortunate, since that word looks either absurd or tautological, depending on the API. We have a convention to *append* the source set name to the file/class name to avoid linkage errors, so `CountDownLatchConcurrent` that would be consistent with the rest of the project, but the `Multiplatform` prefix makes even more sense to me.

kotlinx-coroutines-core/concurrent/test/JobActivationStressTest.kt
```diff
@@ -1,67 +1,51 @@
...
+                var wasStarted = false
+                val d = async(pool + NonCancellable, start = CoroutineStart.LAZY) {
+                    wasStarted = true
+                    throw TestException()
+                }
+                // need to add on completion handler
+                val causeHolder = CompletableDeferred<Throwable?>()
+                // we use synchronization on causeHolder to work around the fact that completion listeners
+                // are invoked after the job is in the final state, so when "d.join()" completes there is
+                // no guarantee that this listener was already invoked
```
dkhalanskyjb, line 27: The comment is outdated.

dkhalanskyjb (changes requested):

dkhalanskyjb:
On second thought, in the names of the primitives, `Concurrent` doesn't give us any new information. It can just be `CountDownLatch`, for example. Less typing, and we are in any case unlikely to want a `CountDownLatch` in the JS test source set, so we won't try using it there or in common code.

test-utils/concurrent/src/ConcurrentThreadUtils.concurrent.kt
murfel: Removed for now, it's for CallbackFlowTest which is currently excluded because there's no timeouts yet.

kotlinx-coroutines-core/native/test/PoolSizes.kt
murfel: `kotlinx-coroutines-core/concurrent/src/internal/PoolSizes.concurrent.kt`
> // Used for testing only, but is in src/internal due to JVM implementation relying on internal constants from src.

test-utils/concurrent/src/SynchronizedSet.kt
murfel: Opted in to call `runBlocking` at `ConcurrentSynchronized.add` usage site, instead of changing the data structure. If someone wants to use the data structure for convenience later, they might be surprised that it's blocking.

kotlinx-coroutines-core/native/test/PoolSizes.kt
dkhalanskyjb: I have seen that comment, but it doesn't answer my concern.

kotlinx-coroutines-core/concurrent/test/JobActivationStressTest.kt
murfel: It's not, we still need to point out why causeHolder is needed. Reworded.

kotlinx-coroutines-core/native/test/PoolSizes.kt
murfel: Why?

murfel: Moved to test/

kotlinx-coroutines-core/concurrent/test/FailFastOnStartTest.kt
murfel: Removed the test itself. Accidentally included again, tests with timeouts are not meant to be touched yet.

test-utils/concurrent/src/ConcurrentThreadUtils.concurrent.kt
murfel: **Concurrent naming thread**

> The name ConcurrentThread makes sense in the context of the concurrent source set, but would likely confuse a stray reader: aren't all threads concurrent? MultiplatformThread could be an improvement in that regard.

Multiplatform prefix is also confusing since it's not multiplatform in terms of the source set.

> A general concern is that the Concurrent* naming used throughout the PR is unfortunate, since that word looks either absurd or tautological, depending on the API. We have a convention to append the source set name to the file/class name to avoid linkage errors, so CountDownLatchConcurrent that would be consistent with the rest of the project, but the Multiplatform prefix makes even more sense to me.

> On second thought, in the names of the primitives, Concurrent doesn't give us any new information. It can just be CountDownLatch, for example. Less typing, and we are in any case unlikely to want a CountDownLatch in the JS test source set, so we won't try using it there or in common code.

I could do that but then in every JVM test which has `import kotlinx.coroutines.testing.*`, I'll have to import the java's primitive explicitly, sometimes with the wildcard import.
```
import java.util.concurrent.CyclicBarrier/CountDownLatch
import java.util.concurrent.*
```

Shall I append the `Concurrent` as a suffix to everything?

No strong preference on my side, though.

dkhalanskyjb: > it's not multiplatform in terms of the source set

It is, it's more than one platform.

> I'll have to import the java's primitive explicitly

Why would we still need Java's primitives if we had our own ones?

murfel: Used MultiplatformThread and lost the Concurrent prefix everywhere

test-utils/concurrent/src/ConcurrentCyclicBarrier.kt
murfel: Please check everything but the barrier implementation and usage. Let me think about the barrier a bit more.

kotlinx-coroutines-core/concurrent/test/flow/SharingStressTest.kt
murfel, line 34: This test is flaky. Failed once locally on native. Cannot reproduce.

kotlinx-coroutines-core/jvm/test/scheduling/DefaultDispatchersTest.kt
```diff
@@ -1,9 +1,12 @@
 package kotlinx.coroutines.scheduling
 
 import kotlinx.coroutines.testing.*
+// TODO: implement CyclicBarrier.reset (in this PR)
+//import kotlinx.coroutines.testing.CyclicBarrier
+import kotlinx.coroutines.testing.CountDownLatch
 import kotlinx.coroutines.*
 import org.junit.Test
-import java.util.concurrent.*
+import java.util.concurrent.CyclicBarrier
```
murfel: This file needs a CyclicBarrier.reset (+ a few of the moved jvm->concurrent need reset(), too, which I will implement and put them back)

kotlinx-coroutines-core/jvm/test/FailingCoroutinesMachineryTest.kt
```diff
@@ -1,11 +1,14 @@
...
 
 import kotlinx.coroutines.testing.*
+// TODO: implement CountDownLatch.await with timeout (in this PR)
+//import kotlinx.coroutines.testing.CountDownLatch
 import org.junit.*
 import org.junit.Test
 import org.junit.runner.*
 import org.junit.runners.*
 import java.util.concurrent.*
+import java.util.concurrent.CountDownLatch
```
murfel: This file needs `CountDownLatch.await` with a timeout, will implement and replace

kotlinx-coroutines-core/jvm/test/jdk8/future/FutureTest.kt
```diff
@@ -190,12 +192,11 @@ class FutureTest : TestBase() {
             depth.andDecrement
         }) {
             assertEquals(1, depth.get(), "Part before first suspension must be wrapped")
-            val result =
```
murfel, line 193: undo accidental reformat on this file

test-utils/native/src/ThreadUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+        if (!starting.compareAndSet(false, true)) {
+            throw IllegalStateException("Cannot call start() on a ConcurrentThread which has already started")
+        }
+        worker = Worker.start(name = name)
+        future = worker!!.execute(TransferMode.SAFE, { block }) { it.run() }
+        started.store(true)
+    }
+
+    actual fun join() {
+        if (started.load() && !joined.load()) {
```
dkhalanskyjb, line 38: This `join` will immediately finish if the thread was unstarted.

test-utils/concurrent/src/ThreadUtils.concurrent.kt
```diff
@@ -0,0 +1,30 @@
...
+
+
+expect class MultiplatformThread(
+    block: Runnable, name: String?
+) {
+    fun start()
+    fun join()
+}
+
+fun ConcurrentThread(name: String?, block: () -> Unit) = MultiplatformThread({ block() }, name)
```
dkhalanskyjb: Inconsistent naming. Shouldn't it be `MultiplatformThread` as well?

test-utils/concurrent/src/SynchronizedSet.kt
dkhalanskyjb: A radical suggestion: since this is only used in a single test, why not inline the implementation? Shouldn't be very intrusive; in fact, I'd expect the test to become clearer.

The reason I'm wary of this structure is that the tests depending on the entities from the implementation is a significant antipattern: there's the risk of the tests implicitly testing themselves. For example, here, we'll need to be careful not to accidentally use `SynchronizedSet` for testing `kotlinx.coroutines.sync.Mutex`: a bug in the mutex would affect the behavior of tests in two aspects that may cancel each other out. We do depend on `kotlinx.coroutines` machinery in test utils when it's simply too useful to make do without it, but in this scenario, it feels easily avoidable.

dkhalanskyjb (changes requested):
Nothing else caught my eye immediately. I'll take a closer look when the CI starts passing.

test-utils/concurrent/src/CountDownLatch.kt
```diff
@@ -0,0 +1,62 @@
...
+package kotlinx.coroutines.testing
+
+import kotlinx.atomicfu.*
+import kotlinx.atomicfu.locks.*
+import kotlin.time.*
+
+// Adapted from kotlinx-atomicfu
+// https://github.com/Kotlin/kotlinx-atomicfu/blob/d09c2b07cd16b0b273bd94edaa4929acd2ec42bc/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt#L64
+
+class CountDownLatch(count: Int) {
```
murfel, line 10: I need to audit this copy-pasted impl

test-utils/concurrent/src/ThreadUtils.concurrent.kt
murfel: By the way, is this what we call a "constructor function" and if so, is it just a style preference compared to creating a secondary constructor within the class?

test-utils/native/src/ThreadUtils.kt
murfel: Yes, to match JVM behaviour.
> join() -> join(0)

> `join(long millis)` This implementation uses a loop of this.wait calls conditioned on this.isAlive.

> `isAlive()` Tests if this thread is alive. A thread is alive if it has been started and has not yet died.

https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#join--

test-utils/concurrent/src/ThreadUtils.concurrent.kt
dkhalanskyjb: We do call it that, but the decision to implement it usually goes beyond style preference.

Here, making it a secondary constructor would prevent us from having `actual typealias` on the JVM, since the Java `Thread` doesn't have such a constructor.

In other cases, we may introduce constructor functions to hide the actual implementation type behind an interface. For example, `fun CoroutineScope` could not have been a constructor, as `CoroutineScope` is an interface. `fun StandardTestDispatcher` returns a `TestDispatcher`, because we may want to remove/update/rename/etc. the `internal class StandardTestDispatcher` while preserving the function's behavior. Tighter API surface → less ABI compatibility work.

test-utils/native/src/ThreadUtils.kt
dkhalanskyjb: Oh! Sorry, my bad. This behavior looks so unnatural to me that I couldn't imagine the JVM doing that and didn't even check. Please disregard my comment, then.

test-utils/concurrent/src/Barrier.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlinx.atomicfu.locks.*
+import kotlin.time.*
+
+/**
+ * Single-use barrier that blocks all participants until they all arrive.
+ *
+ * Adapted from kotlinx-atomicfu
+ * https://github.com/Kotlin/kotlinx-atomicfu/blob/d09c2b07cd16b0b273bd94edaa4929acd2ec42bc/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/BarrierTest.kt#L60
+ */
+class Barrier(private val parties: Int) {
```
murfel, line 15: Re `CyclicBarrier`

I've kept the usages of `java.util.concurrent.CDW/CB` on JVM (which required me explicitly importing them in most cases, so now these files appear in the diff, although nothing has changed for them). This is to keep at least the JVM test definitely correct.

I've avoided re-implementing a true `CyclicBarrier` with the JVM semantics because it's time-consuming and tricky to get it right. Instead, I copy-pasted and reworked `Barrier` and `TwoPhaseBarrier` (confusingly called `CyclicBarrier`) from atomicfu.

`Barrier` is a single-use barrier. It will throw if await is called more than N times.

`TwoPhaseBarrier` is a barrier where no more than one thread can be waking up sleeping threads (so roughly up to 2N-1 awaiters), so basically there is always up to two phases (and up to one of them is triggered). It covers all the other usages that `Barrier` doesn't cover, but does not implement misuse detection. If it is misused, it will be hard to debug.

`CyclicBarrier` supports any number of phases, so if 10N threads call await at the same time, it works. Thus, it is harder to misuse. But it is more tricky to implement than `TwoPhaseBarrier`.

So I've opted to not have `CyclicBarrier` but we can of course implement it in future and replace all barrier usages with it, if we wish to invest the time this way.

test-utils/concurrent/src/CountDownLatch.kt
murfel: Yep, this implementation has issues. It is possible for a thread to enqueue but never get dequeued - this is a memory leak. It is also possible for a thread to get unparked but never get parked.

Will fix it with a lock-free stack in a multiple-producer, single consumer flavour.

murfel: Let's review the latch here, since more people can participate.
https://github.com/Kotlin/kotlinx-atomicfu/pull/577

test-utils/jvm/test/CountDownLatchJvmTest.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Test
+    fun testAwaitAndCountDown() {
+        val latch = CountDownLatch(1)
+        val t1 = thread {
+            latch.await()
+        }
+        val t2 = thread {
+            Thread.sleep(100)
```
dkhalanskyjb: 100 milliseconds in one test is not a lot, but our test suite would take much longer to run if we didn't avoid `Thread.sleep` in tests wherever possible as a policy.

I see a few options here:
* Adopt https://github.com/Kotlin/kotlinx-atomicfu/blob/4e607600049d32d7239fa38c8a0ad663ef64e052/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt#L14-L62 instead. This test actually asserts the behaviors we expect from a latch, giving us some peace of mind.
* `@Ignore` the tests and only run them manually whenever we're changing the implementation of the latch. Since this is not production code, we're not as concerned with its quality and aren't as scared of regressions.
* More radically, have proper tests in `test-utils`, but don't run them by default as part of `gradle clean check`. The tests could take as long as we like.
* Don't test the latch at all. If there are issues with it, we will probably notice with our test suite hanging or behaving more race-y than it should have. `CountDownLatch` regressions can't *only* lead to false positives in our test suite, can they?

I don't have a preference here, I just think 100 ms per run for smoke-testing a non-production piece of code is too wasteful.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
```diff
@@ -11,6 +11,7 @@ import org.junit.*
 import org.junit.Test
 import java.lang.Runnable
 import java.util.concurrent.*
+import java.util.concurrent.CountDownLatch
```
dkhalanskyjb, line 14: Minor: here and in other files, there are some unnecessary imports added.

kotlinx-coroutines-debug/test/RunningThreadStackMergeTest.kt
```diff
@@ -143,15 +114,7 @@ class RunningThreadStackMergeTest : DebugTestBase() {
...
-                "\tat kotlinx.coroutines.debug.StacktraceUtilsKt.verifyDump\$default(StacktraceUtils.kt)\n" +
-                "\tat kotlinx.coroutines.debug.RunningThreadStackMergeTest\$testRunBlocking\$1.invokeSuspend(RunningThreadStackMergeTest.kt)"
+            "Coroutine \"coroutine#1\":BlockingCoroutine{Active}@4bcd176c, state: RUNNING\n" + "\tat java.lang.Thread.getStackTrace(Thread.java)\n" + "\tat kotlinx.coroutines.debug.internal.DebugProbesImpl.enhanceStackTraceWithThreadDumpImpl(DebugProbesImpl.kt)\n" + "\tat kotlinx.coroutines.debug.internal.DebugProbesImpl.dumpCoroutinesSynchronized(DebugProbesImpl.kt)\n" + "\tat kotlinx.coroutines.debug.internal.DebugProbesImpl.dumpCoroutines(DebugProbesImpl.kt)\n" + "\tat kotlinx.coroutines.debug.DebugProbes.dumpCoroutines(DebugProbes.kt)\n" + "\tat kotlinx.coroutines.debug.StacktraceUtilsKt.verifyDump(StacktraceUtils.kt)\n" + "\tat kotlinx.coroutines.debug.StacktraceUtilsKt.verifyDump\$default(StacktraceUtils.kt)\n" + "\tat kotlinx.coroutines.debug.RunningThreadStackMergeTest\$testRunBlocking\$1.invokeSuspend(RunningThreadStackMergeTest.kt)"
         )
     }
```
dkhalanskyjb, line 156: This whole file looks like it was edited by mistake. The new indentation of the stack traces doesn't look like an improvement to me.

kotlinx-coroutines-core/concurrent/test/JobChildStressTest.kt
```diff
@@ -76,35 +77,36 @@ class JobChildStressTest : TestBase() {
...
-                if (it.mod(2) == 0) {
+                if (iteration.mod(2) == 0) {
                     deferred.invokeOnCompletion { }
                 }
                 launch(pool + deferred) {
                     deferred.complete(Unit) // Transition deferred into "completing" state waiting for current child
                     // **Asynchronously** submit task that launches a child so it races with completion
-                    pool.executor.execute {
-                        rogueJob.set(launch(pool + deferred) {
+                    pool.dispatch(currentCoroutineContext()) {
```
dkhalanskyjb: Minor: `EmptyCoroutineContext` would more correctly denote that we don't care which context this is, I think.

kotlinx-coroutines-core/concurrent/test/JoinStressTest.kt
```diff
@@ -54,7 +51,7 @@ class JoinStressTest : TestBase() {
         val results = IntArray(2)
 
         repeat(iterations) {
-            val barrier = CyclicBarrier(4)
+            val barrier = TwoPhaseBarrier(4)
```
dkhalanskyjb: Couldn't a `Barrier` have done the job here as well? With the added protection against the extra `await()` calls it provides, it would be preferable to use it where applicable.

test-utils/concurrent/src/TwoPhaseBarrier.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    fun await() {
+        val wrapper = HandleWrapper(ParkingSupport.currentThreadHandle())
+        val n = queue.enqueue(wrapper)
+        if (n % parties == 0L) {
+            var wokenUp = 0
+            while (wokenUp < parties - 1) {
+                val deq = queue.dequeue() ?: error("CyclicBarrier await failed to dequeue")
+                if (deq.first % parties == 0L) continue
+                if (deq.second.woken.compareAndSet(false, true)) {
```
dkhalanskyjb: Proposal:
```kotlin
val successfullyAwoken = deq.second.woken.compareAndSet(false, true)
check(successfullyAwoken)
```
This way, if the state is somehow broken, we'll learn about it instead of skipping part of the logic as if nothing happened.

test-utils/concurrent/src/Barrier.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    private val count = atomic(0)
+    private val waiters = atomicArrayOfNulls<Any?>(parties - 1)
+
+    fun await() {
+        val myIndex = count.getAndIncrement()
+        if (myIndex >= parties) {
+            // count could overflow and not enter this if, but it's unlikely and has safeguards down the line
+            error("No more than $parties threads can call await() throughout the lifetime of the barrier($parties).")
+        }
```
dkhalanskyjb: Minor, style:
```suggestion
        check(myIndex < parties) {
            "No more than $parties threads can call await() throughout the lifetime of the barrier($parties)."
        }
```
As for the comment: let's consider it impossible that we somehow fail to notice `4*10^9` exceptions flying around in our test suite. If we take this risk seriously, the safeguards aren't entirely robust, as `2^32` `await` calls later, `myIndex` will be `0` again, and the thread will likely happily observe `FINISHED` and exit the `await()` call, even though it has no right to.

kotlinx-coroutines-core/concurrent/test/flow/SharingStressTest.kt
```diff
@@ -166,7 +175,11 @@ class SharingStressTest : TestBase() {
                             if (expected != j) {
                                 if (j == expected + 1) {
                                     // if missing just one -- could be race with cancelled emit
-                                    missingCollects.add(expected)
+                                    runBlocking {
+                                        missingCollectsMutex.withLock {
```
dkhalanskyjb: Is there a reason not to use `synchronized {}` instead of the `runBlocking`/`withLock` combination here?

dkhalanskyjb (changes requested):

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
murfel: This is necessary. Without it, CountDownLatch is ambiguous. And I want to use the JVM latch where possible, because it is definitely correct.

kotlinx-coroutines-core/concurrent/test/flow/SharingStressTest.kt
murfel: What would be the `SynchronizedObject`?

test-utils/concurrent/src/TwoPhaseBarrier.kt
```diff
@@ -0,0 +1,76 @@
...
+package kotlinx.coroutines.testing
+
+import kotlinx.atomicfu.locks.*
+import kotlinx.atomicfu.*
+import kotlin.time.*
+
+// Adapted from kotlinx-atomicfu
+// https://github.com/Kotlin/kotlinx-atomicfu/blob/d09c2b07cd16b0b273bd94edaa4929acd2ec42bc/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/CyclicBarrierTest.kt#L59
+
+class TwoPhaseBarrier(private val parties: Int) {
```
murfel, line 14: Native doesn't compile (compiler backend error). If I comment-out this class and all its usages, it compiles and tests pass.

```
e: org.jetbrains.kotlin.backend.common.CompilationException: Back-end: Please report this problem https://kotl.in/issue
/Users/natasha/IdeaProjects/kotlinx.coroutines/test-utils/concurrent/src/TwoPhaseBarrier.kt:-1:-1
Details: Internal error in file lowering: java.lang.IllegalStateException: actual type is kotlin.Byte, expected kotlin.Boolean
```
https://gist.github.com/murfel/66d74f9cc391aa4725ce355981f81d75#file-gistfile1-txt-L38

If I replace this impl with the original impl from atomicfu, it doesn't compile either.

Any idea why or should I just file an issue in the compiler?

murfel: @fzhinkin could this be due to atomicfu?

dkhalanskyjb: Does replacing `woken = atomic(false)` with `AtomicBoolean` fix the problem? The usage of `wrapper.woken.compareAndSet` looks suspicious to me (in terms of `atomicfu`).

kotlinx-coroutines-core/concurrent/test/flow/SharingStressTest.kt
dkhalanskyjb: Some new value, like `val missingCollectsLock = SynchronizedObject()`.

reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
dkhalanskyjb: Got it, thanks.

test-utils/concurrent/src/TwoPhaseBarrier.kt
fzhinkin: It is https://youtrack.jetbrains.com/issue/KT-83367, isn't it?
You can try compiling the project w/ the Kotlin 2.3.20-Beta2 to check if the issue is gone.

murfel: Oh wow, it is exactly this. I walked a full circle. Thanks!

test-utils/native/src/ThreadUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+
+package kotlinx.coroutines.testing
+
+import kotlinx.coroutines.*
+import kotlin.concurrent.atomics.*
+import kotlin.native.concurrent.*
+
+actual fun assertTrueJvm(value: Boolean) = Unit
+
+
```
dkhalanskyjb: Minor, style, here and everywhere in new files: most of the recently edited codebase is formatted without more than one empty line in a row. The rule of thumb is: "If you want to separate some code, do that with a comment, not with extra whitespace." This convention is occasionally broken, but it's still rare enough for this place in code to jump out at me as unusual. Let's preserve stylistic consistency.

test-utils/native/src/ThreadUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+        }
+        worker = Worker.start(name = name)
+        future = worker!!.execute(TransferMode.SAFE, { block }) { it.run() }
+        started.store(true)
+    }
+
+    actual fun join() {
+        if (started.load() && !joined.load()) {
+            future!!.consume { }
+            worker!!.requestTermination().result
```
dkhalanskyjb: `consume { }` and `.result` throw when called more than once.
Minimal example:
```kotlin
val worker = Worker.start()
val myFuture = worker.execute(TransferMode.SAFE, { }) { }
myFuture.result
myFuture.result
```
Fails with
```
kotlin.IllegalStateException: Future is in an invalid state
```
The implementation is therefore racy and unsafe. If two threads call `.join()` on another thread simultaneously, they can both enter this `if` block and call `consume` on the same `future`, leading to one of them throwing an exception.

test-utils/native/src/FieldWalker.kt
dkhalanskyjb: The implementation is still there.

test-utils/jvm/test/TestBaseTest.kt
```diff
@@ -0,0 +1,22 @@
+import kotlinx.coroutines.*
+import kotlinx.coroutines.testing.*
+import org.junit.*
```
dkhalanskyjb: Minor, style: better to use `kotlin.test` for consistency.

test-utils/jvm/test/TestBaseTest.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class TestBaseTest : TestBase() {
+    @Test
+    fun testThreadsShutdown() {
+        repeat(1000 * stressTestMultiplier) { _ ->
+            initPoolsBeforeTest()
+            val threadsBefore = currentThreads()
+            runBlocking {
+                val sub = launch {
+                    delay(10000000L)
```
dkhalanskyjb: Minor, style: I'm not sure what the intention is here, but this magic constant looks easily avoidable. `awaitCancellation` could signify "wait until the coroutine is cancelled". `delay(Long.MAX_VALUE)` or `delay(Duration.INFINITE)` are other ways of expressing the same idea if you'd prefer to use `delay`.

test-utils/jvm/test/TestBaseTest.kt
```diff
@@ -0,0 +1,22 @@
...
+                }
+                sub.cancel()
+                sub.join()
```
dkhalanskyjb: This tandem of "launch, cancel, join" ultimately does nothing. A coroutine is launched, and then, before it even has a chance to enter the `delay` call, it gets cancelled, and then it is instantaneously joined (since it's completed).

test-utils/jvm/test/TestBaseTest.kt
```diff
@@ -0,0 +1,22 @@
...
+import kotlinx.coroutines.testing.*
+import org.junit.*
+
+class TestBaseTest : TestBase() {
+    @Test
+    fun testThreadsShutdown() {
+        repeat(1000 * stressTestMultiplier) { _ ->
+            initPoolsBeforeTest()
+            val threadsBefore = currentThreads()
+            runBlocking {
```
dkhalanskyjb: The entirety of the `runBlocking { /* */ }` call looks out-of-place here and can be easily replaced by `Thread.sleep(50)`, it would seem. This call doesn't use any threads other than the current one, so it doesn't interact with any of the thread pools that this test seems to focus on, based on its name.

If the intention was to use the extra threads, some coroutines should be launched on `Dispatchers.Default`, and also possibly run `delay` with a non-zero non-infinite constant to make sure the delay-processing thread gets started.

test-utils/jvm/test/TestBaseTest.kt
```diff
@@ -0,0 +1,22 @@
+import kotlinx.coroutines.*
```
dkhalanskyjb: Minor, style: for test, we typically use the package that contains the API under test. This way, in the logs, it's easier to recognize at a glance what exactly failed.

kotlinx-coroutines-core/jvm/test/FailingCoroutinesMachineryTest.kt
```diff
@@ -1,11 +1,14 @@
 package kotlinx.coroutines
 
 import kotlinx.coroutines.testing.*
+// TODO: implement CountDownLatch.await with timeout (in this PR)
```
dkhalanskyjb: Before merging, the fate of this comment has to be decided.

kotlinx-coroutines-core/concurrent/test/JobChildStressTest.kt
```diff
@@ -76,35 +78,36 @@ class JobChildStressTest : TestBase() {
     @Test
     fun testChildAttachmentRacingWithLastChildCompletion() {
         // All exceptions should get aggregated here
-        repeat(N_ITERATIONS) {
+        repeat(N_ITERATIONS) { iteration ->
             val canCloseThePool = CountDownLatch(1)
             runBlocking {
-                val rogueJob = AtomicReference<Job?>()
+                val rogueJob = AtomicReference<Job?>(null)
+
```
dkhalanskyjb: Minor, style: there's an extra newline here.

kotlinx-coroutines-core/concurrent/test/channels/ChannelUndeliveredElementSelectOldStressTest.kt
```diff
@@ -92,10 +98,12 @@ class ChannelUndeliveredElementSelectOldStressTest(private val kind: TestChannel
                     stopSender()
                     launchSender()
                 }
+
                 1 -> { // cancel & restart receiver
                     stopReceiver()
                     launchReceiver()
                 }
+
```
dkhalanskyjb: Minor, style: extra newlines got added here.

kotlinx-coroutines-core/concurrent/test/channels/ChannelUndeliveredElementStressTest.kt
```diff
@@ -92,10 +88,12 @@ class ChannelUndeliveredElementStressTest(private val kind: TestChannelKind) : T
                     stopSender()
                     launchSender()
                 }
+
                 1 -> { // cancel & restart receiver
                     stopReceiver()
                     launchReceiver()
                 }
+
```
dkhalanskyjb: Ditto (there are also other spots, but I won't list them all, to reduce the comment spam).

kotlinx-coroutines-core/concurrent/test/flow/SharedFlowStressTest.kt
```diff
@@ -17,14 +14,19 @@ class SharedFlowStressTest : TestBase() {
     private lateinit var sf: MutableSharedFlow<Long>
     private lateinit var view: SharedFlow<Long>
 
-    @get:Rule
-    val producerDispatcher = ExecutorRule(nProducers)
-    @get:Rule
-    val consumerDispatcher = ExecutorRule(nConsumers)
+    val producerDispatcher = newFixedThreadPoolContext(nProducers, "Producers")
+
```
dkhalanskyjb, line 18: Minor, style: they can also be next to one another here, as is done for the other properties in this class.

kotlinx-coroutines-core/concurrent/test/flow/StateFlowCancellabilityTest.kt
```diff
@@ -1,23 +1,24 @@
...
 class StateFlowCancellabilityTest : TestBase() {
     @Test
     fun testCancellabilityNoConflation() = runTest {
         expect(1)
         val state = MutableStateFlow(0)
-        var subscribed = true
-        var lastReceived = -1
-        val barrier = CyclicBarrier(2)
+        val subscribed = AtomicBoolean(false)
+        val lastReceived = AtomicInt(-1)
```
dkhalanskyjb: Why make them atomic? The barrier should establish a happens-before relationship anyway.

dkhalanskyjb (changes requested):
The PR looks fairly close to getting merged, there are only local concerns.
The one big thing is that needs a cleanup is that `tests-jvm-to-concurrent-move-files` moves some files that this PR moves back.

dkhalanskyjb:
Merged manually: https://github.com/Kotlin/kotlinx.coroutines/commit/2a4ac9a351783474ee532aa6b70eecafce37a720

test-utils/concurrent/src/CountDownLatch.kt
```diff
@@ -12,13 +12,22 @@ class CountDownLatch(count: Int) {
     private val waiters = MPSCQueueLatch<ParkingHandle>()
 
     fun await() {
+        tryAwait(Duration.INFINITE)
+    }
+
+    fun tryAwait(duration: Duration): Boolean {
```
murfel, line 18: @dkhalanskyjb why is it called it `tryAwait`, not just `await`, by the way?

dkhalanskyjb: This is a convention for boolean-returning functions, to indicate that you need to check the result. See: `tryLock`, `tryEmit`, `trySend`, etc.

```kotlin
doSomething()
latch.await(1.seconds)
doSomethingElse()
```
does not look like a mistake: it's normal for operations accepting a deadline to finish with an exception if they fail to execute within the given limit.

With the `try*` convention, this is supposed to look very suspicious during a review:
```kotlin
doSomething()
latch.tryAwait(1.seconds)
doSomethingElse()
```

This entire `try*` convention may become redundant once the return value checker gets enabled by default.

murfel: Ah, in my head it was 'the non-suspending counterpart'

PR #4605, opened 2026-01-24
Port tests from jvm to concurrent: move

murfel:

PR #4607, opened 2026-01-24
Port tests from jvm to concurrent: to be fixed

murfel:


dkhalanskyjb:
In general, we shouldn't store commented-out code blocks in the repository, even temporarily.
Here specifically, commenting out a test that used to pass on the JVM is leaving the codebase in a strictly worse state than it was before the group of change, since now, these tests don't run on any platform.
Instead of this PR, https://github.com/Kotlin/kotlinx.coroutines/pull/4605 should be changed not to include the files that aren't ported to the `concurrentTest` source set.

dkhalanskyjb:
Additionally, some tests still don't pass:
* `kotlinx.coroutines.EventLoopsTest.testSecondThreadRunBlocking` hangs.
* `DispatchersToStringTest` fails (because of a problem with our Native dispatcher implementations not having a `toString`).
* `ChannelSendReceiveStressTestBUFFERED1BROADCASTx10x10.testSendReceiveStress` fails, for a reason we'd need to look into.

PR #4610, opened 2026-01-30
Simplify the build scripts

dkhalanskyjb:
@fzhinkin FYI, this logic is removed:
```kotlin
if (isSnapshotTrainEnabled(rootProject)) {
    // Report Kotlin compiler version when building project
    println("Using Kotlin compiler version: ${KotlinCompilerVersion.VERSION}")
}
```
+ you were interested in https://github.com/Kotlin/kotlinx.coroutines/commit/cfdea221366f528859303e4d7300bb528cbd0f75


buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
```diff
@@ -27,32 +27,30 @@ kotlin {
         }
     }
     jvmToolchain(jdkToolchainVersion)
-    if (nativeTargetsAreEnabled) {
```
qwwdfsad, line 32: (comment to the commit message)
This was a property to easily test JVM IR BE without compiling and running all the native code. Can be safely removed of course

qwwdfsad (approved):
Terrific job!

I've walked through commits one by one, the changes are straightforward and well-documented. I have to admit my Gradle-fu is rusty, so I'd recommend having a second opinion on that

PR #4616, opened 2026-02-05
Fix a memory leak in RxScheduler implementations

### LINKED ISSUE #4615: Memory Leak in CoroutineDispatcher.asScheduler() and DispatcherScheduler
**Describe the bug**

The following code produces a memory leak. E.g. after some time the code crashes with a out of Memory exception:

```
        [...]
        val buggyScheduler =  Dispatchers.Default.asScheduler()
        val workingScheduler =  Schedulers.computation()
        
        val s = if (BUG_ENALBED) buggyScheduler else workingScheduler

        disposable = Flowable.interval(1, 1, TimeUnit.SECONDS, Schedulers.computation())
            .doOnNext { Log.d(TAG, "timer tick every 1  second here; create new Flowable") }
            .switchMap {
                val bigObject = ByteArray(10 * 1024 * 1024) // 10 MiB

                Flowable.interval(5, 1, TimeUnit.MINUTES, s)
                    .doOnNext {
                        // The object "bigObject" is captured in this lambda.
                        Log.d(TAG, "Inner Flowable has fired! size=${bigObject.size}")
                    }
            }
            // The following line is never executed. The upstream Flowable is replaced every one second.
            // So it has never a chance to emit it's first element, because it's set to five minutes.
            //
            // What is the bug?
            // The upstream flowable is replaced. Therefore it's disposed. But the implementation in
            // RxScheduler.java does _not_ removed the scheduled Runnable/coroutine from the task queue
            // even so they are disposed.
            // So every second a new runnable/coroutine is added to the task queue that first runs after 5 minutes.
            // These runnables/coroutines queue up instead of being garbage collected. If the runnables
            // contains/reference big objects you have a memory leak that is visible.
            .doOnNext { Log.d(TAG, "Never executed") }
            .subscribe()
````

The bug is file [RxScheduler.kt](https://github.com/Kotlin/kotlinx.coroutines/blob/b11abdf01d4d5db85247ab365abc72efc7b95062/reactive/kotlinx-coroutines-rx3/src/RxScheduler.kt#L136) in line 135/136

````
    if (delayMillis <= 0) {
        toSchedule.run()
    } else {
        @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE") // do not remove the INVISIBLE_REFERENCE suppression: required in K2
        ctx.delay.invokeOnTimeout(delayMillis, toSchedule, ctx).let { handle = it }
    }
````

The return value of `invokeOnTimeout`, a Disposable, is not used to remove the runnable from the task queue when the outer disposable is disposed.

**Provide a Reproducer**

The full reproducer is in this repository: https://github.com/lengfeld/android-app-kotlinx-coroutines-rx3-leak-reproducer/tree/main?tab=readme-ov-file

Thanks for looking into this.

### PR DESCRIPTION
**Pull request by dkhalanskyjb**
From the RxJava docs on `Scheduler.Worker`:

> Disposing the Scheduler.Worker should cancel all outstanding work
> and allows resource cleanup.

and

> [schedule]'s implementation is encouraged to track the individual
> Runnable tasks while they are waiting to be executed
> (with or without delay) so that dispose() can prevent their
> execution or potentially interrupt them if they are currently
> running.

Fixes #4615

### DIFF
--- a/reactive/kotlinx-coroutines-rx3/test/SchedulerTest.kt
+++ b/reactive/kotlinx-coroutines-rx3/test/SchedulerTest.kt
@@ -516,7 +516,7 @@ class SchedulerTest : TestBase() {
                 }
             }
         }.asScheduler()
-        // Allocates a new worker internally and scheduler a ticker,
+        // Allocates a new worker internally and schedules a ticker,
         // *without* storing the scheduled periodic tasks anywhere.
         // The only way for them to get cancelled is by the worker disposing of them as part of its own disposal.
         val disposable = Flowable.interval(5, 1, TimeUnit.MINUTES, customScheduler)
--- a/reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
+++ b/reactive/kotlinx-coroutines-rx2/test/SchedulerTest.kt
@@ -515,7 +515,7 @@ class SchedulerTest : TestBase() {
                 }
             }
         }.asScheduler()
-        // Allocates a new worker internally and scheduler a ticker,
+        // Allocates a new worker internally and schedules a ticker,
         // *without* storing the scheduled periodic tasks anywhere.
         // The only way for them to get cancelled is by the worker disposing of them as part of its own disposal.
         val disposable = Flowable.interval(5, 1, TimeUnit.MINUTES, customScheduler)

### CODE REVIEW

**Review by tomritzow — COMMENTED** [review]
@

File: reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt [hunk 1]
```diff
@@ -133,7 +133,14 @@ private fun CoroutineScope.scheduleTask(
         toSchedule.run()
     } else {
         @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE") // do not remove the INVISIBLE_REFERENCE suppression: required in K2
-        ctx.delay.invokeOnTimeout(delayMillis, toSchedule, ctx).let { handle = it }
+        ctx.delay.invokeOnTimeout(delayMillis, toSchedule, ctx).let {
+            // If the worker is cancelled, all outstanding work must be cancelled, too.
+            val onWorkerCancellationDisposal = ctx.job.disposeOnCompletion(it)
```
qwwdfsad: Doesn't this code imply that if a worker is *not* cancelled, then all `onWorkerCancellationDisposal` ever observed by this worker will be retained until the worker is disposed?

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
```diff
@@ -133,7 +133,14 @@ private fun CoroutineScope.scheduleTask(
         toSchedule.run()
     } else {
```
qwwdfsad, line 142: [note] while you are here -- `val disposable = Disposables.fromRunnable` seems like a completely redundant allocation most of the time and can be `if`'ed around to use something like `Disposable.empty()`

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
dkhalanskyjb: I believe you are right, yes. Looks like we can't dance around just spawning an extra coroutine for every task, like `scope.launch(start = CoroutineStart.UNDISPATCHED) { delay(...); ... }`. To prop this up with extra callbacks further would just mean recreating structured concurrency anew.

qwwdfsad: Agree. Here it probably can be addressed by invoking disposable handle at the end of `toSchedule` task -- should be enough in most of the cases

dkhalanskyjb: Disposing of a handle unconditionally means the user-visible `Disposable` will get disposed of as well, and someone may check `isDisposed` to see if *their* program logic cancelled a computation. Probably best to avoid that.

qwwdfsad: I've meant something like this: it doesn't leak to a user and properly cleans up otherwise retained `onWorkerCancellationDisposal`

```
Subject: [PATCH] ~
---
Index: reactive/kotlinx-coroutines-rx3/src/RxScheduler.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/reactive/kotlinx-coroutines-rx3/src/RxScheduler.kt b/reactive/kotlinx-coroutines-rx3/src/RxScheduler.kt
--- a/reactive/kotlinx-coroutines-rx3/src/RxScheduler.kt	(revision ca0913f2d895ba3e75c1a7c0c0df8446ffe2f17c)
+++ b/reactive/kotlinx-coroutines-rx3/src/RxScheduler.kt	(date 1771501536406)
@@ -127,7 +127,19 @@
         }
     }
 
-    val toSchedule = adaptForScheduling(::task)
+    var potentiallyLeakingHandle: DisposableHandle? = null
+    val toSchedule = if (delayMillis <= 0) {
+        adaptForScheduling(::task)
+    } else {
+        val r = adaptForScheduling(::task)
+        Runnable {
+            try {
+                r.run()
+            } finally {
+                potentiallyLeakingHandle?.dispose()
+            }
+        }
+    }
     if (!isActive) return Disposable.disposed()
     if (delayMillis <= 0) {
         toSchedule.run()
@@ -136,6 +148,7 @@
         ctx.delay.invokeOnTimeout(delayMillis, toSchedule, ctx).let {
             // If the worker is cancelled, all outstanding work must be cancelled, too.
             val onWorkerCancellationDisposal = ctx.job.disposeOnCompletion(it)
+            potentiallyLeakingHandle = onWorkerCancellationDisposal
             handle = DisposableHandle {
                 it.dispose()
                 onWorkerCancellationDisposal.dispose()
```

note that it's a draft of the idea, not the actual patch (i.e. at least `potentiallyLeakingHandle` is published unsafely here and has a race with task execution)

dkhalanskyjb: I implemented something similar to this. It looks obnoxiously complicated, but the simpler solution with just calling `launch { delay() }` creates a lot of completely pointless work for the dispatcher when the `delay` gets cancelled, so it may be worth it to bite the bullet and deal with the complexity.

I haven't added tests for my bug, because they take too long to get to the OOM condition with just the tiny handlers that get added to the `Job`.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
dkhalanskyjb: Done.

reactive/kotlinx-coroutines-rx2/src/RxScheduler.kt
qwwdfsad: >because they take too long to get to the OOM condition with just the tiny handlers that get added to the Job.

I've asked Claude to do so, and it seems reasonable (tho instead of GC and weak ref I'd suggest using field walker) as an idea:

```

/**
 * Tests that a delayed task scheduled via [Scheduler.scheduleDirect] does not retain
 * its closure after the task completes normally.
 *
 * The bug: `scheduleTask` registers an `invokeOnCompletion` handler on the scope's job
 * to cancel the pending timeout if the job is cancelled. When the timeout fires naturally
 * and the task completes, this completion handler is never removed, leaking the entire
 * closure chain: job -> completionHandler -> timeoutHandle -> taskClosure -> userRunnable.
 */
@Test
fun testScheduleDirectDelayedTaskDoesNotLeak() {
    newFixedThreadPoolContext(1, "test-scheduler").use { dispatcher ->
        val scheduler = dispatcher.asScheduler()
        try {
            val weakRef = scheduleDelayedTaskAndReturnWeakRef(scheduler)
            val deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(5)
            while (weakRef.get() != null && System.nanoTime() < deadline) {
                System.gc()
                Thread.sleep(50)
            }
            assertNull(
                weakRef.get(),
                "Memory leak: delayed task's closure is retained after completion " +
                    "via stale invokeOnCompletion handler on the scheduler's job"
            )
        } finally {
            scheduler.shutdown()
        }
    }
}

/**
 * Schedules a delayed task that captures a large marker object, waits for it to complete,
 * and returns a [WeakReference] to the marker. After this function returns, the marker
 * has no strong references on the call stack — the only path keeping it alive (if any)
 * is through the leaked completion handler chain inside the scheduler.
 */
private fun scheduleDelayedTaskAndReturnWeakRef(scheduler: Scheduler): WeakReference<ByteArray> {
    val marker = ByteArray(1_000_000)
    val weakRef = WeakReference(marker)
    val latch = CountDownLatch(1)
    scheduler.scheduleDirect({
        marker.size // capture marker in the closure
        latch.countDown()
    }, 1, TimeUnit.MILLISECONDS)
    assertTrue(latch.await(5, TimeUnit.SECONDS))
    Thread.sleep(500) // let the launched coroutine finish
    return weakRef
}
```

dkhalanskyjb: The idea is clear, but this test could silently stop meaningfully checking anything in the future when we expose the `Delay` interface to the public. It will become a good idea, I think, to drop a reference to the `Runnable` from the `invokeOnTimeout`'s `DisposableHandle` once it's triggered—after all, client code can store the `DisposableHandle`s indefinitely.

I added a more general `FieldWalker`-powered test.

qwwdfsad (approved):

reactive/kotlinx-coroutines-rx3/test/SchedulerTest.kt
```diff
@@ -487,6 +487,70 @@ class SchedulerTest : TestBase() {
...
+            ): DisposableHandle = object: DisposableHandle {
+                init {
+                    expect(1)
+                }
+                override fun dispose() {
+                    expect(2)
+                }
+            }
+        }.asScheduler()
+        // Allocates a new worker internally and scheduler a ticker,
```
murfel: schedules

reactive/kotlinx-coroutines-rx3/test/SchedulerTest.kt
```diff
@@ -487,6 +487,70 @@ class SchedulerTest : TestBase() {
...
+                expectUnreached()
+            }.subscribe()
+        // Dispose of the worker, which will, in turn, dispose of the outstanding task.
+        disposable.dispose()
+        finish(3)
+    }
+
+    /**
+     * Tests for a memory leak that occurs when disposable handlers added to the scheduler's CoroutineScope for the case
+     * the scheduler is disposed of are not removed when the task completes.
```
murfel, line 533: General - long sentences are hard to parse.

dkhalanskyjb: In general, sure. In the specific case of KDocs for tests, the alternative is to either more verbosely spell out the history behind the test, or to omit the comment altogether, though. Neither option sounds particularly appealing to me. High information density feels justified here.

PR #4618, opened 2026-02-09
Cancellation and timeouts tutorial review

murfel:


docs/topics/cancellation-and-timeouts.md
```diff
@@ -62,29 +56,11 @@ suspend fun main() {
...
-        // Waits for job1 to start before canceling it
-        job1Started.await()
+        // Waits for the coroutine to start before canceling it
+        jobStarted.await()
 
         // Cancels the coroutine, so delay() throws a CancellationException
-        job1.cancel()
-
-        // async returns a Deferred handle, which inherits from Job
-        val job2 = async {
```
murfel, line 72: The first example should be short and easy to comprehend. No need to repeat the same code twice. Added a commend on `async` below instead.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -231,109 +210,19 @@ suspend fun main() {
 
 ### Check for cancellation explicitly
```
murfel: I suggest to get rid of `isActive` and `ensureActive()` examples completely, since they are not the model use-case for them. Mention them briefly, don't provide non-typical code. If someone needs to use them, they will figure it out, these are not complicated APIs.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -580,8 +470,7 @@ suspend fun main() {
 
 ## Timeout
 
-Timeouts allow you to automatically cancel a coroutine after a specified duration.
-They are useful for stopping operations that take too long, helping to keep your application responsive and avoid blocking threads unnecessarily.
+Timeouts allow you to automatically cancel a coroutine after a specified duration. They are useful for stopping operations that take too long, to avoid serving outdated results, mitigate slow connection, or simply free up resources.
```
murfel: "keep your application responsive" is too generic, added more specific reasons

docs/topics/cancellation-and-timeouts.md
```diff
@@ -593,7 +482,7 @@ import kotlin.time.Duration.Companion.milliseconds
 suspend fun slowOperation(): Int {
     try {
         delay(300.milliseconds)
-        return 5
+        return 1
```
murfel: Simplifying. Too many random numbers, alongside the timeout numbers, make it hard to keep track. Even better approach could have been to return a string, to avoid numbers altogether.

docs/topics/coroutines-cancellation.md
```diff
@@ -457,6 +346,7 @@ class ScreenWithFileContents(private val scope: CoroutineScope) {
         scope.cancel()
     }
 }
```
murfel, line 364: Still confused by this example. As I understood, the intention was to contrast the usage of `updateUi` from within `displayFile` and from some other function. But why is it relevant that `updateUi` throws?

docs/topics/cancellation-and-timeouts.md
```diff
@@ -24,13 +21,11 @@ implements `Job` and supports the same cancellation behavior.
...
 
 When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) the next time it checks for cancellation.
-For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
+The [Suspension points and cancellation](#suspension-points-and-cancellation) section covers precisely when this happens, but for now it is enough to know that all suspend functions in Kotlin coroutines, such as `delay()` and `awaitCancellation()` explained below, check for cancellation if they suspend.
 
-> You can use the [`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function to suspend a coroutine until it's canceled.
->
-{style="tip"}
+[`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function suspends a coroutine until it's canceled. It is equivalent to `delay(Duration.INFINITE)`.
```
murfel: Avoid using the tip box for information which is key to understanding the article, if it's possible to introduce it more naturally as part of the main text.

Also, introducing it here allows to use it in the example instead of `delay`, as it is the idiomatic way to do that.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -2,13 +2,10 @@
...
-Cancellation lets you stop a coroutine before it completes.
+Cancellation lets you request to stop a coroutine before it completes.
 It stops work that's no longer needed, such as when a user closes a window or navigates away in a user interface while a coroutine is still running.
-You can also use it to release resources early and to stop a coroutine from accessing objects past their disposal.
-
-> You can use cancellation to stop long-running coroutines that keep producing values even after other coroutines no longer need them, for example, in [pipelines](channels.md#pipelines).
->
-{style="tip"}
+You can use it to release resources early and to stop a coroutine from accessing objects past their disposal.
+You can also use cancellation to stop long-running coroutines that keep producing values even after other coroutines no longer need them, such as a coroutine collecting an infinite [flow](flow.md).
```
murfel: I'd like to remove mentions of channels since we don't want to advertise them.

Looking back, not sure if the flow example is good. Possibly we should just remove any mentions of channels or flows, and say "producer coroutine" - it hints at the right idea for those who know it, and doesn't encourage looking up outdated material for those who don't.

docs/topics/coroutines-cancellation.md
```diff
@@ -40,19 +35,18 @@ import kotlin.time.Duration
 suspend fun main() {
```
murfel, line 43: I'd like to hide `suspend fun main() {` behind the `//sampleStart` and then decrease the indentation, to have more text fit into the playground window. I'll do it as the very last change, to keep the diff down. Any objections?

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: The flow example definitely feels off to me, since there's no logical connection between consuming values and producing anything.
Also, without the connection to channels, the focus on producing values looks misplaced in general: there are many other sorts of long-running tasks that benefit from cancellation. Sending heartbeats, running scheduled tasks, updating a state to reflect the newest reading (like the clock UI)—all of these may need to be stopped.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -24,13 +21,11 @@ implements `Job` and supports the same cancellation behavior.
 You can call the `cancel()` function manually, or it can be invoked automatically through cancellation propagation when a parent coroutine is canceled.
 
 When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) the next time it checks for cancellation.
-For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
+The [Suspension points and cancellation](#suspension-points-and-cancellation) section covers precisely when this happens, but for now it is enough to know that all suspend functions in Kotlin coroutines, such as `delay()` and `awaitCancellation()` explained below, check for cancellation if they suspend.
```
dkhalanskyjb: > all suspend functions in Kotlin coroutines

Do you mean the `kotlinx.coroutines` library? Right now, it reads to me like "all suspend functions in coroutines in Kotlin", but if so, the statement is inaccurate.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -96,8 +72,9 @@ suspend fun main() {
 {kotlin-runnable="true" id="manual-cancellation-example"}
 
 In this example, [`CompletableDeferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/) is used as a signal that the coroutine has started running.
-The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed. This way, cancellation happens only after the coroutine has started running.
-The coroutine created by `.async()` doesn't have this check, so it may be canceled before it can run the code inside its block.
+The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed. This way, cancellation happens only after the coroutine has started running. Without this check, the coroutine could be canceled before it runs the code inside its block. You don't need to have this check to cancel a coroutine, but it is included here to have a consistent example which always prints the explanation lines.
+
+Similarly, a coroutine created by `async` can be canceled, `val deferred = async { ... }`, `deferred.cancel()`.
```
dkhalanskyjb: This is missing a mention of `Deferred` inheriting from a `Job`. I think it's valuable to highlight that in some form, as nothing else tells even an attentive reader that cancellation is exactly same concept for `async` and `launch`—they could end up as disjointed concepts.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -163,9 +140,9 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
 ## Make coroutines react to cancellation {id="cancellation-is-cooperative"}
 
 In Kotlin, coroutine cancellation is _cooperative_.
-This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation), [checking for cancellation explicitly](#check-for-cancellation-explicitly), or using the special case [`yield()`](#yield-often-in-non-suspending-code) function.
```
dkhalanskyjb: `yield` performs a suspension, so it falls into the first category, adding it explicitly is excessive. There is a tiny bit of special behavior to `yield()` when it's used with `Dispatchers.Unconfined`, but that should be irrelevant to the discussion in a general-strokes description.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -163,9 +140,9 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
...
 
-In this section, you can learn how to create cancelable coroutines.
+Coroutines which react to cancellations are sometimes referred to as cancelable coroutines.
```
dkhalanskyjb: The idea that this section teaches writing coroutines that react to cancellation looks important to me. It justifies the title and prepares the reader for what's to come. If you dislike using the phrase "cancelable coroutine" without giving it a precise definition, we could try avoiding the phrase altogether.

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: Good call. I didn't think this through during the initial discussion of the document, but indeed, focusing on `isActive` and `ensureActive` doesn't make as much sense as it used to. See, before prompt cancellation, you used to have to call `ensureActive()` in this scenario:
```kotlin
withContext(Dispatchers.Main) {
  withContext(Dispatchers.IO) {
      // compute the UI update
  }
  ensureActive()
  updateUi() // can throw if the screen was disposed of
}
```
Now, this `ensureActive` happens automatically, so its role is significantly less prominent.

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: In my view, it's the exact opposite. `0`, `1`, or `2` can reasonably come up during a real-world program execution, so the eye is accustomed to them. `5` or `14` jump out at you. If you see `14` in the output, you can quickly find the corresponding `14` in the code and figure out where it came from.

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: Instead of "mitigating slow connections", doesn't it make more sense to say "react to network failures"? The mitigation for a slow connection is patience and increasing the timeouts. Timeouts can help against dropped packets or being blocked by a firewall.

docs/topics/coroutines-cancellation.md
dkhalanskyjb: The reason it's relevant is that without prompt cancellation, `updateUi` can be called on a disposed screen and crash the program.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -372,20 +261,20 @@ On the JVM, some functions, such as `Thread.sleep()` or `BlockingQueue.take()`,
...
 ```kotlin
 import kotlinx.coroutines.*
 
 //sampleStart
 suspend fun main() {
-        val childStarted = CompletableDeferred<Unit>()
-        val childJob = launch {
+        val started = CompletableDeferred<Unit>()
+        val job = launch {
```
dkhalanskyjb: `childJob` + `childStarted` look better to me, as `job` looks too similar to `coroutineContext.job`, which is entirely different and is the `Job` of the currently executing coroutine. `childJob` here does look inconsistent with other examples, though, so maybe we could specify `child` in those as well?

docs/topics/cancellation-and-timeouts.md
```diff
@@ -231,109 +210,19 @@ suspend fun main() {
...
-    }
-}
-```
-{kotlin-runnable="true" id="ensurective-example"}
-
-#### yield()
-
-The [`yield()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html) function suspends the coroutine and checks for cancellation before resuming.
-Without suspending, coroutines on the same thread run sequentially.
+Thus `yield()` achieves two goals: frees up resources to allow others to run, and allows cancellations to propagate.
```
dkhalanskyjb: "Resources" isn't the best wording, as it often refers to locks or files, and we don't want `yield()` to release either of those. The paragraph just below is the more precise and useful statement to the same effect. Also, cancellation propagation between coroutines happens without `yield`, `yield` only allows *reacting* to cancellation in a specific coroutine to which cancellation was already propagated. The goal of this paragraph is unclear to me, it looks like it only reiterates what was already said, but in less precise and useful terms.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -231,109 +210,19 @@ suspend fun main() {
...
-> You can use the [`cancelAndJoin()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/cancel-and-join.html) function to cancel a coroutine and wait for it to finish in a single call.
->
-{style="note"}
+### `yield()` often in non-suspending code
 
-#### ensureActive()
+In CPU-intensive computations and suspending code which is unlikely to suspend, call `yield()` periodically to allow cancellations to propagate.
 
-Use the [`ensureActive()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/ensure-active.html) function to check for cancellation and stop the current computation by throwing `CancellationException` if the coroutine is canceled:
+The [`yield()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html) function suspends the coroutine, releasing the thread and giving other coroutines a chance to run on it. Suspending the coroutine lets it check for cancellation and throw `CancellationException` if it's canceled. Without suspending, coroutines on the same thread run sequentially.
```
dkhalanskyjb: The addition of "releasing the thread" is very welcome!

The "Without suspending" sentence, both here and in the current version, feels a bit excessive to me, as it's focusing on a single-threaded scenario in the same paragraph where the basic function of `yield` is described. While we're at it, we could either remove this sentence or make it a separate paragraph where we would also mention multi-threaded contexts and explain that, say, `Dispatchers.Default`'s threads can all be taken, and `yield()` is a way to let other coroutines leave the queue earlier.

dkhalanskyjb (changes requested):

docs/topics/cancellation-and-timeouts.md
murfel: Kept the definition AND the original sentence. I don't like this phrase but it's already used, so we might as well demystify it.

docs/topics/cancellation-and-timeouts.md
murfel: Changed the wording. The goal is to highlight that `yield` does two things.

docs/topics/cancellation-and-timeouts.md
murfel: Yep, reworded and added `runInterruptible`. I meant to summarise all subsections in this section.

docs/topics/cancellation-and-timeouts.md
murfel: Agreed, removed for now. A separate paragraph might be quite useful. Although, I refrained from it for now because this section feels already a bit awkward: `yield` primary goal is to release resources and reacting to cancellation is a secondary benefit. Yet from the beginning we advertise it as "use it to react to cancellations" and then give an example where there's no cancellations at all. We don't lie, but we jump from point to point.

I'd keep it as is for now, unless someone wants to take a go at rewriting the whole section.

docs/topics/cancellation-and-timeouts.md
murfel: > `childJob` + `childStarted` look better to me, as `job` looks too similar to `coroutineContext.job`, which is entirely different and is the `Job` of the currently executing coroutine.

I don't have this impression at all. Moreover, we haven't covered `coroutineContext.job` up to here. We also use `job` as a variable name in the very first example and talk about `Job` as the return type of `launch`.

Now that you mentioned it, I can see how it could be confusing. But I don't think the intended audience going to have this problem, and for simplicity I'd like to stick with `job`.

I went over examples and I suggest this rule: always call it a `job`, but when parent-child relationship are important, call it `parentJob` or `childJobs`.

murfel: Another argument against a lonely `childJob`: when I hide these bits of the examples to fight indentation on the playground, it will no longer make sense whose child it is. It's just a coroutine now.

```
suspend fun main() {
        ...
    }
}
```

docs/topics/cancellation-and-timeouts.md
murfel: Yeah, it's all a bit too generic. I meant that if you detect a slow connection via a timeout, you can choose to walk away from it and use cache instead. Reworded.

I don't get the react to network failures example. Isn't catching an exception a reaction to network failures?

docs/topics/cancellation-and-timeouts.md
murfel: Agree, but in this case there's also `300`, `15`, `100`, `100`, and the combination of all numbers makes it overwhelming. Replaced with A and B.

docs/topics/coroutines-cancellation.md
murfel: Say `withContext` returned `contents` correctly. Now `updateUi(contents)` is about to start executing and the scope is cancelled. `updateUI` is not a suspend function, so it won't react to a cancellation and attempt to update the UI elements anyways past their disposal time. No?

dkhalanskyjb: Only the UI thread is allowed to remove UI components and cancel the coroutine scopes related to those components. If the scope isn't cancelled after `withContext` returns, it can't get cancelled before the `updateUi` call, because `updateUi` has the sole ownership of the thread. Is this the information you were missing?

murfel: YES!

docs/topics/coroutines-cancellation.md
murfel: `withContext` as well (almost everywhere, unless it is important which dispatcher we use), so these will be hidden:

```
suspend fun main() {
    }
}
```

docs/topics/coroutines-cancellation.md
murfel: I want to take a level of abstraction to avoid drawing attention to irrelevant details, but also make the UI elements and the controller more specific. Here's my sketch, feel free to redo it or just comment. Possibly we need to transfer the comments into text, given its amount, and keep the example tiny.

```kotlin
// Defines a coroutine scope that uses the UI thread
class ScreenWithButtons(private val scope: CoroutineScope) {
    fun loadAndUpdateButtons(filename: String) {
        scope.launch {
            val buttonNames = withContext(Dispatchers.IO) {
                readLines(filename) // suspend function, can be canceled
            }
            // If withContext() returned values, it wasn't canceled.
            // Now this coroutine has the UI thread,
            // so no one can cancel this component's scope and dispose of buttons
            // until this coroutine suspends for the next time.
            // It's safe to call updateUi here, as buttons are guaranteed to exist.
            //
            // If the screen is cancelled while withContext() is executing,
            // the control will never reach the assignment of withContext return value
            // to buttonNames, and so updateUi will never be executed, either.
            updateUi(buttonNames)
        }
    }

    private fun updateUi(buttonNames: List<String>) {
        // Updates buttons with the specified names.

        // Throws if buttons no longer exist,
        // due to being disposed of
        // after the user left the screen.
    }

    // Only callable from the UI thread
    fun leaveScreen() {
        // Cancels the scope when leaving the screen
        // You can no longer update the UI
        scope.cancel()
    }
}

// UI controller code
setHandler(Event.ScreenClosed) {
    // Always executing from the UI thread
    withContext(Dispatchers.Main) {
        screenWithButtons.leaveScreen()
        buttons.dispose()
    }
}
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -163,9 +141,9 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
 ## Make coroutines react to cancellation {id="cancellation-is-cooperative"}
 
 In Kotlin, coroutine cancellation is _cooperative_.
-This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly). We will also cover some typical usage patterns such as periodically calling the [`yield()`](#yield-often-in-non-suspending-code) suspending function in CPU-intensive code, and using [`runInterruptible()`](#interrupt-blocking-code-when-coroutines-are-canceled) to interrupt blocking code when coroutines are canceled.
```
dkhalanskyjb: [Minor, style] The sentences aren't in agreement syntactically: "Coroutines react to cancellation in specific cases. We will also describe other things." The transition can be improved: "Coroutines react to cancellation in specific cases. There are standard approaches for making coroutines cancellable: ..."

docs/topics/cancellation-and-timeouts.md
```diff
@@ -231,109 +211,19 @@ suspend fun main() {
...
-        // Waits until the sorting coroutine finishes
-        // before accessing the shared list to avoid data races
-        listSortingJob.join()
-        println("The list is probably sorted: $unsortedList")
-    }
-}
-//sampleEnd
-```
-{kotlin-runnable="true" id="isactive-example"}
+In certain scenarios, you may use the [`isActive`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/is-active.html) property or the [`ensureActive()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/ensure-active.html) function to check if the coroutine was canceled. `isActive` is `false` when the coroutine is canceled and `ensureActive()` throws `CancellationException` if the coroutine is canceled.
```
dkhalanskyjb: [Minor, style] It's surprising to me that this subsection is above `yield()`, since we agreed on treating `yield()` as more important. As far as I can see, there is no dependency between the two sections, so we may order them from most to least significant freely.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -231,109 +211,19 @@ suspend fun main() {
...
-```
-{kotlin-runnable="true" id="ensurective-example"}
-
-#### yield()
-
-The [`yield()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html) function suspends the coroutine and checks for cancellation before resuming.
-Without suspending, coroutines on the same thread run sequentially.
+Thus `yield()` achieves two goals: lets other coroutines run, and allows checking for cancellations.
 
 Use `yield` to allow other coroutines to run on the same thread or thread pool before one of them finishes:
```
dkhalanskyjb: [Minor, style] Looks like redundancy in directly neighboring sentences? The `Use yield` sentence may be removed without losing anything, it feels like.

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: It's also important that there is no general way to deal with blocking code—`runInterruptible` only deals with *some* specific instances of blocking code, and only on the JVM.

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: > I don't think the intended audience going to have this problem, and for simplicity I'd like to stick with `job`.

It looks like we view our audience differently, then. I'd estimate that quite a large portion of the audience doesn't start new projects from scratch but gets onboarded into an existing project, tries working with `kotlinx.coroutines` there, gives up and goes to tutorials. This means that they are likely to have some chaotic exposure to unpredictable parts of our API.

> when I hide these bits of the examples to fight indentation on the playground, it will no longer make sense whose child it is. It's just a coroutine now.

Any `launch` invocation without an explicit receiver is the child of the caller, by convention. If you see `something; launch { somethingElse() }` anywhere, this is a child coroutine (or a stylistic mistake!).

docs/topics/cancellation-and-timeouts.md
```diff
@@ -580,8 +471,7 @@ suspend fun main() {
 
 ## Timeout
 
-Timeouts allow you to automatically cancel a coroutine after a specified duration.
-They are useful for stopping operations that take too long, helping to keep your application responsive and avoid blocking threads unnecessarily.
+A timeout allows you to automatically cancel a coroutine after a specified duration. It is useful for stopping operations that take too long. For example, if a request to download a picture from a server times-out, you can choose to retry, or fallback to the local cache.
```
dkhalanskyjb: [Minor, style] It's preferable to keep sentences on separate lines, to help git tooling determine what changed when. It's line-based by default in most cases.

docs/topics/coroutines-cancellation.md
dkhalanskyjb: > `                readLines(filename) // suspend function, can be canceled`

That's probably wrong. In most cases, this exact code shape would contain a blocking call, because otherwise, why wouldn't `readLines` itself perform `withContext(Dispatchers.IO)` instead of requiring that from the caller?

In this specific scenario, the example would actually become weaker if `readLines` were a `suspend` function, because then, the race this section describes would trigger less often.

> `            // If the screen is cancelled while withContext() is executing,`

This is restating the same thing twice in a contrapositive. That can be helpful if you preface the second part with `In other words, ` or something like that, so that people know not to look for new information there.

Other than that, thanks, this is an improvement w.r.t. clarity!

dkhalanskyjb (commented):
I think most technical concerns have been resolved, so we can ask for the writer team's review next. Or maybe ask *them* to rephrase this instead? This could end up being more streamlined in the end.

docs/topics/cancellation-and-timeouts.md
murfel: Agreed on the audience. Semantically, it doesn't make sense to indicate that it is a child, if we don't refer to a particular parent or trying to stress that certain coroutines are siblings, etc. If it's always a child, it doesn't matter that it's a child.

docs/topics/coroutines-cancellation.md
murfel: Edited that, and added some more hints.

docs/topics/cancellation-and-timeouts.md
murfel: Will apply to the whole file after all approvals.

docs/topics/cancellation-and-timeouts.md
murfel: All in all, you want to eradicate confusion of `val job` vs `Job`, I want simple names for readability.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -2,13 +2,10 @@
...
 
-Cancellation lets you stop a coroutine before it completes.
+Cancellation lets you request to stop a coroutine before it completes.
 It stops work that's no longer needed, such as when a user closes a window or navigates away in a user interface while a coroutine is still running.
-You can also use it to release resources early and to stop a coroutine from accessing objects past their disposal.
-
-> You can use cancellation to stop long-running coroutines that keep producing values even after other coroutines no longer need them, for example, in [pipelines](channels.md#pipelines).
->
-{style="tip"}
+You can use it to release resources early and to stop a coroutine from accessing objects past their disposal.
```
daniCsorbaJB:
```suggestion

You can use cancellation to release resources early and to stop a coroutine from accessing objects past their disposal.
```

daniCsorbaJB: just an extra line break to break it up and then add cancellation to the sentence for clarity

docs/topics/cancellation-and-timeouts.md
```diff
@@ -2,13 +2,10 @@
...
+You can also use cancellation to stop long-running coroutines, for example, sending heartbeats, running scheduled tasks, updating a state to reflect the newest reading (like the clock UI), etc.
```
daniCsorbaJB:
```suggestion
You can also use it to stop long-running coroutines that perform repeated work, for example:

* Sending heartbeats
* Running scheduled tasks
* Updating a state to reflect the newest reading, such as in a clock UI
```

daniCsorbaJB: Not sure if the repeated work part helps, but these all feel like it fits that criterion 🤔

docs/topics/coroutines-cancellation.md
```diff
@@ -24,13 +21,11 @@ implements `Job` and supports the same cancellation behavior.
...
 
 When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) the next time it checks for cancellation.
-For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
+The [Suspension points and cancellation](#suspension-points-and-cancellation) section covers precisely when this happens, but for now it is enough to know that all suspend functions in the `kotlinx.coroutines` library, such as `delay()` and `awaitCancellation()` explained below, check for cancellation if they suspend.
 
-> You can use the [`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function to suspend a coroutine until it's canceled.
->
-{style="tip"}
+[`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function suspends a coroutine until it's canceled. It is equivalent to `delay(Duration.INFINITE)`.
```
daniCsorbaJB:
```suggestion
Suspending functions in the `kotlinx.coroutines` library, such as the `delay()` function, check for cancellation when they suspend.

> For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
>
{style="tip"}

You can use the [`awaitCancellation()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/await-cancellation.html) function to suspend a coroutine until it's canceled. It's equivalent to calling `delay(Duration.INFINITE)`.
```

docs/topics/coroutines-cancellation.md
daniCsorbaJB: I see some consistency issues in hiding `suspend fun main()`, but it's not a big concern, and I think readability is more important, so let's go for it. 👍 

I'm not certain on the `withContext()` part, though 🤔 Wasn't the original intention here to ensure that people see at a glance which dispatcher is used in the examples? (alongside with the examples always specifying a dispatcher explicitly — that's not a concern here) 🤔 
if that's not a big issue, I think we can hide that as well behind sampleStart.

docs/topics/coroutines-cancellation.md
```diff
@@ -24,13 +21,11 @@ implements `Job` and supports the same cancellation behavior.
 You can call the `cancel()` function manually, or it can be invoked automatically through cancellation propagation when a parent coroutine is canceled.
 
 When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) the next time it checks for cancellation.
-For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
+The [Suspension points and cancellation](#suspension-points-and-cancellation) section covers precisely when this happens, but for now it is enough to know that all suspend functions in the `kotlinx.coroutines` library, such as `delay()` and `awaitCancellation()` explained below, check for cancellation if they suspend.
```
daniCsorbaJB: As a general comment: 
The docs purposefully avoid similar phrases like: explained below — these are quite fragile, because the surrounding structure can change, and don't add much to the reader, since below is not exact. (a link is exact)

I think the original is more concise and better captures the overall style, which should be something the readers are already used to. The information you are proposing is a great addition I added a [suggestion](https://github.com/Kotlin/kotlinx.coroutines/pull/4618/changes#r3511386564) how I think we could add it. 👍

docs/topics/coroutines-cancellation.md
daniCsorbaJB: I'd suggest using a tip to point readers to more information — it helps it pop and it's the same as saying "The user doesn't need to know this now, and they can read more now if they prefer"

docs/topics/cancellation-and-timeouts.md
```diff
@@ -40,19 +35,18 @@ import kotlin.time.Duration
...
-            
+        val job: Job = launch {
             println("The coroutine has started")
 
             // Completes the CompletableDeferred,
             // signaling that the coroutine has started running
-            job1Started.complete(Unit)
+            started.complete(Unit)
             try {
                 // Suspends indefinitely
```
daniCsorbaJB:
```suggestion

```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -40,19 +35,18 @@ import kotlin.time.Duration
...
-                // Without cancellation, this call would never return
-                delay(Duration.INFINITE)
+                // This call will never return unless the coroutine is canceled
```
daniCsorbaJB:
```suggestion
                // Suspends until the coroutine is canceled
```

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB:
```suggestion
```

daniCsorbaJB: I think we can delete the Duration import 🤔

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: WDYT would this work instead?

General note: We should avoid future tense unless necessary. The present tense is usually clearer because it describes behavior directly. Future tense markers like "will" can sometimes make behavior sound less certain, and they often add extra unnecessary words.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -96,8 +73,9 @@ suspend fun main() {
 {kotlin-runnable="true" id="manual-cancellation-example"}
 
 In this example, [`CompletableDeferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/) is used as a signal that the coroutine has started running.
-The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed. This way, cancellation happens only after the coroutine has started running.
-The coroutine created by `.async()` doesn't have this check, so it may be canceled before it can run the code inside its block.
+The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed. This way, cancellation happens only after the coroutine has started running. Without this check, the coroutine could be canceled before it runs the code inside its block. You don't need to have this check to cancel a coroutine, but it is included here to have a consistent example which always prints the explanation lines.
```
daniCsorbaJB:
```suggestion
The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed. This way, cancellation happens only after the coroutine has started running. Without this check, the coroutine may be canceled before it runs the code inside its block. You don't need this check with `CompletableDeferred` to cancel a coroutine. It's included here to make the example consistent, so the coroutine always starts and prints its messages before it's canceled.
```

docs/topics/cancellation-and-timeouts.md
```diff
@@ -96,8 +73,9 @@ suspend fun main() {
...
+
+Similarly, a coroutine created by `async` can be canceled, `val deferred = async { ... }`, `deferred.cancel()`. The `async` builder returns a `Deferred` handle, which inherits from `Job`. Hence, the cancellation works in exactly the same way for `Deferred` as it does for `Job`.
```
daniCsorbaJB:
```suggestion
```

daniCsorbaJB: I think we can remove these lines, because we already covered this concept at the beginning:

> The [.async()](https://github.com/Kotlin/kotlinx.coroutines/blob/9746d58d1e8f267bb81592fbf2a2e094e665f570/docs/topics/coroutines-basics.md#coroutinescope-async) function returns a [Deferred](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/), which implements Job and supports the same cancellation behavior.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -109,7 +87,7 @@ The coroutine created by `.async()` doesn't have this check, so it may be cancel
 ### Cancellation propagation
 
 [Structured concurrency](coroutines-basics.md#coroutine-scope-and-structured-concurrency) ensures that canceling a coroutine also cancels all of its children.
-This prevents child coroutines from working after the parent has already stopped.
+This prevents child coroutines from working after the parent has been requested to stop.
```
daniCsorbaJB:
```suggestion
This prevents child coroutines from continuing their work after their parent coroutine is canceled.
```

daniCsorbaJB: (suggestion)
I think requested to stop works, it's a nod back to the change at the beginning, but reading it this feels more natural - wdyt?

docs/topics/coroutines-cancellation.md
```diff
@@ -163,9 +141,12 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
 ## Make coroutines react to cancellation {id="cancellation-is-cooperative"}
 
 In Kotlin, coroutine cancellation is _cooperative_.
-This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+Coroutines react to cancellation only when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
```
daniCsorbaJB, line 160: not sure if we have to remove "This means" here. it reinforces that this is a new concept for readers. that said I don't feel strongly about this, I think it works without it as well. why did you decide to remove it? 🤔

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
-                }
+            // withContext will check for cancellation before entering the block
+            // and after the block returns
+            val buttonNames = withContext(Dispatchers.IO) {
+                readLines(filename) // A blocking call, cannot be canceled here
             }
-            // It's safe to call updateUi here,
-            // In case of cancellation, withContext() wouldn't return any values
-            updateUi(contents)
+            // If withContext() returned a value, it wasn't canceled.
```
daniCsorbaJB: This comment is very long. 😢  I'd suggest creating a description before the example.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -580,8 +482,7 @@ suspend fun main() {
 
 ## Timeout
 
-Timeouts allow you to automatically cancel a coroutine after a specified duration.
-They are useful for stopping operations that take too long, helping to keep your application responsive and avoid blocking threads unnecessarily.
+A timeout allows you to automatically cancel a coroutine after a specified duration. It is useful for stopping operations that take too long. For example, if a request to download a picture from a server times-out, you can choose to retry, or fallback to the local cache.
```
daniCsorbaJB:
```suggestion
A timeout allows you to automatically cancel a coroutine after a specified duration. You can use it to stop operations that take too long. 
For example, if a request to download an image from a server times out, you can retry the request or fall back to the local cache.
```

docs/topics/coroutines-cancellation.md
```diff
@@ -163,9 +141,12 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
...
 
 In Kotlin, coroutine cancellation is _cooperative_.
-This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+Coroutines react to cancellation only when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+There are standard approaches for making coroutines cancelable, such as periodically calling the [`yield()`](#yield-often-in-non-suspending-code) suspending function in CPU-intensive code.
+On the JVM, you can use [`runInterruptible()`](#interrupt-blocking-code-when-coroutines-are-canceled) with some blocking calls that support thread interruption.
 
-In this section, you can learn how to create cancelable coroutines.
+Coroutines that react to cancellation are sometimes referred to as **cancelable coroutines**.
+In this section, you will learn how to create cancelable coroutines.
```
daniCsorbaJB:
```suggestion
In this section, you can learn how to create cancelable coroutines.
```

daniCsorbaJB: I think that if there is an issue with the original sentence, then the right approach is to remove it. Replacing "can" with "will" makes this read more like a course/tutorial, so I would strongly advise against it. I have no issue with the original, but I believe it doesn't add much, so feel free to remove.

docs/topics/coroutines-cancellation.md
```diff
@@ -229,114 +212,18 @@ suspend fun main() {
 >
 {style="tip"}
 
-### Check for cancellation explicitly
-
-If a coroutine doesn't [suspend](#suspension-points-and-cancellation) for a long time, it doesn't stop when it's canceled unless it explicitly checks for cancellation.
+### `yield()` often in non-suspending code
```
daniCsorbaJB:
```suggestion
### The `yield()` suspending function
```

docs/topics/coroutines-cancellation.md
```diff
@@ -163,9 +141,12 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
 ## Make coroutines react to cancellation {id="cancellation-is-cooperative"}
 
 In Kotlin, coroutine cancellation is _cooperative_.
-This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+Coroutines react to cancellation only when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+There are standard approaches for making coroutines cancelable, such as periodically calling the [`yield()`](#yield-often-in-non-suspending-code) suspending function in CPU-intensive code.
```
daniCsorbaJB: I find the logic strange here and I much prefer the original. I understand we want to decrease the relevance of ensureActive and isActive, so I suggest removing the link 🤔 — alternatively, we could integrate the yield function in the title and cover it there, and keep the link.

I think this section should serve these functions:
* Explain what cooperative cancellation is as a concept (since there are resources that link here)
* (optional, we can drop this) Let the reader know what they can learn in the subsections

To achieve this I think we have two ways. Either mentioning the subsections briefly in a sentence or add a generic sentence that encompasses all of them like we had before:

> In this section, you can learn how to make coroutines react to cancellation using suspension points, the `yield()` function, explicit cancellation checks, and the `runInterruptible()` function on the JVM.

daniCsorbaJB: A related question:
- Do we want to describe yield as adding a suspension point or as an explicit check? 🤔 To me it kind of feels like both, but please correct me if I'm wrong.

I'm asking because if we could integrate it in "Check for cancellation explicitly" it would make this part easier to grasp, but if it's not correct then let's keep it as a separate one.

docs/topics/coroutines-cancellation.md
```diff
@@ -366,26 +253,33 @@ fun main() {
 
 In this example, each coroutine uses `yield()` to let other coroutines run between iterations.
 
+### Check for cancellation explicitly
+
+You almost never need to check for cancellation explicitly. In most cases, you can use `yield()`.
```
daniCsorbaJB: This is feels really vague 🤔 I understand the intention, but this needs more context. Why shouldn't the user check for it? What are the scenarios when the users might need it? 

Depending on what's your opinion on [this](https://github.com/Kotlin/kotlinx.coroutines/pull/4618/changes#r3520549879): I believe we should rewrite this section.

Alternatively, we could sidestep it, if we want to make it an official recommendation. 

> We don't recommend checking for cancellation explicitly unless...

 (if our stance is strong and we believe it warrants it)

Or something like: 
> Avoid checking for cancellation explicitly unless you need to check for cancellation without suspending.

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
-import java.nio.file.*
-import java.nio.charset.*
-import kotlinx.coroutines.*
-import java.io.*
-
 // Defines a coroutine scope that uses the UI thread
-class ScreenWithFileContents(private val scope: CoroutineScope) {
-    fun displayFile(path: Path) {
+class ScreenWithButtons(private val scope: CoroutineScope) {
+    // Should only be called from the UI thread
```
daniCsorbaJB: general note: 
I suggest never using should. It's not exact, so it either lacks information or weakens the sentence. It should either explain why shouldn't the user do it, or if it doesn't matter, omit it.

docs/topics/cancellation-and-timeouts.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
+    // Should only be called from the UI thread
+    fun loadAndUpdateButtons(filename: String) {
         scope.launch {
-            val contents = withContext(Dispatchers.IO) {
-                Files.newBufferedReader(
-                    path, Charset.forName("US-ASCII")
-                ).use {
-                    it.readLines()
-                }
+            // withContext will check for cancellation before entering the block
```
daniCsorbaJB:
```suggestion
            // withContext() checks for cancellation before entering the block
```

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
-            val contents = withContext(Dispatchers.IO) {
-                Files.newBufferedReader(
-                    path, Charset.forName("US-ASCII")
-                ).use {
-                    it.readLines()
-                }
+            // withContext will check for cancellation before entering the block
+            // and after the block returns
+            val buttonNames = withContext(Dispatchers.IO) {
+                readLines(filename) // A blocking call, cannot be canceled here
```
daniCsorbaJB:
```suggestion
                // This is a blocking call, and can't be canceled here
                readLines(filename)
```

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
+            // It's safe to call updateUi here, as buttons are guaranteed to exist.
+            //
+            // In other words, if the screen is canceled while withContext() is running,
+            // the coroutine throws CancellationException instead of assigning a value
+            // to buttonNames, so updateUi() is never called.
+            updateUi(buttonNames)
         }
     }
 
+    // Should only be called from the UI thread
```
daniCsorbaJB:
```suggestion
```

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
-    private fun updateUi(contents: List<String>) {
-      contents.forEach { line -> addOneLineToUi(line) }
-    }
-  
-    private fun addOneLineToUi(line: String) {
-        // Placeholder for code that adds one line to the UI
+    private fun updateUi(buttonNames: List<String>) {
+        // Updates buttons with the specified names.
+
+        // Throws if buttons no longer exist,
```
daniCsorbaJB:
```suggestion
        // Throws an exception if the buttons no longer exist
```

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
+        // due to being disposed of
```
daniCsorbaJB:
```suggestion
        // because they are disposed of after the screen closes
```

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
+        // after the user left the screen.
```
daniCsorbaJB:
```suggestion
```

docs/topics/coroutines-cancellation.md
```diff
@@ -419,48 +312,57 @@ It prevents your code from continuing in a canceled coroutine's scope, such as u
...
     fun leaveScreen() {
         // Cancels the scope when leaving the screen
         // You can no longer update the UI
         scope.cancel()
     }
 }
+
+// UI controller code
+setHandler(Event.ScreenClosed) {
+    // Always executes from the UI thread
```
daniCsorbaJB:
```suggestion
    // Runs on the UI thread
```

daniCsorbaJB (commented):
Thank you for the additions I left some comments. I couldn't finish the yield() section because of some open questions from my side.

docs/topics/cancellation-and-timeouts.md
dkhalanskyjb: > Semantically, it doesn't make sense to indicate that it is a child, if we don't refer to a particular parent or trying to stress that certain coroutines are siblings, etc. If it's always a child, it doesn't matter that it's a child.

A job is not always a child:

```kotlin
coroutineScope {
    // The *current* job (may or may not be someone's child),
    // but in any case, it's not the current coroutine's child
    val job = currentCoroutineContext().job
}

coroutineScope {
    // The *child* job of the current coroutine
    val job = launch { }
}
```

Whether a job is the job of the currently running coroutine, its child, or someone else's job is important in practice, and I'd use `job`, `childJob`, and `otherJob` for these scenarios, respectively.

docs/topics/cancellation-and-timeouts.md
murfel: It's only an intro, I don't want to draw attention to this by using a list.

docs/topics/coroutines-cancellation.md
murfel: This potential hedging interferes with the flow of the text. There isn't much value in a discipline to fix an unlikely potential problem if it's hurting the immediate value. Moreover, the surrounding structure is unlikely to change. These docs will probably never be updated in future given where the world is headed. Even without the world, these docs saw little attention in past. Re your suggestion - please read my comment at the same line as your suggestion.

docs/topics/coroutines-cancellation.md
murfel: (Please read my comment at the same line)

docs/topics/cancellation-and-timeouts.md
murfel: Agreed with the general idea, yet here "will never return" is used to explain what "suspends indefinitely" means.

docs/topics/cancellation-and-timeouts.md
murfel: That's misleading, you don't need this check to cancel a coroutine regardless of CompletableDerrefed.

docs/topics/cancellation-and-timeouts.md
murfel: Dmitry asked me to include these lines, since I removed the identical example with async, and I agree that it's nice to re-iterate this and show the syntax.

docs/topics/coroutines-cancellation.md
murfel: Cursive reinforces the newness of the concept. "This means" is hence unnecessary.

docs/topics/coroutines-cancellation.md
murfel: It is vague, yet explaining this will go into too much boring or complicated detail. Basically, use yield. It is prescribed as a spell here.

docs/topics/coroutines-cancellation.md
murfel: Yield is both, and it is described as both, so no problem here.

murfel: I don't understand your first comment, but the suggested text is not correct, since yield is both.

murfel:
This is the final look. Any last minute comments? Please note I'll spend no more than 10 minutes reading and fixing.  I'm merging ~Monday.

docs/topics/coroutines-cancellation.md
daniCsorbaJB: I'm raising this again because I agree with your sentiment, but this line doesn't fit this though:

> The [Suspension points and cancellation](https://github.com/Kotlin/kotlinx.coroutines/blob/c74267417d125cb61bcd2be189d1b68cd48fab46/docs/topics/cancellation-and-timeouts.md#suspension-points-and-cancellation) section covers precisely when this happens, but for now it is enough to know that all suspend functions in the kotlinx.coroutines library, such as delay() and awaitCancellation() explained below, check for cancellation if they suspend.

This parses really strangely: It's too long, so it's difficult to follow. It also mixes in information that's neither described nor necessary to understand the example or the section.

We should bring this into focus, as this is the important part to understand the example and section:

> Suspending functions in the `kotlinx.coroutines` library, such as the `delay()` function, check for cancellation when they suspend.

And this is a link where the reader can learn more (at this point, optional content, which is the perfect use case for a tip): 

> For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).

docs/topics/coroutines-cancellation.md
daniCsorbaJB: I'm not sure I understand the stance we are taking here.

What I'm saying is that our approach should depend on whether we want to officially recommend against it or not. If we have a (strong) opinion either way, we should state it explicitly. 

"You almost never need something" is a strange approach. There are many things we almost never need, yet we would like a description of how to do them.

docs/topics/coroutines-cancellation.md
daniCsorbaJB: My issue with this version is that it adds redundant information. 

To make it a bit more focused, I'd propose two options

`Option #1`:
Keep the beginning (except change this part to bring the "react" part into focus

> There are standard approaches for making coroutines react to cancellation, such as periodically calling the [`yield()`](#yield-often-in-non-suspending-code) suspending function in CPU-intensive code.

then remove everything from below this and replace with:

> In this section, you can learn how these approaches work and when to use each one.

`Option #2`:

Remove this sentence entirely: "There are standard approaches for making coroutines react to cancellation...."

and just remove the explicit checking from the short intro, putting the focus on yield() and suspension points.

> In this section, you can learn how to make coroutines react to cancellation using suspension points, the `yield()` function, and the `runInterruptible()` function on the JVM.

docs/topics/coroutines-cancellation.md
daniCsorbaJB: I still think a link is more exact than "below". If a reader wants to read more about something now, we should point them to the relevant section.

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: Hmm 🤔 I don't think we need suspends indefintely though. "Suspends until the coroutine is canceled" describes what's happening. 

If we want to keep it for clarity though let's describe the behavior with present tense: 

> // This call never returns unless the coroutine is canceled

docs/topics/cancellation-and-timeouts.md
daniCsorbaJB: 🤔 good point. I think without the example we could show the syntax, but I'd suggest a very minimal example instead. 

Similarly, you can manually cancel a coroutine created by the `async()` builder function:

```kotlin
val deferred = async { /* ... */ }
deferred.cancel()
```

docs/topics/coroutines-cancellation.md
```diff
@@ -229,184 +217,94 @@ suspend fun main() {
...
-                ++i
+// runBlocking uses the current thread for running all coroutines
+runBlocking {
+    val coroutineCount = 5
+    repeat(coroutineCount) { coroutineIndex ->
+        launch {
+            val id = coroutineIndex + 1
+            repeat(5) { iterationIndex ->
+                val iteration = iterationIndex + 1
+                // Temporarily suspends to give other coroutines a chance to run
```
daniCsorbaJB:
```suggestion
                // Suspends temporarily to give other coroutines a chance to run
```

docs/topics/coroutines-cancellation.md
```diff
@@ -229,184 +217,94 @@ suspend fun main() {
...
-{kotlin-runnable="true" id="ensurective-example"}
-
-#### yield()
-
-The [`yield()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html) function suspends the coroutine and checks for cancellation before resuming.
-Without suspending, coroutines on the same thread run sequentially.
+{kotlin-runnable="true" id="yield-example"}
 
-Use `yield` to allow other coroutines to run on the same thread or thread pool before one of them finishes:
+In this example, each coroutine uses `yield()` to let other coroutines run between iterations.
```
daniCsorbaJB, line 269:
```suggestion
In this example, each coroutine uses the `yield()` function to let other coroutines run between iterations.
```

docs/topics/coroutines-cancellation.md
```diff
@@ -163,9 +146,12 @@ Here, `CompletableDeferred` signals that the child coroutines are launched befor
 ## Make coroutines react to cancellation {id="cancellation-is-cooperative"}
 
 In Kotlin, coroutine cancellation is _cooperative_.
-This means that coroutines only react to cancellation when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+Coroutines react to cancellation only when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
+There are standard approaches for making coroutines cancelable, such as periodically calling the [`yield()`](#yield-often-in-non-suspending-code) suspending function in CPU-intensive code.
+On the JVM, you can use [`runInterruptible()`](#interrupt-blocking-code-when-coroutines-are-canceled) with some blocking calls that support thread interruption.
 
-In this section, you can learn how to create cancelable coroutines.
+Coroutines that react to cancellation are sometimes referred to as **cancelable coroutines**.
```
daniCsorbaJB: Is there a reason to mention this? I suggest removing this line and modify the next sentence if you think the it sounds like a term 🤔 i.e.: In this section, you can learn how to make coroutines react to cancellation. — but as I said before I think we can safely remove that line as well

daniCsorbaJB (commented):
Adding some additional suggestions and comments

docs/topics/coroutines-cancellation.md
```diff
@@ -24,80 +21,66 @@ implements `Job` and supports the same cancellation behavior.
 You can call the `cancel()` function manually, or it can be invoked automatically through cancellation propagation when a parent coroutine is canceled.
 
 When a coroutine is canceled, it throws a [`CancellationException`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-cancellation-exception/) the next time it checks for cancellation.
-For more information about how and when this happens, see [Suspension points and cancellation](#suspension-points-and-cancellation).
+The [Suspension points and cancellation](#suspension-points-and-cancellation) section covers precisely when this happens, but for now it is enough to know that all suspend functions in the `kotlinx.coroutines` library, such as `delay()` and `awaitCancellation()` explained below, check for cancellation if they suspend.
```
daniCsorbaJB: Another general comment: the page uses contractions, so we should use them consistently. (it is -> it's | does not -> doesn't etc.)

dkhalanskyjb:
@murfel,

> This is the final look. Any last minute comments? Please note I'll spend no more than 10 minutes reading and fixing. I'm merging ~Monday.

It's not for us to decide when this gets merged. The only reason the technical writer team isn't listed as the GitHub code owner of the `docs/` directory is that we want to be able to accept PRs with minor changes without bothering them. In effect, they are code owners, though. They are responsible for the contents on https://kotlinlang.org/docs/ , it's their domain, and we don't have the right to push out a text there until it fully matches their standards.

If you are tired of working on this PR and/or busy with other tasks, that's completely understandable. In that case, it's okay to ask someone else to finish the job. If I understand the situation correctly, trying to push out a version the technical writer team doesn't fully approve of can (rightfully!) lead to the page getting delisted from the website until it can be fixed, and no one wants that—the page is quite useful, we often link to it!

@daniCsorbaJB, could you please pick this up? The process is to make a new branch with `cancellation-and-timeouts-review` as the base and introduce changes there; then, we'll merge your branch back to this one, and finally, this one to `master`.

daniCsorbaJB:
> @daniCsorbaJB, could you please pick this up? The process is to make a new branch with `cancellation-and-timeouts-review` as the base and introduce changes there; then, we'll merge your branch back to this one, and finally, this one to `master`.

Sure thing, on it! 👍 👀

PR #4619, opened 2026-02-17
Optimize attaching the debug probes

dkhalanskyjb:
I see a notable speed improvement even without any benchmarks, simply by running `DebugProbes.withDebugProbes { }` in a loop.

As I see, the change should be completely transparent for the user. `Class.forName` used to run with the class loader of the `ByteBuddyDynamicAttach` class, which itself is loaded by `ByteBuddyDynamicAttach`. This hasn't changed—there is still no other code that can access `ByteBuddyDynamicAttach`, and transitively, the `AttachKt` fields.

### DIFF
--- a/kotlinx-coroutines-debug/src/Attach.kt
+++ b/kotlinx-coroutines-debug/src/Attach.kt
@@ -16,23 +16,30 @@ internal class ByteBuddyDynamicAttach : Function1<Boolean, Unit> {
 
     private fun attach() {
         ByteBuddyAgent.install(ByteBuddyAgent.AttachmentProvider.ForEmulatedAttachment.INSTANCE)
-        val cl = Class.forName("kotlin.coroutines.jvm.internal.DebugProbesKt")
-        val cl2 = Class.forName("kotlinx.coroutines.debug.internal.DebugProbesKt")
-
-        ByteBuddy()
-            .redefine(cl2)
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-debug/src/Attach.kt
```diff
@@ -16,23 +16,30 @@ internal class ByteBuddyDynamicAttach : Function1<Boolean, Unit> {
...
-        ByteBuddy()
-            .redefine(cl2)
-            .name(cl.name)
-            .make()
-            .load(cl.classLoader, ClassReloadingStrategy.fromInstalledAgent())
+        dynamicTypeWithoutProbes.load(targetClassLoader, ClassReloadingStrategy.fromInstalledAgent())
     }
 }
+
+private val targetClassLoader = Class.forName(classNameToOverride).classLoader
```
qwwdfsad: These things probably belong to `ByteBuddyDynamicAttach`'s companion, potentially as `@JvmStatic` encapsulation-wise.

Beware of `targetClassLoader` computed eagerly in that case though

qwwdfsad (approved):
Nice catch!

Nit -- please consider explaining what has been done in the commit message

dkhalanskyjb:
@fzhinkin, are you still interested in reviewing this PR, or can we merge it?

PR #4620, opened 2026-02-18
Port Map-returning Sequence operators to Flow

### LINKED ISSUE #1541: Provide an `associate` terminal operators
Hi, 

Today we realized that despite the terminal operators `toList` and `toSet` being provided for `Flow`, there isn't anything to create a map from a `Flow`.

Could we have `associate`, `associateBy` and `associateWith` for `Flow`, as we have for `Sequence`?

dkhalanskyjb:
Fixes #1541


kotlinx-coroutines-core/common/src/flow/terminal/Collection.kt
```diff
@@ -24,3 +24,327 @@ public suspend fun <T, C : MutableCollection<in T>> Flow<T>.toCollection(destina
...
+ * ```
+ * val names = flowOf("Grace Hopper", "Jacob Bernoulli", "Johann Bernoulli")
+ *
+ * val byLastName = names.associate { it.split(" ").let { (firstName, lastName) -> lastName to firstName } }
+ *
+ * // Jacob Bernoulli does not occur in the map because only the last pair with the same key gets added
+ * println(byLastName) // {Hopper=Grace, Bernoulli=Johann}
+ * ```
+ */
+public suspend fun <T, K, V> Flow<T>.associate(transform: (T) -> Pair<K, V>): Map<K, V> =
```
whyoleg: Curious: why not make all lambdas here `crossinline` and `suspend`, and make the whole function `inline`? **In line** with other [transformation operators](https://github.com/Kotlin/kotlinx.coroutines/blob/5c4733a2bc4f59ca5cfb901a3358dcd50f3954ba/kotlinx-coroutines-core/common/src/flow/operators/Transform.kt).

So the declaration here will be something like:
```kotlin
public suspend inline fun <T, K, V> Flow<T>.associate(
  corssinline transform: suspend (T) -> Pair<K, V>
): Map<K, V>
```

Overall, apart from non-allocating lambda, it will allow calling `suspend` functions in `transform` block, which is not possible right now.

Though I see that not all declarations in coroutines follow this approach. For example, `fold` is `inline`, while `reduce` is not ([source](https://github.com/Kotlin/kotlinx.coroutines/blob/5c4733a2bc4f59ca5cfb901a3358dcd50f3954ba/kotlinx-coroutines-core/common/src/flow/terminal/Reduce.kt#L11-L44)).

(This applies to all other declarations in the file.)

dkhalanskyjb: Thanks for catching this!
I simply didn't know `crossinline` lambdas get inlined, but looking at the bytecode confirms it. This is a major factor.
I'm not sure what you mean by "not profitable" here, but I can imagine some use cases where a `suspend` `valueTransform` could be useful, so let's mark the lambdas as `suspend`, yes.

dkhalanskyjb:
Another question here.

I didn't mark these new APIs as experimental, since they looked obvious, but given that I didn't find the best API on the first try at least, that was overconfidence on my part. Maybe they aren't. However, (doubling down on overconfidence!) *now* I truly don't see any improvements left to be made. Does anyone foresee any potential changes we might theoretically want to introduce later?

kotlinx-coroutines-core/common/src/flow/terminal/Collection.kt
whyoleg: > I'm not sure what you mean by "not profitable" here

I've edited the message; I meant it's **not possible right now**, but it looks like, because of autocomplete/autoreplacement on my side, it became "profitable" for some reason. :)

kotlinx-coroutines-core/common/src/flow/terminal/Collection.kt
```diff
@@ -24,3 +23,341 @@ public suspend fun <T, C : MutableCollection<in T>> Flow<T>.toCollection(destina
...
+ *
+ * println(byLength.keys) // [1, 3, 2, 4]
+ * println(byLength.values) // [[a], [abc, def], [ab], [abcd]]
+ *
+ * val mutableByLength: MutableMap<Int, MutableList<String>> = words.groupByTo(mutableMapOf()) { it.length }
+ * // same content as in byLength map, but the map is mutable
+ * println("mutableByLength == byLength is ${mutableByLength == byLength}") // true
+ * ```
+ */
+public suspend inline fun <T, K> Flow<T>.groupBy(crossinline keySelector: (T) -> K): Map<K, List<T>> =
```
whyoleg:
```suggestion
public suspend inline fun <T, K> Flow<T>.groupBy(crossinline keySelector: suspend (T) -> K): Map<K, List<T>> =
```

One more `suspend` is missing. :)

whyoleg:
> Does anyone foresee any potential changes we might theoretically want to introduce later?

Apart from adding [@IgnorableReturnValue](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-ignorable-return-value/index.html) at some point (which will be safe), I don't.

murfel (approved):
No educated opinion on the experimental status from me.

PR #4624, opened 2026-02-23
Add plus operator functions to MDCContext

schielek:
Continuation of https://github.com/Kotlin/kotlinx.coroutines/pull/4601

See https://github.com/Kotlin/kotlinx.coroutines/issues/4600


schielek:
@dkhalanskyjb

Sorry, had to recreate PR to branch from dev

dkhalanskyjb:
Hi @schielek , I replied under your previous PR: https://github.com/Kotlin/kotlinx.coroutines/pull/4601#discussion_r2852435450, PTAL

schielek:
Hi @dkhalanskyjb. Thank you very much for your valuable feedback.

I would like to explore some API options with you and hear your preferred choice.
In the following simplified examples, I will focus solely on the API, not on the implementation.
E.g. the code will define constructors, where the arguments are unused.

## Secondary constructor
We discussed and rejected this already, I still wanted to keep it for completeness.

```kotlin
class MDCContext(contextMap: MDCContextMap = MDC.getCopyOfContextMap()) {
    constructor(vararg pairs: Pair<String, String>): this(
        MDC.getCopyOfContextMap().let {
            if (pairs.isEmpty()) it else (it ?: emptyMap()) + pairs
        }
    )
}
```

**Pros**
* Concise API for extending current MDC `MDCContext("key" to "value")`.

**Cons**
* Consumers might be confused that the existing MDC is extended and not set to the values provided by `pairs`.

## Constructor extension
The goal is to support `vararg pairs: Pair<String, String>` directly in the constructor for convenience,
while maintaining the existing `contextMap` parameter with its default value.

It is required that:
* Both parameters must be part of the constructor
* The default value of `contextMap` must be visible from the function signature (no implicit hiding the constructor invocation)
* The change must be non-breaking and bytecode compatible with the current API (not discussed in this example).
* The following calls are possible
  ```kotlin
  MDCContext() // existing
  MDCContext(mapOf("a" to "1")) // existing
  MDCContext("a" to "1") // new
  MDCContext(mapOf("a" to "1"), "b" to "2") // new
  ```

### Simply extending the constructor with an additional vararg argument
Simply adding another argument does not work:
```kotlin
class MDCContext(contextMap: MDCContextMap = MDC.getCopyOfContextMap(), vararg pairs: Pair<String, String>)

fun desiredConstructorInvocations() {
    MDCContext()
    MDCContext(mapOf("a" to "1"))
    MDCContext("a" to "1") // ❌ error: Argument type mismatch: actual type is 'Pair<String, String>', but 'Map<String, String>?' was expected.
    MDCContext(mapOf("a" to "1"), "b" to "2")
}

fun undesiredConstructorInvocations() {
    MDCContext(contextMap = MDC.getCopyOfContextMap(), "a" to "1") // ⚠️ repeating of default argument
    MDCContext(pairs = arrayOf("a" to "1")) // ⚠️ wrapping varargs in array
}
```

### Secondary constructor with reversed order
```kotlin
class MDCContext(contextMap: MDCContextMap = MDC.getCopyOfContextMap(), vararg pairs: Pair<String, String>) {
    @Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
    constructor(vararg pairs: Pair<String, String>, contextMap: MDCContextMap = MDC.getCopyOfContextMap()): this(contextMap, *pairs)
}

fun desiredConstructorInvocations() {
    MDCContext()
    MDCContext(mapOf("a" to "1"))
    MDCContext("a" to "1")
    MDCContext(mapOf("a" to "1"), "b" to "2")
}

fun ambiguousConstructorInvocations() {
    MDCContext(mapOf("a" to "1"), "a" to "2") // ✅ {a:2} - Primary constructor
    MDCContext("a" to "1", contextMap = mapOf("a" to "2")) // ⚠️ {a:1} - Secondary constructor (awkward argument order)
    MDCContext("a" to "1", mapOf("b" to "2")) // Does not compile, which is good I guess?
}
```
**Pros:**
* Works for all desired constructor invocations.

**Cons:**
* Can be misleading since `contextMap` and `pairs` have a relationship (`pairs` overrides `contextMap`).
  By convention the parameter that is overridden should always come first.
* Black magic?

## Functions
These solutions drop the idea of having the key-value-pairs as part of a constructor completely.

### Extension function `withEntries`
```kotlin
class MDCContext(val contextMap: MDCContextMap = MDC.getCopyOfContextMap())

fun MDCContext.withEntries(vararg pairs: Pair<String, String>): MDCContext = MDCContext(
    this.contextMap.let { if (pairs.isEmpty()) it else (it ?: emptyMap()) + pairs }
)

fun desiredConstructorInvocations() {
    MDCContext()
    MDCContext(mapOf("a" to "1"))
    MDCContext().withEntries("a" to "1")
    MDCContext(mapOf("a" to "1")).withEntries("b" to "2")
}
```
**Pros:**
* No constructor ambiguity or complexity.

**Cons:**
* Slightly more verbose syntax for the caller.
* Creates an intermediate object.

### 4. Factory function `extendCurrent`
We could introduce a factory function on the companion object to handle the vararg case.

```kotlin
class MDCContext(val contextMap: MDCContextMap = MDC.getCopyOfContextMap()) {
    companion object {
        fun extendCurrent(vararg pairs: Pair<String, String>): MDCContext = MDCContext(
            MDC.getCopyOfContextMap().let {
                if (pairs.isEmpty()) it else (it ?: emptyMap()) + pairs
            }
        )
    }
}

fun desiredConstructorInvocations() {
    MDCContext()
    MDCContext(mapOf("a" to "1"))
    MDCContext.extendCurrent("a" to "1")
}
```

**Pros**
* Clear naming reflects behavior

**Cons**
* Slightly more verbose API
* No combination of first and second argument possible

dkhalanskyjb:
I hadn't realized that the default first parameter with `vararg` following it would cause problems, but on second thought, it is actually good that this doesn't work!
```kotlin
MDCContext(mapOf("a" to "b"))
MDCContext(      "a" to "b" )
```
These look like they should be identical, but they aren't: the second one will implicitly read the current MDC context, the first one will replace it instead.

This suggests we have been on the wrong path from the start, and what we need is a separate function, not a constructor. Sorry for not noticing that earlier and sending you on an unproductive path.

Luckily, this frees us from needing to do dark magic. The function-based solutions you present make a lot of sense to me.

> `withEntries`

`plus`? Like for `Map`: https://github.com/JetBrains/kotlin/blob/d75477d5337719cc9096feef6b6f703fa0d0558f/libraries/stdlib/src/kotlin/collections/Maps.kt#L670

> `extendCurrent`

Or just `current`: `MDCContext.current()` looks reasonable if nothing was passed, and `MDCContext.current("a" to "b")` does still convey the idea (if we call the `vararg` something like `overrides`).

Both options are fine by me, though I'm slightly in favor of `plus`, for the following reasons.

By introducing `current()`, we would be creating two ways of obtaining the current `MDCContext`: `MDCContext.current()` and simply `MDCContext()`. The latter one is more succinct, which makes `MDCContext.current()` look unnatural—unless we deprecate `MDCContext()`. We could do that and replace `MDCContext(...)` with `MDCContext.current()` and `MDCContext.fromMap()`.

However, the no-parameter `MDCContext()` call is the most commonly used way to construct an `MDCContext`: https://grep.app/search?f.lang=Kotlin&words=true&q=MDCContext%28 I think it's reasonable for it to have a short form. We haven't received any complaints from users who thought `MDCContext()` would be empty but was instead the current context, so forcing them to spell out `current()` looks like a negative-value change.

schielek:
I'll make it very short: I agree with everything you said.

I updated the PR to introduce a plus operator instead. I chose to go for all the overloads which `Map` has (`Map`, `Pair`, `Sequence<Pair>`, `Iterable<Pair>`).

I am quite happy with the API. It is not as short as originally planned, but a good compromise between preserving existing behavior, being explicit, and being relatively concise.

> Sorry for not noticing that earlier and sending you on an unproductive path.

Don't feel sorry. You did not send me on an unproductive path. As mentioned before, your feedback has been incredibly valuable and you do a phenomenal job at keeping quality standards high for this library.
Don't let your guard down when reviewing the next iteration of this PR ;)

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
```diff
@@ -44,9 +52,10 @@ public typealias MDCContextMap = Map<String, String>?
  * ```
  * launch(MDCContext()) {
  *     MDC.put("key", "value") // This update will be captured
- *     withContext(MDCContext()) {
+ *     withContext(MDCContext() + mapOf("otherKey" to "otherValue")) { // This update will be captured as well
```
dkhalanskyjb:
```suggestion
 *     withContext(MDCContext() + "otherKey" to "otherValue") { // This update will be captured as well
```
Here and elsewhere in the documentation, since there's a more succinct overload accepting a pair, why not use it?

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
```diff
@@ -24,6 +24,14 @@ public typealias MDCContextMap = Map<String, String>?
  * }
  * ```
  *
+ * You can derive a new [MDCContext] with additional entries via the `+` operators.
+ *
+ * ```
+ * launch(MDCContext() + mapOf("key" to "value")) { // Propagate existing MDC with additional key-value pairs
+ *     logger.info { "..." }                        // The MDC context contains the mapping here
```
dkhalanskyjb:
```suggestion
 *     logger.info { "..." } // The MDC context contains the mapping here
```
For consistency with the rest of the documentation.

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
```diff
@@ -106,3 +112,44 @@ public class MDCContext(
...
+public operator fun MDCContext.plus(map: Map<String, String>): MDCContext = MDCContext((contextMap ?: emptyMap()) + map)
+
+/**
+ * Returns a new [MDCContext] with [pair] added to this context.
+ *
+ * [pair] is a key-value entry where `pair.first` is the key and `pair.second` is the value.
+ * If the key `pair.first` is already present in this context, its existing value is replaced
+ * with `pair.second`.
+ */
+public operator fun MDCContext.plus(pair: Pair<String, String>): MDCContext = MDCContext((contextMap ?: emptyMap()) + pair)
```
dkhalanskyjb, line 135: This operator makes perfect sense. I believe we can introduce it as non-`ExperimentalCoroutinesApi` in this form.

The other operators have a questionable aspect to them: they all accept collections that may be empty, and when that happens, they create an `emptyMap()`.

Intuitively, `null` and an empty map should behave equivalently, so usually, this shouldn't matter.
The behavior a user observes *is* different, though: `MDC.clear()` + `MDC.getCopyOfContextMap()` returns `null`, but if you stick a `MDC.setContextMap(emptyMap())` between the two, `getCopyOfContextMap` starts returning `{}`. Does anyone rely on this distinction? I really don't know.

What behavior is correct for `MDC.clear(); val context = MDCContext() + emptyMap()`, then?
- One could say an empty right-hand side should return the left-hand side unchanged—then, the right thing to do is `if (map.isEmpty()) this else MDCContext((contextMap ?: emptyMap()) + map)`. This is simple and intuitive, like in math: `x + 0 = x`.
- Alternatively, one could say that `withContext(MDCContext() + myMap) { getCopyOfContextMap()!! }` should work consistently regardless of whether `myMap` is empty. This corresponds to the current behavior, which makes the behavior more predictable and stable.

Neither of the options looks *incorrect* to me. This calls for `@ExperimentalCoroutinesApi`, so that if we make a wrong choice now, we have an easy way to modify the behavior later. Unless, of course, you can think of a way in which one of the two behaviors is clearly problematic.

integration/kotlinx-coroutines-slf4j/test/MDCContextPlusOperatorTest.kt
```diff
@@ -0,0 +1,152 @@
...
+                { MDCContext(mapOf("key" to "value")) + arrayOf("key" to "new", "b" to "2") },
+                mapOf("key" to "new", "b" to "2")
+            ),
+            arrayOf(
+                "MDCContext() + empty Array",
+                { MDCContext() + emptyArray<Pair<String, String>>() },
+                emptyMap<String, String>()
+            ),
+        )
+    }
```
dkhalanskyjb, line 121: This test suite is stylistically inconsistent with how we structure tests. Normally, this would be a test per test case, like
```kotlin
// MDCContext(existing) + Array<Pair> overrides shared key
@Test
fun testPlusArrayOverriding() {
    assertEquals(
        mapOf("key" to "new", "b" to "2"),
        (MDCContext(mapOf("key" to "value")) + arrayOf("key" to "new", "b" to "2")).contextMap
    )
}
```
It's also perfectly fine to group batches together into a single test, like
```kotlin
@Test
fun testPlusArray() {
    // four `assertEquals` calls
}
```
The primary benefit of this free-form structure when compared to the parametric approach is the extra flexibility: it's trivial to add new test cases down the line, for example, those of `MDCContext()` is called with no parameter in a non-empty context (which does sound like an important case, actually).

schielek: Done

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
schielek: Since `contextMap` is of type `Map`, it felt most natural for me to use the `Map` `+` `Map` variant. But I guess `Map` `+` `Pair` would also do. We need to include parentheses though:
```kotlin
withContext(MDCContext() + mapOf("otherKey" to "otherValue")) {}
// vs.
withContext(MDCContext() + ("otherKey" to "otherValue")) {}
```
I still prefer the `Map` variant, but I don't have a strong opinion on this matter. I'll let you take the final call.

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
schielek: ### Experimental annotation
It's probably best to introduce all variants with `@ExperimentalCoroutinesApi`. This leaves us the option to roll back everything in case somebody comes around with an even better solution.

### null issue
Both arguments can be made here. Tough call. Life would be easier if either `getCopyOfContextMap()` would not return `null` or Kotlin would have non-empty collections.
Looking at the implementation of `Map.plus(Map)` from the kotlin stdlib, a new map is created every time. This means `map + emptyMap() !== map`. This operation is not a no-op.
```kotlin
public operator fun <K, V> Map<out K, V>.plus(pairs: Sequence<Pair<K, V>>): Map<K, V> =
    LinkedHashMap(this).apply { putAll(pairs) }.optimizeReadOnlyMap()
```
Interestingly there is an empty check on the left hand operator:
```kotlin
public operator fun <K, V> Map<out K, V>.plus(pair: Pair<K, V>): Map<K, V> =
    if (this.isEmpty()) mapOf(pair) else LinkedHashMap(this).apply { put(pair.first, pair.second) }

public operator fun <K, V> Map<out K, V>.plus(pairs: Iterable<Pair<K, V>>): Map<K, V> =
    if (this.isEmpty()) pairs.toMap() else LinkedHashMap(this).apply { putAll(pairs) }
```
This makes `withContext(MDCContext() + myMap) { getCopyOfContextMap()!! }` consistent with the kotlin stdlib.

schielek: > What behavior is correct for MDC.clear(); val context = MDCContext() + emptyMap(), then?

It starts with the problem that the result of `MDC.clear()` is undefined. According to the SLF4J docs it "clears" the context. SLF4J's `BasicMDCAdapter` implementation will result in an empty map:
```java
// BasicMDCAdapter
public void clear() {
    Map<String, String> map = inheritableThreadLocalMap.get();
    if (map != null) {
        map.clear();
        inheritableThreadLocalMap.remove();
    }
}
```

But the logback implementation will set the contextMap to `null`:
```java
// LogbackMDCAdapter
@Override
public void clear() {
    readWriteThreadLocalMap.set(null);
    nullifyReadOnlyThreadLocalMap();
}
```

log4j clears the MDC:
```java
// Log4jMDCAdapter
@Override
public void clear() {
    ThreadContext.clearMap();
}
```

For all intents and purposes it seems like SLF4J doesn't make a difference between `null` and an empty map.

schielek: > like in math: `x + 0 = x`.

`x + 1 - 1 = x` should then also apply.

```kotlin
// using logback
MDC.getCopyOfContextMap() // null
MDC.put("key", "value")
MDC.remove("key")
MDC.getCopyOfContextMap() // {}
```

dkhalanskyjb: > It's probably best to introduce all variants with `@ExperimentalCoroutinesApi`

I'm fairly confident in the `+ x to y` overload, but sure, there's no hurry to stabilize it.

> Looking at the implementation of `Map.plus(Map)` from the kotlin stdlib, a new map is created every time.

Yes, and that behavior is beneficial there: https://pl.kotl.in/G1vdIM8Vk No matter what you add, you get a fresh map, and that can be useful. Since the map itself is not exposed from SLF4J, only its copy can be obtained, this isn't applicable to our scenario.

Additionally, note that even with this fresh allocation, `map + mapOf() == map` still works.
In `MDCContext`, if we always allocate a map, `(mdcContext + mapOf()).contextMap == mdcContext.contextMap` won't always hold.

> `x + 1 - 1 = x` should then also apply.

The problem is that there's no `minus` for `Map` that behaves like it does in math, so the analogy doesn't hold. `map.remove("key")` leaves the map unchanged if the map doesn't include the key, which would imply `x - 1 = x`, which isn't right.

schielek:
I have made two changes:
1. Handle empty inputs as no-op
2. Release plus pair variant as stable API

After thinking it through, I come to the conclusion that there is no unexpected outcome for users when treating empty input as no-op. Let's go down that path.

Next steps:
Do you think we should go for additional performance improvements? There is room for improvement in duplicate checks and constructing more instances of `Map` than needed. Do you favor a performant implementation over a simple and readable implementation?
If we go for performance, is it important that `MDCContext::contextMap` is immutable (do not allow cast to `MutableMap`)?

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
```diff
@@ -135,26 +138,47 @@ public operator fun MDCContext.plus(pair: Pair<String, String>): MDCContext = MD
...
  * in this context.
  */
 @ExperimentalCoroutinesApi
-public operator fun MDCContext.plus(pairs: Sequence<Pair<String, String>>): MDCContext = MDCContext((contextMap ?: emptyMap()) + pairs)
+public operator fun MDCContext.plus(pairs: Sequence<Pair<String, String>>): MDCContext = this + pairs.iterator()
 
 /**
  * Returns a new [MDCContext] with all [pairs] from the array added to this context.
  *
+ * If [pairs] is empty, this context is returned unchanged.
```
dkhalanskyjb, line 166: Let's either 1) not document this for now and leave it unspecified, or 2) document this and specify that this behavior may be changed in the future before the API is stabilized.

Since we're unsure about this behavior, we shouldn't advertise it and have people rely on it.

dkhalanskyjb (approved):
Which performance improvements do you have in mind? I don't notice any low-hanging optimization opportunities. Note: it isn't obvious, but `emptyMap()` doesn't allocate anything: https://github.com/JetBrains/kotlin/blob/6e17269d0513eabef2ae7b215c21c32b3d715826/libraries/stdlib/src/kotlin/collections/Maps.kt#L40

In any case, I consider optimizing experimental APIs beyond the obvious a wasted effort: if we mark something as experimental, it means we expect to need to change its behavior, and changing a simple implementation is much less error-prone.

From my point of view, this PR is ready to be merged and is a job well done. Would you like to add some final touches, or is it good to go?

dkhalanskyjb:
Just noticed that the ABI dump wasn't updated to include the new definitions. Please make sure `./gradlew clean build` passes.

schielek:
> Which performance improvements do you have in mind? I don't notice any low-hanging optimization opportunities. 

I think you could improve `MDCContext((contextMap ?: emptyMap()) + pairs)`. When `contextMap` is `null` the expression evaluates to `MDCContext(emptyMap() + pairs))` which could be improved to just `MDCContext(pairs)`.

schielek:
ABI is updated and documentation imporved.

integration/kotlinx-coroutines-slf4j/src/MDCContext.kt
schielek: done

dkhalanskyjb:
This is solid work! Thank you for seeing this through and navigating the design concerns.

PR #4625, opened 2026-02-23
[K/JS] Add exportability for Kotlin/JS with conversion to `AsyncItera…

JSMonk:
…tor`

With the following commit the `Flow` interface could be consumed on the JavaScript / TypeScript side as follows:

```typescript
for await (const value of FLOW_PRODUCER) { ... }
```

As well as created from from either `AsyncIterable`, `AsyncIterator` or an async generator. With an async iterator the API looks like the following:
```typescript
import { Flow } from "coroutines"

Flow.fromAsyncGenerator(async function* () {
  yield await fetch(...)
  yield "Hello, World"
});
```

^KT-80733 Fixed


JSMonk:
@dkhalanskyjb.

**Important note**: it could work only starting 2.3.20 (since `@JsSymbol` is introduced since this version).
So it could be merged only after 2.3.20

dkhalanskyjb (changes requested):
We cannot accept this PR as is. We already have several integrations with other asynchronous systems (for example, everything in the `reactive/` directory), and this PR is inconsistent with that approach.
* Most notably, we usually don't have `Flow` implement any interfaces (even if it could have)—instead, we provide converter functions, so that the concerns are clearly separated.
* `from` functions are questionable: only `from(JsAsyncIterable)` is in line with the cold `Flow` semantics. In any scenario, it's more consistent for `from` to be a extension function on `JsAsyncIterable` instead.
* Also, none of the new code is tested.

If you would like to work on getting this PR to be merged, please take a look at the existing reactive integrations for inspiration and try mimicking their API. Alternatively, please file an issue instead, and we'll look into implementing the integration ourselves, provided there's a demand for it.

JSMonk:
@dkhalanskyjb the functions are not designed to be used from Kotlin, but from the TypeScript side (as a part of the @JsExport functionality):
- For the first item, there is no other way to do so (if we didn't have any Flow to be passable as `AsyncIterable` inside TypeScript code, however, since the structural behavior of the TypeScript type system, I can do it just by adding the Symbol but escaping the `AsyncIterable` inside the super-type list)
- I agree that the part of implementation is not cold, but I disagree with the `extension` function part. Extension functions are designed to be easily used in Kotlin, but not on the other platforms. Since the factory functions are designed to be used from TypeScript, it makes sense to keep them within the static scope. 
- Sure, I will add them; however, it's still required to migrate at least to 2.3.20 (so the code could at least compile).

JSMonk:
@dkhalanskyjb
- Tests are added
- Super type is dropped
- Generator based solution is re-worked with a cold approach

dkhalanskyjb:
Philosophically, the primary issue with the pull request is that, as proposed, the collection of `Flow` in JS is entirely different from the `collect`ion in Kotlin. `collect` is using the caller's coroutine context, collection of asynchronous iterators isn't. `collect` is push-based, `AsyncIterable` is pull-based.

With this mismatch, it's simply incorrect to have the async iterator API as readily available as `collect`. In fact, the less idiomatic + more brittle approach using `JsAsyncIterable` looks *more* streamlined with the proposed change than `collect`! Even in Kotlin, we can't write `for (element in flow)`.

This means that *some* wrapping ceremony is mandatory here, at least to signify the travel between the two worlds. There is a conceptual chasm between `Flow` and `AsyncIterable`, and it should be reflected in the API. The non-context-preserving option needs to be chosen by the user implicitly, not hidden in some magic syntax.

The most idiomatic way to implement the conversion, one that's consistent with the other reactive integrations, is via extension functions, like
```kotlin
fun <T> Flow<T>.asAsyncIterable(): JsAsyncIterable<T>
fun <T> JsAsyncIterable<T>.asFlow(): Flow<T>
```
We do this for integration with the Java entities as well: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-jdk9/kotlinx.coroutines.jdk9/ (see `asFlow` and `asPublisher`). I don't remember any complaints about these functions being somehow unfit for the purpose. They do the job of converting between various asynchronous systems.

The syntax from the Java side is indeed unpleasant, as you mention: you have to write something like `ReactiveFlowKt.asPublisher(flow)`. The way to get around that is to perform the conversion on the Kotlin side, and then expose a pure Java API to the Java code. Still, when you're in a tight spot and have to call the extension function from Java code, you can do that.

Another way would be to introduce a wrapper class:
```kotlin
public class FlowIterable<out T>(public val flow: Flow<T>): JsAsyncIterable
```
This can be called with the same amount of ceremony from Kotlin and JS. It's not as nice as extension functions, as we typically short-circuit chains like `asFlow().asPublisher()` into no-ops to minimize the mismatch, and you can't do that with class wrappers. A class is also more API-heavy than a simple function.

JSMonk:
> Philosophically, the primary issue with the pull request is that, as proposed, the collection of Flow in JS is entirely different from the collection in Kotlin. collect uses the caller's coroutine context; a collection of asynchronous iterators isn't. collect is push-based; AsyncIterable is pull-based.

Exactly the same situation we already have with the Flow conversion to [AsyncIteratorProtocol](https://developer.apple.com/documentation/Swift/AsyncIteratorProtocol) in SwiftExport, and we already convert them in this way, so to not break consistency across platforms - it doesn't make sense to introduce manual convertors.

The core idea behind our work on interop (both JS and Swift) is to make the use of Kotlin declarations on the target platform idiomatic for that platform (not for Kotlin, in our case, JS).

Unfortunately, there is no primitive for push-based async iteration in JavaScript, so the only way to represent Flow in an idiomatic JS way is [AsyncIterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols).

> Another way would be to introduce a wrapper class:
```kotlin
public class FlowIterable<out T>(public val flow: Flow<T>): JsAsyncIterable
```
> This can be called with the same amount of ceremony from Kotlin and JS. It's not as nice as extension functions, as we typically short-circuit chains like asFlow().asPublisher() into no-ops to minimize the mismatch, and you can't do that with class wrappers. A class is also more API-heavy than a simple function.

It requires the user to explicitly call those wrappers (which can't be done from the common code, since the wrapper class would exist only for JS/WasmJS targets), which means tons of expect/actual declarations in the sake of (to be honest, I haven't gotten the point in the sake of what).

Just one more time to highlight the point: the introduced API (at least right now) is not for use inside of Kotlin (I can make it clear by adding a hidden deprecation), but TypeScript.

P.S. I think there could be a need to pass Flow to an imported JavaScript API that expects AsyncIterator/AsyncIterable, so in this case, we definitely will need to introduce a super type of `AsyncIterable` for Flow, but we can postpone this. At this point, I want to resolve the issue with exporting Flow.

dkhalanskyjb:
> we already have with the Flow conversion to [AsyncIteratorProtocol](https://developer.apple.com/documentation/Swift/AsyncIteratorProtocol) in SwiftExport

Thank you, I hadn't realized this. This is probably a problem as well. Could you provide any links to this functionality (code, documentation, KEEPs)? I couldn't find any mentions of this implicit transformation or how it's performed.

> make the use of Kotlin declarations on the target platform idiomatic for that platform (not for Kotlin, in our case, JS).

> It requires the user to explicitly call those wrappers (which can't be done from the common code, since the wrapper class would exist only for JS/WasmJS targets), which means tons of expect/actual declarations in the sake of (to be honest, I haven't gotten the point in the sake of what).

Sorry, this point is unclear. If the whole goal is to use `Flow` from other languages, why is the common code a concern? Yes, `FlowIterable` (or the equivalent) only makes sense for JS, so what would be the point of a common-code declaration exposing it? The interface exposed to JS can include `FlowIterable` without affecting other targets.

> in this case, we definitely will need to introduce a super type of AsyncIterable for Flow

If the functions intended for `AsyncIterable` are present in `Flow` at all—which is the entire contention here—then sure, `AsyncIterable` should also become the supertype, I can't imagine any problems with that.

> in the sake of (to be honest, I haven't gotten the point in the sake of what).

It's for the sake of not breaking the basic contracts. From https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/ (emphasis mine):

> An asynchronous data stream that **sequentially** emits values and completes normally or with an exception.

> By default, flows are sequential and **all flow operations are executed sequentially in the same coroutine**, with an exception for a few operations specifically designed to introduce concurrency into flow execution such as [buffer](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/buffer.html) and [flatMapMerge](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flat-map-merge.html).

Providing a way to collect a `Flow` that breaks the basis of the conceptual model behind `Flow` with no extra ceremony, as the default, just to save an extension function invocation per `Flow` collection, doesn't look like a good compromise to me.

Of course, we are also the authors of the documentation and could change it, but having to add "(except when it's being collected on JS using the native async iterable syntax)" throughout the text suggests the abstractions are broken and that the proposed semantics isn't a good fit.

With JS, we are at least lucky enough to catch this early, before it reaches the users—I don't know what the situation with Swift is.

JSMonk:
> Thank you, I hadn't realized this. This is probably a problem as well. Could you provide any links to this functionality (code, documentation, KEEPs)? I couldn't find any mentions of this implicit transformation or how it's performed.

The code is placed [here](https://github.com/JetBrains/kotlin/blob/master/native/swift/swift-export-standalone/resources/swift/KotlinCoroutineSupport.kt) (all the .kt, .swift, .h files)

> Sorry, this point is unclear. If the whole goal is to use Flow from other languages, why is the common code a concern? 

Since I want to have a common logic exported to all the platforms. Imagine a project with native UI on platforms (SwiftUI, React, as an example).

I have a logic placed in `commonMain` that is shared across all the platforms. I definitely don't want to duplicate it (its main purpose is KMP in this case), so I just want to export it to the platform and use it there. So this is why the common code is a concern.

> It's for the sake of not breaking the basic contracts. From https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/ (emphasis mine):

Could you please demonstrate how usage of the AsyncIterator from JavaScript breaks the basic contract?

dkhalanskyjb:
> The code is placed [here](https://github.com/JetBrains/kotlin/blob/master/native/swift/swift-export-standalone/resources/swift/KotlinCoroutineSupport.kt) (all the .kt, .swift, .h files)

Thanks, I'll read it some time later.

> I just want to export it to the platform and use it there

And by that, do you mean that you want to take a Kotlin module, export its bindings to another language *without writing any extra glue code in Kotlin*, and use it as is? This aspiration fails to work even for Java, to which Kotlin is the closest. Not just in terms of extension functions, which are used quite often, but also in `value` classes that perform unpredictable name mangling.

> Could you please demonstrate how usage of the AsyncIterator from JavaScript breaks the basic contract?

Yes, of course.

```kotlin
flow {
    println("Emitting 10")
    emit(10)
    println("Done processing 10")
    delay(100)
    println("Emitting 2")
    emit(2)
    println("Done processing 2")
}.collect {
    println("Repeat $it times")
    repeat(it) {
        println("Processing $it...")
        delay(100)
    }
    println("Done!")
}
```

The `Flow` semantics describe the order of this code clearly: first, everything before `emit(10)` runs, then, the `collect` lambda runs to completion, then, `Done processing 10` is printed. Unless asynchronous behavior is intentionally introduced by using a hot flow or a concurrency operator, this property is preserved, and it's quite important for flows.

The async iterator in a `for` loop does not and can not work like this: its body will be interleaved with the `flow` processing. 

This got me thinking: it looks at a glance like `AsyncIterable` is similar to the `Channel` in `kotlinx.coroutines`—`ChannelIterator` is an example of iteration over a channel, similar to `AsyncIterable`, and `ReceiveChannel.cancel` can probably be used to emulate `return`/`throw` on the `AsyncIterable`. Isn't `Channel` a better candidate for the `AsyncIterable` interface? We already have `ReceiveChannel.receiveAsFlow` and `Flow.produceIn`—those could provide access to `Flow`.

JSMonk:
> ```kotlin
> flow {
>     println("Emitting 10")
>     emit(10)
>     println("Done processing 10")
>     delay(100)
>     println("Emitting 2")
>     emit(2)
>     println("Done processing 2")
> }.collect {
>     println("Repeat $it times")
>     repeat(it) {
>         println("Processing $it...")
>         delay(100)
>     }
>     println("Done!")
> }
> ```
> 
> The `Flow` semantics describe the order of this code clearly: first, everything before `emit(10)` runs, then, the `collect` lambda runs to completion, then, `Done processing 10` is printed. Unless asynchronous behavior is intentionally introduced by using a hot flow or a concurrency operator, this property is preserved, and it's quite important for flows.
> 
> The async iterator in a `for` loop does not and can not work like this: its body will be interleaved with the `flow` processing.
> 

Thank you so much; With this example, is there any possibility to repeat the same semantic by using Flow on JavaScript (without `AsyncIterator`)?

> This got me thinking: it looks at a glance like `AsyncIterable` is similar to the `Channel` in `kotlinx.coroutines`—`ChannelIterator` is an example of iteration over a channel, similar to `AsyncIterable`, and `ReceiveChannel.cancel` can probably be used to emulate `return`/`throw` on the `AsyncIterable`. Isn't `Channel` a better candidate for the `AsyncIterable` interface? We already have `ReceiveChannel.receiveAsFlow` and `Flow.produceIn`—those could provide access to `Flow`.

Unfortunately, no. This is not my idea; it's just what people ask us. I would be happy to collaborate to make it happen (to make Flow exportable to JavaScript in terms of JavaScript API)

> And by that, do you mean that you want to take a Kotlin module, export its bindings to another language _without writing any extra glue code in Kotlin_, and use it as is? This aspiration fails even for Java, to which Kotlin is closest. Not just in terms of extension functions, which are used quite often, but also in `value` classes that perform unpredictable name mangling.

Let's move forward without spending too much time on this. The SwiftExport team and mine have been working on it for a few years, and we've already gotten good results (at least from user feedback).

dkhalanskyjb:
Hi! The `develop` branch, which is where we're doing development (instead of `master`), now uses Kotlin 2.3.21.

JSMonk:
As discussed with @dkhalanskyjb, the MR was re-worked based on the channel exportability and explicit method converting `Flow` into `AsyncIterable`.

Firstly, the #4632 should be merged, since the `Flow` implementation of `asAsyncIterable` is based on the `Channel` implementation of `Symbol.asyncIterator`.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,266 @@
...
+     * Instead, [AbstractFlow] can be used as the base type to properly ensure flow's properties.
+     *
+     * All default flow implementations ensure context preservation and exception transparency properties on a best-effort basis
+     * and throw [IllegalStateException] if a violation was detected.
+     */
+    @JsExport.Ignore
+    public actual suspend fun collect(collector: FlowCollector<T>)
+
+
+    public fun asAsyncIterable(): JsAsyncIterable<T> =
```
dkhalanskyjb, line 48: This public function is missing a KDoc. In my opinion, in terms of describing the behaviors, it should be sufficient to specify that it's a shorthand for `buffer(0).produceIn(GlobalScope)` (that is, its implementation). A usage example and an explanation of when one needs to use this would be appropriate, too.

Also, this function should be either marked as `@ExperimentalCoroutinesApi` or hidden from Kotlin code. There are too many unclear aspects to it, and there is a potential for its behavior to change, so we can't call this a stable API.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,266 @@
...
+        buffer(0).produceIn(GlobalScope)
+
+    @JsExport.Ignore
+    // Important note: it would be much nicer to place those factory functions outside of Flow
+    // so from both Kotlin and TypeScript side it could be used without importing Flow (like in `flowOf` or `flow`)
+    // However, the described way of exporting factory functions forces the functions always to be exported (even if people don't use them and don't export Flow),
+    // and that may cause bundle size problems (at least right now).
+    // So, until the bundle size problem is solved, we keep those factory functions inside Flow, with possibility to move them outside later.
```
dkhalanskyjb, line 56: Likewise, since we know in advance those functions may need to get moved somewhere else, they need to be marked as experimental.

kotlinx-coroutines-core/wasmWasi/src/flow/Flow.wasm.kt
```diff
@@ -0,0 +1,195 @@
+package kotlinx.coroutines.flow
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.flow.internal.*
+import kotlin.coroutines.*
+
+/**
```
dkhalanskyjb: `actual` definitions of `Flow` shouldn't include copies of the KDocs already present in `common`.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,266 @@
...
+            try {
+                while (true) {
+                    val result = iterator.next().await()
+                    if (result.done) {
+                        completed = true
+                        break
+                    }
+                    emit(result.value.unsafeCast<T>())
+                }
+            } finally {
```
dkhalanskyjb, line 92: Why not cancel the async iterator if `emit` threw an exception?

Note: in this scenario, it *is* fine to treat `CancellationException`s as a normal `return`, because that's consistent with the enormous design problem affecting the entirety of `kotlinx.coroutines`: https://github.com/Kotlin/kotlinx.coroutines/issues/3658 We will need to fix the issue in the entire library later.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,312 @@
...
+        val iterator: JsAsyncIterator<Int> = flow.asAsyncIterable().asDynamic()[js("Symbol.asyncIterator")]()
+        assertNextStepToBe(iterator, 1, done = false)
+        // Call return() to stop iteration early
+        val returnResult = iterator.asDynamic().`return`().unsafeCast<Promise<JsIteratorResult<Int>>>().await()
+        assertEquals(true, returnResult.done)
+        assertNextStepToBe(iterator, done = true)
+    }
+
+    @Test
+    fun testFlowToAsyncIteratorThrow() = runTest {
```
dkhalanskyjb, line 59: Please check that the `flow` receives the exception. For example (note: I haven't tried running this, mistakes are likely):
```kotlin
val deferred = CompletableDeferred<Throwable>()
val flow = flow {
    emit(1)
    try {
        emit(2)
    } catch (e: Throwable) {
        deferred.complete(e)
    }
}
// ...
// after failing the flow
assertSame(error, deferred.await())
```

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,312 @@
...
+            emit(2)
+            emit(3)
+            emit(4)
+            emit(5)
+        }
+        val iterator: JsAsyncIterator<Int> = flow.asAsyncIterable().asDynamic()[js("Symbol.asyncIterator")]()
+        assertNextStepToBe(iterator, 1, done = false)
+        assertNextStepToBe(iterator, 2, done = false)
+        val returnResult = iterator.`return`(42).await()
+        assertEquals(true, returnResult.done)
```
dkhalanskyjb, line 113: Shouldn't `returnResult.value` be 42? Why not check that?

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,312 @@
...
+        val flow = flow {
+            emit(1)
+            emit(2)
+            emit(3)
+        }
+        val iterator: JsAsyncIterator<Int> = flow.asAsyncIterable().asDynamic()[js("Symbol.asyncIterator")]()
+        assertNextStepToBe(iterator, 1, done = false)
+        // Call throw() to cancel the iterator
+        val error = js("new Error('test error')")
+        assertFailsWith<Throwable> { iterator.`throw`(error).await() }
```
dkhalanskyjb, line 69: Please also test the no-argument `throw`.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,312 @@
...
+        assertEquals(true, returnResult.done)
+        assertNextStepToBe(iterator, done = true)
+    }
+
+    @Test
+    fun testFlowToAsyncIteratorThrow() = runTest {
+        val flow = flow {
+            emit(1)
+            emit(2)
+            emit(3)
```
dkhalanskyjb, line 63: We need to verify that `finally` cleanups at the point of cancellation run (and in fact, that cancellation *does* happen).

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,312 @@
...
+        val asyncIterator = createAsyncIteratorWithCleanup(
+            listOf(1, 2, 3, 4, 5),
+            onReturn = { returnCalled = true }
+        )
+        val flow = Flow.from(asyncIterator)
+        val results = mutableListOf<Int>()
+        flow.take(2).collect { results.add(it) }
+        assertEquals(listOf(1, 2), results)
+        yield() // Allow cleanup to happen
+        assertTrue(returnCalled, "return() should be called on cancellation")
```
dkhalanskyjb, line 164: We also need to check that 3, 4, and 5 weren't requested from `asyncIterator` (as opposed to just getting dropped somewhere on the way). To that end, `onReturn` could have the type `(Int) -> Unit`, where `Int` is the index of the last requested element.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,312 @@
...
+            TestException("iterator error")
+        )
+        val flow = Flow.from(asyncIterator)
+        val results = mutableListOf<Int>()
+        try {
+            flow.collect { results.add(it) }
+            fail("Should have thrown TestException")
+        } catch (e: TestException) {
+            assertEquals("iterator error", e.message)
+        }
```
dkhalanskyjb:
```suggestion
        assertFailsWith<TestException> {
            flow.collect { results.add(it) }
        }.apply {
            assertEquals("iterator error", message)
        }
```
?

dkhalanskyjb (changes requested):
The build doesn't pass. Updating the ABI dump should fix that.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: But it cancels the iterator, isn't it?

```kotlin
if (!completed) {
  iterator.asDynamic().`return`().unsafeCast<Promise<*>>().await()                
}
```

JSMonk: Or do you mean to provide the exception thrown by `emit` to the iterator via the `throw` method?

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
JSMonk: Should it happen? As far as I remember, we've discussed that throwing an error inside of `for await` should have no impact on either the Channel state or the Flow state.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
JSMonk: The same question is here. We've discussed that throwing an error inside of the `for await` shouldn't change the state of `Flow`/`Channel`. And the suggestion is actually to check that it's changed. I feel that we should test the opposite, shouldn't we?

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
dkhalanskyjb: Yep, exactly.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
dkhalanskyjb: > We've discussed that throwing an error inside of the for await shouldn't change the state of `Flow`/`Channel`.

Yes, and my position is still the same: the state of `Flow`/`Channel` should change: https://github.com/Kotlin/kotlinx.coroutines/pull/4632#discussion_r3689529820 My point is that such behavior aligns with the intuition and expectations of JS programmers, and we have no reason to diverge from that.

I see now that during the review, I missed the commit that made `return` and `throw` only affect the channel, and because of that, I merged the version that doesn't change the state of `Channel`. It's my mistake, I'm sorry about that.

When it comes to `Flow`, there's no space for debate, the flow has to observe the exception. The reasoning here goes beyond the JS programmer expectations and concerns the core of what a `Flow` is. There are two types of `Flow`:
1) Cold flows, and
2) Hot flows.

Cold flow collection is semantically a self-contained computation, launched for each collector independently and cleaned up when a collector is no longer interested in receiving values.
```kotlin
flow {
  try {
    emit(1)
  } finally {
    delay(100.milliseconds)
    println("Done")
  }
}.first()
// will print `Done` and only return after that
```
What you're proposing is to drop the computation once the result is no longer needed—which leaks closeable resources and leaves the cleanup computations in `finally` block unrun. This isn't aligned with `Flow` semantics and needs to be fixed.

Hot flow collection is shared among collectors—but collectors fail independently. `SharedFlow` doesn't notice any failures in collectors, so it's perfectly safe to cancel collectors indiscriminately. Only the coroutine scope in which the collection is launched can react to failures.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
dkhalanskyjb: Replied in the comment thread above. In short: yes, it's mandatory for a correct implementation of `Flow` collection.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
dkhalanskyjb: Fixed the channel behavior in a separate PR: https://github.com/Kotlin/kotlinx.coroutines/pull/4718 Again, my apologies for merging the previous version without noticing that the channel was no longer getting cancelled.

With that change, the proper cleanup for cold flows in this PR should happen automatically, so we'll be able to keep the straightforward `buffer(0).produceIn` implementation instead of conjuring up something completely separate.

JSMonk: Sorry, but I completely disagree with the behavior and statements. 

1. It's un-debaggable (aka painful) default behavior. It will take hours for Kotlin users to find the `break` place in JavaScript where their `Flow`/`Channel` was canceled.

2. It's not "the intuition and expectations of JS programmers". There is no contract on the `AsyncIterable` level that it should be "closed" after its `AsyncIterator` break/return/throwing (only the iterator itself). Different AsyncIterables behave differently. [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) is closing itself on break (but to protect users, it also guarantees privileged access to itself, so only one listener at a time can be), while [EventEmmitter](https://nodejs.org/docs/latest/api/events.html#class-eventemitter) doesn't close itself, and all the other listeners continue getting new events.

3. We're on the same ground that the AsyncIterable should preserve the `Flow` contracts. And I definitely want to provide users with the control over such a behavior (for both `Flow` and `Channel`). My proposal is to have different implementations depending on the `Flow`, so that if it's the cold one, to cancel it on `break`/`return`/`throw` (shall we also provide the same guarantees for the privileged access as ReadableStream? I feel we could, but not sure to do it in this iteration), for the "hot" flow, to keep the original flow as is without cancelling it. 

4. Regarding `Channel`, it's a question, because I don't know the contract (at least the only thing I see that early break on Kotlin side iterations doesn't cancel the `ReceiveChannel`, it works only in the opposite way: if the channel is closed, the `next` method will throw the `ClosedReceiveChannelException`). However, I still believe that the least painful behavior for the `Channel` is the contract, that `ReceiveChannel` can be canceled only by the `cancel` method (which, btw, is also available from the TypeScript) and not implicitly by the early `return`/`break`/`throw`. 
Also, there is another argument for keeping the `Channel` behavior as it is now: with the `cancel` method available and the current behavior on early return, users have more control and can choose whether to use `cancel` or not (`ReceiveChannel`).
 Changing the behavior will remove this control (so users will not be able to stop listening to the channel without canceling it).

5. From my side, I feel that the discussion is already too long in time, and we definitely will not cover all the user expectations and cases with the discussion, so let we annotate all of the new methods with the experimental annotations (so opt-in is required, keeping the room for changing the behavior in the future after user feedback) and release at least some version to receive actual user feedback.

dkhalanskyjb: 1. Finding a `break` isn't any harder than the place where a concurrent computation putting elements into a channel failed to get cancelled. Generally, thrown exceptions are much easier to debug than resource leaks are. If you believe debugging a thrown exception will take hours for someone, then it will take them days to find out they were supposed to call `cancel()`.

2. The expectation I'm talking about isn't "`for await` renders the data structure unusable" but "I don't have to do anything special to clean up the asynchronous process after exiting from `for await`". `EventEmitter` doesn't require any cleanup procedures and doesn't even support closing it, so it's not relevant here.

3. There is no way to distinguish between cold and hot flows.
Even if it were possible, I'm not a fan of this proposal. Having a simple async iterable is nice for the majority of cases, yet we agreed that configurability could also be useful—and a simple, straightforward `buffer(0).produceIn` as the implementation would play into that, because it would be easy to copy and tweak it to other scenarios. Complex custom implementations for cold flows would go against this.

4. The Kotlin channel contract for iteration is indeed to keep the channel open. The equivalent to the JS `AsyncIterable` semantics is provided by the `consumeEach` function.
Regarding control: users are able to stop listening to the channel without cancelling it by avoiding the `for await` syntax and instead manually calling `receive()`/`receiveCatching()`.

5. I agree that this discussion has taken too much time already, yes. I disagree that we should simply push out something or that user feedback will solve this. We are discussing a corner case in concurrent behavior of a niche API, so most likely, we won't get any feedback on it for a year or so, and what we will get will be inconclusive. This means we have to thoroughly design this behavior upfront.
If you're not interested in designing this API, that's fine. Just let us know, and we'll finish the job internally in the Kotlin Libraries team. We'll implement the rest the PR based on that, though the credit will still go to you.

JSMonk: > Finding a break isn't any harder than the place where a concurrent computation putting elements into a channel failed to get canceled.
1. I would say it's a controversial topic.

> The expectation I'm talking about isn't "for await renders the data structure unusable" but "I don't have to do anything special to clean up the asynchronous process after exiting from for await". EventEmitter doesn't require any cleanup procedures and doesn't even support closing it, so it's not relevant here.
2. It's true. However, there is no distinction at the language/stdlib level in JS between "closable" AsyncIterable and "not closable" AsyncIterable. And the behavior is an implementation detail of a specific type. I'm wondering: if we do a kind of cancellation on the reader side (I'm talking about Kotlin), and the `Flow` implementor is not cancellable (just like the EventEmitter in the example), what would happen to the `Flow`? For example, `StateFlow`. Will it continue to emit values even if one of the "subscribers" sends a cancellation exception in the `collect`? Will it do the same if we throw the error in the `for await ... of`? 
The main thing I'm trying to achieve (and I am concerned about with the "cancel-by-default" behavior) is the following: Imagine I have a Compose ViewModel with a `StateFlow`, and some React hook that listens to changes of the Flow, like this:

```ts
function <T> useStateFlow(stateFlow: StateFlow<T>, maxUpdates: number = Infinity) {
  const [value, updateValue] = useState(stateFlow.value);
  useEffect(() => {
    let remaining = maxUpdates;
    const updates = stateFlow.asAsyncIterable();

    (async () => {
        for await (const newValue of updates) {
          if (remaining-- <= 0) break; // Don't want to break other listeners to the stateFlow (especially inside Kotlin)
          updateValue(newValue);
        }
    })();

    return () => updates.return();
  }, [stateFlow, maxUpdates]);

  return value;
}
```

3. Could we consider having two different implementations inside the `Flow` and the `SharedFlow`? Anyway, an alternative solution is to have an option (like the Node.js streams: https://nodejs.org/api/stream.html#readableiteratoroptions or the mentioned browser ReadableStream: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) to turn the canceling on break/return/throw off (I personally like the solution less, since I feel that should or shouldn't the resource been closed should be defined on the type level rather than on usage side, however, it's at least something). 

4.  So, to do something usable, the user should know about the Flow/Channel specific, which we're avoiding during designing interop. It's just another point to have some controlled behavior on this side.

5. I’d like us to converge on a shippable decision soon rather than continue extended discussions, though I agree we seem to be on the right track. That said, even if my team delegates implementation to yours, final decisions on JS interop remain with us, since we own that layer and are accountable for production outcomes. So, regardless of who implements it, the design choice is ultimately our responsibility.

JSMonk: At this point, I feel that the ideal solution (at least from my side) is the following: the body of `for await`  repeat behavior of the `collect` behavior, so that throwing an error acts the same in `collect` and `for await` within one Flow type

dkhalanskyjb: 1. How is it controversial? An exception obviously tells you 1) which channel got closed, 2) where is at least one of the consumers. A leak is something you first notice when your system starts failing with anomalous memory/file descriptor/network consumption, at which point, the entire program is suspect.
2. As I mentioned above, hot flow collectors fail independently. In other words, `SharedFlow` keeps emitting values to collectors even if one collector gets cancelled or fails, so your example will always work, no matter how the `return`/`throw` methods are going to behave.
3. No, `Flow` and `SharedFlow` can't be meaningfully distinguished. For example, `sharedFlow.map { it * it }` is not of the `SharedFlow` type, even though it behaves like one in every aspect.
4. Clearly, we can't cover both use cases with a single non-parameterized function: either the single-consumer case will leak computations by default, or the multi-consumer case will require manually iterating over `Channel`. If it turns out that exposing a multi-consumer `Channel` across the Kotlin and JS API boundaries is in demand, we can always add a parameter to `asAsyncIterable`, like it's done for `ReadableStream`.
5. Let's try sticking to a collaborative approach and converging on a solution that everyone agrees with. If that fails, sure, let's figure out who should get the last say. From where I'm standing, maintaining `kotlinx-coroutines-core/js` has historically been the library team's responsibility, so we are not going to accept contributions that we disagree with.

JSMonk: I've tried recently to play around cancellation of the channel and cancellation with error if an error happen and found a few things to discuss:

- With the behavior on channels the following `break`:

```typescript
for await (const data of produceChannel()) {
    if (counter-- <= 0) break
    console.log(`Received data: ${data}`)
}
```

producing a `CancellationException` which means that people should wrap it in try catch, which is from my side a big concern and I would like to avoid such API for the end user.

- Since we each time create a new channel for a flow with the `asAsyncIterable`, on the the flow if we call `asAsyncIterable` it will start producing values from the beggining once again, is it an expected behavior we want to provide? Example:
```kotlin
fun produceFlow(): Flow<String> = flow {
    for (i in 1..10) {
        emit("$i")
        delay(1000)
    }
}
```

```typescript
import { produceFlow } from "./example.mjs"

let counter = 5
const flow = produceFlow()

for await (const data of flow.asAsyncIterable()) { // Iterate from the start till 5
   if (counter-- <= 0) break
    console.log(`Received data: ${data}`)
}

for await (const data of flow.asAsyncIterable()) { // Iterate again from the start till 10
    console.log(`Received data twice: ${data}`)
}

console.log("Done")
``` 

With the readable stream there will be an error:
```typescript
let i = 0;

const stream = new ReadableStream<string>({
    start(controller) {
        const id = setInterval(() => {
            i++;
            controller.enqueue(`chunk-${i}`);
            if (i >= 5) {
                clearInterval(id);
                controller.close();
            }
        }, 200);
    },
    cancel(reason) {
        console.log("Stream canceled:", reason);
    },
});

let counter = 2;

// First consumption (break early)
for await (const chunk of stream) {
    console.log("First pass:", chunk);
    if (counter-- <= 0) break;
}

// Second consumption on the SAME stream
for await (const chunk of stream) { // A lot of errors here that we're trying to close a closed stream
    console.log("Second pass:", chunk);
}

console.log("Done");

```

dkhalanskyjb: > With the behavior on channels the following `break`:

Could you please share the complete example? If `produceChannel()` creates a new channel each time, no `CancellationException` should occur here, as the only created iterator stops touching the channel once a `break` happens.

`CancellationException` can only be thrown if the same `Channel` is being iterated on twice. An exception in that case notifies the programmer that they need to restructure their program in a way that the `Channel` isn't used after it's been cancelled. It should almost never be necessary to write `try`/`catch` here.

> it will start producing values from the beggining once again, is it an expected behavior we want to provide?

Yes. Please see https://kotlinlang.org/docs/coroutines-flow.html#collect-a-cold-flow . A cold flow encodes a "computation template" that runs for every collector independently. If we consider every `for await` to be a separate collector, then every one of them should receive all values from the beginning.

JSMonk: * Ah, got it - this behaves as expected with the particular implementation. The `CancellationException` is raised only when entering the second `for await`.
* For `Flow`, it also looks good from my side that this aligns with `collect` semantics.
* I’m still concerned about the default channel closing on `break` without providing privileged access to it.  
 Also, I realized `asAsyncIterable` feels unnecessary for `Flow`, since it creates a new `Channel` per call and offers no meaningful configuration surface there. For `Channel`, however, such configuration *does* make sense, but currently there’s no dedicated API to control that behavior explicitly.
* For usability, I’d like to ask again that we consider:
    * making `Flow` directly `AsyncIterable` by default (no wrapper call when there is nothing to configure, e.g. no scope/cancellation options), and
    * adding an explicit conversion/configuration function for `Channel`, where cancellation behavior is actually configurable (similar to `ReadableStream` behavior options).

JSMonk: So, my proposal here is the following:
- In Channel, add a method similar to `values` in ReadableStream with the option to not cancel the channel (and delegate creating of the iterator to the method with the cancellation option as `true`)
- Inside Flow iterator mimics the API of Channel: default iterator which delegates to the default `Channel` iterator, and `values` that delegate to the channel values.

I'm proposing this specific API to follow the same idea as the:
- [ReadableStream in browser](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration)
- [Stream API in Node.js](https://nodejs.org/api/stream.html#readableiteratoroptions)
- [ReadableStream in Deno](https://docs.deno.com/api/web/streams/#ReadableStream)
- [Stream API from Bun](https://bun.com/docs/runtime/streams)

So the API and control over the semantics would be familiar to the JS devs

dkhalanskyjb: In general—deal, let's do that. Let's figure out how to deal with the remaining edge cases.

## Flow

The reason we decided to go for an explicit function for `Flow` was not exactly that we wanted to parameterize it, but rather, that there's a conceptual mismatch between normal `Flow` collection and the equivalent to `buffer(0).produceIn(GlobalScope)`.
1. Flow executes synchronously, and all cleanup is done by the time `collect` returns. Example: https://pl.kotl.in/y8pdr-kXf
2. Flow executes in the caller's coroutine scope, not in `GlobalScope`.

2) can be worked around by configuring the desired context directly in `.flowOn`, but 1) is unavoidable with this simple implementation.

### Counter-proposal

We could write a custom `AsyncIterable` implementation for `Flow` that deals with the asynchronous cleanup, without delegating to `Channel`. Since `return` and `break` return promises, we can wait until `collect` finishes before resolving them with either the result or an error. This deals with the major remaining mismatch and makes `Flow` collection from JS seamless.

## Channel

Having a separate configurable method sounds okay. I'm not sure about its exact API, though.
* If we're going to expose it to Kotlin, it should look Kotlin-esque, which means something like `fun ReceiveChannel<T>.asAsyncIterable(cancelOnEarlyExit: Boolean): AsyncIterable<T>` (probably with some default value for `cancelOnEarlyExit`).
* However, the `ReadableStream` API is nothing like that, at all. Its `values` accepts an object with a single `preventCancel` field. Maybe that's familiar to JS users (and should be implemented for them as the target audience), but `class ChannelAsyncIteratorOptions(val cancelOnEarlyExit: Boolean)` is extremely non-conventional.

That is to say, either we hide these conversions from Kotlin (which could be okay), or we make two APIs, or the JS developers will have to deal with the Kotlin-esque API.

JSMonk: I've got it for the `Channel` and will prototype soon.
Regarding `Flow`. Would we want to have the version without cleaning (the same as the one in the `Channel` with `preventCancel` parameter)? If so, does it make sense just to use `buffer(0).produceIn(GlobalScope).` for the method?

dkhalanskyjb: No, there's no reason not to await a `Flow` cleanup. The cleanup will run anyway, we're only controlling whether it can run concurrently with the code that comes after `for await`.

JSMonk: Okay, so I re-implemented the `asAsyncIterable` with all the notes and edge cases we've discussed, and the implementation became a bit messy (so could you please review it carefully, cause I'm not sure about some contracts in the Flow).
I also reflected the example you've shared with the Flow and Channel differences in collecting in a few tests, so suggestions regarding extending them are also welcome.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,469 @@
...
+        assertNextStepToBe(iterator, value = 1, done = false)
+        val second = iterator.next()
+        val third = iterator.next()
+        assertEquals(2, second.await().value)
+        assertEquals(3, third.await().value)
+        assertNextStepToBe(iterator, done = true)
+    }
+
+    @Test
+    fun testFlowToAsyncIteratorContinuesAfterCaughtCancellation() = runTest {
```
dkhalanskyjb, line 344: This test is inconsistent with JS, which does allow emissions after a `return` or a `throw` call:
```js
iter = (async function*() { try { yield 1; } catch (error) { console.log(`Nice try ${error}`); } finally { yield "Hi!"; yield "I"; yield "keep"; yield "going!"; } })()
iter.next()
iter.return() // returns: { value: "Hi!", done: false }
```
It's also inconsistent with Kotlin, which allows that, too.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,469 @@
...
+                        yield()
+                        cleanupDone = true
+                    }
+                }
+            }.asAsyncIterable()
+            val first = iterator.next()
+            val second = iterator.next()
+            if (useThrow) {
+                val error = IllegalStateException("Early exit")
+                assertSame(error, assertFailsWith<IllegalStateException> { iterator.`throw`(error).await() })
```
dkhalanskyjb: In JS, this doesn't work:
```js
iter = (async function*() { yield await new Promise(resolve => setTimeout(resolve, 5000)); yield await new Promise(resolve => setTimeout(resolve, 5000)); yield 3; })();
a = iter.next();
b = iter.next();
c = iter.throw();
await b // returns `{ value: undefined, done: false }`
```
I also don't see any logical reasons for this test to pass. The commands are supposed to enter a queue, with `return` and `throw` behaving as if they were written at the points where their `yield` happens. A future `throw` shouldn't invalidate already initiated operations.

kotlinx-coroutines-core/js/test/FlowInteropTest.kt
```diff
@@ -0,0 +1,469 @@
...
+        val error = IllegalStateException("Upstream failed")
+        val iterator = flow {
+            emit(1)
+            gate.await()
+            throw error
+        }.flowOn(Dispatchers.Unconfined).onCompletion { completed.complete(Unit) }.asAsyncIterable()
+        assertNextStepToBe(iterator, value = 1, done = false)
+        // The upstream fails while the iterator is idle (no pending `next()`): the failure must not be lost
+        gate.complete(Unit)
+        completed.await()
```
dkhalanskyjb: `onCompletion` doesn't have any right to fire while the concurrent collection is going on. That's inconsistent with the `Flow` guarantees.

dkhalanskyjb (changes requested):
I wrote a different version of `Flow` iteration to make sure I understand all nuances of the issue at hand: https://github.com/Kotlin/kotlinx.coroutines/commit/cdbd3fb883562e034eb5f5bae3cd108e461c615c

There are several tests that my implementation fails—I described in separate comments why I believe these tests are incorrect.

JSMonk:
@dkhalanskyjb I've re-worked it with the suggestion you've made, with a few changes.

The major one: I'm saving the job with `collectWhile` to use its `isActive` for the done value on `NEXT_ELEMENT`. I feel that if the job is still active, we can't say the iteration is done, isn't it?

A few minor changes: some parts were moved to a function, some classes were replaced with constants for more performance, and some functions were removed.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,253 @@
...
+     *   console.log(value)
+     * }
+     *```
+     *
+     * This API is experimental: behavior and lifecycle semantics may change in future releases.
+     */
+    @ExperimentalCoroutinesApi
+    @JsSymbol("asyncIterator")
+    public fun asAsyncIterable(): JsAsyncIterableIterator<T> {
+        lateinit var mainLoopJob: Job
```
dkhalanskyjb: This variable is completely redundant. The only way the `mainLoopJob.isActive` check in the `NEXT_ELEMENT` branch can happen is if the `elementRequests` channel is closed and an attempt to call `next()` was made. The only way for `elementRequests` to get closed is via the `elementRequests.cancel()` line—which is the last line of `mainLoopJob`, and also a non-suspending one. Since JS is single-threaded, it's guaranteed that `mainLoopJob.isActive` will be false.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,253 @@
...
+    } else {
+        resume(Unit)
+    }
+}
+
+private external interface JsArrayDeque<T> {
+    val length: Int
+    fun push(value: T)
+    fun shift(): T
+}
```
dkhalanskyjb: Some dead code.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,253 @@
...
+
+internal val <T> FlowAsyncIteratorResolution<T>.valueToReturn: T
+    inline get() = value.unsafeCast<T>()
+
+internal val FlowAsyncIteratorResolution<*>.valueToThrow: JsPromiseError
+    inline get() = value.unsafeCast<JsPromiseError>()
+
+@JsPlainObject
+internal external interface FlowAsyncIteratorResolution<T> {
```
dkhalanskyjb: Why is this odd incantation required? Is this meaningfully different from a normal `private class`?

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,253 @@
...
+    val length: Int
+    fun push(value: T)
+    fun shift(): T
+}
+
+internal val <T> FlowAsyncIteratorResolution<T>.valueToReturn: T
+    inline get() = value.unsafeCast<T>()
+
+internal val FlowAsyncIteratorResolution<*>.valueToThrow: JsPromiseError
+    inline get() = value.unsafeCast<JsPromiseError>()
```
dkhalanskyjb, line 237: Ideally, these should be properties of `FlowAsyncIteratorResolution`, and its `value` should be `private`.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,253 @@
...
+
+@JsPlainObject
+internal external interface FlowAsyncIteratorResolution<T> {
+    val resolve: (JsIteratorResult<T>) -> Unit
+    val reject: (JsPromiseError) -> Unit
+    val command: FlowCollectionCommand
+    val value: Any?
```
dkhalanskyjb, line 266: If you're going for the version without `sealed interface`, then both `command` and `value` need to be thoroughly documented so that their range of possible values is obvious.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,253 @@
...
+                                    FlowAsyncIteratorResolution.MUST_RETURN -> false
+                                    FlowAsyncIteratorResolution.NEXT_ELEMENT -> true
+                                    /* Should never happen, but if happens, better to stop collecting */
+                                    else -> true
+                                }
+                            }
+                            currentRequest.resolve(
+                                JsIteratorResult(
+                                    done = true,
+                                    value = if (currentRequest.command == FlowAsyncIteratorResolution.MUST_RETURN) currentRequest.valueToReturn else VOID
```
dkhalanskyjb: Style: our target line width is 120.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -69,106 +70,96 @@ public actual interface Flow<out T> {
...
+                                         */
+                                        currentRequest.valueToThrow.toThrowableOrNull()?.let { throw it }
+                                        false
+                                    }
+
+                                    FlowAsyncIteratorResolution.MUST_RETURN -> false
+                                    FlowAsyncIteratorResolution.NEXT_ELEMENT -> true
+
+                                    /* Should never happen, but if happens, better to stop collecting */
+                                    else -> true
```
dkhalanskyjb: Nope, this should be an `error()`. If something should never happen, it's a bug when it does happen, and we want to learn about it.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -69,106 +70,96 @@ public actual interface Flow<out T> {
...
-                    val continuation = producerContinuation
-                    if (collectionJob == null) {
-                        collectionJob = GlobalScope.launch(start = CoroutineStart.UNDISPATCHED) {
-                            try {
-                                collect(::emitValue)
-                                onFlowCompleted(null)
-                            } catch (e: Throwable) {
-                                onFlowCompleted(e)
+
+        fun startResolutionLoop() {
```
dkhalanskyjb: Why does this need to be a function, and why is this function called after the iterator object is constructed and not immediately? If the intention is to document this block of code, then adding an actual comment is better—you can add arbitrary non-camel-cased prose there.

dkhalanskyjb (changes requested):
In spirit, the proposed restructuring looks fine. It's not great that we're sacrificing maintainability and clarity of an experimental API (that we may need to evolve in the future!) for performance, but I think it still looks fairly clear, it only needs to be cleaned up.

Also, the ABI dump needs to be updated. I'm wondering, though, why this `FlowAsyncIteratorResolution` object shows up in the ABI dump at all—is this a bug in the proposed code (using `external` for something that doesn't require it), or is it a bug in the ABI validator?

JSMonk:
> In spirit, the proposed restructuring looks fine. It's not great that we're sacrificing maintainability and clarity of an experimental API (that we may need to evolve in the future!) for performance, but I think it still looks fairly clear, it only needs to be cleaned up.
> 
> Also, the ABI dump needs to be updated. I'm wondering, though, why this `FlowAsyncIteratorResolution` object shows up in the ABI dump at all—is this a bug in the proposed code (using `external` for something that doesn't require it), or is it a bug in the ABI validator?

Hmmm, I will update, but it definitely looks like a bug in the validator. It could be that the declarations generated with the `js-plain-objects` plugin are treated in some weird way by the validator.

dkhalanskyjb:
@shanshin, could you please take a look at https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildAggregated/6598337 and see why the declarations referred to in the build failure are mentioned at all? They are supposed to be `internal` but leak to the dump for some reason.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: Yes, it's creating an anonymous JS object instead, which means:
- no class code (less code generated)
- no reflection metadata needed for classes (name at least)

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: True, but it's not allowed to do so on three different levels:
- Kotlin doesn't allow having inline getters inside of the interfaces
- Kotlin/JS does not allow non-external members inside external declarations
- The JsPlainObject plugin doesn't allow non-simple properties in declarations marked with @JsPlainObject

shanshin:
> @shanshin, could you please take a look at https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildAggregated/6598337 and see why the declarations referred to in the build failure are mentioned at all? They are supposed to be internal but leak to the dump for some reason.

I can’t say for sure, since we don’t generate the dump content for klib.
But from what I can tell, inline functions are in the dump. Perhaps they were called in some public functions, and therefore they are part of the ABI.

JSMonk:
> @shanshin, could you please take a look at https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_BuildAggregated/6598337 and see why the declarations referred to in the build failure are mentioned at all? They are supposed to be `internal` but leak to the dump for some reason.

It's a bug in the plugin (I've found it). The IR moved functions don't preserve the visibility of the interface (need to be fixed on our side, until that, I re-worked the solution without of the js-plain-object plugin)

kotlinx-coroutines-core/js/src/internal/JsAsyncIteratorProtocol.kt
```diff
@@ -0,0 +1,40 @@
+@file:OptIn(ExperimentalStdlibApi::class)
+@file:Suppress("INVISIBLE_REFERENCE")
```
dkhalanskyjb: Style: it's not great to suppress errors just to get to `VOID`. We could either inline its definition here—it is short—or use some simpler equivalent if possible. Isn't `null` also treated as `undefined`?

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,287 @@
...
+     * for await (const value of flow) {
+     *   console.log(value)
+     * }
+     *```
+     *
+     * This API is experimental: behavior and lifecycle semantics may change in future releases.
+     */
+    @ExperimentalCoroutinesApi
+    @JsSymbol("asyncIterator")
+    public fun asAsyncIterable(): JsAsyncIterableIterator<T> {
```
dkhalanskyjb: Style: what is the value of exposing these return types? Since these are `internal`, I'd expect it to be impossible to use them more reasonably than passing them as opaque values to JS.

So, wouldn't it be better to return `dynamic`?

Currently, this requires `EXPOSED_FUNCTION_RETURN_TYPE`, and I'm not confident the exact behavior we're trying to force here won't break in future compiler versions.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,287 @@
...
+         * Converts a JavaScript AsyncIterator to a Kotlin Flow.
+         *
+         * The resulting flow emits items produced by the iterator until it reports completion.
+         * If a collection is canceled or fails, the iterator's `return()` method (if present) is called
+         * to close the iterator, see [from] for the details.
+         */
+        @JsStatic
+        @JsName("fromAsyncIterator")
+        @ExperimentalCoroutinesApi
+        public fun <T> from(iterator: JsAsyncIterator<T>): Flow<T> =
```
dkhalanskyjb: How are you supposed to use these functions from Kotlin at all? `JsAsyncIterator<T>` is internal, you can't construct it, you can't cast to it. Therefore, you can't call this function. We shouldn't expose any APIs that can't be called anyway, that's simply confusing.

I can imagine a version of `from` to be used from Kotlin that would accept a `dynamic`, and `from` would then dynamically dispatch the call based on whether the value is callable. I'm not sure that's the best Kotlin-side API, though.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,287 @@
...
+         * Converts a JavaScript AsyncIterable to a Kotlin Flow.
+         *
+         * The resulting flow will iterate through all values produced by the async iterable.
+         * If the flow collection is canceled or fails, the iterator's `return()` method (if present) will be called
+         * to properly clean up the async iterable, see [from] for the details.
+         */
+        @JsStatic
+        @ExperimentalCoroutinesApi
+        public fun <T> from(async: JsAsyncIterable<T>): Flow<T> =
+            from(async.asyncIterator())
```
dkhalanskyjb: A strictly better version of this would be `from { async.asyncIterator() }`.

`Flow` can be *cold*—unlike any async JS entity that I found—and converting iterables to a `Flow` creates cold flows. `listOf(1, 2, 3).asFlow()` will return `1`, `2`, and `3` for every caller. The current implementation means `Flow.from([1, 2, 3])` will only iterate over the list in the first `collect`.

For hot flows like `ReadableStream`, we also want to declare different `collect` calls as different subscribers—`Flow` doesn't have any fanout semantics, so `from` shouldn't pretend there's only a single iteration going on if multiple `collect` calls collect the same `Flow`.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,287 @@
...
+         *   the iterator's `return()` method is called (if it is present, as it is optional in the protocol)
+         *   and the promise it returns is awaited. The original exception is then rethrown; if `return()` itself fails,
+         *   its exception is attached to the original one as a suppressed exception and does not replace it.
+         */
+        @JsStatic
+        @JsName("fromAsyncGenerator")
+        @ExperimentalCoroutinesApi
+        public fun <T> from(generator: () -> JsAsyncIterator<T>): Flow<T> = flow {
+            val iterator = generator()
+            val isThereReturnMethod = jsTypeOf(iterator.`return`) == "function"
```
dkhalanskyjb: I struggle to understand why `return` and not `throw` is used to close the iterator when a throwable is thrown.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
dkhalanskyjb: This overload looks useless, as JS async iterators are also normally async iterables, aren't they?

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,287 @@
...
+                                     * Consistency with JS doesn't force us to do anything specific,
+                                     * as JS will throw `undefined` here, but we can't do that.
+                                     */
+                                    currentRequest.valueToThrow.toThrowableOrNull()?.let { throw it }
+                                    false
+                                }
+                                FlowAsyncIteratorResolution.MUST_RETURN -> false
+                                FlowAsyncIteratorResolution.NEXT_ELEMENT -> true
+                                /* Should never happen */
+                                else -> error("Unexpected command value ${currentRequest.command}. It should be either ${FlowAsyncIteratorResolution.MUST_THROW}, ${FlowAsyncIteratorResolution.MUST_RETURN}, or ${FlowAsyncIteratorResolution.NEXT_ELEMENT}")
```
dkhalanskyjb: Style: the target line width is 120, please reformat. There are other spots in this file where this limit is significantly exceeded.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
```diff
@@ -0,0 +1,287 @@
...
+     * Elements are relayed with rendezvous-style backpressure, like in an async generator: each `next()` call
+     * synchronously runs the flow until the next element is emitted (or until the flow suspends), and the flow
+     * stays suspended at `emit` until the next element is requested. Nothing is buffered.
+     *
+     * Early exit via `return` or `throw` cancels the collection and settles the returned promise only after
+     * the flow has finished its cleanup (all `finally` blocks have completed). Unlike async generators,
+     * `return`/`throw` are eager: the collection is cancelled immediately, and pending `next()` promises are
+     * resolved with `{ done: true }` rather than waiting for further elements. `throw(error)` cancels
+     * the collection with a [CancellationException] whose cause is `error`, so the flow cannot recover from it;
+     * the returned promise is rejected with `error`, or with the exception thrown by the flow's cleanup, if any.
```
dkhalanskyjb: This was changed with the latest corrections to the implementation.

dkhalanskyjb (changes requested):
I'm ready to accept the `asAsyncIterable` implementation (though I'm curious why we need to trick the compiler by exposing `internal` types from it).

`from` implementations don't seem to work on the Kotlin side, and their behavior looks weird to me, too.

kotlinx-coroutines-core/js/src/internal/JsAsyncIteratorProtocol.kt
JSMonk: Not really. if `null` is provided, it doesn't trigger the default value setting (if there is a default value parameter inside of a function)

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: It's important to set the proper type inside the generated d.ts file. Setting `dynamic` will change the type to `any`/`unknown`, so at the use-site, there will be no auto-complete for the iterator protocol. This part is essential for the TypeScript generating.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: It's not supposed to be used on the Kotlin side yet. Let me put the deprecation hidden to them

JSMonk: After we stabilize kotlinx.browser it would make sense to add it as a dependency and use the definition of AsyncIterator from there (so that it would be much useful on the Kotlin side as well)

JSMonk: But, to be fair, I would say in this case, `asAsyncIterable` should also be hidden for Kotlin (because it would be impossible to use the return type at this point, before we're using kotlinx-browser)

JSMonk: Btw, in this case, testing of them is terrible(
Since with the kotlinx it would be much better, and the methods are marked with experimental annotation, I propose to keep them with this signature visible for Kotlin, and then to change the return type to the public type from kotlinx-browser (as soon as the stable version is landed)

dkhalanskyjb: Did I get your proposal right?
1. Hide **everything** JS-async-related from Kotlin, since it's difficult or impossible to use.
2. Once `kotlinx-browser` is out, make `kotlinx-coroutines-core-js` depend on that and use its types.

That sounds okay in principle. I'm confused why async iterators, which don't seem at all browser-related, are in `kotlinx-browser`—I'd prefer us to depend on a library without "browser" in the name (`kotlinx-ecmascript2027` or something?), and ideally just the stdlib. I readily admit my ignorance w.r.t. this topic, though.

> Btw, in this case, testing of them is terrible(

How come? Doesn't `val iterator: JsAsyncIterator<Something> = something.asDynamic()[js("Symbol.asyncIterator")]()` work just the same?

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: 1. Agree, we should drop re-throwing of the `CancellationException`
2. Because `throw` doesn't make sense for the `Flow` workflow (as well as `next` with a parameter).
`throw` acts in the following way: in the `yield` point of the generator it replaces the result with throwing an error, like this:
```js
function* test() {
  try { const a = yield; } catch(e) { console.log(e) } console.log("Hello")
}

const it = test();
it.throws(new CustomError()) // log a CustomError and after "Hello"
```

So it acts like `resume(Failed(...))`. Is there any case/contract for this in `Flow`?

3. It's a rare case (for a custom AsyncIterable only, I believe), but yep, let's support it as well.

kotlinx-coroutines-core/js/src/flow/Flow.js.kt
JSMonk: No, some of them are, and some of them are not. 
There is a special type in TypeScript (which we use in the return type of `asAsyncIterable` called `AsyncIterableIterator`, but `AsyncIterator` cannot be `AsyncIterable`)

PR #4626, opened 2026-02-23
integration-testing: add kotlin_repo_url to subproject repositories f…

woainikk:
I'm refactoring train, so the kotlin artefacts from the TeamCity agent are used instead of artefacts from space. This adjustment is needed for it

### DIFF
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -39,6 +39,7 @@ buildscript {
 
     val usingSnapshotVersion = checkIsSnapshotVersion()
     val hasSnapshotTrainProperty = checkIsSnapshotTrainProperty()
+    val kotlinRepoUrl = providers.gradleProperty("kotlin_repo_url").orNull
 
     extra.apply {
         set("using_snapshot_version", usingSnapshotVersion)
@@ -49,7 +50,12 @@ buildscript {
         repositories {
             mavenLocal()
             maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+            maven("https://redirector.kotlinlang.org/maven/dev")
+            if (!kotlinRepoUrl.isNullOrBlank()) {
+                maven(kotlinRepoUrl)
+            }
         }
+
     }
 }
 
@@ -57,12 +63,17 @@ plugins {
     id("org.jetbrains.kotlin.jvm") version extra["kotlin_version"].toString()
 }
 
-repositories {
-    if (extra["using_snapshot_version"] == true) {
+allprojects {
+    repositories {
+        val kotlinRepoUrl = providers.gradleProperty("kotlin_repo_url").orNull
+        if (!kotlinRepoUrl.isNullOrBlank()) {
+            maven(kotlinRepoUrl)
+        }
+        mavenCentral()
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+        maven("https://redirector.kotlinlang.org/maven/dev")
+        mavenLocal()
     }
-    mavenLocal()
-    mavenCentral()
 }
 
 java {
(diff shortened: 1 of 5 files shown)


integration-testing/settings.gradle.kts
```diff
@@ -3,6 +3,11 @@ pluginManagement {
         mavenCentral()
         maven("https://plugins.gradle.org/m2/")
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+        val kotlinRepoUrl = providers.gradleProperty("kotlin_repo_url").orNull
+        if (!kotlinRepoUrl.isNullOrBlank()) {
+            maven(kotlinRepoUrl)
+        }
+        maven("https://redirector.kotlinlang.org/maven/dev")
```
fzhinkin: Why do we need this one here (and here only)?

integration-testing/build.gradle.kts
```diff
@@ -39,6 +39,7 @@ buildscript {
 
     val usingSnapshotVersion = checkIsSnapshotVersion()
     val hasSnapshotTrainProperty = checkIsSnapshotTrainProperty()
+    val kotlinRepoUrl = providers.gradleProperty("kotlin_repo_url").orNull
```
fzhinkin, line 42: BTW, do we need to process other additional parameters here the same way it is done in the main project?

I mean, `kotlin_language_version`, `kotlin_additional_cli_options`, etc.

Not suggesting to support them in this PR, but rather curious if it is something worth doing in the future.

woainikk: If they are not propagated from the main project, then yes, they should be supported

integration-testing/settings.gradle.kts
woainikk: we have this line (line 10) in kotlin-community/dev branch already. it was added in commit c313776b

woainikk: it seems this redirector was added everywheere where we have maven dependency to space-dev repo

integration-testing/build.gradle.kts
fzhinkin: They are not propagated: `integration-testing` is a disjoint Gradle project just residing within the main repo.

integration-testing/settings.gradle.kts
fzhinkin: So, we should also apply https://github.com/Kotlin/kotlinx.coroutines/commit/c313776b5dbfaae1fac580fc11916cffb7cd0171 to `develop`, right?

woainikk: pushed this one also)

integration-testing/build.gradle.kts
woainikk: ok, can you please add this to develop in a separate MR?

fzhinkin: sure, will do

dkhalanskyjb (changes requested):
The `repositories` block is getting unwieldy. Could you please extract it from the subprojects into `integration-testing/build.gradle.kts` to avoid the unnecessary repetition?

integration-testing/build.gradle.kts
```diff
@@ -58,11 +64,28 @@ plugins {
...
+
+        if (!kotlinRepoUrl.isNullOrBlank()) {
+            maven(kotlinRepoUrl)
+        }
+        mavenCentral()
+        maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
+        maven("https://redirector.kotlinlang.org/maven/dev")
+        mavenLocal()
+    }
 }
```
dkhalanskyjb:
```suggestion
allprojects {
    repositories {
        val kotlinRepoUrl = providers.gradleProperty("kotlin_repo_url").orNull

        if (!kotlinRepoUrl.isNullOrBlank()) {
            maven(kotlinRepoUrl)
        }
        mavenCentral()
        maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
        maven("https://redirector.kotlinlang.org/maven/dev")
        mavenLocal()
    }
}
```

PR #4632, opened 2026-03-03
[K/JS] Add exportability of ReceiveChannel and SendChannel for Kotlin…

JSMonk:
…/JS with conversion by `AsyncIterator` protocol

The follow-up MR to the https://github.com/Kotlin/kotlinx.coroutines/pull/4625

The same here, until 2.3.20 it should not be merged since there is no @JsSymbol annotation yet


dkhalanskyjb:
Hi! The `develop` branch, which is where we're doing development (instead of `master`), now uses Kotlin 2.3.21.

dkhalanskyjb:
@JSMonk, could you please rebase the PR, so that we get the CI support while we discuss the changes?

kotlinx-coroutines-core/api/kotlinx-coroutines-core.api
```diff
@@ -725,17 +725,6 @@ public final class kotlinx/coroutines/channels/ChannelIterator$DefaultImpls {
...
-public final class kotlinx/coroutines/channels/ChannelKt {
-	public static final synthetic fun Channel (I)Lkotlinx/coroutines/channels/Channel;
-	public static final fun Channel (ILkotlinx/coroutines/channels/BufferOverflow;Lkotlin/jvm/functions/Function1;)Lkotlinx/coroutines/channels/Channel;
-	public static synthetic fun Channel$default (IILjava/lang/Object;)Lkotlinx/coroutines/channels/Channel;
-	public static synthetic fun Channel$default (ILkotlinx/coroutines/channels/BufferOverflow;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lkotlinx/coroutines/channels/Channel;
-	public static final fun getOrElse-WpGqRn0 (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
-	public static final fun onClosed-WpGqRn0 (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
-	public static final fun onFailure-WpGqRn0 (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
-	public static final fun onSuccess-WpGqRn0 (Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
-}
```
dkhalanskyjb, line 737: Renaming `Channel.kt` to `Channel.common.kt` breaks binary compatibility on the JVM. If we accept this change, all callers of the `Channel()` function will need to be recompiled. This is a significant problem. Luckily, it's easy to solve: `@file:JvmName("ChannelKt")` at the top of `Channel.common.kt` restores the compatibility.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+import kotlinx.coroutines.*
+import kotlinx.coroutines.internal.recoverStackTrace
+import kotlinx.coroutines.selects.*
+import kotlinx.js.JsPlainObject
+import kotlin.internal.*
+import kotlin.js.Promise
+import kotlin.coroutines.EmptyCoroutineContext
+
+@JsImplicitExport(couldBeConvertedToExplicitExport = true)
+public actual interface SendChannel<in E> {
```
dkhalanskyjb: Could you explain the point in making `SendChannel` exportable as well? What does this entail specifically? `ReceiveChannel` is fully clear: it's our counterpart to `AsyncIterable`, we want to be able to use both in `for` loops in both Kotlin and JS, makes sense. `SendChannel` doesn't look like it's special in any regard and doesn't seem to translate to anything specific in JS. So, could we maybe keep it non-`expect`?

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
```diff
@@ -0,0 +1,79 @@
...
+    @Deprecated(
+        level = DeprecationLevel.ERROR,
+        message = "Deprecated in the favour of 'trySend' method",
+        replaceWith = ReplaceWith("trySend(element).isSuccess")
+    ) // Warning since 1.5.0, error since 1.6.0, not hidden until 1.8+ because API is quite widespread
+    public actual fun offer(element: E): Boolean {
+        val result = trySend(element)
+        if (result.isSuccess) return true
+        throw recoverStackTrace(result.exceptionOrNull() ?: return false)
+    }
```
dkhalanskyjb: Minor: this is completely unnecessary, since by the time Kotlin/Wasm/WASI was introduced to this repository, these methods have been deprecated for a long time already. There's no reason to preserve the deprecated methods here.

It would be better not to make the deprecated methods `expect`, but instead, to remove them from `Channel.common.kt` completely and re-add them for compatibility only to the targets where it's required (I guess that's only `concurrent`? Maybe Kotlin/JS as well? What are its compatibility guarantees?) as non-`actual`. Since all of them are either `HIDDEN` or have been an `ERROR` for a long time, we don't even need to bother making them `public`. They can all be `@PublishedApi internal`, with no documentation.

This is not blocking this PR, we can merge it as is and do the necessary cleanup later on our own, but if you'd like to fix this, please do!

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+            },
+            `return` = { value: @UnsafeVariance E ->
+                cancel(null)
+                Promise.resolve(JsIteratorResult(value = value, done = true))
+            }.unsafeCast<() -> Promise<JsIteratorResult<E>>>(),
+            `throw` = { err: dynamic ->
+                cancel(CancellationException("Channel was closed via AsyncIterator#throw method", err))
+                Promise.reject(err)
+            }
+        )
```
dkhalanskyjb: Major: please add KDoc documentation for these methods. They are public, and their behavior w.r.t. structured concurrency is far from being self-evident, so users will naturally want to double-check which guarantees we provide.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+
+@JsName("AsyncIterable")
+internal external interface JsAsyncIterable<out T> {
+    @JsSymbol("asyncIterator")
+    public fun asyncIterator(): JsAsyncIterator<T>
+}
+
+@JsPlainObject
+@JsName("AsyncIterator")
+internal external interface JsAsyncIterator<out T> {
```
dkhalanskyjb, line 144:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols
internal external interface JsAsyncIterator<out T> {
```

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+    public val next: () -> Promise<JsIteratorResult<T>>
+    public val `return`: () -> Promise<JsIteratorResult<T>>
```
dkhalanskyjb:
```suggestion
    // `return` and `throw` must be able to accept either zero arguments or a single one
    public val `return`: () -> Promise<JsIteratorResult<T>>
```

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+            channel.send(3)
+        }
+
+        val result1 = iterator.next().await()
+        assertEquals(1, result1.value)
+
+        // Call throw() to cancel the iterator
+        val error = js("new Error('test error')")
+        try {
+            iterator.`throw`(error).await()
```
dkhalanskyjb: Major: please add a test for a zero-argument invocation of `throw`, as well as a test for a single-argument invocation of `return`.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+                        }
+                    } else {
+                        JsIteratorResult(value = result.getOrThrow(), done = false)
+                    }
+                }
+            },
+            `return` = { value: @UnsafeVariance E ->
+                cancel(null)
+                Promise.resolve(JsIteratorResult(value = value, done = true))
+            }.unsafeCast<() -> Promise<JsIteratorResult<E>>>(),
```
dkhalanskyjb: Major: it's not obvious to me how this construct works for both zero- and one-argument invocations. Is there a clearer way to express the same? Maybe `vararg`, or a `js()` block? If this intricate magic is unavoidable, please drop a code comment explaining why it's fine to invoke a one-argument lambda without any arguments, because to a layman, this looks like a mistake.

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
```diff
@@ -0,0 +1,79 @@
...
+            "for the precise replacement please refer to the 'poll' documentation",
+        replaceWith = ReplaceWith("tryReceive().getOrNull()")
+    ) // Warning since 1.5.0, error since 1.6.0, not hidden until 1.8+ because API is quite widespread
+    public actual fun poll(): E? {
+        val result = tryReceive()
+        if (result.isSuccess) return result.getOrThrow()
+        throw recoverStackTrace(result.exceptionOrNull() ?: return null)
+    }
+
+
```
dkhalanskyjb: Style: here and in other files, the spacing between declarations is inconsistent. Here, there's an extra line, and in other places, there are missing lines between giant, multiline declarations.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+                scope.promise {
+                    val result = receiveCatching()
+                    if (result.isClosed) {
+                        when (val cause = result.exceptionOrNull()) {
+                            null, is CancellationException -> JsIteratorResult(done = true)
+                            else -> throw cause
+                        }
+                    } else {
+                        JsIteratorResult(value = result.getOrThrow(), done = false)
+                    }
```
dkhalanskyjb, line 59: What's the idea behind special-casing `CancellationException` here? Without it, it looks like the logic can be simplified to
```suggestion
                    val result = receiveCatching()
                    JsIteratorResult(value = result.getOrThrow(), done = result.isClosed)
```

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+
+class ChannelInteropTest : TestBase() {
+
+    // ===== Channel to AsyncIterator tests =====
+
+    @Test
+    fun testChannelToAsyncIteratorBasic() = runTest {
+        val channel = Channel<Int>()
+        val iterator: JsAsyncIterator<Int> = channel.asDynamic()[js("Symbol.asyncIterator")]()
+
```
dkhalanskyjb: Style: in `kotlinx.coroutines`, we mostly avoid empty lines in function bodies.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+import kotlin.test.*
+
+class ChannelInteropTest : TestBase() {
+
+    // ===== Channel to AsyncIterator tests =====
+
+    @Test
+    fun testChannelToAsyncIteratorBasic() = runTest {
+        val channel = Channel<Int>()
+        val iterator: JsAsyncIterator<Int> = channel.asDynamic()[js("Symbol.asyncIterator")]()
```
dkhalanskyjb, line 15: Why not `channel.asyncIterator()`?

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+
+        val result1 = iterator.next().await()
+        assertEquals(1, result1.value)
+
+        // Call return() to stop iteration early
+        val returnResult = iterator.`return`().await()
+        assertEquals(true, returnResult.done)
+
+        // Channel should be cancelled
+        assertTrue(channel.isClosedForReceive)
```
dkhalanskyjb: Please also check that the other elements didn't get lost by accident (in this case, that `send(2)` throws an exception; for a channel with a buffer, you'd need to pass `onUndeliveredElement` to the `Channel` constructor).

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+        assertEquals(1, result1.value)
+
+        // Call throw() to cancel the iterator
+        val error = js("new Error('test error')")
+        try {
+            iterator.`throw`(error).await()
+            fail("Should have thrown")
+        } catch (e: Throwable) {
+            // Expected
+        }
```
dkhalanskyjb: This construct won't check anything, because `fail` also simply throws an exception, which will get caught in `catch`. Here's an alternative:
```suggestion
        assertFailsWith<Throwable> {
            iterator.`throw`(error).await()
        }
```

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+
+    @Test
+    fun testChannelToAsyncIteratorThrow() = runTest {
+        val channel = Channel<Int>()
+        val iterator: JsAsyncIterator<Int> = channel.asDynamic()[js("Symbol.asyncIterator")]()
+
+        launch {
+            channel.send(1)
+            channel.send(2)
+            channel.send(3)
```
dkhalanskyjb, line 72: Like in the test above, `send(3)` is unreachable, please verify that.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+        assertEquals(3, result3.value)
+        assertEquals(false, result3.done)
+
+        val result4 = iterator.next().await()
+        assertEquals(true, result4.done)
+    }
+
+    @Test
+    fun testChannelToAsyncIteratorWithRendezvousChannel() = runTest {
+        val channel = Channel<Int>(Channel.RENDEZVOUS)
```
dkhalanskyjb: That's the default capacity for a channel, though, so this was already tested above.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+        val result1 = iterator.next().await()
+        assertEquals(1, result1.value)
+
+        // Next call should throw the exception
+        try {
+            iterator.next().await()
+            fail("Should have thrown exception")
+        } catch (e: IllegalStateException) {
+            assertEquals("Test exception", e.message)
+        }
```
dkhalanskyjb:
```suggestion
        assertFailsWith<IllegalStateException> {
            iterator.next().await()
        }.apply {
            assertEquals("Test exception", message)
        }
```

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,263 @@
...
+        repeat(100) { channel.send(it) }
+        channel.close()
+
+        val iterator: JsAsyncIterator<Int> = channel.asDynamic()[js("Symbol.asyncIterator")]()
+
+        // Read all elements
+        repeat(100) { i ->
+            val result = iterator.next().await()
+            assertEquals(i, result.value)
+            assertEquals(false, result.done)
```
dkhalanskyjb: Style: this snippet repeats often enough that it's worth extracting into a tiny separate function. Then, all the `result*` variables disappear and get replaced by `assertIteratorResult(i, false, iterator.next().await())` or even `awaitIteratorResult(i, false, iterator.next())`.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    public actual suspend fun receive(): E
+    public actual fun cancel(cause: CancellationException?)
+
+    override fun asyncIterator(): JsAsyncIterator<E> =
+        asyncIterator(CoroutineScope(EmptyCoroutineContext))
+
+    @JsExport.Ignore
+    // For Kotlin side only to be able to set up a custom scope for the iterator
+    public fun asyncIterator(scope: CoroutineScope): JsAsyncIterator<E> =
```
dkhalanskyjb: I'm not sure `scope` could be useful here. Let's maybe only expose `asyncIterator()` without parameters for now? The only role of `scope` hereis to propagate the errors coming from the closed channel, but those errors are also available in the channel itself for those interested.

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
JSMonk: So, I can just drop them in all the targets except JVM, right?

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: So, basically, in JS, if the argument isn't provided, the value becomes `undefined` (that is how it works, since we're compiling lambdas in the regular JS functions).
There is no clearer way than defining the same code in a JS block using JS syntax (the "trick" will remain the same). We can't use `vararg` since it's forbidden to use them in lambdas on the Kotlin level

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: When we call `cancel` in kotlinx.coroutines we don't want to trigger an exception on the JS side (we just want to close the channel). `CancallationException,` as far as I know, is just a marker of canceling a task without a semantic of an "exceptional" situation (if `cause` was not provided). So, this logic is needed to treat cancellation just as a closing signal and not as a thing that should be rethrown into JS

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
JSMonk: Because this is an interop test and we want to be sure that the symbol is existing on the JS object (to not appear on a situation when those tests are passing but there is no `Symbol.asyncIterator` on the object, so it can't be used on the JS side with a `for await` statement)

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
JSMonk: I want to explicitly test it in case the default is ever changed (I understand it's impossible, but as I was taught, tests should be written without implying knowledge of the implementation details)

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
JSMonk: It's controversial. In tests, I don't feel that this is an issue, since it's just an explicit assertion that could be extended/dropped in the future. I believe such utilities introduce more problems than they solve. Especially if we take into account that the same utility will be used in the follow-up Flow interop tests. I need to think more about whether we can improve readability without sacrificing explicitness (but I believe we both agree that it's not a blocker)

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
dkhalanskyjb: No, it should stay on Native as well, that is, in the `concurrent` source set (which is JVM + Native). I'm not sure about the JS source set, as I don't know how strict the Kotlin/JS compatibility guarantees are.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
dkhalanskyjb: I agree with that lesson in general, but in this case, it's not an implementation detail, it's part of the function's signature (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-channel.html):
```kotlin
fun <E> Channel(
    capacity: Int = RENDEZVOUS, 
    onBufferOverflow: BufferOverflow = BufferOverflow.SUSPEND, 
    onUndeliveredElement: (E) -> Unit? = null
): Channel<E>(source)
```

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
dkhalanskyjb: Would it make sense to write a pure JS test and invoke it to make sure everything works? The functional tests would stay in pure Kotlin, without `js()` invocations where it's avoidable, and the integration test would clear those doubts.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
dkhalanskyjb: It is indeed not a blocker, yet I don't see why this would be even slightly controversial. All of these places follow the same logic and serve the same purpose: they assert that the return result has the expected shape. If we decide to change this logic in one place, we are going to want to change every other occurrence, as well. That is a dead giveaway that this is unnecessary code duplication as opposed to just logically separate code that happens to look equivalent.

Note also that this is merely part of the test suite, a private convenience. It's not an abstraction that we are exposing to end users. Think of it as a macro.

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: Oh yeah, you're right, sorry.
The `is CancellationException` part still bothers me, though. `channel.close()` (without a `cause`) causes `exceptionOrNull()` to be `null`, and the `is CancellationException` case is only going to fire if the channel was closed *with* a cause that just happened to be a `CancellationException`. In my opinion, it's reasonable to rethrow such explicitly created `CancellationException` instances.

(A notable exception is a channel created by `produce`, which gets closed with a `JobCancellationException`—though, according to https://github.com/Kotlin/kotlinx.coroutines/issues/3658, users expect such exceptions not to be swallowed silently, either).

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: Got it, thanks! Please leave a comment in the code explaining this, so that it's easier to figure out for future readers.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,187 @@
...
+     * `ReceiveChannel` exported to JavaScript can be consumed with `for await ... of`.
+     *
+     * Each call to the iterator's `next` method receives at most one element from this channel:
+     *
+     * * if an element is available, the returned `Promise` is fulfilled with an iterator result
+     *   whose `value` is the received element and whose `done` is `false`;
+     * * if the channel is closed normally, or is cancelled with a [CancellationException], the
+     *   returned `Promise` is fulfilled with an iterator result whose `done` is `true`;
+     * * if the channel is closed with another cause, the returned `Promise` is rejected with that
+     *   cause.
```
dkhalanskyjb:
```suggestion
     * - if an element is available, the returned `Promise` is fulfilled with an iterator result
     *   whose `value` is the received element and whose `done` is `false`;
     * - if the channel is closed normally, or is cancelled with a [CancellationException], the
     *   returned `Promise` is fulfilled with an iterator result whose `done` is `true`;
     * - if the channel is closed with another cause, the returned `Promise` is rejected with that
     *   cause.
```
Stars blur together otherwise.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * Calling the iterator's `return` method cancels this channel and returns a fulfilled
+     * `Promise` with `done` set to `true`. Calling the iterator's `throw` method cancels this
+     * channel and returns a rejected `Promise` with the supplied error.
```
dkhalanskyjb: Please also mention that the channel will be cancelled with the provided cause.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * The coroutines backing calls to `next` are started in a standalone scope with an empty
+     * coroutine context. In particular, they are not children of any caller-provided coroutine
```
dkhalanskyjb:
```suggestion
     * The coroutines backing calls to `next` are started in [GlobalScope].
     * In particular, they are not children of any caller-provided coroutine
```
(with the corresponding change in code)

No need to allocate a new scope, since it's effectively the same as `GlobalScope` anyway.

dkhalanskyjb (changes requested):
Many of the earlier comments weren't addressed, please take a look.
Also, the build currently fails because of the API dump not having been updated.

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: So, we definitely would like to export the `send` method to send data from the TypeScript side; however, there is no JS-specific code, except for the use of annotations declared in the common code.

I've tried to move it back and use the annotations directly inside the common declaration and got the following errors while compiling it to JVM:
```
> Task :kotlinx-coroutines-core:compileKotlinJvm FAILED
e: file:///Users/Artem.Kobzar/Projects/jetbrains/kotlinx.coroutines/kotlinx-coroutines-core/common/src/channels/Channel.common.kt:2:13 [JAVA_MODULE_DOES_NOT_EXPORT_PACKAGE] Symbol is declared in module 'kotlin.stdlib' which does not export package 'kotlin.js'.
e: file:///Users/Artem.Kobzar/Projects/jetbrains/kotlinx.coroutines/kotlinx-coroutines-core/common/src/channels/Channel.common.kt:26:2 [JAVA_MODULE_DOES_NOT_EXPORT_PACKAGE] Symbol is declared in module 'kotlin.stdlib' which does not export package 'kotlin.js'.
```

So, it seems like we still should keep them as expect/actual for adding the annotations

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
JSMonk: Since JS target is stable, the compatibility guarantees are the same as Native

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: I'm not sure what annotations you're referring to, but please try adding `requires kotlin.js;` here: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/module-info.java#L7

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: Hmmm, inside the created `JsAsyncIterator,` we're using `cancel` with cause as `null` and with a `CancellationException("Channel was closed via AsyncIterator#throw method", err)`. As I see in the tests, if `null` is provided, the library creates a `CancellationException` with the `Channel was canceled` message (so the cause is still not null), which means we can't distinguish the user-provided `CancellationException` from the default one.
https://github.com/Kotlin/kotlinx.coroutines/blob/80a5bc71bcefa2b8df0c5e8e7aca60e06f9dc1ac/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt#L1798

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
JSMonk: It would, but I'm not sure I'm ready to rework kotlinx-coroutines infra to run the pure JS tests with all the TeamCity report support.

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: I believe it's something that should be done on the stdlib side (in `module kotlin.stdlib {`). There is no module kotlin.js (the package itself comes from stdlib)

dkhalanskyjb: Weird. Okay, then the way to go is https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-optional-expectation/ . Please introduce `internal` `expect`/`actual` annotations that you'd like to add to JS, alias them to true annotations on JS and omit them everywhere else. Take a look at the `@NoJs` annotation in `kotlinx.coroutines` for inspiration.

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
dkhalanskyjb: Alright, then let's keep the methods on JS, too.

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: I see now where you're coming from.

`Channel` handles two types of events separately: 1) *closing* the channel (which is when the sender promises not to send any new elements), and 2) *cancelling* the channel (where the *receiver* promises not to touch the channel in the future).

It is a violation of contract to first `cancel` a channel and then to attempt to receive from it. If JS can in fact call `next` after a `return`, we need to guard against it, but if a user does this, it's valid for us to throw an exception.

It doesn't seem like JS can call `next` after a `return`, though:

```js
function asyncIterator(items, delay = 100) {
  let i = 0;

  return {
    [Symbol.asyncIterator]() {
      return this;
    },

    async next() {
      console.log("Calling `next`")
      if (i >= items.length) return { done: true };

      await new Promise(resolve => setTimeout(resolve, delay));
      return { value: items[i++], done: false };
    }
  };
}

for await (const value of asyncIterator(["a", "b", "c"])) {
  console.log(value);
  return 3;
} 
```
prints `Calling next`, `a`, and returns 3.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
dkhalanskyjb: I realize that, but I didn't mean anything that drastic. I was only proposing tests like this one:
```kotlin
    @Test
    fun testChannelToAsyncIteratorEarlyReturnWithValueJs() = runTest {
        val channel = Channel<Int>()
        launch {
            channel.send(1)
            channel.send(2)
            channel.send(3)
            channel.send(4)
            channel.send(5)
        }
        val result = testChannelToAsyncIteratorEarlyReturnWithValueJsImpl(channel)
        assertEquals(42, result)
        // Channel should be cancelled
        assertTrue(channel.isClosedForReceive)
        assertFailsWith<CancellationException> { channel.receive() }
            .apply { assertEquals("Channel was cancelled", message) }
    }

    private suspend fun testChannelToAsyncIteratorEarlyReturnWithValueJsImpl(channel: Channel<Int>): Int = js("""
        const returnResult = await (async () => {
            for await (const value of channel) {
                if (value !== 1) {
                    throw "Unexpected value"
                }
                return 42;
            }
        })();
        return returnResult;
    """)
```
Unfortunately, those don't seem possible, because
> `[JSCODE_ERROR] JavaScript error: async/await statements are not supported yet`

Since we're testing the async/await functionality specifically, this is a deal-breaker.

Though it would still be nice to have at least a way to manually test the integration, even without the CI support. I don't know of a simple way to implement even that, though.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
dkhalanskyjb: This can be done by replacing `channel.send(3)` with `expectUnreached()`.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
dkhalanskyjb: The tests still don't check that the async iterator only grabs as many elements as it actually needs. For example, imagine a channel with `1`, `2`, `3` stored inside it. The async iterator grabs `1` and returns it to the user. Question: are `2` and `3` still in the channel, available to other consumers? From the current test suite, you can't tell.

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: Technically, it can; the iterator will just return `done` each time (this is the contract)

```javascript
function asyncIterator(items, delay = 100) {
  let i = 0;

  return {
    [Symbol.asyncIterator]() {
      return this;
    },

    async next() {
      console.log("Calling `next`")
      if (i >= items.length) return { done: true };

      await new Promise(resolve => setTimeout(resolve, delay));
      return { value: items[i++], done: false };
    }
  };
}

const iterator = asyncIterator(["a", "b", "c"])

await (async () => {
  for await (const value of iterator) {
    console.log(value);
    return 3;
  } 
})()

console.log(await iterator.next()) // print { value: 'b', done: false }
```

I believe you wanted to write something like this:
```javascript
async function* elements(items, delay = 100) {
  let i = 0;

  console.log("Calling `next`")

  while (i < items.length) {
    await new Promise(resolve => setTimeout(resolve, delay));
    yield items[i++];
  }
}

const iterator = elements(["a", "b", "c"])

await (async () => {
  for await (const value of iterator) {
    console.log(value);
    return 3
  } 
})()

console.log(await iterator.next()) // print { value: undefined, done: true }
```

JSMonk: So I would like to keep the contract the same.

JSMonk: P.S. throwing something behaves the same:
```javascript
async function* elements(items, delay = 100) {
  let i = 0;

  console.log("Calling `next`")

  while (i < items.length) {
    await new Promise(resolve => setTimeout(resolve, delay));
    yield items[i++];
  }
}

const iterator = elements(["a", "b", "c"])

try {

await (async () => {
  for await (const value of iterator) {
    console.log(value);
    throw 3
  } 
})()

} catch(_) {}

console.log(await iterator.next()) // print { done: true }

```

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
JSMonk: Me neither.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
JSMonk: Right now, the test is checking that after the `return`/`throws` the iterator just returns `done: true`, which means that the iterator itself will not send other data (`2` and `3`) anymore to the JS side. Isn't it enough?

JSMonk: Should I add `expectUnreached` as well?

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: From developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols :

> If an iterator returns a result with done: true, any subsequent calls to next() are expected to return done: true as well, although this is not enforced on the language level.

This means that the correct thing to do here is to use a separate variable to remember when `return`/`throw` was called, and in that case, not even touch the channel but return `done: true` immediately. This solves all our problems in one go: the `is CancellationException` check isn't needed, and the contract is satisfied.

JSMonk: Please, define "correct". In this case, a new iterator could be created, and the channel would continue from where it left off last time (which I don't believe is expected behavior for the user). `AsyncIterator` here serves as a view for the channel, meaning that manipulation of the view affects the Channel itself.

dkhalanskyjb: > the channel would continue from where it left off last time

Nope, that's not what I meant. What I meant was that `return`/`throw` should behave like it does now, cancelling the channel, *but* it should also remember the fact that it was cancelled that way. Future calls to `next` shouldn't touch the channel—they notice that the iterator was cancelled with `return`/`throw` and automatically return `true`.

The current version introduces a pitfall:

```js
// Read the first value
await (async () => {
  for await (const value of channel) {
    return value
  } 
})()

// Print the other values
await (async () => {
  for await (const value of channel) {
    console.log(value);
  } 
})()
```

This piece of code is clearly a mistake on the programmer's part. The programmer intended to stop iterating after the first element in the first loop, but then, continue to receive the other elements. In the current version, this code completes by printing nothing: the channel gets cancelled, and the new iterator returns `done: true` immediately.

The way the `Channel` API prevents this is by throwing an exception from future attempts at `receive`ing an element. This makes the programmer aware of the issue and forces them to restructure the program in a way that the channel/async iterator is consumed only once.

JSMonk: Okay, it's interesting. I thought more on it and at this moment don't think that we should even close the channel 😅

For await statement works more like a "listener" to events, and if we have multiple of them, me as a user don't expect that throwing an error or stopping listening would impact the source of events.

What do we do right now for the `Flow`? Imagine that we have thrown an error inside of the `collect` lambda, what is the impact on the `Flow` itself in this case? 

Regarding this one:
> This piece of code is clearly a mistake on the programmer's part. The programmer intended to stop iterating after the first element in the first loop, but then continue receiving the other elements. In the current version, this code completes without printing anything: the channel is canceled, and the new iterator immediately returns done: true.

With the model I describe, it's definitely not a "mistake". As I said, there is definitely a use case where the `Channel` has one sender and multiple receivers, and as a user I don't expect that throwing an error in one of the receivers will stop receiving data for the others, so I actually think that your first suggestion on having a variable is a good fit for such a behavior.

dkhalanskyjb: > What do we do right now for the `Flow`? Imagine that we have thrown an error inside of the `collect` lambda, what is the impact on the `Flow` itself in this case?

That depends on the `Flow`.
* A cold flow will stop producing elements. It can be restarted anew from different callers. `flowOf(1, 2, 3)` will emit 1, 2, and 3 to as many callers as you'd like.
* A hot flow doesn't react to errors in any one subscriber. It will keep broadcasting elements as long as there's at least one subscriber.

I don't think `Flow` is relevant here, though. It serves a completely different purpose from that of `Channel`.

> As I said, there is definitely a use case where the Channel has one sender and multiple receivers, and as a user I don't expect that throwing an error in one of the receivers will stop receiving data for the others, so I actually think that your first suggestion on having a variable is a good fit for such a behavior.

I never suggested not cancelling the channel, because the one-sender-multiple-receivers use case isn't supported by native JS iterators, either:

```js
async function* fetchUserData(userIds) {
  for (const id of userIds) {
    const response = await new Promise(resolve => setTimeout(resolve, 100));
    yield id;
  }
}

const users = fetchUserData([1, 2, 3, 4, 5]);

await (async () => {
  for await (const user of users) {
    console.log(`Processing ${user}`);
    return 0;
  }
})();
console.log("Got the first element");

await (async () => {
  for await (const user of users) {
    console.log(`Processing ${user}`);
  }
  console.log("Nothing.");
})();
```

It looks like the logic in JS is that each object may only be iterated once, after which it can be cleaned up.

See also https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration:

> By default, exiting the loop will also cancel the stream, so that it can no longer be used. To continue to use a stream after exiting the loop, pass `{ preventCancel: true }` to the stream's `values()` method

If there's a user demand, we could introduce similar configurability in the future, but this parameter isn't very popular in JS: see the number of hits in https://grep.app/search?q=preventCancel%3A vs. https://grep.app/search?q=ReadableStream.

One of the stated goals of this API was familiarity to JS developers who want not to even be aware of the underlying Kotlin. If not for that goal, you'd be right: iterating over a channel in Kotlin using `for (element in channel) { }` will indeed not cancel the channel when one of the consumers stops. In Kotlin, to get the behavior I'm proposing, you'd need to call the `consumeEach` function: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/consume-each.html

Notably, JS also behaves exactly like the current version—which is to say, pretends the remainder of the sequence is empty for future attempts at iterating—but we can do better than that by notifying the programmer of the issue.

> there is definitely a use case where the Channel has one sender and multiple receivers, and as a user I don't expect that throwing an error in one of the receivers will stop receiving data for the others

I agree that such a use case exists, and it is solved by manually invoking `receive` from multiple consumers.

JSMonk: I'm thinking more and more about not canceling it on break/return/throw (so I'm going to re-implement it a bit).

ReadableStream provides guarantees of a single reader:
> While iterating, the stream is locked to prevent other consumers from acquiring a reader (attempting to iterate over a stream that is already locked will throw a TypeError). This lock is released when the loop exits.

And with the behavior, it makes sense to cancel it if the reading was stopped by return/throw. In our case, we don't provide such exclusive read guarantees, so it could be unexpected that `break` in a JS loop canceled the whole channel (imagine trying to debug it). Since we also export the `cancel` method, and it could be done explicitly, let me re-work it to not cancel the whole channel on `break/return/throws`

dkhalanskyjb: `consumeEach` doesn't have the single-reader limitation, either, and no one complained about it.

If you could present an idiomatic JS API exposing async iterators that can be used from multiple receivers, let's discuss it. Otherwise, `break` also breaking the sequence looks like the norm in the JS world. Both of the examples I've looked at—function generators and `ReadableStream`—follow it. Our debugging experience would be at least better than the JS APIs if we throw an exception on repeated subscription attempts.

JSMonk: Additionally, since the discussion started because of the `is CancelationException` check, I also feel that it's not natural for a JS user to get a `CancelationException` if, on any side (Kotlin or TypeScript), the channel is closed without an error (just canceled). As for example, for the following ReadableStream:
```javascript
let controller;

const stream = new ReadableStream({
    start(c) {
        controller = c;
    }
});

// Producer: push one chunk every 300ms
let i = 0;
const producer = setInterval(() => {
    i += 1;
    console.log("enqueue", i);
    controller.enqueue(`chunk-${i}`);
}, 300);

// "Accidental" external close after 1.2s
setTimeout(() => {
    console.log(">>> closing stream from outside");
    clearInterval(producer);
    controller.close(); // normal close, not an error
}, 1200);

(async () => {
    try {
        for await (const chunk of stream) {
            console.log("read:", chunk);
        }
        console.log("for-await ended normally (done: true)");
    } catch (err) {
        // This will NOT run on close; only on actual error
        console.error("for-await failed:", err);
    }
})();
```

We will have just `done: true` at the end:
```
enqueue 1
read: chunk-1
enqueue 2
read: chunk-2
enqueue 3
read: chunk-3
>>> closing stream from outside
for-await ended normally (done: true)
```

Because of that, I still believe that it's an important check to preserve this behavior

JSMonk: Replacing the `close` with an error will produce the error on the reader side:
```javascript
    // controller.close(); // normal close, not an error
    controller.error(new Error("Accidental failure"));
```

dkhalanskyjb: > the channel is closed without an error (just canceled)

Cancelling a channel means closing it with an error.

JSMonk: My point is, that in the following code, we don't want to receive any exception on the JS side:
```kotlin
@JsExport
suspend fun getProducer(): Channel<String> {
    val channel = Channel<String>()

    val producer = launch {
        var i = 0
        while (true) {
            i++
            val chunk = "chunk-$i"
            println("send: $chunk")
            channel.send(chunk)
            delay(300)
        }
    }

    launch {
        delay(1200)
        println(">>> closing channel from outside")
        producer.cancelAndJoin()
        channel.close() 
    }

    return channel
}
```

```javascript

const stream = getProducer()

(async () => {
    try {
        for await (const chunk of stream) {
            console.log("read:", chunk);
        }
        console.log("for-await ended normally (done: true)");
    } catch (err) {
        // This will NOT run on close; only on actual error
        console.error("for-await failed:", err);
    }
})();
```

In this case, we definitely don't want to fall into `catch` with the `CancelationException`

dkhalanskyjb: I agree! Please see https://pl.kotl.in/DObzFuVWs: even without `is CancellationException`, your code will not throw an exception in this scenario.

JSMonk: Awesome. Thank you for noticing it. Let me re-work!

kotlinx-coroutines-core/common/src/utils/JsOptionalExport.kt
```diff
@@ -0,0 +1,5 @@
+package kotlinx.coroutines.utils
```
dkhalanskyjb: We have the `internal` package for such purposes, let's not introduce a new one.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,160 @@
...
+
+        return JsAsyncIterator(
+            next = {
+                GlobalScope.promise {
+                    if (wasEarlyFinished) return@promise JsIteratorResult(done = true)
+
+                    val result = receiveCatching()
+                    result.exceptionOrNull()?.let { throw it }
+
+                    if (result.isClosed) {
```
dkhalanskyjb, line 55: From https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md:

> Do not add extra newlines in function bodies: if you feel that blocks of code should be logically separated, then separate them with a comment instead.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -0,0 +1,160 @@
...
+            "for the precise replacement please refer to the 'poll' documentation",
+        replaceWith = ReplaceWith("tryReceive().getOrNull()")
+    ) // Warning since 1.5.0, error since 1.6.0, not hidden until 1.8+ because API is quite widespread
+    public actual fun poll(): E? {
+        val result = tryReceive()
+        if (result.isSuccess) return result.getOrThrow()
+        throw recoverStackTrace(result.exceptionOrNull() ?: return null)
+    }
+
+
```
dkhalanskyjb: Style: extra newline.

kotlinx-coroutines-core/wasmWasi/src/channels/Channel.kt
```diff
@@ -0,0 +1,57 @@
...
+    @LowPriorityInOverloadResolution
+    @Deprecated(
+        message = "Deprecated in favor of 'receiveCatching'. " +
+            "Please note that the provided replacement does not rethrow channel's close cause as 'receiveOrNull' did, " +
+            "for the detailed replacement please refer to the 'receiveOrNull' documentation",
+        level = DeprecationLevel.ERROR,
+        replaceWith = ReplaceWith("receiveCatching().getOrNull()")
+    ) // Warning since 1.3.0, error in 1.5.0, cannot be hidden due to deprecated extensions
+    public actual suspend fun receiveOrNull(): E? = receiveCatching().getOrNull()
+    @Suppress("DEPRECATION_ERROR")
```
dkhalanskyjb, line 51: Style: missing newline between these two large definitions.

kotlinx-coroutines-core/wasmJs/src/channels/Channel.kt
```diff
@@ -0,0 +1,57 @@
...
+    @LowPriorityInOverloadResolution
+    @Deprecated(
+        message = "Deprecated in favor of 'receiveCatching'. " +
+            "Please note that the provided replacement does not rethrow channel's close cause as 'receiveOrNull' did, " +
+            "for the detailed replacement please refer to the 'receiveOrNull' documentation",
+        level = DeprecationLevel.ERROR,
+        replaceWith = ReplaceWith("receiveCatching().getOrNull()")
+    ) // Warning since 1.3.0, error in 1.5.0, cannot be hidden due to deprecated extensions
+    public actual suspend fun receiveOrNull(): E? = receiveCatching().getOrNull()
+    @Suppress("DEPRECATION_ERROR")
```
dkhalanskyjb, line 51: Ditto.

kotlinx-coroutines-core/js/test/ChannelInteropTest.kt
```diff
@@ -0,0 +1,208 @@
...
+        assertNextStepToBe(iterator, value = 1, done = false)
+        // Call throw() to cancel the iterator
+        val error = js("new Error('test error')")
+        assertFailsWith<Throwable> { iterator.`throw`(error).await() }
+            .apply { assertEquals("test error", message) }
+        // Channel should not be cancelled
+        assertFalse(channel.isClosedForReceive)
+        assertNextStepToBe(iterator, done = true)
+        assertEquals( 2, channel.receive())
+        assertEquals( 3, channel.receive())
```
dkhalanskyjb:
```suggestion
        assertEquals(2, channel.receive())
        assertEquals(3, channel.receive())
```

dkhalanskyjb (changes requested):
Almost there! I have no objections to the API or what it does. There are a couple of small stylistic/maintenance concerns, and once those are fixed, we can merge this!

dkhalanskyjb (approved):

Issue #3458, opened 2022-09-22
Chaining `rxSingle` calls that use `Unconfined` dispatcher and `blockingGet` results in deadlock


byencho:
Consider the follow setup in which two `rxSingle` calls are chained together and the second one uses `blockingGet()`:

```kotlin
fun main() {
    val dispatcher1 = Dispatchers.Unconfined
    val dispatcher2 = Dispatchers.Unconfined

    val disposable = rxSingle(dispatcher1) { }
        .map {
            rxSingle(dispatcher2) { }
                .blockingGet() // <----- This code deadlocks here
        }
        .timeout(100, TimeUnit.MILLISECONDS)
        .subscribe(
            { println("Success") },
            { println("Error : $it") }
        )

    while (!disposable.isDisposed) {
        // Wait for request to end
    }
}
```

(Set aside that `blockingGet()` is not a good practical choice here...this is just to highlight the problem.)

The above triggers the timeout because the `blockingGet()` call is deadlocked:

> Error : java.util.concurrent.TimeoutException: The source did not signal an event for 100 milliseconds and has been terminated.

This happens even if the suspending code inside `rxSingle` operates with a different dispatcher:

```kotlin
fun main() {
    val dispatcher1 = Dispatchers.Unconfined
    val dispatcher2 = Dispatchers.Unconfined

    val disposable = rxSingle(dispatcher1) {
        withContext(Dispatchers.IO) {}
    }
        .map {
            rxSingle(dispatcher2) {
                withContext(Dispatchers.IO) {}
            }
                .blockingGet()
        }
        .timeout(100, TimeUnit.MILLISECONDS)
        .subscribe(
            { println("Success") },
            { println("Error : $it") }
        )

    while (!disposable.isDisposed) {
        // Wait for request to end
    }
}
```

The code completes successfully and prints `"Success"` if either `dispatcher1` or `dispatcher2` is set to `Dispatchers.Default` / `Dispatchers.IO` etc. so it appears that the use of `Unconfined` on both is key to the issue.

My main question is: is this expected? If so, why does it require both requests to use `Unconfined` to see the issue? Is this because of the "event-loop" that gets involved here for nested coroutines:

> Nested coroutines launched in this dispatcher form an event-loop to avoid stack overflows.

[...]

PR #4633, opened 2026-03-04
Follow CoroutineDispatcher.dispatch specification in NamedDispatchers

jDramaix:
This change updates NamedDispatchers to delegate to Dispatchers.Default instead
of executing the block synchronously. This follows the CoroutineDispatcher
specification which states that dispatch should not invoke the block directly.

This also fixes an issue where coroutines were resumed before being suspended
in some tests.

As a consequence, some tests that relied on synchronous execution of
NamedDispatchers have been updated to remove non-deterministic expect calls
or to use runTest instead of withVirtualTime.

### DIFF
--- a/kotlinx-coroutines-core/common/test/flow/NamedDispatchers.kt
+++ b/kotlinx-coroutines-core/common/test/flow/NamedDispatchers.kt
@@ -24,14 +24,6 @@ public object NamedDispatchers {
             // to this, the `dispatch` function documentation states that it cannot invoke the block
             // directly. Our implementation is following the specification by delegating the
             // dispatch to the Default dispatcher.
-            // original code:
-            // stack.push(name)
-            // try {
-            //     block.run()
-            // } finally {
-            //     val last = stack.pop() ?: error("No names on stack")
-            //     require(last == name) { "Inconsistent stack: expected $name, but had $last" }
-            // }
             Dispatchers.Default.dispatch(context) {
                 stack.push(name)
                 try {
--- a/kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
+++ b/kotlinx-coroutines-core/common/test/flow/operators/TimeoutTest.kt
@@ -160,8 +160,6 @@ class TimeoutTest : TestBase() {
     // when `NamedDispatchers` was a fake dispatcher, but it is now a real dispatcher.
     // Hopefully, on JS, this test can simply use `runTest` and doesn't require virtual time. The
     // delay is sufficiently large compared to the timeout to prevent flakiness.
-    // original code:
-    // fun testUpstreamTimeoutIsolatedContext() = withVirtualTime {
     fun testUpstreamTimeoutIsolatedContext() = runTest {
         val flow = flow {
             assertEquals("upstream", NamedDispatchers.name())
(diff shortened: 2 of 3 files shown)


dkhalanskyjb:
The KDoc of `dispatch` needs to be fixed, thanks for noticing that! We will have to update the documentation of `dispatch` to explicitly allow this. The contract is already contradictory (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/dispatch.html):

> The dispatcher guarantees that `block` will eventually execute, typically by dispatching it to a thread pool, using a dedicated thread, or **just executing the block in place**.

> This method **must not immediately call `block`**. Doing so may result in StackOverflowError when dispatch is invoked repeatedly, for example when [yield](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html) is called in a loop.

In any scenario, we will not accept this pull request. It does not improve anything for the end users, since it only changes the code of tests. It also makes the lives of maintainers more difficult, since with your edits, the tests aren't checking the behavior that we want them to check. Everyone loses.

jDramaix:
I'd like to clarify a few points regarding the intent and technical background of this PR.

The initial goal of this Pull Request was primarily to show to @JSMonk (following our online discussion) the patch we had to implement internally to address issues we encountered. This was intended as a draft PR and was not meant for a formal review or to be submitted to the main branch.

The reason we had to implement this patch is that the current implementation of `NamedDispatcher` is problematic in certain environments. Specifically, it behaves like an unconfined dispatcher without being a "real" one, which can lead to coroutines being resumed before they have actually been suspended.

Supporting this behavior in our J2CL (Kotlin/Closure compiler) implementation of Coroutines adds significant and unnecessary complexity. It would require the compiler to maintain state within coroutine intrinsics such as `suspendCoroutineUninterceptedOrReturn`, which essentially duplicates logic already handled by `SafeContinuation`. Since `NamedDispatcher` is a test-only utility, we believe it does not make sense to compromise a clean compiler implementation just to support a "simplification" of tests. Furthermore, this specific behavior (resuming before suspension) may also prove difficult or impossible to support in Kotlin/WASM in the future.

> the tests aren't checking the behavior that we want them to check,

The change does not result in a loss of coverage. The primary objective of these impacted tests is to verify that the correct dispatchers are being utilized for flow operations. Asserting on the exact synchronous execution timing does not add meaningful value to these specific tests. As long as the end result is correct, it confirms that all flow operations executed as expected.

dkhalanskyjb:
> Since `NamedDispatcher` is a test-only utility, we believe it does not make sense to compromise a clean compiler implementation just to support a "simplification" of tests.

Allowing `dispatch` to execute the `block` in-place isn't just for simplifying the tests. Our users rely on the ability to write such a dispatcher, too. Example: https://github.com/androidx/androidx/blob/6d53f95e5d979366cf7935ad7f4f14f76a951ea5/testutils/testutils-ktx/src/main/java/androidx/testutils/DirectDispatcher.kt#L17
Sometimes, such a dispatcher is the best way forward. In particular, it's useful when interacting with other asynchronous systems that do rely on such immediate execution: https://github.com/Kotlin/kotlinx.coroutines/issues/3458#issuecomment-1258060548

PR #4635, opened 2026-03-05
Train setup: get rid of kotlin_snapshot_version parameter. Use kotlin_version instead

woainikk:
I'm doing Train setup refactoring and gonna get rid of kotlin_snapshot_version parameter, as it just duplicates kotlin_version parameter and creates confusion. DO NOT MERGE - will merge as soon as train changes are accepted

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -28,11 +28,6 @@ val gradleProperties = Properties().apply {
 }
 
 fun version(target: String): String {
-    // Intercept reading from properties file
-    if (target == "kotlin") {
-        val snapshotVersion = properties["kotlin_snapshot_version"]
-        if (snapshotVersion != null) return snapshotVersion.toString()
-    }
     val version = "${target}_version"
     // Read from CLI first, used in aggregate builds
     return properties[version]?.let{"$it"} ?: gradleProperties.getProperty(version)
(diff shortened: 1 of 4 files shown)


buildSrc/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ val gradleProperties = Properties().apply {
 fun version(target: String): String {
     // Intercept reading from properties file
     if (target == "kotlin") {
-        val snapshotVersion = properties["kotlin_snapshot_version"]
+        val snapshotVersion = properties["kotlin_version"]
         if (snapshotVersion != null) return snapshotVersion.toString()
     }
```
dkhalanskyjb: This code is now redundant.

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
```diff
@@ -124,8 +124,7 @@ fun Project.configureCommunityBuildTweaks() {
  */
 fun getOverriddenKotlinVersion(project: Project): String? =
     if (isSnapshotTrainEnabled(project)) {
-        project.providers.gradleProperty("kotlin_snapshot_version").orNull
-            ?: error("'kotlin_snapshot_version' should be defined when building with a snapshot compiler")
+        project.providers.gradleProperty("kotlin_version").orNull
     } else {
         null
     }
```
dkhalanskyjb: This whole function can also be removed now (with the corresponding edit to `buildSrc/src/main/kotlin/Projects.kt`).

PR #4636, opened 2026-03-06
Update Flow.cancellable() and collect() KDoc to mention prompt cancel…

### LINKED ISSUE #3803: Flow.cancellable() documentation still describes atomic cancellation instead of prompt cancellation guarantee
**Describe the bug**

Since #1937 `Flow.cancellable()` and `Flow.collect()` (just like the rest of Channels, Job, etc.) use prompt cancellation guarantee as the gold standard for cancellation.

However, the documentation for `Flow.cancellable()` (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/cancellable.html) still mentions:

> Returns a flow which checks cancellation status on each emission

`Flow.collect()` (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/collect.html) doesn't even mention cancellation and the `Flow` main page (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/) only mentions using `AbstractFlow` prevents common errors related to cancellation.

This all seem to imply that a collector's coroutine wouldn't be cancelled until the next emission of a flow and does not really match prompt cancellation guarantee (as far as I understand it.

Should the documentation be updated?

As it currently reads, this implies potential memory leak issues. For example:

A flow that emits somewhat infrequently with a collector that references big objects (e.g., some UI objects). The call to collect is stored in a `val collector = launch { flow.collect { doStuffWithTheUI() } }`. When the collector's cancelled, the coroutine and the big UI it references remain alive until the `flow` emit a new value (which could be a long time afterward).

Am I simply understanding the documentation wrong or should this be made more explicit?

Even though cancellation is cooperative and a coroutine could choose to run for a long time after being cancelled, "prompt cancellation guarantee" implies the coroutine's closure should return/be made available to the GC for collection "promptly" once cancellation is received.

(Apologies if this is the wrong avenue).

**Provide a Reproducer**
No reproduction steps as this is a documentation issue.

glwithu06:
## Summary
Update KDoc for `Flow.cancellable()` and `Flow.collect()` to mention the **prompt cancellation guarantee**, addressing outdated documentation that only described atomic-style cancellation checking.

Fixes #3803

## Changes
- **`Flow.cancellable()`**: Reworded to explicitly state it provides a prompt cancellation guarantee, and clarified when this operator is needed (custom `Flow` implementations not using the `flow` builder).
- **`Flow.collect()`**: Added a "Cancellation" section explaining that `flow` builder and `SharedFlow` implementations are cancellable by default, and that the `cancellable` operator can be used for other implementations.

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Context.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Context.kt
@@ -252,9 +252,13 @@ public fun <T> Flow<T>.flowOn(context: CoroutineContext): Flow<T> {
 }
 
 /**
- * Returns a flow which checks cancellation status on each emission and throws
- * the corresponding cancellation cause if flow collector was cancelled.
- * Note that [flow] builder and all implementations of [SharedFlow] are [cancellable] by default.
+ * Returns a flow that provides a **prompt cancellation guarantee**: on each emission, the
+ * flow collector's cancellation status is checked, and if the collector was cancelled,
+ * the corresponding cancellation cause is thrown.
+ *
+ * The [flow] builder and all implementations of [SharedFlow] are [cancellable] by default.
+ * This operator is useful for custom [Flow] implementations or flows created without the [flow]
+ * builder to ensure they promptly respond to the collector's cancellation.
  *
  * This operator provides a shortcut for `.onEach { currentCoroutineContext().ensureActive() }`.
  * See [ensureActive][CoroutineContext.ensureActive] for details.
--- a/kotlinx-coroutines-core/common/src/flow/Flow.kt
+++ b/kotlinx-coroutines-core/common/src/flow/Flow.kt
@@ -183,6 +183,12 @@ public interface Flow<out T> {
      * myFlow.collect { value -> println("Collected $value") }
      * ```
      *
+     * ### Cancellation
+     *
+     * Flows built with the [flow] builder and all [SharedFlow] implementations provide a
+     * **prompt cancellation guarantee**: the collector's cancellation status is checked on each emission.
+     * For other [Flow] implementations, the [cancellable] operator can be used to ensure the same guarantee.
+     *
      * ### Method inheritance
      *
      * To ensure the context preservation property, it is not recommended implementing this method directly.


kotlinx-coroutines-core/common/src/flow/operators/Context.kt
```diff
@@ -252,9 +252,13 @@ public fun <T> Flow<T>.flowOn(context: CoroutineContext): Flow<T> {
 }
 
 /**
- * Returns a flow which checks cancellation status on each emission and throws
- * the corresponding cancellation cause if flow collector was cancelled.
- * Note that [flow] builder and all implementations of [SharedFlow] are [cancellable] by default.
+ * Returns a flow that provides a **prompt cancellation guarantee**: on each emission, the
+ * flow collector's cancellation status is checked, and if the collector was cancelled,
+ * the corresponding cancellation cause is thrown.
```
dkhalanskyjb, line 257: The problem is, this is not what "prompt cancellation" means. "Prompt cancellation" refers to a very specific scenario where the coroutine is ready to resume with a value after suspending, but throws a `CancellationException` anyway.
The term is introduced here: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/suspend-cancellable-coroutine.html.

`cancellable()` is simply unrelated to that: it checks for cancellation on each emission, and that's it. Likewise, we aren't saying that `ensureActive` provides the prompt cancellation guarantee. The coroutine may not even suspend, but the effect will still get triggered, this is fully orthogonal!
* https://pl.kotl.in/fJ3goL2lh is an example of when `cancellable()` introduces cancelling behaviors even when no suspension happens at all, as the tracking dispatcher shows.
* On the opposite side, `buffer` may introduce asynchronous behavior that won't make cancellation be observed by default, even though it's susceptible to prompt cancellation: https://pl.kotl.in/olmd02Ahx

glwithu06: Thank you for the detailed explanation! I now understand that "prompt cancellation" has a very specific meaning tied to suspendCancellableCoroutine (cancellation at the resume point after suspension), which is orthogonal to what cancellable() does . I'll revisit the documentation with this distinction in mind and update the PR accordingly.

PR #4638, opened 2026-03-09
Fix compilation error after KT-84185

GrigoriiSolnyshkin:


GrigoriiSolnyshkin:
For some reason, this hit was an error instead of warning for me: see [build](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_UserProjectsAggregateK2/894879517?buildTab=artifacts), the diagnostic is `TYPE_ARGUMENTS_NOT_ALLOWED`. Is it true that the project is built with progressive mode?

dkhalanskyjb:
> Is it true that the project is built with progressive mode?

Yes: https://github.com/Kotlin/kotlinx.coroutines/blob/b11abdf01d4d5db85247ab365abc72efc7b95062/buildSrc/src/main/kotlin/GlobalKotlinCompilerOptions.kt#L4

GrigoriiSolnyshkin:
I see. Then it is the expected behavior

fzhinkin (approved):

Issue #4628, opened 2026-02-24
Ability to add a custom message into exception thrown by withTimeout


dant3:
Sorry if this is a duplicate or superseeded by something else.

**Use case**
Consider the code like:

```
withTimeout(timeout1) {
   // some function awaiting for result
}

withTimeout(imeout2) {
   // some other awaiting logic
}
```

There is no way to tell which one of those thrown TimeoutCancellationException especially if the timeout value is the same. Currently the only cumbersome alternatives exist are:

```
try {
   withTimeout ...
} catch (ex: TimeoutCancellationException) {
    throw MyException("this is timeout1");
}

// or...

withTimeoutOrNull(timeout1) {
} ?: throw MyException("timeout1")
```

**The Shape of the API**

This could be just as simple as an optional parameter. Consider junit assertions as example: by default you can assert without any specification and when assertion fails there is not much of the specifics. However you can specify an optional parameter `message` and this message will be added to the AssertionError, making it easier to understand what went south. In case of kotlin coroutines stacktraces can be hard to figure out, so this becomes even more important.

Proposal
```
withTimeout(duration, name = "waiting for X") {
     // waiting for X
}
// produces TimeoutCancellaionException("Timed out waiting for 15000 ms: waiting for X")
```
Alternatively, could use CoroutineName

```
withTimeout(duration, context = CoroutineName("waiting for X")) {
   // waiting for X
}

// produces TimeoutCancellaionException("Coroutine(\"waiting for X\") timed out waiting for 15000 ms")
```

dkhalanskyjb:
Hi! This is a new issue, not a duplicate.

1. The proposed options don't look like idiomatic Kotlin to me. "`doSomething`" and "`doSomethingOrNull`" pairs are common in Kotlin in general, and I don't think a single one of them allows parameterizing `doSomething` with the additional information for the exception message. Yet your arguments are just as applicable to `single`/`singleOrNull`, `first`/`firstOrNull`, etc., are they not? If there are several `first` calls in a row, it may be helpful to know which one failed, for example. Do you see any differences between `withTimeout` and `single` in this regard, or do you consider `userRepo.withId(id).single(name = "the user with the given ID")` a valuable addition as well?
[...]

PR #4639, opened 2026-03-15
Include CoroutineName in TimeoutCancellationException message

wwan13:
This PR includes CoroutineName in TimeoutCancellationException messages when available.

It does not introduce any new public API and follows the direction discussed in #4628.

### DIFF
--- a/kotlinx-coroutines-core/common/test/WithTimeoutTest.kt
+++ b/kotlinx-coroutines-core/common/test/WithTimeoutTest.kt
@@ -205,22 +205,11 @@ class WithTimeoutTest : TestBase() {
         val exception = assertFailsWith<TimeoutCancellationException> {
             withContext(CoroutineName("waiting for x")) {
                 withTimeout(1) {
-                    delay(10)
+                    awaitCancellation()
                 }
             }
         }
-
         assertTrue(exception.message!!.contains("waiting for x"))
-    }
-
-    @Test
-    fun testTimeoutCancellationExceptionWithoutCoroutineName() = runTest {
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/common/src/Timeout.kt
```diff
@@ -182,9 +184,15 @@ public class TimeoutCancellationException internal constructor(
...
-    coroutine: Job
-) : TimeoutCancellationException {
-    val message = (delay as? DelayWithTimeoutDiagnostics)?.timeoutMessage(time.milliseconds)
+    coroutine: Job,
+    coroutineName: String?
+): TimeoutCancellationException {
+    val baseMessage = (delay as? DelayWithTimeoutDiagnostics)?.timeoutMessage(time.milliseconds)
         ?: "Timed out waiting for $time ms"
+    val message = if (coroutineName != null) {
+        "Coroutine \"$coroutineName\" $baseMessage"
```
dkhalanskyjb: `Coroutine "Something" Timed out` doesn't look right. What do you think of decapitalizing the first letter of the message? For example, like so: https://pl.kotl.in/NVZZQ5Q6t

wwan13: Good catch, applied your suggestion. Thanks!

dkhalanskyjb:
Thank you! With this change now in `develop`, do you consider #4628 fixed, or would you like to keep the issue open for now?

wwan13:
Yes, this change resolves the issue, so it can be closed. Thanks!

PR #4641, opened 2026-03-18
Fix most of the unresolved links detected by Dokka/IDEA in the `main` source-sets

whyoleg:
_Based on the work related to https://github.com/Kotlin/dokka/pull/4479 where I tried to enable `failOnWarning=true` for all kotlinx projects._

The unresolved link fixes are split into different categories (and so commits):
- https://github.com/Kotlin/kotlinx.coroutines/commit/4a8a20c8870261f7293e0bf952a12fa061c416b6: accidentally unresolved
- https://github.com/Kotlin/kotlinx.coroutines/commit/741fe194807b69f214b8b4041833e86e41eb70b8: links to not accessible declarations, includes:
    - functions from modules, not from dependencies like `Flow.asPublisher` and `Publisher.asFlow` in the `core` module.
    - declarations from more platform-specific source sets, e.g., `Dispatchers.IO` or `ServiceLoader` in `commonMain`.
- https://github.com/Kotlin/kotlinx.coroutines/commit/998e916693624d6a6fd599aad998d284c7f21f0b: links to `HIDDEN` declarations
    - Note: this is unresolved only with the new KDoc links resolution.
- https://github.com/Kotlin/kotlinx.coroutines/commit/830e8ee7ac9098efeb550920a104a14866a95be7: **most likely** accidentally unresolved
    - **But** I'm not sure if the replacement is correct, as it touches coroutines internals...
    - Also, I haven't found a replacement for `describeRemoveFirst`.
- https://github.com/Kotlin/kotlinx.coroutines/commit/c2b0583e777f15e305847a76190125584b1f64ef: There is no way to make this link resolvable at all, so I'm not sure what to do here. Current state:
    - `InterruptedException` will always be highlighted by IDEA in yellow (warning)
    - `InterruptedException` will always be reported as unresolved by Dokka, but still render as plain text there

There are still unresolved links present in `module-level` documentation, because `README.md` is used both as documentation inside the repo, as well as an input for Dokka module documentation:
- sometimes it contains links to [local files](https://github.com/Kotlin/kotlinx.coroutines/blob/bf9509da7a0eff08e489555254d9662f2cc1894f/ui/kotlinx-coroutines-android/README.md?plain=1#L18-L23), which could not be resolved by Dokka.
- Some files [heavily](https://github.com/Kotlin/kotlinx.coroutines/blob/74cf9ea8598a9ab1328faa52ddaea1bb977d4470/kotlinx-coroutines-core/common/README.md?plain=1#L93-L153) use `knit` ability to add links to the website. At the same time, Dokka does support the resolution, at least `FQN` links in module docs, and even some relative links like `launch`, but not all of them: [`[CoroutineDispatcher]`](https://github.com/Kotlin/kotlinx.coroutines/blob/74cf9ea8598a9ab1328faa52ddaea1bb977d4470/kotlinx-coroutines-core/common/README.md?plain=1#L14) is not resolved for some reason...

**Ideally**, the usage should be split into two distinct files: one for Dokka and, if needed, one for README, like it's done in kotlinx-io ([per-module](https://github.com/Kotlin/kotlinx-io/blob/c06a6016db01df8bea53273a51932a4f4edd1359/core/Module.md)) or in kotlinx-serialization ([per-whole-project](https://github.com/Kotlin/kotlinx.serialization/blob/c0cac7cd981a4712bbc974b5efb59573532f93c7/dokka/moduledoc.md)).

**Note**: all of those links are unresolved and reported by Dokka (`master`) and IDEA (26.1) with [new KDoc links resolution](https://github.com/Kotlin/dokka/issues/4425) based on [KEEP#0389](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0389-kdoc-streamline-ambiguous-KDoc-links.md). With Dokka 2.2.0-Beta, most of those links are also unresolved, where new resolution is **not** enabled yet.

---

Feel free to ask questions or let me know if you just don't agree with something here. I'm open to discussions on how to resolve this. :)

### DIFF
--- a/kotlinx-coroutines-core/concurrent/src/channels/Channels.kt
+++ b/kotlinx-coroutines-core/concurrent/src/channels/Channels.kt
@@ -25,7 +25,7 @@ import kotlin.jvm.*
  *
  * For this operation it is guaranteed that [failure][ChannelResult.failed] always contains an exception in it.
  *
- * @throws `InterruptedException` on JVM if the current thread is interrupted during the blocking send operation.
+ * @throws InterruptedException on JVM if the current thread is interrupted during the blocking send operation.
  */
 public fun <E> SendChannel<E>.trySendBlocking(element: E): ChannelResult<Unit> {
     /*
--- a/kotlinx-coroutines-core/concurrent/src/channels/Channels.kt
+++ b/kotlinx-coroutines-core/concurrent/src/channels/Channels.kt
@@ -25,7 +25,7 @@ import kotlin.jvm.*
  *
  * For this operation it is guaranteed that [failure][ChannelResult.failed] always contains an exception in it.
  *
- * @throws InterruptedException on JVM if the current thread is interrupted during the blocking send operation.
+ * Throws `InterruptedException` on JVM if the current thread is interrupted during the blocking send operation.
  */
 public fun <E> SendChannel<E>.trySendBlocking(element: E): ChannelResult<Unit> {
     /*
--- a/kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
+++ b/kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
@@ -19,7 +19,7 @@ private typealias Node = LockFreeLinkedListNode
  * Important notes:
  * - There are no operations to add items to left side of the list, only to the end (right side), because we cannot
  *   efficiently linearize them with atomic multi-step head-removal operations. In short,
- *   support for `TODO no such function` operation precludes ability to add items at the beginning.
+ *   support for the [remove] operation precludes ability to add items at the beginning.
  * - Previous pointers are not marked for removal. We don't support linearizable backwards traversal.
  * - Remove-helping logic is simplified and consolidated in [correctPrev] method.
  *
--- a/kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
+++ b/kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
@@ -19,7 +19,7 @@ private typealias Node = LockFreeLinkedListNode
  * Important notes:
  * - There are no operations to add items to left side of the list, only to the end (right side), because we cannot
  *   efficiently linearize them with atomic multi-step head-removal operations. In short,
- *   support for [describeRemoveFirst] operation precludes ability to add items at the beginning.
+ *   support for `TODO no such function` operation precludes ability to add items at the beginning.
  * - Previous pointers are not marked for removal. We don't support linearizable backwards traversal.
  * - Remove-helping logic is simplified and consolidated in [correctPrev] method.
  *
--- a/kotlinx-coroutines-core/jvm/src/debug/internal/DebugCoroutineInfoImpl.kt
+++ b/kotlinx-coroutines-core/jvm/src/debug/internal/DebugCoroutineInfoImpl.kt
@@ -24,7 +24,7 @@ internal class DebugCoroutineInfoImpl internal constructor(
     @JvmField public val sequenceNumber: Long
 ) {
     /**
-     * We cannot keep a strong reference to the context, because with the [Job] in the context it will indirectly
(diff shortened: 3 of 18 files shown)


kotlinx-coroutines-core/concurrent/src/channels/Channels.kt
```diff
@@ -25,7 +25,7 @@ import kotlin.jvm.*
  *
  * For this operation it is guaranteed that [failure][ChannelResult.failed] always contains an exception in it.
  *
- * @throws `InterruptedException` on JVM if the current thread is interrupted during the blocking send operation.
+ * @throws InterruptedException on JVM if the current thread is interrupted during the blocking send operation.
```
dkhalanskyjb: Let's remove `@throws` and replace it with `Throws `? Also, keeping the backticks, since that's an identifier and not a plain text.

kotlinx-coroutines-core/concurrent/src/internal/LockFreeLinkedList.kt
```diff
@@ -19,7 +19,7 @@ private typealias Node = LockFreeLinkedListNode
  * Important notes:
  * - There are no operations to add items to left side of the list, only to the end (right side), because we cannot
  *   efficiently linearize them with atomic multi-step head-removal operations. In short,
- *   support for [describeRemoveFirst] operation precludes ability to add items at the beginning.
+ *   support for `TODO no such function` operation precludes ability to add items at the beginning.
```
dkhalanskyjb:
```suggestion
 *   support for the [remove] operation precludes ability to add items at the beginning.
```

kotlinx-coroutines-core/jvm/src/debug/internal/DebugCoroutineInfoImpl.kt
```diff
@@ -24,7 +24,7 @@ internal class DebugCoroutineInfoImpl internal constructor(
     @JvmField public val sequenceNumber: Long
 ) {
     /**
-     * We cannot keep a strong reference to the context, because with the [Job] in the context it will indirectly
+     * We cannot keep a strong reference to the context, because with the [kotlinx.coroutines.Job] in the context it will indirectly
```
dkhalanskyjb: The line is a bit longer than we want—could you `import kotlinx.coroutines.Job` instead?

dkhalanskyjb (approved):
Thanks! Modulo some small nits, it's good to go.

As for module-level documentation, yes, I agree that ideally, it should be separate for Dokka.

whyoleg:
> As for module-level documentation, yes, I agree that ideally, it should be separate for Dokka.

Do you have any ideas (or thoughts) about which information you would like to present there? :)

I mean, there are overall three places currently where we could put some non-declarations-kdoc documentation:

1. `README.md` in GitHub repository
2. module and package descriptions in Dokka's generated API reference
3. "Coroutines guide" and other docs in `docs` folder, which is published to kotlinlang.org

I don't think we should just copy docs/md files from one place to another; we need some structure here, right? Maybe @koshachy also has some suggestions, as I know, that he knows a thing or two about "API reference" vs "concept docs". :)

dkhalanskyjb:
> Do you have any ideas (or thoughts) about which information you would like to present there? :)

It should cover more or less the same topics as the current `README.md` files do. There are many clear improvements to be made to those, but the general idea feels suitable for an API reference. What a module does, which packages it contains, which API entry points are worth learning about first.

On GitHub, these aspects are important, too, but also, an architectural overview or some specific implementation nuances can also be covered. For reference documentation, they don't make much sense, but if someone's actively browsing our repository, they must already be interested in how all of this works.

What do you think about merging this PR as is first, without dealing with module-level documentation at this point? We're rewriting some KDocs right now, so if we don't merge this PR soon, it may need to go through some annoying rebasing.

whyoleg:
> What do you think about merging this PR as is first, without dealing with module-level documentation at this point?

Agree! The PR is good to go from my side in its current state. Meaning, I'm ready to click `merge`. :)

I will create a separate PR for a module-level documentation split, in a naive way, based on your description, in a `failOnWarning` compatible way, and we can discuss it there.

PR #4645, opened 2026-03-25
KT-84955: mark apple x64 tagets as deprecated error

mMaxy:


gradle.properties
```diff
@@ -52,3 +52,5 @@ kotlinx.atomicfu.enableJsIrTransformation=false
 kotlinx.atomicfu.enableNativeIrTransformation=true
 
 org.gradle.caching=true
+
+kotlin.internal.suppressGradlePluginErrors=DeprecatedKotlinNativeTargetsDiagnostic
```
fzhinkin, line 57: @mMaxy, `integration-testing` has its own `gradle.properties` and this diagnostic needs to be disabled there too.

buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
```diff
@@ -29,27 +29,32 @@ kotlin {
...
         mingwX64()
         watchosDeviceArm64()
+
+        // Deprecated
+        @Suppress("DEPRECATION", "DEPRECATION_ERROR")
+        macosX64()
+        @Suppress("DEPRECATION", "DEPRECATION_ERROR")
+        tvosX64()
+        @Suppress("DEPRECATION", "DEPRECATION_ERROR")
+        watchosX64()
```
dkhalanskyjb: Could you please also drop a code comment that will help us decide when to drop the support for these targets entirely? Maybe the KT issue number, or a specific compiler version, etc.

fzhinkin (approved):

PR #4647, opened 2026-03-27
Prevent R8 from removing the job GC anchor in ReadonlySharedFlow/ReadonlyStateFlow

### LINKED ISSUE #4646: R8 removes the `job` GC-anchor field from `ReadonlySharedFlow`/`ReadonlyStateFlow`
The fix for [#2557](https://github.com/Kotlin/kotlinx.coroutines/issues/2557) ([fbac730](https://github.com/pablobaxter/kotlinx.coroutines/commit/fbac73096a61c8b45c995a2320f139a87e5c9333)) added a `private val job: Job?` field to `ReadonlySharedFlow` and `ReadonlyStateFlow` to prevent the sharing coroutine from being garbage collected. The field is annotated `@Suppress("unused")`, which silences the Kotlin compiler but has no effect on bytecode-level optimizers.

R8's full-mode optimizer removes fields that are written but never read. Since `job` is never read, R8 strips the field and its constructor assignment, making the sharing coroutine eligible for GC again.

**This will affect all Android projects once AGP 9.0 is adopted**, because it [drops support for `proguard-android.txt`](https://android-developers.googleblog.com/2025/11/use-r8-to-shrink-optimize-and-fast.html), which had `-dontoptimize`.

## Reproduction

Tested with:
- kotlinx-coroutines 1.10.2
- R8 8.13.19
- Kotlin 2.3.20
- JDK 17

The reproduction code is adapted from the original issue. A `callbackFlow` with a `WeakReference`-based callback (mimicking Android APIs like `SharedPreferences` or `MediaPlayer`) is shared via `stateIn`. Without the `job` GC anchor, the sharing coroutine is collected and emissions stop.

Here are the files to reproduce using a Gradle JVM project that runs R8 standalone, without requiring AGP:

**`settings.gradle.kts`**

```kotlin
rootProject.name = "coroutines-r8-repro"
```

**`build.gradle.kts`**

```kotlin
import java.util.jar.JarFile
import java.net.URI

plugins {
    kotlin("jvm") version "2.3.20"
    application
}

repositories {
    mavenCentral()
}

dependencies {
    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2")
}

application {
    mainClass.set("MainKt")
}

val r8Version = "8.13.19"
val r8Jar = layout.buildDirectory.file("r8/r8.jar")

tasks.register("downloadR8") {
    val dest = r8Jar.get().asFile
    outputs.file(dest)
    doLast {
        dest.parentFile.mkdirs()
        URI("https://storage.googleapis.com/r8-releases/raw/$r8Version/r8.jar").toURL()
            .openStream().use { it.copyTo(dest.outputStream()) }
    }
}

tasks.register<Jar>("fatJar") {
    archiveClassifier.set("all")
    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
    manifest { attributes["Main-Class"] = "MainKt" }
    from(configurations.runtimeClasspath.get().map { if (it.isDirectory) it else zipTree(it) })
    with(tasks.jar.get())
}

tasks.register<JavaExec>("r8") {
    dependsOn("fatJar", "downloadR8")

    val fatJar = tasks.named<Jar>("fatJar").get().archiveFile.get().asFile
    val outputDir = layout.buildDirectory.dir("r8out").get().asFile
    val rulesFile = layout.projectDirectory.file("r8-rules.pro").asFile
    val extractedRules = layout.buildDirectory.file("coroutines-r8.pro").get().asFile

    doFirst {
        val jarFile = JarFile(fatJar)
        val entry = jarFile.getJarEntry("META-INF/com.android.tools/r8/coroutines.pro")
        extractedRules.writeText(jarFile.getInputStream(entry).bufferedReader().readText())
        jarFile.close()
        outputDir.deleteRecursively()
        outputDir.mkdirs()
    }

    classpath = files(r8Jar)
    mainClass.set("com.android.tools.r8.R8")
    args = listOf(
        "--classfile",
        "--release",
        "--no-minification",
        "--pg-conf", rulesFile.absolutePath,
        "--pg-conf", extractedRules.absolutePath,
        "--lib", System.getProperty("java.home"),
        "--output", outputDir.absolutePath,
        fatJar.absolutePath
    )
}

tasks.register<Jar>("r8Jar") {
    dependsOn("r8")
    archiveClassifier.set("r8")
    manifest { attributes["Main-Class"] = "MainKt" }
    from(layout.buildDirectory.dir("r8out"))
}

tasks.register<JavaExec>("runBaseline") {
    dependsOn("fatJar")
    classpath = files(tasks.named<Jar>("fatJar").get().archiveFile)
}

tasks.register<JavaExec>("runOptimized") {
    dependsOn("r8Jar")
    classpath = files(tasks.named<Jar>("r8Jar").get().archiveFile)
}
```

**`r8-rules.pro`**: bundled [coroutines R8 rules](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/com.android.tools/r8/coroutines.pro) are extracted from the jar and provided to the R8 pass

```
-dontwarn **

# App entry point
-keep class MainKt { *; }
```

**`src/main/kotlin/Main.kt`**: similar to original snippet

```kotlin
@file:OptIn(ExperimentalCoroutinesApi::class)

import kotlinx.coroutines.*
import java.lang.ref.WeakReference
import kotlin.coroutines.EmptyCoroutineContext
import kotlin.time.Duration
import kotlin.time.Duration.Companion.milliseconds

fun main(): Unit = runBlocking {
    val jobToCancel = launch {
        someFlow.collect()
    }
    delay(100.milliseconds)
    repeat(5) {
        System.gc()
        System.gc()
        delay(1.seconds)
    }
    jobToCancel.cancel()
}

val someFlow = flow<Int> {
    iterationsFlow(1.seconds).collect {
        println("Got $it")
        currentCoroutineContext().ensureActive()
    }
}.stateIn(
    scope = CoroutineScope(EmptyCoroutineContext),
    started = SharingStarted.Eagerly,
    initialValue = 0
)

fun iterationsFlow(interval: Duration): Flow<Int> = callbackFlow {
    val listener = Listener {
        trySend(it)
    }
    val cb = CallbackRepeatCaller(interval, listener)
    awaitClose {
        cb.forget()
        listener.call(-1)
    }
}

fun interface Listener {
    fun call(iterations: Int)
}

class CallbackRepeatCaller(interval: Duration, listener: Listener) {
    private val ref = WeakReference(listener)

    private val caller = CoroutineScope(EmptyCoroutineContext).launch {
        var iterations = 0
        while (true) {
            ensureActive()
            ref.get()?.call(++iterations)
            delay(interval)
        }
    }

    fun forget() {
        caller.cancel()
        ref.clear()
    }
}
```

**1. Baseline (no R8): sharing coroutine survives GC:**

```bash
./gradlew runBaseline
```
```
Got 1
Got 2
Got 3
Got 4
Got 5
Got 6
```

**2. After R8: sharing coroutine is GC'd:**

```bash
./gradlew runOptimized
```
```
Got 1
```

R8 strips the `job` field because none of the bundled rules protect write-only private fields. You can verify with:

```bash
javap -private -cp build/libs/coroutines-r8-repro-r8.jar kotlinx.coroutines.flow.ReadonlyStateFlow
```

Before R8:
```
private final kotlinx.coroutines.flow.StateFlow<T> $$delegate_0;
private final kotlinx.coroutines.Job job;                          // GC anchor present
```

After R8:
```
public final kotlinx.coroutines.flow.StateFlow $$delegate_0;      // job is gone
```

ProGuard 7.9.0 does not exhibit this behavior with `-optimizations field/removal/writeonly` explicitly enabled, it preserves the field. It is unclear to me whether this is an intentional optimization difference in R8 or a bug.

Regardless, the `job` field should be protected with a `-keep` rule to prevent any optimizer from stripping it.

solevic:
Fixes #4646 

R8 full mode strips fields that are written but never read. The `job` field in `ReadonlySharedFlow` and `ReadonlyStateFlow` is a GC anchor keeping the sharing coroutine alive (added in the fix for #2557), but since it is only assigned and never read, R8 removes it, making the sharing coroutine eligible for collection.

This adds `-keepclassmembers` rules targeting the specific `job` field to all three bundled `.pro` files.

Verified with a standalone R8 Gradle JVM reproduction project available in the original issue.

### DIFF
--- a/kotlinx-coroutines-core/jvm/resources/META-INF/com.android.tools/proguard/coroutines.pro
+++ b/kotlinx-coroutines-core/jvm/resources/META-INF/com.android.tools/proguard/coroutines.pro
@@ -29,3 +29,11 @@
 
 # An annotation used for build tooling, won't be directly accessed.
 -dontwarn org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+
+# Keep GC anchor fields that prevent the sharing coroutine from being collected.
+-keepclassmembers class kotlinx.coroutines.flow.ReadonlySharedFlow {
+    kotlinx.coroutines.Job job;
+}
+-keepclassmembers class kotlinx.coroutines.flow.ReadonlyStateFlow {
+    kotlinx.coroutines.Job job;
+}
--- a/integration-testing/build.gradle.kts
+++ b/integration-testing/build.gradle.kts
@@ -69,6 +69,7 @@ allprojects {
         if (!kotlinRepoUrl.isNullOrBlank()) {
             maven(kotlinRepoUrl)
         }
+        google()
         mavenCentral()
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev")
         maven("https://redirector.kotlinlang.org/maven/dev")
@@ -244,7 +245,8 @@ tasks {
             ":jpmsTest:check",
             "smokeTest:build",
             "java8Test:check",
-            "safeDebugAgentTest:attachAgentWithoutKotlinStdlib"
+            "safeDebugAgentTest:attachAgentWithoutKotlinStdlib",
+            "r8Test:testGcAnchor"
         )
     }
 
--- a/integration-testing/r8Test/build.gradle.kts
+++ b/integration-testing/r8Test/build.gradle.kts
@@ -13,6 +13,8 @@ dependencies {
     "r8"("com.android.tools.build:builder:8.1.0")
 }
 
+// The R8 logic below was taken from `ui/kotlinx-coroutines-android/build.gradle.kts`
+
 val r8OutputDir = layout.buildDirectory.dir("r8out")
 
 val runR8 by tasks.registering(RunR8::class) {
--- a/integration-testing/r8Test/build.gradle.kts
+++ b/integration-testing/r8Test/build.gradle.kts
@@ -1,0 +1,71 @@
+plugins {
+    kotlin("jvm")
+}
+
+val coroutinesVersion = property("coroutines_version").toString()
+
+configurations {
+    create("r8")
+}
+
+dependencies {
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
+    "r8"("com.android.tools.build:builder:8.1.0")
+}
+
+val r8OutputDir = layout.buildDirectory.dir("r8out")
+
+val runR8 by tasks.registering(RunR8::class) {
+    outputDir = r8OutputDir.get().asFile
+    inputConfig = file("r8-rules.pro")
+
+    dependsOn("jar")
+}
+
+tasks.register("testGcAnchor") {
+    dependsOn(runR8)
+
+    doLast {
+        project.javaexec {
+            mainClass.set("GcAnchorKt")
+            classpath = files(r8OutputDir)
+        }
+    }
+}
+
+open class RunR8 : JavaExec() {
+
+    @OutputDirectory
+    lateinit var outputDir: File
+
+    @InputFile
+    lateinit var inputConfig: File
+
+    @InputFiles
+    val jarFile: File = project.tasks.named<Zip>("jar").get().archiveFile.get().asFile
+
+    init {
+        classpath = project.configurations["r8"]
+        mainClass = "com.android.tools.r8.R8"
+    }
+
+    override fun exec() {
+        val arguments = mutableListOf(
+            "--classfile",
+            "--release",
+            "--no-minification", // avoid class name collisions on case-insensitive filesystems
+            "--pg-conf", inputConfig.absolutePath,
+            "--lib", System.getProperty("java.home"),
+            "--output", outputDir.absolutePath,
+        )
+        arguments.addAll(project.configurations["runtimeClasspath"].files.map { it.absolutePath })
+        arguments.add(jarFile.absolutePath)
+
+        args = arguments
+
+        project.delete(outputDir)
+        outputDir.mkdirs()
+
+        super.exec()
+    }
+}
--- a/integration-testing/r8Test/r8-rules.pro
+++ b/integration-testing/r8Test/r8-rules.pro
@@ -1,0 +1,2 @@
+# App entry point
+-keep class GcAnchorKt { *; }
--- a/integration-testing/r8Test/src/main/kotlin/GcAnchor.kt
+++ b/integration-testing/r8Test/src/main/kotlin/GcAnchor.kt
@@ -1,0 +1,44 @@
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.flow.*
+import java.lang.ref.WeakReference
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+/**
+ * Verifies that R8 doesn't strip the GC anchor (`job` field) from ReadonlyStateFlow.
+ *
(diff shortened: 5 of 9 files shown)


kotlinx-coroutines-core/jvm/resources/META-INF/com.android.tools/proguard/coroutines.pro
```diff
@@ -29,3 +29,11 @@
 
 # An annotation used for build tooling, won't be directly accessed.
 -dontwarn org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+
+# Keep GC anchor fields that prevent the sharing coroutine from being collected.
+-keepclassmembers class kotlinx.coroutines.flow.ReadonlySharedFlow {
```
dkhalanskyjb, line 34: Since these classes are `private`, please add a comment to the `ReadonlySharedFlow`/`ReadonlyStateFlow` implementations, explaining that they should not be renamed. Otherwise, we're one small refactoring away from a regression.

dkhalanskyjb (approved):
This change does make the reproducer emit several values instead of just one, so the change clearly works.

Now, changes like these require an integration test. I realize our configuration is complex, so it's perfectly fine to merge the PR as is, and we'll write take it from here later on our own.

If you'd like to work on a test, here are some pointers.
* The `integration-testing` directory is where such tests go.
* `integration-testing/safeDebugAgentTest/build.gradle.kts` is an example of us checking the results of a program's execution to fail a test (though in that case, that program is in Java). I'd expect something similar to be required in this case.
* `ui/kotlinx-coroutines-android/build.gradle.kts` already includes some logic for running R8. Would be nice to keep this consistent. I don't think this code can realistically be reused, though, as integration tests are an entirely separate Gradle project (to ensure that integration tests don't get any leaks from the main project through Gradle).

Should we merge the PR as is, or would you like to attempt adding the test?

kotlinx-coroutines-core/jvm/resources/META-INF/com.android.tools/proguard/coroutines.pro
solevic: Thanks for the review.

I added comments to both `ReadonlySharedFlow` and `ReadonlyStateFlow`.

I'm not familiar with testing byte-code optimization rules, but I can take a look.

Would you prefer something like `R8ServiceLoaderOptimizationTest` that runs actual JUnit tests by performing static analysis on the R8 output OR an example JVM application with the code provided in the reproduction steps for which the stdout would be checked during the gradle task execution?

dkhalanskyjb: The idea is to check this end-to-end—in this case, the behavior of an optimized JVM application. Your reproducer makes a great starting point for that. Instead of checking the stdout, I think it would be cleaner to have the application itself fail if it detects the problem, and then the Gradle task execution could simply detect the failure.

solevic: I pushed an attempt at an integration test for the rule on `ReadonlyStateFlow` with your suggestions.

This creates a module in `integration-testing` with a harness for future R8 integration tests. If needed for completeness, I can amend a second test for `ReadonlySharedFlow`.

The reproducer was simplified to minimal viable reproducing code.

As you mentioned code from `ui/kotlinx-coroutines-android/build.gradle.kts` might not be reusable due to Gradle project hierarchy, I did not attempt any factorization.

Feel free to adjust the emission and GC timings if the tests are flaky during CI.

dkhalanskyjb:
Thank you!

PR #4649, opened 2026-03-30
Fill Implementation-* attributes in Jar's manifests

### LINKED ISSUE #4599: Fill Implementation-Title/Version/Vendor attributes in JAR's Manifest
Having this information will help with improving DX in IDE.

See https://youtrack.jetbrains.com/issue/KT-83393

fzhinkin:
Closes #4599

### DIFF
--- a/buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
@@ -1,6 +1,8 @@
 import org.gradle.api.tasks.testing.logging.*
+import org.gradle.api.tasks.bundling.Jar
 import org.jetbrains.kotlin.gradle.dsl.*
 import org.jetbrains.kotlin.gradle.dsl.abi.ExperimentalAbiValidation
+import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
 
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/kotlin-multiplatform-conventions.gradle.kts
```diff
@@ -146,3 +148,10 @@ tasks.named("jvmTest", Test::class) {
 tasks.check {
    dependsOn(tasks.checkLegacyAbi)
 }
+
+kotlin.targets.withType<KotlinJvmTarget>().configureEach {
+    // Fill attributes for the JVM implementation Jar only
+    tasks.named<Jar>(artifactsTaskName) {
+        fillManifestImplementationAttributes(project)
+    }
+}
```
dkhalanskyjb, line 157: Note to self: it's unpleasant that we have two ways now of searching for the right JVM jar: looking up the `jvmJar` name for multiplatform projects, and accessing `KotlinJvmTarget.artifactsTaskName` to learn that constant. I don't know how to neatly unify that logic, though. The best solution to me is for the Kotlin Gradle plugin to also expose JVM-only Kotlin projects as MPP projects with only the JVM target. Then, `kotlin-jvm-conventions` could be removed altogether.

PR #4651, opened 2026-03-31
Fix module name after KT-69701

Tapchicoma:


Tapchicoma:
I've updated the Gradle version as we are deprecating support for Gradle versions <8.14.4 in the 2.4.0 release.

buildSrc/src/main/kotlin/old-module-name-behavior.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+    }
+}
+
+plugins.withId("org.jetbrains.kotlin.multiplatform") {
+    extensions.configure<KotlinMultiplatformExtension>() {
+        targets.withType<KotlinJvmTarget>() {
+            compilerOptions.moduleName.value(project.name)
+        }
+    }
+}
```
fzhinkin:
```suggestion
// This convention plugin restores behavior that was before changes in https://youtrack.jetbrains.com/issue/KT-69701

plugins.withId("org.jetbrains.kotlin.jvm") {
    extensions.configure<KotlinJvmExtension> {
        compilerOptions.moduleName.value(project.name)
    }
}

plugins.withId("org.jetbrains.kotlin.multiplatform") {
    extensions.configure<KotlinMultiplatformExtension> {
        targets.withType<KotlinJvmTarget> {
            compilerOptions.moduleName.value(project.name)
        }
    }
}
```

settings.gradle.kts
```diff
@@ -13,6 +13,10 @@ pluginManagement {
 
 rootProject.name = "kotlinx.coroutines"
 
+gradle.beforeProject {
+    plugins.apply("old-module-name-behavior")
```
fzhinkin: Why it has to be applied like that instead of applying the plugin in `kotlin-jvm-conventions` and `kotlin-multiplatform-conventions`?

Tapchicoma: This approach is more [isolated projects Gradle feature](https://docs.gradle.org/current/userguide/isolated_projects.html) compatible. 

Your current approach `kotlin-jvm-conventions` uses deprecated `configure` delegate (should be removed in 9.0.0)  and iterates over sub-projects, which violates project isolation.

buildSrc/src/main/kotlin/old-module-name-behavior.gradle.kts
Tapchicoma: `withType<> { .. }` under-the-hood uses `withType<>().all { }` which violates Gradle laziness. Only applied first part of suggestions manually

buildSrc/src/main/kotlin/old-module-name-behavior.gradle.kts
```diff
@@ -0,0 +1,21 @@
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmExtension
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
+
+// This convention plugin restores behavior that was before changes in https://youtrack.jetbrains.com/issue/KT-69701
```
dkhalanskyjb: Could you please explain this change at the *highest* level? I somewhat understand the details of what the PR does, but not the overall motivation.

So, there is some incorrect default on older Kotlin versions, one that's going to be changed to a better one in future versions. Sounds good!

Why does `kotlinx.coroutines` need to stick to the older default, though? If we don't, are we introducing some breaking change for the client code? If so, does *every* multiplatform library need to include such code? That sounds like a large-scale outage waiting to happen!

settings.gradle.kts
fzhinkin: Somewhat unrelated to this particular change, but does it mean we should rewrite our build scripts to something like that?

Tapchicoma: If you have convention plugins that apply to all sub-projects, it's better to use this approach. Though you don't need to do it explicitly right now - we are still working on this Gradle feature support in JS/WASM. Once it lands in KGP - we will come to `kotlinx.` repositories and will try to make them feature compatible.

buildSrc/src/main/kotlin/old-module-name-behavior.gradle.kts
Tapchicoma: This fix is caused by [KT-69701](https://youtrack.jetbrains.com/issue/KT-69701) Gradle: module name is passed inconsistently to different types of compilations. In this issue we've changed default value for `moduleName` in JVM and metadata compilations from `project.name` to `project.group:project.name` to align it with other platforms. 

Using just `project.name` leads to a greater chance of clashes of module names from different sub-projects/libraries.

Why changing the new default value could be needed in projects - there is one particular case that breaks with this change - unconventionally (e.g. not using `-Xfriend-path`) accessing `internal` symbols from a dependency. For example, from Java it is possible to access such symbols with `$module_name` suffix, where `module_name` was changed. 

In the kotlinx.coroutines case the test `CoreAgentTest.testAgentDumpsCoroutines` started to fail (other tests could be fixed by updating test data). After reading the code, I saw a comment that the IDEA debugger relies on internal coroutines symbols to work (related to [IDEA-335375](https://youtrack.jetbrains.com/issue/IDEA-335375) Change coroutine debugger implementation to use a new package), so I decided to keep old behavior to be on the safe side.

dkhalanskyjb: I see, that explains a lot, thank you!

Then, the fix can be applied to `kotlinx.coroutines.core` specifically, without affecting the other subprojects. Only `kotlinx.coroutines.core` needs to expose its internals to reflective access.

Please also add a comment to the build script, explaining the purpose behind changing the defaults, so that we don't have to dig into git history to figure this out later.

Tapchicoma: Done

Tapchicoma:
Need to update test logic to use different test data based on used Kotlin version, build with Kotlin/dev is green

buildSrc/src/main/kotlin/old-module-name-behavior.gradle.kts
dkhalanskyjb: > Then, the fix can be applied to `kotlinx.coroutines.core` specifically, without affecting the other subprojects. Only `kotlinx.coroutines.core` needs to expose its internals to reflective access.

What I meant was, why not move this configuration to `kotlinx-coroutines-core/build.gradle.kts`, which seems to be the only place where module names are important?

Tapchicoma: Done

PR #4655, opened 2026-04-13
Update composing-suspending-functions.md

Karpikova:
Thanks for your great work!
Sincerely,
Maria Pomazkina
LinkedIn: https://www.linkedin.com/in/maria-pomazkina-7972a2a3/


dkhalanskyjb:
The whole section isn't great, but changing this sentence locally doesn't fix any problems, in my opinion. What the sentence is trying to say is that trying to make your concurrent program run sequentially is not a valid reason to use `LAZY`. Also, the goal of `LAZY` start isn't to enable parallel execution—it's to postpone running some computations until their results are necessary, in order to avoid overwhelming the system with a lot of work as soon as some component initializes. In fact, the expected effect is that `LAZY` will *lower* the parallelism.

PR #4656, opened 2026-04-13
Update composing-suspending-functions.md

Karpikova:
The deleted annotation is merely a technical requirement. It does not help to explain the conceptual danger of GlobalScope, so mentioning it distracts from the main point.

Sincerely,  
Maria Pomazkina  
LinkedIn: https://www.linkedin.com/in/maria-pomazkina-7972a2a3/


dkhalanskyjb:
Without mentioning `DelicateCoroutinesApi` in the text, we risk making it unclear how exactly a user should opt into `GlobalScope`. Maybe a compiler flag is needed? Or calling some API at the start of the program? Or possibly, even by just making a mental note?

Here, the sentence makes it clear what you must do and why, and the example that follows explains how exactly to do it.

PR #4657, opened 2026-04-13
Prepare for `kotlin.coroutines.StackTraceRecoverable`

dkhalanskyjb:
We would like to remove `kotlinx.coroutines.CopyableThrowable` in favor of an interface with the same functionality in the standard library. This way, libraries that expose exception classes will not have to depend on `kotlinx.coroutines` to provide stacktrace recovery support.

The likely migration procedure is:
- `kotlinx.coroutines` introduces a hidden code path that reflectively discovers the stdlib method. This is what this commit does. A `kotlinx.coroutines` release is published.
- An stdlib release includes `StackTraceRecoverable`. Any exception classes using `StackTraceRecoverable` will immediately receive stacktrace recovery support.
- A `kotlinx.coroutines` release that switches to the Kotlin version with `StackTraceRecoverable` removes the hidden code path and replaces it by an `is`-check. The same release deprecates `CopyableThrowable` with a warning.
- After the stdlib stabilizes `StackTraceRecoverable`, `CopyableThrowable` is deprecated with an error and eventually removed.

It is slightly concerning that as this PR is written, `CopyableThrowable` takes priority over `StackTraceRecoverable` if both interfaces are implemented. I expect this to be okay, because the semantics of the two interfaces are exactly the same, and their implementations should be equivalent as well. Making priorities the opposite would introduce a slight performance degradation for `CopyableThrowable` implementations and complicate the code.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/ExceptionsConstructor.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/ExceptionsConstructor.kt
@@ -28,6 +28,28 @@ internal fun <E : Throwable> tryCopyException(exception: E): E? {
 
 private fun <E : Throwable> createConstructor(clz: Class<E>): Ctor {
     val nullResult: Ctor = { null } // Pre-cache class
+    /* Prepare for the introduction of `StackTraceRecoverable` in the standard library.
+     * The exact interface to be added is still under discussion,
+     * but the name `StackTraceRecoverable#copyForStackTraceRecovery`,
+     * returning either `(E: Throwable)?` or `Throwable?` is already agreed on.
+     *
+     * To anyone reading this: don't try adding `copyForStackTraceRecovery` in your own code.
+     * This will work for now, but break later.
+     * Once `StackTraceRecoverable` enters the standard library,
+     * `kotlinx.coroutines` will move to using an `is StackTraceRecoverable`-check.
+     * This is the reason this behavior is left undocumented. */
+    try {
+        val copyForStackTraceRecovery = clz.getMethod("copyForStackTraceRecovery")
+        return { e ->
+            runCatching {
+                (copyForStackTraceRecovery.invoke(e) as Throwable?)
+            }.getOrNull()
+        }
+    } catch (_: NoSuchMethodException) {
+        // Expected: there simply was no implementation of `StackTraceRecoverable`
+    } catch (_: SecurityException) {
+        // Restrictive SecurityManager
+    }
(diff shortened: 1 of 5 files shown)


kotlinx-coroutines-core/jvm/src/internal/ExceptionsConstructor.kt
```diff
@@ -28,6 +28,28 @@ internal fun <E : Throwable> tryCopyException(exception: E): E? {
...
+     *
+     * To anyone reading this: don't try adding `copyForStackTraceRecovery` in your own code.
+     * This will work for now, but break later.
+     * Once `StackTraceRecoverable` enters the standard library,
+     * `kotlinx.coroutines` will move to using an `is StackTraceRecoverable`-check.
+     * This is the reason this behavior is left undocumented. */
+    try {
+        val copyForStackTraceRecovery = clz.getMethod("copyForStackTraceRecovery")
+        return { e ->
+            runCatching {
```
sandwwraith, line 44: You could also move `getMethod` under `runCatching` — it seems there are no other exceptions thrown from `getMethod()`. But I guess it's not necessary.

sandwwraith (approved):

sandwwraith:
> CopyableThrowable takes priority over StackTraceRecoverable if both interfaces are implemented.

I think it's totally fine

kotlinx-coroutines-core/jvm/src/internal/ExceptionsConstructor.kt
dkhalanskyjb: I'm keeping `getMethod` outside so that it only gets called once per exception class. The `{ e -> ` lambda is stored in a map and will get invoked for every exception of this type.

kotlinx-coroutines-core/jvm/test/exceptions/StackTraceRecoveryStdlibInterfaceWithTypeBoundsTest.kt
```diff
@@ -0,0 +1,193 @@
...
+            val cause = e.cause
+            assertIs<Copyable>(cause)
+            assertEquals(239, cause.customData)
+        }
+    }
+
+    internal class WithDefault(message: String = "default") : Exception(message)
+
+    @Test
+    fun testStackTraceRecoveredWithCustomMessage() = runTest {
```
sandwwraith: Given that these tests are not related to `StackTraceRecoverableWithTypeBounds`, IMO we shouldn't duplicate them

murfel (approved):

Issue #4602, opened 2026-01-13
Missing main dispatcher hangs if `start = CoroutineStart.ATOMIC`


phiSgr:
I was looking at [this SO question](https://stackoverflow.com/questions/79839679/flowondispatchers-main-blocks-the-main-thread), the accepted answer felt off. Looking at the implementation, I believe the `ATOMIC` start in `ChannelFlow#produceImpl` causes the problem.

**Describe the bug**

Instead of throwing an `IllegalStateException` with the message "Module with the Main dispatcher is missing.", it just hangs.

**Provide a Reproducer**

https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4zLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJpbXBvcnQga290bGlueC5jb3JvdXRpbmVzLipcblxuZnVuIG1haW4oKSA9IHJ1bkJsb2NraW5nPFVuaXQ+IHtcbiAgICBsYXVuY2goRGlzcGF0Y2hlcnMuTWFpbiwgc3RhcnQgPSBDb3JvdXRpbmVTdGFydC5BVE9NSUMpIHtcbiAgICAgICAgcHJpbnRsbihcInN0dWZmXCIpXG4gICAgfVxufSJ9

No-99-Tongji:
I found that using the CoroutineStart.LAZY parameter also directly results in a deadlock, rather than throwing a java.lang.IllegalStateException.

vsartiushkin:
Thank you for the report and the reproducer! 
I have reproduced the issue for the `CoroutineStart.ATOMIC` when the `Dispatchers.Main` is missing for the latest versions Kotlin 2.3.20 and Library 1.10.2.
```
import kotlinx.coroutines.*

fun main() = runBlocking<Unit> {
    // Expected to throw IllegalStateException("Dispatchers.Main is missing..."), but instead it hangs.
    launch(Dispatchers.Main, start = CoroutineStart.ATOMIC) {
    }
}
```

vsartiushkin:
[...]

Issue #4325, opened 2025-01-10
Dispatchers.Main.isMissing throws when dispatcher is missing (since 1.10.0)


consp1racy:
**Describe the bug**

Expected behavior: `Dispatchers.Main.isMissing()` returns true in JVM tests where `Dispatchers.setMain()` wasn't called.

Actual behavior:
```
java.lang.IllegalStateException: Dispatchers.Main was accessed when the platform dispatcher was absent and the test dispatcher was unset. Please make sure that Dispatchers.setMain() is called before accessing Dispatchers.Main and that Dispatchers.Main is not accessed after Dispatchers.resetMain().
	at kotlinx.coroutines.test.internal.TestMainDispatcherJvmKt.reportMissingMainCoroutineDispatcher(TestMainDispatcherJvm.kt:45)
	at kotlinx.coroutines.test.internal.TestMainDispatcherJvmKt.access$reportMissingMainCoroutineDispatcher(TestMainDispatcherJvm.kt:1)
	at kotlinx.coroutines.test.internal.TestMainDispatcherFactory.createDispatcher$lambda$4(TestMainDispatcherJvm.kt:20)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:83)
	at kotlinx.coroutines.test.internal.TestMainDispatcher.getMainDispatcher(TestMainDispatcher.kt:18)
	at kotlinx.coroutines.test.internal.TestMainDispatcher.getDispatcher(TestMainDispatcher.kt:22)
	at kotlinx.coroutines.test.internal.TestMainDispatcher.getImmediate(TestMainDispatcher.kt:28)
	at kotlinx.coroutines.internal.MainDispatchersKt.isMissing(MainDispatchers.kt:62)
	...
```

Probably introduced [here](https://github.com/Kotlin/kotlinx.coroutines/commit/8f8305744e495eed1d371235e96cb5378307a202).

**Provide a Reproducer**

The following method didn't crash before when called in JVM tests where `Dispatchers.setMain()` wasn't called:

```
/**
 * @throws AssertionError when called from the main thread and assertions are enabled
 */
@WorkerThread
fun <T> runBlockingOrThrow(
    context: CoroutineContext = EmptyCoroutineContext,
    block: suspend CoroutineScope.() -> T,
): T {
    contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }

    val mainDispatcher = Dispatchers.Main.immediate
[...]

Issue #880, opened 2018-12-10
Fail fast if coroutine machinery failed to initialize


qwwdfsad:
```
runBlocking {
  withContext(Dispatchers.Main) { }
}
```

currently hangs if the main dispatcher is missing: dispatcher failure occurs **after** establishing a parent-child relationship, so `runBlocking` awaits for its child which never completes.

PR #4662, opened 2026-04-17
Make `Dispatchers.Main` access fail fast

dkhalanskyjb:
Before this change, calling `Dispatchers.Main` used to never fail. Instead, if the `Main` dispatcher was unavailable on the platform, trying to call `dispatch` would throw an exception.

This behavior is problematic. #4602 demonstrates a problem that would never have arisen if calling `Main` failed altogether. Another example is #880.
In #4325, the author attempts to recreate eager throwing. This makes total sense: failures *in* coroutine dispatchers are hard to meaningfully process (#4209), and surfacing errors to the client code is a more reliable way of dealing with them.

With this change, `Dispatchers.Main` will immediately throw an exception if the dispatcher failed to initialize.

Technically, this may be considered a breaking change, because such code is currently possible to use:
```kotlin
var dispatcher = Dispatchers.Main
if (!onAndroid) {
    dispatcher = newSingleThreadContextElement("Main thread")
}
```
With this commit, `Dispatchers.Main` will throw an exception.

Luckily, the behavior before the change contradicts the documentation at
<https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-main.html>:

> Access to this property may throw an `IllegalStateException`
> if no main dispatchers are present in the classpath.

This commit aligns the actual behavior with what's documented.

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
+++ b/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
@@ -2,7 +2,6 @@ package kotlinx.coroutines.internal
 
 import kotlinx.coroutines.*
 import java.util.*
-import kotlin.coroutines.*
 
 /**
  * Name of the boolean property that enables using of [FastServiceLoader].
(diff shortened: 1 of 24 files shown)


kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
```diff
@@ -15,115 +14,56 @@ internal object MainDispatcherLoader {
...
-public object MissingMainCoroutineDispatcherFactory : MainDispatcherFactory {
-    override val loadPriority: Int
-        get() = -1
-
-    override fun createDispatcher(allFactories: List<MainDispatcherFactory>): MainCoroutineDispatcher {
-        return MissingMainCoroutineDispatcher(null)
-    }
-}
+internal class MainDispatcherLoadResult(
+    val dispatcherOrNull: MainCoroutineDispatcher?,
```
qwwdfsad: [nit] `@JvmField` for `Dispatcher.Main` slightly friendlier to the machine

kotlinx-coroutines-test/common/src/internal/TestMainDispatcher.kt
```diff
@@ -15,11 +15,16 @@ internal class TestMainDispatcher(createInnerMain: () -> CoroutineDispatcher):
 {
```
qwwdfsad: Something is off w.r.t. formating, should be

```
internal class TestMainDispatcher(
    createInnerMain: () -> CoroutineDispatcher
) : MainCoroutineDispatcher(), Delay {
```

kotlinx-coroutines-core/jvm/test/FailFastOnStartTest.kt
```diff
@@ -107,4 +105,7 @@ class FailFastOnStartTest : TestBase() {
             finish(3)
         })
     }
+
+    private fun isRejectedExecutionException(e: Throwable): Boolean =
+        e is java.util.concurrent.RejectedExecutionException
```
qwwdfsad: Can be imported?

kotlinx-coroutines-test/common/src/internal/TestMainDispatcher.kt
```diff
@@ -15,11 +15,16 @@ internal class TestMainDispatcher(createInnerMain: () -> CoroutineDispatcher):
...
     internal constructor(delegate: CoroutineDispatcher): this({ delegate })
 
-    private val mainDispatcher by lazy(createInnerMain)
+    private val mainDispatcher: Result<CoroutineDispatcher> by lazy {
```
qwwdfsad, line 17: The reproducer from the original ticket will hang if the test dispatcher is present but unset.

I.e.
```
runBlocking<Unit> {
    launch(Dispatchers.Main, start = CoroutineStart.ATOMIC) {
    }
}
```

in `kotlinx-coroutines-test` module just hangs.

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
```diff
@@ -15,115 +14,56 @@ internal object MainDispatcherLoader {
...
+            val message = buildString {
+                append("Module with the Main dispatcher failed to initialize")
+                preferredFactory.hintOnError()?.let {
+                    append(". ")
+                    append(it)
+                }
+            }
+            return MainDispatcherLoadResult(null, message, cause)
         }
+        return MainDispatcherLoadResult(dispatcher, null, null)
```
qwwdfsad, line 61: (comment to MainCoroutineDispatcher.kt, which wasn't changed) `toStringInternalImpl` can now throw an exception if `Main` is absent.
It is possible either when `MainCoroutineDispatcher` is extended manually (it's public) or for things like `Handler.asCoroutinDispatcher` in tests.

I highly doubt this is a real issue, highlighting in case you've missed.

qwwdfsad (changes requested):

qwwdfsad:
While we are here, this is a breaking change with _some_ impact. Taking the next release will be more breaking than usual, I'd suggest having a dedicated tag for all the changes that should make it into the release, rn it's a bit tough to keep track of everything

kotlinx-coroutines-test/common/src/internal/TestMainDispatcher.kt
dkhalanskyjb: I believe that's unavoidable. It's valid to create a coroutine scope with `Dispatchers.Main` in the system under test as long as you're not starting any coroutines before `setMain` gets called. If we fix the scenario you're outlining, it's at the cost of a comfortable test structure in many scenarios.

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
dkhalanskyjb: I did miss this! It's a real problem: in environments with `ServiceLoader` disabled, you may want to still use a `Dispatchers.Main` manually, and a `toString()` shouldn't fail with a message that `Main` isn't installed.

kotlinx-coroutines-test/common/src/internal/TestMainDispatcher.kt
qwwdfsad: Fair. It's also only an issue with `ATOMIC`, so not a big deal

qwwdfsad: >If we fix the scenario you're outlining, it's at the cost of a comfortable test structure in many scenarios.

Shouldn't it then also be the case for `Dispatchers.Main.immediate` (that fails with "Dispatchers.Main was accessed when the platform dispatcher was absent and the test dispatcher was unset" in the same scenario)

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
qwwdfsad: The fix works in situations where `immediate` does not throw (or throws `UnsupportedOperation`).

E.g. 
```
object MyDispatcher : MainCoroutineDispatcher() {
    override val immediate: MainCoroutineDispatcher get() = this
        block.run()
    }
}

@Test
fun foo() {
    runBlocking<Unit> {
        println(MyDispatcher)
    }
}
```

will fail when ran in an environment where `kotlinx-coroutines-test` is present and will succeed when it's not

kotlinx-coroutines-test/common/src/internal/TestMainDispatcher.kt
dkhalanskyjb: Yes, I agree, but this behavior is unchanged from `master`, which also throws the same exception, so we're being conservative by preserving it here. I'd rather not touch `immediate` on test dispatchers without any demand for that, since it needs to be reworked anyway.

qwwdfsad: > I'd rather not touch ..

Sure.

W.r.t. the issue being unavoidable, filed https://github.com/Kotlin/kotlinx.coroutines/issues/4698 -- a nice onboarding task for @murfel to figure out if the fix is correct

qwwdfsad (changes requested):
Almost there!

kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt
dkhalanskyjb: That's a behavior that's also in `master`, but sure, it's going to be more consistent to ignore all exceptions from `.immediate` in `toString` now. Changed.

qwwdfsad (approved):
thanks!

PR #4663, opened 2026-04-20
Revert "Ensure that inline functions only access atomics from the same class (#4041)"

murfel:
#4041 was created as a temporary fix to mitigate [KT-65554](https://youtrack.jetbrains.com/issue/KT-65554) and [KT-78508](https://youtrack.jetbrains.com/issue/KT-78508). Both are now obsolete.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
@@ -238,7 +238,7 @@ internal open class BufferedChannel<E>(
     /**
      * Abstract send implementation.
      */
-    private inline fun <R> sendImpl(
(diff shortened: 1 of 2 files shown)


dkhalanskyjb:
Good call. `rc01` went fairly smoothly, so we can try including removals of various old workarounds in `rc02`. It will be easier for us to deal with actual breaking changes in releases that are more bold than this one if we get the workarounds out of the way now.

dkhalanskyjb (approved):
The change should be fine:
1) We haven't done anything significant with these files since then.
2) We're on Kotlin 2.2.20, which was published in September, after those issues were marked as obsolete.

I'd wait for the CI to pass just in case, and maybe Filipp has some extra insights regarding this, but in my view, it's good to go.

PR #4664, opened 2026-04-20
Fix a deprecated select's replaceWith

murfel:
[KTIJ-37637](https://youtrack.jetbrains.com/issue/KTIJ-37637) ReplaceWith. Incorrect code replacement for deprecated onTimeout in kotlinx.coroutines

[KT-84371](https://youtrack.jetbrains.com/issue/KT-84371) Incorrect @Deprecated annotation and implementation for onTimeout in kotlinx.coroutines

https://github.com/user-attachments/assets/d98f11a8-ae7c-4c2c-8e23-bf5df6c37028

### DIFF
--- a/kotlinx-coroutines-core/common/src/selects/Select.kt
+++ b/kotlinx-coroutines-core/common/src/selects/Select.kt
@@ -106,7 +106,7 @@ public sealed interface SelectBuilder<in R> {
     @Deprecated(
         message = "Replaced with the same extension function",
         level = DeprecationLevel.ERROR,
-        replaceWith = ReplaceWith(expression = "onTimeout", imports = ["kotlinx.coroutines.selects.onTimeout"])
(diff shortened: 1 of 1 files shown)


murfel:
The replaced code becomes highlighted because it suggests to replace `Long` for `Duration` milliseconds. Is it always safe/equivalent to convert Long to Milliseconds?

It seems like it ([here](https://github.com/JetBrains/kotlin/blob/2.3.20/libraries/stdlib/src/kotlin/time/Duration.kt#L982C17-L982C61)), but I want to double-check. If so, I'll convert it to milliseconds in the `replaceWith`.

dkhalanskyjb (approved):

dkhalanskyjb:
> Is it always safe/equivalent to convert Long to Milliseconds?

No, it's a minor breaking change we'll need to undergo explicitly: https://github.com/Kotlin/kotlinx.coroutines/pull/4256 In essence, `Long.MAX_VALUE / 2` is the largest number of milliseconds representable as a `Duration`, but the overloads accepting the `Long` number of milliseconds also accept the remaining half of the range.

PR #4666, opened 2026-04-26
Fix outdated `apiDump` reference in `CONTRIBUTING`

g000sha256:
`CONTRIBUTING.md` still references the `apiDump` task and the `Binary Compatibility Validator` plugin, both of which were removed in #4487. New contributors following the guide encounter the message `Task 'apiDump' not found`.


dkhalanskyjb:
Thanks!

Issue #4667, opened 2026-04-26
Add `Flow.dropFirst` operator


g000sha256:
The `dropFirst()` operator is more readable and discoverable than `drop(1)` in chains: it's self-documenting (no magic number) and aligns naming with the existing `first()`/`firstOrNull()` family. For example, to skip the `StateFlow` initial value:
```kotlin
stateFlow
    .dropFirst() // skip the initial value
    .collect { change -> handleChange(change) }
```

#### Proposed signature
```kotlin
public fun <T> Flow<T>.dropFirst(): Flow<T>
```

dkhalanskyjb:
I asked the Kotlin Libraries team what they thought of this, and the consensus is to reject this addition.
* To introduce such a function to `Flow`, you'd also need to do that for `List`, `Sequence`, and so on, to preserve consistency across the entire Kotlin.
* In the standard library, there is already a `dropLast`: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/drop-last.html If we add `dropFirst`, we're going to have `dropFirst()`, `drop(n)`, and `dropLast(n)`. Taken together, judging by the name, `dropLast` looks like a counterpart to `dropFirst`—but actually, it's a counterpart to `drop`. This is an additional point of confusion.
* The new function doesn't improve the readability of the provided snippet significantly—to that end, you'd need an even more specialized function, like `fun <T> Flow<T>.skipInitialEmission(): Flow<T> = drop(1)`:
```kotlin
stateFlow
    .skipInitialEmission()
    .collect { change -> handleChange(change) }
```
`dropFirst()` is neither here nor there: it doesn't generalize to dropping more/arbitrary numbers of elements, it doesn't clearly describe what you're trying to achieve by calling the function.

PR #4668, opened 2026-04-26
Add `Flow.dropFirst` operator

g000sha256:
See #4667


kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
```diff
@@ -24,6 +24,21 @@ public fun <T> Flow<T>.drop(count: Int): Flow<T> {
...
+ * More readable than `drop(1)` in chains - for example, to skip the [StateFlow] initial value:
+ *
+ * ```
+ * stateFlow
+ *     .dropFirst() // skip the initial value
+ *     .collect { change -> handleChange(change) }
+ * ```
+ */
+@ExperimentalCoroutinesApi
+public fun <T> Flow<T>.dropFirst(): Flow<T> =
```
ApoloApps, line 39: maybe make it inline?

g000sha256: Can we? This will require adding the `@Suppress("NOTHING_TO_INLINE")` annotation.

ApoloApps: I see no problem unless some Jetbrains maintainer says otherwise. This way we are saving 1 function call instead of having
2 (since underlying drop is already non-inlined). I've suppressed these warnings in other Jetbrains repos and had an ok in all of them so...

dkhalanskyjb:
Rejecting based on https://github.com/Kotlin/kotlinx.coroutines/issues/4667#issuecomment-4334087585

kotlinx-coroutines-core/common/src/flow/operators/Limit.kt
dkhalanskyjb: If we decided to accept this function, a `Suppress` would be okay in this scenario. We have hundreds of them in our codebase anyway.

PR #4670, opened 2026-04-27
Update the release procedure description

dkhalanskyjb:


RELEASE.md
```diff
@@ -45,18 +45,24 @@ To release a new `<version>` of `kotlinx-coroutines`:
 
 0. On [TeamCity integration server](https://teamcity.jetbrains.com/project.html?projectId=KotlinTools_KotlinxCoroutines):
    * Wait until "Build" configuration for committed `version-<version>` branch passes tests.
-   * Run "Deploy (Configure, RUN THIS ONE)" configuration with the corresponding new version:
+   * Run the [`Deployment 3.0/Deploy [RUN THIS ONE]`](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxCoroutines_Deployment30_StartDeployment)c
```
fzhinkin: It's now `Deployment/...` (without `3.0`). Otherwise, it all looks good.

RELEASE.md
```diff
@@ -45,18 +45,24 @@ To release a new `<version>` of `kotlinx-coroutines`:
...
+     configuration with the corresponding new version:
      - Use the `version-<version>` branch
-     - Set the `DeployVersion` build parameter to `<version>`
-   * Wait until all four "Deploy" configurations finish.
-
-1. In [Nexus](https://oss.sonatype.org/#stagingRepositories) admin interface:
-   * Close the repository and wait for it to verify.
-   * Release the repository.
+     - Set the `Version` build parameter to `<version>`
```
fzhinkin, line 51: BTW, if you want, we can automatically validate that the branch name and version parameter match.

dkhalanskyjb: If it's not too much effort, it would be nice. Maybe even extract the version from `gradle.properties`?

I think it's perfectly fine the way it is, though. During the validation step, you'd notice the artifacts having the wrong version.

fzhinkin (approved):

RELEASE.md
fzhinkin: It should not be a huge effort. And we can take the version from properties, indeed. Just let me know what works better and we'll configure it.

PR #4676, opened 2026-05-08
Upgrade Kotlin to 2.3.21

dkhalanskyjb:


kotlinx-coroutines-core/js/src/Promise.js.kt
```diff
@@ -1,5 +1,6 @@
 package kotlinx.coroutines
 
+@Suppress("USELESS_CAST") // on JS, arbitrary objects can be thrown
```
fzhinkin, line 3: But `JsPromiseError` is the type alias to `Throwable`. How it can be anything but throwable?

fzhinkin: Also, maybe we can use stdlib's `JsPromiseError.asJsException` instead of proving our own function for it?

dkhalanskyjb: See `PromiseTest#testAwaitPromiseRejectedWithNonKotlinException`. Removing the type cast here breaks that test. The Kotlin/JS compiler is known to be unsound and allows objects with the type `Throwable` not to be instances of `Throwable`. The Kotlin/JS team does not intend to fix the problem: https://youtrack.jetbrains.com/issue/KT-42465/Provide-a-way-to-catch-everything-in-catch-block-in-common-code Because of this, we end up working around this issue on the library side.

fzhinkin: Got it, thanks

fzhinkin (approved):

PR #4679, opened 2026-05-13
Have `Promise.await` on a rejected promise throw `JsError` on Wasm

### LINKED ISSUE #4678: [Bug] `Promise.await` is not propagating underlying JS error (only on WASM)
### What is happening? What should have happened instead?

In the open source [Kable](https://github.com/JuulLabs/kable) library, when upgrading to kotlinx.coroutines v1.11.0 started seeing [a test fail](https://github.com/JuulLabs/kable/blob/cf9324a874d8ac04c25dd094783e1a82c4c5c2ef/kable-core/src/webTest/kotlin/RequestPeripheralTests.kt#L9-L15) (only on WASM) with:

```
AssertionError: Expected an exception of class IllegalStateException to be thrown, but was Exception: Non-Kotlin exception SecurityError: Failed to execute 'requestDevice' on 'Bluetooth': Must be handling a user gesture to show a permission request. of type 'class DOMException'
```

It was expected that the failure from JS would [propagate from the `await()` as a `JsException`](https://github.com/JuulLabs/kable/blob/cf9324a874d8ac04c25dd094783e1a82c4c5c2ef/kable-core/src/webMain/kotlin/RequestPeripheral.kt#L51-L52) and thus would be able to be caught and [examined to determine the underlying cause](https://github.com/JuulLabs/kable/blob/cf9324a874d8ac04c25dd094783e1a82c4c5c2ef/kable-core/src/webMain/kotlin/RequestPeripheral.kt#L54-L62).

This was previously the case for both JS and WASM, but upon [upgrading to v1.11.0](https://github.com/JuulLabs/kable/pull/1141) of kotlinx.coroutines, only on WASM the exception from JS would [always propagate as `Exception`](https://github.com/Kotlin/kotlinx.coroutines/blob/1.11.0/kotlinx-coroutines-core/wasmJs/src/Promise.wasm.kt#L11) rather than the expected `JsException`.

### Reproducer

https://github.com/JuulLabs/kable/pull/1141 was failing in the initial upgrade commit [ab302ec](https://github.com/JuulLabs/kable/pull/1141/commits/ab302ec4906365bfd09e17decddcfa466bf2404f), until workarounds were implemented.

### Anything else? (optional)

[Workaround](https://github.com/JuulLabs/kable/pull/1141/commits/c81393e343d99b9d5ce1fb53040afef12c6d8660) was to copy the `Promise.await` implementation so that `JsPromiseError.toThrowable` could be changed:

```diff
  private fun JsPromiseError.toThrowable(): Throwable = try {
      unsafeCast<JsReference<Throwable>>().get()
  } catch (_: Throwable) {
-     Exception("Non-Kotlin exception $this of type '${this::class}'")
+     asJsException()
  }
```

dkhalanskyjb:
Fixes #4678

### DIFF
--- a/kotlinx-coroutines-core/js/src/Promise.js.kt
+++ b/kotlinx-coroutines-core/js/src/Promise.js.kt
@@ -1,9 +1,11 @@
 package kotlinx.coroutines
 
-@Suppress("USELESS_CAST") // on JS, arbitrary objects can be thrown
+@Suppress("USELESS_CAST")
 @OptIn(ExperimentalWasmJsInterop::class)
-internal actual fun JsPromiseError.toThrowable(): Throwable = this as? Throwable ?:
-    Exception("Promise resolved with a non-Throwable exception '$this' (type ${this::class})")
+// The cast looks redundant, but the `JsPromiseError` being a type alias to `Throwable` is actually incorrect.
+// A promise can be rejected with any value, not just `Throwable`.
+internal actual fun JsPromiseError.toThrowableOrNull(): Throwable? =
+    this as? Throwable
 
 @OptIn(ExperimentalWasmJsInterop::class)
 internal actual fun Throwable.toJsPromiseError(): JsPromiseError = this
--- a/kotlinx-coroutines-core/wasmJs/src/Promise.wasm.kt
+++ b/kotlinx-coroutines-core/wasmJs/src/Promise.wasm.kt
@@ -3,15 +3,10 @@ package kotlinx.coroutines
 
 import kotlin.coroutines.*
 import kotlin.js.*
+import kotlin.js.toThrowableOrNull as stdlibToThrowableOrNull
 
-@OptIn(ExperimentalWasmJsInterop::class)
-internal actual fun JsPromiseError.toThrowable(): Throwable = try {
-    unsafeCast<JsReference<Throwable>>().get()
-} catch (_: Throwable) {
-    Exception("Non-Kotlin exception $this of type '${this::class}'")
-}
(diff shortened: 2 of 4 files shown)


kotlinx-coroutines-core/js/src/Promise.js.kt
```diff
@@ -1,8 +1,10 @@
 package kotlinx.coroutines
 
 @OptIn(ExperimentalWasmJsInterop::class)
-internal actual fun JsPromiseError.toThrowable(): Throwable = this as? Throwable ?:
-    Exception("Promise resolved with a non-Throwable exception '$this' (type ${this::class})")
+// The cast looks redundant, but the `JsPromiseError` being a type alias to `Throwable` is actually incorrect.
```
fzhinkin, line 5: @JSMonk any chance it will be handled on the K/JS's stdlib side somehow (by aliasing to `Any`, for example)?

fzhinkin (approved):

kotlinx-coroutines-core/js/src/Promise.js.kt
JSMonk: In the general case, it's "incorrect", but only in 1% of cases. Since the design decision for the `catch` parameter was made with `Throwable` as a type (which is also, in the same way, "incorrect"), to be consistent, the same decision was applied to the `catch` function of `Promise`. 

I believe that changing this behavior is a huge breaking change in the sake of cases that were not even reported by users for the last decade, so I don't see any reason to introduce such a breaking change.

JSMonk (approved):
/final

PR #4684, opened 2026-05-26
feat: Add getCompletedOrNull() to Deferred<T>

### LINKED ISSUE #4630: Add getCompletedOrNull to Deferred<T>
**Use case**
In addition to the experimental `Deferred<T>::getCompleted` and `Deferred<T>::getCompletionExceptionOrNull`, it would be very useful to have `getCompletedOrNull` as a way to reduce this snippet.
```kt
if (deferred.isCompleted) {
    deferred.getCompleted()
} else {
    null
}
```

This way, it would be trivial to get an awaitable value from synchronous code and cleaner than the above snippet, or catching the IllegalState thrown by getCompleted

zbnerd:
## Summary

  Adds `getCompletedOrNull(): T?` extension function to `Deferred<T>` that returns the successfully completed result or
  `null` if the deferred has not completed successfully.

  Fixes #4630

  ### Behavior

  | State                    | Return value |
  | ------------------------ | ------------ |
  | Completed successfully   | The result   |
  | Not completed yet        | `null`       |
  | Completed exceptionally  | `null`       |

  ### Implementation

  - **Internal helper**: Added `getCompletedOrNullInternal()` to `JobSupport` for atomic state inspection
  - **Public API**: Added `getCompletedOrNull()` extension function to `Deferred`
  - **Tests**: Added comprehensive test coverage for all completion states

  ### Rationale

  This API fills a gap in the `Deferred` interface:
  - `await()` - suspends, throws on failure
  - `getCompleted()` - throws if not complete, throws on failure
  - `getCompletedOrNull()` - **returns immediately, null for any non-success**

  ### Use Cases

  - Debugging and state inspection
  - Non-blocking status checks
  - Cache snapshots

  ## Test Plan

  - [x] Returns value when completed successfully
  - [x] Returns null when not completed
  - [x] Returns null when cancelled
  - [x] Returns null when completed exceptionally

  ## Design Decisions

  **Extension function vs interface method**: Chose extension function to:
  - Avoid interface modification (ABI safety)
  - Follow existing kotlinx.coroutines patterns
  - Maintain binary compatibility

  **Exception handling approach**: Uses try-catch on public API due to:
  - Cannot access `JobSupport` internal type from public API
  - Simpler implementation for initial version
  - Internal helper available for future optimization

### DIFF
--- a/kotlinx-coroutines-core/common/src/Deferred.kt
+++ b/kotlinx-coroutines-core/common/src/Deferred.kt
@@ -100,3 +100,50 @@ public interface Deferred<out T> : Job {
     @ExperimentalCoroutinesApi
     public fun getCompletionExceptionOrNull(): Throwable?
 }
+
+/**
+ * Returns the successfully completed result of this deferred value, or `null` if this deferred
+ * has not completed successfully.
+ *
+ * This function is a non-throwing convenience accessor for inspecting the state of a `Deferred`
+ * without suspension. It returns the result only if the deferred completed **successfully**.
+ *
+ * The table below summarizes the behavior:
+ *
+ * | State                    | Return value |
+ * | ------------------------ | ------------ |
+ * | Completed successfully   | The result   |
+ * | Not completed yet        | `null`       |
+ * | Completed exceptionally  | `null`       |
+ *
+ * A deferred value is considered to have **completed exceptionally** if it was cancelled
+ * or failed with an exception. In both cases, this function returns `null`.
+ *
+ * To distinguish between these cases, use [isCompleted] and [isCancelled] properties,
+ * or use [getCompletionExceptionOrNull] to retrieve the exception.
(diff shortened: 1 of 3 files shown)


dkhalanskyjb:
Hi! We will not add `getCompletedOrNull` with that exact name, because the name can easily be misinterpreted to mean that its purpose is to handle the errors a `Deferred` can complete with. Additionally, the original request in #4630 *would* throw an error if the `Deferred` completed with an error (https://pl.kotl.in/A9LDcjdlt), which I find to be a much more practical need.

You can help this by describing your use cases in a comment for #4630: what you'd like to achieve with the exact behavior you implemented here.

Additionally, please review the contribution guidelines: https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md

PR #4686, opened 2026-06-02
Manually configure some cache-redirected repositories

Badya:


fzhinkin:
@Badya, I don't know if you love rebasing branches as much as I do, but please, rebase it back to dev (after running tests on the CI side), we'll merge the change there and then I'll merge the dev to `kotlin-community/dev`

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
```diff
@@ -78,7 +78,24 @@ fun getKotlinDevRepositoryUrl(project: Project): String? {
...
 fun addDevRepositoryIfEnabled(rh: RepositoryHandler, project: Project) {
     val devRepoUrl = getKotlinDevRepositoryUrl(project) ?: return
-    rh.maven(devRepoUrl)
+    val version = getKotlinDevVersion(project) ?: return
+    rh.exclusiveContent {
+        forRepository {
+            rh.maven(devRepoUrl)
+        }
+        filter {
+            includeVersionByRegex("org.jetbrains.kotlin", ".*", version)
```
fzhinkin, line 87: @Badya, Why it has to be done this way?

dkhalanskyjb (commented):
The entire `integration-testing/` project needs to be updated, too, I suppose?

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
Badya: This is the best way to use in such a case, and Gradle best practice.
https://docs.gradle.org/current/userguide/filtering_repository_content.html#sec:declaring-content-repositories

Looking up dev deps in central or in cache redirector (which is gonna poll central under the hood) will result in guaranteed `404`, but also will count towards central's request limit bringing up potential throttling and `429 Too Many Requests`.
By using this Gradle is guaranteed not to look up deps matching regexp elsewhere.

Additionally I suspect Gradle to continue trying other repositories on `404` but in the case of `429 Too Many Requests` from central, it just fails the build right away. Provided solution also prevents this behavior regardless of repos declaring order.

fzhinkin: Wouldn't it be enough to simply move the dev repo declaration before central repo and gradle plugins portal?

Badya: > move the dev repo declaration before

If something is wrong with dev repo and gradle fails to find artifacts there, it still gonna try other repos in that case, which we want to avoid, since we know for 100% artifacts are not there.

This is the best way to configure gradle for this case

fzhinkin: But the dev repo is specified in cases when we know for sure that the artifact should be there, isn't it?

fzhinkin:
Still trying to figure out how to avoid copy-pasting the same repository configuration from one build script into another.
Meanwhile, I cherry picked both commits from this PR into `kotlin-community/dev` to unblock our CI.

buildSrc/src/main/kotlin/CommunityProjectsBuild.kt
dkhalanskyjb: In my view, this configuration makes perfect sense. The `kotlin_dev_repo` Gradle property specifies: "You must use the Kotlin development repository for Kotlin!" That means we aren't supposed to use the non-development repository for Kotlin. The proposed config catches the error where 1) `kotlinx.coroutines` accidentally depends on a hardcoded version of Kotlin, or 2) where the TeamCity config got broken and passed a stable instead of the development version of Kotlin, or 3) where the Kotlin Gradle Plugin got broken and accidentally stripped out the dev version suffix during resolution... It's a reasonable way of ensuring we get exactly what we requested, instead of pulling in something undesired from Maven without noticing.

PR #4687, opened 2026-06-03
feat: coroutine flows documentation rewrite

daniCsorbaJB:
This PR adds the first page for the Flows updates.


docs/flow.md
```diff
@@ -1,3 +1,3 @@
 The documentation has been moved to the [https://kotlinlang.org/docs/flow.html](https://kotlinlang.org/docs/flow.html) page.
```
danil-pavlov:
```suggestion
The documentation has been moved to the [https://kotlinlang.org/docs/coroutines-flow.html](https://kotlinlang.org/docs/coroutines-flow.html) page.
```

docs/topics/debug-flow-with-idea.md
```diff
@@ -5,7 +5,7 @@
 
 This tutorial demonstrates how to create Kotlin Flow and debug it using IntelliJ IDEA.
 
-The tutorial assumes you have prior knowledge of the [coroutines](coroutines-guide.md) and [Kotlin Flow](flow.md#flows) concepts.
+The tutorial assumes you have prior knowledge of the [coroutines](coroutines-guide.md) and [Kotlin Flow](coroutines-flow.md#flows) concepts.
```
danil-pavlov:
```suggestion
The tutorial assumes you have prior knowledge of the [coroutines](coroutines-guide.md) and [flows](coroutines-flow.md) concepts.
```

docs/topics/coroutines-basics.md
```diff
@@ -623,5 +626,5 @@ the JVM thread version may throw an out-of-memory error or slow down thread crea
 * Discover more about combining suspending functions in [Composing suspending functions](composing-suspending-functions.md).
 * Learn how to cancel coroutines and handle timeouts in [Cancellation and timeouts](cancellation-and-timeouts.md).
 * Dive deeper into coroutine execution and thread management in [Coroutine context and dispatchers](coroutine-context-and-dispatchers.md).
-* Learn how to return multiple asynchronously computed values in [Asynchronous flows](flow.md).
+* Learn how to return multiple asynchronously computed values in [Asynchronous flows](coroutines-flow.md).
```
danil-pavlov:
```suggestion
* Learn how to return multiple asynchronously computed values in [Flows](coroutines-flow.md).
```

docs/topics/coroutines-guide.md
```diff
@@ -24,7 +24,7 @@ In order to use coroutines as well as follow the examples in this guide, you nee
 * [Cancellation and timeouts](cancellation-and-timeouts.md)
 * [Composing suspending functions](composing-suspending-functions.md)
 * [Coroutine context and dispatchers](coroutine-context-and-dispatchers.md)
-* [Asynchronous Flow](flow.md)
+* [Asynchronous Flow](coroutines-flow.md)
```
danil-pavlov:
```suggestion
* [Flows](coroutines-flow.md)
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        }
+    }
+    println("Creating a cold flow doesn't run it!")
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the `flow()` builder function returns a [`Flow<T>`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/), but it doesn't start executing its block.
+A cold flow is like a recipe: it defines how to produce values, but it only starts producing values when you [collect it](#collect-from-a-cold-flow).
```
danil-pavlov:
```suggestion
A cold flow is like a recipe: it defines how to produce values, but it only starts producing values when you [collect it](#collect-a-cold-flow).
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+In this example, [`CoroutineName`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-name/) adds a name to each coroutine.
+You can use `CoroutineName` for [debugging](coroutine-context-and-dispatchers.md#naming-coroutines-for-debugging), and here it helps show which collector runs each collection.
+
+### Intermediate flow operators
+
+Intermediate operators apply operations to an upstream flow and return a new downstream flow.
+They are cold, so the returned flow doesn't start processing values until it's collected, even when the upstream flow is hot.
+
+The [kotlinx.coroutines](https://github.com/Kotlin/kotlinx.coroutines) library offers a [wide range of intermediate flow operators](coroutines-flow-operators.md) for transforming and processing flows.
```
danil-pavlov, line 194: I think the link to the operators page should be commented out for now?

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+### Flow cancellation
+
+Flow cancellation stops collection when the result is no longer needed, such as when a request times out.
+
+Flow collection is tied to the coroutine that calls the `collect()` function.
+When that coroutine is canceled, collection stops and the upstream flow is canceled too.
```
danil-pavlov:
```suggestion
When that coroutine is canceled, the collection stops, and the upstream flow is canceled too.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+They keep emitting values even when no collector is active, and multiple collectors can collect the same emissions from an already active stream instead of starting a new execution.
+
+A collector of a hot flow is called a *subscriber*.
+
+You can use hot flows when multiple parts of an application need to react to the same stream of updates, such as incoming chat messages, user actions, or UI state changes.
+
+Kotlin provides two hot flow types:
+
+* [`SharedFlow`](#create-a-sharedflow) broadcasts values to multiple subscribers. Use it when you need to broadcast events that happen over time, such as messages or notifications.
+* [`StateFlow`](#create-a-stateflow) is a specialized `SharedFlow` that always holds the latest state value. Use it when you need to represent state that changes over time, such as UI state.
```
danil-pavlov, line 811:
```suggestion
* [`StateFlow`](#create-a-stateflow) is a specialized `SharedFlow` that always holds the latest state value. Use it when you need to represent a state that changes over time, such as UI state.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+### Create a `SharedFlow`
+
+[`SharedFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/) is a hot flow that broadcasts emitted values that happen over time to its subscribers.
+
+You can create `SharedFlow` with the [`MutableSharedFlow()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow.html) function.
+
+A `MutableSharedFlow` exposes functions for emitting values.
+If you expose it directly, code outside the class can emit values into the flow.
+
+To prevent this, store the mutable flow in a private [backing property](properties.md#backing-properties), and expose a read-only `SharedFlow` with the [`.asSharedFlow()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/as-shared-flow.html) function.
```
danil-pavlov:
```suggestion
To prevent this, store the mutable flow in a private [backing property](properties.md#backing-properties) and expose a read-only `SharedFlow` with the [`.asSharedFlow()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/as-shared-flow.html) function.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        // Emits the message to subscribers
+        _messages.emit(message)
+    }
+}
+```
+
+Just like with cold flows, you can use the `collect()` function to collect values from a `SharedFlow`.
+
+You can also configure a `SharedFlow` to immediately replay already emitted values to new subscribers.
+The replay cache works like a small history buffer and stores a fixed number of previous emissions.
```
danil-pavlov:
```suggestion
The replay cache works like a small history buffer, storing a fixed number of previous emissions.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+To set how many previous emissions a new subscriber receives, use the `replay` parameter in `MutableSharedFlow()`:
+
+```kotlin
+// Sets the number of already emitted messages new subscribers receive on subscription
```
danil-pavlov, line 855:
```suggestion
// Sets the number of already emitted messages that new subscribers receive on subscription
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+        // Replays the set amount of last emitted messages to new subscribers
+        replay = MESSAGES_TO_REMEMBER
+    )
+
+    val messages: SharedFlow<Message>
+        get() = _messages.asSharedFlow()
+
+    suspend fun sendMessageToEveryone(message: Message) {
+        // Emits the message to subscribes of the messages flow
```
danil-pavlov:
```suggestion
        // Emits the message to subscribers of the message flow
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        _messages.emit(message)
+    }
+}
+```
+
+Collecting a hot flow doesn't complete by itself, so you must [cancel the collecting coroutines](#cancel-hot-flows) when you no longer need them.
+
+> A hot flow doesn't have a close or cancel operation.
+> Canceling collection only stops that subscriber from collecting.
```
danil-pavlov:
```suggestion
> Canceling a collection only stops that subscriber from collecting.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+#### Use explicit backing fields to expose hot flows
+<primary-label ref="experimental-opt-in"/>
+
+You can use [explicit backing fields](whatsnew23.md#explicit-backing-fields) to expose a read-only `SharedFlow` while keeping a mutable backing field inside the class.
+
+Explicit backing fields define the implementation type in the `field` declaration.
+Inside the class, the compiler smart casts the property to the backing field type, so you can call the `emit()` function without a separate private backing property.
+
+> Explicit backing fields doesn't create the read-only wrapper that `.asSharedFlow()` provides.
```
danil-pavlov:
```suggestion
> Explicit backing fields don't create the read-only wrapper that `.asSharedFlow()` provides.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+}
+```
+{kotlin-runnable="true"}
+
+### Create a `StateFlow`
+
+[`StateFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/) is a hot flow that stores a single state value and emits updates when that value is replaced with a new one.
+New subscribers receive the current value as soon as they start collecting, and then receive new values each time the state updates.
+
+You can use `StateFlow` to represent state that changes over time, such as loading progress, UI state, or the state of an object.
```
danil-pavlov, line 1040:
```suggestion
You can use `StateFlow` to represent a state that changes over time, such as loading progress, UI state, or the state of an object.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+    )
+}
+```
+
+> Setting `value` is thread-safe and replaces the current state, but updating `value` based on its previous value isn't atomic.
+> When a new state depends on the previous state, use `.update()` instead.
+>
+{style="note"}
+
+Similarly to `MutableSharedFlow`, `MutableStateFlow` exposes APIs for emitting updates.
```
danil-pavlov:
```suggestion
Similar to `MutableSharedFlow`, `MutableStateFlow` exposes APIs for emitting updates.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+> This example uses [`GlobalScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/) only to keep the callback-based API short.
+> In your own applications, pass a `CoroutineScope` to the function that starts the work, such as `startLoading()` in this example, and launch the coroutine in that scope so callers can cancel the work when they no longer need it.
+>
+{style="note"}
+
+Since `StateFlow` is a hot flow, collection doesn't complete by itself.
```
danil-pavlov:
```suggestion
Since `StateFlow` is a hot flow,  the collection doesn’t complete on its own.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+You can also update a `StateFlow` by calculating the new state from the current state.
+Use the [`.update()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/update.html) function for these updates.
+The `.update()` function updates the value atomically, which helps when multiple coroutines update the same `MutableStateFlow`.
+
+> If you only need to update a shared value and don't need to observe state changes over time, use the Kotlin Atomics APIs, such as [`AtomicInt`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.concurrent.atomics/-atomic-int/) or [`AtomicReference`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.concurrent.atomics/-atomic-reference/).
+>
+{style="note"}
+
+Here's an example where likes are stored in a `StateFlow` and each new state is calculated from the previous state:
```
danil-pavlov:
```suggestion
Here's an example where likes are stored in a `StateFlow`, and each new state is calculated from the previous state:
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+In this example, `messageHistory` stores the full list of previous messages as the current state.
+When a new message is sent, the `.update()` function creates a new list from the previous history and adds the new message atomically.
+
+> Updating immutable collections by creating a new collection can take more time as the collection grows.
+> You can create persistent collections with the [Experimental](components-stability.md#stability-levels-explained) [`kotlinx.collections.immutable`](https://github.com/Kotlin/kotlinx.collections.immutable) library to make immutable collection updates more efficient.
+>
+{style="tip"}
+
+Since `messageHistory` is a `StateFlow`, subscribers receive the current message history when they start collecting.
+After that, they receive a new list each time a message is sent which changes the chat history.
```
danil-pavlov:
```suggestion
After that, they receive a new list each time a message is sent, which changes the chat history.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+If the upstream flow throws an exception, this collecting coroutine fails.
+Use operators such as `.catch()` or `.retry()` before sharing the flow to [handle upstream exceptions](#handle-exceptions-in-hot-flows) before the collecting coroutine fails.
+
+The built-in [`.shareIn()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/share-in.html) function provides this pattern without requiring you to create a `MutableSharedFlow` yourself.
+It also adds options for controlling when upstream collection starts and stops, and how many previous emissions new subscribers receive.
+
+To use the built-in `.shareIn()` function, provide the following arguments:
+
+1. The coroutine scope where the upstream flow is collected.
```
danil-pavlov:
```suggestion
* The coroutine scope where the upstream flow is collected.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+2. The [`SharingStarted`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-sharing-started/) strategy to control when upstream collection starts and stops.
```
danil-pavlov:
```suggestion
* The [`SharingStarted`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-sharing-started/) strategy that controls when upstream collection starts and stops.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+   For example, [`SharingStarted.Eagerly`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-sharing-started/-companion/-eagerly.html) starts upstream collection immediately in the provided scope, before any subscriber starts collecting.
+3. The optional `replay` value to control how many previous emissions new subscribers receive.
```
danil-pavlov:
```suggestion
* The optional `replay` value that controls how many previous emissions new subscribers receive.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+class Chatroom {
+    // Stores the message flow
+    private val _messages = MutableSharedFlow<Message>()
+
+    // Exposes a read-only SharedFlow with emitted messages
+    // New subscribers don't receive already emitted messages
+    val messages: SharedFlow<Message>
+        get() = _messages.asSharedFlow()
+  
+    // Sends a message to all subscribers of the messages flow
```
danil-pavlov, line 1443:
```suggestion
    // Sends a message to all subscribers of the message flow
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+Hot flows don't stop when a subscriber is canceled.
+
+When you cancel the coroutine that collects a hot flow, you only cancel that subscriber.
+The hot flow can still emit values to other subscribers, and the coroutine that produces those values can keep running.
+
+Hot flows themselves don't have a cancellation operation.
+To cancel a hot flow, cancel the coroutine or scope that produces values for it.
+
+Hot flows created with the `.shareIn()` or `.stateIn()` extension functions keep collecting the upstream flow until the sharing coroutine is canceled.
+To stop collection from the upstream flow, cancel the scope that runs that sharing coroutine.
```
danil-pavlov:
```suggestion
To stop collection from the upstream flow, cancel the scope that runs the sharing coroutine.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+Hot flows don't stop when a subscriber is canceled.
+
+When you cancel the coroutine that collects a hot flow, you only cancel that subscriber.
+The hot flow can still emit values to other subscribers, and the coroutine that produces those values can keep running.
+
+Hot flows themselves don't have a cancellation operation.
+To cancel a hot flow, cancel the coroutine or scope that produces values for it.
+
+Hot flows created with the `.shareIn()` or `.stateIn()` extension functions keep collecting the upstream flow until the sharing coroutine is canceled.
```
danil-pavlov:
```suggestion
Hot flows created with the `.shareIn()` or `.stateIn()` extension functions continue to collect the upstream flow until the sharing coroutine is canceled.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+To stop collection from the upstream flow, cancel the scope that runs that sharing coroutine.
+
+> You can also stop upstream collection automatically when there are no subscribers by using [`SharingStarted.WhileSubscribed()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-sharing-started/-companion/-while-subscribed.html).
```
danil-pavlov:
```suggestion
> You can also automatically stop upstream collection when there are no subscribers using [`SharingStarted.WhileSubscribed()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-sharing-started/-companion/-while-subscribed.html).
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+            }
+        }
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, [`CoroutineName`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-name/) adds a name to each coroutine.
+You can use `CoroutineName` for [debugging](coroutine-context-and-dispatchers.md#naming-coroutines-for-debugging), and here it helps show which collector runs each collection.
```
danil-pavlov:
```suggestion
You can use `CoroutineName` for [debugging](coroutine-context-and-dispatchers.md#naming-coroutines-for-debugging). Here, it helps show which collector runs each collection.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        .collect()
+}
+```
+{kotlin-runnable="true"}
+
+In this example, the `.onEach()` operator runs upstream from `.catch()`, so the `.catch()` operator handles the exception when the `require()` check fails for `'5'`.
+
+#### Restart the upstream flow after an exception
+
+Some operations may fail temporarily, such as a network request that loses connection.
```
danil-pavlov:
```suggestion
Some operations may temporarily fail, such as a network request that loses its connection.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+suspend fun main() {
+    (0..1000).asFlow().myTake(3).collect {
+        println("Got $it")
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, the `.myTake()` function emits values from the upstream flow until it emits the requested number of values.
```
danil-pavlov:
```suggestion
In this example, the `.myTake()` function emits values from the upstream flow until it emits the requested number.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+Kotlin provides two hot flow types:
+
+* [`SharedFlow`](#create-a-sharedflow) broadcasts values to multiple subscribers. Use it when you need to broadcast events that happen over time, such as messages or notifications.
+* [`StateFlow`](#create-a-stateflow) is a specialized `SharedFlow` that always holds the latest state value. Use it when you need to represent state that changes over time, such as UI state.
+
+### Create a `SharedFlow`
+
+[`SharedFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/) is a hot flow that broadcasts emitted values that happen over time to its subscribers.
+
+You can create `SharedFlow` with the [`MutableSharedFlow()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow.html) function.
```
danil-pavlov:
```suggestion
You can create a `SharedFlow` with the [`MutableSharedFlow()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow.html) function.
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+suspend fun main() {
+    withContext(Dispatchers.Default) {
+        val post = Post(15)
+        val notifyingJob = launch {
+            post.numberOfLikes.collect {
+                drawUpdatedNumberOfLikes(it)
+            }
+        }
+        // Simulates users that like the post
```
danil-pavlov:
```suggestion
        // Simulates users who like the post
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In a flow, values move from an emitter toward a collector, from *upstream* to *downstream*.
+Intermediate operators collect an upstream flow, apply an operation to its values, and return a new downstream flow.
+That downstream flow can become the upstream flow for the next collector.
+
+![Parts of a flow: emitter, intermediate operator (optional), collector. Values move from upstream to downstream.](flow-upstream-downstream.svg){width="700px"}
```
danil-pavlov:
```suggestion
![Parts of a flow: emitter, intermediate operator (optional), collector. Values move from upstream to downstream.](flow-upstream-downstream.svg){width=700}
```

docs/kc.tree
```diff
@@ -8,7 +8,7 @@
 		<toc-element topic="cancellation-and-timeouts.md" toc-title="Cancellation"/>
 		<toc-element topic="composing-suspending-functions.md"/>
 		<toc-element topic="coroutine-context-and-dispatchers.md"/>
-		<toc-element topic="flow.md"/>
+		<toc-element topic="coroutines-flow.md"/>
```
danil-pavlov: JFYI, the old link "flow.md" is also used in

api-guidelines-readability.md:17:46; api-guidelines-readability.md:17:81; api-guidelines-simplicity.md:40:10; coroutines-overview.md:67:27

Maybe it also makes sense to update the old redirect and add a new one to kotlin-website's `site.yml` file?

docs/topics/coroutines-flow.md
danil-pavlov: nitpick: the upstream/downstream line at the bottom is not aligned with the text bubbles above

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+* [**Hot flows**](#hot-flows) emit values independently of collectors and share the same stream of values with all collectors.
+
+> You can use the [Turbine library](https://github.com/cashapp/turbine) to test Kotlin flows.
+> It simplifies collecting and asserting flow emissions in unit tests, including completion and failure cases.
+> 
+{style="tip"}
+
+## Cold flows
+
+Like [sequences](sequences.md), cold flows are lazy.
```
danil-pavlov, line 56: maybe a link to an explanation about 'lazy'?

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+//sampleStart
+fun main() {
+    // Creates a flow from provided values
+    val predefinedPageFlow = flowOf("Page 1", "Page 2", "Page 3")
+    // Creates a flow from a range
+    val generatedPageFlow = (1..3).asFlow()
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov: not sure if this one should be runnable, no output

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        }
+    } catch (e: MyFlowException) {
+        println("Flow collection failed with $e")
+        // Rethrows the exception to the caller
+        throw e
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov: here I only get "Cannot convert undefined or null to object"

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+}
+
+suspend fun main() {
+    loadBlob("https://example.com/").collect {
+        println("Got '$it'")
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov: Is it expected to get all 100% or not? (getting "Loading data failed with java.io.IOException: Failed to load!
Got 'LoadingState$Failed@7466d320' after 84%)

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+            require(!it.isDigit()) { "Digits are not allowed!" }
+            println("Got '$it'")
+        }
+        .catch { e ->
+            println("Caught an exception: $e")
+        }
+        .collect()
+}
+```
+{kotlin-runnable="true"}
```
danil-pavlov, line 524: MismatchedInputException error

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        }
+        delay(100.milliseconds)
+
+        // Cancels the coroutine that collects the flow
+        job.cancel()
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov, line 649: "Cannot convert undefined or null to object" error

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+}
+
+suspend fun main() {
+    (0..1000).asFlow().myTake(3).collect {
+        println("Got $it")
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov, line 699: java.io.IOException: Cannot run program

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+}
+
+suspend fun main() {
+    val flow1 = (0..3).asFlow().onEach { delay(20.milliseconds) }
+    val flow2 = (6..9).asFlow().onEach { delay(50.milliseconds) }
+    flow1.myMerge(flow2).collect { println(it) }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov, line 749: Cannot convert undefined or null to object

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+  
+    // Removes the buffer so sending and processing interleave from the start
+    oneHundredNumbers.buffer(0).collect {
+        println("Processing $it")
+        delay(10.milliseconds)
+    }
+}
+//sampleEnd
+```
+{kotlin-runnable="true"}
```
danil-pavlov, line 790: Cannot convert undefined or null to object

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+        }
+        // Delays to make sure people have enough time to chat
+        delay(100.milliseconds)
+        // Cancels readers because SharedFlow collection doesn't finish by itself
+        messageReaders.forEach { it.cancel() }
+    }
+
+}
+```
+{kotlin-runnable="true"}
```
danil-pavlov, line 952: Cannot convert undefined or null to object

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+//sampleEnd
+```
+{kotlin-runnable="true"}
+
+In this example, when you call the `derivedFlowsScope.cancel()` function, `totalMessages` stops collecting updates from `messageHistory`.
+
+The `sendMessageToEveryone()` function still updates `messageHistory` because the coroutine that calls it wasn't canceled.
+As a result, `totalMessages.value` keeps the last collected size, while `chatroom.messageHistory.value.size` shows the actual number of messages.
+
+#### Handle exceptions in hot flows
```
danil-pavlov:
```suggestion
### Handle exceptions in hot flows
```

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+            // Starts collecting when the first subscriber appears
+            // and stops when the last subscriber disappears
+            SharingStarted.WhileSubscribed(),
+            // Sets the initial state before the first upstream emission
+            null,
+        )
+```
+
+#### Cancel hot flows
```
danil-pavlov:
```suggestion
### Cancel hot flows
```

danil-pavlov: Otherwise the header is not visible in the right-hand menu, which is inconsistent with Cold flows

docs/topics/coroutines-flow.md
```diff
@@ -0,0 +1,1708 @@
...
+
+> Updating immutable collections by creating a new collection can take more time as the collection grows.
+> You can create persistent collections with the [Experimental](components-stability.md#stability-levels-explained) [`kotlinx.collections.immutable`](https://github.com/Kotlin/kotlinx.collections.immutable) library to make immutable collection updates more efficient.
+>
+{style="tip"}
+
+Since `messageHistory` is a `StateFlow`, subscribers receive the current message history when they start collecting.
+After that, they receive a new list each time a message is sent which changes the chat history.
+
+### Convert cold flows to hot flows
```
danil-pavlov, line 1371: Do you place it here as another option to create hot flows? Maybe explicitly present it as "Create hot flows from cold flows?"

danil-pavlov (changes requested):

danil-pavlov (changes requested):
Please check that runnable snippets behave as expected

docs/topics/coroutines-flow.md
daniCsorbaJB: Based on the discussion in the channel, we'll add this page, but only with the current content (without any updates)   — I'll add it when pushing the changes to fix the link issue

docs/topics/coroutines-flow.md
daniCsorbaJB: 🤔 here, state refers to uncountable concepts so I'd go without an article

docs/topics/coroutines-flow.md
daniCsorbaJB: hmm 🤔 I prefer without a comma here because it's a compound predicate

docs/topics/coroutines-flow.md
daniCsorbaJB: ah nice catch on the typo! 🙏 — the flow should be messages (message is a single emitted value)

docs/topics/coroutines-flow.md
daniCsorbaJB: hmm it should be again a concept here, but I see what could be the issue (stops that subscriber makes it read a bit like we are talking about a specific situation and not a general one) 🤔 

I rewrote it as: 

> Canceling collection only stops the corresponding subscriber from collecting.

docs/topics/coroutines-flow.md
daniCsorbaJB: same as above - should be an uncountable state

docs/topics/coroutines-flow.md
daniCsorbaJB: same here, the flow is messages (individual value is message)

docs/topics/coroutines-flow.md
daniCsorbaJB: Both are correct, just a nit why I prefer the original: it contrasts the new idea more: You can stop X with this - you can also stop x automatically with that.

(agree with removing the by! 👍 🙏 )

docs/topics/coroutines-flow.md
daniCsorbaJB: I think that's a bit less specific 🤔 (I'm worried it could be misread as `.take(2)` means until it emits the number `2` not two values.

daniCsorbaJB: so perhaps let's change it to specified number of values to be more clear? 🤔

docs/topics/coroutines-flow.md
daniCsorbaJB: ah cheers — I aligned it 👍

docs/kc.tree
daniCsorbaJB: good idea 🤔 — we could also add accepts-web-file-names="flow.html"
I'll open up separate PRs for those.

docs/topics/coroutines-flow.md
daniCsorbaJB: shouldn't be 👍 I removed runnable.

docs/topics/coroutines-flow.md
daniCsorbaJB: Do we have one? I think since it's a generic programming concept it never made it into the docs/ I never managed to find a generic one (we have descriptions for lazy properties and lazy init in Kotlin/JS 😢  — we have some APIs where it's described in the given context 🤔 I think it should be fine, the next sentence explains immediately what it means in this context.

docs/topics/coroutines-flow.md
daniCsorbaJB: That's weird 🤔 I get: 

Collector threw MyFlowException: Can't process 'a'!
Flow collection failed with MyFlowException: Can't process 'a'!
Exception in thread "main" MyFlowException: Can't process 'a'!
 at FileKt$main$2.emit (File.kt:22) 
 at FileKt$main$2.emit (File.kt:20) 
 at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke (SafeCollector.kt:15)

docs/topics/coroutines-flow.md
daniCsorbaJB: nope not always. that's why we handle it as an expected exception here: 

```
.catch { e ->
    println("Loading data failed with $e")
    if (e is IOException) {
        // Handles an expected exception
        emit(LoadingState.Failed)
    } else {
        // Rethrows unexpected exceptions, so the collect() fails with them
        throw e
    }
```

docs/topics/coroutines-flow.md
daniCsorbaJB: I can't reproduce it 😢

docs/topics/coroutines-flow.md
daniCsorbaJB: I can't reproduce this one either 😢 works for me

docs/topics/coroutines-flow.md
daniCsorbaJB: same — works for me 😢

docs/topics/coroutines-flow.md
daniCsorbaJB: same 😢 works for me.

docs/topics/coroutines-flow.md
daniCsorbaJB: same — should work 😢

docs/topics/coroutines-flow.md
daniCsorbaJB: it is. We added it like this to try to ensure that people read these sections in the context of the 
Convert cold flows to hot flows section, because we were a bit worried that it might be confusing otherwise. 🤔 

Regardless, I agree it works as a 3rd level heading as well and makes navigation easier...

docs/topics/coroutines-flow.md
daniCsorbaJB: That makes sense, and I agree that “create” is technically accurate. 🤔  I still prefer “convert” here because it better reflects the user’s goal iiuc: they start with a cold flow that is too expensive to collect repeatedly, then use `shareIn()` or `stateIn()` to convert it into a hot flow. Even though this creates a new hot flow, the important concept is the conversion from cold to hot.

docs/topics/coroutines-flow.md
danil-pavlov: I don't think so, it's basically 2 coordinated verbs in the main clause "store and expose"

docs/topics/coroutines-flow.md
danil-pavlov: Just from a page-structure point of view, if the main thing is creating hot flows from cold flows, it makes sense to keep it here under Hot flows, next to other "create" options. But if the focus is on the conversion, then I'd place it below on the second level since it is not a part of Hot flows or Cold flows:

```
## Cold flows
## Hot flows
## Convert
```

docs/topics/coroutines-flow.md
danil-pavlov: Yeah, I get it, but it's a standard noun as adjective (noun modifier), always singular, even if it means plural

daniCsorbaJB: hmm 🤔 What I meant is that's the name of the flow: `messages` so calling it message is misleading (that's one value in the `messages` flow )

Like in another example // Sends a message to all subscribers of the messageHistory flow

docs/topics/coroutines-flow.md
danil-pavlov: then I think we should add What's next to that 2nd page at the bottom

docs/topics/coroutines-flow.md
daniCsorbaJB: Oh I get what you mean — I agree 👍

docs/topics/coroutines-flow.md
danil-pavlov: or "until all requested values are emitted."

docs/topics/coroutines-flow.md
danil-pavlov: It works now, maybe 100 is too many? the output is so long on the page

docs/topics/coroutines-flow.md
danil-pavlov: oh, I didn't realize you're referring here to actual variable names! maybe add backticks?

docs/topics/coroutines-flow.md
daniCsorbaJB: good point, let's go in 10 increments -- 10..20..30.. etc. I'll change it

docs/topics/coroutines-flow.md
daniCsorbaJB: Yes definitely, but maybe only add it when we update it (since that one will be fully rewritten) WDYT?

docs/topics/coroutines-flow.md
daniCsorbaJB: 🤔 
WDYT about like this:
    // Sends a message to all subscribers of the flow: messages
we never put backticks to comments for the other coroutines docs (I don't really see it in other docs either 🤔 ) 

or just leave it as is? 🤔

danil-pavlov (approved):

docs/topics/coroutines-flow.md
daniCsorbaJB: that works as well 👍 although it's passive, but I think it's understandable.

docs/topics/coroutines-flow.md
daniCsorbaJB: as discussed on Slack let's keep the original structure then 👍

fzhinkin (approved):
I didn't thoroughly check the doc (although, I red it), only verified the what when to a temporary "under construction" flow ops page, and what is in the new flow page.

It all looks good to me, thanks for making the better guideline!

Issue #4683, opened 2026-05-21
[Feature] make Window.awaitAnimationFrame common for wasm and js


eymar:
### What are you trying to do?

I noticed a nice function in js target - https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/js/src/Window.kt#L20 

I'd like to use this function in webMain source set. 

### What problem does this actually solve?

It would solve the necessity to keep additional expect/actual in the user projects.

dkhalanskyjb:
Yes, we should just move all our `public` APIs from Kotlin/JS to the web target.

PR #4692, opened 2026-07-07
Move W3CWindow coroutine APIs to the common web target

MARKOCEVRLJAKOVIC:
Closes [#4683](https://github.com/Kotlin/kotlinx.coroutines/issues/4683)

## Summary

Moves the public `W3CWindow` coroutine APIs from the JS source set to the common `web` source set, making them available for both JS and WasmJS.

## Changes

- move `W3CWindow.awaitAnimationFrame` to the common `web` source set
- move `W3CWindow.asCoroutineDispatcher` to the common `web` source set
- add common `w3cRequestAnimationFrame` expect/actual declarations
- add JS and WasmJS implementations
- add tests covering dispatcher caching and `awaitAnimationFrame` behavior
- update the public API dumps


dkhalanskyjb:
This PR is unfortunately too far away from being production-ready, so we will not accept it. Some (but not all) of the problems:
* Building this fails, because the links to the API reference got updated (this is tracked by the Knit tool).
* The added tests don't test anything at all: we only run tests under Node.js and don't touch browsers at all.
* The new APIs can't be called and are entirely useless, since `W3CWindow` is `internal`.

In general, this leaves the impression of an LLM-generated pull request. Please don't send us those.

Issue #3594, opened 2023-01-19
Using catch on a Flow prevents flatMapLatest from triggering again


npars:
**Describe the bug**

When `catch {}` is used on a flow it prevents any future events from being emitted after an exception has been caught.

The documentation states:
> Conceptually, the action of catch operator is similar to wrapping the code of upstream flows with try { ... } catch (e: Throwable) { action(e) }.

The implementation does not appear to be inline with this statement. If I placed a try/catch in my upstream flow code it would continue to emit values after an exception has been caught.

This gotcha is especially prevalent in Android code, where you may have a flow for triggering a refresh which then is used with `flatMapLatest {}` to trigger a network request.

See also this StackOverflow post: https://stackoverflow.com/q/74045676

**Provide a Reproducer**

```kt
    // Failing Test 
    @Test
    fun `a catch on the outer flow prevents the flatMapLatest from firing again`() = runTest {
        val collectedResults = mutableListOf<Int>()
        val refreshFlow = MutableStateFlow(0)

        val job = launch {
            refreshFlow
                .flatMapLatest { flow<Int> { throw IllegalArgumentException(it.toString()) } }
                .catch { e -> emit(e.message?.toIntOrNull() ?: -1) }
                .collect { collectedResults.add(it) }
        }

        advanceUntilIdle()

        refreshFlow.emit(1)
        advanceUntilIdle()

        assertThat(collectedResults).containsExactly(0, 1)
        job.cancel()
    }

    // Passing Test 
    @Test
    fun `a catch on the inner flow works as expected`() = runTest {
        val collectedResults = mutableListOf<Int>()
        val refreshFlow = MutableStateFlow(0)

        val job = launch {
            refreshFlow
                .flatMapLatest {
                    flow<Int> { throw IllegalArgumentException(it.toString()) }
                        .catch { e -> emit(e.message?.toIntOrNull() ?: -1) }
                }
                .collect { collectedResults.add(it) }
        }

        advanceUntilIdle()

        refreshFlow.emit(1)
        advanceUntilIdle()

        assertThat(collectedResults).containsExactly(0, 1)
        job.cancel()
    }
```

JakeWharton:
> If I placed a try/catch in my upstream flow code it would continue to emit values after an exception has been caught.

This is the disconnect. You are stating that the try/catch is _within_ the upstream flow code and the documentation is stating the try/catch is _around_ the upstream flow code.

[...]

PR #4693, opened 2026-07-08
Clarify docs on flow completion with exception and catch operator

vsartiushkin:
PR highlights that when the upstream flow throws, it completes, and the original flow can no longer produce the values, even if `.catch` is applied

Unified the usage of  'complete with an exception' across both `Flow` and `.catch`.

Addresses the misconception that arose in #3594

### DIFF
--- a/kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
@@ -10,8 +10,10 @@ import kotlin.jvm.*
 import kotlinx.coroutines.flow.internal.unsafeFlow as flow
 
 /**
- * Catches exceptions in the flow completion and calls a specified [action] with
- * the caught exception. This operator is *transparent* to exceptions that occur
+ * Catches exceptions in the upstream flow completion.
+ * After the upstream flow has completed with an exception, the [action] is called with the caught exception
+ * and may emit fallback values downstream.
+ * This operator is *transparent* to exceptions that occur
  * in downstream flow and does not catch exceptions that are thrown to cancel the flow.
  *
  * For example:
@@ -39,7 +41,7 @@ import kotlinx.coroutines.flow.internal.unsafeFlow as flow
  * }
  * ```
  *
- * The [action] code has [FlowCollector] as a receiver and can [emit][FlowCollector.emit] values downstream.
+ * The [action] code has [FlowCollector] as a receiver and can [emit][FlowCollector.emit] fallback values downstream.
  * For example, caught exception can be replaced with some wrapper value for errors:
  *
  * ```
--- a/kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
+++ b/kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
@@ -11,8 +11,6 @@ import kotlinx.coroutines.flow.internal.unsafeFlow as flow
 
 /**
  * If the upstream flow completes with an exception, this operator catches that exception and calls the specified [action] with it.
- * After completing with an exception, the flow cannot emit any more values, and fallback values may be emitted
- * downstream from the [action] block instead.
  * This operator is *transparent* to exceptions that occur
  * in downstream flow and does not catch exceptions that are thrown to cancel the flow.
  *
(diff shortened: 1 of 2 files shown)


murfel:
More like: this PR highlights that when the upstream flow _throws_, it _completes_, and so the original flow can no longer produce the values, even if `.catch` is used.

murfel:
@vsartiushkin could you link in the issue which prompted this rewording?

kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -10,8 +10,10 @@ import kotlin.jvm.*
 import kotlinx.coroutines.flow.internal.unsafeFlow as flow
 
 /**
- * Catches exceptions in the flow completion and calls a specified [action] with
- * the caught exception. This operator is *transparent* to exceptions that occur
+ * Catches the exception if the upstream flow completes with one, and calls the specified [action] with it.
```
murfel: Should be 'an exception', and the "if completes with **one**" wording is not very clear. I still think this wording is more clear:
> If the flow completes with an exception, catches the exception and calls ...

kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -10,8 +10,10 @@ import kotlin.jvm.*
...
+ * After completing with the exception, the flow cannot emit any more values, and fallback values may be emitted
```
murfel: **an** exception

murfel (approved):

kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
vsartiushkin: The "If ..." wording is much clearer, but breaks the convention that all flow operator's descriptions open with the verb i.e. "Applies" "Returns" "Groups", excluding only the terminal operators.

murfel: I think it's ok to break that for the sake of clarity!

murfel:
Also please change the base branch to `master`: all docs and tutorial changes can go there, so that they appear on the website immediately after merge.

kotlinx-coroutines-core/common/src/flow/operators/Errors.kt
```diff
@@ -10,8 +10,10 @@ import kotlin.jvm.*
 import kotlinx.coroutines.flow.internal.unsafeFlow as flow
 
 /**
- * Catches exceptions in the flow completion and calls a specified [action] with
- * the caught exception. This operator is *transparent* to exceptions that occur
+ * If the upstream flow completes with an exception, this operator catches that exception and calls the specified [action] with it.
+ * After completing with an exception, the flow cannot emit any more values, and fallback values may be emitted
+ * downstream from the [action] block instead.
```
dkhalanskyjb: Keeping this part in the headline doesn't feel right to me. We only observed confusion about this once in the history of the API, and also, this can't logically work any other way, so clarifying this right in the general description of the API's purpose looks off.

In my opinion, the better place for this notice is right after this line:
> `try { ... } catch (e: Throwable) { action(e) }`.

This reinforces the similarity with throwing exceptions in general: once an exception is thrown, you can't go back to the line that `throw`ed and redo the computation.

dkhalanskyjb (changes requested):
Technically, this PR is supposed to target `develop` (https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md#submitting-prs):
> Documentation in markdown files can be updated directly in the master branch, unless **the documentation is in the source code, and the patch changes line numbers**.

That said, let's keep it in `master` and see what happens? The reference documentation publication scheme has changed in recent years, and maybe that rule is outdated. In the worst-case scenario, only the links to `retry` and `retryWhen` will break.

dkhalanskyjb (approved):
Thanks!

dkhalanskyjb:
The experiment was a success: the website rebuilt itself, the line numbers got updated. Changed the rules to account for that: https://github.com/Kotlin/kotlinx.coroutines/commit/04ada74fae2e8914ae92ece34e06e80bb15385e9

PR #4696, opened 2026-07-13
[DO NOT MERGE] Apply ktfmt --kotlinlang-style formatting for demonstration purposes

qwwdfsad:


dkhalanskyjb:
What is the further process with this demonstration? Is there a discussion place where we'd exchange our opinionated takes about which of the newlines the formatter suggests are useful and which aren't?

qwwdfsad:
@dkhalanskyjb so far only internal ones (`#kotlin-team` for general discussion, `#prj-ktfmt` for specific rules/patterns)

PR #4699, opened 2026-07-16
KT-78078: remove watchosArm32

mMaxy:


dkhalanskyjb (approved):
Let's not merge it for now, as this is technically a breaking change. The next few releases won't be on Kotlin 2.5 anyway, so `kotlinx.coroutines` can keep this target for some time.

fzhinkin:
Yep, I'll merge the change to `kotlin-community/dev` only.

PR #4700, opened 2026-07-16
Correctly check version in MavenPublicationMetaInfValidator test

fzhinkin:
One of the tests behavior varies depending on a Kotlin version, and currently the version check is incorrect. It assumes that there's new compiler behavior for Kotlin 2.4 only, while the module names are different starting from 2.4. As a result, the test fails in Kotlin aggregate builds where it is built and executed w/ Kotlin 2.5.x.

### DIFF
--- a/integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
+++ b/integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
@@ -27,7 +27,7 @@ class MavenPublicationMetaInfValidator {
         val clazz = Class.forName("kotlinx.coroutines.android.HandlerDispatcher")
         val expectedEntries = buildSet {
             add("MANIFEST.MF")
-            if (kotlin24OrNewer(kotlinVersion)) {
+            if (kotlinVersion >= "2.4") { // TODO: remove the check after updating Kotlin version to 2.4+
                 add("org.jetbrains.kotlinx_kotlinx-coroutines-android.kotlin_module")
             } else {
                 add("kotlinx-coroutines-android.kotlin_module")
@@ -66,13 +66,4 @@ class MavenPublicationMetaInfValidator {
 
     private val kotlinVersion: String
         get() = System.getProperty("kotlin.version.integration")
-
-    private fun kotlin24OrNewer(version: String): Boolean {
-        val majorVersion = version.substringBefore('.').toIntOrNull() ?: return false
-        if (majorVersion > 2) return true
-        if (majorVersion < 2) return false
-
-        val minorVersion = version.substringAfter('.').substringBefore('.').toIntOrNull() ?: return false
-        return minorVersion >= 4
-    }
 }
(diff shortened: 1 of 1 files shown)


integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
```diff
@@ -66,4 +66,13 @@ class MavenPublicationMetaInfValidator {
...
         get() = System.getProperty("kotlin.version.integration")
+
+    private fun kotlin24OrNewer(version: String): Boolean {
+        val majorVersion = version.substringBefore('.').toIntOrNull() ?: return false
+        if (majorVersion > 2) return true
+        if (majorVersion < 2) return false
+
+        val minorVersion = version.substringAfter('.').substringBefore('.').toIntOrNull() ?: return false
+        return minorVersion >= 4
+    }
```
dkhalanskyjb: Semantically, this function is equivalent to the hypothetical `parseKotlinVersionOrNull(version)?.let { it >= parseKotlinVersion("2.4") } ?: false`. I disagree with the `?: false` part here. Realistically, we aren't going to go back to older Kotlin versions, so any unknown/unparseable string in this part is clearly going to be newer than 2.4. At the very least, then, this should be `?: true`, or maybe even `?: error("Unknown Kotlin version '$version'")`.

I'm not sure even a separate function is even justified, though. `kotlinVersion >= "2.4" /* TODO: replace with `true` when upgrading to Kotlin 2.4+ */` would do the trick with the current versioning scheme right until `2.10`.

integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
```diff
@@ -66,4 +66,13 @@ class MavenPublicationMetaInfValidator {
 
     private val kotlinVersion: String
         get() = System.getProperty("kotlin.version.integration")
+
+    private fun kotlin24OrNewer(version: String): Boolean {
+        val majorVersion = version.substringBefore('.').toIntOrNull() ?: return false
```
Tapchicoma: I would suggest adding a dependency on `kotlin-tooling-core` artifact and using [KotlinToolingVersion](https://github.com/JetBrains/kotlin/blob/ae1166d296d6985816304b5136fac7854dea85e7/libraries/tools/kotlin-tooling-core/src/main/kotlin/org/jetbrains/kotlin/tooling/core/KotlinToolingVersion.kt#L37) instead of a custom parser.

integration-testing/src/mavenTest/kotlin/MavenPublicationMetaInfValidator.kt
fzhinkin: That's fair. Let's use a simple string comparison then, and switch to a different approach when it will be time to update Kotlin to 2.10 (thankfully, we have a few more years).

Adding a dependency on `kotlin-tooling-core` would do the trick, but I'm somewhat reluctant to do that for a single check like this.

Tapchicoma (approved):

dkhalanskyjb (approved):

PR #4702, opened 2026-07-17
update: rewrite the coroutines flow operators page

daniCsorbaJB:
This PR updates the Flow operators page. 

The content has been reviewed by Dmitry Khalanskiy

### DIFF
--- a/docs/topics/coroutines-flow-operators.md
+++ b/docs/topics/coroutines-flow-operators.md
@@ -337,9 +337,7 @@ suspend fun main() = withContext(Dispatchers.Default) {
 {kotlin-runnable="true"}
 
 You can also use the [`.conflate()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/conflate.html) operator, which is a shorthand for `buffer(1, onBufferOverflow = BufferOverflow.DROP_OLDEST)`.
-Use it when you only want to process the latest values, skipping the values emitted while the previous value is being collected.
-
-Here's an example:
+Use it when you only want to process the latest values, skipping the values emitted while the previous value is being collected:
 
 ```kotlin
 import kotlinx.coroutines.*
--- a/docs/topics/coroutines-flow-operators.md
+++ b/docs/topics/coroutines-flow-operators.md
@@ -171,7 +171,7 @@ import kotlinx.coroutines.flow.*
 
 // A simplified custom version of the default .distinctUntilChanged() operator
 fun <T> Flow<T>.myDistinctUntilChanged(): Flow<T> = flow {
-    var lastEmitted: Any? = Any() // A value that is equal only to itself
+    var lastEmitted: Any? = Any() // A value that's equal only to itself
     this@myDistinctUntilChanged.collect { value ->
         if (lastEmitted != value) {
             this@flow.emit(value)
--- a/docs/topics/coroutines-flow-operators.md
+++ b/docs/topics/coroutines-flow-operators.md
@@ -798,7 +798,7 @@ suspend fun main() {
 {kotlin-runnable="true"}
 
 Some terminal operators collect a flow and return a result based on the collected values.
-For example, you can use the [`.first()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/first.html) operator to return the first emitted value and then cancel the collection:
+For example, you can use the [`.first()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/first.html) operator to return the first emitted value and then cancel collection:
 
 ```kotlin
 import kotlinx.coroutines.*
@@ -884,7 +884,7 @@ suspend fun main() {
 ### Collect a flow in a specific `CoroutineScope`
 
 When a screen or another long-lived object needs values from a flow, start the collector in that object's `CoroutineScope`.
-This ensures that canceling the object's `CoroutineScope` when the object is destroyed also cancels the collection.
+This ensures that canceling the object's `CoroutineScope` when the object is destroyed also cancels collection.
 
 To collect a flow in a specific `CoroutineScope`, use the [`.launchIn()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/launch-in.html) terminal operator.
 This operator returns the `Job` of the collecting coroutine.
--- a/docs/topics/coroutines-flow-operators.md
+++ b/docs/topics/coroutines-flow-operators.md
@@ -3,11 +3,10 @@
 [//]: # (title: Flow operators)
 
 Flow operators let you transform and process values in a flow pipeline.
-
 Kotlin provides two main kinds of flow operators:
 
 * [**Intermediate operators**](#intermediate-operators) return a new downstream flow that consumes values from upstream flows and applies an operation to them.
-* [**Terminal operators**](#terminal-operators) trigger execution of the flow pipeline by collecting an upstream flow. They may also return a result.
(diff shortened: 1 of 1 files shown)


docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
-1
-3
-Collected in 758 ms
-```             
+By default, a flow pipeline processes values sequentially.
+The upstream flow emits a value, and collectors process it before the next value is emitted.
 
-<!--- TEST ARBITRARY_TIME -->
+To run the upstream flow concurrently with downstream collection, use concurrent processing operators to introduce a buffer.
+A buffer stores values that the upstream flow has emitted but the collector hasn't processed yet.
```
danil-pavlov:
```suggestion
A buffer stores values that the upstream flow has emitted, but the collector hasn't processed yet.
```

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
-```text
-1
-3
-Collected in 758 ms
-```             
+By default, a flow pipeline processes values sequentially.
+The upstream flow emits a value, and collectors process it before the next value is emitted.
 
-<!--- TEST ARBITRARY_TIME -->
+To run the upstream flow concurrently with downstream collection, use concurrent processing operators to introduce a buffer.
```
danil-pavlov, line 276:
```suggestion
To run the upstream flow concurrently with a downstream collection, use concurrent processing operators to introduce a buffer.
```

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
-import kotlin.system.*
 
-//sampleStart
-fun simple(): Flow<Int> = flow {
-    for (i in 1..3) {
-        delay(100) // pretend we are asynchronously waiting 100 ms
-        emit(i) // emit next value
+// A simplified custom version of the default .distinctUntilChanged() operator
+fun <T> Flow<T>.myDistinctUntilChanged(): Flow<T> = flow {
+    var lastEmitted: Any? = Any() // a value not equal to anything but itself
```
danil-pavlov:
```suggestion
    var lastEmitted: Any? = Any() // A value that is equal only to itself
```

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
-Intermediate operators are applied to an upstream flow and return a downstream flow.
-These operators are cold. A call to such an operator is not
-a suspending function itself. It works quickly, returning the definition of a new transformed flow.
+Kotlin provides two main kinds of flow operators:
 
-The basic operators have familiar names like [map] and [filter].
-An important difference of these operators from sequences is that blocks of
-code inside these operators can call suspending functions.
+* [**Intermediate operators**](#intermediate-operators) return a new downstream flow that consumes values from upstream flows and applies an operation to them.
+* [**Terminal operators**](#terminal-operators) trigger execution of the flow pipeline by collecting an upstream flow. They may also return a result.
```
danil-pavlov:
```suggestion
* [**Terminal operators**](#terminal-operators) trigger the execution of the flow pipeline by collecting an upstream flow. They may also return a result.
```

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
 
-### Processing the latest value
+One operator that introduces this buffer is the [`.buffer()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/buffer.html) operator.
+It lets you configure the buffer capacity and what happens when the buffer is full.
 
-Conflation is one way to speed up processing when both the emitter and collector are slow. It does it by dropping emitted values.
-The other way is to cancel a slow collector and restart it every time a new value is emitted. There is
-a family of `xxxLatest` operators that perform the same essential logic of a `xxx` operator, but cancel the
-code in their block on a new value. Let's try changing [conflate] to [collectLatest] in the previous example:
+Here's an example:
```
danil-pavlov: I think here and everywhere on this page, instead of a separate paragraph for "Here's an example:" it could be "For example" inside the previous one or just a semicolon to introduce the example

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
-<!--- TEST_NAME FlowGuideTest --> 
 <contribute-url>https://github.com/Kotlin/kotlinx.coroutines/edit/master/docs/topics/</contribute-url>
 
 [//]: # (title: Flow operators)
 
-> We're currently revising this page. For an up-to-date guide to flows, start with the [Flow overview](coroutines-flow.md).
->
-{style="note"}
+Flow operators let you transform and process values in a flow pipeline.
```
danil-pavlov:
```suggestion
```

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
 
-It produces the following three lines, each appearing one second after the previous:
+Intermediate operators can fall into the following categories:
 
-```text                                                                    
-response 1
-response 2
-response 3
-```
+* [**Transforming operators**](#transforming-operators) transform values before emitting them downstream.
```
danil-pavlov, line 25:
```suggestion
* [**Transforming operators**](#transforming-operators) change values before emitting them downstream.
```

danil-pavlov: to avoid repetition

docs/topics/coroutines-flow-operators.md
```diff
@@ -1,994 +1,955 @@
...
-{kotlin-runnable="true" kotlin-min-compiler-version="1.3"}
-<!--- KNIT example-flow-01.kt -->
-> You can get the full code [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/test/guide/example-flow-01.kt).
->
-{style="note"}
+Intermediate operators return a new downstream flow that consumes values from upstream flows.
+You can chain several intermediate operators to build a flow pipeline before collecting the final result.
 
-It produces the following three lines, each appearing one second after the previous:
+Intermediate operators can fall into the following categories:
```
danil-pavlov:
```suggestion
Intermediate operators fall into the following categories:
```

danil-pavlov: Or "can act as", "can be classified into", or just "can be"

danil-pavlov (changes requested):

docs/topics/coroutines-flow-operators.md
daniCsorbaJB: can be classified into works well! 👍 let's go with that
with simply fall into my issue would be that these are not actual defined categories, they act more like as arbitrary categories decided based on what the operator does

docs/topics/coroutines-flow-operators.md
daniCsorbaJB: I don't think it needs the article, but it's grammatically correct either way, so let's go with it. 👍

docs/topics/coroutines-flow-operators.md
daniCsorbaJB: hmm 🤔 I'd prefer the repetition even if it's a a bit repetitive. change would imply to me that we are modifying the value itself which might not always be the case.

docs/topics/coroutines-flow-operators.md
daniCsorbaJB: I changed it where we can. 👍 Somewhere, I think it would make the sentence a bit too heavy with another clause.

docs/topics/coroutines-flow-operators.md
daniCsorbaJB: This would slightly change the meaning 🤔

danil-pavlov (approved):

PR #4704, opened 2026-07-21
KT-87873: suppress warning on android native target family

mMaxy:


buildSrc/src/main/kotlin/kotlin-shared-conventions.gradle.kts
```diff
@@ -86,14 +86,19 @@ plugins.withId("org.jetbrains.kotlin.multiplatform") {
...
 
+        // Deprecated for removal: see KT-86581
+        @Suppress("DEPRECATION", "DEPRECATION_ERROR")
+        run {
+            androidNativeArm32()
+            androidNativeArm64()
+            androidNativeX86()
+            androidNativeX64()
+        }
+
```
dkhalanskyjb:
```suggestion
```

PR #4705, opened 2026-07-21
Triggers for debugger-agent to collect async producer stacktraces

murfel:


kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -398,7 +406,7 @@ private class StateFlowImpl<T>(
                 collectorJob?.ensureActive()
                 // Conflate value emissions using equality
                 if (oldState == null || oldState != newState) {
-                    collector.emit(NULL.unbox(newState))
+                    collector.emit(matchStacktrace(this, NULL.unbox<T>(newState)))
```
murfel, line 409: This bit is potentially buggy for debugger-agent: a thread can read newState and get de-scheduled for a while, and then wake up after many updates to state. Then debugger-agent needs to store all the history of states. That's a lot of memory, so debugger-agent can clean non-current state's stacktraces on timeout, e.g. 1 second.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
```diff
@@ -330,7 +330,15 @@ private class StateFlowImpl<T>(
...
             if (expectedState != null && oldState != expectedState) return false // CAS support
             if (oldState == newState) return true // Don't do anything if value is not changing, but CAS -> true
+
+            // Collect the stacktrace before publishing the new state
+            // because it becomes immediately visible for collectors, and hence can be requested from the debugger.
+            // Also need to indicate the state, because the old state might be requested from the debugger
+            // before this state is published.
+            collectStacktrace(this, NULL.unbox<T>(newState))
             _state.value = newState
+            dropStacktrace(this, NULL.unbox<T>(oldState))
```
murfel, line 340: Given that a collector can fall asleep (see comment on L404), this dropStacktrace should be treated as the beginning of a timeout to clean up the stacktrace.

murfel:
Please treat it as a rough prototype rather than a well thought-through suggestion. Many questions remain unanswered, and some are even not asked yet. Such as:
- channel and channelIterator are not supported in the existing patch, should we support it? If we do, we probably won't be able to breakpoint on `Channel.receive()` unless Idea introduces some new UI for it. Although we can do `Channel.tryReceive()`.
- full enumeration of things that need to be supported.
- when debugger-agent should clean it's memory for each flow/channel type
- how debugger-agent is supposed to fuse stacktraces across modifications
- how is this better (for kx.coroutines) than the existing patch
Etc.

murfel:
I'm just sharing this for visibility for now, feel free to and you are welcome to comment, and I will ping you separately when this will be more ready for a thorough review.

kotlinx-coroutines-core/common/src/flow/StateFlow.kt
dkhalanskyjb: The timeout workaround concerns me: plenty of emissions can happen in one second. It's very realistic to get an out-of-memory using that strategy.

A different proposal: before reading the `_state.value`, the collector registers itself: "Hey, I'm about to read the state, please don't remove any stacktraces for the current value and later ones!", then `matchStacktrace` signals: "All done, you can remove whatever you want now, I'll be busy with this value."

kotlinx-coroutines-core/common/src/internal/DebuggerWrappers.kt
```diff
@@ -0,0 +1,23 @@
...
+
+internal fun collectStacktrace(sharedFlow: SharedFlow<*>, index: Long): Any? = null
+internal fun dropStacktrace(sharedFlow: SharedFlow<*>, index: Long): Any? = null
+internal fun matchStacktrace(sharedFlow: SharedFlow<*>, index: Long): Any? = null
+
+internal fun <T> collectStacktrace(stateFlow: StateFlow<T>, state: T): Any? = null
+internal fun <T> matchStacktrace(stateFlow: StateFlow<T>, state: T): T = state
+internal fun <T> dropStacktrace(stateFlow: StateFlow<T>, state: T): Any? = null
+
+internal fun collectStacktrace(channel: Channel<*>, segment: ChannelSegment<*>, index: Int): Any? = null
```
dkhalanskyjb, line 22: A `ChannelSegment` can't belong to several channels at the same time, so `segment` + `index` is enough to uniquely identify a slot—why is a reference to `Channel` necessary?

kotlinx-coroutines-core/common/src/internal/DebuggerWrappers.kt
```diff
@@ -0,0 +1,23 @@
+@file:Suppress("UNCHECKED_CAST")
+
+package kotlinx.coroutines.internal
+
+import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.flow.*
+
+/**
+ * Used by debugger-agent to support asynchronous stack traces
```
qwwdfsad, line 9: Are there any descriptions/use cases that dictate the shape and applicability of this API?

dkhalanskyjb:
> channel and channelIterator are not supported in the existing patch, should we support it? If we do, we probably won't be able to breakpoint on Channel.receive() unless Idea introduces some new UI for it. Although we can do Channel.tryReceive().

I'm wondering what the semantics of modifying the stack trace for channel iterators and `receive` would even be (I also didn't get the point about `tryReceive`). With `Flow`, it's straightforward: we pretend that the lambda passed to `collect` is invoked from inside the sender's `emit`.

However, why would the stack trace get modified in the following snippet?

```kotlin
val iterator = channel.iterator()
if (iterator.hasNext()) {
    // this is probably the point where the stack trace is already different,
    // since `hasNext` is `suspend` and `next` isn't.
    println(iterator.next())
}
```

We're merely receiving something from another coroutine, and intuitively, our call chain should still look the same. It would be weird for it to include the element sender's stack trace, I believe.

Even worse, what about this?

```kotlin
val iterator1 = channel1.iterator()
val iterator2 = channel2.iterator()
iterator1.hasNext()
iterator2.hasNext()
// what's the stack trace here???
```

Generally, normal function calls don't look like something that should modify stack traces—which is the assumption that `suspend fun`ctions already break routinely, but having `hasNext` modify stack traces looks to me like it would make it even worse.

kotlinx-coroutines-core/common/src/internal/DebuggerWrappers.kt
```diff
@@ -0,0 +1,23 @@
...
+ * in sharedFlow, stateFlow, channelFlow, channel, channelIterator.
+ */
+
+internal fun collectStacktrace(sharedFlow: SharedFlow<*>, index: Long): Any? = null
```
qwwdfsad, line 13: probably similar to the previous question, but more specific -- what is the contract/expectation on `index` (here and everywhere) and on return type?

kotlinx-coroutines-core/common/src/internal/DebuggerWrappers.kt
murfel: debugger-agent stores stacktraces in a WeakHashMap mapping from a channel to whatever is needed, so that entry's debugger-agent memory can be reclaimed once the main code no longer uses a channel.

dkhalanskyjb: It can just as well have a `WeakHashMap` mapping from a `ChannelSegment` to that and get even better guarantees.

murfel:
> With `Flow`, it's straightforward: we pretend that the lambda passed to `collect` is invoked from inside the sender's `emit`.

Hmm, I think of this more like this lambda is still executed by our coroutine, but by the way, here's where this element originated. I get what you mean, and now it makes sense to interpret it this way, just naturally it wasn't my interpretation.

Otherwise, fair point, although that could still make sense for `channel.consume` / `channel.consumeEach`.

I think of `channelFlow`, `SharedFlow`, `StateFlow` flows and channels similarly: it's an asynchronous thing which can bring you elements, where sender and receiver are decoupled. So if you can 'look up' where you element originated on one, it makes sense for me to be able to look it up on all the others.

I believe the current implementation of only enriching flows is due to interpreting decoupled emitter/collector flows (`channelFlow`, `SharedFlow`, `StateFlow`) akin to regular context-preserving flow: it's just a funny-control-flow computation. (And also due to channels being less popular.)

dkhalanskyjb:
> I think of this more like this lambda is still executed by our coroutine

This is a useful observation, because it's consistent with one of the important guarantees of `Flow` (see "Context preservation" here: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/)—so do hold onto this concept!

Just the signatures, though, are more in line with what I'm describing: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow-collector/emit.html The `emit` called from `flow {}` *is* the lambda passed to `collect`.

Combined, we can think of flow execution as executing the lambda passed to `collect` in a hidden `withContext(collectorCoroutineContext) { }` from inside `emit`.

> I think of channelFlow, SharedFlow, StateFlow flows and channels similarly

For me, there are two separate classes of asynchronous APIs:
1) Pull-based (the subscriber pulls values from the publisher): channels, promises, futures, `Deferred`;
2) Push-based (the publisher pushes values onto the subscriber): flows (state flows, shared flows, normal flows), observables, publishers.

Yes, they are similar and often interchangeable—for example, it's even possible to convert between `Channel` and `Flow`. From the point of view of the subscriber, the UX is different between the two. You're either 2) passively listening to the publisher—who runs your callbacks at their own convenience in their call frame of choice—or 1) you're executing your own code that happens to involve receiving a value asynchronously.

kotlinx-coroutines-core/common/src/flow/SharedFlow.kt
```diff
@@ -439,6 +439,7 @@ internal open class SharedFlowImpl<T>(
         if (bufferSize > bufferCapacity) dropOldestLocked()
         // keep replaySize not larger that needed
         if (replaySize > replay) { // increment replayIndex by one
+            dropStacktrace(this, replayIndex)
```
murfel: From @mvicsokolova 
```
/** Note: removed dropStacktrace from here:
* launch {
*   // channel -> flow -> SharedFlow
*   val f3 = channelFlow {
*     send('C')
*   }.shareIn(GlobalScope, SharingStarted.Eagerly)
*     .collect {
*       println(it) // Breakpoint!
*      }
*     }
*  SharedFlowImpl calls dropStacktrace(this, 0) before the collector calls matchStacktrace(this, 0).
*  In this shareIn(..., replay = 0) case, index 0 leaves the replay window immediately,
*  but it is still buffered for an existing slow collector.
*/
```

PR #4706, opened 2026-07-21
update: suggestions for cancellation updates after publishing

daniCsorbaJB:
This PR implements the suggestions for the Cancellation review comments

### DIFF
--- a/docs/topics/coroutines-cancellation.md
+++ b/docs/topics/coroutines-cancellation.md
@@ -83,10 +83,8 @@ println("All coroutines have completed")
 
 In this example, [`CompletableDeferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/) is used as a signal that the coroutine has started running.
 The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed.
-This way, cancellation happens only after the coroutine has started running.
-Without this check, the coroutine may be canceled before it runs the code inside its block.
-You don't need this check to cancel a coroutine. 
-It's included here to make the example reproducible, because the coroutine always starts and prints its messages before it's canceled.
+You don't need this check to cancel a coroutine.
+It's included here to make the example reproducible by ensuring that the coroutine starts and prints its messages before it's canceled.
 
 Because `Deferred` implements `Job`, cancellation works the same way for coroutines created by the `async()` coroutine builder function:
 
@@ -161,7 +159,7 @@ If they're canceled first, nothing is printed.
 In Kotlin, coroutine cancellation is _cooperative_.
 Coroutines react to cancellation only when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
 
-In this section, you can learn how to make coroutines react to cancellation using [suspension points](#suspension-points-and-cancellation), the [`yield()` function](#the-yield-suspending-function), and the [`runInterruptible()` function](#interrupt-blocking-code-when-coroutines-are-canceled) on the JVM.
+In this section, you can learn how adding [suspension points](#suspension-points-and-cancellation), such as calls to the [yield()](#the-yield-suspending-function) function, lets coroutines react to cancellation.
 
 ### Suspension points and cancellation
 
@@ -229,8 +227,9 @@ println("All child jobs completed!")
 
 ### The `yield()` suspending function
 
-Without suspending, coroutines on the same thread run sequentially.
-If a coroutine doesn't suspend for a long time, it doesn't stop when it's canceled.
+If a coroutine doesn't suspend, other coroutines can't run on the same thread until it completes.
+As a result, coroutines that don't suspend run sequentially on that thread.
+If a coroutine doesn't suspend for a long time, it also doesn't stop when it's canceled.
 
 In CPU-intensive computations and other code that runs for a long time without suspending, call the [`yield()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/yield.html) function periodically.
 This function releases the current thread and gives other coroutines a chance to run on it.
@@ -334,7 +333,6 @@ Here's an example:
 ```kotlin
 // Defines a coroutine scope that uses the UI thread
 class ScreenWithButtons(private val scope: CoroutineScope) {
-    // Call this function only from the UI thread
     fun loadAndUpdateButtons(filename: String) {
         scope.launch {
             // withContext() checks for cancellation before entering the block
--- a/docs/topics/coroutines-cancellation.md
+++ b/docs/topics/coroutines-cancellation.md
@@ -6,11 +6,11 @@ Cancellation lets you request to stop a coroutine before it completes.
 It stops work that's no longer needed, such as when a user closes a window or navigates away in a user interface while a coroutine is still running.
 
 You can use cancellation to release resources early and to stop a coroutine from accessing objects past their disposal.
-You can also use it to stop long-running coroutines that perform repeated work, for example: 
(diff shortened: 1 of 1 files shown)


docs/topics/coroutines-cancellation.md
```diff
@@ -76,11 +85,15 @@ In this example, [`CompletableDeferred`](https://kotlinlang.org/api/kotlinx.coro
 The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed.
 This way, cancellation happens only after the coroutine has started running.
 Without this check, the coroutine may be canceled before it runs the code inside its block.
-You don't need this check to cancel a coroutine, but it is included here to make the example reproducible, so the coroutine always starts and prints its messages before it's canceled.
+You don't need this check to cancel a coroutine. 
+It's included here to make the example reproducible, because the coroutine always starts and prints its messages before it's canceled.
```
dkhalanskyjb: The causality feels off. Actually, it's:
* We include the check because we want the example to be reproducible,
* The example is reproducible because the coroutine always starts and prints its messages before it's canceled.

Maybe "because" → "so that" here?

docs/topics/coroutines-cancellation.md
```diff
@@ -141,17 +153,15 @@ parentJob.cancel()
...
 ## Make coroutines react to cancellation {id="cancellation-is-cooperative"}
 
 In Kotlin, coroutine cancellation is _cooperative_.
 Coroutines react to cancellation only when they cooperate by [suspending](#suspension-points-and-cancellation) or [checking for cancellation explicitly](#check-for-cancellation-explicitly).
-There are standard approaches for making coroutines cancelable, such as periodically calling the [`yield()`](#yield-often-in-non-suspending-code) suspending function in CPU-intensive code.
-On the JVM, you can use [`runInterruptible()`](#interrupt-blocking-code-when-coroutines-are-canceled) with some blocking calls that support thread interruption.
 
-Coroutines that react to cancellation are sometimes referred to as **cancelable coroutines**.
-In this section, you will learn how to create cancelable coroutines.
+In this section, you can learn how to make coroutines react to cancellation using [suspension points](#suspension-points-and-cancellation), the [`yield()` function](#the-yield-suspending-function), and the [`runInterruptible()` function](#interrupt-blocking-code-when-coroutines-are-canceled) on the JVM.
```
dkhalanskyjb: This list looks weird: "We are buying groceries, a couple of cucumbers, and a gasoline can".
* `yield()` *is* a suspension point, that's why it reacts to cancellation!
* `runInterruptible()` is a JVM-only solution for a very specific JVM-only problem. It's a common enough issue that it's worth discussing in the tutorial, yet it's also much less widely applicable than the concepts of suspension points and `yield` in particular.

docs/topics/coroutines-cancellation.md
```diff
@@ -217,18 +227,17 @@ println("All child jobs completed!")
 >
 {style="tip"}
 
-### `yield()` often in non-suspending code
+### The `yield()` suspending function
 
+Without suspending, coroutines on the same thread run sequentially.
```
dkhalanskyjb: That's coming out of nowhere at this point, and is really only a narrow case of the problem we're trying to avoid. The more general statement is something like this: "If a coroutine does not suspend, it will be the sole user of a thread until it's completed [and hence, if coroutines don't suspend, they will use a thread one after another in turn, while the coroutines later in line haven't even started]".

docs/topics/coroutines-cancellation.md
```diff
@@ -217,18 +227,17 @@ println("All child jobs completed!")
...
 If a coroutine doesn't suspend for a long time, it doesn't stop when it's canceled.
```
dkhalanskyjb: With the additional text above, wouldn't "it *also* doesn't stop" fit better now?

docs/topics/coroutines-cancellation.md
```diff
@@ -321,38 +334,30 @@ Here's an example:
 ```kotlin
 // Defines a coroutine scope that uses the UI thread
 class ScreenWithButtons(private val scope: CoroutineScope) {
-    // Should only be called from the UI thread
+    // Call this function only from the UI thread
```
dkhalanskyjb: This entire comment (before your changes, as well) is out-of-place here. `scope.launch` can be safely run on any thread, it's safe.

dkhalanskyjb (changes requested):

docs/topics/coroutines-cancellation.md
daniCsorbaJB: 🤔 I see what you mean. 

"so that" works imho, but re-reading this part many times I get the impression that we are introducing redundancy with these sentences. 

Do you think we could shorten this part like this? (essentially cutting out two of the preceding sentences — they feel very similar)

In this example, [`CompletableDeferred`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/) is used as a signal that the coroutine has started running.
The coroutine calls `complete()` when it starts executing, and `await()` only returns once that `CompletableDeferred` is completed.
You don't need this check to cancel a coroutine. 
It's included here to make the example reproducible, so that the coroutine always starts and prints its messages before it's canceled.

docs/topics/coroutines-cancellation.md
daniCsorbaJB: I see it, yes 🤔 — my only intention really is to highlight `yield()` briefly here without going into details. Could we drop `runInterruptible()` from the intro, or should we still add a link to it?

WDYT (add suspension point does it feel mechanical? 🤔 to me it sounds clearer than use:

In this section, you can learn how adding [suspension points](#suspension-points-and-cancellation), such as calls to the [`yield()` function](#the-yield-suspending-function), lets coroutines react to cancellation.
(we could add a sentence about runInterruptible() afterward, but it feels a bit tacked on.)

docs/topics/coroutines-cancellation.md
daniCsorbaJB: let's remove it then 👍

docs/topics/coroutines-cancellation.md
dkhalanskyjb: > In this section, you can learn how adding [suspension points](https://github.com/Kotlin/kotlinx.coroutines/pull/4706#suspension-points-and-cancellation), such as calls to the [yield() function](https://github.com/Kotlin/kotlinx.coroutines/pull/4706#the-yield-suspending-function), lets coroutines react to cancellation.

I see no issues with this option!

docs/topics/coroutines-cancellation.md
dkhalanskyjb: I agree this part does look excessive, and your proposed alternative is much more fluent.

docs/topics/coroutines-cancellation.md
daniCsorbaJB: We can switch this around, WDYT about phrasing it like this? (Since we use "coroutines run sequentially" in the later comment, I think it'd be worthwhile to use the same word to reinforce this. We could ofc change it if it sounds strange ):
 
> If a coroutine doesn't suspend, other coroutines can't run on the same thread until it completes. As a result, coroutines that don't suspend run sequentially on that thread.
If a coroutine doesn't suspend for a long time, it also doesn't stop when it's canceled.

docs/topics/coroutines-cancellation.md
daniCsorbaJB: We could probably tighten it a bit more: 

> It's included here to make the example reproducible by ensuring that the coroutine starts and prints its messages before it's canceled.

dkhalanskyjb: Sure, let's go with that!

docs/topics/coroutines-cancellation.md
dkhalanskyjb: That works, yes.

dkhalanskyjb (approved):

PR #4709, opened 2026-07-23
Add calls-in-place contract to `MutableStateFlow` update functions

JakeWharton:


kotlinx-coroutines-core/common/test/flow/sharing/StateFlowTest.kt
```diff
@@ -244,4 +244,37 @@ class StateFlowTest : TestBase() {
...
+
+    @Test
+    fun testGetAndUpdateContract() = runTest {
+        val state = MutableStateFlow(12)
+        var value: Int
+        state.getAndUpdate {
+            value = it
+            it + 1
+        }
+        assertEquals(12, value)
```
JakeWharton, line 256: The assertion itself is mostly useless, but it serves as a read of `value` which otherwise would be invalid without the contract.

dkhalanskyjb:
Nice find, thank you!

Issue #4215, opened 2024-08-14
`runBlocking` executes other tasks instead of its own


dkhalanskyjb:
**Describe the bug**

A guy promises to an old lady that he'll fix a broken staircase, but then arrives to the house and notices that it's in complete disrepair. He calls his old friend to come and redo the circuitry. He doesn't touch the instruments: "My friend will need them, and after he's done, then I'll fix the staircase." The friend arrives and notices that the house is in complete disrepair. He calls another guy to come and fix a door that's hanging on one hinge. Meanwhile, he doesn't touch the instruments: "Let my friend fix the door first, and then I'll do my job". Repeat this situation enough times, and you get hundreds of people in one room, with no one doing anything. Repeat this more, and the house blows up because of how many people are squeezed in there.

These guys were `runBlocking`. The house is the stack space, and the set of tools is the current thread. The old lady is an unfortunate user of our code.

Reported in https://youtrack.jetbrains.com/issue/KT-66219.

Related to https://github.com/Kotlin/kotlinx.coroutines/issues/3982: if we fix this, the reproducer for #3982 will no longer work, though the general issue will likely still be there.

**Provide a Reproducer**

```kotlin
runBlocking {
    repeat(1000) {
        launch {
            try {
                runBlocking {
                    // do nothing
                }
            } catch (e: Throwable) {
                println(e)
            }
        }
    }
}
```

prints

```
java.lang.NoClassDefFoundError: Could not initialize class kotlin.internal.PlatformImplementationsKt
java.lang.StackOverflowError
java.lang.StackOverflowError
java.lang.StackOverflowError
java.lang.StackOverflowError
java.lang.StackOverflowError
java.lang.StackOverflowError
java.lang.StackOverflowError
```

dkhalanskyjb:
(Excerpts from our design meeting with `@qwwdfsad` and `@zuevmaxim`)

One solution to this is to have `runBlocking` prioritize their own tasks. This can be done by each `runBlocking` adding its own `CoroutineContext` element to the coroutines launched inside it, recognizing them by that element, and giving them priority.

[...]

PR #4711, opened 2026-07-27
Simplify the Native multithreaded dispatchers

dkhalanskyjb:
Before, we used to duplicate the functionality of a `Channel` on top of using a `Channel`.
The algorithm is now simplified, stripping away that layer. Most notably, we no longer need to block the thread in a call to `dispatch` while waiting for a preempted worker thread to start its job. This improves liveness and saves us from the spooky action at a distance in sensitive parts of the codebase (#4215).

### DIFF
--- a/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
+++ b/kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
@@ -87,9 +87,18 @@ private class MultiWorkerDispatcher(
     }
 
     /**
-     * `number of tasks - number of workers`
+     * `number of tasks - number of workers`.
+     *
+     * Used only to determine if we need to allocate a new worker when a new task arrives: if there are idling workers,
+     * we can reuse them.
+     *
+     * It is safe for this variable to overflow.
+     * For [tasksMinusWorkers] to exceed [Int.MAX_VALUE], we need to have at least [Int.MAX_VALUE] tasks enqueued,
+     * each of which requested a worker allocation.
+     * Even in the (practically impossible) race with [workersCount] being [Int.MAX_VALUE],
+     * we will correctly allocate every worker.
      */
-    private val tasksMinusWorkers = atomic(0L)
+    private val tasksMinusWorkers = atomic(0)
 
     private fun workerRunLoop() = runBlocking {
         while (true) {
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -80,67 +80,41 @@ private class MultiWorkerDispatcher(
...
-    private fun obtainWorker(): CancellableContinuation<Runnable> =
-        availableWorkers.tryReceive().getOrNull() ?: runBlocking { availableWorkers.receive() }
-
     override fun dispatch(context: CoroutineContext, block: Runnable) {
-        val state = tasksAndWorkersCounter.getAndUpdate {
-            if (it.isClosed())
-                throw IllegalStateException("Dispatcher $name was closed, attempted to schedule: $block")
-            it + 2
+        if (tasksQueue.isClosedForSend) {
+            // a fast path to avoid touching `tasksMinusWorkers` unnecessarily
```
qwwdfsad: I am not sure it's a fast path, rather the opposite -- the branch that presumably never hits. `onClosed` below should be enough, shouldn't it?

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
```diff
@@ -80,67 +80,41 @@ private class MultiWorkerDispatcher(
...
      */
-    private val tasksAndWorkersCounter = atomic(0L)
-
-    @Suppress("NOTHING_TO_INLINE")
-    private inline fun Long.isClosed() = this and 1L == 1L
-    @Suppress("NOTHING_TO_INLINE")
-    private inline fun Long.hasTasks() = this >= 2
-    @Suppress("NOTHING_TO_INLINE")
-    private inline fun Long.hasWorkers() = this < 0
+    private val tasksMinusWorkers = atomic(0L)
```
qwwdfsad: can now probably be `Int`

qwwdfsad (approved):
Nice simplification!

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
dkhalanskyjb: It's a fast path, just for an uncommon case. I see what you mean, though: every `dispatch` needing to check an atomic variable for an uncommon case isn't great. `onClosed` is enough, yes.

kotlinx-coroutines-core/native/src/MultithreadedDispatchers.kt
dkhalanskyjb: It can, yes.

qwwdfsad (approved):

PR #4712, opened 2026-07-27
Replace AtomicFU atomic arrays with stdlib types

fzhinkin:
AtomicFU's array types are deprecated and barely supported by the compiler plugin. These types could be safely replaced with atomic array types from the stdlib.

### DIFF
--- a/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
+++ b/kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
@@ -2817,7 +2817,7 @@ internal class ChannelSegment<E>(id: Long, prev: ChannelSegment<E>?, channel: Bu
     // ########################################
 
     internal fun storeElement(index: Int, element: E) {
-        setElement(index, element)
+        setElementLazy(index, element)
     }
 
     @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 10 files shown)


kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
```diff
@@ -2823,22 +2830,22 @@ internal class ChannelSegment<E>(id: Long, prev: ChannelSegment<E>?, channel: Bu
     }
 
     private fun setElementLazy(index: Int, value: Any?) {
```
dkhalanskyjb: The name of this function is wrong now. Please rename it to `setElement`.

kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
```diff
@@ -2823,22 +2830,22 @@ internal class ChannelSegment<E>(id: Long, prev: ChannelSegment<E>?, channel: Bu
...
-        data[index * 2].lazySet(value)
+        data.storeAt(index * 2, value)
```
dkhalanskyjb: Optional: could we `expect`/`actual` the `lazySet` as an `inline fun` to avoid a potential performance degradation?

dkhalanskyjb (approved):
The only problem blocking the merge is the "lazy" in the name: either the name could be made accurate, or the function should be renamed.

kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
fzhinkin: https://youtrack.jetbrains.com/issue/KT-88050

dkhalanskyjb: Thanks! What I meant was, though, could we get `lazySet` as an `internal` implementation detail in `kotlinx.coroutines` in the meantime?

fzhinkin: Got you! Sure, I'll add it.

kotlinx-coroutines-core/common/src/channels/BufferedChannel.kt
fzhinkin: Now it's lazy again :)

dkhalanskyjb:
Thanks, well done!

PR #4714, opened 2026-07-31
Consolidate repository settings

fzhinkin:
Moved all repository definitions and configuration to a single settings script and included it from all the projects.

### DIFF
--- a/buildSrc/src/main/kotlin/repositories.settings.gradle.kts
+++ b/buildSrc/src/main/kotlin/repositories.settings.gradle.kts
@@ -51,6 +51,7 @@ dependencyResolutionManagement {
             content {
                 includeGroupByRegex("androidx.*")
                 includeGroupByRegex("com\\.google\\.android.*")
+                includeGroupByRegex("com\\.android\\.tools.*")
             }
         }
 
--- a/buildSrc/src/main/kotlin/repositories.settings.gradle.kts
+++ b/buildSrc/src/main/kotlin/repositories.settings.gradle.kts
@@ -32,19 +32,18 @@ dependencyResolutionManagement {
 
     fun RepositoryHandler.configureJsRepositories() {
         val cacheRedirectorEnabled = System.getenv("CACHE_REDIRECTOR")?.toBoolean() == true
-        if (cacheRedirectorEnabled) {
-            ivy("https://cache-redirector.jetbrains.com/nodejs.org/dist/") {
-                name = "Node Distributions at $url"
-                patternLayout { artifact("v[revision]/[artifact](-v[revision]-[classifier]).[ext]") }
-                metadataSources { artifact() }
-                content { includeModule("org.nodejs", "node") }
-            }
-            ivy("https://cache-redirector.jetbrains.com/github.com/yarnpkg/yarn/releases/download") {
-                name = "Yarn Distributions at $url"
-                patternLayout { artifact("v[revision]/[artifact](-v[revision]).[ext]") }
-                metadataSources { artifact() }
-                content { includeModule("com.yarnpkg", "yarn") }
-            }
+        val cacheRedirectorSuffix = if (cacheRedirectorEnabled) "cache-redirector.jetbrains.com/" else ""
+        ivy("https://${cacheRedirectorSuffix}nodejs.org/dist/") {
+            name = "Node Distributions at $url"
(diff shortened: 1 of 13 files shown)


buildSrc/src/main/kotlin/repositories.settings.gradle.kts
```diff
@@ -0,0 +1,102 @@
...
+        if (cacheRedirectorEnabled) {
+            maven("https://cache-redirector.jetbrains.com/maven-central")
+        } else {
+            mavenCentral()
+        }
+    }
+
+    fun RepositoryHandler.mavenLocal(isSnapshotTrainEnabled: Boolean) {
+        mavenLocal {
+            // If it's not a snapshot build, enable local repo only for kotlin and atomicfu artifacts
```
dkhalanskyjb, line 24: That doesn't seem right. Outside `isSnapshotTrainEnabled`, `mavenLocal` is only used in `integration-testing`, so that the CI can work with both new and old branches (which don't know about `publishAllPublicationsToBuildLocalRepository`).

Since it's been a while since the change to the main artifact publication strategy, I now changed the CI config to reflect that, meaning we can avoid using `mavenLocal` when not using the train.

buildSrc/src/main/kotlin/repositories.settings.gradle.kts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fun RepositoryHandler.gradlePortalWithCacheRedirect() {
+        val cacheRedirectorEnabled = System.getenv("CACHE_REDIRECTOR")?.toBoolean() == true
+        if (cacheRedirectorEnabled) {
+            maven("https://cache-redirector.jetbrains.com/maven-central")
+        } else {
+            mavenCentral()
+        }
+    }
```
dkhalanskyjb, line 20: Shouldn't the names be the opposite?

buildSrc/src/main/kotlin/repositories.settings.gradle.kts
```diff
@@ -0,0 +1,102 @@
...
+
+        google {
+            content {
+                includeGroupByRegex("androidx.*")
+                includeGroupByRegex("com\\.google\\.android.*")
+                includeGroupByRegex("com\\.android.*")
+            }
+        }
+
+        mavenLocal(build_snapshot_train?.toBoolean() == true)
```
dkhalanskyjb, line 75: This logic isn't one-to-one with the old one: it used to be the case that *either* `build_snapshot_train` *or* any dependency version being `SNAPSHOT` caused local Maven to be used. I'm assuming this was done to support easily trying out new `atomicfu` versions locally in `kotlinx.coroutines`.

I personally have no use for this logic, so I'm okay with seeing it go.

buildSrc/src/main/kotlin/repositories.settings.gradle.kts
```diff
@@ -0,0 +1,102 @@
...
+                content {
+                    includeVersionByRegex(".*", "(kotlin|atomicfu)", ".*-SNAPSHOT")
+                }
+            }
+        }
+    }
+
+    fun RepositoryHandler.configureJsRepositories() {
+        val cacheRedirectorEnabled = System.getenv("CACHE_REDIRECTOR")?.toBoolean() == true
+        val cacheRedirectorSuffix = if (cacheRedirectorEnabled) "cache-redirector.jetbrains.com/" else ""
```
dkhalanskyjb, line 35: "Prefix"?

buildSrc/src/main/kotlin/repositories.settings.gradle.kts
```diff
@@ -0,0 +1,102 @@
...
+            // If it's not a snapshot build, enable local repo only for kotlin and atomicfu artifacts
+            if (!isSnapshotTrainEnabled) {
+                content {
+                    includeVersionByRegex(".*", "(kotlin|atomicfu)", ".*-SNAPSHOT")
+                }
+            }
+        }
+    }
+
+    fun RepositoryHandler.configureJsRepositories() {
```
dkhalanskyjb, line 33: This function is only used in this file and can be either inlined or made private.

dkhalanskyjb (changes requested):
Having a script that gets inlined into various places in the build doesn't look nice. After all, that isn't a proper abstraction but, basically, text expansion. The wins in this specific case are undeniable, though. Well done!

PR #4715, opened 2026-08-03
Avoid costly stacktrace element initialization on the DebugProbes hot path

qwwdfsad:
Use a hack to detect the type of the underlying continuation instead. It yields twice the performance and avoids reflective lookup to 'label' Field in DebugMetadataKt (KT-57634)

### DIFF
--- a/benchmarks/src/jmh/kotlin/benchmarks/debug/RedundantMaterializationOnResumeBenchmark.kt
+++ b/benchmarks/src/jmh/kotlin/benchmarks/debug/RedundantMaterializationOnResumeBenchmark.kt
@@ -1,0 +1,60 @@
+package benchmarks.debug
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.debug.*
+import org.openjdk.jmh.annotations.*
+import org.openjdk.jmh.annotations.State
+import java.util.concurrent.*
+import java.util.concurrent.atomic.AtomicInteger
+
+// IJPL-251805, KT-57634
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(value = 1)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+open class RedundantMaterializationOnResumeBenchmark {
+
+    @Volatile
+    private var sink = 42 // prevent TCO
+
+    @Param("1", "10", "50")
+    private var depth: Int = 1
+
+    @Param("true", "false")
+    var withDebugger = false
+
+    @Setup
+    fun setup() {
+        DebugProbes.sanitizeStackTraces = false
+        DebugProbes.enableCreationStackTraces = false
+        if (withDebugger) {
+            DebugProbes.install()
+        }
+    }
+
+    @Benchmark
+    fun suspendFewTimes() = runBlocking {
+        repeat(100) {
+            recursion(depth)
+            sink = 239
+        }
+    }
+
+
+    suspend fun recursion(depth: Int) {
+        if (depth - 1 == 0) {
+            stateMachine()
+            sink = 42
+            return
+        }
+        recursion(depth - 1)
+        sink = 42
+    }
+
+    private suspend fun stateMachine() {
+        yield()
+        sink = 42
+    }
+}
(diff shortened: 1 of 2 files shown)


qwwdfsad:
(context: was passing by the internal discussion around https://youtrack.jetbrains.com/issue/IJPL-251805/Terminal-CPU-usage-spikes-at-100-and-kills-other-processes)

benchmarks/src/jmh/kotlin/benchmarks/debug/RedundantMaterializationOnResumeBenchmark.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+// IJPL-251805, KT-57634
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(value = 1)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+open class RedundantMaterializationOnResumeBenchmark {
```
qwwdfsad, line 17: Side note: ping pong on CHM#add/remove on this particular benchmark is very significant. I have an idea how to improve that -- use `Job` as a key instead, avoid removals (->tombstones), should be much faster. Will create a separate PR

benchmarks/src/jmh/kotlin/benchmarks/debug/RedundantMaterializationOnResumeBenchmark.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @Benchmark
+    fun suspendFewTimes() = runBlocking {
+        repeat(100) {
+            recursion(depth)
+            sink = 239
+        }
+    }
+
+
```
dkhalanskyjb: Style: extra newline.

dkhalanskyjb (changes requested):
Good job noticing that the full result `getStackTraceElement()` isn't used, and that obtaining a partial result can be optimized!

We can live with the proposed solution *as a short-term workaround*, but in the long run, there doesn't seem to be a reason not to do this instead:
```kotlin
public interface CoroutineStackFrame {
    public val callerFrame: CoroutineStackFrame?
    public fun getStackTraceElement(): StackTraceElement?
    public val hasStackTraceElement: Boolean
        get() = getStackTraceElement() != null
}
```
(With `BaseContinuationImpl` overriding `hasStackTraceElement` with `true` and `realCaller` checking `hasStackTraceElement`).
The other implementations of `CoroutineStackFrame` can have a more efficient implementation of `hasStackTraceElement`, too (like `TailCallBaseContinuationImpl` always returning `true`).

If you agree, let's figure out which issue we'll use to track the progress with `hasStackTraceElement` and drop the usual comment like `// TODO: workaround for KT-XXXXX`. Then, the PR will be good to go.

dkhalanskyjb (approved):

kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
```diff
@@ -466,7 +467,18 @@ internal object DebugProbesImpl {
...
+         * getStackTraceElement() reflectively reads `label` from the compiler-generated continuation
+         * which tends to be unreasonably slow (KT-57634) because of reflection overhead. We don't need the line number
+         * here or even a stacktrace element, we need only the marker it exists.
+         * By its contract, BaseContinuationImpl always has one, so it's much (~x2) cheaper to check for the type
+         * rather than materializing StackTraceElement.
+         *
+         * Can be removed when KT-43395 is implemented and widely adopted.
+         */
+        @Suppress("INVISIBLE_REFERENCE")
+        return if (caller is BaseContinuationImpl || caller.getStackTraceElement() != null) caller else caller.realCaller()
```
murfel, line 481: Claude suggested that the set of things for which this returns true has increased, like `SafeCollector` implements `ContinuationImpl` and overrides `getStackTraceElement` to return null. I can't understand why it's irrelevant, could you clarify this for my personal understanding? `SafeCollector` can't be the receiver here because of something?

murfel (approved):

kotlinx-coroutines-core/jvm/src/debug/internal/DebugProbesImpl.kt
qwwdfsad: `SafeCollector` is a very obscure hack (see the KDoc for its JVM declaration) that pretends to be a valid `ContinuationImpl` for re-use. It is indeed a `ContinuationImpl` with an empty `gSTE`, but it doesn't matter for correctness here: this method is only needed for `callerInfoCache` (to avoid accidental quadratic behaviour documented in `callerInfoCache`) to speedup the lookup and uses who-is-first-in-real-call-frame logic

PR #4718, opened 2026-08-10
Make JS iteration over ReceiveChannel cancel iteration on early exits

dkhalanskyjb:
Also, mark `asyncIterator` as `ExperimentalCoroutinesApi`.

### DIFF
--- a/kotlinx-coroutines-core/js/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/js/src/channels/Channel.kt
@@ -23,7 +23,7 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
      * Returns a JavaScript `AsyncIterator` for this channel.
      *
      * This method is used to implement the JavaScript async-iteration protocol, so that a
-     * `ReceiveChannel` exported to JavaScript can be consumed with `for await ... of`.
+     * `ReceiveChannel` exported to JavaScript can be [consumed][Channel.consume] with `for await ... of`.
      *
      * Each call to the iterator's `next` method receives at most one element from this channel:
      *
@@ -35,15 +35,18 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
      *   cause.
      *
      * Calling the iterator's `return` method finishes this iterator instance and returns a fulfilled
-     * `Promise` with `done` set to `true`. It does not cancel the underlying channel.
+     * `Promise` with `done` set to `true`. It [cancels][ReceiveChannel.cancel] the channel without a `cause`.
      *
      * Calling the iterator's `throw` method finishes this iterator instance and returns a rejected
-     * `Promise` with the supplied error. It does not cancel the underlying channel.
+     * `Promise` with the supplied error. It [cancels][ReceiveChannel.cancel] the channel
+     * with the `cause` of the [CancellationException] being set to the exception provided to `throw`.
      *
      * The coroutines backing calls to `next` are started in [GlobalScope].
      * In particular, they are not children of any caller-provided coroutine
      * scope and therefore are not bound to the lifetime of any structured-concurrency scope.
      */
+    @OptIn(ExperimentalWasmJsInterop::class)
+    @ExperimentalCoroutinesApi
     override fun asyncIterator(): JsAsyncIterator<E> {
         var wasEarlyFinished = false
         return JsAsyncIterator(
@@ -65,10 +68,16 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
             // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#description
             `return` = { value: E? ->
                 wasEarlyFinished = true
+                cancel()
                 Promise.resolve(JsIteratorResult(value = value, done = true))
             },
             `throw` = { err: dynamic ->
                 wasEarlyFinished = true
+                val cause = err.unsafeCast<JsPromiseError>().toThrowableOrNull()
+                /** Adapted from [ReceiveChannel.cancelConsumed] */
+                cancel(cause?.let {
+                    it as? CancellationException ?: CancellationException("Channel was closed via AsyncIterator#throw method", it)
+                })
                 Promise.reject(err)
             }
         )
--- a/kotlinx-coroutines-core/js/src/channels/Channel.kt
+++ b/kotlinx-coroutines-core/js/src/channels/Channel.kt
@@ -1,5 +1,5 @@
 @file:OptIn(ExperimentalJsExport::class, ExperimentalStdlibApi::class)
-@file:Suppress("EXPOSED_FUNCTION_RETURN_TYPE", "INVISIBLE_REFERENCE", "EXPOSED_SUPER_INTERFACE")
+@file:Suppress("EXPOSED_FUNCTION_RETURN_TYPE", "INVISIBLE_REFERENCE", "EXPOSED_SUPER_INTERFACE", "EXPOSED_PARAMETER_TYPE")
 package kotlinx.coroutines.channels
 
 import kotlinx.coroutines.*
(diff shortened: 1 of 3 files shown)


JSMonk:
@dkhalanskyjb as we discussed in #4625 I've added methods for the `Channel` to configure the cancellation behavior during iteration. I've introduced a few overloads that are hidden in either JS or Kotlin to provide a natural API for both languages.

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb, line 13: Note that `ReceiveChannel` is still `JsAsyncIterable`. We have to choose a single entry point: either the channel itself is iterable, or the users go through a separate function. Having both defeats the purpose of the function to educate the user about the channel cancellation behavior.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -20,10 +20,123 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
...
+     * When iteration exits early (for example via loop `break`/`return`/`throw`, or by calling iterator methods `return`/`throw` directly),
+     * the channel is canceled by default.
+     *
+     * @param options iteration behavior options:
+     * - `preventCancel = true`: early iterator completion does not cancel the channel;
+     * - `preventCancel = false` or omitted: early iterator completion cancels the channel.
+     */
+    @ExperimentalCoroutinesApi
+    // We can't use DeprecationLevel.HIDDEN, because the generated declaration will also be deprecated in .d.ts
+    @LowPriorityInOverloadResolution
```
dkhalanskyjb, line 61: I don't think this does anything. This overload is never in conflict with the other one.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -1,5 +1,5 @@
 @file:OptIn(ExperimentalJsExport::class, ExperimentalStdlibApi::class)
-@file:Suppress("EXPOSED_FUNCTION_RETURN_TYPE", "INVISIBLE_REFERENCE", "EXPOSED_SUPER_INTERFACE")
+@file:Suppress("EXPOSED_FUNCTION_RETURN_TYPE", "INVISIBLE_REFERENCE", "EXPOSED_SUPER_INTERFACE", "EXPOSED_PARAMETER_TYPE")
```
dkhalanskyjb, line 2: `ChannelIteratorOptions` being `internal` but visible as an parameter of a `public` overload is not acceptable. I tried using this from another Kotlin/JS project, and on the Kotlin side, the IDE does not hide the overload that accepts `ChannelIteratorOptions`, doesn't show that it can't be constructed, and in general, behaves as if the overload is entirely valid. This can lead users on a wild goose chase.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -20,10 +20,123 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
...
+     *
+     * When iteration exits early (for example via loop `break`/`return`/`throw`, or by calling iterator methods `return`/`throw` directly),
+     * the channel is canceled by default.
+     *
+     * @param options iteration behavior options:
+     * - `preventCancel = true`: early iterator completion does not cancel the channel;
+     * - `preventCancel = false` or omitted: early iterator completion cancels the channel.
+     */
+    @ExperimentalCoroutinesApi
+    // We can't use DeprecationLevel.HIDDEN, because the generated declaration will also be deprecated in .d.ts
```
dkhalanskyjb, line 60: Alright, does `DeprecationLevel.ERROR` work?

If not, we'll have to settle for just the boolean overload and remove this one. Getting people relying on autocompletion on the Kotlin side into a trap where they try and fail to construct a `ChannelIteratorOptions` object would be a major usability issue.

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb, line 13: With the new function(s), we need to reevaluate `ReceiveChannel` implementing `JsAsyncIterable` directly.

Since both possible `AsyncIterable` behaviors on exit—cancelling or not cancelling—may surprise a subset of users, and we provide a function controlling this, we'd like to emphasize that there's this variability in behavior and encourage checking the options. If someone just writes `for await (const v of channel)`, nothing here indicates a non-trivial choice being made.

I realize `ReadableStream` does this as well—it's both iterable and has a `values` method—but as you pointed out, its default cancellation behavior is more obvious, due to it being exclusive to a single consumer.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -20,10 +20,123 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
...
+     * scope and therefore are not bound to the lifetime of any structured-concurrency scope.
+     *
+     * @param options iteration behavior options:
+     * - `preventCancel = true`: early iterator completion does not cancel the channel;
+     * - `preventCancel = false` or omitted: early iterator completion cancels the channel.
+     */
+    @ExperimentalCoroutinesApi
+    // We can't use DeprecationLevel.HIDDEN, because the generated declaration will also be deprecated in .d.ts
+    @LowPriorityInOverloadResolution
+    @JsName("values") // We use "values" here to mimic the ReadableStream API: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream
```
dkhalanskyjb, line 131: Is `ReadableStream.values` an `AsyncIterator` or an `AsyncIterable`? I couldn't easily find this in the documentation.

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -20,10 +20,123 @@ public actual interface ReceiveChannel<out E> : JsAsyncIterable<E> {
...
+     * with the `cause` of the [CancellationException] being set to the exception provided to `throw`.
+     *
+     * The coroutines backing calls to `next` are started in [GlobalScope].
+     * In particular, they are not children of any caller-provided coroutine
+     * scope and therefore are not bound to the lifetime of any structured-concurrency scope.
+     *
+     * @param options iteration behavior options:
+     * - `preventCancel = true`: early iterator completion does not cancel the channel;
+     * - `preventCancel = false` or omitted: early iterator completion cancels the channel.
+     */
```
dkhalanskyjb, line 127: Since we don't expect Kotlin users to call this function, we shouldn't generate an API reference page for it, so please `/** @suppress */` it.

dkhalanskyjb (commented):
I can't request changes for my own pull request, which complicates the review. Please consider making another branch on top of mine and opening a separate PR. No need to bother making it pretty, it's only for code review.

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: As far as I remember, [we've agreed](https://github.com/Kotlin/kotlinx.coroutines/pull/4625#discussion_r3801943488) on mimicking the JS API, such as:
- [ReadableStream in browser](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#async_iteration)
- [Stream API in Node.js](https://nodejs.org/api/stream.html#readableiteratoroptions)
- [ReadableStream in Deno](https://docs.deno.com/api/web/streams/#ReadableStream)
- [Stream API from Bun](https://bun.com/docs/runtime/streams)

Each representative of such an API provides a default Symbol.iterator (meaning it is an AsyncIterable) and an additional function for configuring the closing/cancellation behavior. So, I propose keeping it as is (per the previously agreed API).

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: I believe on the IDE level it will show the overload at the end of the overloads list, isn't it? (so it's a more UX related addition)

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: It's an intentional change. 
1. Making the `ChannelIteratorOptions` `internal` prevents using the API with the options in Kotlin.
2. Yes, unfortunately, this is the design problem we have with the `@Deprecated` annotation. I believe we will fix it soon, so that `LowPriorityInOverloadResolution` can be replaced with `DeprecationLevel.HIDDEN`, and it will not cause a TypeScript-side deprecation. But since there is currently no better solution (or is there any?), I propose keeping it as is.

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: The issue with the `@Deprecated` annotation is that, on the TypeScript side, the declaration will also be marked as `@deprecated` (despite the level specified for the annotation). Right now, there is no API that could help us to hide some member on the Kotlin side, but keep it on the JS/TS side (we're working on it, to solve it at least internally)

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: It's an `AsyncIterator` in Node.js: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8443606b19593c505271bda9ad1238e10a2ef78a/types/node/stream/web.d.ts#L187

And it's both `AsyncIterator` and `AsyncIterable` in the browser: https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/019469b3d455389782e34584554d9e54b74d9c2b/baselines/ts5.5/dom.asynciterable.generated.d.ts#L14

JSMonk: So, possibly, we can drop `asAsyncIterable` and mimic browser behavior, so that it returns an iterator which is at the same time `AsyncIterable` (minimizing the API)

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: No, I don't see any such effect.
image

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: In that case, the new `values` function is pointless, though. No one will need to call it if `Channel` is `JsAsyncIterable` already. Our goal to educate people about `Channel` being cancellable will not be reached, and the debugging pitfall that you expect will still be present.

On the other hand, if we accept `Channel` as `JsAsyncIterable` without any converters, we can remove the proposed `asAsyncIterable`/`values` functions for now, as the language is clearly not ready to handle them properly. Once the compiler learns to expose a function to TypeScript but not Kotlin, we can revisit this, see if there's actually user demand for iterating over a `Channel` without cancelling it on JS, and if so, add the functions. Sounds good to me!

kotlinx-coroutines-core/js/src/channels/Channel.kt
dkhalanskyjb: I understand that the change is intentional, my point is that we can't accept this change with how broken it is on the Kotlin side.

kotlinx-coroutines-core/js/src/channels/Channel.kt
JSMonk: Interesting 🤔
Let me think more about what options we have to hide the member (or at least put it at the end of the overload list

JSMonk: Even uglier, but more efficient (as I said, it would be a temporary solution until we add proper support for DeprecationLevel.HIDDEN for the JsExport and 2.5.0 becomes more widely used) is to add on the Kotlin side some escaped (or two underscores) name and the @JsName annotation, so that the declaration is visible but not quite usable.

Or another thing that came to mind is to use an opt-in with an internal class (does that make any sense?)

JSMonk: I would like to combine this thing: ugly name + JsName + internal opt-in.
After 2.5.0 becomes usable (maybe around 2.6.20), we would switch it to @Deprecated with DeprecationLevel.HIDDEN (since we are going to support it in 2.5.0)

dkhalanskyjb: That could work, though an ugly name is still not invisible, which is the downside of this approach. Is there a specific reason why we need to add `.values` as soon as possible and can't wait for 2.5.0 instead?

JSMonk: I don't have any forcing reasons.
Maybe except the possibility that users would need this at some point, and to have it, they need to update to a specific version of the compiler (but it's a small downside)

JSMonk: Okay, somehow, I managed to make it work with the DeprecationLevel.HIDDEN and an empty message

kotlinx-coroutines-core/js/src/channels/Channel.kt
```diff
@@ -148,6 +240,24 @@ internal external interface JsAsyncIterator<out T> {
     public val `throw`: (value: Any?) -> Promise<JsIteratorResult<T>>
 }
 
+/**
+ * Options for customizing channel async-iteration behavior.
+ */
+@JsExport
+@Deprecated(message = "", level = DeprecationLevel.HIDDEN)
```
dkhalanskyjb: Please drop a comment explaining why the deprecation message is intentionally empty. Otherwise, we may forget about this and add some text here in the future for readability.

JSMonk (approved):

PR #4721, opened 2026-08-11
Enable Configuration Cache 

qwwdfsad:
It speeds up the rerun of a single test by a whopping second-point-something, making it sub-second on my machine and much more pleasant to run any tests.

I was hoping this is 20 minute adventure, but here we are: I suggest reviewing it commit-by-commit (they are self-contained and incremental). Publication is byte-to-byte compatible, Kover, Dokka, `jmhJar` and animalsniffer can be run without any violations

### DIFF
--- a/buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/dokka-conventions.gradle.kts
@@ -14,11 +14,13 @@ configure(subprojectWithDokka) {
     configureDokkaPlugins()
     configureDokkaTemplatesDir()
     condigureDokkaSetup()
+    configureDokkaGeneratorMemory()
 }
 
 // For top-level multimodule collection
 configureDokkaPlugins()
 configureDokkaTemplatesDir()
+configureDokkaGeneratorMemory()
 
 dependencies {
     subprojectWithDokka.forEach {
@@ -49,6 +51,14 @@ private fun Project.condigureDokkaSetup() {
                 localDirectory = rootDir
                 remoteUrl("https://github.com/kotlin/kotlinx.coroutines/tree/master")
             }
+        }
+    }
+}
+
+private fun Project.configureDokkaGeneratorMemory() {
+    dokka {
+        dokkaGeneratorIsolation = ProcessIsolation {
+            maxHeapSize = "1g"
         }
     }
 }
--- a/buildSrc/src/main/kotlin/kover-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kover-conventions.gradle.kts
@@ -10,7 +10,7 @@ val expectedCoverage = mutableMapOf(
     // These have lower coverage in general, it can be eventually fixed
     "kotlinx-coroutines-swing" to 70, // awaitFrame is not tested
     "kotlinx-coroutines-javafx" to 35, // JavaFx is not tested on TC because its graphic subsystem cannot be initialized in headless mode
-
+    "kotlinx-coroutines-debug" to 50, // TODO: temp for cc
     // Reactor has lower coverage in general due to various fatal error handling features
     "kotlinx-coroutines-reactor" to 75
 )
--- a/gradle.properties
+++ b/gradle.properties
@@ -41,7 +41,7 @@ baksmali_version=2.2.7
 kotlin.incremental.multiplatform=true
 kotlin.native.ignoreDisabledTargets=true
 
-org.gradle.jvmargs=-Xmx3g
+org.gradle.jvmargs=-Xmx3g -XX:+UseParallelGC
 
 kotlinx.atomicfu.enableJvmIrTransformation=true
 kotlinx.atomicfu.enableJsIrTransformation=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -41,9 +41,9 @@ baksmali_version=2.2.7
 kotlin.incremental.multiplatform=true
 kotlin.native.ignoreDisabledTargets=true
 
-org.gradle.jvmargs=-Xmx3g -XX:+UseParallelGC
+org.gradle.jvmargs=-Xmx3g -XX:+UseParallelGC \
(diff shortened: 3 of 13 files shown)


qwwdfsad:
Also, if you are debugging tests via IJ often enough, you might want to disable coroutines debugger in IJ, otherwise switches between run and debug will be invalidating cc with `Calculating task graph as configuration cache cannot be reused because environment variable 'DEBUGGER_ENABLED' has changed.`

dkhalanskyjb:
tl;dr: it's broken now, but may be worth it? Please fix the CI build (setting a larger `-Xmx` may suffice?), and we'll be in a better position to judge.

1. I can confirm there are some performance improvements when rerunning a single test, using the invocation `./gradlew --no-build-cache :kotlinx-coroutines-core:cleanJvmTest :kotlinx-coroutines-core:jvmTest --tests '**testCancellingWhenInterrupted'`.

* `configuration-cache`, rerunning a test: 1 second.
* `configuration-cache`, rerunning a test after adding/removing a newline: 2 seconds.
* `develop`, rerunning a test: 2 seconds.
* `develop`, rerunning a test after adding/removing a newline: 2 seconds.

All in all, the wins are marginal. The only tests I'm running without recompiling are stress test, and their own execution time dominates whatever gains this presents.

2. In exchange, the build now requires more memory. On my machine, `./gradlew clean check` fails after consuming the JVM heap space, unless I raise the `-Xmx` parameter in `gradle.properties`. On TeamCity, the build fails with some weird errors, which could also be caused by an OOM. Worth checking.

3. However! For some reason, enabling the configuration cache makes tasks run in parallel for `./gradlew --no-build-cache clean check`.
`linuxX64Test` is running in parallel with `linuxX64WorkerTest` and `jvmTest`, making the build finish a lot faster (4 minutes with 8g heap memory + the cache enabled, 14 minutes with 8g heap memory + no configuration cache). This side effect is odd. On the part of the library itself, there should be no state leakage between different processes, so it may be fine.

4. The output from the tasks being interleaved is unpleasant. Example:

```
DebugProbesTest > testAsync PASSED

DebugProbesTest > testAsyncWithSanitizedProbes PASSED

DebugProbesTest > testAsyncWithProbes PASSED

DebugProbesTest > testMultipleConsecutiveProbeResumedAndLaterRunning PASSED

DebugProbesTest > testMultipleConsecutiveProbeResumed PASSED

(node:2924195) ExperimentalWarning: WASI is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

> Task :kotlinx-coroutines-reactive:test

PublisherCompletionStressTest > testCompletion PASSED

PublisherMultiTest > testConcurrentStress PASSED

PublisherMultiTest > testConcurrentStressOnSend PASSED

w: [:kotlinx-coroutines-jdk9:dokkaGenerateModuleHtml] Couldn't resolve link: [../kotlinx-coroutines-reactive] in module documentation (:kotlinx-coroutines-jdk9/main)

> Task :kotlinx-coroutines-rx3:test

IntegrationTest > testNumbers[ctx=UNCONFINED, delay=true] PASSED

LeakedExceptionTest > testFlowable PASSED

> Task :android-unit-tests:test

CustomizedRobolectricTest > testComponent PASSED

CustomizedRobolectricTest > testComponentAfterReset PASSED

FirstMockedMainTest > testComponent PASSED
```

whyoleg:
> However! For some reason, enabling the configuration cache makes tasks run in parallel for ./gradlew --no-build-cache clean check.

That's expected behavior, and one of the cool features of configuration cache!
Without it, Gradle can only run tasks in parallel from different projects when `org.gradle.parallel=true` ([docs](https://docs.gradle.org/current/userguide/performance.html#sec:enable_parallel_execution)). Or, if the workers' machinery is used during task execution ([docs](https://docs.gradle.org/current/userguide/worker_api.html)), e.g., kotlin compilation or Dokka execution uses workers, those could be parallelized even inside a single project, even without configuration cache.
The parallelism for both of those is controlled by the `org.gradle.workers.max` property (Default is the number of CPU processors.)

But with the configuration cache, Gradle can execute tasks in parallel within a single project, even if it doesn't use workers! ([docs](https://docs.gradle.org/current/userguide/configuration_cache.html#parallel_task_execution))

Test tasks are one of them, publication is another, and there are definitely more tasks that could be parallelized with it!

> The output from the tasks being interleaved is unpleasant.

This could also happen now, with tasks from different workers, or if you enable `org.gradle.parallel=true`. As far as I remember, there were several improvements to test reporting in the latest Gradle versions, so it might be worth updating to the latest Gradle version and rechecking it.

Also, there is an option that could further reduce configuration time: `org.gradle.configuration-cache.parallel=true`. This will also execute the configuration load/store phase in parallel. ([docs](https://docs.gradle.org/current/userguide/configuration_cache_enabling.html#config_cache:usage:parallel))
It's incubating, but it works fine in [cryptography-kotlin](https://github.com/whyoleg/cryptography-kotlin) at my side for more than a year.

---

While I understand that the end goal of this PR is to improve "edit-build-test" time (and for this, CC is really helpful!), you might also want to explore enabling parallelism (`org.gradle.parallel=true`) to improve CI time or the full build.
In the end, here are all the flags you need to enable to be _as fast as possible_ (for the cost of memory usage):
```properties
org.gradle.parallel=true
org.gradle.caching=true
org.gradle.configuration-cache=true
org.gradle.configuration-cache.parallel=true
```

I would also suggest updating to the latest Gradle version anyway, as it frequently improves things. I could help here, if you want, just say it. :)

qwwdfsad:
@whyoleg thanks! That was the plan: merge https://github.com/Kotlin/kotlinx.coroutines/pull/4723, rebase on it, then discuss the performance impact (which is a net win IMO).

>org.gradle.configuration-cache.parallel=true

I've tried this one, but the speedup is negligible, and I am not sure we don't have an implicit ordering dependency (i.e. we have plenty `subprojects` and `allprojects` in the build), so this one I omitted

qwwdfsad:
Experimenting with memory-constrained CI, will figure out, properly reorder commits and make it ready for review

PR #4726, opened 2026-08-13
Join backgroundScope after cancel in runTest

### LINKED ISSUE #4685: [Bug] `backgroundScope` coroutines can trigger `UncaughtExceptionsBeforeTest`
### What is happening? What should have happened instead?

Currently, the `backgroundScope` in a `runTest` block is cancelled, but may leave coroutines running.
https://github.com/Kotlin/kotlinx.coroutines/blob/8564f65764d3d05893cec026c6e94250e2b23874/kotlinx-coroutines-test/common/src/TestBuilders.kt#L369

I believe its job should be `.join()`ed as well. Otherwise, subsequent tests can throw `kotlinx.coroutines.test.UncaughtExceptionsBeforeTest`.

### Reproducer

See https://pl.kotl.in/okOqxIhaC
```kotlin
import kotlinx.coroutines.*
import kotlinx.coroutines.test.*

fun main() {
   runTest {
       backgroundScope.launch(Dispatchers.Default) {
           withContext(NonCancellable) {
               delay(1_000)
               error("hmm")
           }
       }

       delay(50)
   }

   Thread.sleep(1_500)

   runTest {
       println("I'm innocent")
   }
}
```

Code like this can occasionally crash with `kotlinx.coroutines.test.UncaughtExceptionsBeforeTest`, although I haven't been able to reproduce it in the playground.

### Anything else? (optional)

I originally ran into this with a Ktor server launched in the `backgroundScope`  when running tests in-process via the Mill build system, using isolated classloaders for the tests themselves but a shared classloader for ~clean dependencies, including `kotlinx-coroutines-test`. This led to a race between the Ktor server shutting down and Mill unloading the isolated classloader for the given test. When Mill was faster, the leftover coroutines in `backgroundScope` started running into `NoClassDefFoundError`s, in turn triggering `UncaughtExceptionsBeforeTest` with the _next_ `runTest` invocation, whatever that may have been.

AzazelSensei:
cancel() plus advanceUntilIdle does not wait for NonCancellable work on Dispatchers.Default. That job then finishes after the test and the next runTest throws UncaughtExceptionsBeforeTest.

I join the background job after draining the test scheduler, so real-dispatcher leftovers are collected in the same test.

Fixes #4685

### DIFF
--- a/kotlinx-coroutines-test/common/src/TestBuilders.kt
+++ b/kotlinx-coroutines-test/common/src/TestBuilders.kt
@@ -368,6 +368,7 @@ public fun TestScope.runTest(
         } finally {
             backgroundScope.cancel()
             testScheduler.advanceUntilIdleOr { false }
+            backgroundScope.coroutineContext[Job]?.join()
             val uncaughtExceptions = scope.leave()
             throwAll(timeoutError ?: scope.getCompletionExceptionOrNull(), uncaughtExceptions)
         }
@@ -400,6 +401,7 @@ public fun TestScope.runTest(
         runTestCoroutineLegacy(it, dispatchTimeoutMs.milliseconds, TestScopeImpl::tryGetCompletionCause, testBody) {
             backgroundScope.cancel()
             testScheduler.advanceUntilIdleOr { false }
+            backgroundScope.coroutineContext[Job]?.join()
             it.legacyLeave()
         }
     }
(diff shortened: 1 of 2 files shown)


AzazelSensei:
No maintainer in sight. Closing so this doesn't sit here forever.

PR #4727, opened 2026-08-14
Make kotlinx.coroutines.debug timeout tests deterministic

dkhalanskyjb:
Introduced a simple lightweight alternative to VirtualTimeSource or TestCoroutineScheduler specifically for testing the JUnit 4/5 extensions for dumping coroutine state on a timeout.

We could not easily reuse TestCoroutineScheduler because the code awaiting the timeout isn't (and shouldn't be!) running in a coroutine.
That would mean polluting the coroutine dumps,
which we are checking.

VirtualTimeSource could have been used, allowing us to write `delay` instead of `fakeDelay`. However, that would have required either sending tasks to the `DefaultExecutor` or writing our own copy of that based on `parkNanos`/`nanoTime` and co.

1. Using `DefaultExecutor` for tracking high-level test timeouts in production isn't ideal, because a timeout indicates that the system is in a weird state, making everything suspect. `DefaultExecutor` could have been flooded with tasks. Using a separate thread like we do now is appropriate in this scenario.
2. Writing a copy of `DefaultExecutor` is trickier than the current implementation.
3. A combined approach is possible: use `DefaultExecutor` + `VirtualTimeSource` only for tests, use a separate thread + real time in production. That was attempted, but looked very convoluted.

### DIFF
--- a/kotlinx-coroutines-debug/src/junit/CoroutinesTimeoutImpl.kt
+++ b/kotlinx-coroutines-debug/src/junit/CoroutinesTimeoutImpl.kt
@@ -134,24 +134,21 @@ internal class FakeDelayEventQueue {
 
     fun processEventQueueAndAwaitCompletion(testTimeoutMs: Long) {
         while (true) {
-            val (timeMs, tasks) = synchronized(this@FakeDelayEventQueue) {
+            val tasks = synchronized(this@FakeDelayEventQueue) {
                 val currentEvents = events ?: return
-                currentEvents.eventPriorityQueue.pollFirstEntry().also {
-                    if (it == null) {
+                val (timeMs, tasks) = currentEvents.eventPriorityQueue.pollFirstEntry()
+                    ?: run {
                         (this@FakeDelayEventQueue as Object).wait()
                         continue
                     }
+                if (timeMs >= testTimeoutMs) {
+                    throw TimeoutException()
                 }
-            }
-            if (timeMs > testTimeoutMs) {
-                throw TimeoutException()
+                currentEvents.currentTime = timeMs
+                tasks
             }
             for (task in tasks) {
                 task.resumeWith(Result.success(Unit))
-            }
-            synchronized(this@FakeDelayEventQueue) {
-                val currentEvents = events ?: continue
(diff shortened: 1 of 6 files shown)


kotlinx-coroutines-debug/src/junit/CoroutinesTimeoutImpl.kt
```diff
@@ -75,3 +97,96 @@ private fun Throwable.attachStacktraceFrom(thread: Thread) {
...
+            val (timeMs, tasks) = synchronized(this@FakeDelayEventQueue) {
+                val currentEvents = events ?: return
+                currentEvents.eventPriorityQueue.pollFirstEntry().also {
+                    if (it == null) {
+                        (this@FakeDelayEventQueue as Object).wait()
+                        continue
+                    }
+                }
+            }
+            if (timeMs > testTimeoutMs) {
```
qwwdfsad: probably should be `>=`? Since we are working with deadlines, they should overlap/be inclusive

kotlinx-coroutines-debug/src/junit/CoroutinesTimeoutImpl.kt
```diff
@@ -75,3 +97,96 @@ private fun Throwable.attachStacktraceFrom(thread: Thread) {
...
+                throw TimeoutException()
+            }
+            for (task in tasks) {
+                task.resumeWith(Result.success(Unit))
+            }
+            synchronized(this@FakeDelayEventQueue) {
+                val currentEvents = events ?: continue
+                currentEvents.currentTime = maxOf(currentEvents.currentTime, timeMs)
```
qwwdfsad: You need to advance time before resuming tasks from the timer, otherwise it seems to interact incorrectly with unconfined dispatchers:

```
    @Test
    @CoroutinesTimeout(11)
    fun noClassTimeout() = runBlocking {
            fakeDelay(10)
            fakeDelay(10)
        }
    }
```

doesn't fail, though it should

qwwdfsad (changes requested):

qwwdfsad (approved):

PR #4730, opened 2026-08-17
feat: Create a new debugging page

daniCsorbaJB:
This PR adds a new coroutine debugging page to the docs.
It's based on existing content located [here](github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-debug) and [here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/topics/debugging.md).

It also adds information for a new 2.4.20 release feature the `StackTraceRecoverable` interface

The related issues are here:

* [KEDOC-320](https://youtrack.jetbrains.com/issue/KEDOC-320) Add a coroutines debug page to kotlinlang.org
* [KT-86595](https://youtrack.jetbrains.com/issue/KT-86595) Introduce StackTraceRecoverable interface into the standard library


docs/topics/debugging.md
```diff
@@ -1,109 +0,0 @@
-**Table of contents**
```
dkhalanskyjb, line 1: Since this file was linked to directly and wasn't part of the guide, this renaming breaks some existing links. Please keep this file. We could either avoid this renaming—after all, since it's the coroutines guide, it's self-evident that this is also coroutine debuggnig—or keep a file similar to `docs/debugging.md` that would redirect readers to the new location (maybe even to the rendered guide?). I don't have a preference.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+
+[//]: # (title: Debug coroutines)
+
+Debugging applications that use coroutines can be challenging because multiple coroutines can run concurrently, suspend on one thread, and resume on another.
+Their execution order and the threads they use can also change between runs, making it difficult to follow the execution of a particular coroutine.
+
+On the JVM, you can use the following features to make debugging coroutines easier:
+
+* [Debug mode](#enable-debug-mode) adds a unique name to each coroutine so you can identify it in a debugger and diagnostic output.
+* [Stack trace recovery](#stack-trace-recovery) adds information about where an exception is rethrown or caught.
```
dkhalanskyjb: That's very misleading. Let's discuss this internally, we'll probably converge on solid wording faster that way.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+* The [debug agent](#the-debug-agent) tracks active coroutines, reports their state, and more.
+
+Debug mode and stack trace recovery are available in the [`kotlinx-coroutines-core`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/) module.
+The debug agent is available in the [`kotlinx-coroutines-debug`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/) module.
+
+> The debug agent isn't supported on Android.
+> 
+> Additionally, R8 version 1.6.0 or later disables debug mode in optimized release builds to reduce the size of the resulting binary.
```
dkhalanskyjb: We can safely omit the R8 version from this tutorial page. 1.6.0 is severely outdated: please see the table at https://developer.android.com/build/kotlin-support

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+   >
+   {style="note"}
+
+2. In the **Run/Debug Configurations** dialog, enter `-Dkotlinx.coroutines.debug` in the **VM options** field:
+
+   ![Adding the -Dkotlinx.coroutines.debug option to a run/debug configuration in IntelliJ IDEA](run-debug-configuration.png){width="600"}
+
+3. Click **OK**.
+
+Running your code with Java assertions enabled also activates debug mode automatically.
```
dkhalanskyjb: I'd expect this point to be made to be much earlier. Unit tests run with enabled assertions by default, so many readers don't need to do anything at all—if the code being debugged runs in a test, the debug mode is already on.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+
+![Recovered exception stack trace with information about where the exception is rethrown](with-stack-trace-recovery.png){width="600"}
+
+### Stack trace recovery for custom exceptions
+<primary-label ref="experimental-opt-in"/>
+
+Stack trace recovery can copy an exception automatically when its class has a public constructor that accepts a message, a cause, both, or no arguments.
+
+If you want the `kotlinx.coroutines` library to recover the stack trace of an exception that requires additional constructor arguments,
+such as a line number or an error code, implement the `StackTraceRecoverable` interface.
```
dkhalanskyjb: Without a doubt, I'd expect this to be a link to the stdlib API reference, but it looks like that documentation wasn't published yet! Could you please drop a comment like `<!-- TODO: add a link to StackTraceRecoverable -->` for us not to forget about it in the future?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+	at kotlinx.coroutines.debug.junit4.CoroutinesTimeoutStatement$evaluate$$inlined$runWithTimeoutDumpingCoroutines$1.call(CoroutinesTimeoutImpl.kt:79)
+	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
+	at java.base/java.lang.Thread.run(Thread.java:1575)
+```
+{collapsible="true" collapsed-title="CoroutinesTimeout example output"}
+
+### Limitations on Android
+
+The debug agent isn't supported on Android.
+However, `kotlinx-coroutines-test` can add `kotlinx-coroutines-debug` to an Android project as a transitive dependency.
```
dkhalanskyjb: This is no longer an issue. https://github.com/Kotlin/kotlinx.coroutines/commit/2706a76ae793c40482fed6e23ba30fecf2ae63b4 removed the `kotlinx-coroutines-debug` dependency from `kotlinx-coroutines-test` in 1.6.0.

The instructions could still be useful, but now, there isn't an obvious way to trigger the error.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+    ) {
+        exclude(
+            group = "org.jetbrains.kotlinx",
+            module = "kotlinx-coroutines-debug",
+        )
+    }
+}
+```
+
+This resolves the conflict if no other dependency adds the debug module.
```
dkhalanskyjb: At least this is outdated now and can be removed!

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+Coroutine "coroutine#4":StandaloneCoroutine{Active}@335eadca, state: SUSPENDED
+	at DebugAgentExampleKt$loadUserProfile$2$2.invokeSuspend(DebugAgentExample.kt:15)
+User profile coroutine hierarchy:
+"coroutine#2":StandaloneCoroutine{Active}, continuation is SUSPENDED at line DebugAgentExampleKt$main$1$loadingJob$1.invokeSuspend(DebugAgentExample.kt:27)
+	"coroutine#3":StandaloneCoroutine{Active}, continuation is SUSPENDED at line DebugAgentExampleKt$loadUserProfile$2$1.invokeSuspend(DebugAgentExample.kt:14)
+	"coroutine#4":StandaloneCoroutine{Active}, continuation is SUSPENDED at line DebugAgentExampleKt$loadUserProfile$2$2.invokeSuspend(DebugAgentExample.kt:15)
+```
+{collapsible="true" collapsed-title="Debug mode example output"}
+
+### Print active coroutines when a JUnit 4 test times out
```
dkhalanskyjb: There's also the equivalent for JUnit 5: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug.junit5/-coroutines-timeout/

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class UserProfileTest {
+    @get:Rule
+    val timeout = CoroutinesTimeout.seconds(1)
+
+    private suspend fun loadUserProfile() {
+        withContext(Dispatchers.IO) {
+            // Simulates an operation that doesn't complete
+            delay(Long.MAX_VALUE)
```
dkhalanskyjb: We recommend using the `Duration`-based overloads for `delay` now, even IDEA highlights this code. `delay(Duration.INFINITE)` is the modern equivalent.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
+	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
+	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
+	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
+	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
+	at kotlinx.coroutines.debug.junit4.CoroutinesTimeoutStatement$evaluate$$inlined$runWithTimeoutDumpingCoroutines$1.call(CoroutinesTimeoutImpl.kt:79)
+	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
+	at java.base/java.lang.Thread.run(Thread.java:1575)
+```
+{collapsible="true" collapsed-title="CoroutinesTimeout example output"}
```
dkhalanskyjb: Will this be uncollapsed by default? Looks like a lot of visual noise that doesn't assist our exposition. Could we edit out the boilerplate parts? `// ... some frames were omitted for brevity`, etc.?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+            loadUserProfile()
+        }
+
+        // Gives the child coroutines time to suspend
+        delay(1.seconds)
+
+        // Prints all active coroutines
+        DebugProbes.dumpCoroutines()
+
+        println("\nUser profile coroutine hierarchy:")
```
dkhalanskyjb: Maybe we should stick a `println()` (without an argument) just above, or possibly `println("============")`, so that the output is easier to understand at a glance?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+<primary-label ref="experimental-opt-in"/>
+
+The [`kotlinx-coroutines-debug`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/) module provides a debug agent for JVM applications.
+The agent tracks coroutines as they are created, suspended, and resumed.
+
+The [`DebugProbes`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug/-debug-probes/) API is the main entry point for the debug agent.
+You can use it to print active coroutines and their current state.
+The output includes stack traces that show where each coroutine was created and where it is suspended.
+
+You can also use it to print the coroutine hierarchy for a specific `Job` or `CoroutineScope`.
```
dkhalanskyjb: You can print hierarchies without any debug mechanisms, just by traversing the `parent`/`children` links in a `Job`. What `DebugProbes` allow is printing coroutine dumps for a *subset of all coroutines* that corresponds to a specific hierarchy.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+
+Additionally, you can use the debug agent in production environments to monitor an application's state and improve its observability.
+When you do so, set the [`DebugProbes.enableCreationStackTraces`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug/-debug-probes/enable-creation-stack-traces.html) property to `false` to reduce performance overhead.
```
dkhalanskyjb: I feel that the focus here is slightly off. It being possible to use `DebugProbes` in production isn't a feature but a fact of life—someone may want to enable this in production as well for some reason. If they do, we have a warning for them: creating stack traces are slow and should be avoided because of how severe the resulting slowdown is.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,488 @@
...
+    
+    // Normally, you don't need to call this function directly unless you're testing its behavior
+    // The kotlinx.coroutines library invokes it automatically during stack trace recovery
+    val copy = original.copyForStackTraceRecovery()
+
+    println(copy.message)
+    // When editing line 15: Unexpected token
+
+    println(copy.cause == original)
+    // true
```
dkhalanskyjb: This example is reasonable for describing a *stdlib function*—where the documentation should be preferably library-agnostic—yet we're in `kotlinx.coroutines` here and can demonstrate the exact change in the behavior (like the general stacktrace recovery example above).

Note that https://play.kotlinlang.org/ won't be of help—the `kotlinx.coroutines` version is stuck at 1.7.3 there, and only `kotlinx.coroutines` 1.11.0+ supports `StackTraceRecoverable`.

dkhalanskyjb (changes requested):

docs/topics/debugging.md
daniCsorbaJB: > or keep a file similar to docs/debugging.md that would redirect readers to the new location (maybe even to the rendered guide?).

yes good point — let's go with this option — I'll do it similarly to the rest of the old docs like for cancellation:

> The documentation has been moved to the [https://kotlinlang.org/docs/coroutines-cancellation.html](https://kotlinlang.org/docs/coroutines-cancellation.html) page.

docs/topics/coroutines-debugging.md
daniCsorbaJB: thank you 👍  good point — 

I moved it before the "to enable part"
WDYT would it help users to mention the part that unit tests run with assertions enabled by default here? 🤔

docs/topics/coroutines-debugging.md
daniCsorbaJB: yes, definitely I have reminders for all of these (the already published What's new included)

docs/topics/coroutines-debugging.md
daniCsorbaJB: nice removed it 👍

docs/topics/coroutines-debugging.md
daniCsorbaJB: 👍 I'll rewrite this part slightly then and add a subsection/info for that as well

docs/topics/coroutines-debugging.md
daniCsorbaJB: yes — I put it there as some extra optional content - but it's collapsed by default. the reader can expand it if they want/collapse it afterward

image

docs/topics/coroutines-debugging.md
daniCsorbaJB: sure! sounds good to me 👍

docs/topics/coroutines-debugging.md
daniCsorbaJB: 👍 got it — rewrote this part and the specific list bullets as well to clarify:

You can also use it to print a coroutine dump for the hierarchy of a specific `Job` or `CoroutineScope`.

docs/topics/coroutines-debugging.md
daniCsorbaJB: I see what you mean. What do you think about phrasing it like this? 🤔 

> If you enable `DebugProbes` in a production environment, it can significantly reduce your application's performance when it creates a stack trace for each new coroutine.
To avoid this overhead, set [`DebugProbes.enableCreationStackTraces`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug/-debug-probes/enable-creation-stack-traces.html) to `false`.

docs/topics/coroutines-debugging.md
dkhalanskyjb: Yes. Then, I'd expect most people to go: "Great, then, I can skip this section!" It's always nice when your documentation manages to provide an answer more or less immediately.

docs/topics/coroutines-debugging.md
dkhalanskyjb: Hm, alright. Just to double-check: is it redundant that `coroutines-and-channels.md` describes the initial state of the collapsible explicitly as `{initial-collapse-state="collapsed" collapsible="true"}`?

docs/topics/coroutines-debugging.md
dkhalanskyjb: That works, yes.

docs/topics/coroutines-debugging.md
daniCsorbaJB: it is 👍 😄 ([for example](https://kotlinlang.org/docs/type-safe-builders.html)) — now you only have to specify it the other way (if you want it to be collapsible but expanded by default)

dkhalanskyjb (approved):

docs/images/run-debug-configuration.png
sarahhaggarty, line 1: Can you crop this image so the rounded edges with transparent corners don't show?

docs/topics/coroutines-and-channels.md
```diff
@@ -2,6 +2,10 @@
 
 [//]: # (title: Coroutines and channels − tutorial)
 
+> We plan to revise this tutorial. For an up-to-date guide to getting started with coroutines, see [Coroutines basics](coroutines-basics.md).
```
sarahhaggarty:
```suggestion
> An upcoming update will revise this tutorial. In the meantime, for an up-to-date guide to getting started with coroutines, see [Coroutines basics](coroutines-basics.md).
```
To avoid "we".

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+>
+>
+{style="note"}
+
+## Enable debug mode
+
+Debug mode assigns a unique name to every launched coroutine.
+You can see the coroutine names in a Java debugger, in the coroutine's string representation, and in a thread's name while it runs the coroutine.
+
+> Debug mode has negligible runtime overhead, so you can keep it enabled to simplify logging and diagnostics.
```
sarahhaggarty: Does this have to be a tip? I think it would be a useful addition to the previous paragraph.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+>
+{style="tip"}
+
+When you run your code with Java assertions enabled, the `kotlinx.coroutines` library automatically enables debug mode.
+Unit tests run with assertions enabled by default, so you don't need to enable debug mode explicitly for them.
+
+To enable debug mode explicitly, pass the `-Dkotlinx.coroutines.debug` VM option in your build tool or IDE run configuration.
```
sarahhaggarty: What does VM mean here? And can we refer to any useful Maven or Gradle materials we have? Or add tabs to demonstrate how?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+3. Click **OK**.
+
+## Stack trace recovery
+
+When a coroutine receives an exception from another coroutine through a suspending function such as `Deferred.await()`,
+the exception's stack trace doesn't contain the stack frames from the receiving coroutine.
+Without these stack frames, the stack trace doesn't show where `await()` is called or which functions lead to that call.
+
+The `kotlinx.coroutines` library adds this information using _stack trace recovery_, which creates a copy of the exception with additional stack frames.
```
sarahhaggarty, line 55:
```suggestion
The `kotlinx.coroutines` library restores this missing information through _stack trace recovery_, which creates a copy of the exception with additional stack frames.
```

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+   > If you don't have a run/debug configuration, select **Current File** in the **Run widget**, then select **More Actions** | **Run with Parameters** to open the run configuration settings.
+   >
+   {style="note"}
+
+2. In the **Run/Debug Configurations** dialog, enter `-Dkotlinx.coroutines.debug` in the **VM options** field:
+
+   ![Adding the -Dkotlinx.coroutines.debug option to a run/debug configuration in IntelliJ IDEA](run-debug-configuration.png){width="600"}
+
+3. Click **OK**.
```
sarahhaggarty: Let's combine this with the previous instruction. It's so short it's not worth a whole other step :)

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+## Stack trace recovery
+
+When a coroutine receives an exception from another coroutine through a suspending function such as `Deferred.await()`,
+the exception's stack trace doesn't contain the stack frames from the receiving coroutine.
+Without these stack frames, the stack trace doesn't show where `await()` is called or which functions lead to that call.
```
sarahhaggarty: Could you extend this sentence a little more with something like "so it's hard to understand what caused the exception."? It feels like it could use just a little more explanation of why missing these stack frames is a problem.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+The `kotlinx.coroutines` library adds this information using _stack trace recovery_, which creates a copy of the exception with additional stack frames.
+
+When the receiving coroutine resumes, it throws the copy instead of the original exception, which becomes the cause of the copy.
```
sarahhaggarty: Can you separate this into two sentences? It's a bit hard to determine what "which" is referring to easily. Out of curiosity, why does the exception become the cause of the copy? Does it make it easier to debug somehow? This part isn't clear to me.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+If the original exception has suppressed exceptions, they remain attached to it instead of being copied.
+This can prevent cycles in the exception chain and crashes in some frameworks.
```
sarahhaggarty:
```suggestion
Keeping them attached to the original exception can prevent cycles in the exception chain and crashes in some frameworks.
```
Just to make it clear what "this" is referring to :)

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+Debug mode enables stack trace recovery by default.
+To disable stack trace recovery in debug mode, pass the `-Dkotlinx.coroutines.stacktrace.recovery=false` VM option.
+
+Here's an example that demonstrates the difference between stack traces with and without stack trace recovery:
```
sarahhaggarty:
```suggestion
Here's an example to demonstrate the difference between stack traces with and without stack trace recovery:
```

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+    }
+}
+
+suspend fun main() {
+    UserProfileService.awaitUserProfile()
+}
+```
+
+In this example, the `parseUserProfile()` function throws the exception in the coroutine started by the `.async()` builder function.
+The coroutine that calls `awaitUserProfile()` receives the exception through `Deferred.await()`.
```
sarahhaggarty:
```suggestion
The coroutine that calls `awaitUserProfile()` receives the exception through the `await()` function.
```
Should this and all the other mentions of `await()` have `Deferred` removed?

docs/images/without-stack-trace-recovery.png
sarahhaggarty, line 1: Can you add a box or a bracket to highlight the lines that are added to the stack trace?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+![Exception stack trace without stack frames from the receiving coroutine](without-stack-trace-recovery.png){width="600"}
+
+With stack trace recovery enabled, the stack trace also includes the `Deferred.await()` call in the `awaitUserProfile()` function:
+
+![Recovered exception stack trace with stack frames from the receiving coroutine](with-stack-trace-recovery.png){width="600"}
+
+### Stack trace recovery for custom exceptions
+<primary-label ref="experimental-opt-in"/>
+
+Stack trace recovery can copy an exception automatically when its class has a public constructor that accepts a message, a cause, both, or no arguments.
```
sarahhaggarty:
```suggestion
Stack trace recovery can copy an exception automatically when its class has a public constructor that accepts a message, a cause, both, or no arguments at all.
```
To make it sound more natural.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+If you want the `kotlinx.coroutines` library to recover the stack trace of an exception that requires additional constructor arguments,
+such as a line number or an error code, implement the `StackTraceRecoverable` interface.
```
sarahhaggarty: Is there an API reference link we can add for the interface?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+        // Stack trace recovery creates a copy of the exception,
+        // adds the calling coroutine's stack frames, and throws the copy
+        fileEdit.await()
+    }
+}
+```
+
+With debug mode enabled, the output contains the recovered copy followed by the original exception as its cause.
+
+```text
```
sarahhaggarty, line 180: To be consistent, can you make this also a screenshot? I think you can also add it under a collapsible.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+<primary-label ref="experimental-opt-in"/>
+
+The [`kotlinx-coroutines-debug`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/) module provides a debug agent for JVM applications.
+The agent tracks coroutines as they are created, suspended, and resumed.
+
+The [`DebugProbes`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug/-debug-probes/) API is the main entry point for the debug agent.
+You can use it to print active coroutines and their current state.
+The output includes stack traces that show where each coroutine was created and where it is suspended.
+
+You can also use it to print a coroutine dump for the hierarchy of a specific `Job` or `CoroutineScope`.
```
sarahhaggarty, line 209: I think this can be combined with the paragraph above since they are both about the API.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+If you enable `DebugProbes` in a production environment, it can significantly reduce your application's performance when it creates a stack trace for each new coroutine.
```
sarahhaggarty, line 211: Is it a valid use case to have DebugProbes enabled at all in production? Shouldn't it be reserved for debugging and in this case, the performance impact is expected?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+```
+
+</tab>
+</tabs>
+
+### Track coroutines with the debug agent
+
+To start tracking coroutines with the debug agent, you can either:
+
+* Call the [`DebugProbes.install()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug/-debug-probes/install.html) function before starting the coroutines you want to track.
```
sarahhaggarty: If this option triggers a warning in newer JDK versions, shouldn't `-javaagent` be the first bullet as the preferred method?

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+            // Retains one copy of each Byte Buddy Agent file
+            pickFirsts += setOf(
+                "win32-x86-64/attach_hotspot_windows.dll",
+                "win32-x86/attach_hotspot_windows.dll",
+            )
+        }
+    }
+}
+```
```
sarahhaggarty, line 559: WDYT about adding a What's next? section with links to the tutorials you have about debugging?

docs/topics/debugging.md
```diff
@@ -1,109 +1,3 @@
...
+The documentation has been moved to the [https://kotlinlang.org/docs/coroutines-debugging.html](https://kotlinlang.org/docs/coroutines-debugging.html) page.
```
sarahhaggarty, line 1:
```suggestion
The documentation has been moved to the [Debug coroutines](https://kotlinlang.org/docs/coroutines-debugging.html) page.
```

docs/debugging.md
```diff
@@ -1 +1 @@
-The documentation has been moved to the [topics/debugging.md](topics/debugging.md).
\ No newline at end of file
+The documentation has been moved to the [https://kotlinlang.org/docs/coroutines-debugging.html](https://kotlinlang.org/docs/coroutines-debugging.html) page.
```
sarahhaggarty, line 1:
```suggestion
The documentation has been moved to the [Debug coroutines](https://kotlinlang.org/docs/coroutines-debugging.html) page.
```

docs/topics/coroutines-basics.md
```diff
@@ -235,7 +235,7 @@ You may notice that the output order and thread names may change each time you r
 > You can display coroutine names next to thread names in the output of your code for additional information.
 > To do so, pass the `-Dkotlinx.coroutines.debug` VM option in your build tool or IDE run configuration.
 >
-> See [Debugging coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/master/docs/topics/debugging.md) for more information.
+> See [Debugging coroutines](coroutines-debugging.md) for more information.
```
sarahhaggarty:
```suggestion
> See [Debug coroutines](coroutines-debugging.md) for more information.
```

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+
+To enable debug mode explicitly, pass the `-Dkotlinx.coroutines.debug` VM option in your build tool or IDE run configuration.
+
+In IntelliJ IDEA, follow these steps to enable debug mode:
+
+1. In the **Run widget**, select the run/debug configuration you want to update, then select **More Actions** | **Edit**:
+
+   ![Selecting Edit from the More Actions menu for a run configuration in IntelliJ IDEA](coroutines-debug-mode-more-options.png){width="600"}
+
+   > If you don't have a run/debug configuration, select **Current File** in the **Run widget**, then select **More Actions** | **Run with Parameters** to open the run configuration settings.
```
sarahhaggarty, line 39: Can you add a screenshot for this as well? And maybe move it under a collapsible? Like in this KMP tutorial: https://kotlinlang.org/docs/multiplatform/multiplatform-create-first-app.html#run-your-application-on-android

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,561 @@
...
+	at kotlinx.coroutines.BuildersKt.runBlockingK$default(Unknown Source)
+	at UserProfileTest.loadsUserProfile(UserProfileTest.kt:18)
+	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
+	at kotlinx.coroutines.debug.junit5.CoroutinesTimeoutExtension$interceptInvocation$$inlined$runWithTimeoutDumpingCoroutines$1.call(CoroutinesTimeoutImpl.kt:79)
+	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
+	at java.base/java.lang.Thread.run(Thread.java:1575)
+```
+{collapsible="true" collapsed-title="CoroutinesTimeout JUnit5 example output"}
+
+### Limitations on Android
```
sarahhaggarty: I'm a bit confused with this section. We say that the debug agent doesn't support Android but then we advise how to resolve a build failure. Is it so that you can use other parts of the debug module that aren't the debug agent? 🤔

sarahhaggarty (changes requested):
Nice work! I see why this is a cool feature to document. I have some clarification questions that I need your help with :)

docs/topics/coroutines-and-channels.md
daniCsorbaJB: sure  works for me 👍  although for these messages/notes, I generally don't mind the "we". (it's not unclear like in tutorials who the we refers to)

docs/topics/coroutines-debugging.md
daniCsorbaJB: Hmm 🤔 "restores missing information" might work conceptually, but I think simply “adds this information” is more precise because those stack frames were never part of the original exception’s stack trace.

docs/topics/coroutines-debugging.md
daniCsorbaJB: nope - not yet unfortunately — but we'll update these when we'll have them.

docs/topics/coroutines-debugging.md
daniCsorbaJB: good point 👍 let's flip it. (it's just easier to demonstrate in a code snippet on the website the other way, but much better flipped so good point)

docs/topics/coroutines-debugging.md
daniCsorbaJB: I'd prefer not to — some of these are really long (one of them is 80 lines~ in JUnit) so the screenshot would be unreadable. it's cool for the stack trace one, especially with your proposal, but for the rest I much prefer this format because it would either get squashed during resizing or be an absolute monster 😅

docs/topics/coroutines-debugging.md
daniCsorbaJB: sure, someone might find it a valid use case 👍 
The real performance impact is from the stack traces, which they should disable to avoid the overhead, but otherwise — someone might want to do it.

docs/debugging.md
daniCsorbaJB: The other pages already use this format, so I kept it for consistency. The full URL makes it clear that the link points to kotlinlang.org rather than another file in the repository.

That said, I think we could consider changing all of them to something like (but probably in a separate PR for clarity):

The documentation has been moved to the [Debug coroutines] page on kotlinlang.org. or something similar

docs/topics/coroutines-debugging.md
daniCsorbaJB: 🤔 I think the title is the culprit here. 

While there is no obvious way to trigger this error, if I understand correctly, it may still be useful for some edge cases. 

this configuration only resolves the resource conflict if `kotlinx-coroutines-debug` is present in an Android project's dependency graph somehow. It doesn't enable the debug agent or other functionality from the module.

I renamed the section to:
### Resolve `kotlinx-coroutines-debug` resource conflicts on Android

docs/topics/coroutines-debugging.md
daniCsorbaJB: no strong opinion on this one — sure works for me either way 👍

docs/topics/coroutines-debugging.md
daniCsorbaJB: > And can we refer to any useful Maven or Gradle materials we have?

Not sure, I haven't found anything specific — I found it in the [Gradle docs](https://docs.gradle.org/current/userguide/application_plugin.html#sec:application_usage) though.

> What does VM mean here?

It's an argument passed to the JVM. We could specify it as: 

> To enable debug mode explicitly, configure your build tool or IDE run configuration to pass the `-Dkotlinx.coroutines.debug` argument to the JVM that runs your application.

docs/topics/coroutines-debugging.md
daniCsorbaJB: sure thing 👍 : I added:
 
> Without these stack frames, the stack trace doesn't show where `await()` is called or which functions lead to that call, which can make debugging difficult.

with something like "it's hard to understand what caused the exception" — is a bit too fuzzy, imho, but a more specific explanation would just repeat too much: "...which makes it hard to trace which functions lead to that call and which function received the exception from an await() call."

docs/topics/coroutines-debugging.md
daniCsorbaJB: > Can you separate this into two sentences?

Sure thing — good point! 👍 💯  

> Out of curiosity, why does the exception become the cause of the copy? Does it make it easier to debug somehow? This part isn't clear to me.

🤔 Ah good question: no, throwing a copy doesn't make it easier to debug, it's just an important piece of information, but the reason behind it is implementation-related, so it's not important for the user. 

The following sentences expand on this a little, but the main reason is that (if I understand correctyl) that multiple coroutines can receive the same exception instance, for example by awaiting the same Deferred, then stack trace recovery would modify that shared exception, and the coroutines could overwrite one another's recovered stack frames. (which would make a mess)

docs/topics/coroutines-debugging.md
daniCsorbaJB: actually I'd prefer the other way around 😅 I only used it without Deferred once in the doc since we already mentioned it beforehand. but for clarity I'd prefer it with Deferred here as well. I don't think anyone would misunderstand, but it's clearer.

docs/images/without-stack-trace-recovery.png
daniCsorbaJB: ofc, really good idea! added it

docs/topics/coroutines-debugging.md
daniCsorbaJB: sure 👍  added a bullet point for the IntelliJ IDEA ones

docs/topics/coroutines-debugging.md
daniCsorbaJB: added it 👍 

hmm 🤔 there it's only possible because it's on lvl 4 I think. 😢 or I don't see it how to make it collapsible otherwise.

docs/topics/coroutines-debugging.md
```diff
@@ -0,0 +1,562 @@
...
+                "win32-x86/attach_hotspot_windows.dll",
+            )
+        }
+    }
+}
+```
+
+## What's next
+
+* Learn how to debug coroutines in IntelliJ IDEA in [Debug coroutines using IntelliJ IDEA](debug-coroutines-with-idea.md) and [Debug Kotlin Flow using IntelliJ IDEA](debug-flow-with-idea.md).
```
sarahhaggarty:
```suggestion
Learn how to debug coroutines in IntelliJ IDEA in [Debug coroutines using IntelliJ IDEA](debug-coroutines-with-idea.md) and [Debug Kotlin Flow using IntelliJ IDEA](debug-flow-with-idea.md).
```

docs/topics/coroutines-debugging.md
sarahhaggarty: I like the new description, thanks! And that Gradle link looks good. What about this one for Maven? https://maven.apache.org/configure

docs/debugging.md
sarahhaggarty: If the other pages are like this, then don't worry about it. It's not worth updating all instances of it :)

docs/topics/coroutines-debugging.md
sarahhaggarty: The screenshot is small so I think it's OK without a collapsible!

docs/topics/coroutines-debugging.md
sarahhaggarty: I like it! This new name makes it much clearer to me. Thanks!

sarahhaggarty (approved):
Thanks so much for the changes! If you can add the build tools references, I think this will be good to go! 🚀

docs/topics/coroutines-debugging.md
daniCsorbaJB: sounds good to me 👍 I added them

PR #4732, opened 2026-08-20
Upgrade kotlinx.benchmark to 0.4.17

Egorand:
https://youtrack.jetbrains.com/issue/KT-80897 removes a Kotlin Gradle plugin API that was used by kotlinx.benchmark prior to version 0.4.15, so once kotlinx.coroutines upgrades to KGP 2.5.x it'll run in binary compatibility issues. This change prevents that by upgrading to a safe kotlinx.benchmarks version.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -68,7 +68,12 @@ dependencies {
         exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-jdk7")
         exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib")
     }
-    implementation("org.jetbrains.kotlinx:kotlinx-benchmark-plugin:${version("benchmarks")}")
+    // kotlinx-benchmark-plugin may depend on a newer version of kotlin-stlib than the one bundled by Gradle,
+    // leading to Gradle complaining about an incompatible version of Kotlin. Exclude the kotlin-stdlib
+    // dependency to default to the one bundled by Gradle.
+    implementation("org.jetbrains.kotlinx:kotlinx-benchmark-plugin:${version("benchmarks")}") {
+        exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib")
+    }
     implementation("org.jetbrains.kotlinx:kotlinx-knit:${version("knit")}")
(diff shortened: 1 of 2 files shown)


buildSrc/build.gradle.kts
```diff
@@ -68,7 +68,12 @@ dependencies {
         exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-jdk7")
         exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib")
     }
-    implementation("org.jetbrains.kotlinx:kotlinx-benchmark-plugin:${version("benchmarks")}")
+    // kotlinx-benchmark-plugin may depend on a newer version of kotlin-stlib than the one bundled by Gradle,
+    // leading to Gradle complaining about an incompatible version of Kotlin. Exclude the kotlin-stdlib
+    // dependency to default to the one bundled by Gradle.
```
dkhalanskyjb, line 73: This doesn't make much sense to me. If the plugin is using the capabilities of the new Kotlin stdlib version, removing its dependency can break the plugin, meaning we aren't allowed to remove the dependency.

dkhalanskyjb (changes requested):
`master` is not the right branch to target here. Please use `develop` if this is a change intended for the current state of the codebase, and `kotlin-community/dev` if it's a workaround for `kotlinx.coroutines` failing as a project used to run compiler/tooling integration tests against.

buildSrc/build.gradle.kts
Egorand: Yeah, makes sense. I'm gonna upgrade kotlinx.benchmarks to 0.4.16 instead - it doesn't use the KGP API that's going away in 2.5.0, and builds against an older version of Kotlin, compatible with the Gradle version coroutines uses.

Egorand:
> `master` is not the right branch to target here. Please use `develop` if this is a change intended for the current state of the codebase, and `kotlin-community/dev` if it's a workaround for `kotlinx.coroutines` failing as a project used to run compiler/tooling integration tests against.

Thanks, makes sense! Let me re-submit a PR targeting `kotlin-community/dev`, as that's indeed a better fit for what I'm trying to do.

Edit: #4733

PR #4733, opened 2026-08-24
Upgrade kotlinx.benchmarks to 0.4.16

Egorand:
https://youtrack.jetbrains.com/issue/KT-80897 removes a Kotlin Gradle plugin API that was used by kotlinx.benchmark prior to version 0.4.15, so once kotlinx.coroutines upgrades to KGP 2.5.x it'll run in binary compatibility issues. This change prevents that by upgrading to a safe kotlinx.benchmarks version.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ kotlin_version=2.3.21
 atomicfu_version=0.32.1
 
 # Dependencies
-benchmarks_version=0.4.13
+benchmarks_version=0.4.16
 benchmarks_jmh_version=1.37
 junit_version=4.12
 junit5_version=5.7.0


fzhinkin:
Is there a reason to update it only to 0.4.16 instead of the latest available version?

It's expected that only 0.4.18 will support 2.5.0 in the nearest future.

Egorand:
> Is there a reason to update it only to 0.4.16 instead of the latest available version?
> 
> It's expected that only 0.4.18 will support 2.5.0 in the nearest future.

kotlinx.benchmarks 0.4.17 (current latest) is built against Kotlin 2.2.0, while coroutines is using Gradle 8.14.4, which bundles Kotlin 2.0.x, so when I try to upgrade to 0.4.17, Gradle complains about version incompatibility:

```
> Task :buildSrc:compilePluginsBlocks FAILED
e: file:///Users/egor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.2.0/fdfc65fbc42fda253a26f61dac3c0aca335fae96/kotlin-stdlib-2.2.0.jar!/META-INF/kotlin-stdlib-jdk7.kotlin_module Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0.
e: file:///Users/egor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.2.0/fdfc65fbc42fda253a26f61dac3c0aca335fae96/kotlin-stdlib-2.2.0.jar!/META-INF/kotlin-stdlib-jdk8.kotlin_module Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0.
e: file:///Users/egor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/2.2.0/fdfc65fbc42fda253a26f61dac3c0aca335fae96/kotlin-stdlib-2.2.0.jar!/META-INF/kotlin-stdlib.kotlin_module Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0.
```

It'll probably make sense to upgrade coroutines to Gradle 9.x to unlock compatibility with newer versions of benchmarks. I tried quickly bumping coroutines to Gradle 9.7.1 (latest) and ran into configuration warnings, so I'm not sure how much effort this will require.

fzhinkin:
@Egorand, please use `develop` as the target branch. Gradle is already updated to `9.6.1` there. And I'll merge it into the community branch soon (this week, probably).

As of kotlinx-benchmark's compatibility with Gradle 8.x - it's quite a shame as we explicitly declare it. 😢 
I'll check if there's anything I can do about it.

Egorand:
> @Egorand, please use `develop` as the target branch. Gradle is already updated to `9.6.1` there. And I'll merge it into the community branch soon (this week, probably).
> 
> As of kotlinx-benchmark's compatibility with Gradle 8.x - it's quite a shame as we explicitly declare it. 😢 I'll check if there's anything I can do about it.

Sure! #4734. Syncing kotlin-community/dev with develop unblocks my work, thanks!

PR #4734, opened 2026-08-24
Upgrade kotlinx.benchmarks to 0.4.17

Egorand:
https://youtrack.jetbrains.com/issue/KT-80897 removes a Kotlin Gradle plugin API that was used by kotlinx.benchmark prior to version 0.4.15, so once kotlinx.coroutines upgrades to KGP 2.5.x it'll run into binary compatibility issues. This change prevents that by upgrading to a safe kotlinx.benchmarks version.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ kotlin_version=2.3.21
 atomicfu_version=0.32.1
 
 # Dependencies
-benchmarks_version=0.4.13
+benchmarks_version=0.4.17
 benchmarks_jmh_version=1.37
 junit_version=4.12
 junit5_version=5.7.0


fzhinkin:
@Egorand, could you please update it to `0.4.18` instead?

Egorand:
> @Egorand, could you please update it to `0.4.18` instead?

Ah, sorry, didn't notice the new release. Updated to 0.4.18!

dkhalanskyjb:
The build fails with
> `:kotlinx-coroutines-core:linuxX64BenchmarkBenchmarkGenerate`
> `> 'byte[] org.jetbrains.kotlin.library.KotlinLibrary.getModuleHeaderData()'`

Something is wrong with the benchmark plugin, it seems?

Egorand:
> The build fails with
> 
> > `:kotlinx-coroutines-core:linuxX64BenchmarkBenchmarkGenerate`
> > `> 'byte[] org.jetbrains.kotlin.library.KotlinLibrary.getModuleHeaderData()'`
> 
> Something is wrong with the benchmark plugin, it seems?

Looks like benchmarks is using the `MetadataLibrary.moduleHeaderData` API ([code](https://github.com/Kotlin/kotlinx-benchmark/blob/c37c4bd4ad3e2ed3810e1773700f0f1ded154f79/plugin/main/src/kotlinx/benchmark/gradle/CopyKlibMetadataModuleDescriptorFactoryImpl.kt#L74)) removed in Kotlin 2.3.20 ([commit](https://github.com/JetBrains/kotlin/commit/ddb73a97ae3be1e1491f57182f21e754b9278ed1)). coroutines uses Gradle 9.6.1, which bundles Kotlin 2.3.21, and that leads to a NoSuchMethodError. @fzhinkin is bumping benchmarks to at least Kotlin 2.3.21 an option?

fzhinkin:
No, it's not an option. The plugin has to be reimplemented, which, hopefully, will happen within a few next weeks.

Egorand:
> No, it's not an option. The plugin has to be reimplemented, which, hopefully, will happen within a few next weeks.

Got it. Would it be okay to bump to 0.4.17 in the meantime? I'm working on deleting a deprecated Kotlin Gradle plugin API that benchmarks 0.4.13 uses, and the stale dependency in coroutines is currently a blocker.

Egorand:
@fzhinkin @dkhalanskyjb I propose upgrading benchmarks 0.4.17 for now (updated the code) to unblock the KGP API deprecation and make coroutines compatible with KGP 2.5.0 once it lands. Once there's a newer compatible version of the benchmarks plugin we can perform another upgrade. Please let me know if you have any concerns.

dkhalanskyjb:
I don't mind. If the build passes, I'll be happy to merge this PR. There aren't any downsides to this, after all.

dkhalanskyjb (approved):

PR #4735, opened 2026-08-25
docs(core): fix duplicated words in internal comments

zjuzhongwen:
Fix three duplicated words in internal comments covering job state, the null surrogate, and platform exception handling.

### DIFF
--- a/kotlinx-coroutines-core/common/src/JobSupport.kt
+++ b/kotlinx-coroutines-core/common/src/JobSupport.kt
@@ -1549,7 +1549,7 @@ private class ResumeAwaitOnCompletion<T>(
         val state = job.state
         assert { state !is Incomplete }
         if (state is CompletedExceptionally) {
-            // Resume with with the corresponding exception to preserve it
+            // Resume with the corresponding exception to preserve it
             continuation.resumeWithException(state.cause)
         } else {
             // Resuming with value in a cancellable way (AwaitContinuation is configured for this mode).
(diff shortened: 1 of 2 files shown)


kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.common.kt
```diff
@@ -23,7 +23,7 @@ internal expect fun propagateExceptionFinalResort(exception: Throwable)
  * Deal with exceptions that happened in coroutines and weren't programmatically dealt with.
  *
  * First, it notifies every [CoroutineExceptionHandler] in the [platformExceptionHandlers] list.
- * If one of them throws [ExceptionSuccessfullyProcessed], it means that that handler believes that the exception was
+ * If one of them throws [ExceptionSuccessfullyProcessed], it means that handler believes that the exception was
```
dkhalanskyjb: That's valid English. "it means that [**that handler** believes ...]"

dkhalanskyjb (changes requested):

kotlinx-coroutines-core/common/src/internal/CoroutineExceptionHandlerImpl.common.kt
zjuzhongwen: Thanks for the clarification! I misunderstood the second "that" as a duplicate word. 

I've restored the original wording. Please review it again.

Issue #4643, opened 2026-03-23
Deprecate `NonCancellable` in favor of a new scoped function


dkhalanskyjb:
`NonCancellable` is problematic.

First, it can be passed to `launch`/`async`/`produce`/etc. and break structured concurrency.  This is not what you want: see https://github.com/Kotlin/kotlinx.coroutines/pull/4435 for details.

The only correct usage of `NonCancellable` is as the `withContext` parameter. This is an idiom:

```kotlin
try {
    // do something
} finally {
    withContext(NonCancellable) {
        // cleaning up resources whose `close` is `suspend`
    }
}
```

`NonCancellable` is not particularly good at that, as well. `withContext` uses *prompt cancellation*. If `withContext(NonCancellable)` suspends on the final `}` (waiting for its child coroutines, or dispatching back to the original dispatcher), then it will throw a `CancellationException` on exit. That will override the original exception thrown inside `try`: https://pl.kotl.in/u5gIaCPQ5

What we need instead is:

```kotlin
suspend fun <T> nonCancellable(action: suspend CoroutineScope.() -> T): T =
    withContext(NonCancellable) {
        coroutineScope {
            action()
        }
    }
```

This looks like the only reasonable usage pattern for `NonCancellable`, so the additional flexibility presented by it being a `CoroutineContext` element is harmful.

PR #4736, opened 2026-08-26
Fix warnings

dkhalanskyjb:
The warnings were mostly introduced by the new deprecated `launch` and `async` overloads, though there are a couple of others.


kotlinx-coroutines-core/common/test/flow/channels/ChannelFlowTest.kt
```diff
@@ -204,4 +204,4 @@ class ChannelFlowTest : TestBase() {
         myFlow.collect()
         finish(4)
     }
-}
+}
```
qwwdfsad: newline

kotlinx-coroutines-core/common/test/NonCancellableTest.kt
```diff
@@ -40,6 +40,7 @@ class NonCancellableTest : TestBase() {
     @Test
     fun testNonCancellableWithException() = runTest {
         expect(1)
+        @Suppress("DEPRECATION") // async(NonCancellable) is deprecated, but we test its behavior here
```
qwwdfsad, line 43: We expect everybody who relies on `NonCancellable` to suppress after the next major release, correct?

qwwdfsad (approved):

kotlinx-coroutines-core/common/test/NonCancellableTest.kt
dkhalanskyjb: We do want to get rid of `NonCancellable` altogether in favor of a separate function if that's what you mean: https://github.com/Kotlin/kotlinx.coroutines/issues/4643

Issue #4698, opened 2026-07-15
Fail fast for CoroutineStart.ATOMIC when an underlying dispatcher throws an exception


qwwdfsad:
If a dispatcher throws an exception when dispatching a coroutine,  the following code will fail:
```
val brokenDispatcher = object : CoroutineDispatcher() {
    override fun dispatch(context: CoroutineContext, block: Runnable) = TODO()
}

runBlocking<Unit> {
    launch(brokenDispatcher, ) {
    }
}
```
because of the special handling introduced in #880.
But `launch(brokenDispatcher, CoroutineStart.ATOMIC)` will silently hang without any errors or diagnostics, though nothing prevents us from using the same fail-fast mechanism there.

E.g. the following patch presumably addresses the issue: https://gist.github.com/qwwdfsad/d221e99eea6767bbe6f6be135373e5ff 

See also: [this discussion](https://github.com/Kotlin/kotlinx.coroutines/pull/4662#discussion_r3480259475)

PR #4737, opened 2026-08-30
Fail fast when CoroutineStart.ATOMIC dispatch fails

h2jinee:
Fixes #4698.

`CoroutineStart.ATOMIC` currently uses `startCoroutine`. If the initial dispatch throws, the exception can bypass the fail-fast handling and leave the coroutine incomplete.

This change wraps the atomic start in `runSafely` while preserving its non-cancellable start semantics. It also adds a regression test for a dispatcher that throws during initial dispatch.

Tested with:

- `./gradlew :kotlinx-coroutines-core:jvmTest`
- `./gradlew :kotlinx-coroutines-core:checkKotlinAbi`


dkhalanskyjb:
Hi! Thank you for the effort, but it already wasn't challenging for us to copy `@qwwdfsad`'s solution into the codebase. The hard part is figuring out the tradeoffs. If you have an opinion on this, please share it under #4698.

PR #4739, opened 2026-09-02
fix: use unsafeCast instead of `as` for the no runtime interfaces

JSMonk:
The commit is fixing a compilation error appearing during metadata compilation

### DIFF
--- a/kotlinx-coroutines-core/web/src/Promise.kt
+++ b/kotlinx-coroutines-core/web/src/Promise.kt
@@ -62,8 +62,7 @@ public fun <T: JsAny?> Deferred<T>.asPromise(): Promise<T> {
  */
 @OptIn(ExperimentalWasmJsInterop::class)
 public fun <T: JsAny?> Promise<T>.asDeferred(): Deferred<T> {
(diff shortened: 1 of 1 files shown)


kotlinx-coroutines-core/web/src/Promise.kt
```diff
@@ -63,7 +63,7 @@ public fun <T: JsAny?> Deferred<T>.asPromise(): Promise<T> {
 @OptIn(ExperimentalWasmJsInterop::class)
 public fun <T: JsAny?> Promise<T>.asDeferred(): Deferred<T> {
     @Suppress("UNCHECKED_CAST", "UNCHECKED_CAST_TO_EXTERNAL_INTERFACE")
```
dkhalanskyjb: These suppressions can be removed.

dkhalanskyjb (changes requested):
Please target the `develop` branch. See https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md for other details on how to contribute to `kotlinx.coroutines`.

dkhalanskyjb (approved):

PR #4741, opened 2026-09-02
Fix external interface is check community dev

JSMonk:


fzhinkin:
For the context: without this change, compilation fails with Kotlin 2.5:
```
JS_NO_RUNTIME_FORBIDDEN_AS_CAST] Runtime type casts ('as'/'as?') are forbidden for '@JsNoRuntime' interfaces. Consider using `unsafeCast` instead.
```

Issue #4358, opened 2025-02-19
Mark `CompletionStage<T>.await()` as `@DelicateCoroutinesApi`


dovchinnikov:
This is unexpected that `await` cancels the `CompletableFuture` on coroutine cancellation. Since it's named like `Deferred.await`, it's expected to behave the same way without requiring to convert to `Deferred` first.

dovchinnikov:
This could be already done before the decision on #4329.

dkhalanskyjb:
I'd prefer to decide on #4329 in the near future. If we decide to deprecate `CompletionStage.await` anyway, then forcing everyone into an opt-in first is going to create unnecessary friction.

PR #4742, opened 2026-09-02
Rename the `await` overloads that cancel the source to `consume`

dkhalanskyjb:
Fixes #4329
Fixes #4358

See the issue for the rationale.

### DIFF
--- a/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
+++ b/integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
@@ -232,18 +232,24 @@ public fun <T> Deferred<T>.asListenableFuture(): ListenableFuture<T> {
 }
 
 /**
- * Awaits completion of `this` [ListenableFuture] without blocking a thread.
+ * Await the result of this [ListenableFuture] without blocking the thread or cancel it if the result is not needed.
  *
- * This suspend function is cancellable.
+ * If the future completes, this function either returns its result [T] or throws the exception thrown by the task,
+ * depending on whether the future was successful.
+ * However, if the call to [consume] is cancelled before the future completes,
+ * the future also gets cancelled.
  *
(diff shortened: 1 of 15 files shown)


integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -232,18 +232,24 @@ public fun <T> Deferred<T>.asListenableFuture(): ListenableFuture<T> {
...
- *
- * This suspend function is cancellable.
- *
- * If the [Job] of the current coroutine is cancelled while this suspending function is waiting, this function
- * stops waiting for the future and immediately resumes with [CancellationException][kotlinx.coroutines.CancellationException].
- *
- * This method is intended to be used with one-shot Futures, so on coroutine cancellation, the Future is cancelled as well.
- * If cancelling the given future is undesired, use [Futures.nonCancellationPropagating] or
- * [kotlinx.coroutines.NonCancellable].
+ * Await the result of this [ListenableFuture] without blocking the thread or cancel it if the result is not needed.
```
murfel, line 235: "cancel if the result is not needed" reads a bit too unspecific, albeit brief. Possibly the full form would work better: "Cancel it if the calling coroutine is cancelled while awaiting".

integration/kotlinx-coroutines-guava/src/ListenableFuture.kt
```diff
@@ -232,18 +232,24 @@ public fun <T> Deferred<T>.asListenableFuture(): ListenableFuture<T> {
...
+ *
+ * If the future completes, this function either returns its result [T] or throws the exception thrown by the task,
+ * depending on whether the future was successful.
+ * However, if the call to [consume] is cancelled before the future completes,
```
murfel, line 239: if the coroutine calling [consume] is cancelled

PR #4744, opened 2026-09-03
Optimize `CoroutineContext.foldCopies` and `newCoroutineContext`

chaoren:
Optimize `CoroutineContext.foldCopies` and `newCoroutineContext` by reusing the original context to avoid unnecessary allocations and skipping redundant traversals.

Previously, if either context contained a `CopyableThreadContextElement` (CTCE), `foldCopies` reconstructed the entire context tree from scratch starting from `EmptyCoroutineContext`. This allocated new context nodes for every element, even unmodified ones, and performed key lookups in the child context on every step.

Key changes:
- Reuse the original context as the accumulator, preserving unmodified elements and avoiding allocations from rebuilding context trees from scratch.
- When the child context has no CTCEs, copy the parent's CTCEs in a single fold and append the child context directly, avoiding key lookups.
- When both contexts contain CTCEs, fold and merge directly onto the existing context instead of starting from `EmptyCoroutineContext`.
- For `withContext`, fold the added context directly onto the parent context, copying only new or merged CTCEs.
- Fast-path `hasCopyableElements()` for single-element contexts to avoid allocating a `fold` lambda.
- Add `CoroutineContextMergingBenchmark` to `:benchmarks`.

### Benchmark Results

Executed on JDK 26 via JMH (`@Warmup(8, 1s)`, `@Measurement(10, 1s)`). Timings in `ns/op`:

#### `withContext` (Context Switching)

| Scenario | raw `+` | Before | After | Speedup |
| :--- | :---: | :---: | :---: | :---: |
| `+ EMPTY` | 1.1 | 1.2 | 1.1 | 1.1x |
| `+ E` | 43.3 | 42.7 | 41.4 | 1.0x |
| `+ CTCE` | 42.7 | 304.5 | 132.7 | **2.3x** |
| `CTCE + EMPTY` | 1.1 | 1.2 | 1.1 | 1.1x |
| `CTCE + E` | 42.7 | 41.6 | 42.0 | 1.0x |
| `CTCE + CTCE` | 43.7 | 331.9 | 126.9 | **2.6x** |
| `MERGE_HEAD` | 57.8 | 319.7 | 211.5 | **1.5x** |
| `MERGE_TAIL` | 19.8 | 279.2 | 93.5 | **3.0x** |

#### `launch` (Coroutine Creation)

| Scenario | Before | After | Speedup |
| :--- | :---: | :---: | :---: |
| `+ EMPTY` | 85.2 | 81.4 | 1.0x |
| `+ E` | 136.7 | 132.2 | 1.0x |
| `+ CTCE` | 541.7 | 307.8 | **1.8x** |
| `CTCE + EMPTY` | 862.4 | 285.6 | **3.0x** |
| `CTCE + E` | 1051.0 | 406.8 | **2.6x** |
| `CTCE + CTCE` | 1064.7 | 427.5 | **2.5x** |
| `MERGE_HEAD` | 654.6 | 392.8 | **1.7x** |
| `MERGE_TAIL` | 498.4 | 260.5 | **1.9x** |


dkhalanskyjb (changes requested):
On my computer with JDK 21, the benchmarks for `withContext()` + empty context actually got slower. Everything else did get improved or stayed the same, and `withContext(EmptyCoroutineContext)` isn't a priority, so this seems to be a strict improvement. `coroutineScope` and similar scoped builders that don't modify the context don't go through the same code path.

The code changes itself look good, but the new version is underdocumented. `fold`-heavy code isn't very readable, and comments can help to make sense of the general picture. In addition to that, some new code looks redundant: `appendContext === EmptyCoroutineContext || !appendContext.hasCopyableElements()` is equivalent to `appendContext.hasCopyableElements()`, and `hasCopyableElements` can be replaced with its `else` branch. Comments are needed to indicate that these are intentionally introduced fast paths and not editing artifacts.

chaoren:
@dkhalanskyjb I added some comments. Please take another look.

chaoren:
Friendly ping @dkhalanskyjb, could you take another look when you have a moment?

dkhalanskyjb (approved):
Looks good to me, thanks! @qwwdfsad wanted to review this, too, though, so I won't merge it for now.

benchmarks/src/jmh/kotlin/benchmarks/CoroutineContextMergingBenchmark.kt
```diff
@@ -0,0 +1,130 @@
...
+import kotlinx.coroutines.*
+import org.openjdk.jmh.annotations.*
+
+private data class E(val id: Int) : CoroutineContext.Element, CoroutineContext.Key<E> {
+    override val key = this
+}
+
+private data class CTCE(val id: Int, val value: Int = 1) :
+    CopyableThreadContextElement<Unit>, CoroutineContext.Key<CTCE> {
+    override val key = this
```
qwwdfsad: @chaoren, could you please elaborate on whether non-constant keys are something you rely on in production, or is it here for the sake of the benchmark?

qwwdfsad: Because if it's the former, then here is the snippet that used to work but changes the behaviour:

```
@OptIn(ExperimentalCoroutinesApi::class, DelicateCoroutinesApi::class)
class CopyForChildKeyTest2 {

    private data class Copyable(private val name: String) :
        CopyableThreadContextElement<String>, CoroutineContext.Key<Copyable> {

        override val key: CoroutineContext.Key<*> get() = this

        override fun updateThreadContext(context: CoroutineContext): String = name

        override fun restoreThreadContext(context: CoroutineContext, oldState: String) {}

        override fun copyForChild(): CopyableThreadContextElement<String> = Copyable("Copy($name)")

        override fun mergeForChild(overwritingElement: CoroutineContext.Element): CoroutineContext =
            Copyable("Merge($name)")
    }

    @Test
    fun testShowcase() {
        val scope = CoroutineScope(Copyable("original"))
        val childContext = scope.newCoroutineContext(EmptyCoroutineContext)
        val copyable = childContext.fold(emptyList<String>()) { acc, element ->
            if (element is CopyableThreadContextElement<*>) acc + element.toString() else acc
        }
        assertEquals(listOf("Copyable(copy(original))"), copyable)
    }
}

```

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -40,60 +40,67 @@ public actual fun CoroutineContext.newCoroutineContext(addedContext: CoroutineCo
...
 
+    // Slow-path: appendContext contains one or more CTCEs. We'll need two fold passes over both originalContext and
+    // appendContext to merge CTCEs found in both contexts, copy CTCEs found in one context but not the other, and
+    // append any remaining elements in appendContext to originalContext.
     var leftoverContext = appendContext
-    val folded = originalContext.fold<CoroutineContext>(EmptyCoroutineContext) { result, element ->
-        if (element !is CopyableThreadContextElement<*>) return@fold result + element
-        // Will this element be overwritten?
+    // Fold originalContext into itself, overwriting CTCEs with copies or merges while leaving other elements alone.
+    val folded = originalContext.fold(originalContext) { result, element ->
```
qwwdfsad, line 86: This is technically a breaking change.

Previously, merge could have returned `EmptyCoroutineContext`, meaning that "no such element, please remove it from the child's context".
Now there is no such option anymore.

Two questions:

1) @dkhalanskyjb are you okay with this change? If so, let's then follow up in a separate PR with a documentation update. Also, see the comment to the benchmark, if any of `copy`/`merge` return an element with a different key, all bets are off. We can document it is prohibited and even assert (will require @chaoren to look into that, though)
2) @chaoren is this change deliberate? Are you aware (I assume you are one of the biggest `CopyableThreadContextElement` clients) of merge returning `EmptyCoroutineContext`?

qwwdfsad (commented):
I am conceptually fine with a speedup, but it definitely requires a follow up -- either we should explicitly prohibit changing keys for copies on the fly or, if it's actually a usable feature, then it's a breaking change

benchmarks/src/jmh/kotlin/benchmarks/CoroutineContextMergingBenchmark.kt
chaoren: Sorry, that's just a mistake in the benchmark. We don't rely on non-constant keys. I'll update the benchmark.

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
chaoren: We only have a handful of `CopyableThreadContextElement` implementations. None of them ever return `EmptyCoroutineContext`. Most of them actually just return `copyForChild()`.

dkhalanskyjb: The exact scenario you're outlining is already prohibited, since `mergeForChild`'s documentation unambiguously states (https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-copyable-thread-context-element/merge-for-child.html):

> Returns a `CopyableThreadContextElement`

`EmptyCoroutineContext` isn't a valid value to return.

However, as far as I can tell, nothing can stop you from returning a `CopyableThreadContextElement` with a different key. The PR does change the behavior in that scenario: before, only the element with the key returned from `mergeForChild` would appear in the resulting context, but now, the original element would be preserved, too.

This is technically a breach of contract, as the docs also state:

> [...] to **replace** this CopyableThreadContextElement in the child coroutine's context

(emphasis mine)

A good find, though I'm indeed inclined to see this as a documentation issue.

chaoren:
I think we can preserve behavior by checking the key of the copied/merged context against the original element and remove the original if the key changed.

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
chaoren: I added a `replace` function to check if the replacement's key matches the original. This should restore the original behavior. How does this look?

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
```diff
@@ -40,62 +40,78 @@ public actual fun CoroutineContext.newCoroutineContext(addedContext: CoroutineCo
...
-    return folded + leftoverContext
 }
 
+/**
+ * Replaces the element associated with [key] in this context with [replacement].
+ *
+ * If [replacement] contains [key], appending it directly overwrites the old element. Otherwise (e.g. key changed or
+ * [replacement] is [EmptyCoroutineContext]), [key] is explicitly removed via [minusKey] first.
+ */
+private fun CoroutineContext.replace(key: CoroutineContext.Key<*>, replacement: CoroutineContext): CoroutineContext =
```
qwwdfsad, line 106: Style (mostly for the sake of future contributions): please avoid (especially LLM-produced) comments to private functions that are self-explanatory.

Here the whole KDoc is redundant

qwwdfsad (approved):
I have one minor comment and it's good to go. Thanks!

Initiated a CI build as well

kotlinx-coroutines-core/jvm/src/CoroutineContext.kt
chaoren: Done. Removed the KDoc.

PR #4748, opened 2026-09-04
Prevent SMARTCAST_IMPOSSIBLE error on 2.4 update

fzhinkin:
Updating Kotlin version to 2.4 will inevitably cause compilation error in code that currently compiles without any warnings.


lppedd:
@fzhinkin just for the sake of my own knowledge, what's the Kotlin change that causes this? I see there is `assertNotNull` immediately above that.

Issue #4743, opened 2026-09-03
[Bug] Coroutine Workers can lose submitted tasks when the keepalive timeout is exceeded


knisht:
### What is happening? What should have happened instead?

If a coroutine worker decides to terminate itself due to exceeded keepalive timeout, it does not inspect its local queue. This can lead to lost coroutines: the attached reproducer shows a flaky behavior where some coroutines get lost -- they are stuck either in `CREATED` or `SUSPENDED` state.

### Reproducer

```kotlin

import kotlinx.coroutines.*
import org.junit.Test
import kotlin.test.assertEquals
import kotlin.time.Duration
import kotlin.time.Duration.Companion.milliseconds
import kotlin.time.Duration.Companion.seconds

class CustomKeepaliveTest {
    @Test
    fun testFixtureDispatcherRunsEveryTaskUnderSameLoad() {
        System.setProperty("kotlinx.coroutines.scheduler.keep.alive.sec", "1")
        System.setProperty("kotlinx.coroutines.scheduler.core.pool.size", "2")
        val lost = countLostTasks(cpu = Dispatchers.Default, blocking = Dispatchers.IO)
        assertEquals(0, lost)
    }
}

private const val ITERATIONS = 10

private fun countLostTasks(cpu: CoroutineDispatcher, blocking: CoroutineDispatcher): Int {
    @Suppress("RAW_SCOPE_CREATION")
    val detached = CoroutineScope(SupervisorJob() + cpu)
    try {
        var lost = 0
        runBlocking {
            repeat(ITERATIONS) { iteration ->
                println("Iteration #$iteration begin; lost: $lost")
                val finished = withTimeoutOrNull(12.seconds) {
                    coroutineScope {
                        val victim = async(cpu) {
                            val child = detached.async { 1 }
                            withContext(blocking) { Thread.sleep(1_010) }
                            child.await()
                        }
                        repeat(2) {
                            launch(cpu) { spinFor(4_000.milliseconds) }
                        }
                        victim.await()
                    }
                }
                if (finished == null) {
                    lost++
                }
                println("Iteration #$iteration done; lost: $lost")
            }
        }
        return lost
    } finally {
        detached.cancel()
    }
}

private fun spinFor(duration: Duration) {
    val end = System.nanoTime() + duration.inWholeNanoseconds
    while (System.nanoTime() < end) {
        Thread.onSpinWait()
    }
}
```

vsalavatov:
[...]

PR #4750, opened 2026-09-07
Properly hand-off tasks from the local queue during the termination s…

qwwdfsad:
…equence

Fixes #4743

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -416,7 +416,7 @@ internal class CoroutineScheduler(
         }
     }
 
-    fun createTask(block: Runnable, taskContext: TaskContext): Task {
+    private fun createTask(block: Runnable, taskContext: TaskContext): Task {
         val nanoTime = schedulerTimeSource.nanoTime()
         if (block is Task) {
             block.submissionTime = nanoTime
@@ -893,6 +893,16 @@ internal class CoroutineScheduler(
                 workers.setSynchronized(lastIndex, null)
             }
             state = WorkerState.TERMINATED
+            /*
+             * At this point, worker cannot be observed by other threads (though its queue can be in the process of stealing from).
+             * It can be the case that the local queue is not empty if during the keep-alive period all workers were
+             * busy with their own work and didn't have a chance to steal any task, there are no CPU permits and local queue
+             * contains only CPU work
+             */
+            while (true) {
+                val task = localQueue.poll() ?: break
+                dispatch(task)
+            }
         }
 
         fun findTask(mayHaveLocalTasks: Boolean): Task? {
--- a/kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
@@ -16,7 +16,7 @@ class CoroutineSchedulerWorkerTerminationTest : TestBase() {
     @Test(timeout = 20_000L)
     fun testWorkerTerminationDuringOversubscription() {
         repeat(25 * stressTestMultiplierSqrt) { iteration ->
-            testWorkerTerminationDuringOversubscription(iteration)
+            doTest()
         }
     }
 
@@ -28,7 +28,7 @@ class CoroutineSchedulerWorkerTerminationTest : TestBase() {
      * - No one to process it
      * - Termination of such worker
      */
-    private fun testWorkerTerminationDuringOversubscription(iteration: Int) {
+    private fun doTest() {
         val corePoolSize = 2
         val tasksCount = corePoolSize + 1
         val dispatcher = SchedulerCoroutineDispatcher(
--- a/kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
+++ b/kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
@@ -1,0 +1,95 @@
+package kotlinx.coroutines.scheduling
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.testing.TestBase
+import kotlinx.coroutines.testing.stressTestMultiplier
+import kotlinx.coroutines.testing.stressTestMultiplierSqrt
+import org.junit.Test
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.milliseconds
+
+class CoroutineSchedulerWorkerTerminationTest : TestBase() {
+
+    // Note: right now test is timing sensitive because it awaits for the worker termination
+    @Test(timeout = 20_000L)
+    fun testWorkerTerminationDuringOversubscription() {
+        repeat(25 * stressTestMultiplierSqrt) { iteration ->
+            testWorkerTerminationDuringOversubscription(iteration)
+        }
+    }
+
+    /**
+     * Emulates oversubscription + worker termination pattern.
+     * What we want to recreate:
+     * - IO worker
+     * - CPU task in local queue
+     * - No one to process it
+     * - Termination of such worker
+     */
+    private fun testWorkerTerminationDuringOversubscription(iteration: Int) {
+        val corePoolSize = 2
+        val tasksCount = corePoolSize + 1
+        val dispatcher = SchedulerCoroutineDispatcher(
+            corePoolSize,
+            maxPoolSize = tasksCount,
+            idleWorkerKeepAliveNs = TimeUnit.MILLISECONDS.toNanos(50)
+        )
+
+        val workers = ConcurrentHashMap.newKeySet<Thread>()
+        val allWorkersCreated = CountDownLatch(tasksCount)
+        val blockingTasksBlocker = CountDownLatch(1)
+        val cpuThreads = ConcurrentHashMap.newKeySet<Thread>()
+        val cpuThreadsReachedThis = CountDownLatch(corePoolSize)
+        val cpuThreadsBlocked = CountDownLatch(1)
+        val cpuThreadsFinished = CountDownLatch(tasksCount)
+        val cpuScope = CoroutineScope(SupervisorJob() + dispatcher)
+
+        fun recordWorker() {
+            if (workers.add(Thread.currentThread())) allWorkersCreated.countDown()
+        }
+        repeat(tasksCount) {
+            cpuScope.launch {
+                // It would've been so much easier if a dispatch of a CPU task from a blocking thread
+                // wouldn't been redirecting everything to the global queue.
+                // Anyway: spam tasks until there are three blocking tasks properly created and blocking the whole scheduler.
+                while (allWorkersCreated.count > 0) {
+                    dispatcher.dispatchWithContext(Runnable {
+                        recordWorker()
+                        blockingTasksBlocker.await()
+                    }, BlockingContext, false)
+
+                    yield()
+                }
+
+                /*
+                 * Here we have the following state (test-wise, not line-wise), dear reader:
+                 * - 3 out of 3 threads are blocked fully in the IO (BlockingContext) task
+                 * - Each of these three threads has a CPU task in its local queue
+                 * - These CPU tasks are, well, this very line of code
+                 * - We release these blocking tasks, making this two out of three (!) executing
+                 * - They record their thread and notify test coordinator
+                 * - Test coordinator waits for the thread to shut down
+                 * - Then releases 2 CPU tasks, 3rd one should execute
+                 */
+                cpuThreads += Thread.currentThread()
+                cpuThreadsReachedThis.countDown()
+                cpuThreadsBlocked.await()
+                cpuThreadsFinished.countDown()
+            }
+        }
+
+        allWorkersCreated.await() // Wait all
+        blockingTasksBlocker.countDown() // Ublock all blocking
+        cpuThreadsReachedThis.await() // Wait all
+
+        // Quickly terminates, thanks keep alive timeout
+        val retiringWorker = workers.single { it !in cpuThreads }
+        retiringWorker.join()
+
+        cpuThreadsBlocked.countDown() // Unblock
+        cpuThreadsFinished.await() // Wait
+        dispatcher.close()
+    }
+}


kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.milliseconds
+
+class CoroutineSchedulerWorkerTerminationTest : TestBase() {
+
+    // Note: right now test is timing sensitive because it awaits for the worker termination
+    @Test(timeout = 20_000L)
+    fun testWorkerTerminationDuringOversubscription() {
```
qwwdfsad, line 17: The fix is trivial; it's the test that I spent most of my time on.

I suggest reviewing this carefully, I've reiterated on it quite a few times

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+                 */
+                cpuThreads += Thread.currentThread()
+                cpuThreadsReachedThis.countDown()
+                cpuThreadsBlocked.await()
+                cpuThreadsFinished.countDown()
+            }
+        }
+
+        allWorkersCreated.await() // Wait all
+        blockingTasksBlocker.countDown() // Ublock all blocking
```
vsalavatov, line 84: small typo

dkhalanskyjb: Brand recognition for https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -893,6 +893,16 @@ internal class CoroutineScheduler(
...
             state = WorkerState.TERMINATED
+            /*
+             * At this point, worker cannot be observed by other threads (though its queue can be in the process of stealing from).
+             * It can be the case that the local queue is not empty if during the keep-alive period all workers were
+             * busy with their own work and didn't have a chance to steal any task, there are no CPU permits and local queue
+             * contains only CPU work
+             */
+            while (true) {
+                val task = localQueue.poll() ?: break
+                dispatch(task)
```
dkhalanskyjb, line 904: By going through the `dispatch` logic, we'll observe the following effects:
```kotlin
trackTask()
task.submissionTime = nanoTime
globalCpuQueue.addLast(task)
signalCpuWork()
```
1) Won't calling `trackTask()` twice lead to double-counting it? It's not that important, as it's just test code, but it is still an eyesore.
2) Updating the `nanoTime` reduces the "staleness" of the task, which does have some effect on the decision to steal it or not. Isn't doing that strictly worse than not doing that?
3) Using `globalCpuQueue` does seem reasonable in this case!
4) The significance of `signalCpuWork()` isn't clear to me. ~We're assuming that all CPU workers are taken, so no need to signal them.~ EDIT: well, assuming a nasty race, we should signal the work, though. If this worker waits for an eternity to do its cleanup, there may be no CPU workers by the time the tasks enter the global queue.

To sum it up: won't it be strictly better, as well as semantically clearer, to do this instead?
```kotlin
localQueue.offloadAllWorkTo(globalCpuQueue)
signalCpuWork() // EDIT
```

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
+package kotlinx.coroutines.scheduling
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.testing.TestBase
+import kotlinx.coroutines.testing.stressTestMultiplier
+import kotlinx.coroutines.testing.stressTestMultiplierSqrt
+import org.junit.Test
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CountDownLatch
```
dkhalanskyjb, line 9: Style: we have our own `kotlinx.coroutines.testing.CountDownLatch` now.

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
+package kotlinx.coroutines.scheduling
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.testing.TestBase
+import kotlinx.coroutines.testing.stressTestMultiplier
+import kotlinx.coroutines.testing.stressTestMultiplierSqrt
+import org.junit.Test
```
dkhalanskyjb, line 7: Style:
```suggestion
import kotlin.test.*
```

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+     */
+    private fun doTest() {
+        val corePoolSize = 2
+        val tasksCount = corePoolSize + 1
+        val dispatcher = SchedulerCoroutineDispatcher(
+            corePoolSize,
+            maxPoolSize = tasksCount,
+            idleWorkerKeepAliveNs = TimeUnit.MILLISECONDS.toNanos(50)
+        )
+
```
dkhalanskyjb, line 39: From https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md :
> Do not add extra newlines in function bodies: if you feel that blocks of code should be logically separated, then separate them with a comment instead.

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: 1) Might be the case
2) Shouldn't matter because the task will be offloaded to the global CPU queue, where timings do not matter

>To sum it up: won't it be strictly better, as well as semantically clearer, to do this instead?

That was my first idea, but I decided against it. The problem is that the moment you start applying tailored solutions (like the originally proposed snippet with a plain submit), it's easy to miss a subtle invariant (such as signalling). Taking this is a *very* occasional code path, I decided to keep it absolutely straightforward and, so to speak, to empty the kettle -- thus the regular dispatch that does a bit of extra.

dkhalanskyjb: Your proposed solution extracts a task without executing it and then adds it back, and that is not straightforward at all. Now, the scheduler has two separate paths through which a task can leave it altogether (before being immediately re-added)—one of which is very occasional, indeed. To see that out-of-band task extraction from the scheduler is non-obvious, consider that for blocking tasks, we would also need to call `decrementBlockingTasks`.

To me, moving a task between queues is much more natural, even if the case here is a bit weird in that it *pushes* a task to some other queue, whereas in the rest of the scheduler, tasks are *pulled*.

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.milliseconds
+
+class CoroutineSchedulerWorkerTerminationTest : TestBase() {
+
+    // Note: right now test is timing sensitive because it awaits for the worker termination
+    @Test(timeout = 20_000L)
+    fun testWorkerTerminationDuringOversubscription() {
+        repeat(25 * stressTestMultiplierSqrt) { iteration ->
```
dkhalanskyjb, line 18: Style: unused.

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Test
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.milliseconds
+
+class CoroutineSchedulerWorkerTerminationTest : TestBase() {
+
+    // Note: right now test is timing sensitive because it awaits for the worker termination
+    @Test(timeout = 20_000L)
```
dkhalanskyjb, line 16: Style: clearer to time out in `cpuThreadsFinished.await()`, to highlight which exact part is expected to hang on a regression (and which hang points to some different issue).

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+                    dispatcher.dispatchWithContext(Runnable {
+                        recordWorker()
+                        blockingTasksBlocker.await()
+                    }, BlockingContext, false)
+
+                    yield()
+                }
+
+                /*
+                 * Here we have the following state (test-wise, not line-wise), dear reader:
```
dkhalanskyjb, line 67:
```suggestion
                 * Before this line executes in any of the coroutines, the state we want is:
```

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+            idleWorkerKeepAliveNs = TimeUnit.MILLISECONDS.toNanos(50)
+        )
+
+        val workers = ConcurrentHashMap.newKeySet<Thread>()
+        val allWorkersCreated = CountDownLatch(tasksCount)
+        val blockingTasksBlocker = CountDownLatch(1)
+        val cpuThreads = ConcurrentHashMap.newKeySet<Thread>()
+        val cpuThreadsReachedThis = CountDownLatch(corePoolSize)
+        val cpuThreadsBlocked = CountDownLatch(1)
+        val cpuThreadsFinished = CountDownLatch(tasksCount)
```
dkhalanskyjb: Style: seems unnecessary, since we can instead await the coroutines. `withTimeout` will be able to handle the timeout, with no JUnit tricks.

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+                        blockingTasksBlocker.await()
+                    }, BlockingContext, false)
+
+                    yield()
+                }
+
+                /*
+                 * Here we have the following state (test-wise, not line-wise), dear reader:
+                 * - 3 out of 3 threads are blocked fully in the IO (BlockingContext) task
+                 * - Each of these three threads has a CPU task in its local queue
```
dkhalanskyjb, line 69:
```suggestion
                 * - Each of these three threads has a CPU task in its local queue
                 *   (may not be true, the CPU task may be in the global queue, making the test non-deterministic)
```

dkhalanskyjb (commented):
The change in general and the test in particular both make sense. The test could benefit from differentiating between the events that is *sure* to happen and the state that *may* be observed with some probability—right now, the comments state, if we are to believe them, that the test deterministically detects the issue.

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
murfel: It's not equivalent to the `java.util.concurrent.CountDownLatch` though, ours is not reusable. We made it because we needed a quick hack to port the tests and not spend tons of time ~~verifying the port of the j.u.c implementation~~ (edit: we can still make it reusable, without importing the whole j.u.c data structure, but it also required a bit of careful pondering, so I didn't do it because we didn't need it). I will document it now.

dkhalanskyjb: Could it be that you are confusing a barrier with a latch? `CountDownLatch` isn't reusable anyway, you can only fully lower it once.

murfel: Oh, of course.

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
murfel: ok

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
qwwdfsad: I would prefer not to touch it:

* It doesn't change anything in this particular PR, it is a mechanical change for the sake of a mechanical change. If we wish to use _only_ our own test primitives in our own tests, I'd prefer to do it in a more consolidated manner
* For the code (the test, invariants) that is already subtle enough, I would strongly, strongly prefer to use a primitive that is 100% battle-tested and has as many precise (semantics-wise and implementation-flavour-wise) details as possible. ATM, `kotlinx.coroutines.testing.CountDownLatch` is not deliberately tested, and I'd like to avoid clashing correctness issues in `CoroutineScheduler` with potential correctness issues in k.c.t.CDL

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: >localQueue.offloadAllWorkTo(globalCpuQueue)
>signalCpuWork() 

Here we offload _some_ number of tasks, but wake up only a single worker, leading to the same problem in disguise. The subtle invariant that is easy to miss and the thing I don't want to wrestle with.

>To see that out-of-band task extraction from the scheduler is non-obvious, consider that for blocking tasks

Yep, but blocking tasks cannot end up in this queue at this point.

The trade-off for me is the following -- on the one hand, we reuse `dispatch` machinery, which can interfere with tests (albeit unlikely) and also comes with a bit of potential overhead, on the other -- we have to maintain a set of subtle invariants in two places instead of one, and at the point of fixing, and two attempts out of three missed an invariant and suggested a fix with an actual bug. Anyway, that's two unpleasant things to choose from that have a potato-potato flavour. I prefer the one with `dispatch`

qwwdfsad: Maybe the winning move is not to play this game -- instead of redispatching, do something like `  if (localQueue.size != 0) return` in `tryTerminateWorker`. WDYT?

dkhalanskyjb: > Yep, but blocking tasks cannot end up in this queue at this point.

That's beside the point, though, I merely tried to highlight that we must maintain two places with tricky invariants in any case: either on entering the global queue or on leaving it. `dispatch` doesn't save us from the problem in general. The fact that `dispatch` has more overhead and may interfere with tests is what causes me to prefer not doing that.

> `return`

I don't understand immediately how that would work. A thread doesn't get a CPU permit, doesn't have anything else to do, gets bored, and wants to terminate. Then it notices it still has work, wakes up—and then what? Keep parking and unparking until it happens to snatch a CPU permit or someone steals all its tasks?

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
dkhalanskyjb: The idea was to do it for the sake of having a wider area that tests our common-code thread parking/unparking in extreme scenarios like this one.

Though that's not important on the JVM, as we're only wrapping the usual JVM thread parking. So, it's fine to leave this as is.

> ATM, `kotlinx.coroutines.testing.CountDownLatch` is not deliberately tested

It's somewhat tested: https://github.com/Kotlin/kotlinx-atomicfu/blob/master/atomicfu/src/concurrentTest/kotlin/kotlinx/atomicfu/locks/LatchTest.kt

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: >I don't understand immediately how that would work. 

While the system makes no CPU progress (e.g. all CPU permits are blocked), a thread without a permit that has some local CPU tasks in its queue will be kept alive (waking up every `idleWorkerKeepAliveNs`) until system makes some progress

kotlinx-coroutines-core/jvm/test/scheduling/CoroutineSchedulerWorkerTerminationTest.kt
```diff
@@ -0,0 +1,88 @@
...
+                    dispatcher.dispatchWithContext(Runnable {
+                        recordWorker()
+                        blockingTasksBlocker.await()
+                    }, BlockingContext, false)
+                    yield()
+                }
+                /*
+                 * Before this line executes in any of the coroutines, the state we want is:
+                 * - 3 out of 3 threads are blocked fully in the IO (BlockingContext) task
+                 *- Each of these three threads has a CPU task in its local queue
```
dkhalanskyjb:
```suggestion
                 * - Each of these three threads has a CPU task in its local queue
```

dkhalanskyjb (approved):

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: Seems like it would be a preferable fix only after #4752

PR #4751, opened 2026-09-07
Remove mayHaveLocalTasks from CoroutineScheduler

qwwdfsad:
This property was a completely redundant 'performance optimization' that saves two uncontended atomic loads on a path full of atomic reads and writes. Its presence also introduces a subtle inconsistency into the liveness guarantee (went unnoticed for six years, unlikely to ever affect a thing).

Minor risk, no reward, harder to reason about the code. Let's just remove it

### DIFF
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -717,7 +717,7 @@ internal class CoroutineScheduler(
                     executeTask(task)
                     continue
                 } else {
-                    mayHaveLocalTasks = false
+                    mayHaveLocalTasks = localQueue.size > 0
                 }
                 /*
                  * No tasks were found:
--- a/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
+++ b/kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
@@ -365,7 +365,7 @@ internal class CoroutineScheduler(
         globalCpuQueue.close()
         // Finish processing tasks from globalQueue and/or from this worker's local queue
         while (true) {
-            val task = currentWorker?.findTask(true)
+            val task = currentWorker?.findTask()
                 ?: globalCpuQueue.removeFirstOrNull()
                 ?: globalBlockingQueue.removeFirstOrNull()
                 ?: break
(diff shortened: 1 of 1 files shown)


qwwdfsad:
The first commit fixes a bug. The second (and the PR description) is about removing the actual root cause.

I have a stress test, of course, but it's very weirdly specific, and I see no value in committing it as there are no more `mayHaveLocalTasks`.

The ~rough shape of the test:
* 2 permits, 3 threads max total
* T1 is always busy (because of our min 2 threads rule)
* T2 slams CPU tasks into its queue
    * At some point, T2 submits a blocking task. It triggers the creation of T3
    * T2 executes this blocking task (meanwhile T3 steals things occasionally)
    * This blocking task creates P (as in poison) -- a CPU task, submitted to the global queue, that T3 takes on
    * Blocking task over. T2 goes to sleep -- it tries to poll its local tasks, but cannot do it because it doesn't have a permit. **This is where mayHaveLocalTasks invariant breaks**.
    * T2 goes to sleep
* T3 executes P. P produces a blocking task to T3 local queue. T3 wakes up T2, jumps on a blocking task, releases the permit]
* T2 acquires a permit and starts looking for work. It cannot find it _because it never looks into its local queue_. Fail

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
```diff
@@ -703,21 +702,16 @@ internal class CoroutineScheduler(
...
-                val task = findTask(mayHaveLocalTasks)
+                val task = findTask()
                 // Task found. Execute and repeat
                 if (task != null) {
                     rescanned = false
                     minDelayUntilStealableTaskNs = 0L
                     executeTask(task)
                     continue
-                } else {
-                    mayHaveLocalTasks = false
```
murfel, line 720: Was supposed to be: `mayHaveLocalTasks = localQueue.size > 0` (from the first commit)

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
murfel, line 899: I guess it should have also propagated `mayHaveLocalTasks` here.

murfel (approved):

murfel:
The initial issue and the suggested scenario make sense.

The removal makes sense, too. I'm only a tiny bit afraid that this will introduce a slightly different pattern which could surface some other scheduler bug, so let's be on lookout.

kotlinx-coroutines-core/jvm/src/scheduling/CoroutineScheduler.kt
qwwdfsad: Good catch, I think that's right

Issue #4580, opened 2025-12-03
The test coroutine scheduler isn't closed when a test ends


dkhalanskyjb:
**Describe the bug**

The `kotlinx-coroutines-test` test framework may fail to run some code to completion if a test dispatcher was passed to the system under test.
The recommended approach is to pass a `CoroutineScope` to the system under test, which allows the test to know that some operations in the system under test may still be running. When only a dispatcher is passed, a new task may arrive there by the time `runTest` is no longer processing the tasks in the test scheduler.

**Provide a Reproducer**

```kotlin
import kotlinx.coroutines.test.*
import kotlinx.coroutines.*
import kotlin.time.Duration.Companion.milliseconds
import kotlin.concurrent.thread

suspend fun waitForSomethingAsynchronousToHappen() {
    val latch = CompletableDeferred<Unit>()
    thread {
        Thread.sleep(100)
        latch.complete(Unit)
        println("Latch lowered")
    }
    latch.await()
}

fun runSomeTestThing(dispatcher: CoroutineDispatcher) {
    GlobalScope.launch(dispatcher) {
        try {
            println("Starting a new task")
            waitForSomethingAsynchronousToHappen()
        } finally {
            println("Cleaning up")
        }
    }
}

fun main() {
    val dispatcher = StandardTestDispatcher()
    TestScope(dispatcher).runTest {
        runSomeTestThing(dispatcher)
    }
    println("Test finished")
    Thread.sleep(300)
    println("Tired of waiting")
}
```

prints

```
Starting a new task
Test finished
Latch lowered
Tired of waiting
```

The cleanup never runs.

oaksprout:
Would you want this one done? — Fix a bug in kotlinx.coroutines' test framework where a background task's cleanup code can silently never run if the system under test was only handed a `CoroutineDispatcher` (not a `CoroutineScope`) to launch on. If yes, we'll do it and open a PR; if not, say so (or just ignore this) and we won't touch it.

How it works: an AI agent does the work against your write-up; it's independently checked in a clean container (your test suite, network off) before anything reaches you; a human reviews it, submits it, and answers the review. You keep full control — nothing lands unless you merge it.

Recent examples of the same flow on other projects: https://github.com/pypa/pip-audit/pull/1089, https://github.com/sigstore/sigstore-python/pull/1845, https://github.com/openstreetmap/iD/pull/12631.

[...]

Issue #4382, opened 2025-03-11
`withContext` ingnores cancellation from `withTimeout` when the dispatcher cannot execute the body


zuevmaxim:
**Describe the bug**

The following code hangs, while I expected it to output Timeout. When I uncomment the commented code, it works as I expected, so `withContext` handles timeouts differently. I wonder if it can be handled correctly without a deadlock here.

**Provide a Reproducer**

```
package kt.coroutines

import kotlinx.coroutines.*
import kotlinx.coroutines.channels.Channel
import java.util.concurrent.Executors
import kotlin.coroutines.resume

fun main() {
    val executor = Executors.newSingleThreadExecutor { r -> Thread(r, "MyThread") }
    val dispatcher = executor.asCoroutineDispatcher()
    runBlocking(dispatcher) {
        runBlocking { // block on the dispatcher thread
            val c = Channel<String>()
            launch(Dispatchers.Default) {
                try {
                    withTimeout(1000) {
//                        suspendCancellableCoroutine { cont ->
//                            executor.submit {
//                                c.trySend("Switched to dispatcher")
//                                cont.resume(Unit)
//                            }
//                        }
                        withContext(dispatcher) { // cannot switch to dispatcher as it is locked, but can it timeout?
                            c.send("Switched to dispatcher")
                        }
                    }
                } catch (_: TimeoutCancellationException) {
                    c.send("Timeout")
                }
            }
            println(c.receive())
        }
    }
    executor.shutdown()
}
```

version `1.9.0`

Issue #2818, opened 2021-07-10
MutableSharedFlow rendezvous


Dominaezzz:
I expected this bit of code to deadlock. (The `onEach` is a workaround for #2817)
```kotlin
val stream = MutableSharedFlow<Unit>()
stream/* .onEach {} */.buffer(Channel.RENDEZVOUS).produceIn(this)
yield()
stream.emit(Unit)
// stream.emit(Unit)
```

This did not deadlock until I commented out the second emission.

Is this by design? The docs don't seem to suggest this.

> [emit][MutableSharedFlow.emit] call to such a shared flow suspends until all subscribers receive the emitted value

I haven't received anything from the channel so this shouldn't happen.

I need this behaviour so I need to decide between using a workaround or a re-design?
If this won't be fixed then I'll redesign but if it will be fixed/implemented then I justify using a workaround.

## Mini investigation
So API aside, I had a look at the code and the culprit seems to be in `SharedFlow.collect`.
https://github.com/Kotlin/kotlinx.coroutines/blob/65e1b84f742a4fff14c67a11a873a4977a0ba539/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt#L343-L352 .

`tryTakeValue`/`awaitValue` "pops" the value out of the buffer for the current subscriber, (which effectively counts as the receipt the docs talked about) then the value is emitted to the collector.

Looking at `tryTakeValue`.
https://github.com/Kotlin/kotlinx.coroutines/blob/65e1b84f742a4fff14c67a11a873a4977a0ba539/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt#L593-L609

In theory, that could be changed to a two step process, one step to peek the value in the buffer (i.e. `getPeekedValueLockedAt`), then the value is emitted and another step to pop/release the value in the buffer (i.e. slot index increment and `updateCollectorIndexLocked`).

It means locking twice instead of once but it's probably best to measure than listen to whatever I was about hypothesise here lol.

vsartiushkin:
Thank you for the report and the reproducer!
I have reproduced the issue with Kotlin 2.3.21 and Library 1.11.0 with the following code:
```kotlin
val stream = MutableSharedFlow<Unit>()
stream.onEach {}.buffer(Channel.RENDEZVOUS).produceIn(this) 
yield()
stream.emit(Unit) // Expected to suspend at the first emit
stream.emit(Unit) // Actually suspends at the second emit
```

PR #4753, opened 2026-09-09
Add behaviour pinning tests for the known issues

vsartiushkin:
This PR adds tests that pin the current behaviour in cases from the known but not yet addressed issues.

The goal is to monitor on each build if the proposed changes have altered the behaviour in any of these cases.
This should help identify if any other issue was silently fixed by the changes, or whether it may unwillingly affect a known case. 

The first test batch adds tests that pin the cases from these issues:
at ‎`kotlinx-coroutines-core/common/test/pinnedBugs/`
#1578
#2817
#2818
#4382
#4383
#4698
at `kotlinx-coroutines-test/common/test/pinnedBugs`
#4580
 
~~This is a draft PR to request feedback on the idea and its implementation.~~


dkhalanskyjb:
Notably, the two hanging tests don't actually require spawning a new thread. They are not *blocking*, they demonstrate coroutines that aren't making *progress*. This is good for us, as coroutines can be simply dropped and garbage-collected (don't do that in production)! Here's an efficient way to leak a coroutine: https://pl.kotl.in/Xp7kA5fgV. We try making "the next step" for a hundred times, and if that's not enough, we assume it's just stuck, then.

kotlinx-coroutines-core/jvm/test/pinnedBugs/Gh1578PinnedBugTest.kt
```diff
@@ -0,0 +1,47 @@
...
+ * runBlocking hangs if it has a child that never completes
+ */
+class Gh1578PinnedBugTest : TestBase() {
+
+    @Test
+    fun testRunBlockingHangsWithManualSupervisorJobChild() {
+        val completed = AtomicBoolean(false)
+        val thread = Thread {
+            try {
+                runBlocking {
```
dkhalanskyjb: Can be replaced with a `coroutineScope { }`, and the problem will still reproduce. Then, the same `spinAwaitingCompletion` approach can be utilized.

kotlinx-coroutines-core/jvm/test/pinnedBugs/Gh1578PinnedBugTest.kt
```diff
@@ -0,0 +1,47 @@
+package kotlinx.coroutines.pinnedBugs
+
+import kotlinx.coroutines.testing.*
+import kotlinx.coroutines.*
+import java.util.concurrent.atomic.AtomicBoolean
```
dkhalanskyjb: We're going for Kotlin atomics these days: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.concurrent.atomics/-atomic-boolean/
That would allow making moving this test to `common` (though I believe `AtomicBoolean` isn't even necessary here).

kotlinx-coroutines-core/jvm/test/pinnedBugs/Gh4698PinnedBugTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    private val brokenDispatcher = object : CoroutineDispatcher() {
+        override fun dispatch(context: CoroutineContext, block: Runnable) = TODO("dispatcher is broken")
+    }
+
+    @Test
+    fun testAtomicStartOnBrokenDispatcherHangsInsteadOfFailingFast() {
+        val threadName = "gh4698-hung-thread"
+        ignoreLostThreads(threadName) // The thread hangs and is impossible to terminate
+
```
dkhalanskyjb: Style, from https://github.com/Kotlin/kotlinx.coroutines/blob/master/CONTRIBUTING.md:

> Follow the [Kotlin Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html). Use 4 spaces for indentation. Do not add extra newlines in function bodies: if you feel that blocks of code should be logically separated, then separate them with a comment instead.

This looks like a simple code style rule that shouldn't take too much effort to follow. No need to make these tests look pretty beyond that, though! It's expected that they will look nasty, as they come early in the issue processing pipeline.

dkhalanskyjb (changes requested):
The overall approach looks solid.
An important improvement would be to move these tests to the most general group of targets that they affect. In case of these tests, that means `common/test/`, since they are all completely deterministic and don't rely on data races. Then, if some target *does* accidentally behave as we want it to, we'll learn about it, and that's an important data point.

kotlinx-coroutines-core/common/test/pinnedBugs/Gh4382PinnedBugTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        val deferred = GlobalScope.async(dispatcher) {
+            hangingTest()
+        }
+        repeat(attempts) {
+            yield()
+            if (deferred.isCompleted) {
+                fail("Expected a hanging test, got ${deferred.await()}")
+            }
+        }
+    }
```
dkhalanskyjb: I'd rather expect this `spinAwaitingCompletion` to end up somewhere in `test-utils` and get used from the pinned tests. Is there a reason to duplicate it in several files?

dkhalanskyjb (commented):
This looks very reasonable, thanks!

kotlinx-coroutines-core/common/test/pinnedBugs/Gh4382PinnedBugTest.kt
vsartiushkin: The only reason is to not leave any other traces connected to these tests, so when the tests become outdated the cleanup will be trivial and no loose ends will be left.
If storing the `spinAwaitingCompletion` in `test-utils` is acceptable - I will move it there.

dkhalanskyjb: Sure, please go ahead.

dkhalanskyjb (approved):

PR #4757, opened 2026-09-14
Enable configuration cache with a small Gradle parallelism level

qwwdfsad:


murfel (approved):
Updated your comment, please check if you are happy with it!

murfel:
The CI did fail once out of 4 runs (two good here, one bad from my commit, one more good when you merged to develop), so possibly we will need to  set workers.max to 1 after all soonish.

murfel:
@qwwdfsad suggests tackling flaking tests one by one, as in #4760, as they appear, to see if we can manage with the workers.max count set to 2.

PR #4760, opened 2026-09-16
Make testListingActiveCoroutinesOnTimeout deterministic

qwwdfsad:
Use an unconfined dispatcher to step into the test body immediately, so first it launches all coroutines, then suspends, and only then the timeout timer is armed


murfel (approved):
Please check if you are happy with the new comment!

kotlinx-coroutines-test/common/test/RunTestTest.kt
```diff
@@ -112,7 +113,10 @@ class RunTestTest {
                 assertFalse((e.message ?: "").contains(name2))
             }
         }) {
-            runTest(timeout = 10.milliseconds) {
+            // Supplying UnconfinedTestDispatcher effectively ignores yield in the runTest implementation,
+            // so that the body is executed as much as possible before the timeout is set up.
```
nikolaykasyanov: `so that as much of the body as possible is executed before...` wdyt?

murfel:
```suggestion
            // so that as much of the body as possible is executed before the timeout is set up.
```

murfel: Thank you! Feel free to use the github's interface to suggest a change: click the file-diff +- document icon in the comment editor. See example below. This way it would take me one second to accept your change from the github UI!

If you are commenting from the Intellij interface to github, here's the markdown syntax:

````
```suggestion
replacement line here
```
````

PR #4761, opened 2026-09-16
Make WithTimeout*Test deterministic

qwwdfsad:
By using virtual time instead of real one


qwwdfsad:
I think after this one, we can try https://github.com/Kotlin/kotlinx.coroutines/pull/4721/changes/587ea44c8761481c5635e10014434bf66273c441 and actually increase the parallelism to see how it goes

dkhalanskyjb:
I'm not a fan of not having any tests that check the real-time behaviour on `delay` or `withTimeout`.
WDYT about having a separate test run dealing with very large durations, like `10.seconds`? If we enter the paradise of parallelised builds, then it won't take any extra time.

qwwdfsad:
@dkhalanskyjb the rest of the tests are still using the real time. I've only changed ones that are not deterministic

murfel:
Would it make sense to apply [587ea44](https://github.com/Kotlin/kotlinx.coroutines/commit/587ea44c8761481c5635e10014434bf66273c441) and increase the parallelism on top of this branch? And if there are more non-deterministic withtimeout tests discovered, then fix them in one go on this branch, too?

qwwdfsad:
@murfel I would prefer to go one by one with useful, isolated changes. I don't expect applying the patch, reviewing it and increasing parallelism to be pain- or problem- free, but it doesn't affect the current PR a bit, so a follow-up is much more preferable

